PR #5, opened 2020-08-12
implemented setting default cpu requests if none are set

meik99:


pkg/controller/builder/pod_builder.go
```diff
@@ -27,6 +28,13 @@ func BuildActiveGatePodSpecs(
 		}
 	}
 
+	if acitveGatePodSpec.Resources.Requests == nil {
+		acitveGatePodSpec.Resources.Requests = corev1.ResourceList{}
+	}
+	if _, hasCPUResource := acitveGatePodSpec.Resources.Requests[corev1.ResourceCPU]; !hasCPUResource {
+		acitveGatePodSpec.Resources.Requests[corev1.ResourceCPU] = *resource.NewScaledQuantity(1, -1)
```
DTMad, line 35: What exactly does this "NewScaledQuantity do with those parameters?

PR #6, opened 2020-09-02
Use containers/image library

meik99:


lrgar:
Added the "containers_image_storage_stub" tag to the Go commands. This disables the storage transport that was causing the broken builds since TravisCI VMs don't seem to have the devmapper module.

Skopeo does the same thing.

pkg/controller/version/docker.go
```diff
@@ -24,62 +27,87 @@ func NewDockerVersionChecker(currentImage, currentImageId string, dockerConfig *
...
+	return currentDigest.Algorithm().String() == latestDigest.Algorithm().String() &&
+		currentDigest.Hex() == latestDigest.Hex(), nil
 }
 
-func getDigest(reference name.Reference, authOption remote.Option) (string, error) {
-	img, err := remote.Image(reference, authOption)
+func (dockerVersionChecker *DockerVersionChecker) getDigest(ref types.ImageReference) (*digest.Digest, error) {
+	systemContext := dockerVersionChecker.makeSystemContext(ref.DockerReference())
+	imageSource, err := ref.NewImageSource(context.TODO(), systemContext)
+	defer closeImageSource(imageSource)
```
lrgar, line 68: Can `imageSource` be set when there is an error? Otherwise makes sense to defer the `imageSource.Close()` directly after the `if`.

pkg/controller/version/docker_integration_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	"github.com/Dynatrace/dynatrace-activegate-operator/pkg/controller/parser"
+	"github.com/stretchr/testify/assert"
+	"os"
+	"testing"
+)
+
+func TestDockerVersionChecker_DockerHub(t *testing.T) {
+	alpineTagNew := "alpine:3.12.0"
```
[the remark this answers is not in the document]
meik99: In those tests, the images are not tasted against e.g. "latest" but against an already known digest. I.e. the test here is to make sure that for a different tag, a different digest is returned and not that for ":<version>" the  same digest is returned as for ":latest"

The variable names have been changed to make this clearer

pkg/controller/version/docker_integration_test.go
```diff
@@ -0,0 +1,130 @@
+//+build integration
```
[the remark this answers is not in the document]
meik99: Should I delete or deactivate them then?

pkg/controller/version/docker.go
```diff
@@ -24,62 +27,85 @@ func NewDockerVersionChecker(currentImage, currentImageId string, dockerConfig *
...
+		return &types.SystemContext{}
 	}
 
-	credentials, hasCredentials := dockerConfig.Auths[registryURL]
+	registryName := strings.Split(dockerReference.Name(), "/")[0]
+	credentials, hasCredentials := dockerVersionChecker.dockerConfig.Auths[registryName]
+
 	if !hasCredentials {
-		return remote.WithAuthFromKeychain(authn.DefaultKeychain)
+		registryURL := "https://" + registryName
```
lrgar, line 98: To reiterate my previous comment, I don't have the protocol `https://` on the Dockerconfig from my local machine, are you certain you'll see it on Auths's map?

pkg/controller/version/docker_integration_test.go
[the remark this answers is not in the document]
lrgar: Ah, I see now.

pkg/controller/version/docker.go
```diff
@@ -24,62 +27,87 @@ func NewDockerVersionChecker(currentImage, currentImageId string, dockerConfig *
...
 
-	return remote.WithAuth(authn.FromConfig(authn.AuthConfig{
-		Username: credentials.Username,
-		Password: credentials.Password,
-	}))
+	return &types.SystemContext{
+		DockerAuthConfig: &types.DockerAuthConfig{
+			Username: credentials.Username,
+			Password: credentials.Password,
+		}}
```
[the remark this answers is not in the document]
lrgar: As spoken offline, IdentityToken doesn't seem to be used for Kubernetes environments. And after I look at this somewhat more, Auth was the one I've seen used before, not IdentityToken, sorry. But as mentioned, only user/password pairs are used on Kubernetes.

pkg/controller/version/docker_integration_test.go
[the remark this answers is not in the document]
lrgar: OK, let's leave them here for now until we implement the corresponding E2E tests.

pkg/controller/version/docker.go
meik99: I have added a check so both variants are covered:
```
registryName := strings.Split(dockerReference.Name(), "/")[0]
credentials, hasCredentials := dockerVersionChecker.dockerConfig.Auths[registryName]
 ```
checks without protocol. If it does not find it 
```
registryURL := "https://" + registryName
credentials, hasCredentials = dockerVersionChecker.dockerConfig.Auths[registryURL]
```
checks with prepended protocol

Comment (lrgar) [reply pkg/controller/version/docker.go:98]: Ah, missed that, sorry. All good then.

**Review by lrgar — DISMISSED** [review]

File: pkg/controller/version/docker.go [hunk 6]
```diff
@@ -24,62 +27,85 @@ func NewDockerVersionChecker(currentImage, currentImageId string, dockerConfig *
...
 		return false, err
 	}
 
-	registryURL := "https://" + reference.Context().RegistryStr()
-	authOption := getAuthOption(dockerVersionChecker.dockerConfig, registryURL)
-
-	latestDigest, err := getDigest(latestReference, authOption)
+	//Using ImageID instead of Image because ImageID contains digest of image that is used while Image only contains tag
+	currentReference, err := alltransports.ParseImageName(transportImageName)
+	//reference, err := name.ParseReference(strings.TrimPrefix(dockerVersionChecker.currentImageId, "docker-pullable://"))
```
DTMad, line 48: Shouldn't this line be used?

meik99: Switched to `alltransports.ParseImageName(...)` because Skopeo uses the same call for it's `inspect` command

DTMad (approved):

PR #7, opened 2020-09-28
Update deployment manifest and add linting

DTMad:


deploy/openshift/clusterrole-monitoring-patch.yaml
```diff
@@ -0,0 +1,11 @@
+- op: add
```
lrgar, line 1: We don't need this patch by the way, the default SCC set by OpenShift (`restricted`) is fine.

DTMad: Removed

lrgar (approved):

PR #8, opened 2020-10-05
Extend unit tests

meik99:


DTMad (approved):

pkg/controller/activegate/activegate_update.go
```diff
@@ -14,15 +15,33 @@ import (
...
 	secret *corev1.Secret) (*reconcile.Result, error) {
-	if !instance.Spec.DisableActivegateUpdate &&
+	if instance == nil {
+		return nil, fmt.Errorf("instance is nil")
+	} else if !instance.Spec.DisableActivegateUpdate &&
 		instance.Status.UpdatedTimestamp.Add(UpdateInterval).Before(time.Now()) {
 		log.Info("checking for outdated pods")
 		// Check if pods have latest activegate version
-		outdatedPods, err := r.findOutdatedPods(log, instance, isLatest)
+		outdatedPods, err := r.updateService.FindOutdatedPods(r, log, instance)
```
lrgar, line 44: Can't it call itself?

```suggestion
		outdatedPods, err := updateService.FindOutdatedPods(r, log, instance)
```

pkg/controller/activegate/activegate_update.go
```diff
@@ -14,15 +15,33 @@ import (
...
+	UpdatePods(
+		r *ReconcileActiveGate,
+		pod *corev1.Pod,
+		instance *dynatracev1alpha1.ActiveGate,
+		secret *corev1.Secret) (*reconcile.Result, error)
+}
+
+type activeGateUpdateService struct{}
+
+func (updateService *activeGateUpdateService) UpdatePods(
```
lrgar, line 33: Suggest shorter name, e.g., `us`

pkg/controller/activegate/activegate_update.go
```diff
@@ -14,15 +15,33 @@ import (
...
+		instance *dynatracev1alpha1.ActiveGate) ([]corev1.Pod, error)
+	IsLatest(logger logr.Logger, image string, imageID string, imagePullSecret *corev1.Secret) (bool, error)
+	UpdatePods(
+		r *ReconcileActiveGate,
+		pod *corev1.Pod,
+		instance *dynatracev1alpha1.ActiveGate,
+		secret *corev1.Secret) (*reconcile.Result, error)
+}
+
+type activeGateUpdateService struct{}
```
lrgar, line 31: I don't mind an extra object for a particular purpose, but two objects querying each other, e.g., the reconciler calling the service, and then service calling the reconciler is confusing.

I'd suggest either leaving them on the same type, or focusing the service to avoid contacting the reconciler.

pkg/controller/activegate/activegate_update.go
```diff
@@ -66,7 +85,7 @@ func (r *ReconcileActiveGate) findOutdatedPods(
 				logger.Error(err, err.Error())
 			}
 
-			isLatest, err := isLatestFn(logger, instance.Spec.Image, status.ImageID, imagePullSecret)
+			isLatest, err := r.updateService.IsLatest(logger, instance.Spec.Image, status.ImageID, imagePullSecret)
```
lrgar, line 88:
```suggestion
			isLatest, err := updateService.IsLatest(logger, instance.Spec.Image, status.ImageID, imagePullSecret)
```

pkg/controller/activegate/activegate_update.go
```diff
@@ -41,17 +60,17 @@ func (r *ReconcileActiveGate) updatePods(
...
-	isLatestFn func(logger logr.Logger, image string, imageID string, imagePullSecret *corev1.Secret) (bool, error)) ([]corev1.Pod, error) {
+	instance *dynatracev1alpha1.ActiveGate) ([]corev1.Pod, error) {
 	pods, err := r.findPods(instance)
 	if err != nil {
 		logger.Error(err, "failed to list pods")
 		return nil, err
 	}
 
-	var outdatedPods []corev1.Pod
+	outdatedPods := make([]corev1.Pod, 0)
```
lrgar, line 73: Was this a warning from some tool? My understanding is that the former is [preferred](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices).

pkg/controller/activegate/activegate_update_test.go
```diff
@@ -21,33 +22,183 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	mockActivegateVersion = "1.200.0"
```
lrgar, line 26:
```suggestion
	mockActiveGateVersion = "1.200.0"
```

pkg/controller/activegate/activegate_update_test.go
```diff
@@ -21,33 +22,183 @@ import (
...
+)
+
 func init() {
 	_ = apis.AddToScheme(scheme.Scheme) // Register OneAgent and Istio object schemas.
 	_ = os.Setenv(k8sutil.WatchNamespaceEnvVar, _const.DynatraceNamespace)
 }
 
+type mockIsLatestUpdateService struct{}
```
lrgar, line 34: Alternatively you can use a mocking framework, we use one on `dtclient/mock_client.go`

pkg/controller/builder/pod_builder_test.go
```diff
@@ -0,0 +1,119 @@
+package builder
+
+import (
+	"testing"
+
+	"github.com/Dynatrace/dynatrace-activegate-operator/pkg/apis/dynatrace/v1alpha1"
```
lrgar, line 6:
```suggestion
	dynatracev1alpha1 "github.com/Dynatrace/dynatrace-activegate-operator/pkg/apis/dynatrace/v1alpha1"
```

pkg/controller/version/docker_test.go
```diff
@@ -0,0 +1,53 @@
...
+	assert.Equal(t, types.SystemContext{}, *noAuth)
+
+	type auth struct {
+		Username string
+		Password string
+	}
+	auths := make(map[string]struct {
+		Username string
+		Password string
+	})
```
lrgar, line 34:
```suggestion
	auths := make(map[string]auth)
```

pkg/controller/version/docker_test.go
```diff
@@ -0,0 +1,53 @@
...
+	auths["localhost"] = auth{
+		Username: "username",
+		Password: "password",
+	}
+	versionChecker.dockerConfig = &parser.DockerConfig{
+		Auths: auths}
+	withAuth := versionChecker.makeSystemContext(reference.DockerReference())
+	assert.Equal(t, withAuth.DockerAuthConfig.Username, "username")
+	assert.Equal(t, withAuth.DockerAuthConfig.Password, "password")
+}
```
lrgar, line 53: Suggest using single lines for most of these, e.g.,

```
auths["localhost.com"] = auth{Username: "username", Password: "password"}
versionChecker.dockerConfig = &parser.DockerConfig{Auths: auths}
```

pkg/controller/parser/docker_config_parser_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestNewDockerConfig(t *testing.T) {
+	t.Run("NewDockerConfig", func(t *testing.T) {
+		auths := make(map[string]struct {
```
lrgar, line 13: In another class, a type `auth` is used over `struct {Username string, Password string}`.

pkg/controller/parser/token_parser.go
```diff
@@ -23,15 +27,9 @@ func NewTokens(secret *corev1.Secret) (*Tokens, error) {
...
-	apiToken, err = ExtractToken(secret, _const.DynatraceApiToken)
-	if err != nil {
-		return nil, err
-	}
-
-	paasToken, err = ExtractToken(secret, _const.DynatracePaasToken)
-	if err != nil {
-		return nil, err
-	}
+	//Errors would have been caught by verifySecret
```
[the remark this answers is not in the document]
meik99: `verifySecret` already calls `ExtractToken`. If it would throw an error in lines 31 or 32, it would have already thrown one in 26. Calling `ExtractToken` twice for each token is done so I can catch any error in a single line instead of catching it every call.

pkg/controller/version/docker_test.go
meik99: This results in the following error message when compiling:
`cannot use auths (type map[string]auth) as type map[string]struct { Username string; Password string } in field value`

pkg/controller/activegate/activegate_update.go
meik99: The reconciler could have a different service than the calling one

pkg/controller/activegate/activegate_update.go
meik99: The reconciler could have a different service than the calling one

pkg/dtclient/dynatrace_client_test.go
```diff
@@ -234,3 +234,15 @@ func TestIgnoreNonCurrentlySeenHosts(t *testing.T) {
 	require.Equal(t, "HOST-42", info.entityID)
 	require.Equal(t, "1.195.0.20200515-045253", info.version)
 }
+
+func createTestDynatraceClient(t *testing.T, handler http.HandlerFunc) (*httptest.Server, Client) {
```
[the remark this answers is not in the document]
meik99: Why is `Handler` preferred here instead of  `HandlerFunc`?

pkg/controller/version/docker_test.go
meik99: I created a new type `DockerConfigAuth` to simplify those calls

DTMad (approved):

pkg/controller/activegate/activegate_update.go
lrgar: So that `activeGateUpdateService.UpdatePods()` can call `FindOutdatedPods()` from a `updateService` that is not itself? Is this done for testing?

pkg/controller/activegate/activegate_update_test.go
lrgar: I don't mind leaving it for follow-up change.

pkg/dtclient/activegate_query.go
```diff
@@ -65,26 +64,27 @@ func (dtc *dynatraceClient) QueryOutdatedActiveGates(query *ActiveGateQuery) ([]
...
-		params += "networkZone=" + query.NetworkZone + "&"
-	}
-	if query.NetworkAddress != "" {
-		params += "networkAddress=" + query.NetworkAddress + "&"
-	}
-	if query.UpdateStatus != "" {
-		params += "updateStatus=" + query.UpdateStatus + "&"
+func (query *ActiveGateQuery) buildQueryParams() string {
+	values := url.Values{}
+	if query != nil {
```
lrgar, line 69: Can it still be nil? (I mean, you can create a nil ActiveGateQuery instance and call its method, but in practice is harder.)

pkg/controller/parser/token_parser.go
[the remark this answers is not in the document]
lrgar: I see. Fair enough.

lrgar (approved):

pkg/dtclient/activegate_query.go
meik99: Regarding the update service:

It would be possible to create a service A that calls `FindOutdatedPods` with a `reconciler` that uses a service B
It doesn't happen in code, but I do like the option that this can be done.
If it causes more problems than it solves I have nothing against removing that, i.e., to let the `updateService` call itself

meik99: Regarding activegate query:

The `dynatraceClient.QueryActiveGates` call takes a pointer to the query, and handing it nil would result in only the default parameters being used, which imo is expected. 
Otherwise a nil check has to either be introduced in that call or everytime before calling `dynatraceClient.QueryActiveGates`

PR #11, opened 2020-10-13
added crd for type connectionconfiguration

meik99:


DTMad (commented):
LGTM - however please check if adaptions to the roles or clusterroles are necessary

PR #16, opened 2020-10-19
rename activegate to dynakube

meik99:
renamed activegate-crd to dynakube-crd
added kubernetesMonitoring property and moved properties accordingly


deploy/cr.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: dynatrace.com/v1alpha1
```
DTMad, line 1: Please move the fields from the ConnectionInformation in here too

DTMad (dismissed):

pkg/apis/dynatrace/v1alpha1/dynakube_types.go
```diff
@@ -0,0 +1,127 @@
...
+// +k8s:openapi-gen=true
+type DynaKubeSpec struct {
+	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+	BaseDynaKubeSpec `json:""`
+
+	OneAgentAPMSpec OneAgentAPMSpec `json:"oneAgentApm,omitempty"`
+
+	OneAgent OneAgentSpec `json:"oneAgent,omitempty"`
+
+	// Enables Kubernetes Monitoring"
```
lrgar, line 18: Extra quotes at the end.

pkg/apis/dynatrace/v1alpha1/dynakube_types.go
```diff
@@ -0,0 +1,127 @@
...
+
+// DynaKubeSpec defines the desired state of DynaKube
+// +k8s:openapi-gen=true
+type DynaKubeSpec struct {
+	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+	BaseDynaKubeSpec `json:""`
+
+	OneAgentAPMSpec OneAgentAPMSpec `json:"oneAgentApm,omitempty"`
+
+	OneAgent OneAgentSpec `json:"oneAgent,omitempty"`
```
lrgar, line 16: We don't need these, OneAgent features won't be supported on the DynaKube CRD anytime soon (or ever.) Remove the types as well.

pkg/apis/dynatrace/v1alpha1/dynakube_types.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"github.com/operator-framework/operator-sdk/pkg/status"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// DynaKubeSpec defines the desired state of DynaKube
+// +k8s:openapi-gen=true
+type DynaKubeSpec struct {
+	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+	BaseDynaKubeSpec `json:""`
```
lrgar, line 12: No need for an extra type since it isn't being inherited by anyone else.

The OneAgent CRDs may eventually, but let's leave that problem for later.

pkg/controller/builder/pod_builder.go
```diff
@@ -10,13 +10,13 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-func BuildActiveGatePodSpecs(instance *v1alpha1.ActiveGate, tenantInfo *dtclient.TenantInfo, kubeSystemUID types.UID) (corev1.PodSpec, error) {
-	sa := MonitoringServiceAccount
+func BuildActiveGatePodSpecs(instance *v1alpha1.DynaKube, tenantInfo *dtclient.TenantInfo, kubeSystemUID types.UID) (corev1.PodSpec, error) {
+	serviceAccount := MonitoringServiceAccount
```
lrgar, line 14: There is a pseudo-convention in Go to use shorter names for local variables if they don't really matter. To be fair the guideline is quite arbitrary, but I agree that it makes things easier to read, since there is less noise.

Global variables can have full names.

lrgar (changes requested):

deploy/cr.yaml
```diff
@@ -1,44 +1,95 @@
...
+
+  # Optional: Sets Network Zone for OneAgent and ActiveGate pods
+  #
+  # networkZone: name-of-my-network-zone
+
+  # Enables and configures an ActiveGate instance that allows monitoring
+  # of Kubernetes environments
+  #
+  kubernetesMonitoring:
+  #   Enable Kubernetes monitoring functionality
```
lrgar, line 37: I suggest increasing the indentation for this fields to make it easier to read.

lrgar (approved):

DTMad (approved):

PR #17, opened 2020-10-27
extending properties for Kubernetes monitoring

meik99:


deploy/cr.yaml
```diff
@@ -81,38 +81,37 @@ spec:
...
+  #  args: []
 
   #   Optional: additional environment variables to add to the ActiveGate Pods.
   #
   #   env: []
 
   #   Optional: Set activation group for ActiveGate
   #
-  #   group: ""
+  #  group: "mygroup"
```
DTMad, line 109: Please don't put anything here - there's a default value called "default" which gets automatically applied in case nothing is set.

DTMad (changes requested):

deploy/crds/dynatrace.com_dynakubes_crd.yaml
```diff
@@ -68,6 +83,112 @@ spec:
...
+                            type: object
+                          secretKeyRef:
+                            description: Selects a key of a secret in the pod's namespace
+                            properties:
+                              key:
+                                description: The key of the secret to select from.  Must
+                                  be a valid secret key.
+                                type: string
+                              name:
+                                description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
```
DTMad, line 173: typo? referent

pkg/controller/builder/pod_builder_test.go
```diff
@@ -34,9 +34,18 @@ func TestBuildActiveGatePodSpecs(t *testing.T) {
 		assert.NotNil(t, specs)
 		assert.Equal(t, 1, len(specs.Containers))
 		assert.Equal(t, serviceAccountName, specs.ServiceAccountName)
-		assert.Equal(t,
-			*resource.NewScaledQuantity(1, -1),
-			activeGateSpec.KubernetesMonitoringSpec.Resources.Requests[corev1.ResourceCPU])
+		assert.True(t,
+			activeGateSpec.KubernetesMonitoringSpec.Resources.Requests[corev1.ResourceCPU].Equal(
+				*resource.NewScaledQuantity(1, -1)))
```
DTMad, line 39: You can use the predefined type "resource" here - see
https://github.com/Dynatrace/dynatrace-oneagent-operator/commit/faee78b07dc8ffe124e793a73699150313900835#diff-6404509afd56daa425812290afd70394d934e45a776dcdb36087716a660c113aL204

DTMad (changes requested):

deploy/crds/dynatrace.com_dynakubes_crd.yaml
meik99: That origins from the documentation of the `LocalObjectReference` struct provided by the Kubernetes api. 
```
// LocalObjectReference contains enough information to let you locate the
// referenced object inside the same namespace.
type LocalObjectReference struct {
	// Name of the referent.
	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
	// TODO: Add other useful fields. apiVersion, kind, uid?
	// +optional
	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
}
```

pkg/controller/activegate/activegate_create.go
```diff
@@ -11,18 +12,56 @@ import (
...
+		configMap := &corev1.ConfigMap{}
+		err := r.client.Get(context.TODO(), client.ObjectKey{
+			Namespace: _const.DynatraceNamespace,
+			Name:      _const.CustomPropertiesConfigMapName,
+		}, configMap)
+		if err != nil {
+			if k8serrors.IsNotFound(err) {
+				configMap = &corev1.ConfigMap{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      _const.CustomPropertiesConfigMapName,
```
lrgar, line 36: You can have more than one DynaKube object, requiring more than one config map. You'd also need to detect changes if the field on DynaKube changes.

lrgar (approved):

DTMad (approved):

lrgar (approved):

PR #21, opened 2020-11-03
update with labels instead of hashes

meik99:


.gitignore
```diff
@@ -79,3 +79,4 @@ tags
 /debug.sh
 /tmp/
 /reapply-all.sh
+
```
DTMad, line 82: Newline?

DTMad (dismissed):

.gitignore
meik99: lines should end in newlines / linebreaks according to [POSIX convention](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)

pkg/controller/activegate/version_label.go
```diff
@@ -0,0 +1,48 @@
...
+			imagePullSecret, err := dao.GetImagePullSecret(r.client, pod)
+			if err != nil {
+				// Something wrong with pull secret, exit function entirely
+				return err
+			}
+
+			dockerConfig, err := parser.NewDockerConfig(imagePullSecret)
+			// If an error is returned, try getting the image anyway
+
+			versionLabel, err2 := version.GetVersionLabel(status.Image, dockerConfig)
```
lrgar, line 30: Do you actually get something on dockerConfig is there was an error? Or just using nil there?

pkg/controller/activegate/activegate_controller.go
```diff
@@ -142,12 +142,25 @@ func (r *ReconcileActiveGate) Reconcile(request reconcile.Request) (reconcile.Re
...
+	}
+
+	pods, err := r.findPods(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	err = r.setVersionLabel(pods)
+	if err != nil {
+		if _, isStatusError := err.(*errors.StatusError); isStatusError {
```
lrgar, line 157: For reference there is `errors.As()` on the Go library. Although probably better to do this check before creating the StateFul set?

pkg/controller/activegate/version_label.go
meik99: On error, `dockerConfig` is `nil`

lrgar (approved):

PR #24, opened 2020-11-05
Make install.sh run in Bourne Shell

baichinger:


DTMad (approved):

deploy/install.sh
```diff
@@ -170,18 +161,26 @@ addK8sConfiguration() {
 EOF
   )
 
-  curl -sS -X POST "${API_URL}/config/v1/kubernetes/credentials" \
+  response=$(curl -sS -X POST "${API_URL}/config/v1/kubernetes/credentials" \
```
arthfl, line 164: `$()` should be quoted :-)

arthfl (changes requested):
Also `json=$(...)` should be quoted :-)

deploy/install.sh
```diff
@@ -45,26 +45,27 @@ for arg in "$@"; do
...
 fi
 
 set -u
 
 applyOneAgentOperator() {
-  if ! "${CLI}" get ns dynatrace &>/dev/null; then
-    if [[ "${CLI}" == "kubectl" ]]; then
+  "${CLI}" get ns dynatrace >/dev/null 2>&1
+  if [ $? -ne 0 ]; then
+    if [ "${CLI}" == "kubectl" ]; then
```
arthfl, line 68: `shellcheck` complains here:
```text
   if [ "${CLI}" == "kubectl" ]; then
                  ^-- SC2039: In POSIX sh, == in place of = is undefined.
```

arthfl (changes requested):

DTMad (dismissed):

DTMad (approved):

PR #25, opened 2020-11-16
Fix dynatrace-operator functionality

DTMad:


pkg/controller/factory/secret_creation.go
```diff
@@ -9,11 +9,14 @@ import (
...
+	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-// CreateOrUpdateSecretIfNotExists creates a secret in case it does not exist or updates it if there are changes
-func CreateOrUpdateSecretWithAPIReader(c client.Client, r client.Reader, secretName string, targetNS string, data map[string][]byte, secretType corev1.SecretType, log logr.Logger) error {
+// CreateOrUpdateSecretWithAPIReader creates a secret in case it does not exist or updates it if there are changes
+func CreateOrUpdateSecretWithAPIReader(c client.Client, r client.Reader, scheme *runtime.Scheme, owner metav1.Object, targetNS string, data map[string][]byte, secretType corev1.SecretType, log logr.Logger) error {
+	secretName := owner.GetName() + "-pull-secret"
```
lrgar, line 19: I'll presume this is the only one calling this function, but otherwise you'd want to send the name as a parameter. If that isn't the case, good for now.

lrgar (approved):

pkg/controller/factory/secret_creation.go
```diff
@@ -46,6 +55,6 @@ func CreateOrUpdateSecretWithAPIReader(c client.Client, r client.Reader, secretN
 	return nil
 }
 
-func CreateOrUpdateSecret(c client.Client, secretName string, targetNamespace string, data map[string][]byte, secretType corev1.SecretType, log logr.Logger) error {
-	return CreateOrUpdateSecretWithAPIReader(c, c, secretName, targetNamespace, data, secretType, log)
+func CreateOrUpdateSecret(c client.Client, r client.Reader, scheme *runtime.Scheme, owner metav1.Object, targetNS string, data map[string][]byte, secretType corev1.SecretType, log logr.Logger) error {
+	return CreateOrUpdateSecretWithAPIReader(c, r, scheme, owner, targetNS, data, secretType, log)
```
meik99, line 59: This function should use `c` only and delegate to `CreateOrUpdateSecretWithAPIReader`  by calling it with the client for both, reader and client. If this feature is not needed anymore, feel free to delete this function
```suggestion
func CreateOrUpdateSecret(c client.Client, scheme *runtime.Scheme, owner metav1.Object, targetNS string, data map[string][]byte, secretType corev1.SecretType, log logr.Logger) error {
	return CreateOrUpdateSecretWithAPIReader(c, c, scheme, owner, targetNS, data, secretType, log)
```

pkg/controller/factory/secret_creation.go
```diff
@@ -9,11 +9,14 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-// CreateOrUpdateSecretIfNotExists creates a secret in case it does not exist or updates it if there are changes
-func CreateOrUpdateSecretWithAPIReader(c client.Client, r client.Reader, secretName string, targetNS string, data map[string][]byte, secretType corev1.SecretType, log logr.Logger) error {
+// CreateOrUpdateSecretWithAPIReader creates a secret in case it does not exist or updates it if there are changes
+func CreateOrUpdateSecretWithAPIReader(c client.Client, r client.Reader, scheme *runtime.Scheme, owner metav1.Object, targetNS string, data map[string][]byte, secretType corev1.SecretType, log logr.Logger) error {
```
meik99, line 18: With so many arguments that have to be passed, introducing a struct to group them together would be better imo

meik99 (changes requested):

meik99 (approved):

lrgar (approved):

PR #43, opened 2020-12-03
Recognize upgrades for ActiveGate instances with image hashes

lrgar:
On this PR, the Operator will use image hashes to identify when there is a new ActiveGate version available.

Also, the update rollout logic has been updated considerably - rather than looking at individual Pods and restarting them, the Operator will set an annotation on the PodSpec of the StatefulSet, which will cause for the Pods to be rolled out automatically by Kubernetes.

Other changes:
* Remove ability to disable auto-updates for the time being on the CR, as well as to set an ActiveGate version for the time being. The tag can be added to the image field as an alternative.
* Auto-updates can be disabled by setting the environment variable  `OPERATOR_DEBUG_DISABLE_UPDATES` to `true` on the Operator Deployment.
* The ActiveGate image can be set through `.spec.activeGate.image` field now.
* Cleaned up the DynaKube CR example quite a bit.
* The Operator will update the status section only if there were changes to the StatefulSet, or there was an update.
* StatefulSet diffing now works.

Given the simplified logic, I've deleted everything else that wasn't used.


config/samples/cr.yaml
```diff
@@ -30,90 +30,83 @@ spec:
...
+    # See more here: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/
+    #
+    # priorityClassName: PRIORITYCLASS
 
-  #   Optional: arguments to add to the ActiveGate instances.
-  #   Available options: https://www.dynatrace.com/support/help/shortlink/linux-custom-installation
-  #   Limitations: https://www.dynatrace.com/support/help/shortlink/oneagent-docker#limitations
-  #
-  #  args: []
+    # Optional: resource settings for OneAgent container. Consumption of the OneAgent heavily depends
```
DTMad, line 75:
```suggestion
    # Optional: resource settings for ActiveGate container. Consumption of the ActiveGate heavily depends
```

config/samples/cr.yaml
```diff
@@ -30,90 +30,83 @@ spec:
...
+    #
+    # resources:
+    #   requests:
+    #     cpu: 100m
+    #     memory: 512Mi
+    #   limits:
+    #     cpu: 300m
+    #     memory: 1.5Gi
+
+    # Optional: labels are customer defined labels for OneAgent Pods to structure workloads as desired.
```
DTMad, line 86:
```suggestion
    # Optional: labels are customer defined labels for ActiveGate Pods to structure workloads as desired.
```

DTMad (approved):

PR #47, opened 2020-12-04
Don't include custom labels on selectors

lrgar:
Once the StatefulSet has been created, selector labels can't be updated anymore, so let's not include the custom ones there. We don't need them there anyway.


controllers/kubemon/statefulset.go
```diff
@@ -52,7 +52,9 @@ func newStatefulSet(instance *dynatracev1alpha1.DynaKube, kubeSystemUID types.UI
 		Spec: appsv1.StatefulSetSpec{
 			Replicas:            instance.Spec.KubernetesMonitoringSpec.Replicas,
 			PodManagementPolicy: appsv1.ParallelPodManagement,
-			Selector:            buildLabelSelector(instance),
+			Selector: &metav1.LabelSelector{
+				MatchLabels: BuildLabelsFromInstance(instance),
+			},
```
meik99, line 57:
```suggestion
			Selector: &metav1.LabelSelector{ MatchLabels: BuildLabelsFromInstance(instance) },
```

meik99 (approved):

DTMad (approved):

PR #52, opened 2020-12-22
better stacktraces

meik99:
adds a "prettyfier" to the default logger and wraps some errors with a stacktrace

### DIFF
--- a/controllers/activegate/token_parser.go
+++ b/controllers/activegate/token_parser.go
@@ -2,6 +2,7 @@ package activegate
 
 import (
 	"fmt"
+	"github.com/pkg/errors"
 	"strings"
 
 	dynatracev1alpha1 "github.com/Dynatrace/dynatrace-operator/api/v1alpha1"
@@ -24,7 +25,7 @@ func NewTokens(secret *corev1.Secret) (*Tokens, error) {
 	var err error
 
 	if err = verifySecret(secret); err != nil {
-		return nil, err
+		return nil, errors.WithStack(err)
 	}
 
 	//Errors would have been caught by verifySecret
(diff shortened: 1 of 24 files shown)


controllers/activegate/token_parser.go
```diff
@@ -43,7 +44,7 @@ func verifySecret(secret *corev1.Secret) error {
 		dtclient.DynatracePaasToken} {
 		_, err := ExtractToken(secret, token)
 		if err != nil {
-			return fmt.Errorf("invalid secret %s, %s", secret.Name, err)
+			return errors.Errorf("invalid secret %s, %s", secret.Name, err)
```
DTMad, line 47: Don't we need errors.WithStack here?

dtclient/client.go
```diff
@@ -3,11 +3,11 @@ package dtclient
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"errors"
 	"net/http"
 	"net/url"
 	"strings"
 
+	"github.com/pkg/errors"
```
DTMad, line 10: Do we need the import here?

controllers/activegate/token_parser.go
meik99: basically every method in the pkg/errors package adds a stacktrace [https://github.com/pkg/errors/blob/master/errors.go](https://github.com/pkg/errors/blob/master/errors.go)

dtclient/client.go
meik99: yes, there are two "errors" packages, "errors" and "pkg/errors". Since "errors", the package previously imported here, also includes a `.New` method, the lines where it was used are not shown in the changes.
E.g. see line 117 `return nil, errors.New("tokens are empty")`.

"errors" does not include stacktraces afaik, but "pkg/errors" does.

DTMad (approved):

PR #71, opened 2021-02-25
Set default for useUnprivilegedMode to true

DTMad:


lrgar (commented):
Just in case, there is a difference between, having the default as true in our examples and, if nothing is set, then the default is true.

Customers shouldn't need to set the DNS Policy.

lrgar (approved):

PR #73, opened 2021-02-26
Add CSI Driver

lrgar:
This PR includes an implementation for the CSI Driver,

* `cmd/csidriver/` - starting point for the CSI Driver binary. Runs a GRPC server handling requests from Kubelet, and a Provisioner reconciler in charge of downloading OneAgent packages.
* `controllers/csi/driver/` - includes the implementation for the GRPC server. It implements the interface required by Kubelet, although most of it is not used and not implemented. Logic happens in `NodePublishVolume()` and `NodeUnpublishVolume()`. The server is also implemented using `controller-runtime`'s framework.
* `controllers/csi/provisioner` - classic reconciler for DynaKube objects, in charge of downloading and unzipping the OneAgent package, as well as downloads.
* `dtclient/` - added a new method to download the OneAgent package.

The CSI Driver runs as a DaemonSet, with three containers.
1. There is the CSI Driver itself.
2. A Node Registrar in charge of registering the CSI Driver plugin with Kubelet.
3. A livenessprobe checking if things are running.

The former container is included on the Operator image, and the other two are provided by Kubernetes, and we're currently using those for the time being.

The CSI Driver shouldn't be added to OCP 3.11, so for now I've only added the manifests to run it with Kubernetes. In a future PR we can split OCP configuration for 3.11 and 4.x.


controllers/csi/driver/server.go
```diff
@@ -0,0 +1,327 @@
+/*
```
DTMad, line 1: Is there a reason we have several methods which seem to be auto-generated but unimplemented?

controllers/csi/driver/server.go
```diff
@@ -0,0 +1,327 @@
...
+	podUID := volCtx[podUIDContextKey]
+	if podUID == "" {
+		return nil, status.Error(codes.InvalidArgument, "No Pod UID included with request")
+	}
+
+	flavor := volCtx["flavor"]
+	if flavor == "" {
+		flavor = "default"
+	}
+	if flavor != "default" && flavor != "musl" {
```
DTMad, line 167: Maybe think about moving those flavours to consts since we reuse it a lot of times

controllers/csi/driver/server.go
lrgar: We need to fulfill an Go interface, even if we don't do anything. They are used for normal CSI drivers with PersistentVolumes. Our CSI driver supports only ephemeral volumes, so most of them aren't needed.

controllers/csi/driver/server.go
```diff
@@ -0,0 +1,327 @@
...
+			Target: filepath.Join(targetPath, "datastorage"),
+		},
+	); err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("failed to mount OneAgent volume: %s", err.Error()))
+	}
+
+	return &csi.NodePublishVolumeResponse{}, nil
+}
+
+func (svr *CSIDriverServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublishVolumeRequest) (*csi.NodeUnpublishVolumeResponse, error) {
```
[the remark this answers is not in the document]
lrgar: If no arguments is used, then I can remove all the names. But when only some are used you need to put a `_`. I think I'll leave them be instead.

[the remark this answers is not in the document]
DTMad: sounds good!

controllers/csi/driver/server.go
DTMad: OK

DTMad (approved):

PR #78, opened 2021-03-10
Change CSV generation to use make file instead of script

DTMad:


DTMad:
> There is a lot that the old script does that just `make bundle` doesn't.
> 
> We now have two different CSVs for each platform.
> 
> They are also versioned. I think `make bundle` just put the files in a given directory, without versions?

You are totally right!
Did overlook this. Now adapted the PR to also focus on the what we did before and also do it for k8s and ocp.

lrgar (approved):

PR #84, opened 2021-03-12
Enable k8s monitoring by default

DTMad:


deploy/install.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 CLI="kubectl"
-ENABLE_K8S_MONITORING="false"
+ENABLE_K8S_MONITORING="true"
```
baichinger, line 6: With this change K8s monitoring cannot be disabled anymore. Option `--enable-k8s-monitoring` can therefore be removed as well.

lrgar: Do we want to support disabling it?

baichinger (approved):

PR #85, opened 2021-03-16
Improvements for upgrades of immutable images

lrgar:
Logic:
* Similar to changes in https://github.com/Dynatrace/dynatrace-oneagent-operator/pull/376, I'm reusing the logic we had already to detect upgrades with ActiveGate images, to do the same with immutable OneAgent images.
  * Specifically, we keep on the DynaKube status section the last version we saw, and periodically check for the version label on the image-to-deploy, to see if there have been changes.
* ActiveGate upgrades can be disabled with the `OPERATOR_ACTIVEGATE_DISABLE_UPDATES` environment variable. The reconciler supported this behavior, but nothing to actually enable it.
* Do not stop the reconciler in case of UseImmutableImage changes, and also removing the cluster version checks. As described on the other PR, this may end up with the Operator not deploying anything under certain conditions.
* Phase: we migrated the phase-setting logic of the OneAgent Operator of being set when upgrading, or when there is a new deployment in progress (note that currently AG deployments are ignored for this field.) For the upgrade case, we were saving the DynaKube object before the rest of the reconciliation logic, which also saves the DynaKube object, causing an conflict error (hence discarding status changes at the end.) We must only save once by reconciliation, so I've removed this logic for upgrades. It's cosmetic anyway.

Code:
* `controllers/dynakube/updates/`: moved the upgrade logic for immutable images here.
* `helpers.go`: in an attempt to handle the increasing utilities, I've added this file to contain simple DynaKube-related methods. So if you have a function that depends solely on the DynaKube object, and has no arguments, just add it there. As it's now you need to guess which package has each utility, if any.
* `version/`: Removed code for versions that aren't used anymore (and that had another implementation in `controller/dtversion/`.)


DTMad (approved):
LGTM

PR #86, opened 2021-03-17
Enable routing and accept cluster name in install script

DTMad:


deploy/install.sh
```diff
@@ -50,6 +56,21 @@ if [ -z "$PAAS_TOKEN" ]; then
   exit 1
 fi
 
+if [ -z "$CLUSTER_NAME" ]; then
+  echo "Error: cluster name not set!"
+  exit 1
+fi
+
+if ! echo "$CLUSTER_NAME" | grep -q "$CLUSTER_NAME_REGEX"; then
```
lrgar, line 64: `grep` doesn't match regex by default, you want to use `-E` in Linux. No idea about Mac.

Bash also supports regexes, but you'd need to update the shebang.

deploy/install.sh
```diff
@@ -146,14 +171,28 @@ EOF
...
 
+checkForExistingCluster() {
+  response="$(curl -sS -X GET "${API_URL}/config/v1/kubernetes/credentials" \
+    -H "accept: application/json; charset=utf-8" \
+    -H "Authorization: Api-Token ${API_TOKEN}" \
+    -H "Content-Type: application/json; charset=utf-8")"
+
+  if echo "$response" | grep -Fqe "\"name\":\"${CLUSTER_NAME}\""; then
+    echo "Error: Cluster already exists!"
+    exit 1
```
lrgar, line 189: Why exit? Basically forces the customer to delete entry from Dynatrace to run the command twice, even if there is no change for the connection at all.

deploy/install.sh
```diff
@@ -146,14 +171,28 @@ EOF
...
   fi
 }
 
+checkForExistingCluster() {
+  response="$(curl -sS -X GET "${API_URL}/config/v1/kubernetes/credentials" \
+    -H "accept: application/json; charset=utf-8" \
+    -H "Authorization: Api-Token ${API_TOKEN}" \
+    -H "Content-Type: application/json; charset=utf-8")"
+
+  if echo "$response" | grep -Fqe "\"name\":\"${CLUSTER_NAME}\""; then
```
lrgar, line 187: I'm only accepting a customer argument to be used as a regex on this command because it's being validated before.

lrgar (changes requested):

deploy/install.sh
DTMad: Added -E

deploy/install.sh
DTMad: This is on purpose. We will not overwrite customer configuration. In case he provides the same name he needs to adapt

lrgar (approved):

PR #87, opened 2021-03-18
Feature/metadata for code modules

gkrenn:
Added env variable for oneagent to know clusterId and version

Currently One Agent Version can not be evaluated as the dynakube object does not have a oneagent object if in app-only mode. Should the object be created in the future, the version will be included in the env variable.

Already tested feature in own test cluster.


webhook/server/server.go
```diff
@@ -75,10 +77,16 @@ func registerInjectEndpoint(mgr manager.Manager, ns string, podName string) erro
 		return err
 	}
 
+	var clusterNS corev1.Namespace
+	if err := mgr.GetAPIReader().Get(context.TODO(), client.ObjectKey{Name: "kube-system"}, &clusterNS); err != nil {
+		return err
+	}
+
```
meik99, line 84: A function which returns the kube-system UID  already exists in controllers/kubesystem.
I suggest using it to avoid code duplication

meik99 (changes requested):

webhook/server/server.go
gkrenn: Ok, thx

meik99 (approved):

deployment_metadata/deployment_metadata.go
```diff
@@ -0,0 +1,64 @@
+package deployment_metadata
```
lrgar, line 1: Packages shouldn't have underscores on them according to [Go guidelines](https://golang.org/doc/effective_go#package-names).

deployment_metadata/deployment_metadata.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+type DeploymentMetadata struct {
+	OperatorScriptVersion string
+	OrchestratorID        string
+	ContainerImageVersion string
+	OrchestrationTech     string
+}
+
+func NewDeploymentMetadata(operatorScriptVersion string, orchestratorId string, containerImageVersion string) *DeploymentMetadata {
```
lrgar, line 25:
```suggestion
func NewDeploymentMetadata(operatorScriptVersion string, orchestratorID string, containerImageVersion string) *DeploymentMetadata {
```

Also, we don't need to provide the OneAgent version to our environment, the agents themselves send them already. And if, as I presume it's the case, if used for OneAgent DaemonSets, it can also interfere OneAgent updates where Pods may end up being restarted twice.

In fact I think I just ran out into this bug.

As for the script version, you can also just use version.Version directly.

webhook/server/server.go
```diff
@@ -75,10 +79,16 @@ func registerInjectEndpoint(mgr manager.Manager, ns string, podName string) erro
 		return err
 	}
 
+	var UID types.UID
+	if UID, err = kubesystem.GetUID(mgr.GetAPIReader()); err != nil {
```
lrgar, line 83:
```suggestion
	if UID, err = kubesystem.GetUID(mgr.GetClient()); err != nil {
```

In general prefer `.GetClient()` over `.GetAPIReader()` - the former has a cache and should be used when possible (i.e., basically all cases except when accessing objects in foreign namespaces.)

Not that important for this case in particular, since it's an one-time operation.

lrgar (changes requested):

lrgar:
Removed the OneAgent version from the metadata, also setting the static fields, Operator, and the Operator version, directly rather than having them configurable (since nothing needs it at the moment.)

DTMad (approved):

DTMad (approved):

lrgar (approved):

PR #88, opened 2021-03-18
Bugfix/message routing

gkrenn:
Because our selector for the dynatrace service was to general, traffic was probably routed to the wrong endpoint/msg-routing-pod if more than one dynakube object existed in the cluster


baichinger:
Will `.spec.selector.matchLabels` of the STS be updated in a separate PR?

PR #90, opened 2021-03-22
Service selector for routing matches all instances - new keys

gkrenn:
changed labels on the four locations again
- Label of sts
- match label of sts
- label of pod template
- selector of service

Example of labels: 
```
    dynatrace.com/component: activegate
    operator.dynatrace.com/feature: msgrouting
    operator.dynatrace.com/instance: dynakube
```

only for the stateful set as it is immutable, we need to destroy it by hand, if the labels are not correct

### DIFF
--- a/controllers/capability/reconciler.go
+++ b/controllers/capability/reconciler.go
@@ -20,6 +20,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+const (
+	OldKeyActiveGate = "activegate"
+)
+
 type Reconciler struct {
 	client.Client
 	Instance                         *v1alpha1.DynaKube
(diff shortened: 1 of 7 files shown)


DTMad (approved):

lrgar (approved):

controllers/capability/reconciler.go
```diff
@@ -159,6 +168,23 @@ func (r *Reconciler) updateStatefulSetIfOutdated(desiredSts *appsv1.StatefulSet)
 	return true, err
 }
 
+func (r *Reconciler) deleteStatefulSetIfOldLabelsAreUsed(desiredSts *appsv1.StatefulSet) (bool, error) {
+	currentSts, err := r.getStatefulSet(desiredSts)
+	if err != nil {
+		return false, err
+	}
+
+	if _, ok := currentSts.Labels[OldKeyActiveGate]; !ok {
```
baichinger, line 177: Is is sufficient to compare the desired labels with the ones given in the StatefulSet.

PR #92, opened 2021-03-22
Update deleteStatefulSet to compare with desired labels

gkrenn:
deleteStatefulSetIfOldLabelsAreUsed now generally compares if the labels have changed

### DIFF
--- a/controllers/capability/reconciler.go
+++ b/controllers/capability/reconciler.go
@@ -20,10 +20,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-const (
-	OldKeyActiveGate = "activegate"
-)
-
 type Reconciler struct {
 	client.Client
 	Instance                         *v1alpha1.DynaKube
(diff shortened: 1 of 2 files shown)


controllers/capability/reconciler.go
```diff
@@ -174,15 +170,17 @@ func (r *Reconciler) deleteStatefulSetIfOldLabelsAreUsed(desiredSts *appsv1.Stat
 		return false, err
 	}
 
-	if _, ok := currentSts.Labels[OldKeyActiveGate]; !ok {
-		return false, nil
+	for k := range desiredSts.Labels {
```
lrgar, line 173: Alternatively, you can go with `reflect.DeepEqual(currentSts.Labels, desiredSts.Labels)`

controllers/capability/reconiler_test.go
```diff
@@ -216,14 +216,20 @@ func TestReconcile_DeleteStatefulSetIfOldLabelsAreUsed(t *testing.T) {
...
 	desiredSts, err = r.buildDesiredStatefulSet()
 	require.NoError(t, err)
+	correctLabels := make(map[string]string)
+	for key, value := range desiredSts.Labels {
+		correctLabels[key] = value
+	}
+	delete(desiredSts.Labels, KeyActiveGate)
+	delete(desiredSts.Labels, KeyDynatrace)
+	delete(desiredSts.Labels, KeyFeature)
+	desiredSts.Labels["activegate"] = "dynakube"
```
lrgar, line 228: If the intention is just to replace the whole map, I may suggest doing,

```go
correctLabels := desiredSts.Labels
desiredSts.Labels = map[string]string { "activegate": "dynakube" }
```

lrgar (approved):
Added some suggestions, but good otherwise.

controllers/capability/reconiler_test.go
gkrenn: Yes, that's way better

controllers/capability/reconciler.go
gkrenn: Ok, thx

Did not know that

lrgar (approved):

DTMad (approved):

PR #94, opened 2021-03-23
Check token scope install script

DTMad:


deploy/install.sh
```diff
@@ -176,18 +175,59 @@ EOF
...
+    exit 1
+  fi
+}
+
+apiRequest() {
+  method=$1
+  url=$2
+  json=$3
+
+  response="$(curl -sS -X ${method} "${API_URL}""${url}" \
```
lrgar, line 219:
```suggestion
  response="$(curl -sS -X ${method} "${API_URL}${url}" \
```

deploy/install.sh
```diff
@@ -176,18 +175,59 @@ EOF
...
+  fi
+
+  if ! echo "$responseAPI" | grep -q "ReadConfig"; then
+    echo "Error: API token does not have config read permission!"
+    exit 1
+  fi
+
+  responsePaaS=$(apiRequest "POST" "/v1/tokens/lookup" "${jsonPaaS}")
+
+  if echo "$responsePaaS" | grep -q "Token does not exist"; then
```
lrgar, line 208: Technically, you could check for the `InstallerDownload` permission for PaaS tokens.

lrgar (dismissed):

deploy/install.sh
DTMad: Since the PaaSToken can only have InstallerDownload permission checking for it's existence should be sufficient here

lrgar: Except when the customer uses an API token as PaaS token, which is not that uncommon.

lrgar (approved):

PR #96, opened 2021-03-25
Fix message routing service not delete

gkrenn:
Fix 2 bugs where msg routing service not deleted: 
- After deletion of dynakube object, msgrouting not destroyed
- After disabling of routing in CR-file, msgrouting not destroyed

### DIFF
--- a/controllers/capability/routing/reconciler.go
+++ b/controllers/capability/routing/reconciler.go
@@ -20,8 +20,8 @@ import (
 )
 
 const (
-	module            = "msgrouting"
-	StatefulSetSuffix = "-" + module
+	Module            = "msgrouting"
+	StatefulSetSuffix = "-" + Module
 	capabilityName    = "MSGrouter"
 	containerPort     = 9999
 	DTDNSEntryPoint   = "DT_DNS_ENTRY_POINT"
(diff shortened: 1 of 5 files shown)


controllers/capability/routing/reconciler.go
```diff
@@ -94,11 +95,16 @@ func (r *Reconciler) Reconcile() (update bool, err error) {
 }
 
 func (r *Reconciler) createServiceIfNotExists() (bool, error) {
-	service := createService(r.Instance, module)
-	err := r.Get(context.TODO(), client.ObjectKey{Name: service.Name, Namespace: service.Namespace}, &service)
+	service := createService(r.Instance, Module)
+
+	if err := controllerutil.SetControllerReference(r.Instance, service, r.Scheme()); err != nil {
```
lrgar, line 100: Technically you only need to set this reference when creating, i.e., you didn't find it.

As it's now, you're doing this operation even if you create it or not.

tldr: put it inside the if below.

controllers/dynakube/dynakube_controller.go
```diff
@@ -260,6 +260,16 @@ func (r *ReconcileDynaKube) reconcileRouting(rec *utils.Reconciliation, dtc dtcl
 		if err := r.ensureDeleted(&sts); rec.Error(err) {
 			return false
 		}
+
+		ser := corev1.Service{
```
lrgar, line 264: I don't actually care, but just so you know that the short name for services in Kubernetes is `svc`.

lrgar (changes requested):

lrgar (approved):

PR #104, opened 2021-03-26
Add PodSecurityPolicy for routing

DTMad:


lrgar (dismissed):
As long as you tested this behavior in an actual environment with PSP enforcement enabled, I'm fine with it.

config/common/routing/role-routing.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: dynatrace-routing
+  namespace: dynatrace
+rules:
+  - apiGroups:
+      - policy
+    resources:
+      - podsecuritypolicies
+    resourceNames:
+      - dynatrace-kubernetes-monitoring
```
lrgar, line 12:
```suggestion
      - dynatrace-routing
```

lrgar (changes requested):

lrgar (approved):

PR #106, opened 2021-03-29
Adapt install script to not require cluster name as mandatory parameter

DTMad:


deploy/install.sh
```diff
@@ -93,7 +94,32 @@ applyDynatraceOperator() {
 }
 
 applyDynaKubeCR() {
-  cat <<EOF | "${CLI}" apply -f -
+  if [ -n "$CLUSTER_NAME" ]; then
```
lrgar, line 97: If I'm not mistaken, `-n` means non-empty, and this example doesn't use `CLUSTER_NAME`, so perhaps it's the other way around?

deploy/install.sh
```diff
@@ -142,8 +169,33 @@ addK8sConfiguration() {
     exit 1
   fi
 
-  json="$(
-    cat <<EOF
+  if [ -n "$CLUSTER_NAME" ]; then
```
lrgar, line 172: (Kind of) Same as above.

deploy/install.sh
```diff
@@ -142,8 +169,33 @@ addK8sConfiguration() {
...
+  "eventsFieldSelectors": [
+    {
+      "label": "Node events",
+      "fieldSelector": "involvedObject.kind=Node",
+      "active": true
+    }
+  ],
+  "workloadIntegrationEnabled": true,
+  "eventsIntegrationEnabled": false,
+  "activeGateGroup": "${CLUSTER_NAME}",
```
lrgar, line 189: I presume you don't want to use this field here, and only on the else.

deploy/install.sh
```diff
@@ -162,22 +214,28 @@ addK8sConfiguration() {
...
+  fi
 
   response=$(apiRequest "POST" "/config/v1/kubernetes/credentials" "${json}")
 
-  if echo "$response" | grep "${CLUSTER_NAME}" >/dev/null 2>&1; then
-    echo "Kubernetes monitoring successfully setup."
-  else
-    echo "Error adding Kubernetes cluster to Dynatrace: $response"
+  if [ -n "$CLUSTER_NAME" ]; then
+    if echo "$response" | grep "${CLUSTER_NAME}" >/dev/null 2>&1; then
```
lrgar, line 223: So no checks if connection name is used?

deploy/install.sh
```diff
@@ -162,22 +214,28 @@ addK8sConfiguration() {
...
   fi
 }
 
 checkForExistingCluster() {
   response=$(apiRequest "GET" "/config/v1/kubernetes/credentials" "")
-  if echo "$response" | grep -FEq "\"name\":\"${CLUSTER_NAME}\""; then
-    echo "Error: Cluster name already exists!"
-    exit 1
+
+  if [ -n "$CLUSTER_NAME" ]; then
```
lrgar, line 234: Technically you can use the connection name instead here to verify.

lrgar (changes requested):

deploy/install.sh
```diff
@@ -66,6 +74,9 @@ if [ -z "$CLUSTER_NAME" ]; then
     echo "Error: cluster name too long!"
     exit 1
   fi
+  CONNECTION_NAME=$CLUSTER_NAME
```
lrgar, line 77:
```suggestion
  CONNECTION_NAME="$CLUSTER_NAME"
```

deploy/install.sh
```diff
@@ -142,8 +174,32 @@ addK8sConfiguration() {
     exit 1
   fi
 
-  json="$(
-    cat <<EOF
+  if [ -z "$CLUSTER_NAME" ]; then
+    CONNECTION_NAME="$(echo "${K8S_ENDPOINT}" | awk -F[/:] '{print $4}')"
+
```
lrgar, line 179: No need to set it.
```suggestion
```

lrgar (approved):

deploy/install.sh
```diff
@@ -175,7 +230,8 @@ EOF
 
 checkForExistingCluster() {
   response=$(apiRequest "GET" "/config/v1/kubernetes/credentials" "")
-  if echo "$response" | grep -FEq "\"name\":\"${CLUSTER_NAME}\""; then
+
+  if echo "$response" | grep -FEq "\"name\":\"${CONNECTION_NAME}\""; then
```
baichinger, line 234: Causes `grep: conflicting matchers specified` for GNU grep 3.6.

baichinger (changes requested):

lrgar (approved):

baichinger (approved):

PR #107, opened 2021-03-29
Updated set-host-id-source parameter

gkrenn:
Set the variable `--set-host-id-source=auto` if in infra monitoring mode

### DIFF
--- a/controllers/oneagent/daemonset.go
+++ b/controllers/oneagent/daemonset.go
@@ -19,6 +19,8 @@ func prepareArgs(instance *dynatracev1alpha1.DynaKube, fs *dynatracev1alpha1.Ful
 	}
 
 	if feature == InframonFeature {
+		args = append(args, "--set-host-id-source=k8s-node-name")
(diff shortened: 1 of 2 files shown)


controllers/oneagent/daemonset.go
```diff
@@ -19,7 +19,7 @@ func prepareArgs(instance *dynatracev1alpha1.DynaKube, fs *dynatracev1alpha1.Ful
 	}
 
 	if feature == InframonFeature {
-		args = append(args, "--set-host-id-source=k8s-node-name")
+		args = append(args, "--set-host-id-source=auto")
```
lrgar, line 22: The other way around, when feature is not InframonFeature, then you set it to `auto`. When it's in infrastructure monitoring, it must be `k8s-node-name`

lrgar (changes requested):

lrgar (approved):

DTMad (approved):

PR #121, opened 2021-04-19
Add -k to curl and fix quoting for modularization

DTMad:


deploy/install.sh
```diff
@@ -312,11 +315,19 @@ apiRequest() {
...
+      -H "Authorization: Api-Token ${API_TOKEN}" \
+      -H "Content-Type: application/json; charset=utf-8" \
+      -d "${json}")"
+  else
+    response="$(curl -sS -X ${method} "${API_URL}${url}" \
+      -H "accept: application/json; charset=utf-8" \
+      -H "Authorization: Api-Token ${API_TOKEN}" \
+      -H "Content-Type: application/json; charset=utf-8" \
+      -d "${json}")"
+  fi
```
baichinger, line 330: The if-else block can be avoided by introducing a curl options variable. This would improve readability and maintainability a lot.

deploy/install.sh
```diff
@@ -117,9 +117,9 @@ applyDynaKubeCR() {
     apiUrl: ${API_URL}
     skipCertCheck: ${SKIP_CERT_CHECK}
 EOF
-  )
+  )"
 
-  classicFS=$(
+  classicFS="$(
     cat <<EOF
```
baichinger, line 124: Please remove all those unnecessary empty lines.

deploy/install.sh
```diff
@@ -312,11 +315,19 @@ apiRequest() {
   url=$2
   json=$3
 
-  response="$(curl -sS -X "${method}" "${API_URL}${url}" \
-    -H "accept: application/json; charset=utf-8" \
-    -H "Authorization: Api-Token ${API_TOKEN}" \
-    -H "Content-Type: application/json; charset=utf-8" \
-    -d "${json}")"
+  if "$SKIP_CERT_CHECK" = "true"; then
+    response="$(curl -k -sS -X ${method} "${API_URL}${url}" \
```
arthfl, line 319: you could reduce duplication if you just make the `curl` statement part of the if switch:
```suggestion
if "$SKIP_CERT_CHECK" = "true"; then
    curl_command="curl -k"
else
    curl_command="curl"
fi

response="$(${curl_command}" -sS -X ${method} "${API_URL}${url}" \
￼      -H "accept: application/json; charset=utf-8" \
￼      -H "Authorization: Api-Token ${API_TOKEN}" \
￼      -H "Content-Type: application/json; charset=utf-8" \
￼      -d "${json}")"
```

deploy/install.sh
```diff
@@ -312,11 +315,19 @@ apiRequest() {
...
+      -d "${json}")"
+  else
+    response="$(curl -sS -X ${method} "${API_URL}${url}" \
+      -H "accept: application/json; charset=utf-8" \
+      -H "Authorization: Api-Token ${API_TOKEN}" \
+      -H "Content-Type: application/json; charset=utf-8" \
+      -d "${json}")"
+  fi
 
   echo "$response"
```
arthfl, line 332: Is the reponse you get here speaking for itself or would it make sense to prepend this with some information?

arthfl (changes requested):

deploy/install.sh
DTMad: They are necessary since otherwise when concatenating the strings the 2nd string would land on the same line as the first one

arthfl: Not if there's a proper linebreak at the end :-)

deploy/install.sh
```diff
@@ -312,7 +315,13 @@ apiRequest() {
   url=$2
   json=$3
 
-  response="$(curl -sS -X "${method}" "${API_URL}${url}" \
+  if "$SKIP_CERT_CHECK" = "true"; then
+    curl_command="curl -k"
+  else
+    curl_command="curl"
+  fi
+  set -x
```
arthfl, line 323: Why the `set -x` here?

DTMad: Added for debugging - already removed

deploy/install.sh
DTMad: This is not really possible since we are using this request for every different api call we are doing within the script. So having proper handling would mean identifying the responses upfront, parsing the response as json and then getting the correct (and probably interesting part) out of it

arthfl (dismissed):

deploy/install.sh
baichinger: Its due to how it is fumbled together. Which by the way is hard to understand.

arthfl (approved):

PR #124, opened 2021-04-22
Use label selectors to inject into namespaces

meik99:


codemodules/match.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/Dynatrace/dynatrace-operator/dtlabels"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func MatchForNamespace(codeModuleDynakubes []dynatracev1alpha1.DynaKube, namespace *corev1.Namespace) (*dynatracev1alpha1.DynaKube, error) {
+	var matchingModules []dynatracev1alpha1.DynaKube
+
+	for _, codeModule := range codeModuleDynakubes {
+		matching, err := dtlabels.IsMatching(
```
lrgar, line 14: Just to confirm since I just ran into it, `metav1.LabelSelectorAsSelector()` does some of the logic here, e.g.,

```go
sel, err := metav1.LabelSelectorAsSelector(&dk.Spec.CodeModules.Selector)
if sel.Matches(labels.Set(dk.Labels)) {
  // Do something.
}
```

webhook/server/server.go
```diff
@@ -125,36 +125,14 @@ func (m *podInjector) Handle(ctx context.Context, req admission.Request) admissi
...
-	inject := utils.GetField(ns.Annotations, dtwebhook.AnnotationInject, "true")
-	inject = utils.GetField(pod.Annotations, dtwebhook.AnnotationInject, inject)
-	if inject == "false" {
-		return admission.Patched("")
-	}
-
-	oaName := utils.GetField(ns.Labels, dtwebhook.LabelInstance, "")
-	if oaName == "" {
-		return admission.Errored(http.StatusBadRequest, fmt.Errorf("no DynaKube instance set for namespace: %s", req.Namespace))
+	oa, response := m.FindDynakubeForPod(ctx, pod)
```
lrgar, line 130: Note that this high-throughput function. It's currently O(\<number of DynaKubes\>*\<number of Pods\>). Whether how relevant that is needs to be calculated.

There are alternatives, such as the Namespace controller maintaining a ConfigMap with the mappings from namespace to DynaKube that can be queried once.

webhook/server/server_match-selectors_test.go
```diff
@@ -0,0 +1,422 @@
+package server
```
lrgar, line 1: Underscores on file names.

meik99: Do you mean it should be `server-match-selectors_test.go`, `server_match_selectors_test.go` or something entirely different?

controllers/namespace/namespace_controller.go
```diff
@@ -83,23 +85,26 @@ func (r *ReconcileNamespaces) Reconcile(ctx context.Context, request reconcile.R
...
-		return reconcile.Result{}, nil
+	codeModuleDynakubes, err := codemodules.FindCodeModules(ctx, r.client)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to query CodeModule enabled Dynaekubes: %w", errors.WithStack(err))
 	}
 
-	oaName := ns.Labels[webhook.LabelInstance]
-	if oaName == "" {
-		return reconcile.Result{}, nil
+	apm, err := codemodules.MatchForNamespace(codeModuleDynakubes, &ns)
```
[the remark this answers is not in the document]
meik99: renamed

controllers/namespace/namespace_controller.go
```diff
@@ -124,7 +129,7 @@ func (r *ReconcileNamespaces) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, fmt.Errorf("failed to query tokens: %w", err)
 	}
 
-	script, err := newScript(ctx, r.client, apm, tkns, imNodes, r.namespace)
+	script, err := newScript(ctx, r.client, *apm, tkns, imNodes, r.namespace)
```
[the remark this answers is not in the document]
meik99: changed

webhook/server/server_match-selectors_test.go
meik99: changed to `server_match_selectors_test.go` for now

codemodules/match.go
meik99: changed and removed redundant code.
Thank you for that btw, I looked for something like that, but couldn't find anything.

webhook/server/server_match-selectors_test.go
lrgar: Yes, all in underscore.

baichinger:
Please revise:
* Remove `webhook/server/flavor.go` as it isn't used. Might be a left-over from ancient days...
* `AnnotationInject` in `webhook/config.go` can be removed as it is no longer used.

webhook/server/server.go
meik99: @baichinger, what is an example of a conflicting configuration? If you mean finding two dynakubes that are responsible for a single namespace, it already prints an error accordingly and does not inject

lrgar:
> AnnotationInject in webhook/config.go can be removed as it is no longer used.

I'd call this a bug, since it means there is no way to *not* instrument a Namespace by default except for a few Pods.

controllers/namespace/namespace_controller.go
```diff
@@ -83,23 +85,26 @@ func (r *ReconcileNamespaces) Reconcile(ctx context.Context, request reconcile.R
...
 	}
 
-	tokenName := utils.GetTokensName(&apm)
-	if !apm.Spec.CodeModules.Enabled {
-		r.ensureSecretDeleted(tokenName, targetNS)
+	tokenName := utils.GetTokensName(dk)
+	if !dk.Spec.CodeModules.Enabled {
+		// If secret does not exist, this method throws a not found error
+		// Suppress the error in that case
+		_ = r.ensureSecretDeleted(tokenName, targetNS)
```
baichinger, line 107: Is this comment even true? According to the implementation this should actually be not the case.

meik99: My mistake, I saw that the error was ignored (not suppressed) before and thought it was because of a potential not found error, because I overlooked `!k8serrors.IsNotFound(err)`  in the condition. I'll remove the comment, but I wonder why the error is ignored in the first place

baichinger:
> > AnnotationInject in webhook/config.go can be removed as it is no longer used.
> 
> I'd call this a bug, since it means there is no way to _not_ instrument a Namespace by default except for a few Pods.

Use case accepted. But would rather invert the logic: instrument a namespace (by selector) and opt-out on a per pod basis.

lrgar:
> Use case accepted. But would rather invert the logic: instrument a namespace (by selector) and opt-out on a per pod basis.

That was also supported, opt-out by default, and opt-in when Namespace had the annotation set to false.

lrgar:
It's probably worth to not inject into anything in kube-system and dynatrace (or whatever the webhook Pod is running on) namespaces.

controllers/namespace/namespace_controller.go
```diff
@@ -83,23 +85,24 @@ func (r *ReconcileNamespaces) Reconcile(ctx context.Context, request reconcile.R
...
+		return reconcile.Result{}, fmt.Errorf("failed to query CodeModule enabled Dynaekubes: %w", errors.WithStack(err))
 	}
 
-	oaName := ns.Labels[webhook.LabelInstance]
-	if oaName == "" {
-		return reconcile.Result{}, nil
+	dk, err := codemodules.MatchForNamespace(codeModuleDynakubes, &ns)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("could not match CodeModule with namespace: %w", errors.WithStack(err))
 	}
```
baichinger, line 96: Beware of excessive error logging and re-triggering the reconciler:
1. Does this justify an error message including a stack trace?
2. Will an immediate reconcile solve the problem or would it be sufficient to wait a couple more minutes?

Therefore, I suggest to log an info message and reschedule the event after a couple of minutes.

webhook/server/server.go
```diff
@@ -303,6 +284,38 @@ func (m *podInjector) Handle(ctx context.Context, req admission.Request) admissi
...
+
+	oa, err := codemodules.MatchForNamespace(codeModuleDynakubes, namespace)
+	if err != nil {
+		return nil, admission.Errored(http.StatusInternalServerError, err)
+	}
+	if oa == nil {
+		// If no DynaKube matches the namespaces labels, do not inject
+		return nil, admission.Patched("")
+	}
+	return oa, admission.Response{}
```
baichinger, line 316: Newline please.

webhook/server/server.go
```diff
@@ -303,6 +284,38 @@ func (m *podInjector) Handle(ctx context.Context, req admission.Request) admissi
...
 
+func (m *podInjector) FindDynakubeForPod(ctx context.Context, pod *corev1.Pod) (*dynatracev1alpha1.DynaKube, admission.Response) {
+	codeModuleDynakubes, err := codemodules.FindCodeModules(ctx, m.client)
+	if err != nil {
+		logger.Error(err, "error when trying to find Dynakubes with CodeModules enabled")
+
+		// If CodeModules is not enabled, or cannot be found, cannot inject
+		return nil, admission.Patched("")
+	}
+	if len(codeModuleDynakubes) <= 0 {
```
baichinger, line 295: else-if

webhook/server/server.go
```diff
@@ -303,6 +284,38 @@ func (m *podInjector) Handle(ctx context.Context, req admission.Request) admissi
...
+	if err != nil {
+		logger.Error(err, "error when trying to find namespace of pod", "pod", pod.Name, "namespace", pod.Namespace)
+		return nil, admission.Errored(http.StatusInternalServerError, err)
+	}
+
+	oa, err := codemodules.MatchForNamespace(codeModuleDynakubes, namespace)
+	if err != nil {
+		return nil, admission.Errored(http.StatusInternalServerError, err)
+	}
+	if oa == nil {
```
baichinger, line 312: else-if

webhook/server/server.go
```diff
@@ -303,6 +284,38 @@ func (m *podInjector) Handle(ctx context.Context, req admission.Request) admissi
 	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
 }
 
+func (m *podInjector) FindDynakubeForPod(ctx context.Context, pod *corev1.Pod) (*dynatracev1alpha1.DynaKube, admission.Response) {
```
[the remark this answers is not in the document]
meik99: changed

webhook/server/server.go
meik99: changed

webhook/server/server.go
meik99: changed

webhook/server/server.go
meik99: changed

webhook/server/server.go
```diff
@@ -303,6 +284,38 @@ func (m *podInjector) Handle(ctx context.Context, req admission.Request) admissi
...
+	codeModuleDynakubes, err := codemodules.FindCodeModules(ctx, m.client)
+	if err != nil {
+		logger.Error(err, "error when trying to find Dynakubes with CodeModules enabled")
+
+		// If CodeModules is not enabled, or cannot be found, cannot inject
```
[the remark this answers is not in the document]
meik99: removed comment

controllers/namespace/namespace_controller.go
meik99: changed

codemodules/find.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"context"
+
+	dynatracev1alpha1 "github.com/Dynatrace/dynatrace-operator/api/v1alpha1"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func FindCodeModules(ctx context.Context, clt client.Client) ([]dynatracev1alpha1.DynaKube, error) {
```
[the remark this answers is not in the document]
meik99: merged

DTMad (approved):

PR #127, opened 2021-04-22
Implement garbage collection for CSI driver

DTMad:


baichinger:
A short description how the GC works would be very helpful.

In order to prevent errors all the strings should be replaced by constants.

lrgar:
Description:

This PR implements Garbage Collection for binaries of unused OneAgent versions.

The idea here is based on keeping track of `NodePublishVolume()` and `NodeUnpublishVolume()` operations, that represent when a volume is mounted to a Pod, and unmounted from it, respectively.

We have two directories,
**PODS_DIR**: `/tmp/data/<Environment ID>/gc/<OneAgent version>/` where we have a file for each CSI-provisioned Pod.
**VOLS_DIR**: `/tmp/gc/` that points volume IDs (provided by Kubernetes) to the corresponding file in a).

The workflow is,
* `NodePublishVolume()` gets called.
  * We create a file on **PODS_DIR**, where the file name is the UID of the Pod we'll instrument.
  * We create a file on **VOLS_DIR**, where the file name is the volume ID and the contents is the path of the file we created on **PODS_DIR**.
* `NodeUnpublishVolume()` gets called.
  * We delete both files in **PODS_DIR** and **VOLS_DIR**.
  * Note that we don't get the Pod UID on this operation, which is why we need the link file to know what to delete on **VOLS_DIR**.
* Provisioner reconciler runs,
  * For each subdirectory we find inside `/tmp/data/<Environment ID>/gc/` (i.e., a **PODS_DIR**), that is not the latest. We look at files inside, and if there is none, then we delete Glibc and Musl binaries directories.

If the CSI Driver Pod gets restarted,
* We don't have access to existing **PODS_DIR** and **VOLS_DIR** directories from the non-existing container, in which case we ignore the missing files and continue.
* As I understand from Linux, the mounts that came from the deleted CSI Driver container will remain valid, and the source directories for the volume mounts will be deleted once they are not mounted by anyone, that happens when all of the containers that were using them are deleted.
* Note that the CSI Driver will receive `NodeUnpublishVolume()`, and will successfully unmount the directories, regardless of whether it has access or not the source directories.

The goal here is to run just a single command on the reconciler where we list the files to decide whether a particular version is unused or not.

Alternatively, we could keep track of the volume IDs and Pod UIDs in memory. However, if we use a different storage backend for the binaries (e.g., file system), then that information will be lost if the CSI Driver Pod gets restarted.

controllers/csi/driver/server.go
```diff
@@ -225,21 +226,29 @@ func (svr *CSIDriverServer) NodePublishVolume(ctx context.Context, req *csi.Node
...
 		return nil, status.Error(codes.Internal, fmt.Sprintf("failed to mount OneAgent volume: %s", err.Error()))
 	}
 
+	gcFile := filepath.Join(envDir, "gc", string(ver), podUID)
+	if err = ioutil.WriteFile(gcFile, nil, 0770); err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("Failed to create file for garbage collector - error: %s", err))
+	}
+	if err = ioutil.WriteFile(filepath.Join(dtcsi.GarbageCollectionPath, volID), []byte(gcFile), 0770); err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("Failed to create link file for garbage collector - error: %s", err))
+	}
```
meik99, line 250: this creates an empty file at `/tmp/data/<env>/gc/<ver>/<podUID>` and copys the content of the file to `/gc/<volID>`. First of all, I think the target path should be `/tmp/gc/<volID>` if I understood that correctly. Secondly, the error message states that it couldn't link the file, but symlinking is done with `os.Symlink(...)`, wouldn't the above create a copy of the source file?

controllers/csi/driver/server.go
```diff
@@ -256,14 +265,27 @@ func (svr *CSIDriverServer) NodeUnpublishVolume(ctx context.Context, req *csi.No
...
-		Mount{Target: filepath.Join(targetPath, "datastorage")},
+		Mount{Target: filepath.Join(targetPath, dtcsi.AgentConfDir)},
+		Mount{Target: filepath.Join(targetPath, dtcsi.LogDir)},
+		Mount{Target: filepath.Join(targetPath, dtcsi.DatastorageDir)},
 		Mount{Target: targetPath},
 	); err != nil {
 		return nil, status.Error(codes.Internal, fmt.Sprintf("failed to unmount volume: %s", err.Error()))
 	}
 
+	linkFile := filepath.Join(dtcsi.GarbageCollectionPath, volumeID)
```
meik99, line 276: Path points to` /gc/<volumeId`, like previously, are you sure that is the correct path?

controllers/csi/provisioner/reconciler.go
```diff
@@ -101,14 +101,14 @@ func (r *OneAgentProvisioner) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, fmt.Errorf("failed to fetch connection info: %w", err)
 	}
 
-	envDir := filepath.Join(r.opts.DataDir, ci.TenantUUID)
+	envDir := filepath.Join(r.opts.RootDir, dtcsi.DataPath, ci.TenantUUID)
 	verFile := filepath.Join(envDir, "version")
-	tenantFile := filepath.Join(r.opts.DataDir, fmt.Sprintf("tenant-%s", dk.Name))
+	tenantFile := filepath.Join(r.opts.RootDir, dtcsi.DataPath, fmt.Sprintf("tenant-%s", dk.Name))
```
meik99, line 106: what happens with multiple tenants and same dynakube name?

meik99 (changes requested):

controllers/csi/driver/server.go
```diff
@@ -225,21 +226,29 @@ func (svr *CSIDriverServer) NodePublishVolume(ctx context.Context, req *csi.Node
...
-			Source: filepath.Join(envDir, "datastorage", podUID),
-			Target: filepath.Join(targetPath, "datastorage"),
+			Source: filepath.Join(envDir, dtcsi.DatastorageDir, podUID),
+			Target: filepath.Join(targetPath, dtcsi.DatastorageDir),
 		},
 	); err != nil {
 		return nil, status.Error(codes.Internal, fmt.Sprintf("failed to mount OneAgent volume: %s", err.Error()))
 	}
 
+	gcFile := filepath.Join(envDir, "gc", string(ver), podUID)
```
gkrenn, line 244: Would use `GarbageCollectionPath` const

controllers/csi/gc/binaries.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"os"
+	"path/filepath"
+
+	dtcsi "github.com/Dynatrace/dynatrace-operator/controllers/csi"
+	"github.com/go-logr/logr"
+)
+
+func runBinaryGarbageCollection(logger logr.Logger, envID string, latestVersion string) error {
+	gcPath := filepath.Join(dtcsi.DataPath, envID, "gc")
```
gkrenn, line 12: Would use `GarbageCollectionPath` const

controllers/csi/gc/controller.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+func (r *CSIGarbageCollector) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&dynatracev1alpha1.DynaKube{}).
+		Complete(r)
+}
+
+var _ reconcile.Reconciler = &CSIGarbageCollector{}
```
gkrenn, line 45: Why we need this? @baichinger @DTMad

controllers/csi/provisioner/reconciler.go
```diff
@@ -144,6 +144,11 @@ func (r *OneAgentProvisioner) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	if ver != oldVer {
+		gcDir := filepath.Join(envDir, "gc", ver)
```
gkrenn, line 147: gc constant

controllers/csi/provisioner/reconciler.go
```diff
@@ -220,8 +227,6 @@ func installAgent(rlog logr.Logger, dtc dtclient.Client, flavor, arch, targetDir
 }
 
 func unzip(rlog logr.Logger, r *zip.Reader, outDir string) error {
-	const agentConfPath = "agent/conf/"
-
 	os.MkdirAll(outDir, 0755)
```
meik99, line 230: errors shouldn't be implicitly ignored, use `_ = os.MkdirAll(...)` to explicitly ignore them or handle them accordingly

meik99: please double check all similar calls like `os.MkdirAll(...)` or `ioutil.WriteFile(...)`

controllers/csi/gc/controller.go
```diff
@@ -0,0 +1,90 @@
...
+
+	ci, err := dtc.GetConnectionInfo()
+	if err != nil {
+		return r.newReconcileResult(), fmt.Errorf("failed to fetch connection info: %w", err)
+	}
+
+	ver, err := dtc.GetLatestAgentVersion(dtclient.OsUnix, dtclient.InstallerTypePaaS)
+	if err != nil {
+		return r.newReconcileResult(), fmt.Errorf("failed to query OneAgent version: %w", err)
+	}
```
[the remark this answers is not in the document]
DTMad: Once we move the "base fields" like the tenantUUID and the agent version and so on those calls won't be necessary anymore.

controllers/csi/gc/binaries.go
```diff
@@ -0,0 +1,44 @@
...
+	gcDirs, err := os.ReadDir(gcPath)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
DTMad: Changed

controllers/csi/driver/server.go
```diff
@@ -194,20 +195,20 @@ func (svr *CSIDriverServer) NodePublishVolume(ctx context.Context, req *csi.Node
 		return nil, status.Error(codes.FailedPrecondition, fmt.Sprintf("Failed to query namespace %s: %s", nsName, err.Error()))
 	}
 
-	dkName := ns.Labels["oneagent.dynatrace.com/instance"] // TODO(lrgar): replace with constant
+	dkName := ns.Labels[webhook.LabelInstance]
```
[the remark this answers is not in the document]
DTMad: You're right - we need to also change the logic here to consider the matchLabels.
Will create a follow up ticket

controllers/csi/gc/binaries.go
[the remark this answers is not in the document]
DTMad: Changed

controllers/csi/driver/server.go
DTMad: Changed

controllers/csi/provisioner/reconciler.go
DTMad: We can't have multiple tenants within one dynakube instance

meik99: what about same name (not instance) and different namespace?

controllers/csi/driver/server.go
DTMad: Changed

controllers/csi/gc/binaries.go
DTMad: Changed

controllers/csi/provisioner/reconciler.go
DTMad: Changed

controllers/csi/gc/binaries.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+)
+
+func runBinaryGarbageCollection(logger logr.Logger, envID string, latestVersion string, opts dtcsi.CSIOptions) error {
+	gcPath := filepath.Join(opts.RootDir, dtcsi.DataPath, envID, dtcsi.GarbageCollectionPath)
+	gcDirs, err := os.ReadDir(gcPath)
+	if err != nil {
+		return errors.WithStack(err)
+	}
```
baichinger, line 17: There is a race condition upon startup when there isn't yet a version downloaded. Therefore, this shouldn't be an error in the first place. The condition can be logged as an info message.

controllers/csi/gc/controller.go
```diff
@@ -0,0 +1,91 @@
...
+func (r *CSIGarbageCollector) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&dynatracev1alpha1.DynaKube{}).
+		Complete(r)
+}
+
+var _ reconcile.Reconciler = &CSIGarbageCollector{}
+
+func (r *CSIGarbageCollector) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	r.logger.Info("reconciling csi driver", "namespace", request.Namespace, "name", request.Name)
```
baichinger, line 49: The message is misleading. It should be something like _running OneAgent garbage collection_.

controllers/csi/gc/controller.go
baichinger: Safety measure. It helps the compiler to ensure the interface. Newer versions of operator-sdk/kubebuilder seem to not generate this code anymore.

controllers/csi/driver/server.go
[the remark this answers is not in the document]
baichinger: As the label will be removed, the connection to the corresponding DynaKube object will be passed via volume attributes.

controllers/csi/gc/binaries.go
```diff
@@ -0,0 +1,49 @@
...
+		if dir.Name() == latestVersion {
+			continue
+		}
+		subDirs, err := os.ReadDir(filepath.Join(gcPath, dir.Name()))
+		if err != nil {
+			return err
+		}
+		if len(subDirs) == 0 {
+			logger.Info("Garbage collector deleting unused version", "version", dir.Name())
+			err = os.RemoveAll(filepath.Join(opts.RootDir, dtcsi.DataPath, envID, "bin", dir.Name()+"-default"))
```
meik99, line 33: the path should be refactored into a variable to avoid errors, since the same path is used twice
```
binDir := filepath.Join(opts.RootDir, dtcsi.DataPath, envID, "bin")
err = os.RemoveAll(filepath.Join(binDir, dir.Name()+"-default"))
```

controllers/csi/gc/binaries.go
```diff
@@ -0,0 +1,49 @@
...
+			if err != nil {
+				return errors.WithStack(err)
+			}
+			err = os.RemoveAll(filepath.Join(opts.RootDir, dtcsi.DataPath, envID, "bin", dir.Name()+"-musl"))
```
meik99, line 37: the path should be refactored into a variable to avoid errors, since the same path is used twice
```
err = os.RemoveAll(filepath.Join(binDir, dir.Name()+"-musl"))
```

meik99 (dismissed):

baichinger (dismissed):

baichinger (approved):

baichinger (approved):

PR #131, opened 2021-04-29
Add limit to csi driver pods

gkrenn:


config/kubernetes/csi/daemonset-csi.yaml
```diff
@@ -37,6 +37,13 @@ spec:
                   apiVersion: v1
                   fieldPath: spec.nodeName
           imagePullPolicy: Always
+          resources:
+            requests:
+              cpu: 300m
+              memory: 128Mi
+            limits:
+              cpu: 600m
+              memory: 256Mi
```
baichinger, line 46: The limits seems to be quite high. We would reserve half a core on each node.

In order to prevent eviction, it might also help to stick to guaranteed QoS class.

DTMad: @gkrenn please adjust the limits once you have meaningful results and keep limits and requests same so we go with the guaranteed QoS

DTMad (approved):

baichinger (approved):

PR #132, opened 2021-04-30
test BindMount and BindUnmount

meik99:
A set of unit tests for BindMount and BindUnmount functions of the CSI driver with some refactoring

### DIFF
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -19,7 +19,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	runtime "k8s.io/apimachinery/pkg/runtime"
 )
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -19,7 +19,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -19,7 +19,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
baichinger, line 22: Guess this was by accident. Please revert.

controllers/csi/driver/mount.go
```diff
@@ -31,12 +31,29 @@ type Mount struct {
...
+type bindOptions struct {
+	rootDir string
+	mounts  []Mount
+	mounter mount.Interface
+}
+
 func BindMount(rootDir string, mnts ...Mount) error {
 	if err := os.MkdirAll(rootDir, 0750); err != nil {
 		return fmt.Errorf("failed to create root directory %s: %w", rootDir, err)
 	}
```
baichinger, line 43: This `rootDir` option is redundant. Directories should be created (and removed in `BindUnmount`) based on information in `mnts`.

api/v1alpha1/zz_generated.deepcopy.go
meik99: fixed

controllers/csi/driver/mount.go
meik99: I don't get how it is redundant, could you elaborate on that?

baichinger: Directories can be created when iterating over `mnts`.

Additionally, the order of entries in `mnts` is crucial. The first one is actually the root directory and all others depend on existing directories in the root (agent) directory. This code should not depend on that.

Anyways, this can be addressed in a PR on its own.

PR #143, opened 2021-05-19
KubeMon capability name fixed

mjgrzybek:


mjgrzybek:
`MSGrouter` is ok, only `kubernetes_monitoring` was affected.

controllers/activegate/capability/capability.go
```diff
@@ -62,7 +62,7 @@ func NewKubeMonCapability(crProperties *dynatracev1alpha1.CapabilityProperties)
 	return &KubeMonCapability{
 		capabilityBase{
 			moduleName:     "kubemon",
-			capabilityName: "kubernetes-monitoring",
+			capabilityName: "kubernetes_monitoring",
```
baichinger, line 65: Does this impose a dependency on a specific server/AG version?

baichinger: This is actually fixes a regression.

PR #144, opened 2021-05-19
Feature/change default code modules

waodim:
- I changed our old approach where we looked, if we have a empty VolumeSource and just set an EmptyDir
- Now the approach asks, if there is a empty VolumeSource, if so, it sets our CSI Driver


baichinger:
Would really like to see a refactoring of the test. At least the redundant object instantiations should be moved into a shared function.

webhook/server/server.go
```diff
@@ -174,7 +174,9 @@ func (m *podInjector) Handle(ctx context.Context, req admission.Request) admissi
 
 	dkVol := oa.Spec.CodeModules.Volume
 	if dkVol == (corev1.VolumeSource{}) {
-		dkVol.EmptyDir = &corev1.EmptyDirVolumeSource{}
+		dkVol.CSI = &corev1.CSIVolumeSource{
+			Driver: "csi.oneagent.dynatrace.com",
```
meik99, line 178: It's a bit of a nitpick, but a constant should be used for the driver name

meik99 (dismissed):

webhook/server/server.go
baichinger: There is already one: `dtcsi.DriverName` in `controllers/csi/config.go`

meik99 (approved):

DTMad (approved):

gkrenn (approved):

baichinger (dismissed):

DTMad (approved):

baichinger (approved):

PR #146, opened 2021-05-20
Add fields to Dynakube status

meik99:
Added fields to Dynakube status to cache often used values previously retrieved via API/REST calls


controllers/dtpullsecret/generate.go
```diff
@@ -83,5 +83,5 @@ func pullSecretDataFromDockerConfig(dockerConf *dockerConfig) (map[string][]byte
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
-	return map[string][]byte{".dockerconfigjson": dockerConfJson}, nil
+	return map[string][]byte{dockerConfigJson: dockerConfJson}, nil
```
DTMad, line 86: Are you certain that this works?

controllers/dynakube/dynakube_controller.go
```diff
@@ -139,19 +140,29 @@ func (r *ReconcileDynaKube) Reconcile(ctx context.Context, request reconcile.Req
...
 		return
 	}
 
+	err = status.SetDynakubeStatus(rec.Instance, status.Options{
+		Dtc:       dtc,
+		ApiClient: r.apiReader,
+	})
+	if rec.Error(err) {
+		rec.Log.Error(err, "could not set Dynakube status")
+		return
```
DTMad, line 163: Do we really want to return here?
Wouldn't it be possible to create the AG or the OA even without having some of the status fields?

gkrenn (commented):
Seems to be fine if you answer @DTMad questions

controllers/dtpullsecret/generate.go
meik99: Yes, I am. Just replacing a string with a constant of the same string value

controllers/dynakube/dynakube_controller.go
meik99: It would be possible to create them without correctly set fields, but that would result in errors down the line. I thought it would be better to show errors early.

DTMad (approved):

PR #150, opened 2021-05-25
fixup! Fix building the image tag for branch deployments (#136)

toszr:
Solved problem with make deploy from a branch:
`Makefile:19: *** unterminated call to function 'shell': missing ')'.  Stop.`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ ifeq ($(origin IMG),undefined)
 ifeq ($(shell git branch --show-current),master)
 IMG=quay.io/dynatrace/dynatrace-operator:snapshot
 else
-IMG=quay.io/dynatrace/dynatrace-operator:snapshot-$(shell git branch --show-current | sed "s#[^a-zA-Z0-9_-]#-#g")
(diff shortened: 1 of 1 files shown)


DTMad:
Are you sure that this does not happen due to an outdated make cli version?
@baichinger changed this here: https://github.com/Dynatrace/dynatrace-operator/pull/136 and for me it worked after updating my make cli.

toszr:
I'm using GNU make 4.2.1 which is the latest available for Ubuntu 20.04.01 LTS:
```
$ make --version
GNU Make 4.2.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
$ sudo apt install -u make
Reading package lists... Done
Building dependency tree       
Reading state information... Done
make is already the newest version (4.2.1-1.2).
```

baichinger (approved):

PR #151, opened 2021-05-25
Add configuration section for metrics ingest to custom resource

mjgrzybek:
Configuration switched from annotation to CRD-defined one.


mjgrzybek:
Closing due to discussion if `MetricsIngest` is the final form of `*Ingest`.

PR #152, opened 2021-05-26
DT_DEPLOYMENT_METADATA env var added to AG PODs

aorcholski:
DT_DEPLOYMENT_METADATA environment variable added

### DIFF
--- a/controllers/activegate/statefulset.go
+++ b/controllers/activegate/statefulset.go
@@ -9,6 +9,7 @@ import (
 	dynatracev1alpha1 "github.com/Dynatrace/dynatrace-operator/api/v1alpha1"
 	"github.com/Dynatrace/dynatrace-operator/controllers/customproperties"
 	"github.com/Dynatrace/dynatrace-operator/controllers/dtpullsecret"
+	"github.com/Dynatrace/dynatrace-operator/deploymentmetadata"
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


controllers/activegate/statefulset.go
```diff
@@ -195,10 +197,13 @@ func buildVolumeMounts(stsProperties *statefulSetProperties) []corev1.VolumeMoun
 }
 
 func buildEnvs(stsProperties *statefulSetProperties) []corev1.EnvVar {
```
mjgrzybek, line 199: it's used onfly for `containers`, but not for `initContainers`. Open question - should we provide this env vars to `initContainers`? I think yes.

aorcholski: There are no initContainers yet. Will be added by PR "Feature/k8s internal certificate import to the operator". So let's do it after the PR is merged.

DTMad (dismissed):

controllers/activegate/statefulset.go
```diff
@@ -38,6 +39,7 @@ const (
 	DTNetworkZone     = "DT_NETWORK_ZONE"
 	DTGroup           = "DT_GROUP"
 	DTInternalProxy   = "DT_INTERNAL_PROXY"
+	DTDeployMetadata  = "DT_DEPLOY_METADATA"
```
baichinger, line 42: The title of the PR says _DT_DEPLOYMENT_METADATA_, for code modules it is _DT_DEPLOYMENT_METADATA_ as well. Does ActiveGate really use a different one?

aorcholski (commented):
Yes, you are right. Fix committed.

DTMad (approved):

PR #154, opened 2021-05-31
Use multi libc (multidstiro) archive for code modules

baichinger:
The current approach for providing agents based on glibc and/or musl based on a flavor annotation can be replaced and simplified with the combined multidistro archive:
* CSI and emptyDir workflows use new archive only.
* Removed flavor annotation.


dtclient/client.go
```diff
@@ -76,8 +76,9 @@ const (
 
 // Known flavors.
 const (
-	FlavorDefault = "default"
-	FlavorMUSL    = "musl"
+	FlavorDefault     = "default"
+	FlavorMUSL        = "musl"
```
DTMad, line 80: Do we still need those consts?

baichinger: Currently unused. But they are still valid arguments to `GetLatestAgent`.

DTMad (approved):

PR #167, opened 2021-06-10
Feature/ag connection info

mjgrzybek:
Connection info is:
* tenantUUID
* tenantToken
* communicationEndpoint

Operator retrieves connection info from DT cluster using REST API.
Non-secrets are set as environment variables in AG pod.
Secrets are mounted as volumes.

The complementary change was made in AG Dockerfile to make use of these data - if Operator sets them, they override data set by cluster in `entrypoint.properties`.


controllers/tokens/agtokens_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	return map[string][]byte{"tenant-token": []byte(info.Token)}, nil
+}
+
+func isAGTokensSecretEqual(currentSecret *corev1.Secret, desired map[string][]byte) bool {
+	return reflect.DeepEqual(desired, currentSecret.Data)
+}
+
+func BuildAGTokensSecret(instance *dynatracev1alpha1.DynaKube, agTokensSecretData map[string][]byte) *corev1.Secret {
```
baichinger, line 123: No need to export.

controllers/tokens/agtokens_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return fmt.Errorf("could not fetch AG tokens: %w", err)
+	}
+
+	agTokensSecret, err := r.createAGTokensSecretIfNotExists(agTokens)
+	if err != nil {
+		return fmt.Errorf("failed to create or update secret: %w", err)
+	}
+
+	return r.updateAGTokensIfOutdated(agTokensSecret, agTokens)
```
baichinger, line 67: A bit difficult to understand. I would merge those two methods as they the one calls the other anyways.

dtclient/tenant.go
```diff
@@ -15,8 +15,11 @@ type TenantInfo struct {
 	CommunicationEndpoint string
 }
 
-func (dtc *dynatraceClient) GetTenantInfo() (*TenantInfo, error) {
-	url := fmt.Sprintf("%s/v1/deployment/installer/agent/connectioninfo", dtc.url)
+type responseReader func(*dynatraceClient, []byte) (*TenantInfo, error)
+
+func (dtc *dynatraceClient) GetTenantInfo(apiEndpoint string, readResponseForTenantInfo responseReader) (*TenantInfo, error) {
```
baichinger, line 20: No need to export.

controllers/tokens/agtokens_reconciler.go
```diff
@@ -0,0 +1,132 @@
...
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	SecretsName = "dynatrace-tokens"
```
baichinger, line 21: In case multiple environments are configured the token would get overwritten by each reconciliation. Therefore the name of the secret must contain the name of the DynaKube object.

baichinger, line 21: What happens on a token update to the ActiveGate? Does it need to restart or will it update the token internally as well?

api/v1alpha1/dynakube_types.go
```diff
@@ -27,6 +27,10 @@ type DynaKubeSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Custom PullSecret",order=8,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:io.kubernetes:Secret"}
 	CustomPullSecret string `json:"customPullSecret,omitempty"`
 
+	// Tokens for ActiveGates
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="AG Tokens",order=2,xDescriptors="urn:alm:descriptor:io.kubernetes:Secret"
+	AGTokensSecret string `json:"agTokensSecret,omitempty"`
```
baichinger, line 32: Why is this needed? It is not used when assembling the STS.

controllers/tokens/agtokens_reconciler.go
mjgrzybek: AG doesn't currently support tokens rotation during runtime. Restart is required.

controllers/tokens/agtokens_reconciler.go
mjgrzybek: It's a valid point but that's the pattern that's already in use. 
```
$ ag --go IfOutdated
controllers/dtpullsecret/reconciler.go
69:     return r.updatePullSecretIfOutdated(pullSecret, pullSecretData)
82:func (r *Reconciler) updatePullSecretIfOutdated(pullSecret *corev1.Secret, desiredPullSecretData map[string][]byte) error {

controllers/customproperties/reconciler.go
55:                     err = r.updateCustomPropertiesIfOutdated()
76:func (r *Reconciler) updateCustomPropertiesIfOutdated() error {

controllers/tokens/agtokens_reconciler.go
67:     return r.updateAGTokensIfOutdated(agTokensSecret, agTokens)
80:func (r *Reconciler) updateAGTokensIfOutdated(agTokens *corev1.Secret, agTokensSecretData map[string][]byte) error {

controllers/activegate/reconciler/statefulset/reconciler_test.go
168:func TestReconcile_UpdateStatefulSetIfOutdated(t *testing.T) {
174:    updated, err := r.updateStatefulSetIfOutdated(desiredSts)
183:    updated, err = r.updateStatefulSetIfOutdated(desiredSts)
191:    updated, err = r.updateStatefulSetIfOutdated(desiredSts)

controllers/activegate/reconciler/statefulset/reconciler.go
115:    updated, err := r.updateStatefulSetIfOutdated(desiredSts)
160:func (r *Reconciler) updateStatefulSetIfOutdated(desiredSts *appsv1.StatefulSet) (bool, error) {
```
I think such refactoring should be done in separate PR.

controllers/tokens/agtokens_reconciler.go
mjgrzybek: multiple environments == multiple operators? Secret are held in operator's namespace. 
Do we support running multiple instances of `dynatrace-operator` in single namespace?

mjgrzybek: note:
same logic is used for https://github.com/Dynatrace/dynatrace-operator/blob/master/api/v1alpha1/dynakube_types.go#L24

api/v1alpha1/dynakube_types.go
mjgrzybek: related discussion: https://github.com/Dynatrace/dynatrace-operator/pull/167#discussion_r658550010

mjgrzybek: We don't need it because secrets name can be extended with dynakube object name, so it can be calculated dynamically instead of declare in spec? Let's discuss it in linked discussion.

controllers/tokens/agtokens_reconciler.go
mjgrzybek: related discussion: https://github.com/Dynatrace/dynatrace-operator/pull/167#discussion_r659588372

baichinger: > multiple environments == multiple operators? Secret are held in operator's namespace.
> Do we support running multiple instances of `dynatrace-operator` in single namespace?

Single operator handles multiple DynaKube objects. Each of these DynaKubes can belong to different environments. Only one namespace involved.

api/v1alpha1/dynakube_types.go
baichinger: The rationale for having the `Tokens` field is to share the user provided tokens among multiple DynaKube objects. `AGTokensSecret` is a bit different as it is always fetched via Dynatrace API.

mjgrzybek:
Refactoring in progress due to #146

mjgrzybek:
Reopen if needed.

PR #169, opened 2021-06-11
Disable OneAgent binary download when csi driver is not used

meik99:


controllers/csi/provisioner/reconciler.go
```diff
@@ -79,7 +79,7 @@ func (r *OneAgentProvisioner) Reconcile(ctx context.Context, request reconcile.R
 		}
 		return reconcile.Result{}, err
 	} else if dk == nil {
-		rlog.Info("Code modules disabled")
+		rlog.Info("Code modules or csi driver disabled")
 		return reconcile.Result{RequeueAfter: 30 * time.Minute}, nil
```
gkrenn, line 83: Why do we need to requeue this after 30 min? - The dynakube object won't change in the 30 min and if then it will be reconciled anyway

DTMad (approved):

Issue #160, opened 2021-06-07
Host Group parameter not applied


popecruzdt:
The Host Group parameter for the OneAgent isn't applied due to a typo in the install.sh script.

Modify the host group (currently line 160) in install.sh file, remove the double quotes.
OLD:
    - --set-host-group="${CLUSTER_NAME}"
NEW:
    - --set-host-group=${CLUSTER_NAME}

PR #170, opened 2021-06-15
Removed quotes around ${CLUSTER_NAME} in host group parameter.

popecruzdt:
Issue #160
Host Group parameter not applied

Modify the host group in install.sh file, remove the double quotes.
OLD:
- --set-host-group="${CLUSTER_NAME}"
NEW:
- --set-host-group=${CLUSTER_NAME}


arthfl (approved):
Removing the quotes in this context should be fine. `shellcheck` also doesn't complain.

DTMad (approved):

PR #172, opened 2021-06-16
Fix instrumenting third party containers

chrismuellner:
If another webhook gets executed after ours, and adds a side container, then it wasn't instrumented. For example, Istio.

With these changes, the order the webhook gets executed in does not matter anymore.


baichinger:
Can imagine scenarios where this causes problems. There might be even use cases where this is undesired. Therefore suggest to have a feature flag via an annotation to disable this behavior.

api/v1alpha1/feature_flags.go
```diff
@@ -21,11 +21,12 @@ import (
...
-	annotationFeatureDisableActiveGateUpdates = annotationFeaturePrefix + "disable-activegate-updates"
-	annotationFeatureDisableHostsRequests     = annotationFeaturePrefix + "disable-hosts-requests"
-	annotationFeatureOneAgentMaxUnavailable   = annotationFeaturePrefix + "oneagent-max-unavailable"
-	annotationFeatureEnableMetricsIngest      = annotationFeaturePrefix + "enable-metrics-ingest"
+	annotationFeaturePrefix                            = "alpha.operator.dynatrace.com/feature-"
+	annotationFeatureDisableActiveGateUpdates          = annotationFeaturePrefix + "disable-activegate-updates"
+	annotationFeatureDisableHostsRequests              = annotationFeaturePrefix + "disable-hosts-requests"
+	annotationFeatureOneAgentMaxUnavailable            = annotationFeaturePrefix + "oneagent-max-unavailable"
+	annotationFeatureEnableMetricsIngest               = annotationFeaturePrefix + "enable-metrics-ingest"
+	annotationFeatureEnableInstrumentMissingContainers = annotationFeaturePrefix + "enable-instrument-missing-containers"
```
DTMad, line 29: Would probably rename this to something like "enable-webhook-reinvocation-policy"

DTMad (approved):

PR #176, opened 2021-06-17
Disabled classicFullStack

mmayr-at:


DTMad:
@mmayr-at please also change the default here:
https://github.com/Dynatrace/dynatrace-operator/blob/master/config/samples/cr.yaml#L160

DTMad:
Please also add it to the openshift section in the Readme.md

DTMad (approved):

olljanat:
@mmayr-at @DTMad what is plan/roadmap here?

Can we expect that full stack monitoring will be implemented on some other way (hopefully on way which does need current oneagent at all) as this setting name is `classicFullStack` and now it is even disabled by default?

DTMad:
There will be a replacement where we change the way of injection in behind and improve usability.
However, it will still appear with the same visibility and insights you are used to.

I can't give an exact plan yet but it will be released with v0.3.0

olljanat:
That is nice to hear. I can also participate for piloting when any kind of beta/etc version is available.

I would be very happy to see that it would be implemented on way that it can support https://github.com/talos-systems/talos

ActiveGate and Kubernetes connection already works on it  but OneAgent cannot as it have long list requirements for binaries and libraries which does not exist on Talos.


PS. Just to clarifying: Company where I'm working on is Dynatrace customer and currently running Docker Swarm on private cloud. Eventually we will migrate to Kubernetes and it would be nice to be able to use Talos capabilities to minimize effort needed to maintain it without losing Dynatrace support.

DTMad:
We don't know if we will have some kind of beta - if so I'll let you know.

The data the agent and the AG delivers will not change. 
It will be a change of how we do the injection within Kubernetes to solve pain points we used to have in the past and to make our injection cloud native.

PR #179, opened 2021-06-21
Update deprecated apis

chrismuellner:
- Remove webhook for openshift 3.11
- Add separate CRD for openshift 3.11
- Adapt makefile to separate openshift 3.11 case from default


DTMad (approved):

Makefile
```diff
@@ -85,10 +86,22 @@ deploy-ocp: manifests kustomize
...
+	cd config/deploy && $(KUSTOMIZE) create
+	cd config/deploy && $(KUSTOMIZE) edit add base ../openshift3.11
+	cd config/deploy && $(KUSTOMIZE) edit set image "quay.io/dynatrace/dynatrace-operator:snapshot"=${IMG}
+	$(KUSTOMIZE) build config/deploy | oc apply -f -
+
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) paths="./..." output:crd:artifacts:config=config/crd/bases
-	python3 ./hack/customize_crds.py
+	#python3 ./hack/customize_crds.py
```
DTMad, line 101: In that script we used to remove a few parts which get added to the CRD with newer versions.
I guess for the new default we won't need it anymore but most probably for OCP 3.11? If not feel free to get rid of the script as well

controllers/webhookcerts/controller.go
```diff
@@ -80,6 +82,7 @@ type ReconcileWebhookCertificates struct {
 	logger    logr.Logger
 	namespace string
 	now       time.Time
+	config    *rest.Config
```
DTMad, line 85: Why do we need this here?

DTMad (changes requested):

Makefile
chrismuellner: I don't think it's need anymore, since we explicitly use the old crdVersion for OCP 3.11

controllers/webhookcerts/controller.go
chrismuellner: No longer required, was used to get a discoveryClient for checking package versions

DTMad (dismissed):

baichinger (approved):

PR #190, opened 2021-07-14
Feature/report types of deploymentin operator

waodim:


deploymentmetadata/deploymentmetadata.go
```diff
@@ -14,34 +15,106 @@ const (
...
+	}
+
+	if metadata.DynaKube.Spec.InfraMonitoring.Enabled {
+		metadata.osAgentDeployed = "true"
+	}
+
+	if metadata.DynaKube.Spec.ClassicFullStack.Enabled {
+		metadata.osAgentDeployed = "true"
+		metadata.codeModulesDeployed = ""
+	}
```
meik99, line 115: should the metrics also report "false" if the features are not set?

go.mod
```diff
@@ -5,15 +5,20 @@ go 1.16
 require (
 	github.com/container-storage-interface/spec v1.3.0
 	github.com/containers/image/v5 v5.9.0
+	github.com/containers/storage v1.23.7 // indirect
+	github.com/daixiang0/gci v0.2.9 // indirect
```
chrismuellner, line 9: Did you add this by accident?

waodim: Yes, I will remove it, thank you.

DTMad:
Please rework as discussed

DTMad:
As discussed internally we will adapt the behaviour so we report the usage within each component

PR #197, opened 2021-07-21
Poc/sqlite instead of fs for metadata for csi

0sewa0:


cmd/csidriver/main.go
```diff
@@ -87,8 +88,13 @@ func main() {
...
-		log.Error(err, "unable to create garbage collector directory for CSI Driver")
+	access := metadata.SqliteAccess{}
+	if err := access.Setup(); err != nil {
+		log.Error(err, "failed to setup database storage for CSI Driver")
+		os.Exit(1)
+	}
+	if err := metadata.CheckStorageCorrectness(mgr.GetClient(), &access, log); err != nil {
+		log.Error(err, "failed to correct database storage for CSI Driver")
 		os.Exit(1)
 	}
```
baichinger, line 99: If housekeeping fails for whatever reason, the driver will terminate. This might result in an outage which is not automatically recoverable.

controllers/csi/metadata/correctness.go
```diff
@@ -0,0 +1,65 @@
...
+		var pod corev1.Pod
+		if err := cl.Get(context.TODO(), client.ObjectKey{Name: podName}, &pod); !k8serrors.IsNotFound(err) {
+			continue
+		}
+		if err := access.DeleteVolume(podNames[podName]); err != nil {
+			return err
+		}
+		pruned++
+	}
+	log.Info("CSI volumes database is corrected", "prunedRows", pruned)
```
baichinger, line 42: Instead of logging a plain number it would be more valuable to log the orphan volumes/pods. Same below for tenants.

controllers/csi/metadata/sqlite.go
```diff
@@ -0,0 +1,308 @@
...
+
+	getDynakubesStatement = `
+	SELECT UUID, Dynakube
+	FROM tenants;
+	`
+)
+
+var (
+	log    = logger.NewDTLogger().WithName("storage")
+	dbPath = filepath.Join(dtcsi.DataPath, "csi.db")
```
baichinger, line 91: Overwriting this variable in tests is fragile. Would make it an argument to the constructor.

cmd/csidriver/main.go
```diff
@@ -87,8 +88,13 @@ func main() {
...
 
-	if err := fs.MkdirAll(filepath.Join(csiOpts.RootDir, dtcsi.GarbageCollectionPath), 0770); err != nil {
-		log.Error(err, "unable to create garbage collector directory for CSI Driver")
+	access := metadata.SqliteAccess{}
```
baichinger, line 91: Not using the constructor for this seems odd. As already pointed out, the initialization (including the setup step) of this object should be revised.

controllers/csi/metadata/metadata.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Return a new Volume if all fields are set.
+func NewVolume(id, podUID, version, tenantUUID string) *Volume {
+	if id == "" || podUID == "" || version == "" || tenantUUID == "" {
+		return nil
+	}
+	return &Volume{id, podUID, version, tenantUUID}
+}
+
+type Access interface {
```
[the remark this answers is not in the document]
0sewa0: My idea was that we always use it with the package name, so `metadata.Access` in which case it refers to a struct that can access the metadata. If we called it `Metadata` then we would use it as `metadata.Metadata` which is very redundant.

controllers/csi/metadata/sqlite.go
```diff
@@ -0,0 +1,306 @@
+package metadata
```
thschue, line 1: I think it would be a better solution to use a map or a custom resource to achieve this instead of bringing the complexity of this in there ...

thschue (changes requested):
Let's talk about the solution, there might be more reasonable ways to achieve this ...

controllers/csi/metadata/sqlite.go
0sewa0: Using maps (tenants/volume map) could work well up until `GetUsedVersions(tenantUUID string)`, you would have to iterate over all of the volume-map to find the versions associated with a tenant, a 3rd map just for the purpose of storing what versions does a tenant have would be yet again a bit annoying because you can't really store a key: list AFAIK (you can store the list as a string) in a configmap.

A smaller issue is that this would create 2(or 3) configmap/node, which feels a bit much 😄 
And creating a custom resource to solve this issue feels a bit more complex to me.

however, I'm open for suggestions 😃

controllers/csi/metadata/correctness.go
```diff
@@ -0,0 +1,67 @@
...
+	dynatracev1alpha1 "github.com/Dynatrace/dynatrace-operator/api/v1alpha1"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Checks if the entries in the storage are actually valid
+// Removes not valid entries
+func CheckStorageCorrectness(cl client.Client, access Access, log logr.Logger) error {
```
meik99, line 15: If it not only checks the storage but also removes invalid entries, it should be called something like `CorrectStorage`

controllers/csi/metadata/correctness.go
```diff
@@ -0,0 +1,67 @@
...
+		return err
+	}
+	if err := checkTenantsCorrectness(cl, access, log); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Removes volume entries if their pod is no longer exists
+func checkVolumesCorrectness(cl client.Client, access Access, log logr.Logger) error {
```
meik99, line 26: same as before, should be renamed to something like `correctVolumes`

controllers/csi/metadata/correctness.go
```diff
@@ -0,0 +1,67 @@
...
+			return err
+		}
+		pruned = append(pruned, volumeID+"|"+podName)
+	}
+	log.Info("CSI volumes database is corrected (volume|pod)", "prunedRows", pruned)
+	return nil
+}
+
+// Removes tenant entries if their dynakube no longer exists
+func checkTenantsCorrectness(cl client.Client, access Access, log logr.Logger) error {
```
meik99, line 48: same as before, should be renamed to something like `correctTenants`

controllers/csi/metadata/sqlite.go
```diff
@@ -0,0 +1,306 @@
...
+	a := SqliteAccess{}
+	err := a.Setup(path)
+	if err != nil {
+		log.Error(err, "Failed to connect to the database, err: %s", err.Error())
+		return nil, err
+	}
+	return &a, nil
+}
+
+// Connects to the database via the provided driver and path to the database.
```
meik99, line 106: unnecessary comment

controllers/csi/metadata/sqlite.go
```diff
@@ -0,0 +1,306 @@
...
+
+func (a *SqliteAccess) DeleteTenant(tenantUUID string) error {
+	err := a.executeStatement(deleteTenantStatement, tenantUUID)
+	if err != nil {
+		err = fmt.Errorf("couldn't delete tenant, UUID %s, err: %s", tenantUUID, err)
+	}
+	return err
+}
+
+// Gets a Tenant from the database, return (nil, nil) if the tenant is not in the database.
```
meik99, line 171: unnecessary comment

controllers/csi/metadata/sqlite.go
```diff
@@ -0,0 +1,306 @@
...
+	var dynakube string
+	err := a.querySimpleStatement(getTenantStatement, tenantUUID, &latestVersion, &dynakube)
+	if err != nil {
+		err = fmt.Errorf("couldn't get tenant for UUID %s, err: %s", tenantUUID, err)
+	}
+	return NewTenant(tenantUUID, latestVersion, dynakube), err
+}
+
+// Gets a Tenant from the database via its dynakube, return (nil, nil) if the tenant is not in the database.
+// Needed during NodePublishVolume.
```
meik99, line 183: unnecessary comment

controllers/csi/metadata/sqlite.go
```diff
@@ -0,0 +1,306 @@
...
+			volume.VolumeID,
+			volume.PodName,
+			volume.Version,
+			volume.TenantUUID,
+			err)
+	}
+	return err
+}
+
+// Gets a Volume from the database, return (nil, nil) if the volume is not in the database.
```
meik99, line 207: unnecessary comment

controllers/csi/metadata/sqlite.go
```diff
@@ -0,0 +1,306 @@
...
+
+func (a *SqliteAccess) DeleteVolume(volumeID string) error {
+	err := a.executeStatement(deleteVolumeStatement, volumeID)
+	if err != nil {
+		err = fmt.Errorf("couldn't delete volume VolumeID %s, err: %s", volumeID, err)
+	}
+	return err
+}
+
+// Gets all unique versions present in the `volumes` database in map.
```
meik99, line 227: unnecessary comment

controllers/csi/provisioner/reconciler.go
```diff
@@ -96,31 +104,52 @@ func (r *OneAgentProvisioner) Reconcile(ctx context.Context, request reconcile.R
...
+		}
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	}
 
 	return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 }
 
+func tenantChanged(old, new metadata.Tenant) bool {
```
meik99, line 149:
```suggestion
func hasTenantChanged(old, new metadata.Tenant) bool {
```

meik99 (changes requested):

DTMad (approved):

PR #198, opened 2021-07-21
Move unzip to klauspost/compress

chrismuellner:
- Add error for GetLatestAgent on non 2xx http status code
- Add test for GetLatestAgent
- Move unzip to klauspost/compress


DTMad (approved):

controllers/csi/provisioner/agent.go
```diff
@@ -57,85 +57,75 @@ func installAgent(installAgentCfg *installAgentConfig) error {
...
-		}
+			mode := file.Mode()
 
-		if zipf.FileInfo().IsDir() {
-			return fs.MkdirAll(path, mode)
-		}
+			// Mark all files inside ./agent/conf as group-writable
+			if file.Name != agentConfPath && strings.HasPrefix(file.Name, agentConfPath) {
+				mode |= 020
+			}
```
baichinger, line 103: Why do we need this?

chrismuellner: There is a [unit test](https://github.com/Dynatrace/dynatrace-operator/blob/master/controllers/csi/provisioner/agent_test.go#L201) that asserts this behavior, so I kept the same functionality.

PR #206, opened 2021-08-03
Bugfix/cloud native cfg map ns and dk

aorcholski:
ConfigMaps updates triggered by any change of a namespace or a DynaKube object

### DIFF
--- a/namespacesmapper/namespacesmapper.go
+++ b/namespacesmapper/namespacesmapper.go
@@ -3,6 +3,7 @@ package namespacesmapper
 import (
 	"context"
 	"fmt"
+
 	dynatracev1alpha1 "github.com/Dynatrace/dynatrace-operator/api/v1alpha1"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
--- a/namespacesmapper/namespacesmapper.go
+++ b/namespacesmapper/namespacesmapper.go
@@ -1,0 +1,270 @@
+package namespacesmapper
+
+import (
+	"context"
+	"fmt"
+	dynatracev1alpha1 "github.com/Dynatrace/dynatrace-operator/api/v1alpha1"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


namespacesmapper/namespacesmapper.go
```diff
@@ -0,0 +1,317 @@
+package namespacesmapper
```
DTMad, line 1: Since this file already gets pretty long and complex would it make sense to split it into two files - one for dynakubes, one for namespaces?

namespacesmapper/dynakubes.go
```diff
@@ -0,0 +1,134 @@
...
+	diErr := doMapFromDynaKube(ctx, clt, opns, dataIngestMapName, dk,
+		func(dk dynatracev1alpha1.DynaKube) *metav1.LabelSelector {
+			return dk.Spec.DataIngestSpec.Selector
+		}, nsList)
+
+	cmErr := doMapFromDynaKube(ctx, clt, opns, codeModulesMapName, dk,
+		func(dk dynatracev1alpha1.DynaKube) *metav1.LabelSelector {
+			return dk.Spec.CodeModules.Selector
+		}, nsList)
+
```
meik99, line 77: I don't really like how this file handles mapping for data ingest **and** codemodules while still being specifically made for those two. would it be possible, for example, to abstract common code and then create more specific code for either data ingest and code modules?

namespacesmapper/namespaces.go
```diff
@@ -0,0 +1,163 @@
...
+			return dk.Spec.DataIngestSpec.Selector
+		})
+
+	cmErr := doMapFromNamespace(ctx, clt, opns, ns, codeModulesMapName,
+		func(dk dynatracev1alpha1.DynaKube) bool {
+			return dk.Spec.CodeModules.Enabled
+		},
+		func(dk dynatracev1alpha1.DynaKube) *metav1.LabelSelector {
+			return dk.Spec.CodeModules.Selector
+		})
```
meik99, line 98: Same as before, pasting the previous comment for reference.

> I don't really like how this file handles mapping for data ingest and codemodules while still being specifically made for those two. would it be possible, for example, to abstract common code and then create more specific code for either data ingest and code modules?

meik99 (changes requested):
Missing unit tests for namespacesmapper/dynakubes.go, namespacesmapper/mapper.go and not enough for namespacesmapper/namespaces.go.

PR #210, opened 2021-08-06
implemented validation webhook for dynakube objects

meik99:


cmd/operator/main.go
```diff
@@ -77,7 +77,7 @@ func main() {
 
 	signalHandler := ctrl.SetupSignalHandler()
 
-	startWebhookAndBootstrapperIfDebugFlagSet(startupInfo{
+	startWebhookIfDebugFlagSet(startupInfo{
 		cfg:           cfg,
```
gkrenn, line 81: In addition to remove the bootstrapper from the function name we should remove the function `startBootstrapperManager` too as the bootstrapper no longer exists

config/common/operator/deployment-operator.yaml
```diff
@@ -25,7 +25,7 @@ spec:
         - name: dynatrace-operator
           args:
             - operator
-          image: quay.io/dynatrace/dynatrace-operator:snapshot
+          image: quay.io/dynatrace/dynatrace-operator:snapshot-validating-webhook
```
DTMad, line 28: I guess this has been added by accident - please remove

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/containers/image/v5 v5.9.0
 	github.com/evanphx/json-patch v4.9.0+incompatible
 	github.com/go-logr/logr v0.3.0
+	github.com/google/go-licenses v0.0.0-20210727195037-24e2421190d8 // indirect
```
DTMad, line 10: Did you add this on purpose? 
What is it used for?

DTMad (changes requested):

webhook/validation/validation.go
```diff
@@ -0,0 +1,103 @@
...
+	clt    client.Client
+}
+
+// InjectClient implements the inject.Client interface which allows the manager to inject a kubernetes client into this handler
+func (validator *dynakubeValidator) InjectClient(clt client.Client) error {
+	validator.clt = clt
+	return nil
+}
+
+func (validator *dynakubeValidator) Handle(_ context.Context, request admission.Request) admission.Response {
```
0sewa0, line 49: We have the following in `oneagent_utils.go`:
```go
func validate(cr *dynatracev1alpha1.DynaKube) error {
	var msg []string
	if cr.Spec.APIURL == "" {
		msg = append(msg, ".spec.apiUrl is missing")
	}
	if len(msg) > 0 {
		return errors.New(strings.Join(msg, ", "))
	}
	return nil
}
```
checking this here would nice. (so we could remove it from the oneagent_controller)

build/deploy_local.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+if [[ ! "${TAG}" ]]; then
```
chrismuellner, line 3: `make deploy-local-easy` already does something similar.
We should decide on one of the 2

go.mod
meik99: That has probably been added by the deploy_local script, since it installs go-licenses, which generates the third-party licenses of the libraries we use. Those licenses are then expected by our Dockerfile and bundled with the image, before it is pushed to quay

webhook/validation/validation.go
meik99: Do you mean moving the apiUrl validation from oneagent_utils to the new webhook logic?

build/deploy_local.sh
meik99: `make deply-local` generates an image given a tag and complains if no tag is set. `make deploy-local-easy` generates a tag from the git branch and sets it, before calling the deploy_local script. imo those two can coexist. 

The first use case is, you want to generate an image and care about the tag and want to use a custom one, or you want two images with different tags from the same branch, with only minor changes.
The second one is, you don't care about the tag and just want an image based on whatever is the current branch-name.

cmd/operator/main.go
meik99: removed

webhook/validation/validation.go
meik99: moved functionality

config/common/operator/deployment-operator.yaml
meik99: removed

cmd/operator/manager.go
```diff
@@ -0,0 +1,48 @@
+package main
```
DTMad, line 1: Please add the license header

cmd/operator/validation.go
```diff
@@ -0,0 +1,23 @@
+package main
```
DTMad, line 1: Please add the license header

cmd/operator/watcher.go
```diff
@@ -0,0 +1,81 @@
+package main
```
DTMad, line 1: Please add the license header

DTMad (changes requested):
Do we need a separate pod for the validation webhook?
Wouldn't it be possible to have it running as a part of the operator to avoid additional resources?

cmd/operator/manager.go
meik99: added

cmd/operator/validation.go
meik99: added

cmd/operator/watcher.go
meik99: added

cmd/operator/validation.go
meik99: added

PR #225, opened 2021-08-13
Moves logic in init.sh into functions

0sewa0:


controllers/namespace/init.sh.tmpl
```diff
@@ -23,89 +24,120 @@ im_nodes=(
...
+			fi
+			writeToConfFile ${conf_host} ${container_conf_file}
+		fi
+	done
+}
+
+####### MAIN #######
+printf "\nSetting fail code\n"
+setFailCode
+printf "\Downloading OneAgent\n"
```
waodim, line 140: Maybe it's a nitpick, but I think a \n is missing.

Besides that everything looks fine to me.

DTMad (dismissed):

chrismuellner (approved):

PR #227, opened 2021-08-18
Added debug command to makefile

gkrenn:
Added easy way to debug webhook

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,29 @@ test: generate fmt vet manifests
 manager: generate fmt vet
 	go build -o bin/manager ./cmd/operator/
 
+manager-amd64: export GOOS=linux
+manager-amd64: export GOARCH=amd64
+manager-amd64: generate fmt vet
+	go build -o bin/manager-amd64 ./cmd/operator/
+
+WEBHOOK_NS=dynatrace
+
+debug-webhook-setup:
+	@echo Preparing webhook deployment for debugging
+	kubectl -n $(WEBHOOK_NS) patch deploy dynatrace-webhook -p '{"spec":{"template":{"spec":{"containers":[{"name":"webhook", "image":"quay.io/gkrenn/testoperator:operatorwithtar","command":["/bin/bash", "-c", "sleep 999999999"]}]}}}}'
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -46,6 +46,29 @@ test: generate fmt vet manifests
...
+manager-amd64: export GOOS=linux
+manager-amd64: export GOARCH=amd64
+manager-amd64: generate fmt vet
+	go build -o bin/manager-amd64 ./cmd/operator/
+
+WEBHOOK_NS=dynatrace
+
+debug-webhook-setup:
+	@echo Preparing webhook deployment for debugging
+	kubectl -n $(WEBHOOK_NS) patch deploy dynatrace-webhook -p '{"spec":{"template":{"spec":{"containers":[{"name":"webhook", "image":"quay.io/gkrenn/testoperator:operatorwithtar","command":["/bin/bash", "-c", "sleep 999999999"]}]}}}}'
```
DTMad, line 58: Please don't use images from your private quay repos here

DTMad: sleep 999999999 ?

Makefile
```diff
@@ -46,6 +46,29 @@ test: generate fmt vet manifests
...
+	kubectl -n $(WEBHOOK_NS) patch deploy dynatrace-webhook --type json -p '[{"op": "remove", "path": "/spec/template/spec/containers/0/readinessProbe"}]'
+	@echo
+	@echo Please wait till the webhook is running again
```
DTMad, line 61: There's kubectl wait command - probably we could make use of this here?

DTMad (changes requested):

Makefile
gkrenn: I'll move it to our dynatrace-operator quay repo with a special tag

sleep is used as a placeholder application to prevent the container from restarting.
If we wouldn't provide a new command it would start the webhook provided in the image.

Makefile
gkrenn: Thanks, will have a look at that

gkrenn: Not possible as we would need to know the name of the pod we are waiting for

gkrenn: `sleep 3 && kubectl wait --for=condition=ready --timeout=60s -n $(WEBHOOK_NS) pod/$(shell (kubectl -n $(WEBHOOK_NS) get pods | awk '/webhook/ && /Running/' | cut -d" " -f1))` - would also not be possible as shell always gets executed before anything else

PR #228, opened 2021-08-20
improve readonly support

meik99:


build/deploy_local.sh
```diff
@@ -13,7 +13,7 @@ go_build_args=(
 base_image="dynatrace-operator"
 out_image="quay.io/dynatrace/dynatrace-operator:${TAG}"
 
-CGO_ENABLED=1 go build "${go_build_args[@]}" -o ./build/_output/bin/dynatrace-operator ./cmd/operator/
+go build "${go_build_args[@]}" -o ./build/_output/bin/dynatrace-operator ./cmd/operator/
```
0sewa0, line 16: won't this break the deploy for the CSI driver ?

meik99: no, because travis deploys using deploy.sh, not deply_local.sh

0sewa0: yes, but this will break the local deploy :)

PR #239, opened 2021-09-02
remove beta affinity for Kubernetes >= 1.14

meik99:


controllers/oneagent/daemonset/affinity.go
```diff
@@ -0,0 +1,36 @@
...
+	return []corev1.NodeSelectorTerm{
+		kubernetesArchOsSelectorTerm(),
+	}
+}
+
+func kubernetesArchOsSelectorTerm() corev1.NodeSelectorTerm {
+	return corev1.NodeSelectorTerm{
+		MatchExpressions: []corev1.NodeSelectorRequirement{
+			{
+				Key:      "kubernetes.io/arch",
```
chrismuellner, line 25: We have constants for these in `activegate/reconciler/statefulset/statefulset.go` so we could reuse these in some way.

Can we remove the beta affinity for that `statefulset` as well?

chrismuellner (changes requested):

controllers/oneagent/daemonset/affinity.go
meik99: removed from other deployments and refactored

chrismuellner (approved):

PR #240, opened 2021-09-02
TLS certificate can be set for AG jetty server to avoid self signed certs usage

mjgrzybek:


api/v1alpha1/dynakube_types.go
```diff
@@ -87,6 +87,12 @@ type ActiveGateSpec struct {
 	// implementation from the Dynatrace environment set as API URL.
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Image",order=10,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:text"}
 	Image string `json:"image,omitempty"`
+
+	// Optional: the name of a secret containing ActiveGate TLS cert+key and password. If not set, self-signed certificate is used.
+	// server.p12: certificate+key pair in pkcs12 format
+	// password: passphrase to read server.p12
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="TlsSecretName",order=10,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:text"}
+	TlsSecretName string `json:"TlsSecretName,omitempty"`
```
meik99, line 95:
```suggestion
	TlsSecretName string `json:"tlsSecretName,omitempty"`
```

config/crd/default/bases/dynatrace.com_dynakubes.yaml
```diff
@@ -51,6 +51,11 @@ spec:
               activeGate:
                 description: General configuration about ActiveGate instances
                 properties:
+                  TlsSecretName:
```
meik99, line 54:
```suggestion
                  tlsSecretName:
```

controllers/activegate/capability/capability.go
```diff
@@ -146,10 +190,14 @@ func NewRoutingCapability(crProperties *dynatracev1alpha1.CapabilityProperties)
 			},
 		},
 	}
+
+	setTlsConfig(agSpec, &c.capabilityBase)
```
meik99, line 194: please add unit tests to check whether tls config is applied correctly if a secret name is provided and that it is not applied if no secret name is given.

controllers/activegate/capability/capability.go
```diff
@@ -126,14 +132,52 @@ func NewKubeMonCapability(crProperties *dynatracev1alpha1.CapabilityProperties)
...
+			},
 		},
 	}
+
+	setTlsConfig(agSpec, &c.capabilityBase)
+
+	return c
+}
+
+func setTlsConfig(agSpec *dynatracev1alpha1.ActiveGateSpec, c *capabilityBase) {
```
meik99, line 145: would be nicer as a method with a pointer receiver. E.g.:
`func (c *capabilityBase) setTlsConfig(agSpec)`

meik99 (changes requested):

controllers/activegate/capability/capability.go
mjgrzybek: Added UT for first scenario.
Regarding second scenario - just by definition, secret cannot be applied if no secret name is given.

meik99: my use case for the second scenario is the following:

In the future, some developer is working on the setTlsConfig.
They change some stuff here, maybe refactor something there.
Then they forget to reinstate the  check `if agSpec.TlsSecretName != "" {`
Now a VolumeSource is always created, even with an empty SecretName, breaking the deployment of the capability

In my mind, that is a mistake that is easily catch-able by a simple unit test.

mjgrzybek: Ok, I get it now. It's already covered in `"default"` [case](https://github.com/Dynatrace/dynatrace-operator/pull/240/files#diff-ab178ad5f17da576726643ba0e3f6e9c442efc4f2dc621d2d4c40267f2108f29R256), where `want` has no `VolumeMount` set (`nil`).
Test [fails](https://github.com/Dynatrace/dynatrace-operator/pull/240/files#diff-ab178ad5f17da576726643ba0e3f6e9c442efc4f2dc621d2d4c40267f2108f29R313) if `VolumeMount` is created because `reflect.DeepEqual` returns false as `VolumeMount` was not expected.

meik99 (approved):

PR #246, opened 2021-09-13
Fix webhook certificates with existing secret

chrismuellner:
* Check webhook configurations for certificate (and not just secret)
* Add unit test to check scenario

### DIFF
--- a/controllers/certificates/webhook_reconciler.go
+++ b/controllers/certificates/webhook_reconciler.go
@@ -217,7 +217,7 @@ func (r *ReconcileWebhookCertificates) getDomain() string {
 	return fmt.Sprintf("%s.%s.svc", webhook.DeploymentName, r.namespace)
 }
 
-// checkMutatingWebhookConfigurations loads webhook configurations and checks certificates exist and are valid
+// checkMutatingWebhookConfigurations checks certificates exist and are valid
 func (r *ReconcileWebhookCertificates) checkMutatingWebhookConfigurations(
 	mutatingWebhookConfiguration *admissionregistrationv1.MutatingWebhookConfiguration,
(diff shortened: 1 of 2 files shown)


controllers/certificates/webhook_reconciler.go
```diff
@@ -83,20 +83,36 @@ func (r *ReconcileWebhookCertificates) Reconcile(ctx context.Context, request re
...
+	mutatingWebhookConfiguration, err := r.getMutatingWebhookConfiguration(ctx)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	validatingWebhookConfiguration, err := r.getValidatingWebhookConfiguration(ctx)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	hasMissingCertificate := r.checkMutatingWebhookConfigurations(
```
gkrenn, line 95: Would even check if the CABundle has the right public certificate

gkrenn (changes requested):
See my single comment

controllers/certificates/webhook_reconciler.go
```diff
@@ -206,13 +217,32 @@ func (r *ReconcileWebhookCertificates) getDomain() string {
 	return fmt.Sprintf("%s.%s.svc", webhook.DeploymentName, r.namespace)
 }
 
+// checkMutatingWebhookConfigurations loads webhook configurations and checks certificates exist and are valid
```
gkrenn, line 220:
```suggestion
// checkMutatingWebhookConfigurations checks certificates exist and are valid
```

gkrenn (changes requested):

gkrenn (approved):

PR #224, opened 2021-08-13
Add resource requests and limits to CSI DaemonSet

chrismuellner:
csi driver now uses:
```yaml
requests:
  cpu: 200m
  memory: 100Mi
limits:
  cpu: 200m
  memory: 100Mi
```

PR #247, opened 2021-09-13
Set resource limits/requests for csi daemonset to same values

chrismuellner:
* Improve QoS for CSI Driver by setting same values on limits/requests (same as: #224)
* Update manifests to match code


waodim (approved):
Seems fine to me.

PR #254, opened 2021-09-17
refactor: added files to generate a crd for a v1-DynaKube as preparation for cr simplification

meik99:


chrismuellner (commented):
Should `v1alpha1` get a [deprecation warning](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-deprecation)?

meik99:
> Should `v1alpha1` get a [deprecation warning](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-deprecation)?

While I do agree, I couldn't find any info on a kubebuilder or operator annotation to generate this field. Manually setting it would be pointless, as it would be overwritten every time we re-generat the crd

0sewa0:
nitpick: why are we skipping "v1beta1" ? :)

0sewa0 (approved):

PR #262, opened 2021-09-24
Add resources to all csi driver containers

chrismuellner:
* Move `driver` limits/requests to same constant
* Add default resource requests/limits to `registrar` and `liveness-probe` container

If no mismatched values are configured via the `csi-resources` annotation, the csi driver pod now gets the `qosClass` of `Guaranteed`

### DIFF
--- a/controllers/csi/daemonset.go
+++ b/controllers/csi/daemonset.go
@@ -231,16 +231,7 @@ func prepareDriverResources(client client.Client, operatorNS string, logger logr
 		}
 	}
 
-	return corev1.ResourceRequirements{
-		Requests: corev1.ResourceList{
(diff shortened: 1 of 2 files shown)


controllers/csi/daemonset.go
```diff
@@ -26,10 +26,14 @@ const (
...
-	MaxCPU    = 200
-	MaxMemory = 100
+	DriverDefaultCPU    = 200
+	DriverDefaultMemory = 100
+
+	RegistrarDefaultCPU    = 5
+	RegistrarDefaultMemory = 10
+
+	LivenessProbeDefaultCPU    = 5
+	LivenessProbeDefaultMemory = 10
```
meik99, line 36: do these constants have to be exposed?

meik99 (approved):
nice refactoring

PR #265, opened 2021-09-29
Adding cloud native full stack and renaming code_modules

luhi-DT:
Changes are just to add the deployment types we have, to the metadata const. Afterwards the new DeploymentTypeCN (Cloud Native Fullstack) can be used.


deploymentmetadata/deploymentmetadata.go
```diff
@@ -15,10 +15,11 @@ const (
...
 
-	DeploymentTypeCodeModules = "code_modules"
-	DeploymentTypeFS          = "classic_fullstack"
-	DeploymentTypeHM          = "host_monitoring"
-	DeploymentTypeAG          = "active_gate"
+	DeploymentTypeAM = "application_monitoring"
+	DeploymentTypeFS = "classic_fullstack"
+	DeploymentTypeCN = "cloud_native_fullstack"
+	DeploymentTypeHM = "host_monitoring"
+	DeploymentTypeAG = "active_gate"
```
baichinger, line 22: Full names for those constants would greatly improve readability.

luhi-DT: That's true, I will rename them.

0sewa0 (approved):
👍

PR #266, opened 2021-09-29
Implements the new activegate section

0sewa0:
Implements the new activeGate section where each capability we want to turn on is put into a list, but the rest of the config they share, and capabilities get deployed in 1 pod in 1 container. If the customer wants the capabilities to have their on pods then multiple Dynakube need to be created.


aorcholski (changes requested):
"make deploy" command does not create serviceaccount/dynatrace-multi so statfulset reports an error.
dynakube object :
```
kind: DynaKube                                                                                                                                                                                                     
metadata:                                                                                                                                                                                                          
  name: dk                                                                                                                                                                                                         
  namespace: dynatrace                                                                                                                                                                                             
spec:                                                                                                                                                                                                              
  apiUrl: https://...
  activeGate:                                                                                                                                                                                                      
    capabilities:                                                                                                                                                                                                  
      - "routing" 
```

api/v1beta1/activegate_types.go
```diff
@@ -4,19 +4,62 @@ import (
...
-const (
-	Routing    ActiveGateCapability = "routing"
-	KubeMon    ActiveGateCapability = "kubernetes-monitoring"
-	DataIngest ActiveGateCapability = "data-ingest"
+type ActiveGateCapability struct {
+
+	// The name of the capability known by the user, mainly used in the CR
+	DisplayName CapabilityDisplayName
+
+	// The name of used for marking the pod for given capability
```
chrismuellner, line 14:
```suggestion
	// The name used for marking the pod for given capability
```

api/v1beta1/activegate_types.go
```diff
@@ -4,19 +4,62 @@ import (
...
+	}
 )
 
 type ActiveGateSpec struct {
 
 	// Activegate capabilities enabled (routing, kubernetes-monitoring, data-ingest)
-	Capabilities []ActiveGateCapability `json:"capabilities,omitempty"`
+	Capabilities []CapabilityDisplayName `json:"capabilities,omitempty"`
+
+	// Amount of replicas for your DynaKube
```
chrismuellner, line 46:
```suggestion
	// Amount of replicas for the ActiveGate
```

api/v1beta1/dynakube_conversion.go
```diff
@@ -183,8 +183,6 @@ func (dst *DynaKube) ConvertFrom(srcRaw conversion.Hub) error {
 }
 
 func convertToDeprecatedActiveGateCapability(dst *CapabilityProperties, src *v1alpha1.CapabilityProperties) {
-	dst.Enabled = true
```
chrismuellner, line 186: Is this not needed?

config/crd/default/patches/webhook_in_dynakubes.yaml
```diff
@@ -13,4 +13,4 @@ spec:
           name: dynatrace-webhook
           path: /convert
       conversionReviewVersions:
-      - v1
+      - v1beta1
```
chrismuellner, line 16: does not belong in this PR

config/common/activegates/serviceaccount-activegates.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: dynatrace-activegates
```
chrismuellner, line 4: Wouldn't `dynatrace-activegate` fit better (similar to the other Service Accounts)?

controllers/activegate/capability/capability.go
```diff
@@ -57,12 +73,12 @@ func (c *capabilityBase) GetConfiguration() Configuration {
 	return c.Configuration
 }
 
-func (c *capabilityBase) GetModuleName() string {
-	return c.moduleName
+func (c *capabilityBase) GetShortName() string {
```
chrismuellner, line 76:
```suggestion
func (c *capabilityBase) ShortName() string {
```

controllers/activegate/capability/capability.go
```diff
@@ -57,12 +73,12 @@ func (c *capabilityBase) GetConfiguration() Configuration {
...
+	return c.shortName
 }
 
-func (c *capabilityBase) GetCapabilityName() string {
-	return c.capabilityName
+func (c *capabilityBase) GetArgName() string {
```
chrismuellner, line 80:
```suggestion
func (c *capabilityBase) ArgName() string {
```

chrismuellner (changes requested):

config/common/activegates/serviceaccount-activegates.yaml
0sewa0: but this can have multiple activegates inside :)

controllers/activegate/capability/capability.go
0sewa0: then I should redo all the other functions in the interface, which will result in more files changed, if you are up for it then I will do it 😛

config/common/activegates/serviceaccount-activegates.yaml
chrismuellner: isn't it still one activegate that has multiple capabilities?

config/crd/default/patches/webhook_in_dynakubes.yaml
0sewa0: its already in master, so removing it here does nothing

config/common/activegates/serviceaccount-activegates.yaml
0sewa0: 😞 I hate naming things

api/v1beta1/dynakube_conversion.go
0sewa0: Its not part of CapabilityProperties, moved it out before the func call

aorcholski (approved):

chrismuellner (approved):

controllers/activegate/capability/capability.go
```diff
@@ -30,169 +40,235 @@ type Configuration struct {
...
 		},
 	}
+	if dk == nil || !dk.ActiveGateMode() {
+		return &mc
+	}
+	mc.enabled = true
+	mc.properties = &dk.Spec.ActiveGate.CapabilityProperties
+	capabilityNames := []string{}
+	for _, capName := range dk.Spec.ActiveGate.Capabilities {
+		cap := activeGateCapabilities[capName]()
```
aorcholski, line 155: There is no protection against invalid DisplayName in [capName]. Operator dies:
```pod/dynatrace-operator-7b66c9cdf9-8czlm   0/1     CrashLoopBackOff   5 (61s ago)   9m5s```

```E1001 11:21:16.848245       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)```

I added 'dummy' capability using 'kubectl edit mydynakube' command:

```
spec:
  activeGate:
    capabilities:
    - kubernetes-monitoring
    - dummy
```

**Review by aorcholski — CHANGES_REQUESTED** [review]

File: controllers/activegate/capability/capability.go (hunk 8 above)
Comment (chrismuellner) [reply controllers/activegate/capability/capability.go:155]: This will be fixed with an updated validation webhook

Comment (0sewa0) [reply controllers/activegate/capability/capability.go]: yes, the validation webhook will catch this, but I also added a check so in case we want to support this on openshift 3.11 (that doesn't have webhooks) we can do so

File: controllers/activegate/capability/capability.go [hunk 9]
```diff
@@ -152,7 +152,12 @@ func NewMultiCapability(dk *dynatracev1beta1.DynaKube) *MultiCapability {
 	mc.properties = &dk.Spec.ActiveGate.CapabilityProperties
 	capabilityNames := []string{}
 	for _, capName := range dk.Spec.ActiveGate.Capabilities {
-		cap := activeGateCapabilities[capName]()
+		capabilityGenerator, ok := activeGateCapabilities[capName]
+		if !ok {
+			continue
```
chrismuellner, line 157: I guess ignoring invalid capabilities is fine

chrismuellner (approved):

PR #270, opened 2021-09-30
build: replace kustomize generation of manifest files with Helm generation

meik99:


chrismuellner:
Will be done in a different PR

PR #276, opened 2021-10-04
Add log message for monitored namespaces for cloudNative and appOnly

luhi-DT:


mapper/mapper.go
```diff
@@ -112,6 +114,7 @@ func updateLabels(matches bool, dynakube *dynatracev1beta1.DynaKube, namespace *
 		if !ok || oldDkName != dynakube.Name {
 			updated = true
 			addNamespaceInjectLabel(dynakube.Name, namespace)
+			log.Info("Started monitoring Namespace", "Namespace", namespace.Name, "Dynakube", dynakube.Name)
```
DTMad, line 117:
```suggestion
			log.Info("Started monitoring namespace", "Namespace", namespace.Name, "Dynakube", dynakube.Name)
```

mapper/mapper.go
```diff
@@ -102,7 +104,7 @@ func updateNamespace(namespace *corev1.Namespace, dkList *dynatracev1beta1.DynaK
 	return updated, nil
 }
 
-func updateLabels(matches bool, dynakube *dynatracev1beta1.DynaKube, namespace *corev1.Namespace) (bool, error) {
+func updateLabels(matches bool, dynakube *dynatracev1beta1.DynaKube, namespace *corev1.Namespace, log logr.Logger) (bool, error) {
```
meik99, line 107: since this function has a lot of parameters now, consider refactoring some or all of them into a struct and using that as a parameter

the same or a similar struct should also be used for the `updateNamespace` function, as it too has a lot of paramteres

meik99 (dismissed):

mapper/mapper.go
0sewa0: I'm not a fan of creating structs just to pass them as variables, you are just moving the bloat to a different place.

furthermore it kinda asks the question "why isn't this a method of the struct ?"

luhi-DT: Imho this would be a little overkill, since I would need two structs one for `updateNamespace` and one for `update Labels`. Futhermore I think it would add more complexity to the code, just for a shorter parameter list.

PR #284, opened 2021-10-05
Provides the hostGroup to the special agents

0sewa0:


initgeneration/initgeneration.go
```diff
@@ -168,9 +170,29 @@ func (g *InitGenerator) prepareScriptForDynaKube(dk *dynatracev1beta1.DynaKube,
...
+func getHostGroup(dk *dynatracev1beta1.DynaKube) string {
+	var hostGroup string
+	if dk.CloudNativeFullstackMode() && dk.Spec.OneAgent.CloudNativeFullStack.Args != nil {
+		for _, arg := range dk.Spec.OneAgent.CloudNativeFullStack.Args {
+			split := strings.Split(arg, "=")
+			if len(split) != 2 {
+				continue
+			}
+			key := split[0]
+			value := split[1]
```
meik99, line 186: consider refactoring these lines into their own function

meik99 (dismissed):

0sewa0:
we decided against using a separate field in the CR, because that would let the customer set it in 2 different places (in args and in the new field), which would just introduce more logic trying to make sure that the user doesn't do anything dumb.

PR #291, opened 2021-10-07
Additional dimensions for data ingest

aorcholski:
Pods enriched with additional dims for mint:
- dt.kubernetes.workload.kind
- dt.kubernetes.workload.name
Webhook recursively iterates over the list of OwnerReferences to find the top most Controller to report it in dt_metadata.properties(-.json) files.

```	// There cannot be more than one managing controller.```
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L249

### CODE REVIEW

File: config/common/webhook/role-webhook.yaml [hunk 1]
```diff
@@ -50,3 +50,4 @@ rules:
       - get
       - update
       - create
+
```
meik99, line 53:
```suggestion

```

initgeneration/init.sh.test-sample
```diff
@@ -100,6 +100,25 @@ tenant ${host_tenant}
...
+
+  echo "{
+\"k8s.pod.uid\": \"${K8S_PODUID}\",
+\"k8s.pod.name\": \"${K8S_PODNAME}\",
+\"k8s.namespace.name\": \"${K8S_NAMESPACE}\",
+\"dt.kubernetes.workload.kind\": \"${DT_WORKLOAD_KIND}\",
+\"dt.kubernetes.workload.name\": \"${DT_WORKLOAD_NAME}\",
+\"dt.kubernetes.cluster.id\": \"${cluster_id}\"
+}" >> "${1}/${2}.json"
+}
```
meik99, line 120: would it be possible to use `jq` here or is that not available on the image?

initgeneration/init.sh.tmpl
```diff
@@ -101,6 +101,25 @@ tenant ${host_tenant}
...
+dt.kubernetes.cluster.id ${cluster_id}" >> "${1}/${2}.properties"
+
+  echo "{
+\"k8s.pod.uid\": \"${K8S_PODUID}\",
+\"k8s.pod.name\": \"${K8S_PODNAME}\",
+\"k8s.namespace.name\": \"${K8S_NAMESPACE}\",
+\"dt.kubernetes.workload.kind\": \"${DT_WORKLOAD_KIND}\",
+\"dt.kubernetes.workload.name\": \"${DT_WORKLOAD_NAME}\",
+\"dt.kubernetes.cluster.id\": \"${cluster_id}\"
+}" >> "${1}/${2}.json"
```
meik99, line 120: would it be possible to use `jq` here or is that not available on the image?

webhook/mutation/pod_mutator.go
```diff
@@ -127,6 +127,46 @@ type podMutator struct {
...
+	for _, owner := range om.OwnerReferences {
+		if owner.Controller != nil && *owner.Controller {
+			obj := &v1.PartialObjectMetadata{
+				TypeMeta: v1.TypeMeta{
+					APIVersion: owner.APIVersion,
+					Kind:       owner.Kind,
+				},
+			}
+			if err := cnt.Get(ctx, client.ObjectKey{Name: owner.Name, Namespace: om.Namespace}, obj); err != nil {
+				log.Error(err, "Failed to query the object", "apiVersion", owner.APIVersion, "kind", owner.Kind, "name", owner.Name, "namespace", om.Namespace)
```
meik99, line 161:
```suggestion
				log.Error(err, "failed to query the object", "apiVersion", owner.APIVersion, "kind", owner.Kind, "name", owner.Name, "namespace", om.Namespace)
```

initgeneration/init.sh.tmpl
aorcholski: jq command is not available

webhook/mutation/pod_mutator.go
aorcholski: small letter used

initgeneration/init.sh.test-sample
```diff
@@ -135,6 +154,7 @@ createConfigurationFiles() {
 		if [[ "${host_tenant}" != "-" ]]; then
 			if [[ "abc12345" == "${host_tenant}" ]]; then
 				writeConfK8To ${container_conf_file}
+				writeConfEnrichmentTo "/var/lib/dynatrace/enrichment" "dt_metadata"
```
0sewa0, line 157: maybe dumb question:
do you only want to do `enrichment` in case of cloudNativeFullstack ?
because  `${host_Tenant}` is only set to anything in case of cloudNativeFullstack and you could have activegates without cloudNativeFullstack

aorcholski: Right, enrichment configuration should be created in both CloudNativeFullStack and ApplicationMonitoring mode.

initgeneration/init.sh.tmpl
```diff
@@ -116,6 +116,25 @@ tenant ${host_tenant}
...
+writeConfEnrichmentTo() {
+  mkdir -p "${1}"
+  echo "k8s.pod.uid ${K8S_PODUID}
+k8s.pod.name ${K8S_PODNAME}
+k8s.namespace.name ${K8S_NAMESPACE}
+dt.kubernetes.workload.kind ${DT_WORKLOAD_KIND}
+dt.kubernetes.workload.name ${DT_WORKLOAD_NAME}
+dt.kubernetes.cluster.id ${cluster_id}" >> "${1}/${2}.properties"
+
+  echo "{
```
chrismuellner, line 128: Just curious here, why do you need the same data in a .properties file/format as well as .json?

mjgrzybek: That's external requirement, see: (OneAgent file-open interface)[https://www.dynatrace.com/support/help/how-to-use-dynatrace/metrics/metric-ingestion/ingestion-methods/enrich-metrics/#oneagent-file-open-interface]

chrismuellner: the link says it's fine to have either.
is the documentation wrong or is one format enough? :smile:

kkreja: I would say that the docs are OK. If you go for .json the indirection file will point you to a .json file if you go for .properties you will get redirected to a .properties file. We support both formats as we want to give some freedom to customers how they would like to use it.

0sewa0: but is there any reason we would like to provide this freedom for the user? 
I already see the confused customer questions asking what is the difference between the 2 approaches or even weirder questions. (I have experience with people confused by multiple options to do the same thing)

IMO we shouldn't provide multiple configuration options for doing the same thing, without any concrete usecase which tells us the user is only able to use one of the possible ways configure and not the other(s). 
(fewer ways to configure ==> simpler docs + simple implementation ==> less confused users + less buggy code)

initgeneration/init.sh.tmpl
```diff
@@ -116,6 +116,25 @@ tenant ${host_tenant}
...
+\"k8s.pod.uid\": \"${K8S_PODUID}\",
+\"k8s.pod.name\": \"${K8S_PODNAME}\",
+\"k8s.namespace.name\": \"${K8S_NAMESPACE}\",
+\"dt.kubernetes.workload.kind\": \"${DT_WORKLOAD_KIND}\",
+\"dt.kubernetes.workload.name\": \"${DT_WORKLOAD_NAME}\",
+\"dt.kubernetes.cluster.id\": \"${cluster_id}\"
+}" >> "${1}/${2}.json"
```
baichinger, line 135: Appending to a properties file is okay. Appending to a JSON file is not. Is someone else allowed to contribute to these files? I propose to have clear ownership and therefore overwrite them.

aorcholski: First of all dt_metadata.properties and dt_metadata.json files must be consistent with each other so both files are overwritten.

initgeneration/init.sh.tmpl
chrismuellner: since it's an external requirement, it's fine to have it duplicated

webhook/mutation/pod_mutator.go
```diff
@@ -117,14 +124,81 @@ func registerHealthzEndpoint(mgr manager.Manager) {
...
+			Name: pod.ObjectMeta.Name,
+			// pod.ObjectMeta.Namespace is empty yet
+			Namespace:       namespace,
+			OwnerReferences: pod.ObjectMeta.OwnerReferences,
+		},
+	}
+	return rootOwner(ctx, cnt, obj)
+}
+
+func rootOwner(ctx context.Context, cnt client.Client, o *v1.PartialObjectMetadata) (string, string, error) {
```
chrismuellner, line 154: use `clt` for the client variable

webhook/mutation/pod_mutator.go
```diff
@@ -237,6 +311,11 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
 	}
 	pod.Annotations[dtwebhook.AnnotationInjected] = "true"
 
+	workloadName, workloadKind, workloadErr := rootOwnerPod(ctx, m.metaClient, pod, req.Namespace)
```
chrismuellner, line 314: would a name like `findRootOwnerPod(...)` improve readability here?

chrismuellner, line 314: is there a reason you don't reuse `err` here?

webhook/mutation/pod_mutator.go
```diff
@@ -117,14 +124,81 @@ func registerHealthzEndpoint(mgr manager.Manager) {
...
+		{Kind: "Deployment", APIVersion: "apps/v1"},
+		{Kind: "ReplicationController", APIVersion: "v1"},
+		{Kind: "StatefulSet", APIVersion: "apps/v1"},
+		{Kind: "DaemonSet", APIVersion: "apps/v1"},
+		{Kind: "Job", APIVersion: "batch/v1"},
+		{Kind: "CronJob", APIVersion: "batch/v1"},
+		{Kind: "DeploymentConfig", APIVersion: "apps.openshift.io/v1"},
+	}
+
+	for _, knownController := range knownWorkloads {
```
chrismuellner, line 195: since this is going to be called a lot, would putting the values into a map and just do a lookup help/be possible here?

chrismuellner (changes requested):

webhook/mutation/pod_mutator.go
```diff
@@ -117,14 +124,81 @@ func registerHealthzEndpoint(mgr manager.Manager) {
...
+	decoder    *admission.Decoder
+	image      string
+	namespace  string
+	apmExists  bool
+	clusterID  string
+	recorder   record.EventRecorder
+}
+
+func rootOwnerPod(ctx context.Context, cnt client.Client, pod *corev1.Pod, namespace string) (string, string, error) {
+	obj := &v1.PartialObjectMetadata{
```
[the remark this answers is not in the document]
aorcholski: Renamed

webhook/mutation/pod_mutator.go
aorcholski: Right!

webhook/mutation/pod_mutator.go
aorcholski: Renamed

aorcholski: There is no reason, err can be reused.

webhook/mutation/pod_mutator.go
aorcholski: It seems that a map has lower efficiency than the list considering such a small number of items to be searched.

config/common/webhook/role-webhook.yaml
```diff
@@ -50,11 +50,3 @@ rules:
...
       - update
       - create
-  - apiGroups:
-      - apps
-    resources:
-      - daemonsets
-    verbs:
-      - get
-      - list
-      - watch
```
chrismuellner, line 60: this rule is missing and leading to errors in the webhook

chrismuellner (changes requested):

config/common/webhook/role-webhook.yaml
mjgrzybek: also `{Kind: "DeploymentConfig", APIVersion: "apps.openshift.io/v1"}` is missing

config/common/webhook/role-webhook.yaml
```diff
@@ -55,6 +55,5 @@ rules:
     resources:
       - daemonsets
     verbs:
-      - get
-      - list
-      - watch
\ No newline at end of file
+      - watch
```
chrismuellner, line 58:
```suggestion
      - list
      - watch
```

chrismuellner (changes requested):

config/common/webhook/clusterrole-webhook.yaml
```diff
@@ -40,4 +40,19 @@ rules:
       - list
       - watch
       - update
-
+  - apiGroups:
+      - ""
+      - "apps"
+      - "batch"
+      - "apps.openshift.io"
```
chrismuellner, line 47: please add the openshift specific rules via patch just when installing for openshift: `apps.openshift.io` and `deploymentconfigs`

config/common/webhook/role-webhook.yaml
aorcholski: Added

config/common/webhook/role-webhook.yaml
aorcholski: Added

chrismuellner (approved):

meik99 (approved):

mjgrzybek (approved):

PR #294, opened 2021-10-08
Ignore base kubernetes/openshift namespaces by default, tolerations for csi via feature-flag

0sewa0:
Labeling system namespaces such as kube-system or openshift-monitoring could cause problems in the cluster.
This is only an issue if the customer doesn't label their cluster accordingly.
(Example csi driver is not on master nodes, openshift-monitoring has pods on master node and if one of these pods restarts then the webhook will try to inject, but because that node lacks a csi driver it won't be able to start up)

~This also introduces a feature flag that will let the customer monitor apps on master nodes, however if csi is used csi pods will also get scheduled for the master nodes.~

Introduces 2 feature-flags:
- `alpha.operator.dynatrace.com/feature-ignored-namespaces`: Set which namespaces should be ignored, defaults to `"[ \"^dynatrace$\", \"^kube-.*\", \"openshift(-.*)?\" ]"`
- ~`alpha.operator.dynatrace.com/feature-csi-master-nodes`: Enable to schedule csi drivers on master nodes~
- `alpha.operator.dynatrace.com/feature-csi-tolerations`: Sets the tolerations on the csi daemonset to the given value

### DIFF
--- a/api/v1beta1/feature_flags.go
+++ b/api/v1beta1/feature_flags.go
@@ -103,7 +103,7 @@ func (dk *DynaKube) FeatureCSITolerations() []corev1.Toleration {
 	return *tolerations
 }
 
-// FeatureIgnoredNamespaces is a feature flag for ignoreing certain namespaces.
+// FeatureIgnoredNamespaces is a feature flag for ignoring certain namespaces.
 // defaults to "[ \"^dynatrace$\", \"^kube-.*\", \"openshift(-.*)?\" ]"
 func (dk *DynaKube) FeatureIgnoredNamespaces() []string {
 	raw, ok := dk.Annotations[annotationFeatureIgnoredNamespaces]


mapper/namespaces.go
```diff
@@ -26,7 +26,7 @@ func NewNamespaceMapper(ctx context.Context, clt client.Client, apiReader client
 
 // MapFromNamespace adds the labels to the targetNs if there is a matching Dynakube
 func (nm NamespaceMapper) MapFromNamespace() (bool, error) {
-	if nm.operatorNs == nm.targetNs.Name {
+	if nm.operatorNs == nm.targetNs.Name || isIgnoredNamespace(nm.targetNs.Name) {
```
meik99, line 29: nice abstraction

chrismuellner (dismissed):

mapper/mapper.go
```diff
@@ -17,6 +18,11 @@ const (
 	UpdatedViaDynakubeAnnotation = "dynatrace.com/updated-via-operator"
 )
 
+var ignoredNamespaces = []string{
+	"^kube-.*",
+	"^openshift(-.*)?",
+}
```
baichinger, line 24: This seems to be very static...

mapper/mapper.go
```diff
@@ -78,11 +84,14 @@ func match(dk *dynatracev1beta1.DynaKube, namespace *corev1.Namespace) (bool, er
 // updateNamespace tries to match the namespace to every dynakube with codeModules
 // finds conflicting dynakubes(2 dynakube with codeModules on the same namespace)
 // adds/updates/removes labels from the namespace.
-func updateNamespace(namespace *corev1.Namespace, dkList *dynatracev1beta1.DynaKubeList, log logr.Logger) (bool, error) {
+func updateNamespace(operatorNs string, namespace *corev1.Namespace, dkList *dynatracev1beta1.DynaKubeList, log logr.Logger) (bool, error) {
 	var updated bool
 	conflict := ConflictChecker{}
 	for i := range dkList.Items {
 		dynakube := &dkList.Items[i]
+		if operatorNs == namespace.Name || (!dynakube.FeatureInjectSystemNamespaces() && isIgnoredNamespace(namespace.Name)) {
```
baichinger, line 92: ... therefore suggest to move `ignoredNamespaces` into the annotation with the following default value:
```
"[ \"^dynatrace$\", \"^kube-.*\", \"openshift(-.*)?\" ]"
```

api/v1beta1/feature_flags.go
```diff
@@ -70,3 +79,23 @@ func (dk *DynaKube) GetFeatureEnableWebhookReinvocationPolicy() string {
 func (dk *DynaKube) FeatureIgnoreUnknownState() bool {
 	return dk.Annotations[annotationFeatureIgnoreUnknownState] == "true"
 }
+
+// FeatureCSIOnMasterNodes is a feature flag that will cause the csi driver(if enabled) to be scheduled on master nodes.
+func (dk *DynaKube) FeatureCSIOnMasterNodes() bool {
+	return dk.Annotations[annotationFeatureCSIOnMasterNodes] == "true"
+}
+
+// FeatureIgnoredNamespaces is a feature flag for ignoreing certain namespaces.
```
chrismuellner, line 88:
```suggestion
// FeatureIgnoredNamespaces is a feature flag for ignoring certain namespaces.
```

api/v1beta1/feature_flags.go
```diff
@@ -70,3 +79,23 @@ func (dk *DynaKube) GetFeatureEnableWebhookReinvocationPolicy() string {
...
+// defaults to "[ \"^dynatrace$\", \"^kube-.*\", \"openshift(-.*)?\" ]"
+func (dk *DynaKube) FeatureIgnoredNamespaces() []string {
+	raw, ok := dk.Annotations[annotationFeatureIgnoredNamespaces]
+	if !ok || raw == "" {
+		return defaultIgnoredNamespaces
```
chrismuellner, line 93: Would a log message informing the user of invalid json be useful here?

mapper/mapper.go
baichinger: _Run on master nodes_ actually means proper support for tolerations. An annotation which unmarshals into a toleration... :exploding_head:

PR #295, opened 2021-10-08
Reimplements the nodeSelector validation check

0sewa0:


webhook/validation/validation.go
```diff
@@ -138,19 +146,32 @@ func hasInvalidActiveGateCapabilities(dynakube *dynatracev1beta1.DynaKube) strin
...
-//	classicNodeSelectorLabels := labels.Set(classicNodeSelectorMap)
-//
-//	return infraNodeSelector.Matches(classicNodeSelectorLabels) || classicNodeSelector.Matches(infraNodeSelectorLabels)
-//}
+func hasConflictingNodeSelector(client client.Client, dynakube *dynatracev1beta1.DynaKube) string {
+	if !dynakube.NeedsOneAgent() {
+		return ""
+	}
+	validDynakubes := &dynatracev1beta1.DynaKubeList{}
+	_ = client.List(context.TODO(), validDynakubes)
```
chrismuellner, line 154: Should we ignore the error here?

webhook/validation/validation.go
```diff
@@ -138,19 +146,32 @@ func hasInvalidActiveGateCapabilities(dynakube *dynatracev1beta1.DynaKube) strin
...
+		nodeSelectorMap := dynakube.NodeSelector()
+		validNodeSelectorMap := validDynakubes.Items[i].NodeSelector()
+		if hasConflictingMatchLabels(nodeSelectorMap, validNodeSelectorMap) {
+			return fmt.Sprintf(errorNodeSelectorConflict, validDynakubes.Items[i].Name)
+		}
+	}
+	return ""
+}
+
+func hasConflictingMatchLabels(labelMap1, labelMap2 map[string]string) bool {
```
chrismuellner, line 165: I would prefer names like `labelMap` and `otherLabelMap`, since the first map is referring to the current Dynakube

webhook/validation/validation.go
```diff
@@ -138,19 +146,32 @@ func hasInvalidActiveGateCapabilities(dynakube *dynatracev1beta1.DynaKube) strin
...
-//	return infraNodeSelector.Matches(classicNodeSelectorLabels) || classicNodeSelector.Matches(infraNodeSelectorLabels)
-//}
+func hasConflictingNodeSelector(client client.Client, dynakube *dynatracev1beta1.DynaKube) string {
+	if !dynakube.NeedsOneAgent() {
+		return ""
+	}
+	validDynakubes := &dynatracev1beta1.DynaKubeList{}
+	_ = client.List(context.TODO(), validDynakubes)
+	for i := range validDynakubes.Items {
+		nodeSelectorMap := dynakube.NodeSelector()
```
chrismuellner, line 156: Doing this once before the loop would be enough.
You could return and skip the call to the k8s api if this returns `nil` then.

chrismuellner (changes requested):

PR #298, opened 2021-10-12
Adds 'internal' into instance label.

0sewa0:
Making it clearer to the users that it shouldn't be changed by hand.


mapper/mapper.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	InstanceLabel                = "dynakube.dynatrace.com/instance"
+	InstanceLabel                = "dynakube.internal.dynatrace.com/instance"
```
chrismuellner, line 16: Please reuse the constant from `webhook` package

PR #311, opened 2021-10-19
Moves toleration feature flag to operator deployment

0sewa0:
The toleration feature flag was moved to the deployment of the operator instead of the dynakube, because this feature flag will affect multiple dynakubes if used, so if left on the dynakubes conflicts could occur. (1 dynakube has it the other doesn't, who wins ?)


controllers/csi/config.go
```diff
@@ -31,6 +31,10 @@ const (
 	DaemonSetName                    = "dynatrace-oneagent-csi-driver"
 	DefaultServiceAccountName        = "dynatrace-oneagent-csi-driver"
 	AnnotationCSIResourcesIdentifier = "dynatrace.com/csi-resources"
+
+	// AnnotationCSITolerations is a feature flag to set tolerations for the csi drive daemonset.
```
chrismuellner, line 35: Not a feature flag anymore

PR #327, opened 2021-10-29
Fix missing certificates on upgrade

chrismuellner:
* Remove patch to add conversion webhook to CRD
* Add conversion webhook to CRD when certificates are ready


baichinger:
This approach introduces a race condition between having the conversion in place and the operator loading the resource. Is it assured that this does not render the custom resource in a broken state?

baichinger:
Data loss is inevitable if a v1alpha1 custom resources gets updated/created when conversion is disabled. Upon write, the custom resource gets stored in v1beta1 format and therefore fields get truncated.

Having no conversion in place - even for a deterministic period of time - is not an option!

chrismuellner:
Different approach in #329

PR #334, opened 2021-11-08
Remove dynamic csi-daemonset

0sewa0:
Having the csi DaemonSet deployed dynamically causes more problems then it solves, so we should just let the user configure the manifest to their liking.

~Still TODO (maybe):~
~- Add a check in the validation webhook to see if there is a csi DaemonSet in the cluster when applying a Dynakube that would use it.~
~- Remove annotations on the operator deployment for the csi daemonset~

Also changes the helper functions for the validation webhook tests so they are more future-proof


config/common/csi/daemonset.yaml
```diff
@@ -0,0 +1,172 @@
...
+        - containerPort: 9809
+          name: healthz
+          protocol: TCP
+        resources:
+          limits:
+            cpu: 10m
+            memory: 15M
+          requests:
+            cpu: 10m
+            memory: 15M
```
meik99, line 111: memory should be increased to 18 MB to avoid running out of memory on startup

config/common/csi/daemonset.yaml
```diff
@@ -0,0 +1,172 @@
...
+        image: quay.io/dynatrace/dynatrace-operator:snapshot
+        imagePullPolicy: Always
+        name: liveness-probe
+        resources:
+          limits:
+            cpu: 5m
+            memory: 15M
+          requests:
+            cpu: 5m
+            memory: 15M
```
meik99, line 135: memory should be increased to 18 MB to avoid running out of memory on startup

meik99 (changes requested):

webhook/validation/validation.go
```diff
@@ -151,6 +164,21 @@ func hasInvalidActiveGateCapabilities(dynakube *dynatracev1beta1.DynaKube) strin
...
+	if !dynakube.NeedsCSIDriver() {
+		return false
+	}
+	csiDaemonSet := appsv1.DaemonSet{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: dtcsi.DaemonSetName, Namespace: dynakube.Namespace}, &csiDaemonSet)
+	if k8serrors.IsNotFound(err) {
+		return true
+	} else if err != nil {
+		logger.Info("error occurred while listing dynakubes", "err", err.Error())
+		return false
```
chrismuellner, line 177: unecessary return

webhook/validation/validation.go
```diff
@@ -22,6 +26,10 @@ const (
 )
 
 const (
+	errorCSIRequired = `
+The DynaKube's specification tries to use the CSI driver while no CSI DaemonSet was deployed.
+Make sure that the deployed kubernetes/openshift.yaml contains the CSI DaemonSet.
```
chrismuellner, line 31: maybe: `The Dynakube's specification requires the CSI driver to work. Make sure you deployed the correct manifests.`

config/common/csi/daemonset.yaml
```diff
@@ -0,0 +1,172 @@
...
+        terminationMessagePath: /dev/termination-log
+        terminationMessagePolicy: File
+        volumeMounts:
+        - mountPath: /csi
+          name: plugin-dir
+      dnsPolicy: ClusterFirst
+      restartPolicy: Always
+      schedulerName: default-scheduler
+      securityContext: {}
+      serviceAccount: dynatrace-oneagent-csi-driver
```
chrismuellner, line 145: can be removed

config/common/csi/daemonset.yaml
```diff
@@ -0,0 +1,172 @@
...
+spec:
+  revisionHistoryLimit: 10
+  selector:
+    matchLabels:
+      internal.oneagent.dynatrace.com/app: csi-driver
+      internal.oneagent.dynatrace.com/component: csi-driver
+  template:
+    metadata:
+      annotations:
+        kubectl.kubernetes.io/default-logs-container: driver
```
chrismuellner, line 17: This annotation is [deprecated](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-default-container) in favor of the following one:
```suggestion
        kubectl.kubernetes.io/default-container: driver
```

chrismuellner (changes requested):

config/common/csi/daemonset.yaml
0sewa0: I think some store or whatever complains if you dont give stuff a serviceAccount
don't want to remove this so close to release :sweat_smile:

chrismuellner: in code `serviceAccount` also wasn't set, only `serviceAccountName` was set:
```
Spec: corev1.PodSpec{
	Containers: []corev1.Container{
		prepareDriverContainer(operatorImage, resourcesMap),
	 	prepareRegistrarContainer(operatorImage, resourcesMap),
	 	prepareLivenessProbeContainer(operatorImage, resourcesMap),
	},
	ServiceAccountName: DefaultServiceAccountName,
	Volumes:            prepareVolumes(),
	Tolerations:        tolerations,
},
```

chrismuellner (approved):

PR #335, opened 2021-11-08
fix: bumped registrar and liveness-probe memory limit to 18 MB to avoid running out of memory

meik99:


chrismuellner:
Done with #334

PR #348, opened 2021-11-12
Change image tag for K8s and OCP on 0.3.0 branch

luhi-DT:
This PR changes the operator image on the release branch, to "snapshot-release-0-3" instead of "snapshot", so kustomize (-k) can be used to apply the operator

### DIFF
--- a/config/kubernetes/kustomization.yaml
+++ b/config/kubernetes/kustomization.yaml
@@ -13,4 +13,3 @@ images:
   - name: quay.io/dynatrace/dynatrace-operator:snapshot
     newName: quay.io/dynatrace/dynatrace-operator
     newTag: snapshot-release-0-3
-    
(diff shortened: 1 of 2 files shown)


0sewa0:
I think the webhook's and csi-driver's image should be changed because they are all the same

luhi-DT:
> csi

csi driver and webhook both should use the kubernetes/openshift folder as base, so this cahnge should replace all the images and change it to the correct one

luhi-DT:
> I think the webhook's and csi-driver's image needs to be changed aswell because they are all the same

csi driver and webhook both should use the kubernetes/openshift folder as base, so this cahnge should replace all the images and change it to the correct one

0sewa0:
oh, I'm dumb :confounded:  its a patch :D

0sewa0 (approved):

baichinger:
@luhi-DT The patch does not get applied to the CSI driver as it is in a completely different tree in the hierarchy.

luhi-DT:
@baichinger you are right, i tested only operator and webhook, i thought it will also replace the image in the csi driver, as the csi driver uses the normal K8s Folder as a base - I made a new PR to fix that. Thanks for the information!

PR #357, opened 2021-11-15
Update csi image to not use snapshot on 0.3.0 branch

luhi-DT:
Updates also the csi driver image and to use the correct release image.


config/kubernetes-csi/kustomization.yaml
```diff
@@ -3,3 +3,7 @@ kind: Kustomization
 bases:
   - ../kubernetes
   - ../common/csi
+images:
```
chrismuellner, line 6: wouldn't it be easier to move this patch into `common/csi` and only have it once?

luhi-DT: it would be yes, you are right

chrismuellner (approved):

PR #363, opened 2021-11-17
Change beta warning condition for applicationMonitoring

luhi-DT:
As applicationMonitoring should only be in beta state, if the CSI Driver is used, then the beta warning should also only be shown if the mentioned conditions are met and not for all applicationMonitoring configurations. Additionally this is mentioned in the readme now.


0sewa0:
you should update the samples :)
https://github.com/Dynatrace/dynatrace-operator/blob/master/config/samples/applicationMonitoring.yml#L1

0sewa0 (approved):

PR #364, opened 2021-11-17
Remove data-ingest from docs

mjgrzybek:
`data-ingest` is not supported in 0.3.0 release, so example usages are removed from config samples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,6 @@ spec:
     capabilities:
       - routing
       - kubernetes-monitoring
-      - data-ingest
 
 ```
 
--- a/config/samples/applicationMonitoring.yml
+++ b/config/samples/applicationMonitoring.yml
@@ -82,7 +82,6 @@ spec:
     capabilities:
       - routing
       - kubernetes-monitoring
-      - data-ingest
 
     # Optional: to use a custom ActiveGate Docker image.
     image: ""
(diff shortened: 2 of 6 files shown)


config/crd/default/bases/dynatrace.com_dynakubes.yaml
```diff
@@ -996,8 +996,7 @@ spec:
                   ActiveGateSpec `json:"activeGate,omitempty"`
                 properties:
                   capabilities:
-                    description: Activegate capabilities enabled (routing, kubernetes-monitoring,
-                      data-ingest)
+                    description: Activegate capabilities enabled (routing, kubernetes-monitoring)
```
chrismuellner, line 999: this is auto-generated by running `make manifests`, you'd have to remove it from `activegate_types.go`

config/crd/ocp311/bases/dynatrace.com_dynakubes.yaml
```diff
@@ -988,8 +988,7 @@ spec:
                   ActiveGateSpec `json:"activeGate,omitempty"`
                 properties:
                   capabilities:
-                    description: Activegate capabilities enabled (routing, kubernetes-monitoring,
```
chrismuellner, line 991: also auto-generated

chrismuellner (changes requested):

chrismuellner (approved):

PR #367, opened 2021-11-18
Rename mentions of beta to preview

luhi-DT:
As we do not want to have cloudNativeFullStack and applicationMonitoring with csiDriver mentioned as 'beta' features, we decided to rename it to 'preview'.


config/samples/multipleDynakubes.yml
```diff
@@ -1,4 +1,4 @@
-# THE CLOUD-NATIVE FULLSTACK MODE AND APPLICATION-MONITORING MODE ENABLED IN THIS YAML ARE STILL IN BETA AND SHOULD NOT BE USED IN A PRODUCTION ENVIRONMENT
+# THE CLOUD-NATIVE FULLSTACK MODE AND APPLICATION-MONITORING MODE ENABLED IN THIS YAML ARE STILL IN PREVIEW AND SHOULD NOT BE USED IN A PRODUCTION ENVIRONMENT
```
chrismuellner, line 1: the applicationMonitoring mode here actually does not use the CSI driver

so either we should use the CSI driver, or remove mention of it being marked as preview

luhi-DT: I think it would make sense to enable the CSIDriver here, to show how it can be used in the future. And as cloudNative is also in preview, customer should not use that yaml in production anyway

chrismuellner (approved):

PR #372, opened 2021-11-19
Use smaller base image

gkrenn:
- replaced `ubi-micro` base image with `ubi-minimal`
- copy curl and unzip to minimal
- set version of base image from `latest`to `8.4`


Dockerfile
```diff
@@ -13,7 +17,39 @@ RUN go get -d ./...
...
+     /usr/lib64/libcom_err.* \
+     /usr/lib64/libldap-* \
+     /usr/lib64/liblber-* \
+     /usr/lib64/libbrotlidec* \
+     /usr/lib64/libunistring.* \
+     /usr/lib64/libkrb5support.* \
+     /usr/lib64/libkeyutils.* \
+     /usr/lib64/libsasl2.* \
+     /usr/lib64/libbz2.* \
+     /usr/lib64/libbrotlicommon.* /usr/lib64/
```
baichinger, line 52: How would this look like if curl is replaced by wget?

gkrenn: Would look similar, we would just have a different set of dependencies and in addition have to change the commands in our init.sh

One solution would be a statically linked curl, but haven't found a trustworthy source for such a build.

Therefore, if we don't want to build our own statically linked curl, I would stick with this solution.

gkrenn:
As coping all the libraries of curl could lead to problems that are hard to detect, we decided to wait until we have moved the init.sh logic to golang.

After the movement we no longer depend on curl and unzip.

PR #382, opened 2021-11-24
Validation webhook shows multiple errors

0sewa0:
Currently the validation webhook only shows the first error it encounters which is not that user friendly if there are multiple problems with the Dynakube. (this would get worse if we add more validation checks)

This change makes it that the user will be given the list of validation errors.
Refactors the codebase by separating the validation checks and their test into separate files so we don't have an ever grooving file(+test file).
Adjusts the nodeSelector check so it recognizes when there are 2 dynakubes which deploys OS OneAgents where atleast 1 of them would deploy a oneagent daemonSet for all nodes (no nodeSelector is set)


webhook/validation/namespace_selector.go
```diff
@@ -0,0 +1,25 @@
...
+`
+
+func conflictingNamespaceSelector(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
+	if !dynakube.NeedAppInjection() {
+		return ""
+	}
+	dkMapper := mapper.NewDynakubeMapper(context.TODO(), dv.clt, dv.apiReader, dynakube.Namespace, dynakube, log)
+	_, err := dkMapper.MatchingNamespaces()
+	if err != nil {
+		log.Info("requested dynakube has conflicting namespaceSelector", "name", dynakube.Name, "namespace", dynakube.Namespace)
```
chrismuellner, line 21: Do we have the information which Dynakubes it's conflicting with here?
We could also add a separate error message when there is already a Dynakube without `namespaceSelector` and the new Dynakube is also without `namespaceSelector` (since then the `namespaceSelector` isn't really conflicting, but more the lack of it)

chrismuellner (changes requested):

mapper/mapper.go
```diff
@@ -17,6 +17,8 @@ import (
 const (
 	InstanceLabel                = dtwebhook.LabelInstance
 	UpdatedViaDynakubeAnnotation = "dynatrace.com/updated-via-operator"
+	ConflictErrorMessage         = "namespace matches two or more DynaKubes which is unsupported. " +
```
chrismuellner, line 20: should be named `errorConflictingNamespace` or similar to match the other errors

webhook/validation/oneagent.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}
+	return ""
+}
+
+func hasConflictingMatchLabels(labelMap, otherLabelMap map[string]string) bool {
+	if labelMap == nil || otherLabelMap == nil {
+		return true
+	}
+	if labelMap != nil && otherLabelMap != nil {
```
chrismuellner, line 70: this can't happen

webhook/validation/preview.go
```diff
@@ -0,0 +1,21 @@
+package validation
+
+import (
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/api/v1beta1"
+)
+
+const (
+	warningCloudNativeFullStack  = `cloudNativeFullStack mode is in PREVIEW. Please be aware that it is NOT production ready and you may run into bugs.`
```
chrismuellner, line 8: could be combined into a single variable

chrismuellner (approved):

PR #386, opened 2021-11-24
`pod_mutator` supports multiple injections; `pod_mutator` refactoring

mjgrzybek:
`DataIngest` is another capability (alongside `OneAgent`) that can be injected to pods by webhook.
`pod_mutator` must be extended to support multiple injections.
Using this opportunity, some refactoring was done, too.

### Changes
- `DataIngest` is injected by default if `OneAgent` is injected
- `DataIngest` injection can be explicitly disabled or enabled (so that it can be injected even if `OneAgent` is not) using annotation: `data-ingest.dynatrace.com/inject=true|false`
- Annotation `dynakube.dynatrace.com/injected` now holds a lexical sorted, comma separated, list of injected capabilities intsead of `"true"`
- `pod_mutator.Handle()` implementation simplified
- `init.sh` supports both injections
-  new UT created: `TestPodPartialInjection`
-  existing UT aligned with changes


webhook/mutation/pod_mutator.go
```diff
@@ -204,264 +224,485 @@ func isWellKnownWorkload(ownerRef metav1.OwnerReference) bool {
...
+		}
+	} else if err != nil {
+		log.Error(err, "failed to query the init secret before pod injection")
+		rsp = admission.Errored(http.StatusBadRequest, err)
+		return &rsp
+	}
+	return nil
+}
+
+func (m *podMutator) CreateInjectionInfo(pod *corev1.Pod) *InjectionInfo {
```
0sewa0, line 698: would fit perfectly in `injection_info.go`, it doesn't need to be a method of `podMutator` as it never uses`m`.
Nitpick: maybe calling it `NewInjectionInfoForPod` (or something) would help understandability when its no longer a method of `podMutator`

webhook/mutation/pod_mutator.go
```diff
@@ -204,264 +224,485 @@ func isWellKnownWorkload(ownerRef metav1.OwnerReference) bool {
...
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: dtwebhook.SecretConfigName,
+				},
+			},
+		},
+	)
+}
+
+func (m *podMutator) getBasicData(pod *corev1.Pod) (string, string, string, string, string) {
```
0sewa0, line 495: When you are returning 5 things you should either give the return values names 
`func (m *podMutator) getBasicData(pod *corev1.Pod) (technologies, installPath, installerURL, failurePolicy, image string)`
or create a struct (`OneAgentInstallConfig` or something) that has these fields

0sewa0:
Just a general question:
I see a lot of methods that don't need to be methods because they don't use the `struct` they are the methods of.
Is there a reason for it ? Because it would make testing these 'methods' easier (not having to create a dummy struct) if they were not methods to begin with. :smile: 
Or maybe storing more things in the struct that we have to pass into multiple methods would make these 'not-really-methods' make more sense (I kinda prefer the not-having-them-as-methods way but I'm not 100% decided on it :sweat_smile:)
(I didn't want to litter the PR with pointing out each 'not-really-a-method')

0sewa0:
to get only `data-ingest` injected I have to create a dynakube that has some kind of application injection + data-ingest configured AND add the annotation on the pod to disable app injection ?

mjgrzybek:
> Just a general question: I see a lot of methods that don't need to be methods because they don't use the `struct` they are the methods of. Is there a reason for it ? Because it would make testing these 'methods' easier (not having to create a dummy struct) if they were not methods to begin with. smile Or maybe storing more things in the struct that we have to pass into multiple methods would make these 'not-really-methods' make more sense (I kinda prefer the not-having-them-as-methods way but I'm not 100% decided on it sweat_smile) (I didn't want to litter the PR with pointing out each 'not-really-a-method')

Ah, working on it - forgot to fix IDE suggestions.

mjgrzybek:
> to get only `data-ingest` injected I have to create a dynakube that has some kind of application injection + data-ingest configured AND add the annotation on the pod to disable app injection ?

`dynakube` should have `applicationMonitoring` or `cloudNativeFullStack` enabled. No need to mention `data-ingest` there.
AND
pods should be annotated:
```
annotations:
        oneagent.dynatrace.com/inject: "false"
        data-ingest.dynatrace.com/inject: "true"
```

initgeneration/init.sh.test-sample
```diff
@@ -152,10 +151,22 @@ createConfigurationFiles() {
 	done
 }
 
+createConfigurationFilesDI() {
```
chrismuellner, line 154: please write out `DataIngest`

initgeneration/init.sh.test-sample
```diff
@@ -125,7 +125,7 @@ dt.kubernetes.cluster.id=${cluster_id}" > "${1}/${2}.properties"
 }" > "${1}/${2}.json"
 }
 
-createConfigurationFiles() {
+createConfigurationFilesOA() {
```
chrismuellner, line 128: please write out `OneAgent`

webhook/mutation/helper.go
```diff
@@ -0,0 +1,75 @@
+package mutation
```
chrismuellner, line 1: we want to prevent `helper` files. please think of another name that fits

webhook/mutation/pod_mutator.go
```diff
@@ -204,264 +224,474 @@ func isWellKnownWorkload(ownerRef metav1.OwnerReference) bool {
...
+	technologies, installPath, installerURL, failurePolicy, image := m.getBasicData(pod)
 
-				if !preloaded {
-					// container does not have LD_PRELOAD set
-					log.Info("instrumenting missing container", "name", c.Name)
+	dkVol, mode := ensureDkVolume(dk)
 
-					deploymentMetadata := deploymentmetadata.NewDeploymentMetadata(m.clusterID, deploymentmetadata.DeploymentTypeApplicationMonitoring)
-					updateContainer(c, &dk, pod, deploymentMetadata, dataIngestFields)
+	ensureInjectionConfigVolume(pod)
```
chrismuellner, line 285: `ensure` sounds like the function verifies the volumes exist. would `setup` work as well?

webhook/mutation/pod_mutator.go
```diff
@@ -204,264 +224,474 @@ func isWellKnownWorkload(ownerRef metav1.OwnerReference) bool {
...
+			updateContainerDI(c, pod, deploymentMetadata, dataIngestFields)
+		}
 	}
-	pod.Annotations[dtwebhook.AnnotationInjected] = "true"
+}
 
-	workloadName, workloadKind, err := findRootOwnerOfPod(ctx, m.metaClient, pod, req.Namespace)
-	if err != nil {
-		return admission.Errored(http.StatusInternalServerError, err)
+func decorateInstallContainerWithDI(ic *corev1.Container, injectionInfo *InjectionInfo, workloadKind string, workloadName string) {
```
chrismuellner, line 341: write out `DataIngest`

webhook/mutation/pod_mutator.go
```diff
@@ -204,264 +224,474 @@ func isWellKnownWorkload(ownerRef metav1.OwnerReference) bool {
...
-	if basePodName == "" {
-		basePodName = pod.Name
+func (m *podMutator) ensureReinvocationPolicy(pod *corev1.Pod, dk dynatracev1beta1.DynaKube, injectionInfo *InjectionInfo, dataIngestFields map[string]string, req admission.Request) admission.Response {
+	var needsUpdate = false
+	var installContainer *corev1.Container
+	for i := range pod.Spec.Containers {
+		c := &pod.Spec.Containers[i]
+
+		oaInjected := false
+		if injectionInfo.enabled(OneAgent) {
```
chrismuellner, line 540: could this check (and the di check) be handled via annotation?
if not, please move them to a function, that returns whether injection is missing

webhook/mutation/pod_mutator.go
```diff
@@ -204,264 +224,474 @@ func isWellKnownWorkload(ownerRef metav1.OwnerReference) bool {
...
+					}
+				}
+			}
+			updateInstallContainerOA(installContainer, i+1, c.Name, c.Image)
+
+			needsUpdate = true
+		}
+
+		if diInjectionMissing {
+			log.Info("instrumenting missing container", "name", c.Name)
```
chrismuellner, line 585: should have different logs for oa/di

webhook/mutation/pod_mutator.go
```diff
@@ -204,264 +224,474 @@ func isWellKnownWorkload(ownerRef metav1.OwnerReference) bool {
...
+	if err != nil {
+		log.Error(err, "failed to query the data-ingest endpoint secret before pod injection")
+		rsp = admission.Errored(http.StatusBadRequest, err)
+		return nil, &rsp
+	}
+	return dataIngestFields, nil
+}
 
-		updateContainer(c, &dk, pod, deploymentMetadata, dataIngestFields)
+func (m *podMutator) getDk(ctx context.Context, req admission.Request, dkName string) (dynatracev1beta1.DynaKube, *admission.Response) {
```
chrismuellner, line 659: `getDynakube`

chrismuellner (changes requested):

webhook/mutation/helper.go
0sewa0: could be `sorting.go` or `sort.go`

webhook/mutation/pod_mutator.go
mjgrzybek: Info if it's enabled is actually based on pods annotations, see: [pod_mutator.go:67](https://github.com/Dynatrace/dynatrace-operator/pull/386/files#diff-e0bb41fcf412fe6b324780881c3a276f59c7e7530df4705e96e33876ff7e0912R66)

chrismuellner: the enabled check is fine, but the injected check is a bit arbitrary.
any of the other volumes/env variables required for the injection could be missing

mjgrzybek: This logic is [untouched](https://github.com/Dynatrace/dynatrace-operator/blob/3402f37059fd97e79dc4dc598f777706366f4a28/webhook/mutation/pod_mutator.go#L294). It should be changed in separate PR imho.
Agree check is redundant but code looks cleaner and is ready for further refactoring - to hide `injectable` logic behind interfaces.

chrismuellner (approved):

0sewa0 (approved):

PR #388, opened 2021-11-25
Refactors logging in the csi packages

0sewa0:
Moves logger to the `config.go` in each package, and have the logger on a per package level.
Also log messages should always start with lowercase.

### DIFF
--- a/controllers/csi/driver/server.go
+++ b/controllers/csi/driver/server.go
@@ -317,11 +317,11 @@ func logGRPC() grpc.UnaryServerInterceptor {
 		}
 		methodName := ""
 		if info.FullMethod == "/csi.v1.Node/NodePublishVolume" {
-			req := req.(csi.NodePublishVolumeRequest)
+			req := req.(*csi.NodePublishVolumeRequest)
 			methodName = "NodePublishVolume"
 			log.Info("GRPC call", "method", methodName, "volume-id", req.VolumeId)
 		} else if info.FullMethod == "/csi.v1.Node/NodeUnpublishVolume" {
-			req := req.(csi.NodeUnpublishVolumeRequest)
+			req := req.(*csi.NodeUnpublishVolumeRequest)
 			methodName = "NodeUnpublishVolume"
 			log.Info("GRPC call", "method", methodName, "volume-id", req.VolumeId)
 		}
--- a/controllers/csi/driver/server.go
+++ b/controllers/csi/driver/server.go
@@ -312,17 +312,24 @@ func (svr *CSIDriverServer) loadVolumeInfo(volumeID string) (*metadata.Volume, e
 
 func logGRPC() grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-		if info.FullMethod == "/csi.v1.Identity/Probe" {
+		if info.FullMethod == "/csi.v1.Identity/Probe" || info.FullMethod == "/csi.v1.Node/NodeGetCapabilities" {
 			return handler(ctx, req)
 		}
-
-		log.Info("GRPC call", "method", info.FullMethod, "request", req)
-
+		methodName := ""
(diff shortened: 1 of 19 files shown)


controllers/csi/driver/server.go
```diff
@@ -335,11 +306,11 @@ func (svr *CSIDriverServer) loadVolumeInfo(volumeID string) (*metadata.Volume, e
 	if volume == nil {
 		return &metadata.Volume{}, nil
 	}
-	svr.log.Info("loaded volume info", "id", volume.VolumeID, "pod name", volume.PodName, "version", volume.Version, "dynakube", volume.TenantUUID)
+	log.Info("loaded volume info", "id", volume.VolumeID, "pod name", volume.PodName, "version", volume.Version, "dynakube", volume.TenantUUID)
 	return volume, nil
 }
 
-func logGRPC(log logr.Logger) grpc.UnaryServerInterceptor {
```
chrismuellner, line 342: could we rework when this is logged as well?
I don't think we have to log every successful grpc call

controllers/csi/provisioner/reconciler.go
```diff
@@ -86,8 +78,9 @@ func (r *OneAgentProvisioner) SetupWithManager(mgr ctrl.Manager) error {
 var _ reconcile.Reconciler = &OneAgentProvisioner{}
 
 func (r *OneAgentProvisioner) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
-	rlog := log.WithValues("namespace", request.Namespace, "name", request.Name)
-	rlog.Info("Reconciling DynaKube")
+	log = log.WithValues("namespace", request.Namespace, "dynakube", request.Name)
+	defer resetLogger()
```
chrismuellner, line 82: we could explicitly log these values instead of having them on the package wide logger?

chrismuellner (changes requested):

controllers/csi/driver/server.go
```diff
@@ -335,23 +306,30 @@ func (svr *CSIDriverServer) loadVolumeInfo(volumeID string) (*metadata.Volume, e
...
 			return handler(ctx, req)
 		}
-
-		log.Info("GRPC call", "method", info.FullMethod, "request", req)
-
+		methodName := ""
+		if info.FullMethod == "/csi.v1.Node/NodePublishVolume" {
+			req := req.(csi.NodePublishVolumeRequest)
+			methodName = "NodePublishVolume"
+			log.Info("GRPC call", "method", info.FullMethod, "volume-id", req.VolumeId)
```
chrismuellner, line 322: wouldn't it be enough to log the `volume-id` and if it's node publish/unpublish here

controllers/csi/driver/server.go
```diff
@@ -335,23 +306,30 @@ func (svr *CSIDriverServer) loadVolumeInfo(volumeID string) (*metadata.Volume, e
...
+			methodName = "NodeUnpublishVolume"
+			log.Info("GRPC call", "method", info.FullMethod, "volume-id", req.VolumeId)
+		}
 		resp, err := handler(ctx, req)
 		if err != nil {
-			log.Error(err, "GRPC call failed")
+			log.Error(err, fmt.Sprintf("%s GRPC call failed", methodName))
 		} else {
-			log.Info("GRPC call successful", "response", resp)
+			log.Info(fmt.Sprintf("%s GRPC call successful", methodName))
```
chrismuellner, line 332: is this log useful?
would removing the info log before the grpc call and just log error/success reduce log spam?

chrismuellner (changes requested):

controllers/csi/driver/server.go
0sewa0: I like `GRPC call`, it shows that the kubernetes made the request
it separates it from other logs

controllers/csi/driver/server.go
0sewa0: there shouldn't be much spam anymore, but yeah I can't come up with a valid use for this :D

chrismuellner (approved):

PR #396, opened 2021-11-30
Refactors logging in most packages

0sewa0:
The webhook package was left out because its has been refactored in another PR.
The kubeobjects package is a helper package so passing in loggers there is fine.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -29,7 +29,7 @@ import (
 )
 
 var (
-	log = logger.NewDTLogger()
+	log = logger.NewDTLogger().WithName("main")
 )
 
 const (
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -45,7 +45,7 @@ func main() {
 	pflag.CommandLine.AddFlagSet(csiDriverFlags())
 	pflag.Parse()
 
-	ctrl.SetLogger(logger.NewDTLogger())
+	ctrl.SetLogger(log)
 
 	version.LogVersion()
 
@@ -79,7 +79,7 @@ func main() {
 		exitOnError(err, "webhook-server setup failed")
 		defer cleanUp()
 	default:
-		log.Error(errBadSubcmd, "Unknown subcommand", "command", subCmd)
+		log.Error(errBadSubcmd, "unknown subcommand", "command", subCmd)
 		os.Exit(1)
 	}
 
--- a/controllers/certificates/certs.go
+++ b/controllers/certificates/certs.go
@@ -67,28 +67,28 @@ func (cs *Certs) ValidateCerts() error {
 
 func (cs *Certs) validateRootCerts(now time.Time) bool {
 	if cs.Data[RootKey] == nil || cs.Data[RootCert] == nil {
-		log.Info("No root certificates found, creating")
+		log.Info("no root certificates found, creating")
 		return true
 	}
 
 	var err error
 
 	if block, _ := pem.Decode(cs.Data[RootCert]); block == nil {
-		log.Info("Failed to parse root certificates, renewing", "error", "can't decode PEM file")
+		log.Info("failed to parse root certificates, renewing", "error", "can't decode PEM file")
 		return true
 	} else if cs.rootPublicCert, err = x509.ParseCertificate(block.Bytes); err != nil {
-		log.Info("Failed to parse root certificates, renewing", "error", err)
+		log.Info("failed to parse root certificates, renewing", "error", err)
 		return true
 	} else if now.After(cs.rootPublicCert.NotAfter.Add(-renewalThreshold)) {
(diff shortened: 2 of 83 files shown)


cmd/operator/main.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 var (
-	log = logger.NewDTLogger()
+	log = logger.NewDTLogger().WithName("main")
```
chrismuellner, line 32: in `ctrl.SetLogger(...)` a new logger without name is used. Could this one be reused there?

controllers/certificates/certs.go
```diff
@@ -70,28 +67,28 @@ func (cs *Certs) ValidateCerts() error {
 
 func (cs *Certs) validateRootCerts(now time.Time) bool {
 	if cs.Data[RootKey] == nil || cs.Data[RootCert] == nil {
-		cs.Log.Info("No root certificates found, creating")
+		log.Info("No root certificates found, creating")
```
chrismuellner, line 70: lower case (rest of file as well)

controllers/config.go
```diff
@@ -0,0 +1,9 @@
+package controllers
+
+import (
+	"github.com/Dynatrace/dynatrace-operator/logger"
+)
+
+var (
+	log = logger.NewDTLogger().WithName("controllers")
```
chrismuellner, line 8: does a logger at this level make sense?
wouldn't it be better to move `dynakube_state.go` into `dynakube/status` so there are no more files on this level?

controllers/customproperties/reconciler.go
```diff
@@ -47,14 +44,14 @@ func (r *Reconciler) Reconcile() error {
 	if r.hasCustomPropertiesValueOnly() {
 		mustNotUpdate, err := r.createCustomPropertiesIfNotExists()
 		if err != nil {
-			r.log.Error(err, fmt.Sprintf("could not create custom properties for '%s'", r.customPropertiesOwnerName))
+			log.Error(err, fmt.Sprintf("could not create custom properties for '%s'", r.customPropertiesOwnerName))
```
chrismuellner, line 47:
```suggestion
			log.Error(err, "could not create custom properties", "ownerName", r.customPropertiesOwnerName))
```

controllers/customproperties/reconciler.go
```diff
@@ -47,14 +44,14 @@ func (r *Reconciler) Reconcile() error {
...
 			return errors.WithStack(err)
 		}
 
 		if !mustNotUpdate {
 			err = r.updateCustomPropertiesIfOutdated()
 			if err != nil {
-				r.log.Error(err, fmt.Sprintf("could not update custom properties for '%s'", r.customPropertiesOwnerName))
+				log.Error(err, fmt.Sprintf("could not update custom properties for '%s'", r.customPropertiesOwnerName))
```
chrismuellner, line 54: I don't think we should use `fmt.Sprintf` in logs

controllers/dynakube/updates/version.go
```diff
@@ -55,13 +55,13 @@ func ReconcileVersions(
 
 	if needsActiveGateUpdate {
 		if err := updateImageVersion(dkState, dk.ActiveGateImage(), &dk.Status.ActiveGate.VersionStatus, &dockerCfg, verProvider, true); err != nil {
-			dkState.Log.Error(err, "Failed to update ActiveGate image version")
+			log.Error(err, "Failed to update ActiveGate image version")
```
chrismuellner, line 58: lower case

controllers/istio/config.go
```diff
@@ -0,0 +1,9 @@
+package istio
+
+import (
+	"github.com/Dynatrace/dynatrace-operator/logger"
+)
+
+var (
+	log = logger.NewDTLogger().WithName("dynakube-istio")
```
chrismuellner, line 8: isn't `istio` enough as name? or `istio-controller` to fit the other controllers

controllers/oneagent/config.go
```diff
@@ -0,0 +1,9 @@
+package oneagent
+
+import (
+	"github.com/Dynatrace/dynatrace-operator/logger"
+)
+
+var (
+	log = logger.NewDTLogger().WithName("dynakube-oneagent")
```
chrismuellner, line 8: `oneagent-controller`?

chrismuellner (changes requested):

controllers/istio/config.go
0sewa0: its not a controller its just named that :sweat_smile: 
the `dynakube` part is there because its called from the `dynakube-controller`

controllers/oneagent/config.go
0sewa0: same as before, its not a controller, and its called from the `dynakube-controller`

controllers/istio/config.go
chrismuellner: makes sense, not really redundant then

chrismuellner (approved):

PR #397, opened 2021-11-30
Clarify autoUpdate comment

chrismuellner:
Previous wording was confusing, so this PR clarifies that the field is used to enable autoUpdate (which is enabled by default)


api/v1beta1/oneagent_types.go
```diff
@@ -98,7 +98,8 @@ type HostInjectSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OneAgent environment variable installer arguments",order=22,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:hidden"}
 	Env []corev1.EnvVar `json:"env,omitempty"`
 
-	// Disable automatic restarts of OneAgent pods in case a new version is available
+	// Optional: Enables automatic restarts of OneAgent pods in case a new version is available
```
0sewa0, line 101:
```suggestion
	// Optional: Enables/Disables automatic restarts of OneAgent pods in case a new version is available
```

0sewa0 (approved):

PR #402, opened 2021-12-03
Reorg codebase

0sewa0:
same as https://github.com/Dynatrace/dynatrace-operator/pull/391


config/manifests/bases/dynatrace-operator.clusterserviceversion.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
     capabilities: Deep Insights
     categories: Monitoring,Logging & Tracing
     containerImage: registry.connect.redhat.com/dynatrace/dynatrace-operator
-    repository: https://github.com/Dynatrace/dynatrace-operator
+    repository: https://github.com/Dynatrace/dynatrace-operator/src
```
chrismuellner, line 9: I don't think this is correct, the repository url didn't change

deploy/install.sh
```diff
@@ -102,12 +102,12 @@ isOCP311() {
 
 applyDynatraceOperator() {
   if [ "${CLI}" = "kubectl" ]; then
-    "${CLI}" apply -f https://github.com/Dynatrace/dynatrace-operator/releases/latest/download/kubernetes.yaml
+    "${CLI}" apply -f https://github.com/Dynatrace/dynatrace-operator/src/releases/latest/download/kubernetes.yaml
```
chrismuellner, line 105: this is the incorrect location

PROJECT
```diff
@@ -4,14 +4,14 @@ layout:
 plugins:
   go.sdk.operatorframework.io/v3: {}
 projectName: dynatrace-operator
-repo: github.com/Dynatrace/dynatrace-operator
+repo: github.com/Dynatrace/dynatrace-operator/src
```
chrismuellner, line 7: repo stays the same

README.md
```diff
@@ -18,7 +18,7 @@ With v0.3.0 the CRD structure has been changed to provide an easier and more und
 - added **hostMonitoring** for only monitoring the host in the cluster without app-only injection
 - (PREVIEW) added **cloudNativeFullStack** mode, which combines **hostMonitoring**, with the webhook based **applicationMonitoring**
 
-For more information please have a look at [our DynaKube Custom Resource examples](https://github.com/Dynatrace/dynatrace-operator/tree/master/config/samples),
+For more information please have a look at [our DynaKube Custom Resource examples](https://github.com/Dynatrace/dynatrace-operator/src/tree/master/config/samples),
```
chrismuellner, line 21: incorrect url

chrismuellner (changes requested):

chrismuellner (commented):
What about the `version` package, it should also be under `src`?

Moving everything that has `go` code under `src` would also make sense (so testing stuff as well)...

0sewa0:
the `version` package is something thats only purpose is to store the version of the operator, we set it during the build process, and we only read it in the go code, so I felt that its fine to be left out

chrismuellner (approved):

PR #403, opened 2021-12-03
Adds editorconfig + pre-commit

0sewa0:
We already have the pre-commit hook in our internal wiki, but having it in the actual repo is nicer.
(I tried to improve the pre-commit, dunno if I succeeded)


.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.go]
+indent_style = space
```
gkrenn, line 12: Would change that to tab as we are currently using tab nearly in every gofile.

.github/pre-commit
```diff
@@ -0,0 +1,55 @@
...
+        ## add any potential changes from our formatting to the
+        ## commit
+        git add $file
+    done
+fi
+
+golangci-lint run --build-tags integration,containers_image_storage_stub
+res=$?
+if [[ $res -ne 0  ]] ; then
+echo "Linting failed"
```
gkrenn, line 34: Nitpick, I don't care if you don't do it, but there should be indentations.

luhi-DT (approved):

gkrenn (approved):

PR #404, opened 2021-12-03
Add support for single token

gkrenn:
If the apiToken has the scope "InstallerDownload" and "DataExport", the paas token can be omitted


controllers/dynakube/dtclient_reconciler.go
```diff
@@ -32,9 +31,8 @@ type tokenConfig struct {
 }
 
 func (r *DynatraceClientReconciler) Reconcile(ctx context.Context, instance *dynatracev1beta1.DynaKube) (dtclient.Client, bool, error) {
-	now := r.Now
```
0sewa0, line 35: wouldn't calling r.Now multiple times and different points in the function cause to have multiple different timestamps, couldn't that cause problems ?

gkrenn: Do not see a problem there. - r.Now can't be called as it isn't a function. - And accessing the same field multiple times is not a problem.

src/controllers/dynakube/dtclient_reconciler.go
```diff
@@ -202,9 +210,28 @@ func (r *DynatraceClientReconciler) Reconcile(ctx context.Context, instance *dyn
...
+	if updateSecret {
+		if err := r.Client.Update(ctx, secret); err != nil {
+			return dtc, updateCR, fmt.Errorf("failed to update secret: %s", err)
+		}
+	}
+
 	return dtc, updateCR, nil
 }
 
+func (r *DynatraceClientReconciler) getTokensFromSecret(ctx context.Context, instance *dynatracev1beta1.DynaKube) (*corev1.Secret, error) {
```
chrismuellner, line 222: this doesn't only get the tokens from the secret
maybe splitting this into loading the secret and then assigning the tokens in a separate stuff would improve readability here

src/controllers/dynakube/dtclient_reconciler_test.go
```diff
@@ -36,8 +36,6 @@ func TestReconcileDynatraceClient_TokenValidation(t *testing.T) {
 		rec := &DynatraceClientReconciler{
 			Client:              c,
 			DynatraceClientFunc: StaticDynatraceClient(dtcMock),
-			UpdatePaaSToken:     true,
```
chrismuellner, line 39: this could use a test to make sure the `apiToken` is used when the `paasToken` is missing

src/controllers/dynakube/dynakube_controller_test.go
```diff
@@ -187,15 +93,48 @@ func TestReconcileActiveGate_Reconcile(t *testing.T) {
...
+		assert.NoError(t, err)
+		assert.NotNil(t, secret)
+		assert.Equal(t, string(secret.Data[dtclient.DynatracePaasToken]), testAPIToken)
+		assert.Equal(t, string(secret.Data[dtclient.DynatraceApiToken]), testAPIToken)
+	})
+}
+
 func TestReconcile_RemoveRoutingIfDisabled(t *testing.T) {
-	mockClient := &dtclient.MockDynatraceClient{}
+	// replacing is possible
```
chrismuellner, line 135: replacing what?

src/initgeneration/initgeneration.go
```diff
@@ -133,9 +133,15 @@ func (g *InitGenerator) generate(ctx context.Context, dk *dynatracev1beta1.DynaK
...
 	var tokens corev1.Secret
+	var paasToken []byte
 	if err := g.client.Get(context.TODO(), client.ObjectKey{Name: dk.Tokens(), Namespace: g.namespace}, &tokens); err != nil {
 		return nil, errors.WithMessage(err, "failed to query tokens")
 	}
+	if len(tokens.Data[dtclient.DynatracePaasToken]) != 0 {
+		paasToken = tokens.Data[dtclient.DynatracePaasToken]
+	} else {
+		paasToken = tokens.Data[dtclient.DynatraceApiToken]
+	}
```
chrismuellner, line 144: could be moved to a function

src/controllers/dynakube/dtclient_reconciler.go
```diff
@@ -96,21 +88,36 @@ func (r *DynatraceClientReconciler) Reconcile(ctx context.Context, instance *dyn
...
+	if r.ApiToken == "" {
+		updateCR = setCondition(&sts.Conditions, metav1.Condition{
+			Type:    apiTokenConf.Type,
+			Status:  metav1.ConditionFalse,
+			Reason:  dynatracev1beta1.ReasonTokenMissing,
+			Message: fmt.Sprintf("Token %s on secret %s missing", apiTokenConf.Key, secretKey),
+		}) || updateCR
+		valid = false
+	} else if r.PaasToken == "" {
+		// if paas token is missing api token is paas token
```
chrismuellner, line 102: should it really be stored twice then?
could we add a flag on the dynakube status to indicate that api/paas token are the same or would that just complicate tings?

deploy/install.sh
```diff
@@ -9,6 +9,7 @@ CONNECTION_NAME=""
 CLUSTER_NAME=""
 CLUSTER_NAME_REGEX="^[-_a-zA-Z0-9][-_\.a-zA-Z0-9]*$"
 CLUSTER_NAME_LENGTH=256
+USES_API_AS_PAAS_TOKEN=false
```
chrismuellner, line 12: I don't think the `install.sh` needs to be updated, but having it can't hurt...

chrismuellner (changes requested):

src/controllers/dynakube/dynakube_controller_test.go
gkrenn: Was just a hint for me during development, will be removed

src/controllers/dynakube/dtclient_reconciler.go
gkrenn: Guess it would not even work, as we need the paasToken field in our secret.

E.g. [env_vars.go](https://github.com/Dynatrace/dynatrace-operator/blob/3dd79f8fbfca65cbdf0414503fe7f03c72c40068/src/controllers/dynakube/oneagent/daemonset/env_vars.go#L43) - The installer env var is grabbed from the secret. FYI: This is only used if the oneagent is not downloaded from the tenant.

So there has to be a paasToken field in the secret.

chrismuellner: ok, shouldn't be done for this pr then

0sewa0 (approved):

chrismuellner (approved):

PR #408, opened 2021-12-08
AG service: port redirection 80->9998 added

mjgrzybek:
If service mesh is used, http rest endpoint is preferred over https.


aorcholski (approved):
Port mapping is OK

PR #410, opened 2021-12-13
Fix install.sh for kubectl 1.23 or later

chrkl:
The install.sh script uses the `--dry-run` mode of kubectl. This flag
without further value has been deprecated with version 1.18 and entirely
removed with 1.23. This change makes the dry-run to run on client side.

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -111,7 +111,7 @@ applyDynatraceOperator() {
     fi
   fi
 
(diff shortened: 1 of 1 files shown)


TigerC10:
Created issue #414 for this before seeing the PR.

TigerC10 (approved):

chrismuellner (approved):

chrismuellner:
Thanks for the fix!

PR #413, opened 2021-12-15
Adding support for applicationMonitoring and cloudNativeFullStack

TigerC10:
Adding support for applicationMonitoring and cloudNativeFullStack with --config-option flag.

Flag name feels a little weird, but it came from this page: https://www.dynatrace.com/support/help/setup-and-configuration/setup-on-container-platforms/kubernetes/set-up-k8s-monitoring#config

Open to suggestions on better names for the flag.

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -5,6 +5,7 @@ set -e
 CLI="kubectl"
 SKIP_CERT_CHECK="false"
 ENABLE_VOLUME_STORAGE="false"
+CONFIG_OPTION="classicFullStack"
 CONNECTION_NAME=""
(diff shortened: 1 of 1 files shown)


TigerC10:
I kind of take issue with that, because the install instructions for ActiveGate (the non-kubernetes version) doesn't include all of the possible options/flags for the installer.  Point is, having extra options/flags doesn't _hurt_.

Also, GKE Autopilot is a thing, where host monitoring is not possible.  Instrumenting "applicationOnly" should 100% be a toggle in the Deploy Dynatrace GUI.  Having this flag in the `install.sh` file enables the toggle to be added later.

PR #421, opened 2021-12-21
Removed modifying webhook configurations if they are not found

meik99:
When the operator is deployed by OLM the webhook configurations cannot be easily queried since they have semi random names.
Since certificates were generated and appended to them, the operator failed when in that case since it could not find the configurations to append them to.
With these changes it does not fail anymore if it can't find the configs, but instead just writes log messages accordingly.

Also some minor adjustments were made to the Makefile, deployments and base-CSV so the operator can more reliably be deployed via OLM


config/common/webhook/deployment-webhook.yaml
```diff
@@ -57,6 +57,8 @@ spec:
               path: /healthz
               port: server-port
               scheme: HTTPS
+            initialDelaySeconds: 60
```
chrismuellner, line 60: this introduces a delay on all platforms. why is this needed now?

src/controllers/certificates/webhook_reconciler.go
```diff
@@ -67,64 +62,58 @@ func (r *ReconcileWebhookCertificates) Reconcile(ctx context.Context, request re
 	r.namespace = request.Namespace
 	r.ctx = ctx
 
-	secret, err := r.getSecret()
+	mutatingWebhookConfiguration, err := r.getMutatingWebhookConfiguration(ctx)
 	if err != nil {
-		return reconcile.Result{}, err
+		log.Info("could not find mutating webhook configuration, this is normal when deployed using OLM")
```
chrismuellner, line 67: shouldn't this stop the remaining reconciliation from executing?
in the case of OLM the certificates do not need to be created manually

chrismuellner: the comment on the tests that we are waiting for the certificate secret would make sense here.

should we rethink if always waiting for the certificate secret before startup makes sense?

src/controllers/certificates/webhook_configuration_test.go
```diff
@@ -0,0 +1,64 @@
+package certificates
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/admissionregistration/v1"
```
chrismuellner, line 7:
```suggestion
	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
```

src/controllers/certificates/certificate_secret_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Dynatrace/dynatrace-operator/src/scheme/fake"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	v1 "k8s.io/api/core/v1"
```
chrismuellner, line 11:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

src/controllers/certificates/certificate_secret.go
```diff
@@ -0,0 +1,130 @@
...
+	"bytes"
+	"context"
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
+	"github.com/Dynatrace/dynatrace-operator/src/webhook"
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/admissionregistration/v1"
```
chrismuellner, line 13:
```suggestion
	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
```

chrismuellner (changes requested):

src/controllers/certificates/webhook_configuration_test.go
meik99: done

src/controllers/certificates/certificate_secret_test.go
meik99: done

src/controllers/certificates/certificate_secret.go
meik99: done

config/common/webhook/deployment-webhook.yaml
chrismuellner: will be removed after logic rework

Makefile
```diff
@@ -199,10 +199,15 @@ bundle: manifests kustomize
...
-	echo 'LABEL com.redhat.delivery.backport=true' >> ./config/olm/$(PLATFORM)/bundle-$(VERSION).Dockerfile
+	echo 'LABEL com.redhat.delivery.backport=true' >> ./config/olm/$(PLATFORM)/bundle-$(VERSION).Dockerfile	
 ifeq ($(PLATFORM), openshift)
 	sed 's/\bkubectl\b/oc/g' ./config/olm/$(PLATFORM)/$(VERSION)/manifests/dynatrace-operator.v$(VERSION).clusterserviceversion.yaml > ./config/olm/$(PLATFORM)/$(VERSION)/manifests/dynatrace-operator.v$(VERSION).clusterserviceversion.yaml.output
 	mv ./config/olm/$(PLATFORM)/$(VERSION)/manifests/dynatrace-operator.v$(VERSION).clusterserviceversion.yaml.output ./config/olm/$(PLATFORM)/$(VERSION)/manifests/dynatrace-operator.v$(VERSION).clusterserviceversion.yaml
+	echo '  com.redhat.openshift.versions: v4.5-v4.9' >> ./config/olm/$(PLATFORM)/$(VERSION)/metadata/annotations.yaml
+	mv ./config/olm/$(PLATFORM)/$(VERSION)/manifests/dynatrace-operator.v$(VERSION).clusterserviceversion.yaml ./config/olm/$(PLATFORM)/$(VERSION)/manifests/dynatrace-operator.clusterserviceversion.yaml
+
+	grep -v 'scorecard' ./config/olm/$(PLATFORM)/$(VERSION)/metadata/annotations.yaml > ./config/olm/$(PLATFORM)/$(VERSION)/metadata/annotations.yaml.output
```
chrismuellner, line 209: this is also relevant if `PLATFORM` is `kubernetes`.

The header `# Annotations for testing.` is also still present.

meik99: adapted makefile

chrismuellner (approved):

PR #425, opened 2021-12-26
Adding k8s-endpoint as command line arg

hackerxben:
Add the possibility to define K8S_ENDPOINT via command line args , If it fails to find the arg the script will default back to the original way of getting the k8s endpoint


chrismuellner:
`install.sh` won't be updated further

PR #427, opened 2022-01-03
Creates symlink for current version in CSI driver in the agent bin

0sewa0:
```
I have no name!@java-glibc-56978775fb-rbpfs:/$ ls -l opt/dynatrace/oneagent-paas/agent/bin/
total 16
drwxr-xr-x 6 root root 4096 Jan  3 15:14 1.233.1.20211231-040645
drwxr-xr-x 2 root root 4096 Jan  3 15:14 any
lrwxrwxrwx 1 root root   23 Jan  3 15:24 current -> 1.233.1.20211231-040645     <== ADDS THIS SYMLINK 
drwxr-xr-x 2 root root 4096 Jan  3 15:14 linux-musl-x86-64
drwxr-xr-x 2 root root 4096 Jan  3 15:14 linux-x86-64
```

### DIFF
--- a/src/controllers/csi/metadata/path_resolver.go
+++ b/src/controllers/csi/metadata/path_resolver.go
@@ -34,6 +34,14 @@ func (pr PathResolver) AgentBinaryDirForVersion(tenantUUID string, version strin
 	return filepath.Join(pr.AgentBinaryDir(tenantUUID), version)
 }
 
+func (pr PathResolver) RelativeSymlinkForVersion(version string) string {
+	return filepath.Join(".", version)
+}
+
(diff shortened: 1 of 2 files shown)


src/controllers/csi/metadata/path_resolver.go
```diff
@@ -34,6 +34,14 @@ func (pr PathResolver) AgentBinaryDirForVersion(tenantUUID string, version strin
 	return filepath.Join(pr.AgentBinaryDir(tenantUUID), version)
 }
 
+func (pr PathResolver) RelativeSymlinkForVersion(version string) string {
```
chrismuellner, line 37: this is not a symlink yet, just the directory

chrismuellner (approved):

PR #433, opened 2022-01-05
Propagates tls cert to oneagents

0sewa0:
If `tlsSecretName` is set in the activeGate section then the value of the `server.crt` key in the provided secret will be propagated into
the oneagent Daemonset, and to the standalone agents. 

The value is put into `/var/lib/dynatrace/oneagent/agent/customkeys/custom.pem`.

### DIFF
--- a/src/controllers/dynakube/oneagent/daemonset/volumes.go
+++ b/src/controllers/dynakube/oneagent/daemonset/volumes.go
@@ -13,6 +13,10 @@ func prepareVolumeMounts(instance *dynatracev1beta1.DynaKube) []corev1.VolumeMou
 		volumeMounts = append(volumeMounts, getCertificateMount())
 	}
 
+	if instance.HasActiveGateTLS() {
+		volumeMounts = append(volumeMounts, getTLSMount())
+	}
+
 	volumeMounts = append(volumeMounts, rootMount)
 	return volumeMounts
 }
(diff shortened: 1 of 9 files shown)


config/samples/applicationMonitoring.yaml
```diff
@@ -137,3 +137,8 @@ spec:
     #     [connectivity]
     #     networkZone=
     #   valueFrom: myCustomPropertiesConfigMap
+
+    # Optional: the name of a secret containing ActiveGate TLS cert+key and password. If not set, self-signed certificate is used.
+    # server.p12: certificate+key pair in pkcs12 format
+    #
+    # tlsSecretName: "my-tls-secret"
```
chrismuellner, line 144: please move the samples into a separate PR as it's not connected to the logic change :)

meik99 (approved):

src/controllers/dynakube/oneagent/daemonset/volumes.go
```diff
@@ -24,6 +28,13 @@ func getCertificateMount() corev1.VolumeMount {
 	}
 }
 
+func getTLSMount() corev1.VolumeMount {
+	return corev1.VolumeMount{
+		Name:      "tls",
+		MountPath: "/var/lib/dynatrace/oneagent/agent/customkeys",
```
mjgrzybek, line 34: Please make it const and reuse it in `pod_mutator.go:717`

mjgrzybek (approved):

PR #439, opened 2022-01-10
Introduces extra check to create "partial" yamls

0sewa0:
Now we have 3 kubernetes yamls: (same for openshift):
- /config/deploy/kubernetes/**kubernetes.yaml**: All the stuff without the CSI driver stuff
- /config/deploy/kubernetes/**kubernetes-csi.yaml**: ONLY the CSI driver stuff
- /config/deploy/kubernetes/**kubernetes-all.yaml**: EVERYTHING (kubernetes.yaml + kubernetes-csi.yaml)

Now the user can easily deploy/remove the CSI driver related resources.

`kubernetes-all.yaml` and `openshift-all.yaml` is used for the local deployment scripts and olm


chrismuellner:
Do we need `kubernetes-all.yaml`?

0sewa0:
> Do we need `kubernetes-all.yaml`?

its mainly for dev stuff, the user doesn't need it AFAIK.
but its nicer in dev scripts to have 1 `kubectl apply` instead of 2

config/helm/chart/default/templates/Common/csi/clusterrole-csi.yaml
```diff
@@ -1,3 +1,5 @@
+{{- $platformIsSet := printf "%s" (required "Platform needs to be set to kubernetes, openshift" (include "dynatrace-operator.platformSet" .))}}
+{{ if eq (include "dynatrace-operator.needCSI" .) "true" }}
```
meik99, line 2: since we now let the user decide if they want to apply *-csi.yaml or not, do we still need the `needCSI` function?

config/helm/chart/default/templates/Common/activegate/serviceaccount-activegate.yaml
```diff
@@ -1,3 +1,5 @@
+{{- $platformIsSet := printf "%s" (required "Platform needs to be set to kubernetes, openshift, google" (include "dynatrace-operator.platformSet" .))}}
+{{ if eq (include "dynatrace-operator.partial" .) "false" }}
```
meik99, line 2: `partial` is not really a meaningful name.
consider changing it to something like, e.g.,  `isCSI`

meik99 (changes requested):

config/helm/chart/default/templates/Common/csi/clusterrole-csi.yaml
0sewa0: still needed for the `helm` stuff

config/helm/chart/default/templates/Common/activegate/serviceaccount-activegate.yaml
0sewa0: my idea was that we could have other stuff in the future that we want to have a separate yaml for so that is why I chosen the name `partial`

if you look at the definition of `dynatrace-operator.partial` at https://github.com/Dynatrace/dynatrace-operator/pull/439/files#diff-212cec223aa4e2a1a063b33607deb1de9a46b96f698d7d9c0e57990f02a30cd7R100
it is `false` if not set to anything and `true` if set to anything.
In the future it could have multiple possible values like "csi" or "webhook" or whatever. So we only have 1 if per file for this instead of multiple.
 
I should replace this `partial=true` with `partial=csi` so it makes more sense (also do some other stuff)😅

chrismuellner (approved):

PR #440, opened 2022-01-11
Removed environment variable specifying the related image for OLM releases

meik99:


src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -27,7 +27,7 @@ const (
 
 	hostRootMount = "host-root"
 
-	relatedImageEnvVar = "RELATED_IMAGE_DYNATRACE_ONEAGENT"
+	relatedImage = "registry.connect.redhat.com/dynatrace/oneagent:latest"
```
chrismuellner, line 30: related image should not be used at all anymore, so the env variable can be fully removed

chrismuellner (changes requested):

src/controllers/dynakube/oneagent/daemonset/daemonset.go
luhi-DT: If possible we could use instead of that relatedImage field the "DEPLOYED_VIA_OLM" directly in the image() function, so we avoid setting images at two different places, and avoid passing variables via the builderInfo

0sewa0 (approved):

PR #441, opened 2022-01-11
AG capability renamed from 'data-ingest' to 'metrics-ingest'

mjgrzybek:
AG's capability renamed from `data-ingest` to `metrics-ingest`
`data-ingest` remains valid in terms of injection type


0sewa0:
if `data-ingest-token` and `data-ingest-endpoint` are meant to be excluded from the change, then all looks fine

0sewa0 (approved):

PR #445, opened 2022-01-12
Adds preview warning for metrics-ingest (+minor fixes)

0sewa0:
When `metrics-ingest` is configured in the activegate section a warning will tell the user that its a preview feature

There were some minor mistakes in the codebase that it also fixes, regarding the validation webhook yaml and `IsActiveGateMode` function

### DIFF
--- a/src/webhook/validation/preview.go
+++ b/src/webhook/validation/preview.go
@@ -7,16 +7,26 @@ import (
 )
 
 const (
-	warningPreview = `%s mode is in PREVIEW. Please be aware that it is NOT production ready and you may run into bugs.`
+	oneAgentModePreviewWarningMessage  = `%s mode is in PREVIEW.`
+	metricsIngestPreviewWarningMessage = `metrics-ingest is in PREVIEW.`
+	basePreviewWarning                 = "PREVIEW features are NOT production ready and you may run into bugs."
 )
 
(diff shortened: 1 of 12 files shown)


src/webhook/validation/preview.go
```diff
@@ -7,16 +7,26 @@ import (
 )
 
 const (
-	warningPreview = `%s mode is in PREVIEW. Please be aware that it is NOT production ready and you may run into bugs.`
+	oneAgentModePreviewWarningMessage  = `%s mode is in PREVIEW.`
```
chrismuellner, line 10: this message could be reused for `metricsIngestPreviewWarningMessage`

src/webhook/validation/validation_test.go
```diff
@@ -196,7 +196,7 @@ func assertAllowedResponseWithoutWarnings(t *testing.T, dynakube *dynatracev1bet
 
 func assertAllowedResponseWithWarnings(t *testing.T, dynakube *dynatracev1beta1.DynaKube, other ...client.Object) {
 	response := assertAllowedResponse(t, dynakube, other...)
-	assert.Equal(t, len(response.Warnings), 1)
+	assert.Greater(t, len(response.Warnings), 1)
```
chrismuellner, line 199: why can't equal be used here? we know what the dynakube looks like

chrismuellner (changes requested):

chrismuellner (approved):

PR #447, opened 2022-01-12
Updates the pre-commit to generate the manifests for master

0sewa0:
You still need to commit the changes by hand, but its harder to forget now


.github/pre-commit
```diff
@@ -8,6 +8,18 @@
...
+## changed since the last commit
+STAGED_YAML_FILES=$(git diff --cached --name-only -- '*.yaml')
+
+## we can check to see if this is empty
+if [[ $STAGED_YAML_FILES == "" ]]; then
+    echo "No Yaml Files to Update"
+## otherwise we generate the manifests for master
+else
+    make manifests
+fi
```
meik99, line 21: Logging that no changes happended isn't necessary imo. Consider changing it to something like this.

```suggestion
if [[ $STAGED_YAML_FILES != "" ]]; then
    echo "YAML files have changed, generating manifests"
    make manifests
fi
```

meik99 (changes requested):

chrismuellner (approved):

PR #448, opened 2022-01-12
Adds dynakube feature flag to disable metadata enrichment

0sewa0:
The new feature flag overrules the pod annotation to enable/disable metadata enrichment.

### DIFF
--- a/src/controllers/dynakube/dynakube_controller.go
+++ b/src/controllers/dynakube/dynakube_controller.go
@@ -254,9 +254,11 @@ func (r *ReconcileDynaKube) reconcileDynaKube(ctx context.Context, dkState *stat
 			return
 		}
 
-		upd, err = dtingestendpoint.NewEndpointSecretGenerator(r.client, r.apiReader, dkState.Instance.Namespace).GenerateForDynakube(ctx, dkState.Instance)
-		if dkState.Error(err) || dkState.Update(upd, defaultUpdateInterval, "new data-ingest endpoint secret created") {
-			return
+		if !dkState.Instance.FeatureDisableMetadataEnrichment() {
+			upd, err = dtingestendpoint.NewEndpointSecretGenerator(r.client, r.apiReader, dkState.Instance.Namespace).GenerateForDynakube(ctx, dkState.Instance)
+			if dkState.Error(err) || dkState.Update(upd, defaultUpdateInterval, "new data-ingest endpoint secret created") {
+				return
+			}
 		}
 	} else {
 		if err := dkMapper.UnmapFromDynaKube(); err != nil {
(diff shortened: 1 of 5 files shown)


src/controllers/dynakube/dynakube_controller.go
```diff
@@ -254,15 +254,26 @@ func (r *ReconcileDynaKube) reconcileDynaKube(ctx context.Context, dkState *stat
 			return
 		}
 
-		upd, err = dtingestendpoint.NewEndpointSecretGenerator(r.client, r.apiReader, dkState.Instance.Namespace).GenerateForDynakube(ctx, dkState.Instance)
-		if dkState.Error(err) || dkState.Update(upd, defaultUpdateInterval, "new data-ingest endpoint secret created") {
-			return
+		if !dkState.Instance.FeatureDisableMetadataEnrichment() {
+			upd, err = dtingestendpoint.NewEndpointSecretGenerator(r.client, r.apiReader, dkState.Instance.Namespace).GenerateForDynakube(ctx, dkState.Instance)
```
chrismuellner, line 258: can `dtingestendpoint.NewEndpointSecretGenerator(r.client, r.apiReader, dkState.Instance.Namespace)` be moved out of the if and reused?

src/webhook/mutation/pod_mutator.go
```diff
@@ -224,6 +224,10 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
 		return *dkResponse
 	}
 
+	if dk.FeatureDisableMetadataEnrichment() {
+		injectionInfo.features[DataIngest] = false
```
chrismuellner, line 228: can this be moved to a function that enables/disables all the features for the `injectionInfo`?

src/webhook/mutation/pod_mutator.go
```diff
@@ -233,9 +237,13 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
 		return *secretResponse
 	}
 
-	dataIngestFields, diResponse := m.ensureDataIngestSecret(ctx, ns, dkName, dk)
-	if diResponse != nil {
-		return *diResponse
+	dataIngestFields := map[string]string{}
+	if injectionInfo.enabled(DataIngest) {
+		var diResponse *admission.Response
+		dataIngestFields, diResponse = m.ensureDataIngestSecret(ctx, ns, dkName, dk)
```
chrismuellner, line 243: can this return an `error` instead of a `response`? the actual error response returned is always the same and could just be returned here once

chrismuellner (changes requested):

src/webhook/mutation/pod_mutator.go
0sewa0: injectionInfo is created before getting the dynakube, so to move this logic up we would always need to get the namespace and dynakube to create the `injectionInfo` which is not nice

chrismuellner (approved):

PR #457, opened 2022-01-16
Refactor nodes controller - splitted Reconcilers

luhi-DT:
Instead of using channels to communicate change (delete, update)  from the corresponding EventHandler events and periodically iterate over all nodes, we should use Reconcile for Node Objects. This PR uses a splitted Reconcile in order to mark graceful shutdown nodes as terminated. One dedicated Reconcile just for deletions in order to detect them safe and the other Reconcile for any other change to a node. 

Both Reconcilers are added in the Add function to ensure that no-one miss-uses the controller. The rest should be the same as in this [PR](https://github.com/Dynatrace/dynatrace-operator/pull/456).

**Unit tests tbd**


src/controllers/nodes/nodes_deletion_controller.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (r *ReconcileNodeDeletion) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Node{}).
+		WithEventFilter(nodeDeletionPredicate()).
+		Complete(r)
+}
+
+func nodeDeletionPredicate() predicate.Predicate {
+	return predicate.Funcs{
```
0sewa0, line 26: I think 2 reconcilers are not necessary.

You can use these predicates to do some wild stuff.
But for our uses I think what we should do is have a predicate just for the delete scenario because all other cases are the kinda the same. (if not we could do the same for those scenarios aswell)

In that predicate with could do the actual logic because the `event.DeleteEvent` has the object you need, you only need to get the name and you basically have all you need.
```		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
			nodeName := deleteEvent.Object.GetName()
			doDeleteLogic(nodeName)
			return false
		},
```
what is missing is all the `stuff` within the node controller `struct`, those things can be passed or you can just pass in the whole `struct` while you are at it 😄 
so it could look like this:
```
type NodeController struct {...}

func (controller *NodeController) SetupWithManager(mgr ctrl.Manager) error {
	return ctrl.NewControllerManagedBy(mgr).
		For(&corev1.Node{}).
		WithEventFilter(nodeDeletionPredicate(controller)).
		Complete(r)
}

func nodeDeletionPredicate(controller *NodeController) predicate.Predicate {
	return predicate.Funcs{
		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
			nodeName := deleteEvent.Object.GetName()
			controller.doDeleteLogic(nodeName)
			return false
		}
	}

func (controller *NodeController) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {...}

func (controller *NodeController) doDeleteLogic(nodeName string)  {...}
```

PR #459, opened 2022-01-17
Adds github actions to test/lint helm chart

0sewa0:


config/helm/testing/lint.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+for directory in "$(pwd)"; do
```
chrismuellner, line 3: is this necessary here since we only have a single chart in this repository?

chrismuellner (approved):

chrismuellner (approved):

PR #461, opened 2022-01-17
Adds token scope check for activeGate features

0sewa0:
In case of select activeGate features the apiToken needs special scopes, if these are not present we notify the user

Automatic kubernetes api monitoring need:
`entities.read`
`entities.write`

Metrics ingest need:
`metrics.ingest`

### DIFF
--- a/src/controllers/dynakube/dtclient_reconciler.go
+++ b/src/controllers/dynakube/dtclient_reconciler.go
@@ -144,6 +144,14 @@ func (r *DynatraceClientReconciler) Reconcile(ctx context.Context, instance *dyn
 			}}
 	}
 
+	if instance.FeatureAutomaticKubernetesApiMonitoring() {
+		tokens[0].Scopes = append(tokens[0].Scopes, dtclient.TokenScopeEntitiesRead, dtclient.TokenScopeEntitiesWrite)
+	}
+
+	if instance.IsActiveGateMode(dynatracev1beta1.MetricsIngestCapability.DisplayName) {
+		tokens[0].Scopes = append(tokens[0].Scopes, dtclient.TokenScopeMetricsIngest)
+	}
+
 	for _, token := range tokens {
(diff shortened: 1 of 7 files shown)


src/controllers/dynakube/dtclient_reconciler.go
```diff
@@ -144,6 +144,14 @@ func (r *DynatraceClientReconciler) Reconcile(ctx context.Context, instance *dyn
 			}}
 	}
 
+	if instance.FeatureAutomaticKubernetesApiMonitoring() {
```
chrismuellner, line 147: check should include `KubernetesMonitoringMode()`

src/dtclient/client.go
```diff
@@ -107,6 +107,9 @@ const (
 	TokenScopeDataExport        = "DataExport"
 	TokenScopeReadConfig        = "ReadConfig"
 	TokenScopeWriteConfig       = "WriteConfig"
+	TokenScopeMetricsIngest     = "metrics.ingest"
```
chrismuellner, line 110: why are these tokens in a different format than the older ones?

chrismuellner (changes requested):

src/dtclient/client.go
0sewa0: its not up to me :)

src/controllers/dynakube/dtclient_reconciler.go
```diff
@@ -144,6 +144,15 @@ func (r *DynatraceClientReconciler) Reconcile(ctx context.Context, instance *dyn
 			}}
 	}
 
+	if instance.IsActiveGateMode(dynatracev1beta1.KubeMonCapability.DisplayName) &&
```
chrismuellner, line 147: this does not include the deprecated kubemon section, use `KubernetesMonitoringMode()` instead

chrismuellner (approved):

PR #466, opened 2022-01-19
Removes entities.write scope check, metrics.ingest scope checked for dataIngestToken

0sewa0:


src/controllers/dynakube/dtclient_reconciler.go
```diff
@@ -146,11 +146,17 @@ func (r *DynatraceClientReconciler) Reconcile(ctx context.Context, instance *dyn
 
 	if instance.KubernetesMonitoringMode() &&
 		instance.FeatureAutomaticKubernetesApiMonitoring() {
-		tokens[0].Scopes = append(tokens[0].Scopes, dtclient.TokenScopeEntitiesRead, dtclient.TokenScopeEntitiesWrite)
+		tokens[0].Scopes = append(tokens[0].Scopes, dtclient.TokenScopeEntitiesRead)
 	}
 
-	if instance.IsActiveGateMode(dynatracev1beta1.MetricsIngestCapability.DisplayName) {
-		tokens[0].Scopes = append(tokens[0].Scopes, dtclient.TokenScopeMetricsIngest)
+	if instance.IsActiveGateMode(dynatracev1beta1.MetricsIngestCapability.DisplayName) && r.DataIngestToken != "" {
```
chrismuellner, line 152:
```suggestion
	if r.DataIngestToken != "" {
```
should give a warning that the token is missing scope, but not block creation of the ActiveGate!

chrismuellner (changes requested):

src/controllers/dynakube/dtclient_reconciler.go
0sewa0: why would this block the creation of the activeGate ?

chrismuellner: it doesn't, I confused it with a different error :slightly_smiling_face:

chrismuellner (approved):

PR #468, opened 2022-01-20
Restructures the csidriver package

0sewa0:
The idea is that in the future the csi driver will be able to mount different types of volumes. (standalone-agent, host-agent etc). 
So the new structure separates the `server` and a set of `Publisher`s
- The `server` handles the parsing of the publish/unplublish requests and decides to call the correct `Publisher` for the request to do its magic.
- A `Publisher` is responsible to do the actual publishing and unpublishing of the volumes.

**Notes**: 
- The names of stuff (package, types etc) is not final so suggestions are welcome.
- Because the 2. publisher is not implemented yet some functions will probably move into the shared package of the publishers to reuse code.


chrismuellner (changes requested):

src/controllers/csi/driver/server.go
```diff
@@ -246,68 +198,14 @@ func (svr *CSIDriverServer) NodeExpandVolume(context.Context, *csi.NodeExpandVol
...
-	volume, err := svr.db.GetVolume(volumeID)
-	if err != nil {
-		return nil, err
-	}
-	if volume == nil {
-		return &metadata.Volume{}, nil
+func isMounted(mounter mount.Interface, targetPath string) (bool, error) {
+	isNotMounted, err := mount.IsNotMountPoint(mounter, targetPath)
+	if os.IsNotExist(err) {
+		isNotMounted = true
```
chrismuellner, line 204: `isNotMounted` variable isn't necessary
could just return true here

src/controllers/csi/driver/volumes/app/publisher.go
```diff
@@ -0,0 +1,175 @@
...
+	return &Publisher{
+		client:  client,
+		fs:      fs,
+		mounter: mounter,
+		db:      db,
+		path:    path,
+	}
+}
+
+type Publisher struct {
```
chrismuellner, line 47: it's confusing that this is named the same way as the interface

chrismuellner (changes requested):

chrismuellner (approved):

PR #481, opened 2022-01-24
ActiveGate dynatrace-api capability added to samples

mjgrzybek:


luhi-DT:
Do we need this also here https://github.com/Dynatrace/dynatrace-operator/blob/master/config/samples/applicationMonitoring.yaml ? :)

chrismuellner (changes requested):
Please add the new capability to the `applicationMonitoring.yaml` sample as well

chrismuellner (approved):

PR #462, opened 2022-01-18
Renames reconcilers/controllers (2. try)

0sewa0:
Same as https://github.com/Dynatrace/dynatrace-operator/pull/417 (merge conflicts were a nightmare, easier to redo everything)

PR #483, opened 2022-01-24
Refactor istio controller

waodim:
> Moved checks for istio availability to kubeobjects.

> Added two new validators, which check for istio

> Adapted dynakubeValidator struct and added rest.Config, in order to be able to make those checks.


src/kubeobjects/istio.go
```diff
@@ -0,0 +1,237 @@
+package kubeobjects
```
0sewa0, line 1: not a fan of having all this stuff in `kubeobjects`, this package is for stuff that are "native" to kubernetes, which is not really true for istio 😄 

all of these could fit in the istio reconciler's package, because they are not used anywhere else (with a few exceptions, and even those don't need to be here)

src/controllers/dynakube/istio/controller.go
```diff
@@ -66,7 +52,7 @@ func (c *Controller) initializeIstioClient(config *rest.Config) (istioclientset.
 // creating/deleting VS & SE for external communications
 func (c *Controller) ReconcileIstio(instance *dynatracev1beta1.DynaKube) (updated bool, err error) {
```
0sewa0, line 53: nitpick: the `istio` Controller is not really a controller, so in this https://github.com/Dynatrace/dynatrace-operator/pull/462 it will be renamed to Reconciler

src/kubeobjects/istio.go
0sewa0: I will leave a bunch of comments where you could put these functions inside the `istio` package, you don't have to follow them, they will be just there to help you with naming and stuff 😉

src/kubeobjects/istio.go
```diff
@@ -0,0 +1,237 @@
...
+	// ServiceEntryGVK => definition of virtual service GVK for oneagent
+	ServiceEntryGVK = schema.GroupVersionKind{
+		Group:   IstioGVRName,
+		Version: "v1alpha3",
+		Kind:    "ServiceEntry",
+	}
+)
+
+// CheckIstioEnabled checks if Istio is installed
+func CheckIstioEnabled(cfg *rest.Config) (bool, error) {
```
0sewa0, line 56: could be in a `validate.go` under the `istio` package, as its used for validating both in the reconcile and in the validation webhook

src/kubeobjects/istio.go
```diff
@@ -0,0 +1,237 @@
...
+		"oneagent":             name,
+		"dynatrace-istio-role": role,
+	}
+}
+
+func isIp(host string) bool {
+	return net.ParseIP(host) != nil
+}
+
+func VerifyIstioCrdAvailability(instance *dynatracev1beta1.DynaKube, config *rest.Config) probeResult {
```
0sewa0, line 99: could be in a `validate.go` under the `istio` package, as its used for validating both in the reconcile and in the validation webhook

src/kubeobjects/istio.go
```diff
@@ -0,0 +1,237 @@
...
+
+	probe, _ = kubernetesObjectProbe(VirtualServiceGVK, instance.GetNamespace(), "", config)
+	if probe == ProbeTypeNotFound {
+		return probe
+	}
+
+	return ProbeTypeFound
+}
+
+func kubernetesObjectProbe(gvk schema.GroupVersionKind,
```
0sewa0, line 115: this could stay in the `kubeobjects` package because its super general and possible can be reused in the future.
maybe in a `objects.go` file ?

src/kubeobjects/istio.go
```diff
@@ -0,0 +1,237 @@
...
+	if name == "" {
+		err = runtimeClient.List(context.TODO(), &objQuery, client.InNamespace(namespace))
+	} else {
+		err = runtimeClient.Get(context.TODO(), client.ObjectKey{Namespace: namespace, Name: name}, &objQuery)
+	}
+
+	return mapErrorToObjectProbeResult(err)
+}
+
+func mapErrorToObjectProbeResult(err error) (probeResult, error) {
```
0sewa0, line 136: connected to the previous functions, should stay with it

src/kubeobjects/istio.go
```diff
@@ -0,0 +1,237 @@
...
+			return ProbeTypeNotFound, err
+		}
+
+		return probeUnknown, err
+	}
+
+	return probeObjectFound, nil
+}
+
+func HandleIstioConfigurationForServiceEntry(instance *dynatracev1beta1.DynaKube,
```
0sewa0, line 150: `serviceentry.go` under the `istio` package
you shouldn't pass in a logger, use the package level logger instead

src/kubeobjects/istio.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		//log.Error(err, "istio: failed to create ServiceEntry")
+		return false, err
+	}
+	log.Info("istio: ServiceEntry created", "objectName", name, "host", communicationHost.Host, "port", communicationHost.Port)
+
+	return true, nil
+}
+
+func HandleIstioConfigurationForVirtualService(instance *dynatracev1beta1.DynaKube,
```
0sewa0, line 173: `virtualservice.go` under the `istio` package
you shouldn't pass in a logger, use the package level logger instead

src/kubeobjects/istio.go
```diff
@@ -0,0 +1,237 @@
...
+		//log.Error(err, "istio: failed to create VirtualService")
+		return false, err
+	}
+	log.Info("istio: VirtualService created", "objectName", name, "host", communicationHost.Host,
+		"port", communicationHost.Port, "protocol", communicationHost.Protocol)
+
+	return true, nil
+}
+
+func createIstioConfigurationForServiceEntry(dynaKube *dynatracev1beta1.DynaKube,
```
0sewa0, line 202: `serviceentry.go` under the `istio` package

src/kubeobjects/istio.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		return err
+	}
+	if sve == nil {
+		return fmt.Errorf("could not create service entry with spec %v", serviceEntry.Spec)
+	}
+	return nil
+}
+
+func createIstioConfigurationForVirtualService(dynaKube *dynatracev1beta1.DynaKube,
```
0sewa0, line 220: `virtualservice.go` under the `istio` package

src/kubeobjects/serviceentry.go
```diff
@@ -1,4 +1,4 @@
-package istio
+package kubeobjects
```
0sewa0, line 1: this was good where it was,
a `ServiceEntry` is an isito specific thingy, should stay "locked away" in the `istio` package 😛

src/kubeobjects/virtualservice.go
```diff
@@ -1,16 +1,16 @@
-package istio
+package kubeobjects
```
0sewa0, line 1: this was good where it was,
a `VirtualService` is an isito specific thingy, should stay "locked away" in the `istio` package 😛

src/kubeobjects/istio.go
```diff
@@ -0,0 +1,237 @@
...
+	for _, apiGroup := range apiGroupList.Groups {
+		if apiGroup.Name == IstioGVRName {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// BuildNameForEndpoint returns a name to be used as a base to identify Istio objects.
+func BuildNameForEndpoint(name string, protocol string, host string, port uint32) string {
```
0sewa0, line 75: this and the next 3 little helper functions could be stored next/near to where they are used
so in the `istio` package 😉

0sewa0 (changes requested):

chrismuellner (changes requested):

src/kubeobjects/istio.go
waodim: Did that and everything else you mentioned. Main problem: if this stays in the kubeobjects package in a objects.go file as suggested, we run into an import cycle error, as this method needs our defined probeResult type which is defined within the istio package. Do you have an idea how to circumvent this problem?

waodim: Thank you very much - it works!

chrismuellner:
The issues with linting should be fixed by getting the changes on the master branch.
Check #461 for the linting errors.

src/kubeobjects/objects.go
```diff
@@ -0,0 +1,57 @@
...
+
+const (
+	ProbeObjectFound ProbeResult = iota
+	ProbeObjectNotFound
+	ProbeTypeFound
+	ProbeTypeNotFound
+	ProbeUnknown
+)
+
+func KubernetesObjectProbe(gvk schema.GroupVersionKind,
```
meik99, line 24: too many paramters

src/controllers/dynakube/istio/virtualservice.go
```diff
@@ -66,3 +78,51 @@ func buildVirtualServiceTLSRoute(host string, port uint32) []*istio.TLSRoute {
...
+
+func createIstioConfigurationForVirtualService(dynaKube *dynatracev1beta1.DynaKube,
+	virtualService *istiov1alpha3.VirtualService, role string,
+	istioClient istioclientset.Interface, scheme *runtime.Scheme) error {
+
+	virtualService.Labels = BuildIstioLabels(dynaKube.GetName(), role)
+	if err := controllerutil.SetControllerReference(dynaKube, virtualService, scheme); err != nil {
+		return err
+	}
+	vs, err := istioClient.NetworkingV1alpha3().VirtualServices(dynaKube.GetNamespace()).Create(context.TODO(), virtualService, metav1.CreateOptions{})
```
meik99, line 119:
```suggestion
	createdVirtualService, err := istioClient.NetworkingV1alpha3().VirtualServices(dynaKube.GetNamespace()).Create(context.TODO(), virtualService, metav1.CreateOptions{})
```

src/controllers/dynakube/istio/istio.go
```diff
@@ -6,86 +6,49 @@ import (
...
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/discovery"
-	"k8s.io/client-go/rest"
 )
 
 var (
-	istioGVRName = "networking.istio.io"
+	IstioGVRName = "networking.istio.io"
```
meik99, line 14: does this need to be exported?

src/controllers/dynakube/istio/istio.go
```diff
@@ -6,86 +6,49 @@ import (
...
-		if apiGroup.Name == istioGVRName {
-			return true, nil
-		}
-	}
-	return false, nil
-}
-
 // BuildNameForEndpoint returns a name to be used as a base to identify Istio objects.
-func buildNameForEndpoint(name string, protocol string, host string, port uint32) string {
+func BuildNameForEndpoint(name string, protocol string, host string, port uint32) string {
```
meik99, line 32: Does this need to be exported?

src/controllers/dynakube/istio/istio.go
```diff
@@ -6,86 +6,49 @@ import (
...
-		}
-
-		return probeUnknown, err
-	}
-
-	return probeObjectFound, nil
-}
-
-func buildIstioLabels(name, role string) map[string]string {
+func BuildIstioLabels(name, role string) map[string]string {
```
meik99, line 44: Does this need to be exported?

src/controllers/dynakube/istio/reconciler.go
```diff
@@ -39,18 +30,18 @@ type IstioReconciler struct {
 
 // NewIstioReconciler - creates new instance of istio controller
 func NewIstioReconciler(config *rest.Config, scheme *runtime.Scheme) *IstioReconciler {
-	c := &IstioReconciler{
+	r := &IstioReconciler{
```
meik99, line 33:
```suggestion
	reconciler := &IstioReconciler{
```

src/controllers/dynakube/istio/istio.go
```diff
@@ -6,86 +6,49 @@ import (
...
 	return map[string]string{
 		"dynatrace":            "oneagent",
 		"oneagent":             name,
 		"dynatrace-istio-role": role,
 	}
 }
 
-func isIp(host string) bool {
+func IsIp(host string) bool {
```
meik99, line 52: Does this need to be exported?

src/controllers/dynakube/istio/reconciler.go
```diff
@@ -112,14 +103,14 @@ func (r *IstioReconciler) reconcileIstioConfigurations(instance *dynatracev1beta
 func (r *IstioReconciler) reconcileIstioRemoveConfigurations(instance *dynatracev1beta1.DynaKube,
 	comHosts []dtclient.CommunicationHost, role string) (bool, error) {
 
-	labelSelector := labels.SelectorFromSet(buildIstioLabels(instance.GetName(), role)).String()
+	labelSelector := labels.SelectorFromSet(BuildIstioLabels(instance.GetName(), role)).String()
 	listOps := &metav1.ListOptions{
 		LabelSelector: labelSelector,
 	}
 
 	seen := map[string]bool{}
```
meik99, line 111:
```suggestion
	seenComHosts := map[string]bool{}
```

src/controllers/dynakube/istio/reconciler.go
```diff
@@ -112,14 +103,14 @@ func (r *IstioReconciler) reconcileIstioConfigurations(instance *dynatracev1beta
...
 	for _, ch := range comHosts {
```
meik99, line 112:
```suggestion
	for _, comHost := range comHosts {
```

meik99 (dismissed):

src/controllers/dynakube/istio/serviceentry.go
```diff
@@ -60,3 +71,44 @@ func buildServiceEntryIP(name, namespace, host string, port uint32) *istiov1alph
 		},
 	}
 }
+
+func handleIstioConfigurationForServiceEntry(instance *dynatracev1beta1.DynaKube,
+	name string, communicationHost dtclient.CommunicationHost, role string, config *rest.Config, namespace string,
+	istioClient istioclientset.Interface, scheme *runtime.Scheme, log logr.Logger) (bool, error) {
+
+	probe, err := kubeobjects.KubernetesObjectProbe(ServiceEntryGVK, instance.GetNamespace(), name, config)
```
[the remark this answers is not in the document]
0sewa0: IMO, its fine here, as its a utility function so creating a separate struct wouldn't be that beneficial

src/controllers/dynakube/istio/istio.go
waodim: No, except for CheckIstioEnabled no function/variable stated from you needed to be exported, changed it, thank you

src/controllers/dynakube/istio/serviceentry.go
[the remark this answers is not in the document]
waodim: I could export the suggested struct which is needed in the `handleIstioConfigurationForVirtualService() `method for example and pass it here as well. What do you guys think about that, or should `KubernetesObjectProbe() `stay as it is?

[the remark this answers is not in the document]
0sewa0: `KubernetesObjectProbe ` should be independent from the istio package if its going to be in the `kubeobjects` package. (if you exported in the istio package and used in the kubeobjects packaged you would get import circle which is bad)

having a guideline of `no more then 2 params for a function` is nice and all, but that is too unrealistic and fragile to enforce seriously IMO. 
Because:
- having random single use structs are worse then a 3 or 4 param function.
- to couple 2 supposedly independent things with a struct, because the required params would match whats in the struct is not a good idea. You just made 2 loosely coupled things tightly coupled for no good reason.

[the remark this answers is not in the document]
waodim: in my opinion you stated very good points, on why it should stay the same. Thank you guys. (and as you mentioned, the import cycle would cause problems anyways)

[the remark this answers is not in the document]
meik99: `KubernetesObjectProbe`  can be made independent from the istio package if the struct is part of the `kubeobjects` package and not of the istio package. Furthermore, having multiple params is worse than having single use structs, because structs encapsulate data into a group which makes code better understandable.

In this case, for example, the paramter list `ServiceEntryGVK, instance.GetNamespace(), name, config` does not say much. However, if you bundle it in a struct it is much more understandable.
E.g.:
```
type KubernetesObjectProbe struct {
  serviceEntry string
  instanceName string
  name string
  config *rest.Config
}
```
The data is now grouped together.
Adding a function, now without parameters, allows for an easy to read call for a result.
```
func (probe *KubernetesObjectProbe) GetProbeResult() (ProbeResult, error) {... }
```

If you need to use it, the code would then look like so:
```
objectProbe := &kubeobjects.KubernetesObjectProbe {
 ...
}
probeResult, err := objectProbe.GetProbeResult()
```

It is more lines of code, but better read- and understandable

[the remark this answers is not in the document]
0sewa0: I agree to disagree :D

Using a struct like so would kinda warrant a constructor, which would result in a function with 4 params in this case 😬 
so we didn't really achieve anything just more code to maintain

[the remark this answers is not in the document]
chrismuellner: 🙂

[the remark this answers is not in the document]
meik99: Reasons to avoid long parameter lists
https://refactoring.guru/smells/long-parameter-list
https://www.arhohuttunen.com/long-parameter-list/
https://medium.com/@huytrongnguyen1985/lessons-learnt-from-the-clean-code-robert-c-martin-cecbe2b09139

If you are worried about long constructors, use the builder pattern
https://konfhub.medium.com/refactoring-long-parameter-list-in-constructors-java-e9c9f2cb1fb

chrismuellner (approved):

PR #485, opened 2022-01-24
Enable automatic kubernetes api monitoring by default in samples

chrismuellner:
* Set feature flag for all samples that use kubernetes monitoring capability


luhi-DT (dismissed):

0sewa0 (approved):

chrismuellner:
Won't do for now, so default scopes of `apiToken` don't change

PR #499, opened 2022-01-25
Moves all the dtclient enpoints into one place

0sewa0:
All dtclient enpoints that we use now are present in 1 place so its easier to see what we use


meik99 (approved):
lgtm

PR #501, opened 2022-01-25
Use different flavor in case of arm64 in the csi-provisioner

0sewa0:
based on: https://github.com/Dynatrace/dynatrace-operator/tree/feature/support-arm64-in-csi-provisioner


src/controllers/csi/provisioner/agent.go
```diff
@@ -120,11 +115,18 @@ func (installAgentCfg *installAgentConfig) installAgent(version, tenantUUID stri
...
 	}()
 
-	log.Info("downloading OneAgent package", "architecture", arch)
-	err = dtc.GetAgent(dtclient.OsUnix, dtclient.InstallerTypePaaS, dtclient.FlavorMultidistro, arch, version, tmpFile)
+	arch := dtclient.ArchX86
+	flavor := dtclient.FlavorMultidistro
+	if runtime.GOARCH == "arm64" {
+		arch = dtclient.ArchARM
+		flavor = dtclient.FlavorDefault
+	}
```
baichinger, line 123: These are constants anyways and therefore suggest to use [(implicit) build constraints](https://pkg.go.dev/cmd/go#hdr-Build_constraints).

src/controllers/csi/provisioner/arch/arm.go
```diff
@@ -0,0 +1,9 @@
+//go:build arm64
+// +build arm64
+
+package arch
+
+import "github.com/Dynatrace/dynatrace-operator/src/dtclient"
+
+const Flavor = dtclient.FlavorDefault
```
chrismuellner, line 8: should be the same order in both files, so it's clearer

chrismuellner (changes requested):

chrismuellner (approved):

PR #512, opened 2022-01-28
Add missing scopes for automatic API monitoring

fabwer:
- Fixes a bug when trying to unmarshal an error response while creating a settings object
- Improves the log message for unauthorized or forbidden api access

### DIFF
--- a/src/dtclient/kubernetes_settings.go
+++ b/src/dtclient/kubernetes_settings.go
@@ -234,13 +234,13 @@ func handleErrorArrayResponseFromAPI(response []byte, statusCode int) error {
 	if statusCode == http.StatusForbidden || statusCode == http.StatusUnauthorized {
 		var se getSettingsErrorResponse
 		if err := json.Unmarshal(response, &se); err != nil {
-			return fmt.Errorf("response error: %d, can't unmarshal json response: %w", statusCode, err)
+			return fmt.Errorf("response error: %d, can't unmarshal json response", statusCode)
 		}
 		return fmt.Errorf("response error: %d, %s", statusCode, se.ErrorMessage.Message)
 	} else {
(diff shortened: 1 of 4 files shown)


src/dtclient/kubernetes_settings.go
```diff
@@ -231,21 +231,29 @@ func createBaseRequest(url, method, apiToken string, body io.Reader) (*http.Requ
 }
 
 func handleErrorArrayResponseFromAPI(response []byte, statusCode int) error {
-	var se []getSettingsErrorResponse
-	if err := json.Unmarshal(response, &se); err != nil {
-		return fmt.Errorf("response error: %d, can't unmarshal json response: %w", statusCode, err)
-	}
+	if statusCode == http.StatusForbidden || statusCode == http.StatusUnauthorized {
+		var se getSettingsErrorResponse
```
chrismuellner, line 235: with the additional checks we hopefully shouldn't get any more forbidden/unauthorized responses, because it should fail before even attempting the request
let's leave it in for now. in the future we could handle responses with these 2 status codes in a reusable way

src/dtclient/kubernetes_settings.go
```diff
@@ -231,21 +231,29 @@ func createBaseRequest(url, method, apiToken string, body io.Reader) (*http.Requ
...
+		if err := json.Unmarshal(response, &se); err != nil {
+			return fmt.Errorf("response error: %d, can't unmarshal json response: %w", statusCode, err)
```
chrismuellner, line 237: maybe the whole stack trace is unnecessary
just returning the response code and that unmarshal failes should be enough

chrismuellner (changes requested):

chrismuellner (approved):

PR #514, opened 2022-01-31
Moves the tls in the correct postion for oneagent daemonset

0sewa0:
The oneagents deployed via a daemonset store their config on the root path of the node

### DIFF
--- a/src/controllers/dynakube/oneagent/daemonset/volumes.go
+++ b/src/controllers/dynakube/oneagent/daemonset/volumes.go
@@ -1,6 +1,8 @@
 package daemonset
 
 import (
+	"path/filepath"
+
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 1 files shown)


src/controllers/dynakube/oneagent/daemonset/volumes.go
```diff
@@ -33,7 +35,7 @@ func getCertificateMount() corev1.VolumeMount {
 func getTLSMount() corev1.VolumeMount {
 	return corev1.VolumeMount{
 		Name:      "tls",
-		MountPath: OneAgentCustomKeysPath,
+		MountPath: filepath.Join("/mnt/root", OneAgentCustomKeysPath),
```
chrismuellner, line 38: isn't it `/mnt/host_root`?

0sewa0: if I understand correctly we mount `/mnt/root` to `mnt/host_root`, but I want to test it to be sure :)

0sewa0: seems to work :D 
```
[root@gke-marcell--test-default-pool-67b1a6d3-m5k1 /]# cat mnt/host_root/var/lib/dynatrace/oneagent/agent/customkeys/custom.pem 
asd
```

0sewa0: my "secret": `kubectl -n dynatrace create secret generic tls-secret --from-literal="server.crt=asd"`

chrismuellner (approved):

PR #519, opened 2022-02-01
Refactor nodes controller

luhi-DT:
In this PR the Nodes Controller is refactored to be an actual Reconciler. Furthermore the delete logic, which formerly has be done by a separated DeleteHandler is now handled as a predicate. This should catch nodeDeletions properly. Like before a ConfigMap is built witch all node information needed and instead of being rebuild everytime Reconcile() is called, the ConfigMap is annotated with a last updated Timestamp, and will only be completely checked for outdated items every 10 Minutes.


src/controllers/nodes/nodes_controller.go
```diff
@@ -10,235 +10,198 @@ import (
...
-				Namespace: controller.namespace,
+				Namespace: r.podNamespace,
 			},
 			Data: map[string]string{},
 		}
 
-		if !controller.local { // If running locally, don't set the controller.
-			deploy, err := kubeobjects.GetDeployment(controller.client, controller.podName, controller.namespace)
+		if !r.runLocal { // If running locally, don't set the controller.
+			deploy, err := kubeobjects.GetDeployment(r.client, os.Getenv("POD_NAME"), r.podNamespace)
```
chrismuellner, line 199: `os.Getenv()` here makes unit testing weird.
better to pass it in via the `ReconcileNodes` struct

src/controllers/nodes/nodes_controller.go
```diff
@@ -10,235 +10,198 @@ import (
...
+	if err := r.updateCache(nodeCache, context.TODO()); err != nil {
+		log.Error(err, "error while updating node cache after deletion:", "node", nodeName)
+		return err
 	}
-
-	return controller.updateCache(c)
+	return nil
 }
 
-func (controller *NodesController) getCache() (*Cache, error) {
```
chrismuellner, line 216: this renamed `controller` name should be used everywhere

chrismuellner (changes requested):

src/controllers/nodes/nodes_controller.go
luhi-DT: Unit Tests will never reach that point, as runLocal is true, when unit testing, thats why os.Getenv() can be used here

src/controllers/nodes/nodes_controller.go
```diff
@@ -10,235 +10,198 @@ import (
...
-	var nodeLst corev1.NodeList
-	if err := controller.client.List(context.TODO(), &nodeLst); err != nil {
+	cachedNodeInfo, err := nodeCache.Get(nodeName)
+	if err != nil {
+		if err == ErrNotFound {
+			// uncached node -> igonoring
+			log.Error(err, "ignoring uncached node on deletion", "node", nodeName)
+			return nil
+		}
+		log.Error(err, "error while getting cachedNode on deletion")
```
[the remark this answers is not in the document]
luhi-DT: NodeDeletion is not a Reconciler, therfore returning the error will not automatically logged as in the Reconcile() function

src/controllers/nodes/nodes_controller.go
```diff
@@ -10,230 +10,192 @@ import (
...
+		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
+			controller.reconcileNodeDeletion(deleteEvent.Object.GetName())
+			return false
+		},
+	}
+}
 
-var unschedulableTaints = []string{"ToBeDeletedByClusterAutoscaler"}
+// blank assignment to verify that NodesController implements reconcile.Reconciler
+var _ reconcile.Reconciler = &NodesController{}
```
0sewa0, line 45: nitpick: not necessary here as you pass it to the manager in line 33, and if the controller didn't fulfil the `Reconciler` interface then your IDE would warn you

src/controllers/nodes/nodes_controller.go
```diff
@@ -10,230 +10,192 @@ import (
...
+
+// NewReconciler returns a new ReconcileDynaKube
+func NewReconciler(mgr manager.Manager) *NodesController {
```
0sewa0, line 48: nitpick: Maybe call it `NewController` as it returns a controller

src/controllers/nodes/nodes_controller.go
```diff
@@ -10,230 +10,192 @@ import (
...
-			if err := controller.onUpdate(node); err != nil {
-				log.Error(err, "failed to reconcile updates", "node", node)
-			}
-		case <-chAll:
-			if err := controller.reconcileAll(); err != nil {
-				log.Error(err, "failed to reconcile nodes")
-			}
+	var node corev1.Node
+	if err := controller.client.Get(ctx, client.ObjectKey{Name: nodeName}, &node); err != nil {
+		// handle deletion of Node
```
0sewa0, line 86: maybe leftover comment from the before times :D

src/controllers/nodes/oneagent_dao.go
```diff
@@ -2,34 +2,29 @@ package nodes
...
 	if err != nil {
 		return nil, err
 	}
-
-	return &dkList, nil
+	return &dynakubeList, nil
 }
 
-func (r *NodesController) filterOneAgentFromList(dkList *dynatracev1beta1.DynaKubeList,
+func (controller *NodesController) filterOneAgentFromList(dkList *dynatracev1beta1.DynaKubeList,
```
0sewa0, line 27: `filterDynaKubeFromList`
you are filtering the dynakubes (using the oneagent instances)

src/controllers/nodes/nodes_controller.go
```diff
@@ -10,230 +10,192 @@ import (
...
-		node := nodeLst.Items[i]
-		nodes[node.Name] = true
-
-		// Sometimes Azure does not cordon off nodes before deleting them since they use taints,
-		// this case is handled in the update event handler
-		if isUnschedulable(&node) {
-			if err = controller.reconcileUnschedulableNode(&node, c); err != nil {
-				return err
-			}
+	// Node is found in the cluster and in cache, send mark for termination, not found node is handled in err check
```
0sewa0, line 158: nitpick: a bit confusing comment because it refers to things before and after it,
(the "err check" happens in line 148 
the "node is found logic" starts at 159 
the comment is at 158)

0sewa0 (changes requested):

src/controllers/nodes/nodes_controller.go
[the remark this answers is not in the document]
chrismuellner: I think returning the error, and logging it in a central location is better than logging it manually.

Also I don't think logging the whole stacktrace for every error is required, should probably be checked on a case by case basis.

0sewa0 (dismissed):

src/controllers/nodes/nodes_controller.go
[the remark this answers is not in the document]
luhi-DT: true, central location is better, changed it to that approach, thanks :)

chrismuellner (approved):

0sewa0 (approved):

PR #538, opened 2022-02-03
Adds warning for missing memory limit for activeGate

0sewa0:
Not having memory limits set on activeGate deployments could cause excess memory usage, so we politely warn the customer about it, but in the end allow it none the less.


src/webhook/validation/activegate.go
```diff
@@ -54,3 +56,27 @@ func invalidActiveGateCapabilities(dv *dynakubeValidator, dynakube *dynatracev1b
...
+		if dynakube.Spec.KubernetesMonitoring.Enabled {
+			if !memoryLimitSet(dynakube.Spec.KubernetesMonitoring.Resources) {
+				return warningMissingActiveGateMemoryLimit
+			}
+		}
+		if dynakube.Spec.Routing.Enabled {
+			if !memoryLimitSet(dynakube.Spec.Routing.Resources) {
+				return warningMissingActiveGateMemoryLimit
+			}
+		}
```
chrismuellner, line 75: this check could be combined

chrismuellner (changes requested):

chrismuellner (approved):

PR #539, opened 2022-02-04
Separate statsd-ingest capability from metrics-ingest

toszr:
Extracted `statsd-ingest` capability from `metrics-ingest` and removed feature flag to control `statsd-ingest`.

### DIFF
--- a/src/api/v1beta1/activegate_types.go
+++ b/src/api/v1beta1/activegate_types.go
@@ -22,7 +22,7 @@ var (
 	RoutingCapability = ActiveGateCapability{
 		DisplayName:  "routing",
 		ShortName:    "routing",
-		ArgumentName: "MSGrouter,extension_controller",
+		ArgumentName: "MSGrouter",
 	}
 
 	KubeMonCapability = ActiveGateCapability{
@@ -42,6 +42,12 @@ var (
 		ShortName:    "dynatrace-api",
 		ArgumentName: "restInterface",
 	}
+
+	StatsDIngestCapability = ActiveGateCapability{
+		DisplayName:  "statsd-ingest",
+		ShortName:    "statsd-ingest",
+		ArgumentName: "extension_controller",
+	}
 )
 
 var ActiveGateDisplayNames = map[CapabilityDisplayName]struct{}{
@@ -49,6 +55,7 @@ var ActiveGateDisplayNames = map[CapabilityDisplayName]struct{}{
 	KubeMonCapability.DisplayName:       {},
 	MetricsIngestCapability.DisplayName: {},
 	DynatraceApiCapability.DisplayName:  {},
+	StatsDIngestCapability.DisplayName:  {},
 }
 
 type ActiveGateSpec struct {
--- a/src/api/v1beta1/activegate_types.go
+++ b/src/api/v1beta1/activegate_types.go
@@ -43,7 +43,7 @@ var (
 		ArgumentName: "restInterface",
 	}
 
-	StatsDIngestCapability = ActiveGateCapability{
+	StatsdIngestCapability = ActiveGateCapability{
 		DisplayName:  "statsd-ingest",
 		ShortName:    "statsd-ingest",
 		ArgumentName: "extension_controller",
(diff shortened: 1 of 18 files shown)


src/api/v1beta1/activegate_types.go
```diff
@@ -42,13 +42,20 @@ var (
...
+		ArgumentName: "extension_controller",
+	}
 )
 
 var ActiveGateDisplayNames = map[CapabilityDisplayName]struct{}{
 	RoutingCapability.DisplayName:       {},
 	KubeMonCapability.DisplayName:       {},
 	MetricsIngestCapability.DisplayName: {},
 	DynatraceApiCapability.DisplayName:  {},
+	//StatsDIngestCapability.DisplayName: {}, // TODO Enable once EEC/StatsD images are pullable from the cluster
```
chrismuellner, line 58: this should be uncommented so it can be enabled

src/api/v1beta1/properties.go
```diff
@@ -295,3 +295,15 @@ func splitArg(arg string) (key, value string) {
...
+func (dk *DynaKube) NeedsStatsD() bool {
+	if dk.FeatureEnableStatsDIngest() {
+		return true
+	}
+	for _, capability := range dk.Spec.ActiveGate.Capabilities {
+		if capability == StatsDIngestCapability.DisplayName {
+			return true
+		}
+	}
+	return false
```
chrismuellner, line 308: Checking both the feature flag and this capability feels redundant.
Remove the feature flag, check just for the capability instead. Marking this capability as preview similar to how it's handled for `metric-ingest` in `validation/preview.go` would be an option on how we could make it clearer that it shouldn't be used.

We have `IsActiveGateMode` for this check already, `NeedsStatsD` should be structured similarly to `KubernetesMonitoringMode`.

chrismuellner (changes requested):

src/api/v1beta1/activegate_types.go
toszr: Actually, it's commented out in a separate commit on purpose. I'd rather enable it no sooner than we have a confirmation EEC/StatsD images are integrated with Dynatrace Docker registry which won't happen until sprint 237+.

chrismuellner: This is enabled already via feature flag, hiding the capability makes no difference

toszr: It actually _does_ make a difference because feature flags are only discoverable through code inspection whereas CR properties are discoverable with `kubectl explain dynakube.spec.activeGate`.

Either way, to keep things simple I'll just remove the flag and enable capability by default to avoid confusion and will worry when and if it's needed again.

src/api/v1beta1/properties.go
toszr: Great, good to know. There was no `IsActiveGateMode()` at the time I wrote this but it's exactly what I need.

Ad feature flag vs capability: I'll go with the warning instead as it will be easier to use and test.

src/api/v1beta1/activegate_types.go
```diff
@@ -42,13 +42,20 @@ var (
 		ShortName:    "dynatrace-api",
 		ArgumentName: "restInterface",
 	}
+
+	StatsDIngestCapability = ActiveGateCapability{
```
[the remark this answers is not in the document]
toszr: I don't have strong feelings about either option, I can make it a `Statsd`. 

BTW Same story for Extension Controller - `EEC` or `Eec`? I thought about following camel case rules more strictly so maybe it's about time to do it.

0sewa0 (changes requested):
why do we have a feature-flag and a capability that enables the same thing ? or did I miss something ?
If not then I would just drop the feature-flag, as its redundant.

toszr:
> why do we have a feature-flag and a capability that enables the same thing ? or did I miss something ? If not then I would just drop the feature-flag, as its redundant.

It was requested by @chrismuellner earlier - I dropped the redundant feature flag as it is no longer needed, please see the new commit from Friday.

There is still a second feature flag `...UseActiveGateImageForStatsd` which is for different purpose and is still relevant (enables single-mode operation with EEC and StatsD built into ActiveGate image, but with custom entry points). It was for a proof of concept last year that was eventually decided against.

0sewa0 (approved):

PR #544, opened 2022-02-07
Raw ActiveGate image support

mjgrzybek:
Operator fetches `tenant uuid`, `tenant token` and `connection endpoints` for given `apiUrl` and applies it to AG via environment variables and mounted secrets.
Operator supports "configured" images as well as "raw" images. Container created using configured image is overridden with parameters as if it was raw.
Added reconciler for parameters mentioned above. Dybakube's `activegate.Reconciler` maintains `<dynakube-name>-activegate-tenant-secret` secret.

Known limitations:
- `dtclient` used by operator is not networkzone aware, so `connection endpoints` may be invalid. It's not a regression, though.


src/controllers/activegate/reconciler/statefulset/statefulset_test.go
```diff
@@ -158,9 +161,9 @@ func TestStatefulSet_Volumes(t *testing.T) {
 		volumes := buildVolumes(stsProperties, getContainerBuilders(stsProperties))
 		expectedSecretName := instance.Name + "-router-" + customproperties.Suffix
 
-		require.NotEmpty(t, volumes)
+		require.Equal(t, 2, len(volumes))
 
-		customPropertiesVolume := volumes[0]
+		customPropertiesVolume := volumes[1]
```
toszr, line 166: Please don't access slice elements by index when you mean something specific, roll out a helper function like `getVolumeByName()` or so (similar to `getContainerByName()`) or use a map.

src/controllers/activegate/reconciler/statefulset/statefulset_test.go
```diff
@@ -196,20 +199,43 @@ func TestStatefulSet_Volumes(t *testing.T) {
...
 			testUID, "", testFeature, "MSGrouter", "",
 			nil, nil, nil,
 		))
-		assert.Equal(t, []corev1.EnvVar{
-			{Name: DTCapabilities, Value: "MSGrouter"},
-			{Name: DTIdSeedNamespace, Value: instance.Namespace},
-			{Name: DTIdSeedClusterId, Value: testUID},
-			{Name: DTDeploymentMetadata, Value: deploymentMetadata.AsString()},
+		assert.Equal(t, 7, len(envVars))
+		assert.Equal(t, envVars, []corev1.EnvVar{
```
toszr, line 210: Please use `assert.Contains()` instead. Note that env variable sequence in this slice doesn't matter so you shouldn't assert that.

src/controllers/dynakube/activegate/generate.go
```diff
@@ -0,0 +1,23 @@
...
+
+import "github.com/pkg/errors"
+
+const (
+	CommunicationEndpointsName = "communication-endpoints"
+	TenantTokenName            = "tenant-token"
+	TenantUuidName             = "tenant-uuid"
+)
+
+func (r *Reconciler) GenerateData() (map[string][]byte, error) {
```
toszr, line 11: Most methods and functions in any programming language generate data. Can you rename it to reflect its purpose more precisely?

src/controllers/dynakube/activegate/reconciler.go
```diff
@@ -0,0 +1,126 @@
...
+func (r *Reconciler) updateAGSecret(agSecret *corev1.Secret, desiredAGSecretData map[string][]byte) error {
+	log.Info("updating secret", "name", agSecret.Name)
+	agSecret.Data = desiredAGSecretData
+	if err := r.Update(context.TODO(), agSecret); err != nil {
+		return fmt.Errorf("failed to update secret %s: %w", agSecret.Name, err)
+	}
+	return nil
+}
+
+func isSecretEqual(currentSecret *corev1.Secret, desired map[string][]byte) bool {
```
toszr, line 109: I wonder if you could move it to `kubeobjects/secret.go`?

src/dtclient/agent_tenant_info.go
```diff
@@ -7,16 +7,16 @@ import (
 	"github.com/pkg/errors"
 )
 
-type TenantInfo struct {
-	ID                    string
+type AgentTenantInfo struct {
```
toszr, line 10: Maybe a common base with `ActiveGateTenantInfo`, or include one in the other?

BTW `Endpoints` use different types in both cases (`string` vs `[]string`), is it necessary?

```suggestion
type TenantInfo struct {
	UUID                  string
	Token                 string
}

type AgentTenantInfo struct {
	TenantInfo
	Endpoints             []string
	CommunicationEndpoint string
}
```

src/dtclient/activegate_tenant_info.go
```diff
@@ -0,0 +1,58 @@
...
+	tenantInfo, err := dtc.readResponseForActiveGateTenantInfo(data)
+	if err != nil {
+		log.Error(err, err.Error())
+		return nil, err
+	}
+	if len(tenantInfo.Endpoints) == 0 {
+		log.Info("tenant has no endpoints")
+	}
+
+	return tenantInfo, nil
```
toszr, line 46: I'm missing `tenantInfo.UUID` validation here, what if it's empty or whitespace-only?

src/dtclient/common_tenant_info_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+func tenantMalformedJson(url string) http.HandlerFunc {
+	return func(writer http.ResponseWriter, request *http.Request) {
+		if request.URL.Path == url {
+			_, _ = writer.Write([]byte("this is not json"))
```
toszr, line 11: What's the point of this syntax?

```suggestion
			writer.Write([]byte("this is not json"))
```

src/dtclient/agent_tenant_info.go
mjgrzybek: Yes, it's necessary because it reflects how REST interfaces are implemented.

src/dtclient/common_tenant_info_test.go
0sewa0: I think this is needed by the linter. 
It makes it explicit that you don't care about the error, which is preferable to just not caring at all

src/controllers/activegate/reconciler/statefulset/statefulset_test.go
mjgrzybek: Actually `assert.ElementsMatch` seems even better, switched to it.

src/dtclient/activegate_tenant_info.go
mjgrzybek: `dtclient` should be refactored - for example generated with swagger. I don't want to manually provide sophisticated logic here because it requires maintenance.
If `tenantInfo.UUID` is empty, or whitespace-only, we can do nothing except printing warning. I see no value to user here because pod creation is failing anyway with obvious reason.

aorcholski (dismissed):
Make sure that for the entrypoint.sh script settings read from Volumes have higher priority than read from entrypoint.properties file.

src/controllers/activegate/reconciler/statefulset/statefulset.go
```diff
@@ -272,6 +288,14 @@ func buildVolumeMounts(stsProperties *statefulSetProperties) []corev1.VolumeMoun
 		volumeMounts = append(volumeMounts, buildProxyMounts()...)
 	}
 
+	volumeMounts = append(volumeMounts, corev1.VolumeMount{
+		Name:      tenantSecretVolumeName,
+		ReadOnly:  true,
+		MountPath: "/var/lib/dynatrace/secrets/tokens/tenant-token",
```
0sewa0, line 294: maybe a `const` for this ?

src/api/v1beta1/activegate_types.go
```diff
@@ -74,6 +74,9 @@ type ActiveGateSpec struct {
 	// Optional: Sets DNS Policy for the ActiveGate pods
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="DNS Policy",order=24,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:text"}
 	DNSPolicy corev1.DNSPolicy `json:"dnsPolicy,omitempty"`
+
+	// Optional: Sets name of secret containing: tenant-uuid, tenant-token, communication-endpoints
+	AGTenantSecret string `json:"tenantSecret,omitempty"`
```
0sewa0: `TenantSecret` should be enough, saying `ActiveGate.AGTenantSecret` feels very redundant

src/controllers/dynakube/activegate/reconciler.go
```diff
@@ -0,0 +1,139 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	AGSecretSuffix             = "-activegate-tenant-secret"
+	CommunicationEndpointsName = "communication-endpoints"
+	TenantTokenName            = "tenant-token"
+	TenantUuidName             = "tenant-uuid"
```
0sewa0, line 23:
```suggestion
	TenantUUIDName             = "tenant-uuid"
```

src/controllers/dynakube/activegate/reconciler.go
```diff
@@ -0,0 +1,139 @@
...
+	return map[string][]byte{
+		TenantUuidName:             []byte(tenantInfo.UUID),
+		TenantTokenName:            []byte(tenantInfo.Token),
+		CommunicationEndpointsName: []byte(tenantInfo.Endpoints),
+	}, nil
+}
+
+func (r *Reconciler) createAGSecretIfNotExists(agSecretData map[string][]byte) (*corev1.Secret, error) {
+	var config corev1.Secret
+	err := r.apiReader.Get(context.TODO(), client.ObjectKey{Name: extendWithAGSecretSuffix(r.instance.Name), Namespace: r.instance.Namespace}, &config)
```
0sewa0, line 88: nitpick: this line could use a few newlines or something to make it shorter :D

src/api/v1beta1/activegate_types.go
0sewa0: also isn't this secret generated by us, why does it have to be in the dynakube ?

src/controllers/dynakube/activegate/reconciler.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+const (
+	AGSecretSuffix             = "-activegate-tenant-secret"
+	CommunicationEndpointsName = "communication-endpoints"
+	TenantTokenName            = "tenant-token"
+	TenantUuidName             = "tenant-uuid"
+)
+
+type Reconciler struct {
```
0sewa0, line 26: The name of `Reconciler` is to general here, this only makes sure that the TenantSecret is present and up to date.

`TenantSecretReconciler` or something would make it clearer.

Also putting `AG` into all function names feels a bit unnecessary

src/controllers/dynakube/activegate/reconciler.go
mjgrzybek: It's less readable imho. 
No two capitalised letters shall follow directly each other in cammelCase.

src/controllers/dynakube/activegate/reconciler.go
toszr: Tip: I merged recently something to achieve that - `kubeobjects.Key()`
https://github.com/Dynatrace/dynatrace-operator/blob/master/src/kubeobjects/object.go#L5

src/controllers/dynakube/activegate/reconciler.go
0sewa0: uh, spicy 🌶️  :D 
I would disagree 😏

toszr: I think we need a guideline note about preferred coding style with regard to camel casing. I remember refactoring a lot of code and APIs to follow `StatsdSomething` instead of `StatsDSomething`. 

You're now suggesting the opposite 😛

0sewa0: Very true 😅 

I just saw a lot of `SomethingID` and `SomethingUUID` in our code so I just wanted to be consistent (that said there is also `SomethingId` and `SomethingUiid` in our code 😛 )

mjgrzybek: @toszr why the opposite? I stick exactly to it by doing `UuidName` instead of `UUIDName` :open_mouth:.
Naming fight, start! :smile:

toszr: Hmm, my comment was to the Marcell's suggestion but apparently I was too slow to post it. So, to clarify:
- I was advised a while ago by @mjgrzybek to switch convention from `StatsDSomething` to `StatsdSomething`
- @0sewa0 is now suggesting to switch from `TenantUuidName` to `TenantUUIDName`

That is why I suggested we have a guideline to simplify reviews.

src/controllers/dynakube/activegate/reconciler.go
mjgrzybek: @toszr I just found out it's already implemented in `controller-runtime` as:
```
package client

// ObjectKeyFromObject returns the ObjectKey given a runtime.Object.
func ObjectKeyFromObject(obj Object) ObjectKey {
	return ObjectKey{Namespace: obj.GetNamespace(), Name: obj.GetName()}
}
```

toszr: Hmm. I like `Key()` more than `ObjectKeyFromObject()` but thanks for finding it - I knew someone must have thought of it before.

mjgrzybek: It can't be used here because name is calculated on the fly so it must be set explicitly.
Added newlines to make it more readable.

src/controllers/dynakube/activegate/reconciler.go
```diff
@@ -0,0 +1,139 @@
...
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	AGSecretSuffix             = "-activegate-tenant-secret"
```
0sewa0, line 20: this is already defined in `src/api/v1beta1/properties.go`

0sewa0: pls define it in 1 place,  also `TenantSecretSuffix` maybe a better name  :)

0sewa0 (approved):

PR #546, opened 2022-02-08
Exposes custom StatsD and EEC images in DynaKube ActiveGate spec

toszr:
ActiveGate spec in DynaKube now offers a way to customize EEC & StatsD Docker image. This is useful for customers who maintain private registries with Dynatrace images (and necessary for the time being until EEC & StatsD images are integrated with Dynatrace Docker registry).

```yaml
# ...
spec:
  activeGate:
    eecImage: registry.lab.dynatrace.org/eec/dynatrace-eec:1.235.0.20220124-150920
    statsdImage: registry.lab.dynatrace.org/datasource/dynatrace-datasource-statsd:0.1.20220113114511
# ...
```


src/api/v1beta1/dynakube_status.go
```diff
@@ -79,10 +96,50 @@ type VersionStatus struct {
 	LastUpdateProbeTimestamp *metav1.Time `json:"lastUpdateProbeTimestamp,omitempty"`
 }
 
+func (verStatus *VersionStatus) GetImageHash() string {
+	return verStatus.ImageHash
```
mjgrzybek, line 100: Should `verStatus.ImageHash` be exported? If yes, getter is not necessary, if no, please change it to lower case.

chrismuellner (changes requested):
These images should be set by feature flag/annotation and not in the CR

src/api/v1beta1/dynakube_status.go
toszr: It must be exported unless you want me to add a setter as well:
image

Alternatively, I can get rid of `.GetImageHash()` etc. and just access the exported fields - I don't have a strong preference here.

toszr:
> These images should be set by feature flag/annotation and not in the CR

Why? They are customer-facing, same as the ActiveGate `image` property in the DynaKube CR. As I learned, some customers prefer to perform manual version upgrades and the same is true for image versions.

chrismuellner:
These images are not yet available on the registry, so having the configurable allows us to test and use them without having to change the whole CR.

src/api/v1beta1/dynakube_status.go
mjgrzybek: For sake of consistency choose either of these: getters+setters or Exported field.
I think the latter is preferred in go world, less boilerplate.

chrismuellner: I'd prefer having getters without get in the name: [EffectiveGo#Getters](https://go.dev/doc/effective_go#Getters)

toszr: Actually, the fields of `VersionStatus` must be exported because so that they can be (un)marshalled from/to JSON or YAML.

I simplified the interfaces a bit and renamed the remaining getters as suggested, thanks.

toszr:
OK, it's a small change I can fit into the first release. Be it annotations then (`custom-eec-image`, `custom-statsd-image`)

src/api/v1beta1/properties.go
```diff
@@ -137,6 +137,40 @@ func (dk *DynaKube) ActiveGateImage() string {
 	return fmt.Sprintf("%s/linux/activegate:latest", registry)
 }
 
+// EecImage returns the Extension Controller image to be used with the dk DynaKube instance.
+func (dk *DynaKube) EecImage() string {
+	if dk.ActiveGateMode() {
```
chrismuellner, line 142: shouldn't this check if the statsd capability is used?

src/controllers/activegate/reconciler/statefulset/container_statsd_test.go
```diff
@@ -31,21 +31,21 @@ func TestStatsd_BuildContainerAndVolumes(t *testing.T) {
 		for _, port := range []int32{
 			consts.StatsdIngestPort, statsdProbesPort,
 		} {
-			assertion.Truef(kubeobjects.PortIsIn(container.Ports, port), "Expected that Statsd container defines port %d", port)
+			assertion.Truef(kubeobjects.PortIsIn(container.Ports, port), "Expected that StatsD container defines port %d", port)
```
chrismuellner, line 34: consistency: `Statsd` or `StatsD`?

src/api/v1beta1/properties.go
```diff
@@ -137,6 +137,40 @@ func (dk *DynaKube) ActiveGateImage() string {
...
+	if dk.Spec.APIURL == "" {
+		return ""
+	}
+
+	registry := buildImageRegistry(dk.Spec.APIURL)
+	return fmt.Sprintf("%s/linux/dynatrace-eec:latest", registry)
+}
+
+// StatsdImage returns the StatsD data source image to be used with the dk DynaKube instance.
+func (dk *DynaKube) StatsdImage() string {
```
chrismuellner, line 158: these methods are pretty similar, logic could be reused

src/controllers/dynakube/updates/version.go
```diff
@@ -36,7 +36,15 @@ func ReconcileVersions(
 		!dk.FeatureDisableActiveGateUpdates() &&
 		dkState.IsOutdated(dk.Status.ActiveGate.LastUpdateProbeTimestamp, ProbeThreshold)
 
-	if !needsActiveGateUpdate && !needsOneAgentUpdate {
+	needsEecUpdate := dk.NeedsActiveGate() &&
```
chrismuellner, line 39: same here, these images shouldn't be updated when they are not required

chrismuellner (changes requested):

toszr:
> PR Description should be updated to reflect move to annotations

Good catch, thanks, done.

src/api/v1beta1/properties.go
toszr: It was needed only to make sure there is an `ActiveGate` spec to look up the image in. I replaced it with `.NeedsStatsd()` anyway for consistency.

src/api/v1beta1/properties.go
toszr: Looked like a bit of an overkill to me, but I don't have a strong opinion here so refactored it to get rid of the redundancy.

src/controllers/dynakube/updates/version.go
toszr: Makes sense

0sewa0 (approved):

PR #549, opened 2022-02-08
Adds support for single token in troubleshoot script

0sewa0:


troubleshoot/troubleshoot.sh
```diff
@@ -151,22 +151,27 @@ function checkDynakube {
...
+    error "token apiToken does not exist in secret '${secret_name}'"
+  else
+    log "secret token apiToken exists"
+  fi
 
+  paas_token=$("${cli}" get secret "$secret_name" \
+  --namespace "${selected_namespace}" \
+  --template="{{.data.paasToken}}")
+  if [[ "$paas_token" == "" || "$paas_token" == "$missing_value" ]]; then
+    log "secret token paasToken doesn't exists using apiToken instead"
```
chrismuellner, line 168:
```suggestion
    log "secret token paasToken doesn't exist; using apiToken instead"
```

chrismuellner (changes requested):

PR #552, opened 2022-02-09
Removes init.sh, used init command instead

0sewa0:
The `initgeneration` package no longer will put the init.sh in to the init secret, instead it creates the json needed for the `init` command and puts that in the secret.

The pod mutation webhook now injects a initContainer that uses the `init` command instead of the init.sh


src/initgeneration/initgeneration.go
```diff
@@ -174,6 +163,10 @@ func getPaasToken(tokens corev1.Secret) []byte {
 	return tokens.Data[dtclient.DynatraceApiToken]
 }
 
+func getAPIToken(tokens corev1.Secret) []byte {
```
chrismuellner, line 166: could this directly return a `string` if it's only used in that way?

src/initgeneration/initgeneration.go
```diff
@@ -174,6 +163,10 @@ func getPaasToken(tokens corev1.Secret) []byte {
...
+	return tokens.Data[dtclient.DynatraceApiToken]
+}
+
 // getInfraMonitoringNodes creates a mapping between all the nodes and the tenantUID for the infra-monitoring dynakube on that node.
```
chrismuellner, line 170: infra monitoring was renamed to host monitoring

src/standalone/run.go
```diff
@@ -11,10 +11,10 @@ import (
 )
 
 const (
-	noHostTenant = "-"
+	NoHostTenant = "-"
```
chrismuellner, line 14: I think the constants in this package should be moved to `config.go`. As it is now that file only contains the logger and having the constants in a central location would be nice.

src/webhook/mutation/pod_mutator_test.go
```diff
@@ -1351,7 +1351,7 @@ func buildResultPod(_ *testing.T, oneAgentFf FeatureFlag, dataIngestFf FeatureFl
 		)
 
 		pod.Spec.Containers[0].VolumeMounts = append(pod.Spec.Containers[0].VolumeMounts,
-			corev1.VolumeMount{Name: dataIngestVolumeName, MountPath: "/var/lib/dynatrace/enrichment"},
+			corev1.VolumeMount{Name: dataIngestVolumeName, MountPath: standalone.EnrichmentPath},
 			corev1.VolumeMount{Name: dataIngestEndpointVolumeName, MountPath: "/var/lib/dynatrace/enrichment/endpoint"},
```
chrismuellner, line 1355: should this also get a constant or use `standalone.EnrichmentPath` as a base?

chrismuellner (changes requested):

src/webhook/mutation/pod_mutator_test.go
0sewa0: it should but, I didn't want to introduce more changes to stuff that are not affected by my changes :D

chrismuellner (approved):

PR #553, opened 2022-02-09
Send Mark for termination event also, when node is not found in the cluster anymore

luhi-DT:
Apparently we send the mark for termination event, also for nodes which we cached, but then could not be found in the cluster after sometime, therefor, we need to check and send the event, when the cache is periodically checked for outdated nodes aswell.

### DIFF
--- a/src/controllers/nodes/nodes_controller.go
+++ b/src/controllers/nodes/nodes_controller.go
@@ -227,6 +227,7 @@ func (controller *NodesController) handleOutdatedCache(nodeCache *Cache) error {
 	}
 
 	for _, cachedNodeName := range nodeCache.Keys() {
+		var cachedNodeInCluster = false
 		for _, clusterNode := range nodeLst.Items {
 			if clusterNode.Name == cachedNodeName {
 				cachedNodeInfo, err := nodeCache.Get(cachedNodeName)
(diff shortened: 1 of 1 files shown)


src/controllers/nodes/nodes_controller.go
```diff
@@ -227,17 +227,26 @@ func (controller *NodesController) handleOutdatedCache(nodeCache *Cache) error {
...
 			if clusterNode.Name == cachedNodeName {
 				cachedNodeInfo, err := nodeCache.Get(cachedNodeName)
 				if err != nil {
 					log.Error(err, "failed to get node", "node", cachedNodeName)
 					return err
 				}
+				cachedNodeInCluster = true
+				// Check if node was seen less than an hour ago, otherwise do not remove from cache
 				controller.removeNodeFromCache(nodeCache, cachedNodeInfo, cachedNodeName)
 				break
```
chrismuellner, line 241: could you return out of the function here and always remove the node from the cache if the loop is fully traversed?
would remove the need for the variable

chrismuellner (changes requested):

src/controllers/nodes/nodes_controller.go
chrismuellner: guess not, need to go through all the nodes 😄

luhi-DT: yes, otherwise first node which is found ends the function

chrismuellner (approved):

PR #555, opened 2022-02-10
Host csi-volume for ReadOnly OneAgent pods

0sewa0:
(+ small csi package reorg 😬 )

1. Adds Host volume publisher
	- Mounts a host path for the host monitoring oneagents in case of readonly mode (enabled via feature-flag: `alpha.operator.dynatrace.com/feature-oneagent-readonly-host-fs: "true"`)
	- Each mount is tenant specific. [so in theory we could have 2(or more) readonly host-monitoring agent on the same node, with a different tenant, at the same time]
	- Uses a new database table to keep track of the previously mounted oneagents host volumes: (can be used for gc in the future)
	```
	CREATE TABLE IF NOT EXISTS storages (
		TenantUUID VARCHAR NOT NULL,
		VolumeID VARCHAR NOT NULL,
		Mounted BOOLEAN NOT NULL,
		LastModified DATETIME NOT NULL,
		PRIMARY KEY (TenantUUID)
	);
    ```
2. Moves `BindConfig` out to the `csivolumes` package so it can be used in the app and host volume packages
   - The dynakube name is no longer acquired using the instance label on the pod's namespace, but using volume attributes. (removes 1 api call, yay)
   - So 2 volume attribute is set by the webhook and daemonset, the name of the dynakube and the name of the volume mode (app/host) used
3. Sets unprivileged user/group for the oneagent daemonset security context in case of readonly mode (root user needed otherwise)

Some weird stuff:
During unpublish we don't really know which mode(app/host) the volume was for, so we try to unmount it both ways, if 1 succeeds we don't try the other if both fail we log a message a move on


src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -306,4 +306,11 @@ func unprivilegedSecurityContext() *corev1.SecurityContext {
 			},
 		},
 	}
+	if dsInfo.instance.FeatureReadOnlyOneAgent() {
+		unprivilegedUser := int64(1001)
+		unprivilegedGroup := int64(2001)
```
chrismuellner, line 311:
```suggestion
		unprivilegedUser := int64(1000)
		unprivilegedGroup := int64(1000)
```

src/controllers/csi/metadata/sqlite.go
```diff
@@ -205,6 +243,46 @@ func (a *SqliteAccess) DeleteVolume(volumeID string) error {
 	return err
 }
 
+// InsertVolume inserts a new Volume
```
chrismuellner, line 246: may need to be adapted to fit method

src/controllers/csi/metadata/sqlite.go
```diff
@@ -205,6 +243,46 @@ func (a *SqliteAccess) DeleteVolume(volumeID string) error {
...
+			storage.VolumeID,
+			storage.TenantUUID,
+			storage.Mounted,
+			storage.LastModified,
+			err)
+	}
+	return err
+}
+
+// UpdateDynakube updates an existing Dynakube by matching the name
```
chrismuellner, line 260: same here

src/controllers/csi/driver/volumes/app/publisher.go
```diff
@@ -167,9 +171,5 @@ func (publisher *AppVolumePublisher) loadVolume(volumeID string) (*metadata.Volu
 	if err != nil {
 		return nil, err
 	}
-	if volume == nil {
-		return &metadata.Volume{}, nil
-	}
-	log.Info("loaded volume info", "id", volume.VolumeID, "pod name", volume.PodName, "version", volume.Version, "dynakube", volume.TenantUUID)
```
chrismuellner, line 173: is this log no longer necessary?

src/controllers/csi/driver/server.go
```diff
@@ -144,32 +146,38 @@ func (svr *CSIDriverServer) NodePublishVolume(ctx context.Context, req *csi.Node
...
 
 func (svr *CSIDriverServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublishVolumeRequest) (*csi.NodeUnpublishVolumeResponse, error) {
 	volumeInfo, err := csivolumes.ParseNodeUnpublishVolumeRequest(req)
 	if err != nil {
 		return nil, err
 	}
-	response, err := svr.publishers[appvolumes.Mode].UnpublishVolume(ctx, volumeInfo)
-	if err != nil {
-		return nil, err
+	for _, publisher := range svr.publishers {
```
chrismuellner, line 170: is trying to unpublish the volume both ways the best idea here?
we should know how it was mounted by checking the database

src/controllers/csi/driver/volumes/host/publisher_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	dtcsi "github.com/Dynatrace/dynatrace-operator/src/controllers/csi"
+	csivolumes "github.com/Dynatrace/dynatrace-operator/src/controllers/csi/driver/volumes"
+	"github.com/Dynatrace/dynatrace-operator/src/controllers/csi/metadata"
+	"github.com/Dynatrace/dynatrace-operator/src/scheme/fake"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
```
chrismuellner, line 16:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

src/controllers/csi/driver/volumes/volume_config.go
```diff
@@ -9,19 +9,25 @@ import (
...
+	VolumeID   string
 	TargetPath string
 }
 
 // Represents the config needed to mount a volume
 type VolumeConfig struct {
 	VolumeInfo
-	Namespace string
-	PodName   string
+	Namespace    string // not really necessary anymore
```
chrismuellner, line 27: do we need it then?

chrismuellner (changes requested):

src/controllers/csi/driver/server.go
0sewa0: that is what actually happen, the unpublish checks if it has info about it in the database and if not it does nothing
(that is why https://github.com/Dynatrace/dynatrace-operator/pull/555#discussion_r805811276 happened)

src/controllers/dynakube/oneagent/daemonset/daemonset.go
0sewa0: groups should start with 2

src/controllers/dynakube/oneagent/daemonset/volumes.go
```diff
@@ -98,14 +100,15 @@ func getCertificateVolume(instance *dynatracev1beta1.DynaKube) corev1.Volume {
 	}
 }
 
-func getCSIStorageVolume() corev1.Volume {
+func getCSIStorageVolume(instance *dynatracev1beta1.DynaKube) corev1.Volume {
```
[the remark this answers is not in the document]
0sewa0: everything in that file uses it that way, so to keep it consistent I would leave it

src/controllers/csi/driver/server.go
chrismuellner: I think this shouldn't be hidden behind the `Unpublish` method
We should extract that check so it's more explicit, either as its own method on each publisher or as a check that's done before the actual unpublish

0sewa0: >  either as its own method on each publisher

I like that

src/controllers/dynakube/oneagent/daemonset/volumes.go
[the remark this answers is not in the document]
chrismuellner: understandable

src/controllers/csi/driver/server.go
```diff
@@ -144,32 +146,38 @@ func (svr *CSIDriverServer) NodePublishVolume(ctx context.Context, req *csi.Node
...
+		return nil, status.Error(codes.Internal, fmt.Sprintf("unknown csi mode provided, mode=%s", volumeCfg.Mode))
 	}
-	return &csi.NodePublishVolumeResponse{}, nil
+	log.Info("publishing volume",
+		"csiMode", volumeCfg.Mode,
+		"target", volumeCfg.TargetPath,
+		"fstype", req.GetVolumeCapability().GetMount().GetFsType(),
+		"readonly", req.GetReadonly(),
+		"volumeID", volumeCfg.VolumeId,
+		"attributes", req.GetVolumeContext(),
```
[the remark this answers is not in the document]
0sewa0: Not really, if I remember correctly it shows up in the log correctly, to make it pretty we would need to unpack it in a safe way, which would probably would be more work than it worth. (or use `Sprintf(...)` 😬 )

chrismuellner (approved):

PR #561, opened 2022-02-14
Adds Extension Controller/StatsD logs to the ActiveGate support archive

toszr:
ActiveGate offers the option to collect a support archive remotely from Debug UI (or the user-facing UI). It is useful when there is no log monitoring enabled in the cluster.

This PR links relevant directories between ActiveGate, Extension Controller and StatsD data source containers using volume mounts so that ActiveGate automatically picks up EEC/StatsD logs when a support archive is generated.

### DIFF
--- a/src/controllers/activegate/reconciler/statefulset/container_eec.go
+++ b/src/controllers/activegate/reconciler/statefulset/container_eec.go
@@ -11,6 +11,7 @@ import (
 
 const eecIngestPortName = "eec-http"
 const eecIngestPort = 14599
+const extensionsLogsDir = "/var/lib/dynatrace/remotepluginmodule/log/extensions"
 
 const activeGateInternalCommunicationPort = 9999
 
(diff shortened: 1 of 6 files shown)


0sewa0:
Just so I understand correctly:
We mount 2 new emptyDir volumes, 1 for stasd logs and 1 for eec logs.
the stasd container can write to its log volume and can read from the eec logs volume
the eec container can write to its log volume and can read from the stasd logs volume
the activegate container can only read from both.

why does the stasd container need to be able to read the logs of the eec and visa-versa ?

toszr:
Close, but it's not exactly what I implemented.

- ActiveGate needs `ro` access to both EEC and StatsD logs (because it must collect them).
- EEC needs access to its own logs volume (`rw`) but in order to maintain bare metal deployment consistency I made it aware (`ro`) of data source's logs as well. (It's for convenience as well).

StatsD only writes its own logs which it must expose to EEC and ActiveGate (does not need access to EEC logs, so only 1 extra volume is mounted).

In theory, I could get rid of the StatsD logs volume mount in the EEC container and it shouldn't affect the deployment. In practice I preferred to keep it consistent with plain OneAgent/ActiveGate deployment modes.

0sewa0 (approved):

PR #565, opened 2022-02-15
Removes unnecessary sa for routing

0sewa0:
This service account is no longer needed.


luhi-DT:
In the make file, dynatrace-routing is still added as an extra service account, when generating the CSV. I guess it can be removed from there aswell :)

luhi-DT (approved):

PR #569, opened 2022-02-15
Prevent creation of duplicate settings for automatic API monitoring

fabwer:
- Adds the functionality to generate a monitored entity ID based on a hash of the kube-system uid

This prevents the creation of multiple settings objects, which occur when POSTing to the Dynatrace settings objects API without a scope, but having the same body otherwise.

Up for discussion:
- This solution, or
- Using dkState with an additional field (like KubernetesMEID), to check if the field is already populated after a successful settings object creation, to then not try to create another settings object.


src/dtclient/kubernetes_settings.go
```diff
@@ -257,3 +265,15 @@ func handleErrorArrayResponseFromAPI(response []byte, statusCode int) error {
 		return fmt.Errorf(sb.String())
 	}
 }
+
+func generateKubernetesMEIdentifier(kubeSystemUUID string) string {
```
baichinger, line 269: Would rather return an extra `error` instead of the implicit empty string error handling logic.

0sewa0 (approved):

PR #580, opened 2022-02-16
Adds security context to the operator/webhook

0sewa0:
# Description
Adds the minimum security context to the operator/webhook deployments

## How can this be tested?
If you deploy it on any cluster, the operator/webhook pods should not be `Blocked`. (stuff should just work)

I tested it on:
- Azure
- GCloud docker
- GCloud containerd
- Openshift 4.8

## Checklist
- [x] Labeled the PR accordingly


config/helm/chart/default/tests/Common/webhook/deployment-webhook_test.yaml
```diff
@@ -103,3 +120,111 @@ tests:
...
+      - equal:
+          path: metadata
+          value:
+            name: dynatrace-webhook
+            namespace: NAMESPACE
+            labels:
+              app.kubernetes.io/instance: RELEASE-NAME
+              app.kubernetes.io/name: RELEASE-NAME
+              dynatrace.com/operator: dynakube
+              helm.sh/chart: dynatrace-operator-0.4.0
```
chrismuellner, line 141: I removed the check for a specific version from the other test case, since it's causing trouble when updating the version. See #577

chrismuellner (changes requested):

chrismuellner (approved):

PR #582, opened 2022-02-16
Maps EEC/StatsD-related annotations to the EEC and data sources runtime properties

toszr:
Extension Controller and data sources (like StatsD) can be configured in OneAgent/ActiveGate bare metal deployments using runtime properties (a.k.a. debug flags). In a bare metal AG setup, the interface is a json file stored at `/var/lib/dynatrace/remotepluginmodule/agent/conf/runtime/runtimeConfiguration`.

This changeset converts EEC/StatsD debug flags defined as `DynaKube` annotations with prefix `alpha.operator.dynatrace.com/extensions.` into valid runtime properties mounted into EEC container via a config map.

Example: 
```
    "alpha.operator.dynatrace.com/extensions.debugExtensionDSstatsdlogoutboundminttraffic": "true"
```
is turned into:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: eec-config
  namespace: dynatrace
data:
  runtimeConfiguration: |
    {
      "revision": 1,
      "booleanMap": {
        "debugExtensionDSstatsdlogoutboundminttraffic": true
      },
      "stringMap": {},
      "longMap": {}
    }
```


src/controllers/activegate/reconciler/capability/eec_runtime_config.go
```diff
@@ -0,0 +1,79 @@
...
+		if parsedLongInt, err := strconv.ParseInt(val, 10, 64); err == nil {
+			longMap[runtimeProp] = parsedLongInt
+		} else if parsedBool, err := strconv.ParseBool(val); err == nil {
+			booleanMap[runtimeProp] = parsedBool
+		} else {
+			stringMap[runtimeProp] = val
+		}
+	}
+
+	return map[string]interface{}{
```
0sewa0, line 42: Couldn't this `map[string]interface{}` be an actual type ?
Something like:
```go
type EecRuntimeConfig struct {
	revision	int	`json:"revision"`
	booleanMap	map[string]bool	`json:"booleanMap"`
	stringMap	map[string]string	`json:"stringMap"`
	longMap	map[string]int64	`json:"longMap"`
}
```

0sewa0 (changes requested):

src/controllers/activegate/reconciler/capability/eec_runtime_config.go
toszr: Sounds great. Why didn't I think about it earlier?

Done 👍

src/controllers/activegate/reconciler/capability/reconciler_eec_runtime_config.go
```diff
@@ -0,0 +1,58 @@
...
+		if updateErr != nil {
+			return false, updateErr
+		}
+		return true, nil
+	}
+	return false, nil
+}
+
+func (r *Reconciler) isEecConfigMapOutdated(installedConfigMap, desiredConfigMap *corev1.ConfigMap) bool {
+	configMapsEqual := reflect.DeepEqual(installedConfigMap.Data, desiredConfigMap.Data) &&
```
chrismuellner, line 51: is comparing both `Data` and `BinaryData` necessary?

src/controllers/activegate/capability/capability.go
```diff
@@ -300,10 +306,11 @@ func statsdIngestBase() *capabilityBase {
...
 		Configuration: Configuration{
-			SetDnsEntryPoint:     true,
-			SetReadinessPort:     true,
-			SetCommunicationPort: true,
-			CreateService:        true,
+			SetDnsEntryPoint:       true,
+			SetReadinessPort:       true,
+			SetCommunicationPort:   true,
+			CreateService:          true,
+			CreateEecRuntimeConfig: true,
```
chrismuellner, line 313: why is this flag needed when it's only applicable for statsd? Do we need another check if we are using statsd?

chrismuellner (changes requested):

chrismuellner:
Annotations starting with `internal.operator...` should be managed by the operator.
If these annotations are defined by the user, we need to find a different name.

src/controllers/activegate/reconciler/statefulset/container_eec.go
```diff
@@ -119,6 +139,7 @@ func (eec *ExtensionController) buildVolumeMounts() []corev1.VolumeMount {
 		{Name: dataSourceMetadata, MountPath: statsdMetadataMountPoint, ReadOnly: true},
 		{Name: eecLogs, MountPath: extensionsLogsDir},
 		{Name: dataSourceStatsdLogs, MountPath: statsDLogsDir, ReadOnly: true},
+		{Name: "eec-config", MountPath: "/var/lib/dynatrace/remotepluginmodule/agent/conf/runtime"},
```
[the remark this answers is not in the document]
toszr: Right, good catch - done

src/controllers/activegate/reconciler/capability/eec_runtime_config_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+		eecConfigMap := CreateEecConfigMap(instance, "activegate")
+		runtimeConfigJson := eecConfigMap.Data["runtimeConfiguration"]
+
+		assert.Equal(t, "dynakube-activegate-eec-config", eecConfigMap.Name)
+
+		require.NotEmpty(t, eecConfigMap.Data)
+		require.NoError(t, json.Unmarshal([]byte(runtimeConfigJson), &runtimeConfig))
+		assert.Equal(t, 1., runtimeConfig["revision"])
+		assert.True(t, runtimeConfig["booleanMap"].(map[string]interface{})["debugExtensionDSstatsdlogoutboundminttraffic"].(bool))
```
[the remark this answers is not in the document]
toszr: I used generic `map[string]interface{}` before to hold bools/strings/ints and that required ugly dynamic casts. This code got outdated now as I switched to a more explicit type `EecRuntimeConfig`  with strongly-typed maps. 

It should be easier on the eyes now.

src/controllers/activegate/reconciler/capability/reconciler_eec_runtime_config.go
```diff
@@ -0,0 +1,58 @@
...
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func (r *Reconciler) createEecConfigMapIfNotExists() (bool, error) {
+	eecConfigMap := CreateEecConfigMap(r.Instance, r.ShortName())
+
+	getErr := r.Get(context.TODO(), client.ObjectKey{Name: eecConfigMap.Name, Namespace: eecConfigMap.Namespace}, eecConfigMap)
```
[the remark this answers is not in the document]
toszr: OK

src/controllers/activegate/capability/capability.go
toszr: It's needed because the config map is only useful if there are relevant annotations. Most client deployments won't even see the `ConfigMap` for EEC or StatsD runtime properties.

toszr:
OK, changed the prefix to `alpha.operator.dynatrace.com/extensions.`.

src/controllers/activegate/reconciler/capability/reconciler_eec_runtime_config.go
toszr: I prefer to compare both for the sake of safety, even though I expect only UTF-8 valid sequences in the property list. Better safe than sorry.

src/controllers/activegate/reconciler/capability/reconciler_eec_runtime_config.go
```diff
@@ -0,0 +1,58 @@
...
+	"reflect"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func (r *Reconciler) createEecConfigMapIfNotExists() (bool, error) {
```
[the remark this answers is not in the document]
toszr: Why not indeed. I remember I mimicked the methods after service create/update but I hated the code in December already so I took the opportunity now to rework it. Thanks for the tip.

src/controllers/activegate/reconciler/statefulset/container_eec.go
```diff
@@ -88,7 +89,26 @@ func (eec *ExtensionController) BuildVolumes() []corev1.Volume {
...
+				},
+			},
+		},
+	}
+}
+
+func BuildEecConfigMapName(instanceName string, module string) string {
+	if len(instanceName) == 0 || len(module) == 0 {
+		err := fmt.Errorf("empty instance or module name not allowed (instance: %s, module: %s)", instanceName, module)
+		log.Error(err, "problem building EEC config map name")
```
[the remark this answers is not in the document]
toszr: Yeah, good idea. Improved error handling around config map name generation.

0sewa0 (approved):

0sewa0 (approved):

PR #585, opened 2022-02-17
Makes readonly oneagent host volume the default

0sewa0:
# Description

- By default the oneagents get csi-volumes and readonly host mount in case of `cloudNative`/`hostMonitoring`.
  - `alpha.operator.dynatrace.com/feature-disable-oneagent-readonly-host-fs` can be used to disable it
- Updates the validation check that makes sure that the csi-driver is present. 
  - `hostMonitoring` needs it if readonly is not disabled
- Updates the template logic in helm to handle this new default and keep intact the "smart" way of deploying the csi-driver.
   - We only deploy the csi stuff in case the mode requires it. So now that `hostMonitoring` uses the csi driver by default, so some changes needed to be made.
- Adds checks during daemonset creation to make sure that classicFullStack daemonset wont be readonly (with opt in feature flag this was not super needed)

## How can this be tested?
Just deploy it with any mode(that deploys a oneagent daemonset) you want and it should just work.
In case of:
- classicFullStack: the oneagent daemonset should NOT have a csi volume and readonly host volume
- cloudNativeFullStack: the oneagent daemonset SHOULD have a csi volume and readonly host volume
- hostMonitoring: the oneagent daemonset SHOULD have a csi volume and readonly host volume

if the `alpha.operator.dynatrace.com/feature-disable-oneagent-readonly-host-fs: "true"` is present on the dynakube the oneagent daemonset should NOT have a csi volume and readonly host volume, no matter what.

if no csidriver is deployed, the validation webhook would deny hostMonitoring dynakubes (without the annotation)

## Checklist
- [x] Labeled the PR accordingly


src/api/v1beta1/properties.go
```diff
@@ -130,6 +130,14 @@ func (dk *DynaKube) StatsdImage() string {
 	return resolveImagePath(newStatsdImagePath(dk))
 }
 
+func (dk *DynaKube) NeedsReadOnlyOneAgents() bool {
+	inSupportedMode := dk.HostMonitoringMode() || dk.CloudNativeFullstackMode()
+	if inSupportedMode && !dk.FeatureDisableReadOnlyOneAgent() {
+		return true
+	}
+	return false
```
chrismuellner, line 138:
```suggestion
return (dk.HostMonitoringMode() || dk.CloudNativeFullstackMode())
     && !dk.FeatureDisableReadOnlyOneAgent()
```

chrismuellner (changes requested):

src/api/v1beta1/properties.go
0sewa0: less lines != better

0sewa0: (I think my version is clearer 😛 )

chrismuellner: I don't like `return true` and `return false` when you can just return the expression.
you can also do:
```go
inSupportedMode := dk.HostMonitoringMode() || dk.CloudNativeFullstackMode()
return inSupportedMode && !dk.FeatureDisableReadOnlyOneAgent()
```

src/api/v1beta1/feature_flags.go
```diff
@@ -37,7 +37,7 @@ const (
 	annotationFeatureUseActiveGateImageForStatsd      = annotationFeaturePrefix + "use-activegate-image-for-statsd"
 	annotationFeatureCustomEecImage                   = annotationFeaturePrefix + "custom-eec-image"
 	annotationFeatureCustomStatsdImage                = annotationFeaturePrefix + "custom-statsd-image"
-	AnnotationFeatureReadOnlyOneAgent                 = annotationFeaturePrefix + "oneagent-readonly-host-fs"
+	AnnotationFeatureDisableReadOnlyOneAgent          = annotationFeaturePrefix + "disable-oneagent-readonly-host-fs"
```
chrismuellner, line 40: default entry in the values file is now a double negation. I think keeping the same name, and disabling it by setting the flag to `false` is better.

src/controllers/csi/metadata/sqlite.go
```diff
@@ -271,14 +277,26 @@ func (a *SqliteAccess) UpdateOsAgentVolume(volume *OsAgentVolume) error {
...
 	var lastModified time.Time
 	err := a.querySimpleStatement(getOsAgentVolumeViaVolumeIDStatement, volumeID, &tenantUUID, &mounted, &lastModified)
 	if err != nil {
-		err = fmt.Errorf("couldn't get storage info for volume id '%s', err: %s", volumeID, err)
+		err = fmt.Errorf("couldn't get osAgentVolume info for volume id '%s', err: %s", volumeID, err)
+	}
+	return NewOsAgentVolume(volumeID, tenantUUID, mounted, &lastModified), err
+}
+
+// GetOsAgentVolumeViaTenantUUID gets an OsAgentVolume by its VolumeID
```
chrismuellner, line 292:
```suggestion
// GetOsAgentVolumeViaTenantUUID gets an OsAgentVolume by its TenantUUID
```

src/webhook/validation/preview.go
```diff
@@ -18,6 +18,9 @@ func oneAgentModePreviewWarning(dv *dynakubeValidator, dynakube *dynatracev1beta
 	} else if dynakube.ApplicationMonitoringMode() && dynakube.NeedsCSIDriver() {
 		log.Info("DynaKube with applicationMonitoring was applied, warning was provided.")
 		return fmt.Sprintf(featurePreviewWarningMessage, "applicationMonitoring")
+	} else if dynakube.HostMonitoringMode() && dynakube.NeedsCSIDriver() {
```
chrismuellner, line 21: I don't think this feature is in preview, a warning that the CSI driver is missing is enough.

chrismuellner (changes requested):

src/api/v1beta1/feature_flags.go
```diff
@@ -138,7 +138,9 @@ func (dk *DynaKube) FeatureCustomStatsdImage() string {
 	return dk.Annotations[annotationFeatureCustomStatsdImage]
 }
 
-// FeatureReadOnlyOneAgent is a feature flag that makes the operator deploy the oneagents in a readonly mode, where the csi-driver provides the volume for logs and such,
+// FeatureReadOnlyOneAgent is a feature flag to specify if the operator needs to deploy the oneagents in a readonly mode,
+// where the csi-driver would provide the volume for logs and such
+// Defaults to true
 func (dk *DynaKube) FeatureReadOnlyOneAgent() bool {
-	return dk.Annotations[AnnotationFeatureReadOnlyOneAgent] == "true"
+	return dk.Annotations[AnnotationFeatureReadOnlyOneAgent] != "false"
```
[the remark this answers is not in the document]
0sewa0: I wanted to write tests for it, so I needed it to be exported

chrismuellner, line 145: this check is way too inclusive
since it's opt-out anyway, we should change the feature flag to `DisableReadOnlyOneAgent` similar to the other features that are opt-out

src/api/v1beta1/properties.go
```diff
@@ -130,8 +130,22 @@ func (dk *DynaKube) StatsdImage() string {
...
 }
 
+func (dk *DynaKube) NeedsReadOnlyOneAgents() bool {
+	inSupportedMode := dk.HostMonitoringMode() || dk.CloudNativeFullstackMode()
+	return inSupportedMode && dk.FeatureReadOnlyOneAgent()
+}
+
 func (dk *DynaKube) NeedsCSIDriver() bool {
-	return dk.CloudNativeFullstackMode() || (dk.ApplicationMonitoringMode() && dk.Spec.OneAgent.ApplicationMonitoring.UseCSIDriver != nil && *dk.Spec.OneAgent.ApplicationMonitoring.UseCSIDriver)
+	if dk.CloudNativeFullstackMode() {
```
chrismuellner, line 139: chaining the checks can look nice if it's formatted correctly, this is pretty confusing/unreadable to me with multiple checks for the same thing

src/controllers/csi/driver/server.go
```diff
@@ -181,6 +181,9 @@ func (svr *CSIDriverServer) NodeUnpublishVolume(ctx context.Context, req *csi.No
 		}
 	}
 	log.Info("VolumeID not present in the database", "volumeID", volumeInfo.VolumeID, "targetPath", volumeInfo.TargetPath)
+	if err != svr.mounter.Unmount(volumeInfo.TargetPath) {
```
chrismuellner, line 184: why this weird check?

chrismuellner (changes requested):

src/controllers/csi/driver/server.go
0sewa0: this only happens if there is no info on the mount in the database, so I try to unmount in hopes that it will just work and let the pod die/restart

0sewa0: I noticed that I was dumb, fixed it 😅

chrismuellner (approved):

PR #586, opened 2022-02-18
Adds security context for EEC and StatsD containers (statsd-ingest)

toszr:
This changeset implements security architecture review remarks regarding the hardening of `statsd-ingest`-related containers in ActiveGate (EEC and StatsD). In particular, `privileged` and `allowPrivilegeEscalation` are disabled for both containers, and all capabilities are dropped by default.

### DIFF
--- a/src/kubeobjects/address_of/address_of.go
+++ b/src/kubeobjects/address_of/address_of.go
@@ -1,0 +1,9 @@
+package address_of
+
+func Int64(i int64) *int64 {
+	return &i
+}
+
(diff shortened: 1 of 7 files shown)


src/kubeobjects/address_of/address_of.go
```diff
@@ -0,0 +1,9 @@
+package address_of
```
0sewa0, line 1: praise the baby Jesus 🙏 
why didn't we do this before 😮

0sewa0 (approved):

PR #588, opened 2022-02-21
Maps certain StatsD-related annotations to EEC/StatsD container resource requests or limits

toszr:
This changeset interprets feature flags similar to:
```
  annotations:
    "alpha.operator.dynatrace.com/feature-activegate-statsd-resources-requests-memory": "500M"
```

to corresponding container resource requests or limits.

Valid flags:
- `alpha.operator.dynatrace.com/feature-activegate-statsd-resources-requests-cpu`
- `alpha.operator.dynatrace.com/feature-activegate-statsd-resources-requests-memory`
- `alpha.operator.dynatrace.com/feature-activegate-statsd-resources-limits-cpu`
- `alpha.operator.dynatrace.com/feature-activegate-statsd-resources-limits-memory`
- `alpha.operator.dynatrace.com/feature-activegate-eec-resources-requests-cpu`
- `alpha.operator.dynatrace.com/feature-activegate-eec-resources-requests-memory`
- `alpha.operator.dynatrace.com/feature-activegate-eec-resources-limits-cpu`
- `alpha.operator.dynatrace.com/feature-activegate-eec-resources-limits-memory`

### DIFF
--- a/src/api/v1beta1/feature_flags_resources.go
+++ b/src/api/v1beta1/feature_flags_resources.go
@@ -1,0 +1,133 @@
+package v1beta1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+const (
+	annotationFeatureActiveGate = annotationFeaturePrefix + "activegate-"
+	annotationFeatureEec        = annotationFeatureActiveGate + "eec-"
+	annotationFeatureStatsd     = annotationFeatureActiveGate + "statsd-"
+)
+
+// FeatureEecResourcesRequests is a feature flag to define CPU or memory requests for the EEC container
+func (dk *DynaKube) FeatureEecResourcesRequests(resourceName corev1.ResourceName) *resource.Quantity {
+	return eecResourceRequirements(dk, "requests-"+resourceName)
+}
+
+// FeatureEecResourcesLimits is a feature flag to define CPU or memory limits for the EEC container
+func (dk *DynaKube) FeatureEecResourcesLimits(resourceName corev1.ResourceName) *resource.Quantity {
+	return eecResourceRequirements(dk, "limits-"+resourceName)
+}
+
+// FeatureStatsdResourcesRequests is a feature flag to define CPU or memory requests for the StatsD container
+func (dk *DynaKube) FeatureStatsdResourcesRequests(resourceName corev1.ResourceName) *resource.Quantity {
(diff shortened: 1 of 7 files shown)


src/controllers/activegate/reconciler/statefulset/container_statsd.go
```diff
@@ -3,10 +3,12 @@ package statefulset
 import (
 	"fmt"
 
+	"github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
```
0sewa0, line 6:
```suggestion
	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
```

src/api/v1beta1/feature_flags_resources.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+const (
+	annotationFeatureActiveGate = annotationFeaturePrefix + "activegate-"
+	annotationFeatureEec        = annotationFeatureActiveGate + "eec-"
+	annotationFeatureStatsd     = annotationFeatureActiveGate + "statsd-"
+)
+
+// FeatureEecResourcesRequests is a feature flag to define CPU or memory requests for the EEC container
+func (dk *DynaKube) FeatureEecResourcesRequests(resourceName corev1.ResourceName) *resource.Quantity {
```
0sewa0, line 15: I understand that we can only enable statsd and eec with an annotation so we need other annotations to configure stuff for it.

But, this kinda feels wrong in a way. 
If we want to give the customer a new fully configurable activegate feature, why don't we just put it in the `activeGate` section, which already has resource limit/requests ?

(The PR with the `extensions` annotations is different because that adds config options that would require changes to the CR otherwise, which we would like to avoid.)

toszr: It was discussed on several occassions with @baichinger, @LukaszHalman and @chrismuellner, with contradictory conclusions. The most recent one was to avoid any changes to CRD and go with feature flags first, mostly due to the ongoing discussions about ActiveGate list (a.k.a. `v1beta2`).

To quote Bernhard (see IF-94, Feb 9th), 
> we must not introduce any new fields or changes to the CRD. Otherwise we end up in a release deadlock - a no-go from my point of view. 

Pity I can't link to our internal JIRA from here...

So:
- in the long run we want StatsD (or, in general, Extensions)-related properties to make it into CRD once we're confident about their stability
- for now we just want to release the feature without messing up the CRD too much (as format migration are already a headache).

0sewa0 (approved):

PR #592, opened 2022-02-22
Add validity check for certificates on webhook startup

luhi-DT:
Currently when a customer has an old certificate in the dynatrace namespace, the webhook will grab that cert and use it until the first renewal of his certificates (after 6 hours), because the validity is not checked, when the webhook grabs the certificates from the secret.

Webhook won't accept any connections/allow pod creations until it reloads certificates, which is after 6 hours. But the new certificates would be already in the cluster in a matter of minutes.

Therefore we should also check if the certificates are valid long enough, if the webhook starts.


src/kubeobjects/certificates.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func ValidateCertificateExpiration(certData []byte, renewalThreshold time.Duration, now time.Time, log logr.Logger) (bool, error) {
+	if block, _ := pem.Decode(certData); block == nil {
+		log.Info("failed to parse certificate", "error", "can't decode PEM file")
+		return false, nil
+	} else if cert, err := x509.ParseCertificate(block.Bytes); err != nil {
+		log.Info("failed to parse certificate", "error", err)
+		return false, err
+	} else if now.After(cert.NotAfter.Add(-renewalThreshold)) {
```
0sewa0, line 18: why don't you use `time.Now()` like you did in the first commit ?
so you don't have to pass in `now` all those times

chrismuellner: that means the time is the same for the whole execution
using `time.Now()` again could lead to unintended effects

src/cmd/operator/watcher.go
```diff
@@ -75,18 +78,29 @@ func (watcher *certificateWatcher) updateCertificatesFromSecret() (bool, error)
 	}
 
 	for _, filename := range []string{"tls.crt", "tls.key"} {
```
gkrenn, line 80: Would replace filenames with constants from cert controller

src/controllers/certificates/certs.go
```diff
@@ -101,23 +103,11 @@ func (cs *Certs) validateServerCerts(now time.Time) bool {
...
-	block, _ := pem.Decode(cs.Data[ServerCert])
-	if block == nil {
-		log.Info("failed to parse server certificates, renewing", "error", "can't decode PEM file")
-		return true
-	}
-
-	cert, err := x509.ParseCertificate(block.Bytes)
-	if err != nil {
-		log.Info("failed to parse server certificates, renewing", "error", err)
+	isValid, err := kubeobjects.ValidateCertificateExpiration(cs.Data[ServerCert], renewalThreshold, now, log)
```
gkrenn, line 106: You could use this validation function in `validateRootCert` too

gkrenn (dismissed):
See my comments

src/controllers/certificates/certs.go
luhi-DT: Sadly not, as `validateRootCert` sets also the parsed certificate to `cs.rootPublicCert`, which would not happen if I only call my function. Generating a new serverCert would fail if a valid root certificate is there, but cs.rootPublicCert is not set.

chrismuellner: can we change the logic on either side so we reuse as much functionality as possible?

luhi-DT: I guess i could refactor more here, wanted to avoid that in order to fix the bug. But could make a function which returns the certificate and a seperated one, which validates the cert validity

0sewa0 (approved):

PR #594, opened 2022-02-22
Includes DT_STATSD_INGEST_URL in the endpoint.properties file injected to pods

toszr:
If `statsd-ingest` capability is enabled for ActiveGate, `endpoint.properties` file is enhanced with 
```
DT_STATSD_INGEST_URL=dynakube-activegate.dynatrace:18125
```

It's used to discover StatsD service endpoint in the cluster (similar to `metrics-ingest` except for StatsD there is no token).


src/controllers/activegate/capability/statsd-ingest/consts.go
```diff
@@ -0,0 +1,11 @@
+package statsdingest
+
+import "github.com/Dynatrace/dynatrace-operator/src/controllers/activegate/internal/consts"
+
+const (
```
chrismuellner, line 5: for the rest of the codebase we have the constants in a `config.go` per package

toszr: Makes sense. I rearranged StatsD- and other ActiveGate-related constants as suggested.

0sewa0 (commented):
looks good, there is a conflict in the tests

toszr:
> looks good, there is a conflict in the tests

Yikes! Had to move the constants to `activegate/capability/config.go` to avoid package name clash. 

Might make sense to rethink ActiveGate file and package structure, similar to what @0sewa0 did with other controllers (why is there a `controllers/activegate` and `controllers/dynakube/activegate`?). Something in the plans, @mjgrzybek @aorcholski?

0sewa0 (approved):

PR #597, opened 2022-02-23
SecurityContext added to ActiveGate container

aorcholski:
All changes related to ActiveGate container:
1) the root filesystem is mounted ReadOnly.
2) security context added:
>  privileged: false
>  allowPrivilegeEscalation: false
>  readOnlyRootFilesystem: true *
>  runAsNonRoot: true
>  capabilities:
>    drop: ["all"]
* enabled by "alpha.operator.dynatrace.com/feature-activegate-readonly-fs":"true" annotation 
3) seccompProfile is set to:
>    type: RuntimeDefault 
4)  AppArmor is enabled via AG pod annotation *
>  container.apparmor.security.beta.kubernetes.io/activegate: runtime/default 
* enabled by "alpha.operator.dynatrace.com/feature-activegate-apparmor":"true" annotation

### DIFF
--- a/src/controllers/activegate/capability/config.go
+++ b/src/controllers/activegate/capability/config.go
@@ -3,6 +3,18 @@ package capability
 const (
 	ActiveGateContainerName = "activegate"
 
+	ActiveGateGatewayConfigVolumeName = "ag-lib-gateway-config"
+	ActiveGateGatewayTempVolumeName   = "ag-lib-gateway-temp"
+	ActiveGateGatewayDataVolumeName   = "ag-lib-gateway-data"
+	ActiveGateLogVolumeName           = "ag-log-gateway"
+	ActiveGateTmpVolumeName           = "ag-tmp-gateway"
+
+	ActiveGateGatewayConfigMountPoint = "/var/lib/dynatrace/gateway/config"
+	ActiveGateGatewayTempMountPoint   = "/var/lib/dynatrace/gateway/temp"
+	ActiveGateGatewayDataMountPoint   = "/var/lib/dynatrace/gateway/data"
+	ActiveGateLogMountPoint           = "/var/log/dynatrace/gateway"
+	ActiveGateTmpMountPoint           = "/var/tmp/dynatrace/gateway"
+
 	HttpsServicePortName = "https"
 	HttpsServicePort     = 443
 	HttpServicePortName  = "http"
--- a/src/controllers/activegate/reconciler/statefulset/container_eec.go
+++ b/src/controllers/activegate/reconciler/statefulset/container_eec.go
@@ -21,8 +21,6 @@ const extensionsRuntimeDir = "/var/lib/dynatrace/remotepluginmodule/agent/conf/r
 const activeGateInternalCommunicationPort = 9999
 
 const (
-	eecAuthToken = "auth-tokens"
-
 	dataSourceStartupArguments = "eec-ds-shared"
 	dataSourceAuthToken        = "dsauthtokendir"
 	eecLogs                    = "extensions-logs"
@@ -74,12 +72,6 @@ func (eec *ExtensionController) BuildContainer() corev1.Container {
 
 func (eec *ExtensionController) BuildVolumes() []corev1.Volume {
 	volumes := []corev1.Volume{
-		{
-			Name: eecAuthToken,
-			VolumeSource: corev1.VolumeSource{
-				EmptyDir: &corev1.EmptyDirVolumeSource{},
-			},
-		},
 		{
 			Name: dataSourceStartupArguments,
 			VolumeSource: corev1.VolumeSource{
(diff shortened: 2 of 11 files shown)


src/controllers/activegate/capability/config.go
```diff
@@ -3,6 +3,18 @@ package capability
 const (
 	ActiveGateContainerName = "activegate"
 
+	ActiveGateVolumeNameGatewayConfig = "ag-lib-gateway-config"
```
mjgrzybek, line 6: plz change consts names to pattern:
`ActiveGateVolumeNameGatewayConfig`-> `ActiveGateGatewayConfigVolumeName`

src/controllers/activegate/reconciler/statefulset/container_eec.go
```diff
@@ -21,8 +21,6 @@ const extensionsRuntimeDir = "/var/lib/dynatrace/remotepluginmodule/agent/conf/r
 const activeGateInternalCommunicationPort = 9999
 
 const (
-	eecAuthToken = "auth-tokens"
```
mjgrzybek, line 24: why is it removed?
EDIT: ok, I see, now it's `ActiveGateVolumeNameGatewayConfig`

src/controllers/activegate/reconciler/statefulset/statefulset.go
```diff
@@ -178,6 +179,10 @@ func buildContainers(stsProperties *statefulSetProperties, extraContainerBuilder
 	return containers
 }
 
+func pointerToBool(value bool) *bool {
```
mjgrzybek, line 182: use `address_of.Bool()` instead

src/kubeobjects/slice.go
```diff
@@ -11,6 +11,24 @@ func MountPathIsIn(volumeMounts []corev1.VolumeMount, mountPathToCheck string) b
 	return false
 }
 
+func MountPathIsReadOnly(volumeMounts []corev1.VolumeMount, mountPathToCheck string) bool {
```
mjgrzybek, line 14: can you merge
`func MountPathIsReadOnly(volumeMounts []corev1.VolumeMount, mountPathToCheck string) bool` and `func MountPathIsReadWrite(volumeMounts []corev1.VolumeMount, mountPathToCheck string) bool` 
to single function:
`MountPathIs(volumeMounts []corev1.VolumeMount, mountPathToCheck string, readOnly bool) bool` ?

chrismuellner (dismissed):
Can't be merged for now as it would break older Activegate versions

src/controllers/activegate/capability/config.go
aorcholski: we agreed to leave it as it is

0sewa0:
If you are brave and want to dive into the depts of ~hell~ helm, then you could make the dynakube have these annotations when `apparmor: true` in the helm values file.  😬 

We do it like this for a deployment, different annotations, same check
https://github.com/Dynatrace/dynatrace-operator/blob/master/config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml#L36-L38

src/controllers/activegate/reconciler/statefulset/statefulset_test.go
```diff
@@ -475,3 +569,19 @@ func buildTestInstance() *dynatracev1beta1.DynaKube {
 		},
 	}
 }
+
+func buildActiveGateVolumeDirectories(readOnly bool) []string {
```
chrismuellner, line 573: does this logic also exist for the implementation? It seems there we always add every volume

chrismuellner (changes requested):

src/kubeobjects/slice.go
```diff
@@ -11,6 +16,15 @@ func MountPathIsIn(volumeMounts []corev1.VolumeMount, mountPathToCheck string) b
 	return false
 }
 
+func MountPathIs(volumeMounts []corev1.VolumeMount, mountPathToCheck string, readOnly bool) bool {
```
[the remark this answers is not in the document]
aorcholski: The function checks if the MountPath exists and is ReadOnly or ReadWrite.

So maybe 
>MountPathIsReadOnlyOrReadWrite(.., mountPathToCheck string, mode bool)
would be better?

ReadOnlyMountPath and ReadWriteMountPath constants are defined for 'mode' parameter to clearly state requirement.

src/controllers/activegate/reconciler/statefulset/statefulset_test.go
aorcholski: Right, the logic no longer exist

aorcholski: Eventually logic is needed. Volumes created if needed (depend on StatsD and RO fs)

src/kubeobjects/slice.go
```diff
@@ -2,6 +2,11 @@ package kubeobjects
 
 import corev1 "k8s.io/api/core/v1"
 
+const (
+	ReadOnlyMountPath  = true
+	ReadWriteMountPath = false
```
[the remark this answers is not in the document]
aorcholski: ReadOnlyMountPath and ReadWriteMountPath constants are defined for 'mode' parameter of MountPathIsReadOnlyOrReadWrite function to clearly state requirement.

0sewa0 (changes requested):
You need to add an SCC so these changes work on Openshift (openshift will complain if you have SecurityContext for a pod but no SCC for it)

Example: https://github.com/Dynatrace/dynatrace-operator/blob/master/config/helm/chart/default/templates/Openshift/operator/securitycontextconstraints.yaml

0sewa0 (dismissed):

chrismuellner (dismissed):

PR #600, opened 2022-02-25
Removes statsd-ingest dependency on metrics-ingest

toszr:
So far, ActiveGate capability `statsd-ingest` enhanced `metrics-ingest` with StatsD-related port, endpoint properties and containers.

This changeset makes it independently controlled (i.e. it should be possible to define `statsd-ingest` alone or together with `metrics-ingest`).


src/util/strings.go
```diff
@@ -0,0 +1,13 @@
+package util
```
chrismuellner, line 1: no `util`/`helper` packages
these should be put somewhere more descriptive

src/api/v1beta1/properties_test.go
```diff
@@ -117,8 +117,20 @@ func TestTokens(t *testing.T) {
 
 func TestTenantUUID(t *testing.T) {
 	t.Run("happy path", func(t *testing.T) {
-		apiUrl := "https://demo.dev.dynatracelabs.com/api"
-		expectedTenantId := "demo"
+		const apiUrl = "https://demo.dev.dynatracelabs.com/api"
```
chrismuellner, line 120: why change this to `const` in all these tests?

src/api/v1beta1/properties.go
```diff
@@ -257,15 +262,19 @@ func tenantUUID(apiUrl string) (string, error) {
...
 	}
 
-	fqdn := parsedUrl.Hostname()
-	hostnameWithDomains := strings.FieldsFunc(fqdn,
-		func(r rune) bool { return r == '.' },
-	)
-	if len(hostnameWithDomains) < 1 {
-		return "", fmt.Errorf("problem getting tenant id from fqdn '%s'", fqdn)
+	// Path = "/e/<token>/api" -> ["e",  "<tenant>", "api"]
+	subPaths := util.Tokenize(parsedUrl.Path, '/')
```
chrismuellner, line 266: is there an advantage to using this instead of `strings.Split()`?

chrismuellner (changes requested):

src/api/v1beta1/properties_test.go
toszr: For consistency with the new test case I added. Constants should be preferred when possible.

src/api/v1beta1/properties.go
toszr: It's based on `strings.FieldsFunc()` which combines adjacent separators and skips empty split results automatically (in particular leading and/or trailing), [see playground example](https://go.dev/play/p/10LFof1C1ql).

`FieldsFunc` have an overcomplicated API which is why I wrapped it in `Tokenize`.

src/util/strings.go
toszr: Where would you put generic algorithms instead? `kubeobjects`?

src/api/v1beta1/properties.go
chrismuellner: I think we can use the STL for this: [url.Parse](https://pkg.go.dev/net/url#Parse)

src/util/strings.go
chrismuellner: If this is needed, we'd need to find a name for a generic helper package similar to `kubeobjects`

src/api/v1beta1/properties_test.go
chrismuellner: I don't think adding `const` to every variable in a test case is good idea
We've only had file/package scoped constants which can be reused for now, and I don't think we should introduce them on a function level

0sewa0: If this was a function unrelated to anything (was a util function used by different parts of the codebase) then it would make sense

AFAIK however this function is only used by `func (dk *DynaKube) TenantUUID() (string, error)` and is package private, so adding `const` feels like adding noise to the code more then anything else IMO

0sewa0: but if we want to be super-duper sure that it doesn't change anything, then its fine 😄

toszr: Matter of taste. I can revert the change if that's what you prefer, no big deal.

src/api/v1beta1/properties.go
toszr: `url.Parse()` serves different purpose - extracting protocol, hostname & domains, path & query arguments. It doesn't help with splitting the path into tokens.

I thought of [`path.Split()` ](https://pkg.go.dev/net/path#Split) as well and it doesn't do what I need. Either way, a helper function is needed, be it local or exported, named `Tokenize()` or something similar.

chrismuellner: I just don't see the advantage of a split function that ignores empty results (unless beginning/end). That sounds like a malformed URL to me that shouldn't be parsed without error.
`path.Split()` is for file paths, so I don't think this should be used here either.

src/util/strings.go
toszr: Removed package util and the `Tokenize()` function (replaced with bare `FieldsFunc()`).

src/api/v1beta1/properties.go
toszr: It's how STL `strings.FieldsFunc()` works - it was implemented and accepted this way before. Removed the `Tokenize()` wrapper to not fuel further discussion.

src/controllers/activegate/reconciler/capability/reconciler.go
```diff
@@ -56,7 +56,7 @@ func NewReconciler(capability capability.Capability, clt client.Client, apiReade
 func setReadinessProbePort() events.StatefulSetEvent {
 	return func(sts *appsv1.StatefulSet) {
 		if activeGateContainer, err := getActiveGateContainer(sts); err == nil {
-			activeGateContainer.ReadinessProbe.HTTPGet.Port = intstr.FromString(capability.HttpsServicePortName)
+			activeGateContainer.ReadinessProbe.HTTPGet.Port = intstr.FromInt(capability.HttpsServicePort)
```
chrismuellner, line 59: this is not in scope of this PR

toszr: Correct, I pushed more than I intended to. It turned out I went too far with disabling ActiveGate `metrics-ingest` features if it's not defined (AG container ports are also needed for probes).

toszr:
Do you have any further comments or can we move it forward?

mjgrzybek (approved):

PR #601, opened 2022-02-28
Change CSI reconciliation to install only if package does not exist

gkrenn:
Process agents that depend on the CSI driver fail monitoring if the underlying files are exchanged.

This fix will prevent installing the agent, if the files already exist on the node.

The bug was triggered by setting a different version on the tenant, but switching back to the old version again, which triggered an install on the provisioner.


src/controllers/csi/provisioner/agent.go
```diff
@@ -48,20 +48,23 @@ func newAgentUpdater(
...
-	targetDir := updater.path.AgentBinaryDirForVersion(tenantUUID, currentVersion)
+	targetVersion := updater.getOneAgentVersionFromInstance()
+	targetDir := updater.path.AgentBinaryDirForVersion(tenantUUID, targetVersion)
 
-	if _, err := updater.fs.Stat(targetDir); currentVersion != version || os.IsNotExist(err) {
-		log.Info("updating agent", "version", currentVersion, "previous version", version)
+	if _, err := updater.fs.Stat(targetDir); os.IsNotExist(err) {
+		log.Info("updating agent",
+			"targetVersion", targetVersion,
+			"installedVersion", installedVersion,
```
chrismuellner, line 59:
```suggestion
			"target version", targetVersion,
			"installed version", installedVersion,
```
consistency

chrismuellner (changes requested):

src/controllers/csi/provisioner/agent_test.go
```diff
@@ -246,6 +209,59 @@ func TestUpdateAgent(t *testing.T) {
 	})
 }
 
+func updateOneagent(t *testing.T, alreadyInstalled bool) {
```
0sewa0, line 212: such abstraction, much reuse 🍪

0sewa0 (approved):

chrismuellner (approved):

PR #602, opened 2022-02-28
Add openshift scc for operator and webhook

0sewa0:
# Description

For openshift you should provide an SCC if you have a securityContext set on your stuff
Also default apparmor breaks to many things, so this removes it and makes it optional

## How can this be tested?
Deploy it on openshift, it should just work.

## Checklist

### DIFF
--- a/config/deploy/openshift/openshift-all.yaml
+++ b/config/deploy/openshift/openshift-all.yaml
@@ -3968,6 +3968,98 @@ users:
 volumes:
   - "*"
 ---
+# Source: dynatrace-operator/templates/Openshift/operator/securitycontextconstraints.yaml
+# Copyright 2021 Dynatrace LLC
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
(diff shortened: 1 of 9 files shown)


config/deploy/openshift/openshift-all.yaml
```diff
@@ -3968,6 +3966,98 @@ users:
...
+metadata:
+  name: dynatrace-webhook
+allowPrivilegedContainer: false
+fsGroup:
+  type: RunAsAny
+priority: 1
+readOnlyRootFilesystem: true
+requiredDropCapabilities:
+  - ALL
+runAsUser:
```
mjgrzybek, line 4040: Can you switch it to `MustRunAsNonRoot`?
See: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#example-policies
```yaml
  runAsUser:
    # Require the container to run without root privileges.
    rule: 'MustRunAsNonRoot'
  seLinux:
    # This policy assumes the nodes are using AppArmor rather than SELinux.
    rule: 'RunAsAny'
```

chrismuellner (approved):

mjgrzybek (approved):

PR #604, opened 2022-03-01
Removes admission.Errored responses from podmutator webhook

0sewa0:
# Description

On OLM the namespaceSelector in the `MutatingWebhookConfiguration` is ignored. This leads to the Webhook intercepting Pod Creation for every pod in the Cluster, independent of the namespaceSelector in the Dynakube and the corresponding label on the namespace managed by the Operator.
The failurePolicy is set to ignore, with a timeout of 2s, but since the Webhook returns an error response before the 2s are up, the pod is blocked from being created.

## How can this be tested?
No matter what, the webhook will not stop the scheduling of a pod because of an error, instead the error's text should be in the message.

## Checklist

### DIFF
--- a/src/webhook/mutation/pod_mutator.go
+++ b/src/webhook/mutation/pod_mutator.go
@@ -601,7 +601,11 @@ func (m *podMutator) getNsAndDkName(ctx context.Context, req admission.Request) 
 
 	dkName, ok := ns.Labels[mapper.InstanceLabel]
 	if !ok {
-		rsp = silentErrorResponse(m.currentPodName, fmt.Errorf("no DynaKube instance set for namespace: %s", req.Namespace))
+		if kubesystem.DeployedViaOLM() {
+			rsp = admission.Patched("")
+		} else {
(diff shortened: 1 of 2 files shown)


src/webhook/mutation/pod_mutator.go
```diff
@@ -794,11 +803,18 @@ func updateContainerDataIngest(c *corev1.Container, pod *corev1.Pod, deploymentM
...
 }
 
 func fieldEnvVar(key string) *corev1.EnvVarSource {
 	return &corev1.EnvVarSource{FieldRef: &corev1.ObjectFieldSelector{FieldPath: key}}
 }
+
+func silentErrorResponse(podName string, err error) admission.Response {
+	rsp := admission.Patched("")
+	message := fmt.Sprintf("Failed to inject into pod: %s because %s", podName, err.Error())
+	rsp.Result.Message = message
```
luhi-DT, line 818: can't you set rsp.Result.Message with the error message directly instead of passing it to the intermediate variable?

luhi-DT (changes requested):

luhi-DT (approved):

PR #609, opened 2022-03-02
Reworked phase/status-field handling

gkrenn:
# Description

In our current implementation the phase dynakube status field should indicate if all oneagents were deployed successfully. This was not the case because our code was searching for an old daemonset name which was no longer there. Therefore if the field was once set to error, it was never reseted. The bug was fixed in this PR.

Additionally, the operator and the deployment of the dynakube is more than just generating oneagents => Advanced the phase field checking: Now, the dynakube status field is 
- set to error if
  - One token is invalid
  - An reconciliation error occurs
- set to deploying if 
  - oneagent or activegate pods are not ready yet
- set to running if
  - none of the other cases happend

To quickly show the state of the deployment, the phase field was added to `additionalPrinterColumns` in our dynakube crd. 

Besides of that, a reconciliation of the dynakube was reworked a little but, to decrease the number of reconciliation the operator needs and unit tests for the dynakube reconciler were added.

## How can this be tested?

Apply the new CRD and start in a monitoring mode you prefer.
In the operator logs you can see: 
- Less reconciliations
- The state changes of the phase field

With `kubectl get dynakube` you should see the additional printer column.

## Checklist


src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,89 @@
...
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	"github.com/Dynatrace/dynatrace-operator/src/controllers/activegate/capability"
+	appsv1 "k8s.io/api/apps/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (controller *DynakubeController) determineDynaKubePhase(instance *dynatracev1beta1.DynaKube) bool {
+	if instance.NeedsActiveGate() {
+		agPods, err := controller.numberOfMissingActiveGatePods(instance)
```
dt-team-kubernetes, line 15:
```suggestion
		activeGatePods, err := controller.numberOfMissingActiveGatePods(instance)
```

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,89 @@
...
+			return updatePhaseIfChanged(instance, dynatracev1beta1.Deploying)
+		}
+		if agPods < 0 {
+			log.Info("activegate sts not yet available", "dynakube", instance.Name)
+			return updatePhaseIfChanged(instance, dynatracev1beta1.Deploying)
+		}
+	}
+
+	if instance.CloudNativeFullstackMode() || instance.ClassicFullStackMode() || instance.HostMonitoringMode() {
+		oaPods, err := controller.numberOfMissingOneagentPods(instance)
```
dt-team-kubernetes, line 31:
```suggestion
		oneAgentPods, err := controller.numberOfMissingOneagentPods(instance)
```

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,89 @@
...
+	return updatePhaseIfChanged(instance, dynatracev1beta1.Running)
+}
+
+func (controller *DynakubeController) numberOfMissingOneagentPods(instance *dynatracev1beta1.DynaKube) (int32, error) {
+	dsActual := &appsv1.DaemonSet{}
+	instanceName := instance.OneAgentDaemonsetName()
+	err := controller.client.Get(context.TODO(), types.NamespacedName{Name: instanceName, Namespace: instance.Namespace}, dsActual)
+
+	if k8serrors.IsNotFound(err) {
+		return -1, nil
```
dt-team-kubernetes, line 55: imo, if there is no deamonset, then there are no pods missing, therefore it should return 0.
Instead of checking for `value < 0`, I'd propagate the error and check in `determineDynakubePhase` if it is a not found error

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if err != nil {
+		return -1, err
+	}
+	return dsActual.Status.CurrentNumberScheduled - dsActual.Status.NumberReady, nil
+}
+
+func (controller *DynakubeController) numberOfMissingActiveGatePods(instance *dynatracev1beta1.DynaKube) (int32, error) {
+	caps := generateActiveGateCapabilities(instance)
```
dt-team-kubernetes, line 64:
```suggestion
	capabilities := generateActiveGateCapabilities(instance)
```

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,89 @@
...
+
+	sum := int32(0)
+	found := false
+
+	for _, c := range caps {
```
dt-team-kubernetes, line 69:
```suggestion
	for _, capability := range caps {
```

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,89 @@
...
+		return -1, err
+	}
+	return dsActual.Status.CurrentNumberScheduled - dsActual.Status.NumberReady, nil
+}
+
+func (controller *DynakubeController) numberOfMissingActiveGatePods(instance *dynatracev1beta1.DynaKube) (int32, error) {
+	caps := generateActiveGateCapabilities(instance)
+
+	sum := int32(0)
+	found := false
```
dt-team-kubernetes, line 67:
```suggestion
	capabilityFound := false
```

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		if err != nil {
+			return -1, err
+		}
+		found = true
+		sum += stsActual.Status.CurrentReplicas - stsActual.Status.ReadyReplicas
+	}
+
+	if !found {
+		return -1, nil
```
dt-team-kubernetes, line 85: same as above; imo, if there are no statefulsets found, no pods are not needed, therefore not missing and it should return 0

dt-team-kubernetes (changes requested):
I was logged in with the team user and only noticed after writing the comments
\- Michael

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -217,23 +221,26 @@ func (controller *DynakubeController) reconcileDynaKube(ctx context.Context, dkS
 		upd, err = oneagent.NewOneAgentReconciler(
 			controller.client, controller.apiReader, controller.scheme, dkState.Instance, daemonset.HostMonitoringFeature,
 		).Reconcile(ctx, dkState)
-		if dkState.Error(err) || dkState.Update(upd, defaultUpdateInterval, "infra monitoring reconciled") {
+		if dkState.Error(err) {
 			return
 		}
+		dkState.Update(upd, defaultUpdateInterval, "infra monitoring reconciled")
```
0sewa0, line 227: `dkState.Update` is a bit weird, at this point its a logging util
also the last setting of the interval will win, (also if we always use `defaultUpdateInterval` always then why pass it in 😅 )

src/controllers/dynakube/dynakube_phase.go
gkrenn: capability is already a package name, would leave c there

meik99:
```suggestion
	for _, activeGateCapability := range caps {
```

src/controllers/dynakube/dynakube_controller.go
meik99: On that note, afaik, `upd` is a boolean. Assuming `defaultUpdateInterval` is refactored to a constant, it would seem prudent to combine the two remaining parameters into a struct, which may or may not also take `dkState`, and call Update with it or, if `dkState` is also in that struct, call it as a receiver function

gkrenn: I agree that the interval parameter should be removed as it does not make sense. 

I think it's fine that it's kind of a logging util, as it adds the benefit that it sets the updated field for you on updates.

gkrenn: For combining the 2 variables into a struct, what would be our advantage?

meik99: > For combining the 2 variables into a struct, what would be our advantage?

increased read- and maintainability of the code base.
Also it's prudent to combine parameters into logical units, i.e., structs, and reduce parameter counts of functions.

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -157,15 +155,17 @@ func (controller *DynakubeController) reconcileDynaKube(ctx context.Context, dkS
...
+	dkState.Update(upd, "Token conditions updated")
 	if dkState.Error(err) {
 		log.Error(err, "failed to check tokens")
 		return
 	}
 	dkState.ValidTokens = true
 	if !dtcReconciler.ValidTokens {
 		dkState.ValidTokens = false
-		log.Info("paas or api token not valid", "name", dkState.Instance.GetName())
+		errMsg := "paas or api token not valid"
```
chrismuellner, line 166: this also happens with an invalid dataingest token, so I'd just change to error message to something like `tokens not valid`

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -183,7 +183,9 @@ func (controller *DynakubeController) reconcileDynaKube(ctx context.Context, dkS
 			// If there are errors log them, but move on.
 			log.Info("Istio: failed to reconcile objects", "error", err)
 		} else if upd {
-			dkState.Update(true, 30*time.Second, "Istio: objects updated")
+			dkState.Update(true, "Istio: objects updated")
+			dkState.RequeueAfter = 30 * time.Second
```
chrismuellner, line 187: constant for the shorter requeue time

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,86 @@
+package dynakube
```
chrismuellner, line 1: wouldn't this fit better in `dynakube/status`?

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,86 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (controller *DynakubeController) determineDynaKubePhase(instance *dynatracev1beta1.DynaKube) bool {
+	if instance.NeedsActiveGate() {
+		activeGatePods, err := controller.numberOfMissingActiveGatePods(instance)
+		if err != nil {
+			log.Error(err, "activegate sts could not be accessed", "dynakube", instance.Name)
```
chrismuellner, line 17: why shorten to `sts`? we have enough space in the logs

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,86 @@
...
+	"context"
+
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	"github.com/Dynatrace/dynatrace-operator/src/controllers/activegate/capability"
+	appsv1 "k8s.io/api/apps/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (controller *DynakubeController) determineDynaKubePhase(instance *dynatracev1beta1.DynaKube) bool {
```
chrismuellner, line 13:
```suggestion
func (controller *DynakubeController) determineDynaKubePhase(dynakube *dynatracev1beta1.DynaKube) bool {
```
since we are dealing with a `dynakube`, we should also name the variable like that. `instance` could be any instance

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -124,7 +120,9 @@ func (controller *DynakubeController) Reconcile(ctx context.Context, request rec
 
 	if dkState.Err != nil {
 		if !dkState.ValidTokens {
-			return reconcile.Result{}, fmt.Errorf("paas or api token not valid")
+			instance.Status.SetPhase(dynatracev1beta1.Error)
+			_ = controller.updateCR(ctx, instance)
+			return reconcile.Result{RequeueAfter: status.DefaultUpdateInterval}, nil
```
chrismuellner, line 125: this feels wrong to access the `defaultUpdateInterval` once here. Would defining it in this file and passing it to the status be an option?

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -217,23 +219,26 @@ func (controller *DynakubeController) reconcileDynaKube(ctx context.Context, dkS
 		upd, err = oneagent.NewOneAgentReconciler(
 			controller.client, controller.apiReader, controller.scheme, dkState.Instance, daemonset.HostMonitoringFeature,
 		).Reconcile(ctx, dkState)
-		if dkState.Error(err) || dkState.Update(upd, defaultUpdateInterval, "infra monitoring reconciled") {
+		if dkState.Error(err) {
 			return
 		}
+		dkState.Update(upd, "infra monitoring reconciled")
```
chrismuellner, line 225:
```suggestion
		dkState.Update(upd, "host monitoring reconciled")
```

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -217,23 +219,26 @@ func (controller *DynakubeController) reconcileDynaKube(ctx context.Context, dkS
...
 	} else if dkState.Instance.CloudNativeFullstackMode() {
 		upd, err = oneagent.NewOneAgentReconciler(
 			controller.client, controller.apiReader, controller.scheme, dkState.Instance, daemonset.CloudNativeFeature,
 		).Reconcile(ctx, dkState)
-		if dkState.Error(err) || dkState.Update(upd, defaultUpdateInterval, "cloud native infra monitoring reconciled") {
+		if dkState.Error(err) {
 			return
 		}
+		dkState.Update(upd, "cloud native infra monitoring reconciled")
```
chrismuellner, line 233:
```suggestion
		dkState.Update(upd, "cloud native fullstack monitoring reconciled")
```

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -244,21 +249,27 @@ func (controller *DynakubeController) reconcileDynaKube(ctx context.Context, dkS
 			log.Error(err, "update of a map of namespaces failed")
 		}
 		upd, err := initgeneration.NewInitGenerator(controller.client, controller.apiReader, dkState.Instance.Namespace).GenerateForDynakube(ctx, dkState.Instance)
-		if dkState.Error(err) || dkState.Update(upd, defaultUpdateInterval, "new init script created") {
+		if dkState.Error(err) {
 			return
 		}
+		dkState.Update(upd, "new init script created")
```
chrismuellner, line 255:
```suggestion
		dkState.Update(upd, "new init secret created")
```

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -244,21 +249,27 @@ func (controller *DynakubeController) reconcileDynaKube(ctx context.Context, dkS
...
+			dkState.Update(upd, "new data-ingest endpoint secret created")
 		} else {
 			err = endpointSecretGenerator.RemoveEndpointSecrets(ctx, dkState.Instance)
 			if dkState.Error(err) {
 				return
 			}
 		}
+		if dkState.Instance.ApplicationMonitoringMode() {
+			dkState.Instance.Status.SetPhase(dynatracev1beta1.Running)
+			dkState.Update(true, "application only reconciled")
```
chrismuellner, line 271:
```suggestion
			dkState.Update(true, "application monitoring reconciled")
```

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -157,15 +155,17 @@ func (controller *DynakubeController) reconcileDynaKube(ctx context.Context, dkS
...
 		log.Error(err, "failed to check tokens")
 		return
 	}
 	dkState.ValidTokens = true
 	if !dtcReconciler.ValidTokens {
 		dkState.ValidTokens = false
-		log.Info("paas or api token not valid", "name", dkState.Instance.GetName())
+		errMsg := "paas or api token not valid"
+		log.Info(errMsg, "name", dkState.Instance.GetName())
+		dkState.Err = fmt.Errorf(errMsg)
```
chrismuellner, line 168: this shouldn't be set explicitly, but use `dkstate.Error(..)`. Is there a reason that `Err` is even exported?

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,86 @@
...
+	if instance.NeedsActiveGate() {
+		activeGatePods, err := controller.numberOfMissingActiveGatePods(instance)
+		if err != nil {
+			log.Error(err, "activegate sts could not be accessed", "dynakube", instance.Name)
+			return updatePhaseIfChanged(instance, dynatracev1beta1.Error)
```
chrismuellner, line 18: wouldn't something similar to `status.SetPhaseOnError` work better here?
Could hide implementation details for the other 3 phases and have the phase logic in a single file instead of split.

src/controllers/dynakube/dynakube_controller_test.go
```diff
@@ -44,6 +45,45 @@ const (
 	testAnotherProtocol = "test-another-protocol"
 )
 
+func TestMonitoringModesDynakube_Reconcile(t *testing.T) {
+	deploymentModes := map[string]dynatracev1beta1.OneAgentSpec{
+		"host":            {HostMonitoring: &dynatracev1beta1.HostMonitoringSpec{}},
```
chrismuellner, line 50: names don't reflect actual monitoring modes we have in the dynakube

chrismuellner (changes requested):

src/controllers/dynakube/dynakube_phase.go
gkrenn: As there are only receiver functions for DynakubeController in there, I would say no

chrismuellner: `phase` is a property of `status` so this should be visible in the file structure as well

src/controllers/dynakube/dynakube_controller.go
gkrenn: In line 121 we check this error in the dynakube controller so it has to be exported, but I agree, in this case the error function should be called

src/controllers/dynakube/dynakube_controller.go
0sewa0: that is what we had originally, which was bad as it always the same value was passed it

gkrenn: Actually, we could just use `dkState.RequeueAfter`

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (controller *DynakubeController) numberOfMissingActiveGatePods(dynakube *dynatracev1beta1.DynaKube) (int32, error) {
+	capabilities := generateActiveGateCapabilities(dynakube)
+
+	sum := int32(0)
+	capabilityFound := false
+
+	for _, activeGateCapability := range capabilities {
+		stsActual := &appsv1.StatefulSet{}
```
0sewa0, line 67: stsActual doesn't mean anything
```suggestion
		capabilityStatefulSet:= &appsv1.StatefulSet{}
```

src/controllers/dynakube/dynakube_phase.go
```diff
@@ -0,0 +1,86 @@
...
+			log.Info("oneagent daemonset is still deploying", "dynakube", dynakube.Name)
+			return updatePhaseIfChanged(dynakube, dynatracev1beta1.Deploying)
+		}
+	}
+
+	return updatePhaseIfChanged(dynakube, dynatracev1beta1.Running)
+}
+
+func (controller *DynakubeController) numberOfMissingOneagentPods(dynakube *dynatracev1beta1.DynaKube) (int32, error) {
+	dsActual := &appsv1.DaemonSet{}
```
0sewa0, line 50:
```suggestion
	oneagentDaemonSet := &appsv1.DaemonSet{}
```

src/controllers/dynakube/dynakube_phase.go
0sewa0: `activeGateStatefulSet` is also good

0sewa0 (approved):

PR #611, opened 2022-03-03
Removes "alpha." from the feature flags + adds deprecation warning for old flags

0sewa0:
# Description

Feature flag shouldn't have the `alpha.` prefix, because they don't mean anything.
And moving them to `beta` or something else will not improve them just introduce more complex logic in the code.

If we want to deprecate 1 or a set of feature-flags we can do it using the validation webhook. 

But because some user might already use some of these we should support them for a release or two, so this PR enables the use of both.
There is an exception when it comes to the new `statsd` feature-flags as they were not released before, so we don't have to support it there.

Also adds a warning in the webhook, if there is a deprecated annotation present in the annotation of the dynakube.

## How can this be tested?
Try to apply a dynakube, with a deprecated feature-flag, it should work but you should also get a warning.
Then do the same but without the `alpha.` part, it should still work, but without the warning on the apply

## Checklist


src/api/v1beta1/feature_flags.go
```diff
@@ -25,12 +25,13 @@ import (
...
-	PublicAnnotationPrefix                            = "alpha.operator.dynatrace.com/"
-	annotationFeaturePrefix                           = PublicAnnotationPrefix + "feature-"
+	DeprecatedFeatureFlagPrefix                       = "alpha."
+	FeatureFlagAnnotationPrefix                       = "operator.dynatrace.com/"
+	annotationFeaturePrefix                           = FeatureFlagAnnotationPrefix + "feature-"
 	annotationFeatureDisableActiveGateUpdates         = annotationFeaturePrefix + "disable-activegate-updates"
 	annotationFeatureDisableHostsRequests             = annotationFeaturePrefix + "disable-hosts-requests"
 	annotationFeatureOneAgentMaxUnavailable           = annotationFeaturePrefix + "oneagent-max-unavailable"
-	annotationFeatureEnableWebhookReinvocationPolicy  = annotationFeaturePrefix + "enable-webhook-reinvocation-policy"
+	AnnotationFeatureEnableWebhookReinvocationPolicy  = annotationFeaturePrefix + "enable-webhook-reinvocation-policy"
```
mjgrzybek, line 34: I don't like it exported among other similar but not exported consts.
Maybe, to make it consistent, all annotations should be exported to make usable in unit tests of other packages?

src/webhook/validation/deprecation.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+)
+
+const (
+	featureDeprecatedWarningMessage = `DEPRACETED: %s`
```
toszr, line 11:
```suggestion
	featureDeprecatedWarningMessage = `DEPRECATED: %s`
```

src/webhook/validation/deprecation.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func deprecatedFeatureFlagFormat(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
+	if dynakube.Annotations == nil {
+		return ""
+	}
+	for annotation := range dynakube.Annotations {
```
toszr, line 18:
```suggestion
	if len(dynatracev1beta1.FlagsWithPrefix(instance, dynatracev1beta1.DeprecatedFeatureFlagPrefix+dynatracev1beta1.FeatureFlagAnnotationPrefix)) > 0 {
```

src/api/v1beta1/feature_flags.go
```diff
@@ -115,33 +111,43 @@ func (dk *DynaKube) getDefaultIgnoredNamespaces() []string {
...
 
 // FeatureDisableReadOnlyOneAgent is a feature flag to specify if the operator needs to deploy the oneagents in a readonly mode,
 // where the csi-driver would provide the volume for logs and such
 // Defaults to false
 func (dk *DynaKube) FeatureDisableReadOnlyOneAgent() bool {
-	return dk.Annotations[AnnotationFeatureDisableReadOnlyOneAgent] == "true"
+	return dk.getFeatureFlagRaw(AnnotationFeatureDisableReadOnlyOneAgent) == "true"
+}
+
+func (dk *DynaKube) getFeatureFlagRaw(annotation string) string {
```
toszr, line 145: Consider moving it to `internal_operator_flags.go`

toszr (dismissed):
Are you 100% sure all the non-alpha flags will be stable from now on and supported forever? Maybe we do want to introduce versioned flags (`beta1.` or so) to be on the safe side? After all they can be used by customers.

src/api/v1beta1/feature_flags.go
```diff
@@ -47,17 +48,17 @@ var (
 
 // FeatureDisableActiveGateUpdates is a feature flag to disable ActiveGate updates.
 func (dk *DynaKube) FeatureDisableActiveGateUpdates() bool {
-	return dk.Annotations[annotationFeatureDisableActiveGateUpdates] == "true"
+	return dk.getFeatureFlagRaw(annotationFeatureDisableActiveGateUpdates) == "true"
```
toszr, line 51: Out of scope of this changeset, but food for thought: `[strconv.ParseBool()](https://pkg.go.dev/strconv#hdr-Numeric_Conversions)`

src/api/v1beta1/feature_flags.go
0sewa0: I feel it wouldnt be in a better place in `internal_operator_flags.go` as that file is for internal flags, and yes this function is a more general function, but I think its fine here :D

src/webhook/validation/deprecation.go
0sewa0: noice 👍

src/api/v1beta1/feature_flags.go
0sewa0: `ParseBool` has an error return so, we would need to handle that, so it wouldn't really improve the code 😅

src/api/v1beta1/feature_flags.go
0sewa0: Made everything public to be consistent, and in a perfect we should have unittests for all of them

src/api/v1beta1/feature_flags.go
toszr: It would improve the user experience of these flags in that you could specify boolean values in a predictable, documented way. For example, if you specify `"True"` and use `.ParseBool()`, it will be interpreted as boolean `true`. How will the current code interpret `"True"`? :)

No big deal though and out of scope of this PR as I mentioned.

src/api/v1beta1/feature_flags.go
toszr: OK. IMO `getFeatureFlagRaw()` and functions for internal flags belong together but some further refactoring would be needed, not mandatory at this point.

mjgrzybek (approved):

PR #618, opened 2022-03-08
Adds database overview to the logs of the csi driver on error and startup

0sewa0:
# Description
We don't have a good way to check whats in the database, so on startup after the correction logic runs we log the content of the database. We also do this when the csi server stops.

## How can this be tested?
When you deploy the csi driver, in the logs there should be a line around the top (~3 line or so) that has the contents of the database on that node.

## Checklist


src/controllers/csi/metadata/sqlite.go
```diff
@@ -95,6 +95,22 @@ const (
 	WHERE TenantUUID = ?;
 	`
 
+	// GET ALL
+	getAllDynakubesStatement = `
+		SELECT Name, TenantUUID, LatestVersion
```
gkrenn, line 100: Would remove one tab to be identical with the previous statements

src/controllers/csi/metadata/sqlite_test.go
```diff
@@ -118,20 +118,69 @@ func TestUpdateDynakube(t *testing.T) {
...
+func TestGetAllDynakubes(t *testing.T) {
+	db := FakeMemoryDB()
+	err := db.InsertDynakube(&testDynakube1)
+	assert.Nil(t, err)
+	err = db.InsertDynakube(&testDynakube2)
+	assert.Nil(t, err)
+
+	dynakubes, err := db.GetAllDynakubes()
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(dynakubes))
```
gkrenn, line 144: Would even check if the returned objects are dk1 & dk2

Similar for the next tests

gkrenn (approved):
Beside of the small changes you could make, seems to be fine

PR #619, opened 2022-03-08
Adds detailed logs for the standalone init

0sewa0:
# Description

Adds a bunch of logs to the standalone init to make debuging easier, and it only runs once so we are not really spamming.
Also the logs should not have any values that can be a security issue (apiToken, paasToken, certs)

## How can this be tested?
In the install-oneagent container the logs are detailed.

## Checklist


src/standalone/secret.go
```diff
@@ -31,6 +31,22 @@ type SecretConfig struct {
 	ClusterID string `json:"clusterID"`
 }
 
+func (secret SecretConfig) logContent() {
```
mjgrzybek, line 34: Conditional secrets printing gives a hint of given field is set or not. 
It should be either never printed or always printed.

0sewa0: every field is always printed (the logger does it like that, and I don't want to list it out by hand)
if its not set then its printed with an empty value
if its set to something ==> "***"

mjgrzybek: Nit: add `const secretValue = "***"` and replace hardocded `"***"` with `secretValue`

mjgrzybek (approved):

PR #647, opened 2022-03-16
Only generate manifests for the selected platform

meik99:
# Description

Please include the following:
- a summary of the change
- which issue is fixed (if there is one)
- relevant motivation and context

Adds conditionals to `make manifests` to avoid building and overwriting manifests for a platform which is currently not being built.
Without these conditions manifests get overwritten with potentially wrong images if built for a release.

## How can this be tested?

Execute `make manifests` since the default PLATFORM is openshift, only openshift-all.yaml, openshift.yaml, openshift-csi.yaml and openshift/kustomization.yaml should be created.
If PLATFORM is set to kubernetes, only the kubernetes specific manifests should be generated.

No specific environment is needed.

## Checklist
- [] Unit tests have been updated/added


0sewa0 (dismissed):

0sewa0:
you reused the `PLATFORM` that was used for only bundles, so now the default value is openshift for `make manifests`, is that good ?

meik99:
> you reused the `PLATFORM` that was used for only bundles, so now the default value is openshift for `make manifests`, is that good ?
> 
> the deploy commands should be updated accordingly

Some values used by the manifests have also been depending on the `PLATFORM` parameter, they were just overwritten before. Now only needed manifests are created.

What needs to be updated concerning the deploy commands?

0sewa0 (approved):

PR #655, opened 2022-03-19
allow for csi tolerations configuration in values.yaml

qgervacio:
## Description

I have the operator 0.5.0 configured using `applicationMonitoring: true` and `useCSIDriver: true`. Due to how we setup our cluster with 2 node groups (1 for `app` workloads, 1 for `ops` workload) the CSI pods are only scheduled in the `app` due to taints in `ops`. This caused a problem when enabling the monitoring on ingress (https://www.dynatrace.com/support/help/setup-and-configuration/setup-on-container-platforms/kubernetes/instrument-ingress-nginx) as the ingress daemonset is configured to have `ops` affinity. Below is the error thrown:

```
MountVolume.SetUp failed for volume "oneagent-bin" : kubernetes.io/csi: mounter.SetUpAt failed to get CSI client: driver name csi.oneagent.dynatrace.com not found in the list of registered 
CSI drivers
```

The update is to allow for CSI tolerations configurations in values.yaml

## How can this be tested?
- Tested in EKS 1.21 on operator and chart 0.5.0 with activegate 1.233.152.20220204-172150
- Also with enabling `applicationMonitoring: true` and `useCSIDriver: true` then running `helm template test . --set apiToken=abc`

### DIFF
--- a/config/helm/chart/default/templates/Common/csi/daemonset.yaml
+++ b/config/helm/chart/default/templates/Common/csi/daemonset.yaml
@@ -37,6 +37,9 @@ spec:
       labels:
         {{- include "dynatrace-operator.csiLabels" . | nindent 8 }}
     spec:
+      {{- if .Values.csi.tolerations }}
+      tolerations: {{ .Values.csi.tolerations | toYaml | nindent 8 }}
+      {{- end }}
       containers:
         # Used to receive/execute gRPC requests (NodePublishVolume/NodeUnpublishVolume) from kubelet to mount/unmount volumes for a pod
         # - Needs access to the csi socket, needs to read/write to it, needs root permissions to do so.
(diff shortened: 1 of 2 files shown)


config/helm/chart/default/templates/Common/csi/daemonset.yaml
```diff
@@ -37,6 +37,9 @@ spec:
       labels:
         {{- include "dynatrace-operator.csiLabels" . | nindent 8 }}
     spec:
+      {{- if .Values.csi.tolerations }}
```
0sewa0, line 40:
```suggestion
      {{- if (.Values.csi).tolerations }}
```

the csi section is quite new in the values.yaml, so just to be safe, this makes it null safe.
`--reuse-values` could break stuff without this (user doesn't have the csi section in there old values.yaml, and use `--reuse-values`, install will fail)

0sewa0 (changes requested):

edogden-pbx:
When is this going to be included, its causing an issue with my test deployment not being able to include the CSI in multiple node groups

0sewa0:
requested changes 24 days ago, no response, closing it

will be done by the team

PR #656, opened 2022-03-21
Optionally parallel downloading csi-provisioner

0sewa0:
# Description

Introduces a way to run multiple reconciles in parallel in the csi-provisioner, hidden behind an ENVVAR (`MAX_PARALLEL_RECONCILES`).

This is trying to speed up the startup incase multiple dynakubes with different tenants are deployed at the same time all needing the csi driver to mount volumes.

### Edge cases we check for:
- Out of range values for `MAX_PARALLEL_RECONCILES` (should be between 1 - 40)
- We requeue a dynakube reconcile request if:
  - we reached the max defined in the envvar for parallel reconciles running at the same time
  - we are already reconciling the requested dynakube in one of the parallel reconciles

### Gotchas:
- The default memoryLimit and cpuLimit were meant for 1 download at a time and not `x`, so those probably need to be raised at high number of parallel downloads
  - memoryLimit can cause `CrashLoopBackOff` because of `OOMKilled`
  - cpuLimit can cause the downloads to be slow, especially at the unzip part. (the pods wont get killed but kubernetes wont give the necessary cpu to do the job in time)
- Error handling suffers a bit, due to the fact that the framework we use for reconciliation is not meant to be used this way, so it can happen that something is failing in the provisioner yet we can only find it out from the logs of the container. 

## How can this be tested?
1. Have a cluster (that didn't have csi-drivers on them before to make sure that we always have to download the binaries)
2. Modify the csi daemonset yaml to have the `MAX_PARALLEL_RECONCILES` set to something more then "1"
3. Deploy as many dynakubes with different tenants as you can do. (`applicationMonitoring` requires less setup, don't forget to use the `useCSIDriver: true`)
4. Deploy sample apps at the same time for all the dynakubes. (be aware of the `namespaceSelector`)

Expected result: 
Parallel downloads for each tenant (grep for `parallel` in the logs)

## Checklist


src/controllers/csi/provisioner/controller.go
```diff
@@ -49,23 +50,25 @@ type OneAgentProvisioner struct {
 	apiReader    client.Reader
 	opts         dtcsi.CSIOptions
 	dtcBuildFunc dynakube.DynatraceClientFunc
-	fs           afero.Fs
 	recorder     record.EventRecorder
 	db           metadata.Access
 	path         metadata.PathResolver
+
+	currentParallel int64
```
mjgrzybek, line 57:
```suggestion
	currentParallelDownloads int64
```

src/controllers/csi/provisioner/controller.go
```diff
@@ -49,23 +50,25 @@ type OneAgentProvisioner struct {
...
+	maxParallel     int64
```
mjgrzybek, line 58:
```suggestion
	maxParallelDownloads     int64
```

src/controllers/csi/provisioner/controller.go
```diff
@@ -76,6 +79,27 @@ func (provisioner *OneAgentProvisioner) SetupWithManager(mgr ctrl.Manager) error
 }
 
 func (provisioner *OneAgentProvisioner) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	fs := afero.NewOsFs()
+	if provisioner.maxParallel > 0 {
+		if provisioner.currentParallel == provisioner.maxParallel {
+			log.Info("max parallel downloads reached, requeued")
+			return reconcile.Result{RequeueAfter: shortRequeueDuration}, nil
+		}
```
mjgrzybek, line 87:
```suggestion

func (provisioner *OneAgentProvisioner) eligibleToParallelDownloads() bool {
	return provisioner.maxParallel > 0
}


func (provisioner *OneAgentProvisioner) parallelDownloadsLimitReached() bool {
	return provisioner. provisioner.currentParallel >= provisioner.maxParallel
}

func (provisioner *OneAgentProvisioner) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
	fs := afero.NewOsFs()
	if provisioner.maxParallel > 0 {
		if provisioner.eligibleToParallelDownloads() {
			provisioner.parallelDownloadsLimitReached() {
				log.Info("max parallel downloads reached, requeued")
				return reconcile.Result{RequeueAfter: shortRequeueDuration}, nil
			}
```

src/controllers/csi/provisioner/controller.go
```diff
@@ -76,6 +79,27 @@ func (provisioner *OneAgentProvisioner) SetupWithManager(mgr ctrl.Manager) error
 }
 
 func (provisioner *OneAgentProvisioner) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	fs := afero.NewOsFs()
+	if provisioner.maxParallel > 0 {
+		if provisioner.currentParallel == provisioner.maxParallel {
```
mjgrzybek, line 84:
```suggestion
		if provisioner.currentParallel >= provisioner.maxParallel {
```
a bit more defensive condition :slightly_smiling_face: .

src/controllers/csi/provisioner/controller.go
```diff
@@ -76,6 +79,27 @@ func (provisioner *OneAgentProvisioner) SetupWithManager(mgr ctrl.Manager) error
...
+		}
+		atomic.AddInt64(&provisioner.currentParallel, 1)
+		log.Info("staring parallel download")
+		go func() {
+			_, err := provisioner.reconcile(ctx, request, fs)
+			if err != nil {
+				log.Error(err, "Problem while provisioning oneagents in parallel")
+			}
+			atomic.AddInt64(&provisioner.currentParallel, -1)
+		}()
```
mjgrzybek, line 96: if `provisioner.reconcile()` panicked, `currentParallel` wouldn't decrease :boom:.
```suggestion
		go func() {
			defer atomic.AddInt64(&provisioner.currentParallel, -1)
			_, err := provisioner.reconcile(ctx, request, fs)
			if err != nil {
				log.Error(err, "Problem while provisioning oneagents in parallel")
			}
		}()
```

src/controllers/csi/provisioner/controller.go
```diff
@@ -49,23 +50,25 @@ type OneAgentProvisioner struct {
...
 	db           metadata.Access
 	path         metadata.PathResolver
+
+	currentParallel int64
+	maxParallel     int64
 }
 
 // NewOneAgentProvisioner returns a new OneAgentProvisioner
-func NewOneAgentProvisioner(mgr manager.Manager, opts dtcsi.CSIOptions, db metadata.Access) *OneAgentProvisioner {
+func NewOneAgentProvisioner(mgr manager.Manager, opts dtcsi.CSIOptions, db metadata.Access, maxParallel int64) *OneAgentProvisioner {
```
mjgrzybek, line 62:
```suggestion
func NewOneAgentProvisioner(mgr manager.Manager, opts dtcsi.CSIOptions, db metadata.Access, maxParallelDownloads int64) *OneAgentProvisioner {
```

src/cmd/operator/csi_driver.go
```diff
@@ -95,7 +96,8 @@ func setupCSIDriver(ns string, cfg *rest.Config) (manager.Manager, func(), error
 		return nil, cleanUp, err
 	}
 
-	if err := csiprovisioner.NewOneAgentProvisioner(mgr, csiOpts, access).SetupWithManager(mgr); err != nil {
+	maxParallel, _ := strconv.ParseInt(os.Getenv("MAX_PARALLEL"), 10, 64)
```
mjgrzybek, line 99:
```suggestion
	maxParallel, _ := strconv.ParseInt(os.Getenv("MAX_PARALLEL_DOWNLOADS"), 10, 64)
```

src/cmd/operator/csi_driver.go
```diff
@@ -95,7 +96,8 @@ func setupCSIDriver(ns string, cfg *rest.Config) (manager.Manager, func(), error
 		return nil, cleanUp, err
 	}
 
-	if err := csiprovisioner.NewOneAgentProvisioner(mgr, csiOpts, access).SetupWithManager(mgr); err != nil {
+	maxParallelDownloads, _ := strconv.ParseInt(os.Getenv("MAX_PARALLEL_DOWNLOADS"), 10, 64)
```
mjgrzybek, line 99: Error handling is necessary here (testcase: `MAX_PARALLEL_DOWNLOADS=`:trollface:)

src/controllers/csi/provisioner/controller.go
```diff
@@ -76,6 +83,34 @@ func (provisioner *OneAgentProvisioner) SetupWithManager(mgr ctrl.Manager) error
...
 
 func (provisioner *OneAgentProvisioner) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	fs := afero.NewOsFs()
+	if provisioner.parallelDownloadsEnabled() {
+		if provisioner.parallelDownloadsLimitReached() {
+			log.Info("max parallel downloads reached, requeued")
+			return reconcile.Result{RequeueAfter: shortRequeueDuration}, nil
+		}
+		atomic.AddInt64(&provisioner.currentParallelDownloads, 1)
+		log.Info("staring parallel download")
```
mjgrzybek, line 93:
```suggestion
		log.Info("starting parallel download")
```

src/cmd/operator/csi_driver.go
meik99, line 99: please don't inline strings like these, refactor it into a constant.

meik99: Alongside the error handling I would appreciate sane limits, i.e., min and max values, to handle cases like "-1".
All of this could be handled in a function which reads the env var, parses it into an int, checks the value, enforces the limits and then returns the final value

src/cmd/operator/csi_driver.go
```diff
@@ -95,7 +96,8 @@ func setupCSIDriver(ns string, cfg *rest.Config) (manager.Manager, func(), error
...
+	if err := csiprovisioner.NewOneAgentProvisioner(mgr, csiOpts, access, maxParallelDownloads).SetupWithManager(mgr); err != nil {
```
meik99, line 100: Could `maxParallelDownloads` be merged into the struct behind csiOpts?

src/controllers/csi/provisioner/controller.go
```diff
@@ -49,23 +50,29 @@ type OneAgentProvisioner struct {
...
-		dtcBuildFunc: dynakube.BuildDynatraceClient,
-		fs:           afero.NewOsFs(),
-		recorder:     mgr.GetEventRecorderFor("OneAgentProvisioner"),
-		db:           db,
-		path:         metadata.PathResolver{RootDir: opts.RootDir},
+		client:               mgr.GetClient(),
+		apiReader:            mgr.GetAPIReader(),
+		opts:                 opts,
+		dtcBuildFunc:         dynakube.BuildDynatraceClient,
+		recorder:             mgr.GetEventRecorderFor("OneAgentProvisioner"),
```
meik99, line 72: Please refactor this string also to a constant

src/controllers/csi/provisioner/controller.go
```diff
@@ -76,6 +83,34 @@ func (provisioner *OneAgentProvisioner) SetupWithManager(mgr ctrl.Manager) error
...
+		}
+		atomic.AddInt64(&provisioner.currentParallelDownloads, 1)
+		log.Info("staring parallel download")
+		go func() {
+			defer atomic.AddInt64(&provisioner.currentParallelDownloads, -1)
+			_, err := provisioner.reconcile(ctx, request, fs)
+			if err != nil {
+				log.Error(err, "Problem while provisioning oneagents in parallel")
+			}
+		}()
```
meik99, line 100: Reafctor into a function like and call as `go functionName()`

src/controllers/csi/provisioner/controller.go
```diff
@@ -76,6 +83,34 @@ func (provisioner *OneAgentProvisioner) SetupWithManager(mgr ctrl.Manager) error
...
+}
+
+func (provisioner *OneAgentProvisioner) parallelDownloadsEnabled() bool {
+	return provisioner.maxParallelDownloads > 0
+}
+func (provisioner *OneAgentProvisioner) parallelDownloadsLimitReached() bool {
+	return provisioner.currentParallelDownloads >= provisioner.maxParallelDownloads
+}
+
+func (provisioner *OneAgentProvisioner) reconcile(ctx context.Context, request reconcile.Request, fs afero.Fs) (reconcile.Result, error) {
```
meik99, line 113: Why is the filesystem not part of the provisioner anymore?

meik99 (changes requested):

src/controllers/csi/provisioner/controller.go
0sewa0: `afero` is not thread safe, (I don't know how or for which functions)
so to be safe each goroutine gets its own `fs` object

src/controllers/csi/provisioner/controller.go
```diff
@@ -76,6 +83,34 @@ func (provisioner *OneAgentProvisioner) SetupWithManager(mgr ctrl.Manager) error
...
 func (provisioner *OneAgentProvisioner) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	fs := afero.NewOsFs()
+	if provisioner.parallelDownloadsEnabled() {
+		if provisioner.parallelDownloadsLimitReached() {
+			log.Info("max parallel downloads reached, requeued")
+			return reconcile.Result{RequeueAfter: shortRequeueDuration}, nil
+		}
+		atomic.AddInt64(&provisioner.currentParallelDownloads, 1)
+		log.Info("staring parallel download")
+		go func() {
```
mjgrzybek, line 94: What if reconciliation failed?

0sewa0: we will be sad, but we don't care, the logs will be still there

(this is not meant to be a permanent long-term solution and won't be documented so not everyone will use it)

mjgrzybek: It changes the logic of `OneAgentProvisioner`. [Currently](https://github.com/Dynatrace/dynatrace-operator/pull/656/files/d2fa7a67887e6ec71a4db5032baccfee711ccff1#diff-f2b8b611648fec932e0790d9a1ae99ad30a8ca183edce27761bbe1271eb71b89L138) it returns an `error` so that request is re-queued to be processed again.
I'd suggest to preserve this logic by adding an atomic bool which indicates if error occurred in any of goroutines.
If that bool==true, return `reconcile.Result{}, errors.New("parallel download failed")`

mjgrzybek, line 94: What if download is still in progress and another reconciliation started?
Add `sync.WorkGroup` to wait for downloads to be finished before exiting reconciliation loop.

0sewa0: that is not how it works 😅

0sewa0: we would need a sync group per dynakube, and not block but requeue them

mjgrzybek: If I understand it correctly, `provisioner.currentParallelDownloads` is a state shared between reconciler loops, so it may happen that two downloads for single tenant are in progress :thinking:.

mjgrzybek: `Reconcile()` should not exit until downloads are done.

0sewa0: we would need to block(which we can't) OR log the error in another reconcile, but because we are running stuff in parallel we might override that error, or need to store multiple of these bools.

this shouldn't be over-engineered

src/controllers/csi/provisioner/controller.go
```diff
@@ -49,33 +53,62 @@ type OneAgentProvisioner struct {
 	apiReader    client.Reader
 	opts         dtcsi.CSIOptions
 	dtcBuildFunc dynakube.DynatraceClientFunc
-	fs           afero.Fs
 	recorder     record.EventRecorder
 	db           metadata.Access
 	path         metadata.PathResolver
+
+	mu                       *sync.Mutex
+	currentParallelDownloads int64
```
0sewa0, line 61: maybe this is redundant now ? 
could just check the len of the map

src/controllers/csi/provisioner/controller.go
```diff
@@ -89,72 +122,110 @@ func (provisioner *OneAgentProvisioner) Reconcile(ctx context.Context, request r
...
 		return reconcile.Result{RequeueAfter: shortRequeueDuration}, nil
 	}
 
+	if provisioner.parallelDownloadsEnabled() {
+		if provisioner.parallelDownloadsLimitReached() {
+			log.Info("max parallel downloads reached, requeued")
+			return reconcile.Result{RequeueAfter: shortRequeueDuration}, nil
+		}
+		if provisioner.isCurrentlyDownloading(tenant) {
+			log.Info("still downloading", "tenant", tenant)
```
mjgrzybek, line 137: nit: it should be able to cancel download it it's stalled using `context.Context`, but that should be done in another PR

src/controllers/csi/provisioner/controller.go
```diff
@@ -49,33 +53,62 @@ type OneAgentProvisioner struct {
...
+		dtcBuildFunc:         dynakube.BuildDynatraceClient,
+		recorder:             mgr.GetEventRecorderFor(eventRecorderName),
+		db:                   db,
+		path:                 metadata.PathResolver{RootDir: opts.RootDir},
+		mu:                   &sync.Mutex{},
+		currentlyDownloading: map[string]bool{},
 	}
 }
 
+func (provisioner *OneAgentProvisioner) setCurrentlyDownloading(name string) {
```
mjgrzybek, line 83:
```suggestion
func (provisioner *OneAgentProvisioner) setCurrentlyDownloading(tenant string) {
```

src/controllers/csi/provisioner/controller.go
0sewa0: that would require that the download would care for that context but it doesn't (that part of the code was written before my time sooo 😬 )

src/controllers/csi/provisioner/controller.go
```diff
@@ -49,33 +53,62 @@ type OneAgentProvisioner struct {
...
-		path:         metadata.PathResolver{RootDir: opts.RootDir},
+		client:               mgr.GetClient(),
+		apiReader:            mgr.GetAPIReader(),
+		opts:                 opts,
+		dtcBuildFunc:         dynakube.BuildDynatraceClient,
+		recorder:             mgr.GetEventRecorderFor(eventRecorderName),
+		db:                   db,
+		path:                 metadata.PathResolver{RootDir: opts.RootDir},
+		mu:                   &sync.Mutex{},
+		currentlyDownloading: map[string]bool{},
```
mjgrzybek, line 79: Map's values are not needed here. We need a Set, which can be created this way:
```suggestion
		currentlyDownloading: map[string]interface{},
```

src/controllers/csi/provisioner/controller.go
mjgrzybek: see: https://github.com/Dynatrace/dynatrace-operator/pull/656#discussion_r831995831

src/controllers/csi/provisioner/controller.go
0sewa0: empty interface gives us nothing, just longer lines of code

also the correct way is `currentlyDownloading: map[string]interface{}{}` which is ugly as sin 😬

0sewa0: we don't use a loop(I delete stuff now), and first result for `golang set` from Google is using a map of bools, so the internet is with me 😛

mjgrzybek: your google search results are profiled up to your needs :stuck_out_tongue:

mjgrzybek: uh, sorry, it's `struct{}`, not `interface{}`!
see: https://www.davidkaya.com/sets-in-golang/

0sewa0: that is the 4. result, therefore I still prefer bool 😬

src/controllers/csi/provisioner/controller.go
```diff
@@ -57,9 +57,8 @@ type OneAgentProvisioner struct {
 	db           metadata.Access
 	path         metadata.PathResolver
 
-	mu                       *sync.Mutex
-	currentParallelDownloads int64
-	currentlyDownloading     map[string]bool
+	mu                   *sync.Mutex
+	currentlyReconciling *kubeobjects.StringSet
```
mjgrzybek, line 61: Why is `StringSet` in `kubeobjects`?

0sewa0: I'm open for suggestion 😄

mjgrzybek: `utils` ofc. :smile:

mjgrzybek: `collections` maybe?

0sewa0: oh no 😨

meik99: So, a map which only purpose is to collect keys to more easily remove them, since golang has no `remove` function for elements in arrays, now has it's own type which, in turn, does nothing but exist?

That seems highly unnecessary and IMO it should just be `map[string]bool`.

Also, please rename `mu` to `mutex`-

0sewa0: 🍿

0sewa0: have it in `dtcsi` and call it `TenantsSet` ? would that work ?
(I like the way it simplifies the code of the provisioner even if just a bit, and the logic is simple so I would keep it)

src/controllers/csi/provisioner/controller.go
```diff
@@ -49,38 +53,72 @@ type OneAgentProvisioner struct {
 	apiReader    client.Reader
 	opts         dtcsi.CSIOptions
 	dtcBuildFunc dynakube.DynatraceClientFunc
-	fs           afero.Fs
 	recorder     record.EventRecorder
 	db           metadata.Access
 	path         metadata.PathResolver
+
+	mu                   *sync.Mutex
```
meik99, line 60:
```suggestion
	mutex                   *sync.Mutex
```

meik99 (changes requested):

mjgrzybek (approved):

PR #661, opened 2022-03-23
ActiveGate webserver's CA mounted properly in OneAgents pods

mjgrzybek:
# Description
OneAgent expects ActiveGate CA cert to be mounted here: `/mnt/dynatrace/certs/activegate/`

## How can this be tested?
1. Create dynakubre `cloudNativeFullstack` monitoring enabled
2. `kubectl patch -ndynatrace dynakubes.dynatrace.com dynakube --type=json -p='[{"op":"replace", "path":"/spec/activeGate
/tlsSecretName", "value":"dk-certs"}]'`
3. OneAgent pods should be in Running state eventually

## Checklist

### DIFF
--- a/src/api/v1beta1/properties.go
+++ b/src/api/v1beta1/properties.go
@@ -94,7 +94,7 @@ func (dk *DynaKube) NeedsStatsd() bool {
 	return dk.IsActiveGateMode(StatsdIngestCapability.DisplayName)
 }
 
-func (dk *DynaKube) HasActiveGateTLS() bool {
+func (dk *DynaKube) HasActiveGateCaCert() bool {
 	return dk.ActiveGateMode() && dk.Spec.ActiveGate.TlsSecretName != ""
 }
 
--- a/src/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/src/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -24,11 +24,11 @@ const (
 	hostRootVolumeName  = "host-root"
 	hostRootVolumeMount = "/mnt/root"
 
-	trustedCaVolumeName  = "dynatrace-cluster-ca"
-	trustedCaVolumeMount = "/mnt/dynatrace/certs"
+	trustedCaCertVolumeName  = "dynatrace-cluster-ca-cert"
+	trustedCaCertVolumeMount = "/mnt/dynatrace/certs"
 
-	agCaVolumeName  = "active-gate-ca"
-	agCaVolumeMount = "/mnt/dynatrace/certs/activegate/"
+	activeGateCaCertVolumeName  = "active-gate-ca"
+	activeGateCaCertVolumeMount = "/mnt/dynatrace/certs/activegate/"
 
 	csiStorageVolumeName  = "osagent-storage"
 	csiStorageVolumeMount = "/mnt/volume_storage_mount"
--- a/src/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/src/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -21,14 +21,14 @@ const (
 	// normal oneagent shutdown scenario with some extra time
 	defaultTerminationGracePeriod = 80
 
-	hostRootVolumeName  = "host-root"
-	hostRootVolumeMount = "/mnt/root"
+	hostRootVolumeName      = "host-root"
+	hostRootVolumeMountPath = "/mnt/root"
 
-	trustedCaCertVolumeName  = "dynatrace-cluster-ca-cert"
-	trustedCaCertVolumeMount = "/mnt/dynatrace/certs"
+	clusterCaCertVolumeName      = "dynatrace-cluster-ca"
+	clusterCaCertVolumeMountPath = "/mnt/dynatrace/certs"
 
-	activeGateCaCertVolumeName  = "active-gate-ca"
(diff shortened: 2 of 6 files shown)


src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -26,11 +24,11 @@ const (
...
 
-	certVolumeName  = "certs"
-	certVolumeMount = "/mnt/dynatrace/certs"
+	trustedCaVolumeName  = "dynatrace-cluster-ca"
+	trustedCaVolumeMount = "/mnt/dynatrace/certs"
 
-	OneAgentCustomKeysPath = "/var/lib/dynatrace/oneagent/agent/customkeys"
-	tlsVolumeName          = "tls"
+	agCaVolumeName  = "active-gate-ca"
+	agCaVolumeMount = "/mnt/dynatrace/certs/activegate/"
```
0sewa0, line 31:
```suggestion
	activeGateCertVolumeMount = "/mnt/dynatrace/certs/activegate/"
```

src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -26,11 +24,11 @@ const (
 	hostRootVolumeName  = "host-root"
 	hostRootVolumeMount = "/mnt/root"
 
-	certVolumeName  = "certs"
-	certVolumeMount = "/mnt/dynatrace/certs"
+	trustedCaVolumeName  = "dynatrace-cluster-ca"
```
meik99, line 27:
```suggestion
	trustedCertVolumeName  = "dynatrace-cluster-cert"
```
I'd suggest using `cert` instead of `ca` in every case since it has more meaning

src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -26,11 +24,11 @@ const (
...
+	trustedCaVolumeMount = "/mnt/dynatrace/certs"
 
-	OneAgentCustomKeysPath = "/var/lib/dynatrace/oneagent/agent/customkeys"
-	tlsVolumeName          = "tls"
+	agCaVolumeName  = "active-gate-ca"
```
meik99, line 30:
```suggestion
	activeGateCertVolumeName  = "active-gate-cert"
```

0sewa0:
what happens if we are not running in read-only mode ? 
shouldn't we use the "old" path or this path works in that case aswell ?

src/webhook/mutation/pod_mutator.go
```diff
@@ -36,6 +35,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
+const OneAgentCustomKeysPath = "/var/lib/dynatrace/oneagent/agent/customkeys"
```
meik99, line 38: Is this constant used anywhere else? otherwise it does not need to be exported

meik99 (changes requested):

src/controllers/dynakube/oneagent/daemonset/daemonset.go
mjgrzybek: "cert" is more generic term, CA is a specific cert :thinking:.
I'll change it to CaCert.

mjgrzybek:
> what happens if we are not running in read-only mode ? shouldn't we use the "old" path or this path works in that case aswell ?

depends on OneAgent version and this is why it's not yet backported to `release-0.5` branch. Let's discuss it offline.

src/controllers/dynakube/oneagent/daemonset/daemonset.go
mjgrzybek: `javaProblemFactoryBuilder`, but ok, I'll follow our [CONTRIBUTING.md](https://github.com/Dynatrace/dynatrace-operator/blob/master/CONTRIBUTING.md) rules :)

src/api/v1beta1/properties.go
```diff
@@ -94,7 +94,7 @@ func (dk *DynaKube) NeedsStatsd() bool {
 	return dk.IsActiveGateMode(StatsdIngestCapability.DisplayName)
 }
 
-func (dk *DynaKube) HasActiveGateTLS() bool {
+func (dk *DynaKube) HasActiveGateCaCert() bool {
```
luhi-DT, line 97: Isn't it a bit confusing to call the field TLSSecret in the CR and the function `HasActiveGateCaCert()`?

src/webhook/mutation/pod_mutator.go
```diff
@@ -36,6 +35,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
+const oneAgentCustomKeysPath = "/var/lib/dynatrace/oneagent/agent/customkeys"
```
luhi-DT, line 38: Can we avoid duplication of that variable?

luhi-DT (dismissed):
When should we merge those changes? When the Agent has the backport to look on both location?

src/api/v1beta1/properties.go
mjgrzybek: It's named wrong, `TLS != CA`. 
It's way more confusing (for me) to understand source code before this change, especially because we have multiple instances of TLS protocol related objects, e.g. CAs or private keys.
I agree naming in CRD shouldn't be changed, though.

PR #662, opened 2022-03-23
Standardized CA certificates handling across components

mjgrzybek:
# Description

OneAgent uses CA certificates when connecting to Dynatrace Cluster and ActiveGate. These are separate certificates.
Goal of this PR is to make it less confusing for developer to distinguish which certs one is dealing with.

`trusted CA` -> `cluster CA`
`ActiveGate TLS` -> `ActiveGate CA`

`dtclient`:
`trusted CA` -> `CA`

## How can this be tested?
Play with certificates.

## Checklist

### DIFF
--- a/src/api/v1beta1/dynakube_types.go
+++ b/src/api/v1beta1/dynakube_types.go
@@ -104,7 +104,7 @@ type DynaKubeSpec struct {
 	// This property only affects certificates used to communicate with the Dynatrace API.
 	// The property is not applied to the ActiveGate
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Trusted CAs",order=6,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:io.kubernetes:ConfigMap"}
-	TrustedCAs string `json:"trustedCAs,omitempty"`
+	ClusterCa string `json:"trustedCAs,omitempty"`
 
 	// Optional: Sets Network Zone for OneAgent and ActiveGate pods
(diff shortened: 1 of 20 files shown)


src/api/v1beta1/dynakube_types.go
```diff
@@ -104,7 +104,7 @@ type DynaKubeSpec struct {
 	// This property only affects certificates used to communicate with the Dynatrace API.
 	// The property is not applied to the ActiveGate
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Trusted CAs",order=6,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:io.kubernetes:ConfigMap"}
-	TrustedCAs string `json:"trustedCAs,omitempty"`
+	ClusterCa string `json:"trustedCAs,omitempty"`
```
0sewa0, line 107: you can't just rename stuff in the CRD, there is a conversion webhook for alpha to beta, but this would cause a problem and would require a new conversion logic, which we don't want to do right now, and especially for the rename of a single filed

luhi-DT:
Please provide a detailed guide or at least something more meaningful than "Play with certificates" in the "How can this be tested section" :) This way we make sure, that everybody can test it, even though he has no knowledge of how to set certificates up 😄

src/api/v1beta1/dynakube_types.go
mjgrzybek: json annotations are preserved. CRDs aren't affected.

0sewa0: that is still not the most optimal way, as that is going to confuse us in a week or two after the change 😅

PR #664, opened 2022-03-24
Mark for termination Event was not sent correctly, because of wrong labels on Daemonset

luhi-DT:
## Description

We currently have the problem, that the mark for termination event is not send, because our node cache is not filled with the nodes, where the oneAgent is deployed on. To understand how the node cache is filled, you need to know that we store the instances of a oneAgent pod in the Dynakube-Status. Every 5 Minutes if the DK is reconciled, we query the cluster for oneAgent pods with defined labels on it. If those labels are not matching, then the pods are not received. 

The case was now, that we had a label missmatch and the pods were not correctly stored in the DK Status. Therfore we were not able to query the DK for a node, when a node got reconciled. As a result, the mark for termination Event never got sent.

Why did this happen?
The labels which are set on the daemonset (in detail the feature label) got changed, from the const which was part of the daemonset.go to deploymentmetadata.(feature). The had a slight difference, e.g classic (old label) and classic_fullstack (new label)

## How can this be tested?
To test if everything works no as expected, you can create a DK with classicFullstack, cloudNative or hostMonitoring and check if the Dynakube Object has set the instances of the OneAgent pods correctly. Furthermore to test if the mark for termination event is sent correctly, you can cordon a Node in GKE for example and check if the host has a label in the Dynatrace UI, that it was marked for termination (or check the Operator Logs in the cluster and search for "Sending mark for termination")

## Checklist

### DIFF
--- a/src/controllers/dynakube/dynakube_controller.go
+++ b/src/controllers/dynakube/dynakube_controller.go
@@ -17,9 +17,9 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/dtversion"
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/istio"
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/oneagent"
+	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/oneagent/daemonset"
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/status"
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/updates"
-	"github.com/Dynatrace/dynatrace-operator/src/deploymentmetadata"
 	"github.com/Dynatrace/dynatrace-operator/src/dtclient"
 	dtingestendpoint "github.com/Dynatrace/dynatrace-operator/src/ingestendpoint"
 	"github.com/Dynatrace/dynatrace-operator/src/initgeneration"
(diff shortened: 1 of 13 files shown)


src/controllers/dynakube/dynakube_controller.go
```diff
@@ -215,21 +215,21 @@ func (controller *DynakubeController) reconcileDynaKube(ctx context.Context, dkS
 
 	if dkState.Instance.HostMonitoringMode() {
 		upd, err = oneagent.NewOneAgentReconciler(
-			controller.client, controller.apiReader, controller.scheme, dkState.Instance, daemonset.HostMonitoringFeature,
+			controller.client, controller.apiReader, controller.scheme, dkState.Instance, deploymentmetadata.DeploymentTypeHostMonitoring,
```
chrismuellner, line 218: I like the reuse of the constant, but I think it should be the other way around.
`deploymentmetadata` should use the constants defined in `oneagent/daemonset/config.go` (or somewhere around the oneagent reconciler)

chrismuellner (dismissed):

src/controllers/dynakube/dynakube_controller.go
luhi-DT: We can not do that, as deploymentmetadata reports to ClusterDataHub, where those constants have expected values, if we change the value of them, I guess our reporting to ClusterDataHub is not working anymore

chrismuellner: the constants would have the same values, just be in a different location?
I'd like the `deploymentmetadata` package to reuse constants that exist in the appropriate packages and not have `deploymentmetadata` spread around the whole code base.

luhi-DT: Moving them out of `deploymentmetadata` would tie them to the package they are moved into, but actually those consts are tied to the deploymentmetadata strongly as a change in them would result in loosing reports ind CDH

meik99:
This opportunity should be used to add an unit tests which checks for this bug, to catch it if labels are changed in the future.
as far as I understand, this unit test should call the `NodesController::Reconcile` function and check that a set of oneagent pods has been added to the cache. A second reconile should be able to update the cache accordingly

0sewa0 (approved):
it works, 
it could still use the unit-test Michael mentioned

meik99 (changes requested):
> it could still use the unit-test Michael mentioned
It must use the unit-test Michael mentioned

0sewa0 (approved):

meik99 (approved):

PR #665, opened 2022-03-25
Add default limits to sample yamls

gkrenn:
# Description

Updating our sample files to include default resources for activate gate containers.

## How can this be tested?

Apply a sample file and check if resources are set on the active gate containers.

## Checklist

### DIFF
--- a/config/samples/applicationMonitoring.yaml
+++ b/config/samples/applicationMonitoring.yaml
@@ -105,9 +105,6 @@ spec:
 
     # Optional: resource settings for ActiveGate container. Consumption of the ActiveGate heavily depends
     # on the workload to monitor; please adjust values accordingly.
-    #
-    # Suggested starting point for resources below:
-    #
     resources:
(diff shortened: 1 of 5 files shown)


luhi-DT:
Should we add a comment, that these limits might be customized depending on the cluster?

config/samples/applicationMonitoring.yaml
```diff
@@ -108,13 +108,13 @@ spec:
     #
     # Suggested starting point for resources below:
```
chrismuellner, line 109: I'd remove this comment, the description above already mentions that these values should be adjusted.
It's not really a suggestion anymore when it's there by default

gkrenn:
Removed the comment now, but a hint to adjust the values is still there: "Consumption of the ActiveGate heavily depends on the workload to monitor; please adjust values accordingly."

chrismuellner (approved):

PR #670, opened 2022-03-26
Remove ExportData token permission if host-requests are disabled

ernstvonoelsen:
…s-requests is true.

# Description

summary of the change:
This PR validates the APITokens' scopes in case that a Dynakube instance is annotated with the `alpha.operator.dynatrace.com/feature-disable-hosts-requests: "true"` even if there is no `ExportData` permission (since this permission is never used in this case).

relevant motivation and context
Dynatrace Operator will be used in our enterprise landscape by several hundred teams to monitor their K8s clusters in a single-tenant Dynatrace Environment. Due to strict data privacy rules, we cannot allow that regular users get such broad read access on v1 API of Dynatrace as the `ExportData` permission scope would provide.
From my analysis of the code, the `ExportData` permission of the provided APIToken is never really required if
automatic kubernetes api monitoring is not enabled and the annotation `alpha.operator.dynatrace.com/feature-disable-hosts-requests: "true"` on the dynakube instance is present.

Since such a least-privilege approach is feasible without losing Dynatrace's monitoring capabilities (one only looses some comfort imho), I suggest to enable it also in code, although not making it the default.

## How can this be tested?
What environment is necessary for the change to be noticeable ?
What needs to be enabled/applied for the change to be noticeable ?

- a `dynakube` secret with an APIToken that grants nothing else than `InstallerDownload` permissions
- the annotation `feature.dynatrace.com/disable-hosts-requests: "true"` on the dynakube instance
- instead of getting a `missing scopes` error on startup, the dynakube instance is spawned without any complaint

## Checklist


src/controllers/dynakube/dtclient_reconciler.go
```diff
@@ -112,14 +112,15 @@ func (r *DynatraceClientReconciler) Reconcile(ctx context.Context, instance *dyn
...
-			},
+			Type:      dynatracev1beta1.APITokenConditionType,
+			Key:       dtclient.DynatraceApiToken,
+			Value:     r.ApiToken,
+			Scopes:    []string{dtclient.TokenScopeInstallerDownload},
 			Timestamp: &r.status.LastAPITokenProbeTimestamp,
 		}}
+		if !instance.FeatureDisableHostsRequests() {
+			tokens[0].Scopes = append(tokens[0].Scopes, dtclient.TokenScopeDataExport)
+		}
```
meik99, line 123: This change would not work if both tokens are provided, since `r.PaasToken` would not be empty.

I'd suggest moving these lines to what is currently line 142, i.e. before `if instance.KubernetesMonitoringMode() &&
		instance.FeatureAutomaticKubernetesApiMonitoring() {` and removing the scope from the first token in the else-branch, i.e. line 131,  to solve this issue.
Please also add an unit test to check for this case.

Otherwise, lgtm

meik99 (changes requested):

src/controllers/dynakube/dtclient_reconciler.go
ernstvonoelsen: done :heavy_check_mark:

meik99 (approved):

ernstvonoelsen:
Hi @meik99, 
what are we missing?

meik99 (dismissed):
I was just confused about an internal process, you're good

ernstvonoelsen:
Hi @chrismuellner thank you for merging the latest `master` branch.

On which end do we have to fix the failing tests? Locally they are green, don't know where the difference stems from?

chrismuellner:
1. You can fix the linting issues by running `make lint`
2. Tests seem to be failing for `dtclient_reconciler_test` which you modified. I'm assuming your logic changed the behavior for other tests. How are you running your tests? Running `go test ./...` should show what tests are failing.

chrismuellner:
Thanks! Just making sure the Actions run through before merging.

PR #672, opened 2022-03-28
Simplifies the matchLabels for Oneagent daemonset

0sewa0:
# Description

The match labels shouldn't change between deployment modes or the operator will throw errors in case the mode is changed on the dynakube and the oneagent daemonset will not change unless deleted by hand.

## How can this be tested?
1. Deploy the operator
2. Deploy a dynakube with a mode that creates a oneagent daemonset
3. Update the dynakube to switch to a different mode

The labels should change, but not the matchLabels

## Checklist

### DIFF
--- a/src/controllers/csi/gc/logs_test.go
+++ b/src/controllers/csi/gc/logs_test.go
@@ -111,7 +111,7 @@ func TestLogGarbageCollector_modificationDateOlderThanTwoWeeks(t *testing.T) {
 	})
 
 	t.Run("is true for timestamp 14 days in past", func(t *testing.T) {
-		isOlder := isOlderThanTwoWeeks(time.Now().AddDate(0, 0, -14))
+		isOlder := isOlderThanTwoWeeks(time.Now().AddDate(0, 0, -15))
 
 		assert.True(t, isOlder)
 	})
--- a/src/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/src/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -143,7 +143,7 @@ func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 	instance := dsInfo.instance
 	podSpec := dsInfo.podSpec()
 	labels := kubeobjects.MergeLabels(
-		BuildLabels(dsInfo.instance.Name, dsInfo.deploymentType),
+		dsInfo.buildLabels(),
 		dsInfo.hostInjectSpec.Labels,
 	)
 	maxUnavailable := intstr.FromInt(instance.FeatureOneAgentMaxUnavailable())
@@ -161,7 +161,7 @@ func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 		},
 		Spec: appsv1.DaemonSetSpec{
 			Selector: &metav1.LabelSelector{
-				MatchLabels: buildMatchLabels(dsInfo.instance.Name),
+				MatchLabels: dsInfo.buildMatchLabels(),
 			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
@@ -179,6 +179,21 @@ func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 	}
 
 	return result, nil
+}
+
+func (dsInfo *builderInfo) buildLabels() map[string]string {
+	return BuildLabels(dsInfo.instance.Name, dsInfo.deploymentType)
+}
+
+// buildMatchLabels produces a set of labels that
+// don't change when switching between oneagent modes
+// or during operator version update
+// as matchLabels are not mutable on a Daemonset
+func (dsInfo *builderInfo) buildMatchLabels() map[string]string {
+	labels := dsInfo.buildLabels()
+	delete(labels, kubeobjects.AppVersionLabel)
+	delete(labels, kubeobjects.FeatureLabel)
+	return labels
 }
 
 func (dsInfo *builderInfo) podSpec() corev1.PodSpec {
--- a/src/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/src/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -142,7 +142,10 @@ func appendHostIdArgument(result *appsv1.DaemonSet, source string) {
 func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 	instance := dsInfo.instance
 	podSpec := dsInfo.podSpec()
-	labels := kubeobjects.MergeLabels(dsInfo.buildLabels(), dsInfo.hostInjectSpec.Labels)
+	labels := kubeobjects.MergeLabels(
+		BuildLabels(dsInfo.instance.Name, dsInfo.deploymentType),
+		dsInfo.hostInjectSpec.Labels,
+	)
 	maxUnavailable := intstr.FromInt(instance.FeatureOneAgentMaxUnavailable())
 	annotations := map[string]string{
 		annotationVersion:      instance.Status.OneAgent.Version,
@@ -158,7 +161,7 @@ func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 		},
 		Spec: appsv1.DaemonSetSpec{
 			Selector: &metav1.LabelSelector{
-				MatchLabels: dsInfo.buildLabels(),
+				MatchLabels: buildMatchLabels(dsInfo.instance.Name),
 			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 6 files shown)


src/controllers/csi/gc/logs_test.go
```diff
@@ -111,7 +111,7 @@ func TestLogGarbageCollector_modificationDateOlderThanTwoWeeks(t *testing.T) {
 	})
 
 	t.Run("is true for timestamp 14 days in past", func(t *testing.T) {
-		isOlder := isOlderThanTwoWeeks(time.Now().AddDate(0, 0, -14))
+		isOlder := isOlderThanTwoWeeks(time.Now().AddDate(0, 0, -15))
```
chrismuellner, line 114: should this be in this PR?

src/controllers/dynakube/oneagent/daemonset/labels.go
```diff
@@ -2,9 +2,14 @@ package daemonset
 
 import "github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
 
-// buildLabels returns generic labels based on the name given for a Dynatrace OneAgent
```
chrismuellner, line 5: we should have more comments, not fewer 😄

src/controllers/dynakube/oneagent/daemonset/labels.go
```diff
@@ -2,9 +2,14 @@ package daemonset
...
 func BuildLabels(name string, feature string) map[string]string {
 	labels := kubeobjects.CommonLabels(name, kubeobjects.OneAgentComponentLabel)
 	labels[kubeobjects.FeatureLabel] = feature
 	return labels
 }
+
+func buildMatchLabels(name string) map[string]string {
+	labels := kubeobjects.CommonLabels(name, kubeobjects.OneAgentComponentLabel)
+	delete(labels, kubeobjects.AppVersionLabel)
```
chrismuellner, line 13: how does that make sense?
that would mean if we upgrade the operator outdated pods would match?

0sewa0: its what we had before,
we never had the version or anything interesting for matchLabels,
and its not important, the operator would update the will the daemonset because the labels changed due to the version, the matchLabels not changing is fine

we could just always delete the previous daemonset, but then there would be a "not nice" period if while both daemonsets exists(1 terminating 1 crashlooping then running)
also I don't know how this would work in case of readonly (we would want to mount the same thing but it might be still used)

src/controllers/dynakube/oneagent/daemonset/labels.go
meik99: the best comment is no comment.
if code needs commenting, most of the time the code needs refactoring.
additionally, each comment needs maintenance the same way code needs maintenance. since comments are mostly overlooked, they tend to contain misinformation after some time. so in the end, a developer has to read the code again anyway.

src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -158,7 +161,7 @@ func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 		},
 		Spec: appsv1.DaemonSetSpec{
 			Selector: &metav1.LabelSelector{
-				MatchLabels: dsInfo.buildLabels(),
+				MatchLabels: buildMatchLabels(dsInfo.instance.Name),
```
meik99, line 164:
```suggestion
				MatchLabels: dsInfo.buildMatchLabels(),
```

src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -142,7 +142,10 @@ func appendHostIdArgument(result *appsv1.DaemonSet, source string) {
 func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 	instance := dsInfo.instance
 	podSpec := dsInfo.podSpec()
-	labels := kubeobjects.MergeLabels(dsInfo.buildLabels(), dsInfo.hostInjectSpec.Labels)
+	labels := kubeobjects.MergeLabels(
+		BuildLabels(dsInfo.instance.Name, dsInfo.deploymentType),
```
meik99, line 146: I'd wrap `BuildLabels(dsInfo.instance.Name, dsInfo.deploymentType)` in `dsInfo.BuildLables()`
```suggestion
		dsInfo.BuildLabels(),
```

looks nicer

src/controllers/dynakube/oneagent/daemonset/labels.go
chrismuellner: > the best comment is no comment.
I disagree. We should at least have a basic understanding of what each function does without having to look up all the different places it's used.
Instead of removing the comment because it's outdated, updating it when the functionality is changed is a better long term solution.

meik99 (changes requested):

src/controllers/dynakube/oneagent/daemonset/labels.go
0sewa0: the best comment is no comment == the best is when no comment is needed

complex functions that are not straight-forward should have some explanation
but `BuildLabels` in the `daemonset` package is rather straight forward, and the only thing I could comment is the actual list of labels that I'm creating which would be just repeating the code but in a different place, which is just asking to get outdated.

good comments answer the question "why are we doing this" and not the "what are we doing"

(if you have to ask "what are we doing" == the code is bad in general )

src/controllers/csi/gc/logs_test.go
0sewa0: creating a separate PR for it would be to much work at this point IMO

src/kubeobjects/daemonset_test.go
```diff
@@ -12,95 +12,74 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const (
```
0sewa0, line 15: the diff looks scary, but I just refactored this according to the guidelines

src/kubeobjects/daemonset_test.go
```diff
@@ -12,95 +12,74 @@ import (
...
-	assert.NotNil(t, dsAfter.Annotations)
-	annotation, ok := dsAfter.Annotations[AnnotationHash]
-	assert.True(t, ok)
-	assert.Equal(t, "new", annotation)
-}
+		updated, err := CreateOrUpdateDaemonSet(fakeClient, log, &oldDaemonSet)
+		require.NoError(t, err)
+		assert.False(t, updated)
+	})
+	t.Run("recreate when exists and changed for immutable field", func(t *testing.T) {
```
0sewa0, line 51: this test case was meant to actually test the change I introduced, 
but unfortunately the client created by `fake.NewClient` doesn't checks for this specific issue, 
so this passes with and without my changes 

should we keep it or not ?

src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -178,6 +181,21 @@ func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
...
 }
 
+func (dsInfo *builderInfo) buildLabels() map[string]string {
+	return BuildLabels(dsInfo.instance.Name, dsInfo.deploymentType)
+}
+
+// buildMatchLabels produces a set of labels that
+// don't change when switching between oneagent modes
+// or during operator version update
+// as matchLabels are not mutable on a Daemonset
```
meik99, line 191: I like this comment, it is good, well written, and useful

src/kubeobjects/daemonset.go
```diff
@@ -23,6 +24,17 @@ func CreateOrUpdateDaemonSet(c client.Client, logger logr.Logger, desiredDs *app
 		return false, nil
 	}
 
+	if !reflect.DeepEqual(currentDs.Spec.Selector.MatchLabels, desiredDs.Spec.Selector.MatchLabels) {
```
meik99, line 27:
```suggestion
	if matchLabelsChanged(currentDs, desiredDs) {
	
	...
	
	func matchLabelsChanged(currentDs, desiredDs) bool {
	    return !reflect.DeepEqual(currentDs.Spec.Selector.MatchLabels, desiredDs.Spec.Selector.MatchLabels)
	}
```

src/kubeobjects/daemonset.go
```diff
@@ -23,6 +24,17 @@ func CreateOrUpdateDaemonSet(c client.Client, logger logr.Logger, desiredDs *app
...
+		logger.Info("immutable section changed on daemonset, deleting and recreating", "name", desiredDs.Name)
+		err = c.Delete(context.TODO(), currentDs)
+		if err != nil {
+			return false, err
+		}
+		logger.Info("deleted daemonset")
+		logger.Info("recreating daemonset", "name", desiredDs.Name)
+		return true, c.Create(context.TODO(), desiredDs)
```
meik99, line 35: refactor into function to keep the parent function more concise and readable.
also, `c` should be changed to a name

src/kubeobjects/daemonset_test.go
meik99: Does it also pass if a fault is introduced in your changes?
For example, if you introduced a trivial problem like dividing by zero in your change?

If the test breaks with an intentionally introduced fault, it validates your logic and should be kept.
Otherwise, the test is not necessary.

src/kubeobjects/daemonset_test.go
meik99: nicely refactored

meik99 (changes requested):

src/controllers/csi/gc/logs_test.go
meik99: I don't quite get what the purpose of this change is, but I agree with @0sewa0, it's such a trivial change it doesn't really need its own PR. It's more like housekeeping

meik99 (approved):

PR #678, opened 2022-03-29
Prunes the makefile

0sewa0:
# Description
Renames/refactors/prunes things in the makefile.

Removed commands that didn't work (properly) (or didn't make sense)
Introduced new targets instead of using `if`s and envvars
Renamed targets that were too vague, inaccurate or unhelpful

Didn't touch the image setting/switching logic. (I couldn't come up with a better solution)

## How can this be tested?
Test out the commands, see if they work.
If removed one of your favourite command, pls comment which one was it and why do you use it 😄   

## Checklist


Makefile
```diff
@@ -74,43 +70,35 @@ run: export POD_NAMESPACE=dynatrace
...
-deploy-ocp: manifests kustomize
+deploy-ocp: manifests-ocp kustomize
 	oc get project dynatrace || oc adm new-project --node-selector="" dynatrace
 	cd config/deploy/openshift && $(KUSTOMIZE) edit set image "quay.io/dynatrace/dynatrace-operator:snapshot"=$(BRANCH_IMAGE)
 	$(KUSTOMIZE) build config/deploy/openshift | oc apply -f -
-	make reset-kustomization-files
 
-deploy-local:
-	./build/deploy_local.sh
+publish-image:
```
meik99, line 91:
```suggestion
push-image:
```

publish sounds like releasing something

Makefile
```diff
@@ -74,43 +70,35 @@ run: export POD_NAMESPACE=dynatrace
...
+	./build/publish_image.sh
 
-deploy-local-easy: export TAG=snapshot-$(shell git branch --show-current | sed "s/[^a-zA-Z0-9_-]/-/g")
-deploy-local-easy:
-	./build/deploy_local.sh
+publish-tagged-image: export TAG=snapshot-$(shell git branch --show-current | sed "s/[^a-zA-Z0-9_-]/-/g")
```
meik99, line 94:
```suggestion
push-tagged-image: export TAG=snapshot-$(shell git branch --show-current | sed "s/[^a-zA-Z0-9_-]/-/g")
```

publish sounds like releasing something

meik99 (changes requested):

Makefile
```diff
@@ -74,43 +70,35 @@ run: export POD_NAMESPACE=dynatrace
 run: generate fmt vet manifests
```
meik99, line 70:
```suggestion
run: generate-crd fmt vet manifests
```

Makefile
```diff
@@ -268,15 +222,12 @@ SERVICE_ACCOUNTS+=--extra-service-accounts dynatrace-dynakube-oneagent-unprivile
...
-bundle-olm: manifests kustomize
+.PHONY: bundle
+bundle: OLM=true
+bundle: manifests kustomize
 	operator-sdk generate kustomize manifests -q --apis-dir ./src/api/
 	cd config/deploy/$(PLATFORM) && $(KUSTOMIZE) edit set image "quay.io/dynatrace/dynatrace-operator:snapshot"=$(OLM_IMAGE)
 	$(KUSTOMIZE) build config/olm/$(PLATFORM) | operator-sdk generate bundle --overwrite --version $(VERSION) $(SERVICE_ACCOUNTS) $(BUNDLE_METADATA_OPTS)
-	make OUT=all reset-kustomization-files
 	operator-sdk bundle validate ./bundle
 	rm -rf ./config/olm/$(PLATFORM)/$(VERSION)
```
meik99, line 232: Would be nice if the lines 232 to 258 (can't select that many lines in GitHub), could be moved to scripts, e.g. python

mjgrzybek: This code is generated and we shouldn't touch it in case it's going to change in the future.
Also, additional building dependency (python, pip modules etc.) would be a pain.

btw, you can mark multiple lines using shift: https://twitter.com/natfriedman/status/1179097330097643521

mjgrzybek: but I agree it looks ugly :smile:.
How about splitting single Makefile to two Makefiles: generated and handmade? The latter can call generated when need.

meik99 (approved):

PR #682, opened 2022-03-31
Add merged entries in helm-index

gkrenn:
# Description

Currently we have 2 repos with different helm indices. This PR merges the index of the current repo with the index of the old one.

## How can this be tested?

Add the current helm-repo to your helm installation and install the oneagent-operator. => This should work now.

## Checklist


config/helm/repos/stable/index.yaml
```diff
@@ -1,5 +1,303 @@
 apiVersion: v1
 entries:
+  dynatrace-oneagent-operator:
```
0sewa0, line 3: do we want the old operator here ? 
(I know we kinda want parity with the helm-chart repo, but this feels icky to me 😅 )

gkrenn: Removed

meik99 (approved):

PR #683, opened 2022-04-01
Remove version from `matchLabels` in Helm and Activegate

0sewa0:
# Description
No label that changes overtime should be part of the matchLabels as those are immutable. Also reintroduces the old labels to be only used as matchLabels for know, will be replaced in the future

## How can this be tested?
After deploy, check if the deployments/daemonset/statefulsets have all the labels (which has the version, feature etc)
but the matchLabels should only have things that don't change (component, name)

## Checklist


src/kubeobjects/labels.go
```diff
@@ -36,3 +40,10 @@ func CommonLabels(dynakubeName string, componentName ComponentLabelValue) map[st
 		AppVersionLabel:   version.Version,
 	}
 }
+
+func MatchLabelsChanged(currentMatchLabels, desiredMatchLabels map[string]string) bool {
```
meik99, line 44:
```suggestion
func LabelsNotEqual(currentMatchLabels, desiredMatchLabels map[string]string) bool {
```

src/controllers/activegate/reconciler/statefulset/reconciler.go
```diff
@@ -158,13 +158,28 @@ func (r *Reconciler) updateStatefulSetIfOutdated(desiredSts *appsv1.StatefulSet)
...
+func (r *Reconciler) recreateStatefulSet(currentSts, desiredSts *appsv1.StatefulSet) (bool, error) {
+	log.Info("immutable section changed on statefulset, deleting and recreating", "name", desiredSts.Name)
+	err := r.Delete(context.TODO(), currentSts)
+	if err != nil {
+		return false, err
+	}
+	log.Info("deleted statefulset")
+	log.Info("recreating statefulset", "name", desiredSts.Name)
+	return true, r.Create(context.TODO(), desiredSts)
+}
```
meik99, line 181:
```suggestion
func (r *Reconciler) recreateStatefulSet(currentSts, desiredSts *appsv1.StatefulSet) (bool, error) {
	log.Info("immutable section changed on statefulset, deleting and recreating", "name", desiredSts.Name)
	
	err := r.Delete(context.TODO(), currentSts)
	if err != nil {
		return false, err
	}
	
	log.Info("deleted statefulset")
	log.Info("recreating statefulset", "name", desiredSts.Name)
	
	return true, r.Create(context.TODO(), desiredSts)
}
```

meik99 (changes requested):

meik99 (approved):

PR #684, opened 2022-04-01
Add codeModulesImage field to Dynakube where CSI Driver is used

waodim:
# Description

Added a image field in the dynakube with the modes enabled, which use the CSIDriver. The field is called "codeModulesImage".

## How can this be tested?
Apply a dynakube with either cloudNativeFullStack enabled with the field "codeModulesImage" set or with applicationMonitoring enabled with the field "codeModulesImage" set. If one decides to apply a dynakube with applicationMonitoring enabled, useCSIDriver field set to false and with the "codeModulesImages" field set, the validation webhook responds with an error.

## Checklist


luhi-DT:
I think you forgot to add general files like CRD and so on

src/api/v1beta1/oneagent_types.go
```diff
@@ -29,6 +29,10 @@ type OneAgentSpec struct {
 }
 
 type CloudNativeFullStackSpec struct {
+	// Optional: the Dynatrace installer container image
+	// Defaults to the registry on the tenant for both Kubernetes and for OpenShift
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Image",order=12,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:text"}
```
0sewa0, line 34:
```suggestion
	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="CodeModulesImage",order=12,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:text"}
```

src/api/v1beta1/oneagent_types.go
```diff
@@ -115,6 +119,11 @@ type HostInjectSpec struct {
 type ApplicationMonitoringSpec struct {
 	AppInjectionSpec `json:",inline"`
 
+	// Optional: the Dynatrace installer container image
+	// Defaults to the registry on the tenant for both Kubernetes and for OpenShift
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Image",order=12,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:text"}
```
0sewa0, line 124:
```suggestion
	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="CodeModulesImage",order=12,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:text"}
```

src/api/v1beta1/properties.go
```diff
@@ -166,6 +166,10 @@ func (dk *DynaKube) Image() string {
 		return dk.Spec.OneAgent.ClassicFullStack.Image
 	} else if dk.HostMonitoringMode() {
 		return dk.Spec.OneAgent.HostMonitoring.Image
+	} else if dk.CloudNativeFullstackMode() {
+		return dk.Spec.OneAgent.CloudNativeFullStack.CodeModulesImage
```
0sewa0, line 170: shouldn't these be in there own func ?  `func (dk *DynaKube) CodeModulesImage() string {...}`
there will be an image field for cloudNative that will use this func

0sewa0 (changes requested):

src/api/v1beta1/properties.go
waodim: you are right, did not think about that. thank you.

src/api/v1beta1/properties.go
```diff
@@ -225,6 +234,11 @@ func (dk *DynaKube) ImmutableOneAgentImage() string {
 		return oneAgentImage // TODO: What to do with the Version field in this case ?
 	}
 
+	codeModulesImage := dk.CodeModulesImage()
```
0sewa0, line 237: do we need this ? @chrismuellner

luhi-DT: I dont think we need that here, CodeModuleImage should be an own function and has nothing in common with the 'image' field of the osAgent

waodim: okay, yes that makes sense. what i do not understand currently is, where is CodeModulesImage() then called? Or is it nowhere called for the moment?

chrismuellner: that's fine, it will be used by the csi driver in the future

src/api/v1beta1/oneagent_types.go
```diff
@@ -29,6 +29,11 @@ type OneAgentSpec struct {
 }
 
 type CloudNativeFullStackSpec struct {
+	// Optional: the Dynatrace installer container image
```
chrismuellner, line 32: the description isn't quite accurate:
- this doesn't default to anything, but needs to be set by the user to a private registry

chrismuellner (changes requested):

src/api/v1beta1/oneagent_types.go
```diff
@@ -29,6 +29,11 @@ type OneAgentSpec struct {
...
+	// Defaults to the registry on the tenant for both Kubernetes and for OpenShift
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="CodeModulesImage",order=12,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:text"}
+	CodeModulesImage string `json:"codeModulesImage,omitempty"`
```
0sewa0, line 35: couldn't this be in the `AppInjectionSpec` ?

src/api/v1beta1/oneagent_types.go
waodim: ah yes thank you for reminding me.

src/api/v1beta1/oneagent_types.go
waodim: so you basically mean to remove the field out of the ApplicationMonitoringSpec and the CloudNativeFullStackSpec and just add it to the AppInjectionSpec?

0sewa0: yas :)

0sewa0:
run `make manifests` to update the the yamls according to the recent changes 😉

0sewa0 (approved):

PR #689, opened 2022-04-04
Add image field for hostAgents to cloudNativeFullStack

luhi-DT:
# Description

This PR adds an image field to cloudNativeFullstack and gives the possibility to add it in the CR. Before that PR it was not possible to set an image for the host agents when using cloudNativeFullStack. 

## How can this be tested?
1. Pull the oneagent Image from your Tenant and push it to a private registry (Quay)
2. Create a CR with cloudNative enabled and add the image you pushed to your private repository before
3. Check if the daemonset and the oneAgents, which got rolled out have the correct image set

## Checklist


0sewa0 (approved):
Should we wait for 0.6 branch to be created, or this is part of it ?

PR #691, opened 2022-04-06
Fix `activegate.customproperties.value` when multi line string is used

chrismuellner:
# Description
When using a multi line string for `customproperties.value` in the values file, invalid yaml was generated by helm.

## How can this be tested?
Use a values file that contains a multi line string in this section:
```yaml
activeGate:
  capabilities:
  - kubernetes-monitoring
  customProperties:
    value: |
      [kubernetes_monitoring]
      key=value
```
Run the following command:
```
helm install dynatrace-operator config/helm/chart/default -f values.yaml --create-namespace -n dynatrace --dry-run
```

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/Common/customresource-dynakube.yaml
+++ b/config/helm/chart/default/templates/Common/customresource-dynakube.yaml
@@ -110,8 +110,7 @@ spec:
       {{- if (.Values.activeGate.customProperties).value }}
       value: |
 {{ (.Values.activeGate.customProperties).value | indent 8 }}
-      {{- end }}
-      {{- if (.Values.activeGate.customProperties).valueFrom }}
+      {{- else if (.Values.activeGate.customProperties).valueFrom }}
       valueFrom: {{ .Values.activeGate.customProperties.valueFrom }}
       {{- end }}
     {{- end }}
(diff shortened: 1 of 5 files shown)


config/helm/chart/default/templates/Common/customresource-dynakube.yaml
```diff
@@ -106,11 +106,12 @@ spec:
     {{- end }}
 
     {{- if .Values.activeGate.customProperties }}
-      {{- if .Values.activeGate.customProperties.value }}
-    customProperties:
-      value: {{ .Values.activeGate.customProperties.value }}
-      {{- else if .Values.activeGate.customProperties.valueFrom }}
     customProperties:
+      {{- if (.Values.activeGate.customProperties).value }}
```
meik99, line 110: This situation would call for a unit tests, please add one to the Helm unittest suite for the custom resource

meik99 (changes requested):

config/helm/chart/default/templates/Common/customresource-dynakube.yaml
chrismuellner: I didn't check off the unit test entry yet 😉

meik99 (approved):

PR #694, opened 2022-04-07
Fixes using wrong tag for immutable image when version is set

meik99:
# Description

## Summary

This PR truncates the build date of a supplied OneAgent version when used as a tag for the immutable image.
It implements the change for the `ImmutableOneAgentImage()` DynaKube property.

## Which issue is fixed

When the OneAgent version is set in the custom resource and the CSI driver is used, the image cannot be pulled since the version has the wrong format for the image.

The CSI driver expects the version to be in a format like "1.231.288.20220218-170442". If the version is not in this format, the CSI driver cannot download it from the API since the version has to include the build date.

However, the same version field is used to overwrite the version tag for the oneagent images that are used. These image tags do not contain the build date.
Therefore, the version can only be correct for either the API or the Image tag.

### Environment necessary

Fresh Kubernetes cluster

### Recreating the error

On a fresh cluster,  
create the dynatrace namespace,  
deploy the operator from the master branch or use the 0.5 release,  
create the dynakube secret,  
modify a sample to use cloud native full stack and set the version field to a valid version, e.g., the one described above,  
watch the pods fail due to a image pull back off error,  
modify the sample to use a version without the build date and apply,  
watch the pods be created successfully, however the CSI driver will fail in the logs due to the version being invalid

### Confirming the fix

On a fresh cluster,  
create the dynatrace namespace,  
deploy the operator from fix/truncate-version-build-date branch,  
create the dynakube secret,  
modify a sample to use cloud native full stack and set the version field to a valid version, e.g., the one described above,  
the pods should start successfully, the version tag should be the same as the one in the CR but without the build date,  
the CSI driver does not show error logs regarding the version

## Checklist

### DIFF
--- a/src/api/v1beta1/properties.go
+++ b/src/api/v1beta1/properties.go
@@ -232,12 +232,27 @@ func (dk *DynaKube) ImmutableOneAgentImage() string {
 	}
 
 	tag := "latest"
-	if ver := dk.Version(); ver != "" {
-		tag = ver
+	if version := dk.Version(); version != "" {
+		truncatedVersion := truncateBuildDate(version)
+		tag = truncatedVersion
 	}
 
 	registry := buildImageRegistry(dk.Spec.APIURL)
 	return fmt.Sprintf("%s/linux/oneagent:%s", registry, tag)
+}
+
+func truncateBuildDate(version string) string {
+	const versionSeperator = "."
+	const buildDateIndex = 3
+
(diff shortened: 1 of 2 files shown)


luhi-DT:
Should we have also an check in the validation webhook, that the version provided has a compatible and correct format?

meik99:
> Should we have also an check in the validation webhook, that the version provided has a compatible and correct format?

We could consider that as an additional feature.
This PR should only fix the already existing bug

PR #706, opened 2022-04-07
Add an validator to verify correctness of ProxyUrl string/secret

aorcholski:
Description
ActiveGate Proxy URL has to be properly [encoded](https://en.wikipedia.org/wiki/Percent-encoding) URI. ActiveGate POD fails if Proxy URL contains special characters (for instance "?) unencoded. To improve user experience I propose to add an Validator to reject a CR if the Proxy value or Proxy secret is invalid. 

How can this be tested?
Use any unencoded special character(s) in the Proxy Url (proxy.value, proxy secret) and try to apply your CR - should be rejected.

## Checklist


src/agproxysecret/secret.go
```diff
@@ -129,7 +129,7 @@ func (agProxySecretGenerator *ActiveGateProxySecretGenerator) proxyUrlFromUserSe
 func parseProxyUrl(proxy string) (host string, port string, username string, password string, err error) {
 	proxyUrl, err := url.Parse(proxy)
 	if err != nil {
-		return "", "", "", "", err
+		return "", "", "", "", fmt.Errorf("could not parse proxy URL")
```
meik99, line 132: use `errors.New()` from the `pkg/errors` library

src/webhook/validation/activegate.go
```diff
@@ -66,6 +81,65 @@ func missingActiveGateMemoryLimit(dv *dynakubeValidator, dynakube *dynatracev1be
...
 
+func invalidActiveGateProxyUrl(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
+	if dynakube.Spec.Proxy != nil {
+		if len(dynakube.Spec.Proxy.ValueFrom) > 0 {
+			var proxySecret corev1.Secret
+			err := dv.clt.Get(context.TODO(), client.ObjectKey{Name: dynakube.Spec.Proxy.ValueFrom, Namespace: dynakube.Namespace}, &proxySecret)
+			if k8serrors.IsNotFound(err) {
+				return errorMissingActiveGateProxySecret
+			} else if err != nil {
+				return fmt.Sprintf("error occurred while reading PROXY secret indicated in the Dynakube specification (%s)", err.Error())
```
meik99, line 92: use `errors.Wrap()` from the `pkg/errors` libraries

src/webhook/validation/activegate.go
```diff
@@ -66,6 +81,65 @@ func missingActiveGateMemoryLimit(dv *dynakubeValidator, dynakube *dynatracev1be
...
+	return ""
+}
+
 func memoryLimitSet(resources corev1.ResourceRequirements) bool {
 	return resources.Limits != nil && resources.Limits.Memory() != nil
 }
+
+// proxyUrl is valid if
+// 1) encoded
+// 2) "`\ are escaped using \
```
meik99, line 112: does that mean it must be either escaped or encoded or must it be both?
If it must be both, would that then work with the dtclient, since it would take the password unencoded?

src/webhook/validation/activegate.go
```diff
@@ -66,6 +81,65 @@ func missingActiveGateMemoryLimit(dv *dynakubeValidator, dynakube *dynatracev1be
...
+			return evalErrorMessage
+		}
+	}
+	return ""
+}
+
+// 'eval' command is used by entrypoint.sh:readSecret function to return its result.
+// For this reason quotation mark ", backtick ` and backslash \ characters have to be escaped using backslash.
+func isEvalEscapeNeeded(str string) bool {
+	previousChar := '\000'
```
meik99, line 128: introduce a local constant for the null character just so it is more obvious that that magic number is supposed to be that character

src/webhook/validation/activegate.go
```diff
@@ -66,6 +81,65 @@ func missingActiveGateMemoryLimit(dv *dynakubeValidator, dynakube *dynatracev1be
...
+		if isEvalEscapeNeeded(password) {
+			return evalErrorMessage
+		}
+	}
+	return ""
+}
+
+// 'eval' command is used by entrypoint.sh:readSecret function to return its result.
+// For this reason quotation mark ", backtick ` and backslash \ characters have to be escaped using backslash.
+func isEvalEscapeNeeded(str string) bool {
```
meik99, line 127: as far as I understand, this function would effectively be a block-list, blocking the three known characters. It would be more secure to restructure it to an allow-list, checking that every character is one of the known ones, i.e., letters, numbers and tested special characters and then checking if they are appropriately escaped

src/webhook/validation/activegate.go
```diff
@@ -19,6 +24,16 @@ Make sure you correctly specify the ActiveGate capabilities in your custom resou
...
+
+	errorInvalidActiveGateProxyUrl = `The DynaKube's specification has an invalid Proxy URL value set. Make sure you correctly specify the URL in your custom resource.`
+	errorInvalidEvalCharacter      = `The DynaKube's specification has an invalid Proxy password value set. Make sure you correctly escape quotation mark, backtick and backslash characters using backslash.`
+
+	errorMissingActiveGateProxySecret = `The Proxy secret indicated by the DynaKube specification doesn't exist.`
+
+	errorInvalidProxySecretFormat = `The Proxy secret indicated by the DynaKube specification has an invalid format. Make sure you correctly creates the secret.`
+
+	errorInvalidProxySecretUrl           = `The Proxy secret indicated by the DynaKube specification has an invalid URL value set. Make sure you correctly specify the URL in the secret.`
+	errorInvalidProxySecretEvalCharacter = `The Proxy secret indicated by the DynaKube specification has an invalid Proxy password value set. Make sure you correctly escape quotation mark, backtick and backslash characters using backslash.`
```
meik99, line 36: With so many specifc errors (and also logic), this code and other proxy validation code should probably belong into a separate file, e.g., "validation/proxy.go"

meik99 (changes requested):

src/webhook/validation/activegate.go
aorcholski: my fault, characters can not be escaped

src/webhook/validation/activegate.go
aorcholski: my fault, characters can not be escaped because of dtclient. Finally, there are two invalid characters : apostrophe and backtick.

src/webhook/validation/activegate.go
aorcholski: done

src/webhook/validation/activegate.go
aorcholski: Characters can not be escaped due to dtclient (which uses the same password) so the logic is simpler - apostrophe and backtick are forbidden. The function has been renamed to evalIsStringInvalid

src/agproxysecret/secret.go
aorcholski: Done

src/webhook/validation/activegate.go
aorcholski: Done

src/webhook/validation/proxy_url.go
```diff
@@ -0,0 +1,75 @@
...
+	return ""
+}
+
+// 'eval' command is used by entrypoint.sh:readSecret function to return its result.
+// For this reason apostrophe ' and backtick ` characters has to be escaped using backslash.
+// On the other hand the operator needs a pure password at the same time.
+// Finally, these 2 characters are forbidden.
+func evalIsStringInvalid(str string) bool {
+	for _, char := range str {
+		if char == '\'' || char == '`' {
```
meik99, line 70: This implements a blocklist again, please rewrite that function to return true anytime it encounters an unkown character

meik99 (changes requested):

src/webhook/validation/proxy_url.go
meik99: Something like this would implement an allowlist, only permitting the characters in the brackets to be used

```
func isStringValidForEval(str string) bool {
	regex := regexp.MustCompile(`^[a-zA-Z0-9./?,*;\\ ]+$`)
	return regex.MatchString(str)
}

func TestAllowList(t *testing.T) {
	assert.True(t, isStringValidForEval("a simple password"))
	assert.True(t, isStringValidForEval("a simple password w1th sp?cial ;; chara/cters"))
	assert.False(t, isStringValidForEval("A ` backtick"))
	assert.False(t, isStringValidForEval("A ' single quote"))
}

```

aorcholski: regex is used

src/webhook/validation/proxy_url_test.go
```diff
@@ -0,0 +1,164 @@
...
+				},
+				Data: map[string][]byte{
+					"proxy": []byte(invalidPlainTextProxyUrl),
+				},
+			})
+	})
+
+	t.Run(`invalid proxy secret url - entrypoint.sh`, func(t *testing.T) {
+		assert.Equal(t, true, isStringValidForAG("password"))
+		assert.Equal(t, true, isStringValidForAG("test~!@#^*()_-|}{[]\":;?><./pass"))
```
meik99, line 140: sorry I didn't see that during the first review. There is `assert.True` and `assert.False` for these kinds of assertions. Otherweise lgmt

meik99 (changes requested):

src/webhook/validation/proxy_url_test.go
aorcholski: good point, fixed

meik99 (approved):

meik99 (approved):

PR #713, opened 2022-04-08
Add highavailability option for webhook via helm

waodim:
# Description

Introduces a new option for the webhook. The flag is called enableHighAvailability and can be found in the helm charts in the values.yaml file. Changes the replicas of the webhook deployment from 1 to 2 and also introduces a PodDisruptionBudget with minAvailable set to 1. Also topologySpreadConstraints are set.

## How can this be tested?

Set the flag to true, render a chart and apply the output file.

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
+++ b/config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
@@ -21,7 +21,7 @@ metadata:
   labels:
     {{- include "dynatrace-operator.webhookLabels" . | nindent 4 }}
 spec:
-  replicas: {{ (.Values.webhook).enableHighAvailability | ternary 2 1 }}
+  replicas: {{ (.Values.webhook).highAvailability | ternary 2 1 }}
   revisionHistoryLimit: 1
   selector:
     matchLabels:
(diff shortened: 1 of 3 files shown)


config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
```diff
@@ -21,7 +21,11 @@ metadata:
   labels:
     {{- include "dynatrace-operator.webhookLabels" . | nindent 4 }}
 spec:
+  {{- if (.Values.webhook).enableHighAvailability }}
```
chrismuellner, line 24: I think the `ternary` operator could improve readability here: [helm docs](https://helm.sh/docs/chart_template_guide/function_list/#ternary)

config/helm/chart/default/values.yaml
```diff
@@ -56,6 +56,7 @@ webhook:
   limits:
     cpu: 300m
     memory: 128Mi
+  enableHighAvailability: false
```
meik99, line 59: imo it would suffice if it was called `highAvailability` since the bool value already implies a enable / disable behaviour

config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
   labels:
     {{- include "dynatrace-operator.webhookLabels" . | nindent 4 }}
 spec:
-  replicas: 1
+  replicas: {{ (.Values.webhook).enableHighAvailability | ternary 2 1 }}
```
meik99, line 24: I think the ternary operator is less readable, but that's just a matter of preference. 
If it stays that way I am also fine with it

config/helm/chart/default/values.yaml
waodim: you are right, makes total sense. thank you

meik99 (approved):

PR #715, opened 2022-04-08
Gets kubeconfig only when subcommand needs it

0sewa0:
# Description
If the users pod has `automountServiceAccountToken: false` then the standalone init will fail.
Probably because https://github.com/Dynatrace/dynatrace-operator/blob/master/src/cmd/operator/main.go#L55-L60 will run which needs these tokens it.

However the standalone init does not use this config.

## How can this be tested?
Deploy a pod were we would inject with the standalone init that has the `automountServiceAccountToken: false` set.

## Checklist

### DIFF
--- a/src/cmd/operator/main.go
+++ b/src/cmd/operator/main.go
@@ -52,18 +52,17 @@ func main() {
 	version.LogVersion()
 
 	namespace := os.Getenv("POD_NAMESPACE")
-	cfg, err := config.GetConfig()
-	if err != nil {
-		log.Error(err, "")
-		os.Exit(1)
-	}
-
 	var mgr manager.Manager
 	var cleanUp func()
 
 	subCmd := getSubCommand()
 	switch subCmd {
 	case operatorCmd:
+		cfg, err := config.GetConfig()
+		if err != nil {
+			log.Error(err, "")
+			os.Exit(1)
+		}
 		if !kubesystem.DeployedViaOLM() {
 			// setup manager only for certificates
 			bootstrapperCtx, done := context.WithCancel(context.TODO())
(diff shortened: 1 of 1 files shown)


src/cmd/operator/main.go
```diff
@@ -52,18 +52,17 @@ func main() {
...
-		os.Exit(1)
-	}
-
 	var mgr manager.Manager
 	var cleanUp func()
 
 	subCmd := getSubCommand()
 	switch subCmd {
 	case operatorCmd:
+		cfg, err := config.GetConfig()
```
chrismuellner, line 61: can this be done in a more reusable way? same 5 lines in 3 of the 4 sub commands...

baichinger:
The proposed solutions seems overly complex compared to what is actually needed. A dedicated binary for the `init` sub-command would be easy to understand, maintain, and cleanly separate concerns.

dt-team-kubernetes:
> The proposed solutions seems overly complex compared to what is actually needed. A dedicated binary for the `init` sub-command would be easy to understand, maintain, and cleanly separate concerns.

As far as I understand, the complexity comes from calling the function three times instead of once, for the subcommands that need it. However, in my opinion, the complexity of creating, publishing and maintaining a separate binary if far greater then the one introduced by this solution.

meik99:
> The proposed solutions seems overly complex compared to what is actually needed. A dedicated binary for the `init` sub-command would be easy to understand, maintain, and cleanly separate concerns.

As far as I understand, the complexity comes from calling the function three times instead of once, for the subcommands that need it. However, in my opinion, the complexity of creating, publishing and maintaining a separate binary if far greater then the one introduced by this solution.

meik99 (approved):

PR #717, opened 2022-04-08
Add ability to supply custom annotations to DynaKube via values.yaml

dfallert-rcd:
# Description

This will allow users of the helm chart to add annotations to the `DynaKube` object via the helm chart itself.  Among other things, this will allow users to set feature flags such as `alpha.operator.dynatrace.com/feature-automatic-kubernetes-api-monitoring: "true"` via the helm chart, instead of needing to take post-install action.

## How can this be tested?

in the values.yaml under `activeGate.annotations`, provide a list of annotations you would like added to the `DynaKube` object.
```YAML
activeGate:
  capabilities:
  - routing
  - kubernetes-monitoring
  - metrics-ingest
  annotations: 
    - 'alpha.operator.dynatrace.com/disable-metadata-enrichment: "false"'
    - 'alpha.operator.dynatrace.com/feature-automatic-kubernetes-api-monitoring: "true"'
```
perform a `helm template` command with the `--debug` flag, and supply the necessary parameters. 
Search for the DynaKube object and observe the annotations

```YAML
kind: DynaKube
metadata:
  annotations:
    alpha.operator.dynatrace.com/disable-metadata-enrichment: "false"
    alpha.operator.dynatrace.com/feature-automatic-kubernetes-api-monitoring: "true"
    helm.sh/hook: post-install,post-upgrade
```

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/Common/customresource-dynakube.yaml
+++ b/config/helm/chart/default/templates/Common/customresource-dynakube.yaml
@@ -31,6 +31,9 @@ metadata:
     {{- if (.Values.activeGate).readOnlyFs }}
     alpha.operator.dynatrace.com/feature-activegate-readonly-fs: "true"
     {{- end }}
+    {{- range (.Values.activeGate).annotations }}
+    {{ . }}
+    {{- end }}
   name: {{ .Values.name }}
   namespace: {{ .Release.Namespace }}
   labels:
--- a/config/helm/chart/default/values.yaml
+++ b/config/helm/chart/default/values.yaml
@@ -139,6 +139,7 @@ activeGate:
   priorityClassName: ""
   apparmor: false
   readOnlyFs: false
+  annotations: []
 
 # deprecated
 kubernetesMonitoring:


sebv7:
Would be nice if someone can review this. We are also waiting for this ability.

gowrisankar22:
we are also looking for the same.

arno790:
it would be nice to have this feature as soon as possible

chrismuellner:
There are plans to change CR generation in the future.

If you want to include annotations now, you can install the operator by passing no `apiUrl` and `autoCreateSecret=false` to just install operator components, and then apply a Dynakube manually after.

You can also annotate the Dynakube resource via the `kubectl annotate` command after the helm chart is installed.

dfallert-rcd:
> There are plans to change CR generation in the future.
> 
> If you want to include annotations now, you can install the operator by passing no `apiUrl` and `autoCreateSecret=false` to just install operator components, and then apply a Dynakube manually after.
> 
> You can also annotate the Dynakube resource via the `kubectl annotate` command after the helm chart is installed.

Doing things manually defeats the entire purpose of the operator, and makes building automation around this more complex than it needs to be. This PR is a rather simple solution to a common grievance with this helm chart installation, is there more I need to add to it in order to pass review?

0sewa0:
>Doing things manually defeats the entire purpose of the operator, and makes building automation around this more complex than it needs to be. This PR is a rather simple solution to a common grievance with this helm chart installation, is there more I need to add to it in order to pass review?

While I agree that this a simple and good change, please consider the following:
- Even if we merged this change right now, you could only use in the next release which is not going to happen right now, as releasing takes time.
- We are thinking about a different way of handling the CR, because the current way we are doing things(with the CR and CRD) is not viable on all platforms we want to support. So adding this change prematurely might end up changing nothing, as it might be changed again in the short term (which would cause confusion).

martin-ducar-gd:
Look @Oberon00 we are pretty big customer of Dynatrace just trying to onboard ~100 K8s clusters and this is major pain in the a**. Similar PR raised here: https://github.com/Dynatrace/dynatrace-operator/pull/727 this is really putting the image of the company in bad light its simple tactical solution to a problem which when not solved here takes very ugly hack/workaround. I understand you plan to redo the process of install, but I assume it will take weeks instead of days. This on other hand is simple change taking hours to finalize. This is actually a bug in your infra because without annotation ```yaml
alpha.operator.dynatrace.com/feature-automatic-kubernetes-api-monitoring: "true"
```
The operator won't register the cluster in the dynatrace, this was found during debugging with assigned support engineer from your side. So this should actually ideally go to release-* branch which I would leave up to you to decide. Hence why the other PR was raised to release-0.5. This will get to you either from here or from support ticket we need this escalated and fixed as soon as possible. Thank you

DTMad:
Hey,

I would like to turn your attention to the Github repositories for both Helm and Cert-Manager respectively. The issue we are contending with is explained therein. In issue 7735 https://github.com/helm/helm/issues/7735 on the Helm repository you can see a long conversation about Helm’s inability to upgrade CRD’s, which is the reason we originally chose to kubectl apply the CRD instead of installing it with Helm. The problem there, as our friends at Jet Stack learned two years ago, as you can read about here: https://github.com/cert-manager/cert-manager/pull/2646#discussion_r388202951 is that this creates an automation problem, like you are experiencing, in which the custom resource values file begins to grow in complexity, becoming more difficult for the repository owners to maintain and update. So you basically shift the upgrade / maintenance problem from the CRD to the CR. What we see, then, is that many Helm charts, like cert-manager, offer different options, and different solutions, some of which rely on tools like Bazel, as you can read about here: [CRDs - cert-manager Documentation](https://cert-manager.io/docs/contributing/crds/#versions).

If you really want to use your solution, we suggest you fork our repo and use it. In the meantime, we are going to solve this problem the way we stated. We are also thinking about adding the self-registering ActiveGate annotations as default values.

Assuming we were in sync with the PR, it would take a few weeks for us to test and approve it, so the changes we’re talking about, which solve your specific annotation problem, and require some level of automation regardless, are due to land in the same general timeframe.

If this is completely unacceptable, we would like to refer your company contact to our product manager, Matthew Reider ([matthew.reider@dynatrace.com](mailto:matthew.reider@dynatrace.com)) who is happy to discuss the rationale for this decision. Of course, you, as a customer, just expect things to work, as easily as possible, and we totally understand that. The decisions we are making eventually make us, your trusted software vendor, go faster, maintain things more easily, be more flexible, and absolutely increase your ability to automate things in less complicated ways. We hope you trust that we’ve considered this very carefully, as has the Helm team, the Cert Manager team, and a dozen other open source projects that run into issues stemming from the fact that Helm is quite imperfect when it comes to managing custom resources, and their related custom resource definitions.

Thanks,
Marco

tonedefdev:
I know this issue is closed already, but @DTMad when leveraging Flux or ArgoCD deploying CRDs are feasible and overcome a lot of the shortcomings that you have with Helm alone due to those applications using the native k8s reconciliation to loop to bring everything into desired state. 

Today, I create our own Helm chart with this chart as a dependency and include the CRD as a template. ArgoCD deploys it without issue, and any future CRD updates should be fine, as OPA Gatekeeper deploys its policies as CRDs and I have been able to update those with ArgoCD without any issues.

This annotation would be nice to include for those of us who are using Flux or ArgoCD so that we can automate this deployment end-to-end. I'll fork it and create my own Helm chart, but figured I'd add that as a reason for including this so we don't have to 100% maintain the chart on our own.

PR #725, opened 2022-04-14
Add highavailability option for webhook via helm

waodim:
# Description

Introduces a new option for the webhook. The flag is called enableHighAvailability and can be found in the helm charts in the values.yaml file. Changes the replicas of the webhook deployment from 1 to 2 and also introduces a PodDisruptionBudget with minAvailable set to 1. Also topologySpreadConstraints are set.

## How can this be tested?
Set the flag to true, render a chart and apply the output file.

## Checklist


config/helm/chart/default/values.yaml
```diff
@@ -85,7 +86,6 @@ classicFullStack:
 
 cloudNativeFullStack:
   enabled: false
-  image: ""
```
chrismuellner, line 88: `image` field exists on master

chrismuellner (approved):

PR #729, opened 2022-04-15
Implements copying codeModules from image.

0sewa0:
# Description
This change gets the image specified in the `codeModulesImage` field and unpacks it's contents into the right place using the csi-driver. The outcome is meant to be the same as getting the zip from the tenant, but this way the user can scan the contents of the image they want us to install.

Some extra logic had to be added where the version is important (creating the symlink) as the image not necessarily has this information plainly available.

## How can this be tested?
1. Create an image with the codeModules inside that you want to use. 
2. Set the `codeModulesImage` to that image. 
    - make sure that the operator can "pull" that image, use a custom pull-secret if needed
    - **WARNING**: if you use a custom-pull secret, but still want to get the (host) oneagent images from your tenant, you must have the necessary creds in your custom-pull secret to access the tenant, and cloudNative injection will fail if you don't have running oneagents in your cluster.  (Using applicationMonitoring  with csi-driver for testing is simpler 😉 .)
3.  Deploy the dynakube and some sample apps, and make sure everything works as expected.
    - You can exec into the pod of the csi-driver and check if everything was downloaded properly. (`/data/<tenant>/bin/<image-tag>`)

## Checklist


src/installer/url.go
```diff
@@ -0,0 +1,93 @@
...
+		fileSize = stat.Size()
+	}
+
+	log.Info("saved OneAgent package", "dest", tmpFile.Name(), "size", fileSize)
+	log.Info("unzipping OneAgent package")
+	if err := installer.extractZip(tmpFile, targetDir); err != nil {
+		return fmt.Errorf("failed to unzip file: %w", err)
+	}
+	log.Info("unzipped OneAgent package")
+	return nil
```
meik99, line 48: could be refactored into its own function. Also use errors.New() for new errors. If errors.New() does not have formatting options some function in the `pkg/errors` package will, please double check other error returns as well

src/installer/zip.go
```diff
@@ -0,0 +1,158 @@
...
+			defer func() { _ = dstFile.Close() }()
+
+			srcFile, err := file.Open()
+			if err != nil {
+				return err
+			}
+			defer func() { _ = srcFile.Close() }()
+
+			_, err = io.Copy(dstFile, srcFile)
+			return err
```
meik99, line 72: refactor into its own function

src/installer/zip.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	defer func() { _ = reader.Close() }()
+
+	gzipReader, err := gzip.NewReader(reader)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = gzipReader.Close() }()
+
+	tarReader := tar.NewReader(gzipReader)
```
meik99, line 99: can be moved into its own function creating the tarReader

src/installer/zip.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		target := filepath.Join(targetDir, header.Name)
+
+		// Check for ZipSlip: https://snyk.io/research/zip-slip-vulnerability
+		if !strings.HasPrefix(target, targetDir) {
+			return fmt.Errorf("illegal file path: %s", target)
+		}
+
+		switch header.Typeflag {
```
meik99, line 115: move the switch into a function, the cases could be moved into their own functions as well

src/controllers/csi/provisioner/agent.go
```diff
@@ -42,71 +49,95 @@ func newAgentUpdater(
...
 		path:      path,
+		apiReader: apiReader,
 		recorder:  recorder,
 		dk:        dk,
 		installer: agentInstaller,
 	}
 }
 
-func (updater *agentUpdater) updateAgent(installedVersion, tenantUUID string, previousHash string, latestProcessModuleConfigCache *processModuleConfigCache) (string, error) {
+func (updater *agentUpdater) updateAgent(ctx context.Context, latestVersion, tenantUUID string, previousHash string, latestProcessModuleConfigCache *processModuleConfigCache) (string, error) {
```
meik99, line 59: Move the context and latestVersion into the agentUpdater constructor. The previousHash string is not used anymore in this function. latestVersion is also only used for logging, maybe the logging could be done somewhere else or in different way so this parameter does not have to be passed as well

src/controllers/dynakube/updates/version.go
```diff
@@ -51,75 +54,55 @@ func ReconcileVersions(
...
 		defer func() {
-			_ = os.Remove(path.Join(dtversion.TmpCAPath, dtversion.TmpCAName))
+			_ = os.Remove(TmpCAPath)
 		}()
 	}
 	upd = true // updateImageVersion() always updates the status
 
 	if needsActiveGateUpdate {
-		if err := updateImageVersion(dkState, dk.ActiveGateImage(), &dk.Status.ActiveGate.VersionStatus, &dockerCfg, verProvider, true); err != nil {
+		if err := updateImageVersion(dkState, dk.ActiveGateImage(), &dk.Status.ActiveGate.VersionStatus, dockerConfig, verProvider, true); err != nil {
```
meik99, line 75: This function must be refactored somehow. Not only is it called with six parameters, the parameters are almost the same. Also the `allowDowngrades` flag seems weird. Additionally, these condtionals could be moved into a function.

meik99 (changes requested):

src/controllers/dynakube/updates/version.go
0sewa0: The updates package should be refactored in another PR

src/dockerconfig/docker_config.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	err := json.Unmarshal(config, &dockerConf)
+	if err != nil {
+		return nil, err
+	}
+
+	return dockerConf.Auths, nil
+}
+
+func (config *DockerConfig) SaveCustomCAs(ctx context.Context, kubeClient client.Reader, dynakube dynatracev1beta1.DynaKube, path string) error {
```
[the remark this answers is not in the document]
0sewa0: I would keep the context in the params, it feels weird to hide in the struct, even-though that it would work considering the way we currently use it

src/installer/symlink.go
```diff
@@ -0,0 +1,78 @@
...
+			return nil
+		}
+		folderName := filepath.Base(path)
+		if regexp.MustCompile(versionRegexp).Match([]byte(folderName)) {
+			log.Info("found version", "version", folderName)
+			version = folderName
+			return iofs.ErrExist
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
0sewa0: `aferoFs.Walk(targetDir, walkFiles)`
does not return anything and the `walkFiles` func needs to have the params list `path string, info iofs.FileInfo, err error`, so to actually get the `version` and use it I have to create it as a "closure" (the `version = folderName` is the most important part, and the whole reason we are doing this)

src/installer/zip.go
0sewa0: closing the readers would be a pain, so I would rather not do it. (it would introduce more complexity because of it)

src/installer/zip.go
0sewa0: tried it out, the end result just made things a tiny bit more complicated. you basically have to pass in everything and couldn't make it general enough

src/installer/symlink.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	iofs "io/fs"
+	"path/filepath"
+	"regexp"
+
+	"github.com/spf13/afero"
+)
+
+const (
+	// example match: 1.239.14.20220325-164521
```
chrismuellner, line 12: 😄

chrismuellner (changes requested):

src/installer/installer.go
```diff
@@ -68,203 +74,30 @@ func NewOneAgentInstaller(
 func (installer *OneAgentInstaller) InstallAgent(targetDir string) error {
 	log.Info("installing agent", "target dir", targetDir)
 	installer.props.fillEmptyWithDefaults()
-	if err := installer.installAgent(targetDir); err != nil {
+	var err error
+	if installer.props.ImageInfo != nil {
```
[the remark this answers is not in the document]
0sewa0: look at what you made me do: [`57c11c4` (#729)](https://github.com/Dynatrace/dynatrace-operator/pull/729/commits/57c11c4b56c35ec44cf5658ff11f6ff38120eb37) 😨

src/installer/image/destination.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func getDestinationInfo(imageCacheDir string) (*types.SystemContext, *types.ImageReference, error) {
+	destinationRef, err := getDestinationReference(imageCacheDir)
+	if err != nil {
+		return nil, nil, err
+	}
+	destinationCtx := buildDestinationContext(imageCacheDir)
+	return destinationCtx, &destinationRef, nil
+}
```
meik99, line 17: why combine these functions? Returning three values is not very nice

src/installer/image/source.go
```diff
@@ -0,0 +1,47 @@
...
+
+	sourceRef, err := getSourceReference(imageRef)
+	if err != nil {
+		log.Info("failed to get source reference", "image", pullInfo.Image, "imageRef", imageRef)
+		return nil, nil, err
+	}
+	log.Info("got source reference", "image", pullInfo.Image)
+
+	sourceCtx := buildSourceContext(imageRef, cacheDir, pullInfo.DockerConfig)
+	return sourceCtx, &sourceRef, nil
```
meik99, line 26: similar to before, returning more than one values with one error is weird

src/installer/image/source.go
```diff
@@ -0,0 +1,47 @@
...
+	ref = reference.TagNameOnly(ref)
+
+	return ref, nil
+}
+
+func getSourceReference(named reference.Named) (types.ImageReference, error) {
+	return docker.NewReference(named)
+}
+
+func buildSourceContext(imageRef reference.Named, cacheDir string, dockerConfig dockerconfig.DockerConfig) *types.SystemContext {
```
meik99, line 43:
```suggestion
func createSourceContext(imageRef reference.Named, cacheDir string, dockerConfig dockerconfig.DockerConfig) *types.SystemContext {
```

src/installer/image/unpack.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+		manifests = append(manifests, ociManifest)
+
+	} else if mimeType == ocispec.MediaTypeImageIndex {
+		ociManifests, err := unmarshallImageIndex(installer.fs, imageCacheDir, manifestBlob)
+		if err != nil {
+			return nil, err
+		}
+		manifests = append(manifests, ociManifests...)
+	}
```
meik99, line 66: An implementation that does not need to be changed that much when a new mime type is added would be preferred.

Maybe a map of functions to be called on specific types?

src/installer/image/unpack.go
```diff
@@ -0,0 +1,135 @@
...
+	for _, entry := range manifests {
+		for _, layer := range entry.LayerInfos() {
+			switch layer.MediaType {
+			case ocispec.MediaTypeImageLayerGzip:
+				sourcePath := filepath.Join(imageCacheDir, "blobs", layer.Digest.Algorithm().String(), layer.Digest.Hex())
+				if err := zip.ExtractGzip(installer.fs, sourcePath, targetDir); err != nil {
+					return err
+				}
+			case ocispec.MediaTypeImageLayerZstd:
+				return fmt.Errorf("MediaTypeImageLayerZstd is not implemented")
```
meik99, line 81: similar to above, an implementation that is more robust to changes would be preferred

src/installer/image/unpack.go
```diff
@@ -0,0 +1,135 @@
...
+	for _, descriptor := range index.Manifests {
+		manifestFile, err := aferoFs.ReadFile(filepath.Join(imageCacheDir, "blobs", descriptor.Digest.Algorithm().String(), descriptor.Digest.Hex()))
+		if err != nil {
+			return nil, err
+		}
+		ociManifest, err := manifest.OCI1FromManifest(manifestFile)
+		if err != nil {
+			return nil, err
+		}
+		manifests = append(manifests, ociManifest)
```
meik99, line 108:
```suggestion
		manifests = append(manifests, createOciFromManifest(...))
		
...		
func createOciFromManifest(...) {
        manifestFile, err := aferoFs.ReadFile(filepath.Join(imageCacheDir, "blobs", 
        descriptor.Digest.Algorithm().String(), descriptor.Digest.Hex()))
		if err != nil {
			return nil, err
		}
		ociManifest, err := manifest.OCI1FromManifest(manifestFile)
		if err != nil {
			return nil, err
		}
		return ociManifest
}
```

src/installer/symlink/symlink.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	iofs "io/fs"
+	"path/filepath"
+	"regexp"
+
+	"github.com/spf13/afero"
+)
+
+func CreateSymlinkIfNotExists(fs afero.Fs, targetDir string) error {
```
meik99, line 11: Create a symlink from where?

src/installer/url/installer.go
```diff
@@ -0,0 +1,73 @@
...
+		dtc:   dtc,
+		props: props,
+	}
+}
+
+type urlInstaller struct {
+	fs    afero.Fs
+	dtc   dtclient.Client
+	props *Properties
+}
```
meik99, line 41: Struct should be above constructor.

Also:
```suggestion
type urlInstaller struct {
	filesystem    afero.Fs
	dynatraceClient   dtclient.Client
	properties *Properties
}
```

src/installer/zip/gzip.go
```diff
@@ -0,0 +1,95 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/klauspost/compress/gzip"
+	"github.com/spf13/afero"
+)
+
+func ExtractGzip(fs afero.Fs, sourceFilePath, targetDir string) error {
```
meik99, line 15: this function is way too long, it must be refactored

src/installer/zip/zip.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to determine file info: %w", err)
+	}
+
+	reader, err := zip.NewReader(sourceFile, fileInfo.Size())
+	if err != nil {
+		return fmt.Errorf("failed to open ZIP file: %w", err)
+	}
+
+	_ = fs.MkdirAll(targetDir, 0755)
```
meik99, line 30: error isn't handled

meik99 (dismissed):

src/api/v1beta1/properties.go
```diff
@@ -21,7 +21,7 @@ import (
 	"net/url"
 	"strings"
 
-	"github.com/Dynatrace/dynatrace-operator/src/dtclient"
+	"github.com/Dynatrace/dynatrace-operator/src/dtclient/types"
```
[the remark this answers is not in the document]
0sewa0: to avoid circular dependency

src/controllers/csi/provisioner/agent.go
```diff
@@ -1,112 +1,158 @@
...
-		log.Info("updating ruxitagentproc.conf on latest installed version")
-		if err := updater.installer.UpdateProcessModuleConfig(targetDir, latestProcessModuleConfigCache.ProcessModuleConfig); err != nil {
-			return "", err
-		}
+
+	return updatedVersion, nil
+}
+
+func (updater *agentUpdater) initUrlInstaller() {
+	// To allow mocking
```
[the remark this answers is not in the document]
0sewa0: but we create the installer when we update/install something, so you can't mock it for the "main" `updateAgent` function

src/api/v1beta1/properties.go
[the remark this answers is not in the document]
0sewa0: actually it works without this 😄

src/api/v1beta1/properties.go
```diff
@@ -233,6 +233,20 @@ func (dk *DynaKube) Version() string {
...
+		return ""
+	}
+	if dynakube.CodeModulesImage() != "" {
+		codeModulesImage := dynakube.CodeModulesImage()
+		return strings.Split(codeModulesImage, ":")[1]
+	}
+	if dynakube.Version() != "" {
+		return dynakube.Version()
+	}
+	return dynakube.Status.LatestAgentVersionUnixPaas
```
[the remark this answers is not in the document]
0sewa0: `flavor != type`
we use default flavor in case of arm, and default type

chrismuellner (approved):

PR #731, opened 2022-04-15
Remove possibility to install operator with install.sh

gkrenn:
# Description

install.sh is no longer maintained. - As old tenants reference this script, we can't remove it. Instead we display a message that this installation method is outdated.

## How can this be tested?

Run `bash deploy/install.sh`. - Only a msg should be shown.

## Checklist


deploy/install.sh
```diff
@@ -2,367 +2,14 @@
...
-printf "\nWait for webhook to become available\n"
-waitForWebhook
-printf "\nApplying DynaKube CustomResource...\n"
-applyDynaKubeCR
-printf "\nAdding cluster to Dynatrace...\n"
-addK8sConfiguration
+bold=$(tput bold)
+normal=$(tput sgr0)
+
+echo ${bold}This script is deprecated - installation aborted.${normal}
```
chrismuellner, line 8: I don't think we should format anything in bold in the output. Just the normal text is ok.

chrismuellner (changes requested):

deploy/install.sh
gkrenn: Done

0sewa0 (approved):

PR #737, opened 2022-04-21
Add feature flag for ignoring proxy within oneAgent and activeGate

luhi-DT:
# Description
Currently we have the problem, when a proxy is set in the custom resource, it will get applied to all the components (operator, oneAgent and activeGate). This is so far intended, but if there is an activeGate with the purpose of routing agents traffic, then the oneAgent will also try to connect to the activeGate using the proxy, which causes problems

## How can this be tested?
Setup the operator and set a proxy in the customResource. This proxy will get applied to dtclient, oneAgents and activeGate (if defined in the same CR). If you add one or both featureFlags

- "operator.dynatrace.com/feature-activegate-ignore-proxy=true"
- "operator.dynatrace.com/feature-oneagent-ignore-proxy=true"

proxy should not be set for activeGate and/or oneAgent

## Checklist


src/controllers/activegate/reconciler/statefulset/statefulset.go
```diff
@@ -525,3 +525,7 @@ func isCustomPropertiesNilOrEmpty(customProperties *dynatracev1beta1.DynaKubeVal
 		(customProperties.Value == "" &&
 			customProperties.ValueFrom == "")
 }
+
+func needsProxy(stsProperties *statefulSetProperties) bool {
+	return !stsProperties.DynaKube.FeatureActiveGateIgnoreProxy() && stsProperties.HasProxy()
```
meik99, line 530: IIRC, `stsProperties` embeds the DynaKube, so the call can be simplified to 
```suggestion
	return !stsProperties.FeatureActiveGateIgnoreProxy() && stsProperties.HasProxy()
```

meik99: could then be further refactored into a property of DynaKube called, e.g., `ActiveGateNeedsProxy()`.
I.e.,
```
// In properties.go
func (dynakube *DynaKube) ActiveGateNeedsProxy() bool {
  return !stsProperties.FeatureActiveGateIgnoreProxy() && stsProperties.HasProxy()
}

// In statefulset.go
if stsProperties.ActiveGateNeedsProxy() {
```

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -316,7 +316,7 @@ func (controller *DynakubeController) reconcileActiveGate(ctx context.Context, d
 
 func (controller *DynakubeController) reconcileActiveGateProxySecret(ctx context.Context, dynakubeState *status.DynakubeState) bool {
 	gen := agproxysecret.NewActiveGateProxySecretGenerator(controller.client, controller.apiReader, dynakubeState.Instance.Namespace, log)
-	if dynakubeState.Instance.HasProxy() {
+	if !dynakubeState.Instance.FeatureActiveGateIgnoreProxy() && dynakubeState.Instance.HasProxy() {
```
meik99, line 319: The property mentioned before can then be used here too.  See the comment for the activegate proxy flag

src/controllers/dynakube/oneagent/daemonset/arguments.go
```diff
@@ -36,7 +36,7 @@ func (dsInfo *builderInfo) appendNetworkZoneArg(args []string) []string {
 }
 
 func (dsInfo *builderInfo) appendProxyArg(args []string) []string {
-	if dsInfo.hasProxy() {
+	if !dsInfo.instance.FeatureOneAgentIgnoreProxy() && dsInfo.hasProxy() {
```
meik99, line 39: can also be refactored into a Dynakube property. See the comment for the activegate proxy flag

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -316,7 +316,7 @@ func (controller *DynakubeController) reconcileActiveGate(ctx context.Context, d
 
 func (controller *DynakubeController) reconcileActiveGateProxySecret(ctx context.Context, dynakubeState *status.DynakubeState) bool {
```
meik99, line 317: No unit test was changed when this methods logic changed. Please edit or add an unit test accordingly

meik99 (changes requested):

0sewa0 (approved):

PR #744, opened 2022-04-26
Improve kuttl test suit

gkrenn:
# Description

Simple improvements for our kuttl tests: 
- Add makefile commands for kuttl
- Add check for required env variables in makefile
- Correctly set timeout in our teststeps
- Updated kind node images

## How can this be tested?

Run 
- `make create-kind-cluster` - Kind cluster should be created
- `make kuttl-oneagent` - All tests should pass

## Checklist


src/testing/kuttl/oneagent/hostMonitoring/01-errors.yaml
```diff
@@ -1,23 +1,23 @@
...
-  numberUnavailable: 2
----
-apiVersion: apps/v1
-kind: DaemonSet
-metadata:
-  name: dynakube-oneagent
-  namespace: dynatrace
-status:
-  numberUnavailable: 3
+#apiVersion: apps/v1
```
0sewa0, line 1: why comment out ?

gkrenn: Deleted now

Makefile
```diff
@@ -55,18 +55,26 @@ helm-test:
...
+
+delete-kind-cluster:
+	kind delete cluster --name kind
 
 kuttl-install:
 	hack/e2e/install-kuttl.sh
 
-kuttl-oneagent: deploy
+kuttl-oneagent: kuttl-check-mandatory-fields deploy
 	kubectl -n dynatrace wait pod --for=condition=ready -l app.kubernetes.io/component=webhook
```
mjgrzybek, line 68: Makefile shouldn't be used to prepare environment for tests if framework is able to do it on it's own.
Things like applying CRD, waiting for webhook etc should be defined in kuttl testsuite.

Makefile
```diff
@@ -55,18 +55,26 @@ helm-test:
...
 	kubectl kuttl test --config src/testing/kuttl/oneagent/oneagent-test.yaml
-# CLEAN-UP
+	# CLEAN-UP
 	kubectl delete dynakube --all -n dynatrace
-	kubectl -n dynatrace wait pod --for=delete -l app.kubernetes.io/component=oneagent --timeout=500s
+	-kubectl -n dynatrace wait pod --for=delete -l app.kubernetes.io/component=oneagent --timeout=500s
 	kubectl delete -f config/deploy/kubernetes/kubernetes-all.yaml
 
+kuttl-check-mandatory-fields:
+	hack/do_env_variables_exist.sh "APIURL APITOKEN PAASTOKEN"
```
mjgrzybek, line 76: Plz make is a first assert in kuttl and remove from Makefile

src/testing/kuttl/oneagent/oneagent-test.yaml
```diff
@@ -1,8 +1,9 @@
 apiVersion: kuttl.dev/v1beta1
 kind: TestSuite
 startKIND: false
-kindConfig: kind.yaml
+kindConfig: src/testing/kuttl/kind.yaml
 skipDelete: false
+skipClusterDelete: true
```
mjgrzybek, line 6: it's better to delete cluster here than in Makefile

mjgrzybek (changes requested):

Makefile
gkrenn: The problem is that the framework is not really able to do what we want.

There are 3 options with downsides that I could think of: 
- Run `make deploy` in the testsuit command section. => No option to delete the operator 
- Add manifests to `manifestsDirs` => No way to wait until our operator is running, => first tests will fail due to operator not yet ready
- Add command to every test directory => same code in every test folder. + the operator should be running during the whole test suite

Therefore I would stick to the makefile.

src/testing/kuttl/oneagent/oneagent-test.yaml
meik99: According to docs, this flag only skips deleting a mocked control plane or KIND clusters. The former doesn't matter since it's a mock, the latter should be discarded after a test run anyway. So I'd agree with @mjgrzybek here.

https://kuttl.dev/docs/testing/reference.html#testsuite

src/testing/kuttl/oneagent/appmonitoring/04-assert.yaml
```diff
@@ -0,0 +1,3 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 200
```
meik99, line 3: why are those steps needed?

Makefile
meik99: why would we need to assert that in every test case? The tests failing would be indication enough, wouldn't they?

Makefile
mjgrzybek: @gkrenn 
> * Run `make deploy` in the testsuit command section. => No option to delete the operator

Why is cleanup needed? Kind cluster is deleted after tests is done anyway

> * Add manifests to `manifestsDirs` => No way to wait until our operator is running, => first tests will fail due to operator not yet ready

AG tests contain assert checking if operator is fully operational, see: https://github.com/Dynatrace/dynatrace-operator/blob/master/src/testing/kuttl/activegate/tests/capabilities-walkthrough/10-setup.yaml#L4-L7

> Add command to every test directory => same code in every test folder. + the operator should be running during the whole test suite

Operator (running for whole test suite) can be started using `manifestsDir` from https://kuttl.dev/docs/testing/reference.html#testsuite

src/testing/kuttl/oneagent/oneagent-test.yaml
gkrenn: Agree

Makefile
gkrenn: The test failing would be enough, there would not be an advantage for pipeline tests

But it is a great advantage for manual testing, as everytime Michal or I started the tests we forgot to the set the variables  => therefore had to wait to get an error and wonder why the test failed.

So in my opinion, we should check that, even if the tests would be sufficient.

src/testing/kuttl/oneagent/appmonitoring/04-assert.yaml
meik99: I meant why are those steps needed in general? They do not test anything, they just add a timeout as the last test step. Does that wait for something to happen?

Makefile
```diff
@@ -55,22 +55,25 @@ helm-test:
 helm-lint:
 	cd config/helm && ./testing/lint.sh
 
+create-kind-cluster:
+	kind create cluster --config ./src/testing/kuttl/oneagent/kind.yaml
+
+delete-kind-cluster:
+	kind delete cluster --name kind
```
mjgrzybek, line 62: It can be removed because it's not used now.

gkrenn: Removed

Makefile
```diff
@@ -55,22 +55,19 @@ helm-test:
...
-kuttl-oneagent: deploy
-	kubectl -n dynatrace wait pod --for=condition=ready -l app.kubernetes.io/component=webhook
+kuttl-oneagent: kuttl-check-mandatory-fields
 	kubectl kuttl test --config src/testing/kuttl/oneagent/oneagent-test.yaml
-# CLEAN-UP
-	kubectl delete dynakube --all -n dynatrace
-	kubectl -n dynatrace wait pod --for=delete -l app.kubernetes.io/component=oneagent --timeout=500s
-	kubectl delete -f config/deploy/kubernetes/kubernetes-all.yaml
+
+kuttl-check-mandatory-fields:
```
0sewa0, line 69: smurt 👍

0sewa0 (approved):

PR #748, opened 2022-04-27
Updates csi garbage-collector logic to work with images

0sewa0:
# Description

The image cache needs to be cleaned up otherwise the we can flood the node's filesystem with many 200MB+ image dirs.

This deletes any cached image that haven't been pulled for 2 weeks.
Every pull for an image changes its `index.json` even if the image was already cached, so its a good indicator of usage.

## How can this be tested?
Well this is kinda hard, as for the "thing" to happen you need to wait for 2 weeks.
What you can do is change the time locally to 10 mins or something, 
build the operator and use it, then create a dynakube that has `codeModulesImage` set.

## Checklist

### DIFF
--- a/src/api/v1beta1/properties.go
+++ b/src/api/v1beta1/properties.go
@@ -233,6 +233,7 @@ func (dk *DynaKube) Version() string {
 	return ""
 }
 
+// The dynakube.Version is not take into account when using cloudNative to avoid confusion
 func (dynakube DynaKube) CodeModulesVersion() string {
 	if !dynakube.CloudNativeFullstackMode() && !dynakube.ApplicationMonitoringMode() {
 		return ""
@@ -241,7 +242,7 @@ func (dynakube DynaKube) CodeModulesVersion() string {
 		codeModulesImage := dynakube.CodeModulesImage()
 		return strings.Split(codeModulesImage, ":")[1]
 	}
-	if dynakube.Version() != "" {
+	if dynakube.Version() != "" && !dynakube.CloudNativeFullstackMode() {
 		return dynakube.Version()
 	}
 	return dynakube.Status.LatestAgentVersionUnixPaas
--- a/src/controllers/csi/gc/binaries.go
+++ b/src/controllers/csi/gc/binaries.go
@@ -7,7 +7,7 @@ import (
 	"github.com/spf13/afero"
 )
 
-func (gc *CSIGarbageCollector) runBinaryGarbageCollection(tenantUUID string, latestVersion string) {
+func (gc *CSIGarbageCollector) runBinaryGarbageCollection(pinnedVersions pinnedVersionSet, tenantUUID string, latestVersion string) {
 	fs := &afero.Afero{Fs: gc.fs}
 	gcRunsMetric.Inc()
 
@@ -27,7 +27,7 @@ func (gc *CSIGarbageCollector) runBinaryGarbageCollection(tenantUUID string, lat
 
 	for _, version := range storedVersions {
 		shouldDelete := isNotLatestVersion(version, latestVersion) &&
-			shouldDeleteVersion(version, usedVersions)
+			shouldDeleteVersion(version, usedVersions) && !pinnedVersions[version]
 
 		if shouldDelete {
 			binaryPath := gc.path.AgentBinaryDirForVersion(tenantUUID, version)
--- a/src/controllers/csi/gc/binaries.go
+++ b/src/controllers/csi/gc/binaries.go
@@ -27,7 +27,7 @@ func (gc *CSIGarbageCollector) runBinaryGarbageCollection(pinnedVersions pinnedV
 
 	for _, version := range storedVersions {
 		shouldDelete := isNotLatestVersion(version, latestVersion) &&
-			shouldDeleteVersion(version, usedVersions) && !pinnedVersions[version]
+			shouldDeleteVersion(version, usedVersions) && pinnedVersions.isNotPinned(version)
 
 		if shouldDelete {
 			binaryPath := gc.path.AgentBinaryDirForVersion(tenantUUID, version)
--- a/src/controllers/csi/gc/controller.go
+++ b/src/controllers/csi/gc/controller.go
@@ -7,8 +7,6 @@ import (
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	dtcsi "github.com/Dynatrace/dynatrace-operator/src/controllers/csi"
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/csi/metadata"
-	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube"
-	"github.com/Dynatrace/dynatrace-operator/src/dtclient"
 	"github.com/spf13/afero"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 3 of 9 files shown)


src/controllers/csi/gc/images.go
```diff
@@ -0,0 +1,87 @@
+package csigc
+
+import (
+	"os"
+	"path/filepath"
+	"time"
+
+	_os "github.com/Dynatrace/dynatrace-operator/src/controllers/csi/gc/os"
```
mjgrzybek, line 8: I don't like it starting with underscore but I have no better idea. `dt_os` maybe?
How about moving it to more general level, like ~~`dynatrace-operator/src/utils`~~ `dynatrace-operator/src/os`?

src/controllers/csi/gc/images_test.go
```diff
@@ -0,0 +1,57 @@
...
+		require.NoError(t, err)
+		require.NotNil(t, modTime)
+	})
+}
+
+func createTestImageCacheDir() afero.Fs {
+	fs := afero.NewMemMapFs()
+	_ = fs.MkdirAll(image.CacheDir, 0755)
+	_ = fs.Mkdir(filepath.Join(image.CacheDir, testImageDigest), 0755)
+	_ = afero.WriteFile(fs, filepath.Join(image.CacheDir, testImageDigest, "index.json"), []byte("{}"), 0644)
```
meik99, line 55: nitpick: those errors should be returned also and checked with `require.NoError`. Admittedly, since its only used in tests, that wouldn't make much of a difference

src/controllers/csi/gc/images.go
```diff
@@ -0,0 +1,79 @@
...
+	"path/filepath"
+	"time"
+
+	"github.com/Dynatrace/dynatrace-operator/src/installer/image"
+	"github.com/spf13/afero"
+)
+
+var maxImageAge = time.Hour * 24 * 14 // 14 days
+
+func (gc *CSIGarbageCollector) runImageGarbageCollection() {
```
meik99, line 14:
```suggestion
func (garbageCollector *CSIGarbageCollector) runImageGarbageCollection() {
```

meik99 (changes requested):

src/controllers/csi/gc/images.go
```diff
@@ -0,0 +1,79 @@
...
+	imageDirs, _ := getImageCacheDirs(gc.fs)
+	if imageDirs == nil {
+		return
+	}
+
+	toDelete := collectUnusedImageCaches(gc.fs, imageDirs)
+
+	deleteImageCaches(gc.fs, toDelete)
```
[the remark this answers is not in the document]
0sewa0: yes, but as its a super simple function, I would prefer to keep it like this as it needs less boilerplate to test

src/controllers/csi/gc/images.go
```diff
@@ -0,0 +1,79 @@
...
+	"time"
+
+	"github.com/Dynatrace/dynatrace-operator/src/installer/image"
+	"github.com/spf13/afero"
+)
+
+var maxImageAge = time.Hour * 24 * 14 // 14 days
+
+func (gc *CSIGarbageCollector) runImageGarbageCollection() {
+	imageDirs, _ := getImageCacheDirs(gc.fs)
```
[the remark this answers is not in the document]
0sewa0: yes, but as its a super simple function, I would prefer to keep it like this as it needs less boilerplate to test

src/controllers/csi/gc/images.go
```diff
@@ -0,0 +1,79 @@
...
+	return imageDirs, err
+}
+
+func collectUnusedImageCaches(fs afero.Fs, imageDirs []os.FileInfo) []string {
+	var toDelete []string
+	for _, imageDir := range imageDirs {
+		if !imageDir.IsDir() {
+			continue
+		}
+		modificationTime, err := getRelevantModificationTime(fs, imageDir.Name())
```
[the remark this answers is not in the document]
0sewa0: yes, but as its a super simple function, I would prefer to keep it like this as it needs less boilerplate to test

src/api/v1beta1/properties.go
```diff
@@ -241,7 +242,7 @@ func (dynakube DynaKube) CodeModulesVersion() string {
 		codeModulesImage := dynakube.CodeModulesImage()
 		return strings.Split(codeModulesImage, ":")[1]
 	}
-	if dynakube.Version() != "" {
+	if dynakube.Version() != "" && !dynakube.CloudNativeFullstackMode() {
```
meik99, line 245: Nitpick: the check if version is empty or not could be a property to enhance readability

```suggestion
	if dynakube.HasVersion() && !dynakube.CloudNativeFullstackMode() {
```

src/controllers/csi/gc/controller.go
```diff
@@ -59,35 +59,53 @@ func (gc *CSIGarbageCollector) Reconcile(ctx context.Context, request reconcile.
...
+	log.Info("running image garbage collection")
+	gc.runImageGarbageCollection()
 
 	return reconcileResult, nil
 }
+
+// getAllPinnedVersionsForTenantUUID returns all pinned versions for a given tenantUUID.
+// A pinned version is either the:
+// - image tag/digest set in the dynakube for the csi codeModules
+// - version set for in the dynakube for the csi codeModules. (currently not supported on cloudNative)
```
meik99, line 96:
```suggestion
// A pinned version is either:
// - the image tag or digest set in the custom resource
// - the version set in the custom resource if applicationMonitoring is used
```

src/controllers/csi/gc/images.go
```diff
@@ -0,0 +1,79 @@
...
+			toDelete = append(toDelete, filepath.Join(image.CacheDir, imageDir.Name()))
+		}
+	}
+	return toDelete
+}
+
+// getRelevantModificationTime returns the last modification time of an image in the cache
+// an image is not a single file but a directory of several files.
+// Most of which don't change when the image is tried to be pulled again.
+// The index.json of the image is a file that is modified in reliable way.(after every pull)
```
meik99, line 59:
```suggestion
// getRelevantModificationTime returns the last modification time of an extracted image in the cache.
// Most of the extracted files do not change after an update.
// Therefore, the index.json is checked for its modification time, as it is always updated after an image pull.
```

src/controllers/csi/gc/binaries.go
```diff
@@ -27,7 +27,7 @@ func (gc *CSIGarbageCollector) runBinaryGarbageCollection(tenantUUID string, lat
 
 	for _, version := range storedVersions {
 		shouldDelete := isNotLatestVersion(version, latestVersion) &&
-			shouldDeleteVersion(version, usedVersions)
+			shouldDeleteVersion(version, usedVersions) && !pinnedVersions[version]
```
meik99, line 30: nitpick: since `pinnedVersionSet` is already a type, it could be extended with a method such as `isNotPinned(version string)` which could be used in this condition to enhance readability.

```suggestion
			shouldDeleteVersion(version, usedVersions) && pinnedVersions.isNotPinned(version)
```

src/controllers/csi/gc/images.go
0sewa0: AFAIK the image is not really 'extracted' in the `data/cache/<digest>`, its just everything the image comes with, the blobs are still compressed

src/controllers/csi/gc/binaries.go
0sewa0: done

src/controllers/csi/gc/controller.go
0sewa0: done

meik99 (approved):

PR #754, opened 2022-04-28
Skip cache removal when file does not exist

chrismuellner:
# Description
Install agent binaries when installed version is `""` (empty), but the `<tenant>/bin/<target version>` directory exists. This means the previous attempt to install agent failed (e.g. OOM during unzip), only created the `bin` folder but did not update the database with the correct version.

## How can this be tested?
Dynakube that needs CSI driver is deployed:
1. CSI driver crashes during unzip (after creating `bin` folder with version)
2. CSI driver restarts and tries to update `ruxitagentproc.conf` without an installed version

## Checklist

### DIFF
--- a/src/controllers/csi/provisioner/agent.go
+++ b/src/controllers/csi/provisioner/agent.go
@@ -53,7 +53,8 @@ func (updater *agentUpdater) updateAgent(installedVersion, tenantUUID string, pr
 	targetVersion := updater.getOneAgentVersionFromInstance()
 	targetDir := updater.path.AgentBinaryDirForVersion(tenantUUID, targetVersion)
 
-	if _, err := updater.fs.Stat(targetDir); os.IsNotExist(err) || installedVersion == "" {
+	_, err := updater.fs.Stat(targetDir)
+	if os.IsNotExist(err) || installedVersion == "" {
 		log.Info("updating agent",
 			"target version", targetVersion,
 			"installed version", installedVersion,
--- a/src/controllers/csi/provisioner/agent.go
+++ b/src/controllers/csi/provisioner/agent.go
@@ -53,7 +53,7 @@ func (updater *agentUpdater) updateAgent(installedVersion, tenantUUID string, pr
 	targetVersion := updater.getOneAgentVersionFromInstance()
 	targetDir := updater.path.AgentBinaryDirForVersion(tenantUUID, targetVersion)
 
-	if _, err := updater.fs.Stat(targetDir); os.IsNotExist(err) {
+	if _, err := updater.fs.Stat(targetDir); os.IsNotExist(err) || installedVersion == "" {
 		log.Info("updating agent",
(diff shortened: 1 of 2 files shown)


src/controllers/csi/provisioner/processmoduleconfig.go
```diff
@@ -76,6 +77,7 @@ func (provisioner *OneAgentProvisioner) readProcessModuleConfigCache(tenantUUID
 		return nil, errors.Wrapf(err, "error closing file after reading processModuleConfigCache")
 	}
 
+	var processModuleConfig processModuleConfigCache
```
meik99, line 80: nitpick: should be declared on top to have a nicer overview of declared variables

src/controllers/csi/provisioner/processmoduleconfig.go
```diff
@@ -55,10 +55,11 @@ func (provisioner *OneAgentProvisioner) getProcessModuleConfig(dtc dtclient.Clie
...
 
 func (provisioner *OneAgentProvisioner) readProcessModuleConfigCache(tenantUUID string) (*processModuleConfigCache, error) {
-	var processModuleConfig processModuleConfigCache
 	processModuleConfigCacheFile, err := provisioner.fs.Open(provisioner.path.AgentRuxitProcResponseCache(tenantUUID))
 	if err != nil {
-		provisioner.removeProcessModuleConfigCache(tenantUUID)
+		if !os.IsNotExist(err) {
+			provisioner.removeProcessModuleConfigCache(tenantUUID)
+		}
 		return nil, err
```
meik99, line 63:
```suggestion
		if os.IsNotExist(err) {		
     		return nil, err
		}
		provisioner.removeProcessModuleConfigCache(tenantUUID)
```

Double negation is confusing

src/controllers/csi/provisioner/agent.go
```diff
@@ -53,7 +53,7 @@ func (updater *agentUpdater) updateAgent(installedVersion, tenantUUID string, pr
 	targetVersion := updater.getOneAgentVersionFromInstance()
 	targetDir := updater.path.AgentBinaryDirForVersion(tenantUUID, targetVersion)
 
-	if _, err := updater.fs.Stat(targetDir); os.IsNotExist(err) {
+	if _, err := updater.fs.Stat(targetDir); os.IsNotExist(err) || installedVersion == "" {
```
meik99, line 56: nitpick: since there is now a second condition to the error handling, it would be better to assign `err` before the condition instead of inside it.
```suggestion
    _, err := updater.fs.Stat(targetDir)
	if os.IsNotExist(err) || installedVersion == "" {
```

meik99 (changes requested):

waodim (approved):
apart from the points which @meik99 brought up, which are already resolved anyway, everything looks fine to me.

0sewa0 (approved):

PR #755, opened 2022-04-29
Remove yamls that are generated in `manifests` build target

mjgrzybek:
# Description
Removed files from the repo:
- `config/deploy/kubernetes/kubernetes-all.yaml`
- `config/deploy/openshift/openshift-all.yaml`

because they are generated in `manifests` build target.

## How can this be tested?
1. `make manifests`
2. verify if files are generated `find ./config -name "*-all.yaml"`

## Checklist


0sewa0:
AFAIK these are used by the pipelines, so it should not be removed

0sewa0:
@mjgrzybek 
I removed it once, 
then people screamed, 
I told them to use `make manifests`, 
but then they would have to clone the repo + need to have helm installed, 
they didn't like the idea 😭  
(I wanted it gone before it was cool 😎 )

PR #757, opened 2022-04-29
Create symlink from filesystem instead of latest version on Dynakube

chrismuellner:
# Description
Inspired by #729 
When there's a mismatch between latest version on the Dynakube and the version of the binaries download, the symlink would point to a different version than existed on the file system.

## How can this be tested?
Check `current` symlink on injected pods and verify it's valid. 

## Checklist

### DIFF
--- a/src/installer/symlink/symlink.go
+++ b/src/installer/symlink/symlink.go
@@ -1,0 +1,64 @@
+package symlink
+
+import (
+	iofs "io/fs"
+	"path/filepath"
+	"regexp"
+
+	"github.com/spf13/afero"
+)
+
+func CreateSymlinkForCurrentVersionIfNotExists(fs afero.Fs, targetDir string) error {
+	var relativeSymlinkPath string
+	var err error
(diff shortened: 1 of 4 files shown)


0sewa0 (approved):

src/installer/symlink/symlink.go
```diff
@@ -0,0 +1,64 @@
...
+			return nil
+		}
+		folderName := filepath.Base(path)
+		if regexp.MustCompile(versionRegexp).Match([]byte(folderName)) {
+			log.Info("found version", "version", folderName)
+			version = folderName
+			return iofs.ErrExist
+		}
+		return nil
+	}
```
meik99, line 59: Could be moved to its own function declaration, that would make the surrounding function leaner

chrismuellner: reasonable suggestion, but logic was just copied from `master` without changes. Don't want to introduce issues here.

PR #759, opened 2022-05-02
Add ClusterRoles for OLM activegate

luhi-DT:
# Description

ActiveGate need new SecurityContextConstraints, but just linking the ServiceAccount with the SCCs is not enough to have the corresponding permissions, when generating the CSV. Therefore if a an activeGate is deployed via OperatorHub, the statefulset cannot create pods, as it violates SCCs. ClusterRole and RoleBinding for dynatrace-activegate SA are only created if generated for OLM.

## How can this be tested?
```
make bundle VERSION=0.X.X PLATFORM=openshift
```
Should also include now SCCs for `dynatrace-activegate` and `dynatrace-kubernetes-monitoring`

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring.yaml
+++ b/config/helm/chart/default/templates/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring.yaml
@@ -79,4 +79,15 @@ rules:
       - /livez
     verbs:
       - get
+  {{- if eq (default false .Values.olm) true}}
+  - apiGroups:
+      - security.openshift.io
+    resourceNames:
+      - host
+      - privileged
+    resources:
+      - securitycontextconstraints
+    verbs:
+      - use
+  {{ end }}
 {{ end }}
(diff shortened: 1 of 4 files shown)


config/helm/chart/default/templates/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring.yaml
```diff
@@ -79,4 +79,15 @@ rules:
       - /livez
     verbs:
       - get
+  {{- if eq (default false .Values.olm) true}}
```
chrismuellner, line 82:
```suggestion
  {{- if .Values.olm }}
```

chrismuellner: guess we have that check everywhere, can be left as is

luhi-DT: We have that check everywhere, where we check for olm == true :)

0sewa0 (approved):

PR #760, opened 2022-05-02
Replace the latest oneagent version endpoint to use the versions endpoint

0sewa0:
# Description
The latest agent version endpoint we use can't be trusted to give us a version that we can actually download. (it provides the latest version for a different kind of installer).

So instead of that we use the get all versions endpoint and find the latest from that.

## How can this be tested?
Deploy cloudnative on a cluster where
`/deployment/installer/agent/unix/paas/latest/metainfo` returns a version that is not present in 
the output `/deployment​/installer​/agent​/versions​/unix/paas` 
and it shouldn't error anymore.

## Checklist

### DIFF
--- a/src/dtclient/agent_version.go
+++ b/src/dtclient/agent_version.go
@@ -10,7 +10,15 @@ import (
 
 // GetLatestAgentVersion gets the latest agent version for the given OS and installer type.
 func (dtc *dynatraceClient) GetLatestAgentVersion(os, installerType string) (string, error) {
-	versions, err := dtc.GetAgentVersions(os, installerType, arch.Flavor, arch.Arch)
+	var flavor string
+	// Default installer type has no "multidistro" flavor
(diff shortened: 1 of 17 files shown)


src/dtclient/agent_version.go
```diff
@@ -1,34 +1,34 @@
...
-
-	responseData, err := dtc.getServerResponseData(resp)
-	if err != nil {
-		return "", err
+	if len(versions) == 0 {
+		return "", errors.New("no agent versions found")
 	}
 
-	return dtc.readResponseForLatestVersion(responseData)
+	latestVersion := dtversion.VersionInfo{}
```
meik99, line 21: should be initialized with the first version in the array

meik99 (changes requested):

chrismuellner (approved):

PR #766, opened 2022-05-05
Remove unnecessary ports in manifests

waodim:
# Description

Removed unnecessary ports in the k8s manifest. The functionality of the webhook and the operator remains the same even without those ports.

## How can this be tested?
Simply apply the manifest without those ports and watch the behavior.

## Checklist


0sewa0:
run `make manifests` so the openshift_all will also get updated

0sewa0 (approved):
tried it out, noticed no problems (validation and pod mutattion worked as expected)

PR #783, opened 2022-05-10
Move github actions code into dedicated bash files

gkrenn:
# Description

Bash scripts are defined in the yaml file of the Actions directly. This means there's no syntax highlighting and testing them is problematic.

Therefore we moved ci scripts to dedicated shell files.

## How can this be tested?

Look on our action pipeline, if it still works. 

## Checklist


hack/build/ci/install_go_test_dependencies.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+installCGODependencies() {
```
chrismuellner, line 3: please move each function into a separate script

chrismuellner (changes requested):

hack/build/ci/install_go_test_dependencies.sh
gkrenn: The idea behind this "partial" split into only some scripts was that we have one script for each step in the ci pipeline. But yeah, agree that we should ev. split functions into files if they are called from the outside, and even not only from the ci pipeline.

gkrenn: Moved now all the functions into own scripts, besides of some functions in prepareVariablesAndDependencies. But this functions are just there to make the code look better and are not called from outside the script.

.github/workflows/ci.yml
```diff
@@ -61,13 +60,12 @@ jobs:
       - name: Download dependencies
         id: depdownload
         run: |
-          sudo apt-get update
-          sudo apt-get install -y libdevmapper-dev libbtrfs-dev libgpgme-dev
+          hack/build/ci/install_cgo_dependencies.sh
+          hack/build/ci/install_kubebuilder.sh
```
chrismuellner, line 64: why do we need kubebuilder now in these steps?

.github/workflows/ci.yml
```diff
@@ -114,73 +111,36 @@ jobs:
             ${{ runner.os }}-go-build-
       - name: Prepare
         id: prep
+        env:
+          GITHUB_EVENT_NAME: ${{ github.event_name }}
```
chrismuellner, line 115: `GITHUB_EVENT_NAME` should already exist: [Default Environment Variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables)

hack/build/ci/create_go_build_args.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+version=$1
```
chrismuellner, line 3: could use checks if the parameters are not empty/exist

hack/build/push_image.sh
```diff
@@ -7,24 +7,16 @@ if [[ ! "${TAG}" ]]; then
...
 
-
-args="${go_build_args[*]}"
 if [[ "${LOCALBUILD}" ]]; then
   export CGO_ENABLED=1
   export GOOS=linux
   export GOARCH=amd64
 
-  go build "$args" -o ./build/_output/bin/dynatrace-operator ./src/cmd/operator/
+  go build "$go_build_args" -o ./build/_output/bin/dynatrace-operator ./src/cmd/operator/
```
chrismuellner, line 19: `push_image` shouldn't build the image, but that's a different problem.
since `ci/create_go_guild_args` is not only used for ci, it should be moved one folder up.

chrismuellner (changes requested):

.github/workflows/ci.yml
gkrenn: Previously this was done in `test.sh` - separated the installation into an own file.

.github/workflows/ci.yml
```diff
@@ -61,13 +60,12 @@ jobs:
...
-          sudo apt-get update
-          sudo apt-get install -y libdevmapper-dev libbtrfs-dev libgpgme-dev
+          hack/build/ci/install_cgo_dependencies.sh
+          hack/build/ci/install_kubebuilder.sh
       - name: Run Unit tests
         id: unittest
         run: |
-          chmod +x ./hack/build/test.sh
-          ./hack/build/test.sh
+          go test -cover -tags integration,containers_image_storage_stub ./...
```
chrismuellner, line 68: this could also be moved into a separate file.
having no more bash logic in the action manifests is the goal here!

hack/build/push_image.sh
```diff
@@ -7,24 +7,16 @@ if [[ ! "${TAG}" ]]; then
 fi
 
 commit=$(git rev-parse HEAD)
```
chrismuellner, line 9: does `GITHUB_SHA` (default env variable) result in the same value?

chrismuellner (changes requested):

.github/workflows/ci.yml
gkrenn: Removed

hack/build/ci/create_go_build_args.sh
gkrenn: implemented

hack/build/push_image.sh
gkrenn: agree

moved

hack/build/push_image.sh
gkrenn: this script does not run in our github ci, therefore `GITHUB_SHA` does not exist here

.github/workflows/ci.yml
gkrenn: I'm not really a friend of creating a bash script for one line, but created it for now

.github/workflows/ci.yml
```diff
@@ -115,72 +112,33 @@ jobs:
...
-          # Get licenses if no cache exists
-          if ! [ -d ./third_party_licenses ]; then
-            go get github.com/google/go-licenses && go-licenses save ./... --save_path third_party_licenses --force
-          fi
-
-          # Get dependencies
-          go get -d ./...
-          ls -la $HOME/go/pkg/mod
-          cp -r $HOME/go/pkg/mod .
+          hack/build/ci/prepare_variables_and_download_go_build_deps.sh
```
meik99, line 115:
```suggestion
          hack/build/ci/prepare-build-environment.sh
```

Generally I am more a fan of dashes instead of underscores, because if text is selected with Ctrl+Shift the selection is more granular since it stops at dashes. However, that's just a personal preference

.github/workflows/ci.yml
```diff
@@ -216,22 +175,8 @@ jobs:
       - name: Upload image to Quay
         env:
           # reuse variables defined during build job
-          LABELS: ${{ needs.builddockerimage.outputs.labels }}
-          TAG: ${{ needs.builddockerimage.outputs.version }}
+          TAG: ${{ needs.builddockerimage.outputs.docker_image_tag }}
         run: |
           docker load --input /tmp/operator-amd64.tar
```
meik99, line 180: Why was does line not moved into the script?

gkrenn:
Messed up with git history, will create a new PR

PR #785, opened 2022-05-10
Improve yamls generation in Makefile 

mjgrzybek:
## Description
Basic refactoring of Makefile around yamls generation

- use variables
- remove temp variables
- better targets granularity
- ~`grep -v` replaced with `sed`~
- more descriptive yaml filenames
- no more overriding of `kubernetes.yaml`

## How can this be tested?
`make manifests`

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,18 @@ MASTER_IMAGE = quay.io/dynatrace/dynatrace-operator:snapshot
 BRANCH_IMAGE = quay.io/dynatrace/dynatrace-operator:snapshot-$(shell git branch --show-current | sed "s/[^a-zA-Z0-9_-]/-/g")
 OLM_IMAGE ?= registry.connect.redhat.com/dynatrace/dynatrace-operator:v${VERSION}
 
+KUBERNETES_OTHERS_YAML=config/deploy/kubernetes/kubernetes-others.yaml
+KUBERNETES_CRD_YAML=config/deploy/kubernetes/kubernetes-crd.yaml
+KUBERNETES_CSIDRIVER_YAML=config/deploy/kubernetes/kubernetes-csidriver.yaml
+KUBERNETES_OLM_YAML=config/deploy/kubernetes/kubernetes-olm.yaml
+KUBERNETES_ALL_YAML=config/deploy/kubernetes/kubernetes-all.yaml
+
+OPENSHIFT_OTHERS_YAML=config/deploy/openshift/openshift-others.yaml
+OPENSHIFT_CRD_YAML=config/deploy/openshift/openshift-crd.yaml
+OPENSHIFT_CSIDRIVER_YAML=config/deploy/openshift/openshift-csidriver.yaml
+OPENSHIFT_OLM_YAML=config/deploy/openshift/openshift-olm.yaml
+OPENSHIFT_ALL_YAML=config/deploy/openshift/openshift-all.yaml
+
 # Options for 'bundle-build'
 ifneq ($(origin CHANNELS), undefined)
 BUNDLE_CHANNELS := --channels=$(CHANNELS)
@@ -111,23 +123,30 @@ push-tagged-image: push-image
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: manifests-k8s manifests-ocp
 
-manifests-k8s: generate-crd kustomize
+manifests-k8s: manifests-k8s-crd manifests-k8s-csidriver
+	cp "$(KUBERNETES_CRD_YAML)" "$(KUBERNETES_OLM_YAML)"
+	cat "$(KUBERNETES_CRD_YAML)" "$(KUBERNETES_CSIDRIVER_YAML)" > "$(KUBERNETES_ALL_YAML)"
+
+manifests-k8s-crd: generate-crd controller-gen kustomize
 	# Create directories for manifests if they do not exist
 	mkdir -p config/deploy/kubernetes
 
-	# Generate kubernetes.yaml
+	# Generate kubernetes-crd.yaml
 	helm template dynatrace-operator config/helm/chart/default \
 		--namespace dynatrace \
 		--set platform="kubernetes" \
 		--set manifests=true \
 		--set olm="${OLM}" \
 		--set autoCreateSecret=false \
-		--set operator.image="$(MASTER_IMAGE)" > config/deploy/kubernetes/kubernetes.yaml
-	grep -v 'app.kubernetes.io/managed-by' config/deploy/kubernetes/kubernetes.yaml > config/deploy/kubernetes/tmp.yaml
-	grep -v 'helm.sh' config/deploy/kubernetes/tmp.yaml > config/deploy/kubernetes/kubernetes.yaml
-	rm config/deploy/kubernetes/tmp.yaml
+		--set operator.image="$(MASTER_IMAGE)" > "$(KUBERNETES_OTHERS_YAML)"
 
-	# Generate kubernetes-csi.yaml
+	sed -i "/app.kubernetes.io\/managed-by/d" "$(KUBERNETES_OTHERS_YAML)" 
+	sed -i "/helm.sh/d" "$(KUBERNETES_OTHERS_YAML)" 
+
+	$(KUSTOMIZE) build config/crd | cat - "$(KUBERNETES_OTHERS_YAML)" > "$(KUBERNETES_CRD_YAML)"
+
+manifests-k8s-csidriver:
+	# Generate kubernetes-csidriver.yaml
 	helm template dynatrace-operator config/helm/chart/default \
 		--namespace dynatrace \
 		--set partial="csi" \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -111,42 +123,48 @@ push-tagged-image: push-image
...
 		--set olm="${OLM}" \
 		--set autoCreateSecret=false \
-		--set operator.image="$(MASTER_IMAGE)" > config/deploy/kubernetes/kubernetes.yaml
-	grep -v 'app.kubernetes.io/managed-by' config/deploy/kubernetes/kubernetes.yaml > config/deploy/kubernetes/tmp.yaml
-	grep -v 'helm.sh' config/deploy/kubernetes/tmp.yaml > config/deploy/kubernetes/kubernetes.yaml
-	rm config/deploy/kubernetes/tmp.yaml
+		--set operator.image="$(MASTER_IMAGE)" > "$(KUBERNETES_OTHERS_YAML)"
+
+	sed -i "/app.kubernetes.io\/managed-by/d" "$(KUBERNETES_OTHERS_YAML)" 
+	sed -i "/helm.sh/d" "$(KUBERNETES_OTHERS_YAML)"
```
meik99, line 144: `sed -i` does not work on Mac

Makefile
```diff
@@ -111,42 +123,48 @@ push-tagged-image: push-image
...
-		--set operator.image="$(MASTER_IMAGE)" > config/deploy/kubernetes/kubernetes-csi.yaml
-	grep -v 'app.kubernetes.io/managed-by' config/deploy/kubernetes/kubernetes-csi.yaml > config/deploy/kubernetes/tmp.yaml
-	grep -v 'helm.sh' config/deploy/kubernetes/tmp.yaml > config/deploy/kubernetes/kubernetes-csi.yaml
-	rm config/deploy/kubernetes/tmp.yaml
+		--set operator.image="$(MASTER_IMAGE)" > "$(KUBERNETES_CSIDRIVER_YAML)"
 
-	$(KUSTOMIZE) build config/crd | cat - config/deploy/kubernetes/kubernetes.yaml > temp
-	mv temp config/deploy/kubernetes/kubernetes.yaml
+	sed -i "/app.kubernetes.io\/managed-by/d" "$(KUBERNETES_CSIDRIVER_YAML)" 
+	sed -i "/helm.sh/d" "$(KUBERNETES_CSIDRIVER_YAML)"
```
meik99, line 160: sed -i does not work on Mac

Makefile
```diff
@@ -158,11 +176,14 @@ manifests-ocp: generate-crd controller-gen kustomize
...
 		--set autoCreateSecret=false \
 		--set createSecurityContextConstraints="true" \
-		--set operator.image="$(MASTER_IMAGE)" > config/deploy/openshift/openshift.yaml
-	grep -v 'app.kubernetes.io/managed-by' config/deploy/openshift/openshift.yaml > config/deploy/openshift/tmp.yaml
-	grep -v 'helm.sh' config/deploy/openshift/tmp.yaml > config/deploy/openshift/openshift.yaml
-	rm config/deploy/openshift/tmp.yaml
+		--set operator.image="$(MASTER_IMAGE)" > "$(OPENSHIFT_OTHERS_YAML)"
+
+	sed -i "/app.kubernetes.io\/managed-by/d" "$(OPENSHIFT_OTHERS_YAML)"
+	sed -i "/helm.sh/d" "$(OPENSHIFT_OTHERS_YAML)"
```
meik99, line 182: sed -i does not work on Mac

Makefile
```diff
@@ -172,16 +193,10 @@ manifests-ocp: generate-crd controller-gen kustomize
...
-	rm config/deploy/openshift/tmp.yaml
-
-	$(KUSTOMIZE) build config/crd | cat - config/deploy/openshift/openshift.yaml > temp
-	mv temp config/deploy/openshift/openshift.yaml
+		--set operator.image="$(MASTER_IMAGE)" > "$(OPENSHIFT_CSIDRIVER_YAML)"
 
-	cat config/deploy/openshift/openshift.yaml > config/deploy/openshift/openshift-olm.yaml
-	cat config/deploy/openshift/openshift.yaml config/deploy/openshift/openshift-csi.yaml > config/deploy/openshift/openshift-all.yaml
+	sed -i "/app.kubernetes.io\/managed-by/d" "$(OPENSHIFT_CSIDRIVER_YAML)" 
+	sed -i "/helm.sh/d" "$(OPENSHIFT_CSIDRIVER_YAML)"
```
meik99, line 199: sed -i does not work on Mac

meik99 (changes requested):

Makefile
mjgrzybek: Please don't flood with same comment every line, this pollutes whole review, thx

Makefile
mjgrzybek: Marking as resolved to make only single instance of this comment visible.

Makefile
mjgrzybek: Marking as resolved to make only single instance of this comment visible.

Makefile
mjgrzybek: Marking as resolved to make only single instance of this comment visible.

Makefile
```diff
@@ -13,6 +13,18 @@ MASTER_IMAGE = quay.io/dynatrace/dynatrace-operator:snapshot
...
 OLM_IMAGE ?= registry.connect.redhat.com/dynatrace/dynatrace-operator:v${VERSION}
 
+KUBERNETES_OTHERS_YAML=config/deploy/kubernetes/kubernetes-others.yaml
+KUBERNETES_CRD_YAML=config/deploy/kubernetes/kubernetes-crd.yaml
+KUBERNETES_CSIDRIVER_YAML=config/deploy/kubernetes/kubernetes-csidriver.yaml
+KUBERNETES_OLM_YAML=config/deploy/kubernetes/kubernetes-olm.yaml
+KUBERNETES_ALL_YAML=config/deploy/kubernetes/kubernetes-all.yaml
+
+OPENSHIFT_OTHERS_YAML=config/deploy/openshift/openshift-others.yaml
+OPENSHIFT_CRD_YAML=config/deploy/openshift/openshift-crd.yaml
```
meik99, line 23:
```suggestion
OPENSHIFT_YAML=config/deploy/openshift/openshift.yaml
```

Makefile
```diff
@@ -13,6 +13,18 @@ MASTER_IMAGE = quay.io/dynatrace/dynatrace-operator:snapshot
 BRANCH_IMAGE = quay.io/dynatrace/dynatrace-operator:snapshot-$(shell git branch --show-current | sed "s/[^a-zA-Z0-9_-]/-/g")
 OLM_IMAGE ?= registry.connect.redhat.com/dynatrace/dynatrace-operator:v${VERSION}
 
+KUBERNETES_OTHERS_YAML=config/deploy/kubernetes/kubernetes-others.yaml
+KUBERNETES_CRD_YAML=config/deploy/kubernetes/kubernetes-crd.yaml
```
meik99, line 17:
```suggestion
KUBERNETES_YAML=config/deploy/kubernetes/kubernetes.yaml
```

Makefile
mjgrzybek: @meik99 can you try: `brew install gnu-sed` ?
src: https://stackoverflow.com/questions/25124969/sed-replacement-command-not-working-on-mac?msclkid=5bf06597d05011ec875a7ffeb5fb8f85#comment119427508_41029130

Makefile
```diff
@@ -13,6 +13,18 @@ MASTER_IMAGE = quay.io/dynatrace/dynatrace-operator:snapshot
 BRANCH_IMAGE = quay.io/dynatrace/dynatrace-operator:snapshot-$(shell git branch --show-current | sed "s/[^a-zA-Z0-9_-]/-/g")
 OLM_IMAGE ?= registry.connect.redhat.com/dynatrace/dynatrace-operator:v${VERSION}
 
+KUBERNETES_OTHERS_YAML=config/deploy/kubernetes/kubernetes-others.yaml
```
meik99, line 16: I would suggest using a tmp file to do the file manipulations which can be discarded afterwards.
There is no need as far as I can see to keep a file with a specific name

meik99 (changes requested):

Makefile
meik99: I don't have a mac

0sewa0: I will try 😉

meik99: > Please don't flood with same comment every line, this pollutes whole review, thx

I do that so occurrences of the same detail are not overlooked and they can be ticked off one by one

mjgrzybek: > > Please don't flood with same comment every line, this pollutes whole review, thx
> 
> I do that so occurrences of the same detail are not overlooked and they can be ticked off one by one

I'd suggest using general comment instead.

Makefile
mjgrzybek: These intermediate results are git-ignored and it allows developer to understand the intermediate results. I'd prefer to leave it as it is now.

Makefile
luhi-DT: I think @meik99 is right, as CRD makes me think of there is only the CRD included.

Makefile
luhi-DT: I think @mjgrzybek is right here, they are git-ignored and now you can actually see the results in each and check, in which file an error happened in case

Makefile
mjgrzybek: Actually this was my intention because I overlooked the fact it's actually `CRD`+`OTHERS` merged :see_no_evil:. Thanks for pointing it. 
Anyway, we should call these files different than "kubernetes.yaml" in the future because it means literally nothing.

luhi-DT: At least for me it makes sense, since `kubernetes.yaml` sounds to me like the default installation file, i would apply to run the operator (everything which is needed to run on kubernetes) and thats what it actually is. I would suggest to keep it as `kubernetes.yaml` and  `openshift.yaml` for now :)

Makefile
```diff
@@ -111,42 +124,49 @@ push-tagged-image: push-image
...
+manifests-k8s: manifests-crd manifests-k8s-csidriver
+	cp "$(KUBERNETES_CRD_AND_OTHERS_YAML)" "$(KUBERNETES_OLM_YAML)"
+	cat "$(KUBERNETES_CRD_AND_OTHERS_YAML)" "$(KUBERNETES_CSIDRIVER_YAML)" > "$(KUBERNETES_ALL_YAML)"
+
+manifests-crd: generate-crd controller-gen kustomize
 	# Create directories for manifests if they do not exist
 	mkdir -p config/deploy/kubernetes
 
-	# Generate kubernetes.yaml
+	# Generate kubernetes-crd.yaml
```
luhi-DT, line 135: small nitpick its `kubernetes.yaml` now

PR #786, opened 2022-05-13
Add basic support for activeGate auth token

luhi-DT:
# Description
This change introduces the activeGate auth token support. It basically adds the ability to create an auth token within the operator if the feature flag `feature.dynatrace.com/enable-activegate-authtoken: "true"` is set on the DynaKube.
This is the first PR of the whole activeGate auth token support feature, as token rotation will be included at a later stage.

## How can this be tested?
1. Create an API Token with the correct permission `activeGateTokenManagement.create`
2. Create a Dynakube that also creates an activeGate and add `feature.dynatrace.com/enable-activegate-authtoken: "true"` to it
3. check if the activeGate gets the correct secret mounted and check the logs of the activeGate if it uses the created auth Token

## Checklist


src/api/v1beta1/properties.go
```diff
@@ -359,6 +364,11 @@ func (dk *DynaKube) HostGroup() string {
 	return hostGroup
 }
 
+// UseActiveGateAuthToken returns if the activeGate should get an authToken mounted
```
0sewa0, line 367:
```suggestion
// UseActiveGateAuthToken returns if the activeGate should get an authToken mounted
// By default its should be used if the `dk.spec.apiToken` has the correct scopes,
// but otherwise the user can also opted out using a feature-flag
```

0sewa0 (changes requested):
My reviews aside, I reeally don't like the idea of having both a status field based on the scopes and a feature-flag to turn the thing on-off.
This means we have 2 different knobs that kinda do the same thing. (and rather confusing)
The user can opt out 2 different ways, by not having the scopes for their token OR adding the feature-flag.
(I guess) most users still have to opt-in, because they need to update the tokens.

I know the reason for having this is so that the Deployment Page just creates a token with the correct scopes by default and the authToken will be used unless explicitly turned off.
Couldn't we have just 1 feature flag to turn the thing on and the Deployment page could add that annotation to the dynakube by default instead ? 

In this case, if turned on and scopes are not correct, we complain (like with the automatic kubernetes monitoring registration).
(And in the future when having this new scope is the norm, we could flip the default value of the feature-flag to true, doing so the Deployment page is safe as even if the still set the feature-flag because it will only be redundant)

src/api/v1beta1/properties.go
0sewa0: (writing this comment made me write the little essay above 😛 )

mjgrzybek: it's an essay already, so actual scope- and feature-flag- names can be mentioned here explicitly

src/controllers/dynakube/dtclient_reconciler.go
```diff
@@ -158,6 +159,8 @@ func (r *DynatraceClientReconciler) Reconcile(ctx context.Context, instance *dyn
 		})
 	}
 
+	tokens[0].OptionalScopes = append(tokens[0].Scopes, dtclient.TokenScopeActiveGateTokenCreate)
```
mjgrzybek, line 162: `tokens` should be a `map[TokenType]tokenConfig` instead of `[]tokenConfig` to avoid relying on order in the slice

```
type TokenType int64

const (
	ApiToken TokenType = iota
	PaasToken 
)
```

mjgrzybek (changes requested):

src/controllers/dynakube/activegate/secrets/authtoken_reconciler.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (r *AuthTokenReconciler) createSecret(secretData map[string][]byte) (*corev1.Secret, error) {
+	secret := kubeobjects.CreateSecret(r.instance.ActiveGateAuthTokenSecret(), r.instance.Namespace, secretData)
+	if err := controllerutil.SetControllerReference(r.instance, secret, r.scheme); err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	err := r.Create(context.TODO(), secret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create secret '%s': %w", extendWithAGSecretSuffix(r.instance.Name), err)
```
meik99, line 85:
```suggestion
		return nil, errors.Errorf("failed to create secret '%s': %w", extendWithAGSecretSuffix(r.instance.Name), err)
```

src/controllers/dynakube/activegate/secrets/authtoken_reconciler.go
```diff
@@ -0,0 +1,88 @@
...
+		&config)
+	if k8serrors.IsNotFound(err) {
+		log.Info("creating activeGateAuthToken secret")
+		agSecretData, err := r.getActiveGateAuthToken()
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch activeGateAuthToken: %w", err)
+		}
+		return r.createSecret(agSecretData)
+	}
+	return &config, err
```
meik99, line 64:
```suggestion
	return &config, errors.WithStack(err)
```

src/controllers/dynakube/activegate/secrets/authtoken_reconciler.go
```diff
@@ -0,0 +1,88 @@
...
+func (r *AuthTokenReconciler) createSecretIfNotExists() (*corev1.Secret, error) {
+	var config corev1.Secret
+	err := r.apiReader.Get(context.TODO(),
+		client.ObjectKey{Name: r.instance.ActiveGateAuthTokenSecret(), Namespace: r.instance.Namespace},
+		&config)
+	if k8serrors.IsNotFound(err) {
+		log.Info("creating activeGateAuthToken secret")
+		agSecretData, err := r.getActiveGateAuthToken()
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch activeGateAuthToken: %w", err)
```
meik99, line 60:
```suggestion
			return nil, errors.Errorf("failed to fetch activeGateAuthToken: %w", err)
```

src/controllers/dynakube/activegate/secrets/authtoken_reconciler.go
```diff
@@ -0,0 +1,88 @@
...
+		instance:  instance,
+		apiToken:  apiToken,
+		dtc:       dtc,
+	}
+}
+
+func (r *AuthTokenReconciler) Reconcile() error {
+	_, err := r.createSecretIfNotExists()
+	if err != nil {
+		return fmt.Errorf("failed to create or update activeGateAuthToken secret: %w", err)
```
meik99, line 45:
```suggestion
		return errors.Errorf("failed to create activeGateAuthToken secret: %w", err)
```

src/controllers/dynakube/activegate/secrets/tenant_info_reconciler.go
```diff
@@ -100,7 +99,7 @@ func (r *TenantSecretReconciler) updateSecretIfOutdated(secret *corev1.Secret, d
 }
 
 func (r *TenantSecretReconciler) createSecret(secretData map[string][]byte) (*corev1.Secret, error) {
-	secret := BuildAGSecret(r.instance, secretData)
+	secret := kubeobjects.CreateSecret(extendWithAGSecretSuffix(r.instance.Name), r.instance.Namespace, secretData)
```
meik99, line 102: create the struct directly here or rename the function

src/controllers/dynakube/dtclient_reconciler.go
```diff
@@ -279,6 +285,15 @@ func (r *DynatraceClientReconciler) setAndLogCondition(conditions *[]metav1.Cond
...
 }
 
+func (r *DynatraceClientReconciler) HandleOptionalScope(optionalScopes []string, tokenScopes dtclient.TokenScopes) {
+	for _, scope := range optionalScopes {
+		switch scope {
+		case dtclient.TokenScopeActiveGateTokenCreate:
+			r.status.ActiveGate.UseAuthToken = tokenScopes.Contains(scope)
+		}
+	}
+}
```
meik99, line 295:
```suggestion
func (r *DynatraceClientReconciler) updateUseAuthTokenStatus(tokenScopes dtclient.TokenScopes) {
	    r.status.ActiveGate.UseAuthToken = tokenScopes.Contains(dtclient.TokenScopeActiveGateTokenCreate)
}
```

src/controllers/dynakube/dtclient_reconciler.go
```diff
@@ -226,6 +230,8 @@ func (r *DynatraceClientReconciler) CheckToken(dtc dtclient.Client, token tokenC
 		return true
 	}
 
+	r.HandleOptionalScope(token.OptionalScopes, ss)
```
meik99, line 233: please rename the variables `s` and `ss` to something meaningful

src/dtclient/activegate_auth_token.go
```diff
@@ -0,0 +1,82 @@
...
+func (dtc *dynatraceClient) GetActiveGateAuthToken(dynakubeName string) (*ActiveGateAuthTokenInfo, error) {
+	body := &ActiveGateAuthTokenParams{
+		Name:           dynakubeName,
+		SeedToken:      false,
+		ActiveGateType: activeGateType,
+	}
+	bodyData, err := json.Marshal(body)
+	if err != nil {
+		return nil, err
+	}
```
meik99, line 35: can be moved into its own function

src/dtclient/activegate_auth_token.go
```diff
@@ -0,0 +1,82 @@
...
+	)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	response, err := dtc.httpClient.Do(request)
+	if err != nil {
+		log.Info("failed to retrieve ActiveGateAuthToken")
+		return nil, err
+	}
```
meik99, line 51: can be moved into its own function

src/dtclient/activegate_auth_token.go
```diff
@@ -0,0 +1,82 @@
...
+
+	data, err := dtc.getServerResponseData(response)
+	if err != nil {
+		return nil, dtc.handleErrorResponseFromAPI(data, response.StatusCode)
+	}
+
+	authTokenInfo, err := dtc.readResponseForActiveGateAuthToken(data)
+	if err != nil {
+		return nil, err
+	}
```
meik99, line 68: can be moved into its own function

src/dtclient/dynatrace_client.go
```diff
@@ -76,6 +76,21 @@ func (dtc *dynatraceClient) makeRequest(url string, tokenType tokenType) (*http.
...
+	}
+	req.Header.Add("Accept", "application/json")
+	req.Header.Add("Authorization", fmt.Sprintf("Api-Token %s", apiToken))
+
+	if method == http.MethodPost {
+		req.Header.Add("Content-Type", "application/json")
+	}
+
+	return req, nil
+}
```
meik99, line 92: good move

meik99 (changes requested):

src/controllers/dynakube/dtclient_reconciler.go
luhi-DT: Good suggestion, thanks!

src/controllers/certificates/certificate_secret.go
```diff
@@ -31,7 +31,7 @@ func (certSecret *certificateSecret) setSecretFromReader(ctx context.Context, ap
 		return errors.WithStack(err)
 	}
 	if secret == nil {
-		secret = kubeobjects.CreateEmptySecret(buildSecretName(), namespace)
+		secret = kubeobjects.CreateSecret(buildSecretName(), namespace, map[string][]byte{})
```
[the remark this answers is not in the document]
luhi-DT: It saves a few lines of code, I will keep it

src/kubeobjects/secret.go
```diff
@@ -114,13 +114,14 @@ func GetSecret(ctx context.Context, apiReader client.Reader, name string, namesp
...
 
-func CreateEmptySecret(name string, namespace string) *corev1.Secret {
+func CreateSecret(name string, namespace string, data map[string][]byte) *corev1.Secret {
 	return &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
 			Namespace: namespace,
 		},
-		Data: map[string][]byte{},
+		Type: corev1.SecretTypeOpaque,
```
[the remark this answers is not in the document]
luhi-DT: Good input, thanks

src/dtclient/activegate_auth_token.go
```diff
@@ -0,0 +1,101 @@
...
+
+	authTokenInfo, err := dtc.handleAuthTokenResponse(response)
+	if err != nil {
+		log.Info("failed to handle ag-auth-token response")
+		return nil, err
+	}
+
+	return authTokenInfo, nil
+}
```
meik99, line 46: beautiful :heart:

src/api/v1beta1/dynakube_status.go
```diff
@@ -94,6 +94,7 @@ var _ VersionStatusNamer = (*ActiveGateStatus)(nil)
 
 type ActiveGateStatus struct {
 	VersionStatus `json:",inline"`
+	UseAuthToken  bool `json:"useAuthToken,omitempty"`
```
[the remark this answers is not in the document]
luhi-DT: dont need it anymore

meik99 (approved):

0sewa0 (approved):

PR #794, opened 2022-05-20
Fix error masking in standalone runner

luhi-DT:
# Description
Currently our error masking within the standalone runner is broken. The runner always logs that an error has been masked even though there is no error. This happened because, in the check for masking the error `consumeErrorIfNecessary()`, we do not check for the error itself but for the pointer, which is never nil. Additionally the defer function set the wrong variable to nil because `defer` runs a function **after** the return statement but **before** the function is actually returned, thus enabling modify returned results, but only named return results can be accessed normally. This caused that the error was set to nil, but not actual returned error, because go already evaluated the return result.

More here: https://stackoverflow.com/questions/48680222/what-is-the-difference-between-named-return-value-and-normal-return-value

## How can this be tested?
**Before this PR**
Deploy cloudNative or applicationMonitoring on a cluster and deploy sample apps. If you now check the logs of the initContainer, you can see the error log at the end of the run. Additionally if an error happens within the run and the failure policy is set to `silent`, the sample app would crash, because the initContainer is not finished successfully. 

**After this PR**
Deploy cloudNative or applicationMonitoring on a cluster and deploy sample apps. If you now check the logs of the initContainer, should not see any false error log anymore. Additionally if an error happens within the run and the failure policy is set to `silent`, the sample app is not crashing anymore and the initContainer finishes with 0 even though there was an error. 

## Checklist

### DIFF
--- a/src/standalone/run.go
+++ b/src/standalone/run.go
@@ -57,32 +57,32 @@ func NewRunner(fs afero.Fs) (*Runner, error) {
 	}, nil
 }
 
-func (runner *Runner) Run() error {
+func (runner *Runner) Run() (resultedError error) {
 	log.Info("standalone agent init started")
-	var err error
-	defer runner.consumeErrorIfNecessary(&err)
+	defer runner.consumeErrorIfNecessary(&resultedError)
 
-	if err = runner.setHostTenant(); err != nil {
+	if err := runner.setHostTenant(); err != nil {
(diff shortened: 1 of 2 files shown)


0sewa0 (approved):
🪄

waodim (approved):
seems fine to me

PR #801, opened 2022-05-25
Update `controller-gen` to `v0.9.0`

chrismuellner:
# Description
CRD was generated using an outdated version of `controller-gen`.

Removed `+kubebuilder:validation:MaxProperties=1` annotation for `OneAgentSpec` which as incompatible.

## How can this be tested?
Upgrade `controller-gen` and check if the generated CRD is the same:
1. `go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.0`
2. `make manifests`

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ ifeq ($(origin IMG),undefined)
 	endif
 endif
 
-CRD_OPTIONS ?= "crd:preserveUnknownFields=false, crdVersions=v1"
+CRD_OPTIONS ?= "crd:crdVersions=v1"
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -220,7 +220,7 @@ ifeq (, $(shell which controller-gen))
 	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$CONTROLLER_GEN_TMP_DIR ;\
 	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0 ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.0 ;\
```
mjgrzybek, line 223:
```suggestion
	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.0 ;\
```

mjgrzybek (changes requested):

Makefile
mjgrzybek: btw, I don't think Makefile should install tools. I'd leave it just checking if tool is available. 
Now it only installs it, but doesn't check if already existing one is in correct version.
Update is required to be done manually anyway.

chrismuellner:
Those checks are generated when the project is generated with `operator-sdk`.
May be different now than when this Makefile was first generated.

0sewa0:
> I don't think Makefile should install tools.
I agree, but I think the main purpose of it is for the CI

mjgrzybek:
Ok, so now let's just bump version and we can discuss it when refactoring Makefile.

mjgrzybek (approved):

PR #803, opened 2022-05-31
Fixes missing "crd:" prefix for crd options

meik99:
# Description

`controller-gen` produces an error that `crdVersions` is not recognized.
This is due to the prefix `crd:` missing.

## How can this be tested?

`make install` must run successfully

## Checklist


mjgrzybek (changes requested):
I think `build` target should depend on `manifests` to always update manifests when building. Manifests depend on go files anyway.

mjgrzybek (approved):

PR #804, opened 2022-05-31
Move complicated makefile logic into separate scripts

aorcholski:
Installation of kustomize/controller-gen commands and the 'bundle' target moved to shell scripts.

The command.sh script installs a command if the corresponding package is provided. It also returns the full command path if the command is found.

The contents of the 'bundle' target has been moved to bundle.sh script.

## How can this be tested?
Manual test performed - outcome of 'make bundle' command has not changed.

## Checklist


hack/build/bundle.sh
```diff
@@ -0,0 +1,68 @@
...
+--extra-service-accounts dynatrace-kubernetes-monitoring
+--extra-service-accounts dynatrace-activegate
+)
+
+function splitOperatorSdkOptions() {
+    # shellcheck disable=SC2034
+    OPERATOR_SDK_OPTIONS=("$@")
+}
+
+eval splitOperatorSdkOptions "${SERVICE_ACCOUNTS[@]}" "${BUNDLE_CHANNELS}" "${BUNDLE_DEFAULT_CHANNEL}"
```
meik99, line 34: what kind of magic do these bash commands do?

aorcholski: eval .... filters out empty values from the list of parameters.

I agree it's tricky...

0sewa0 (approved):

PR #805, opened 2022-06-01
Add auth token rotation for the activeGate

luhi-DT:
# Description
This change introduces the activeGate auth token rotation support. It adds the functionality that auth tokens are created with an expiration Time (60 days) and also it adds the ability to automatically rotate the tokens if they are older than 30 days.
Additionally I introduced a mechanism to restart the activeGate if the content of the auth token secret (new Auth Token) changes, by adding the auth token content to the activeGate hash annotation.

## How can this be tested?
1. Create an API Token with the correct permission `activeGateTokenManagement.create`
2. Create a Dynakube that also creates an activeGate and add `feature.dynatrace.com/enable-activegate-authtoken: "true"` to it
3. check if the activeGate gets the correct secret mounted and check the logs of the activeGate if it uses the created auth Token
4. Delete the auth-token secret in the operator namespace and check if the activeGate is restarted, when the new auth token is created
5. Wait for 30 days (or modify it when testing it and lower the rotation interval) and check if the auth token gets automatically deleted and recreated again. Additionally the activeGate pod should be restarted and get the new auth Token mounted

## Checklist


src/controllers/activegate/reconciler/statefulset/reconciler.go
```diff
@@ -117,13 +118,13 @@ func (r *Reconciler) buildDesiredStatefulSet() (*appsv1.StatefulSet, error) {
 		return nil, errors.WithStack(err)
 	}
 
-	cpHash, err := r.calculateCustomPropertyHash()
+	agHash, err := r.calculateActiveGateHash()
```
mjgrzybek, line 121:
```suggestion
	agHash, err := r.calculateActiveGateConfigurationHash()
```
`ActiveGateHash` may by confusing because one may think it's image hash.

mjgrzybek (changes requested):

config/deploy/kubernetes/kubernetes-all.yaml
```diff
@@ -3681,6 +3681,7 @@ rules:
       - watch
       - create
       - update
+      - delete
```
meik99, line 3684: If the auth token is outdated, wouldn't it suffice to just update the secret?
Then we wouldn't need this permission

config/helm/chart/default/templates/Common/operator/role-operator.yaml
```diff
@@ -108,6 +108,7 @@ rules:
       - watch
       - create
       - update
+      - delete
```
meik99, line 111: Same as before

src/controllers/activegate/reconciler/statefulset/reconciler.go
meik99:
```suggestion
	activegateHash, err := r.calculateActiveGateHash()
```

src/controllers/activegate/reconciler/statefulset/reconciler.go
```diff
@@ -200,25 +201,49 @@ func (r *Reconciler) deleteStatefulSetIfOldLabelsAreUsed(desiredSts *appsv1.Stat
 	return false, nil
 }
 
-func (r *Reconciler) calculateCustomPropertyHash() (string, error) {
-	customProperties := r.capability.CustomProperties
-	if customProperties == nil || (customProperties.Value == "" && customProperties.ValueFrom == "") {
-		return "", nil
+func (r *Reconciler) calculateActiveGateHash() (string, error) {
+	customPropData, err := r.getCustomPropertyHash()
```
meik99, line 205:
```suggestion
	customPropertyData, err := r.getCustomPropertyHash()
```

src/controllers/activegate/reconciler/statefulset/reconciler.go
```diff
@@ -200,25 +201,49 @@ func (r *Reconciler) deleteStatefulSetIfOldLabelsAreUsed(desiredSts *appsv1.Stat
...
-	if _, err = hash.Write([]byte(data)); err != nil {
+	if _, err := hash.Write([]byte(customPropData + authTokenData)); err != nil {
 		return "", errors.WithStack(err)
 	}
 
 	return strconv.FormatUint(uint64(hash.Sum32()), 10), nil
 }
 
+func (r *Reconciler) getCustomPropertyHash() (string, error) {
+	if !r.needsCustomPropertyHash(r.capability.CustomProperties) {
```
meik99, line 228: customPropertyHash does not need to be a receiver function of r since it only depends on whatever CustomProperties are supplied

src/controllers/activegate/reconciler/statefulset/reconciler.go
```diff
@@ -200,25 +201,49 @@ func (r *Reconciler) deleteStatefulSetIfOldLabelsAreUsed(desiredSts *appsv1.Stat
...
+		return "", nil
+	}
+
+	customPropData, err := r.getDataFromCustomProperty(r.capability.CustomProperties)
```
meik99, line 232: There is probably already a kubeobjects function to get secret data. The called function could be refactored to use it since it wouldn't be that much of a change to be out of scope

src/controllers/activegate/reconciler/statefulset/reconciler.go
mjgrzybek, line 228:
```suggestion
	if isCustomPropertyEmpty(r.capability.CustomProperties) {
```
It doesn't need to be `Reconciler` method.
Hash is still needed, it's just calculated for an empty content.

src/controllers/activegate/reconciler/statefulset/reconciler.go
```diff
@@ -236,3 +261,22 @@ func (r *Reconciler) getDataFromCustomProperty(customProperties *dynatracev1beta
...
+	var secret corev1.Secret
+	err := r.Get(context.TODO(), client.ObjectKey{Name: r.Instance.ActiveGateAuthTokenSecret(), Namespace: namespace}, &secret)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	dataBytes, ok := secret.Data[secrets.ActiveGateAuthTokenName]
+	if !ok {
+		return "", errors.Errorf("no auth-token found on secret '%s' on namespace '%s'", r.Instance.ActiveGateAuthTokenSecret(), namespace)
+	}
```
meik99, line 280: refactor to use kubeobjects.GetSecret. Also, since this is a common use case, an addtional function in kubeobjects can be added to specifically extract the data and then be used here

src/controllers/activegate/reconciler/statefulset/reconciler.go
```diff
@@ -236,3 +261,22 @@ func (r *Reconciler) getDataFromCustomProperty(customProperties *dynatracev1beta
...
+	return string(dataBytes), nil
+}
+
+func (r *Reconciler) needsCustomPropertyHash(customProperties *dynatracev1beta1.DynaKubeValueSource) bool {
```
meik99, line 284: Does not need to be a receiver function

src/controllers/activegate/reconciler/statefulset/statefulset.go
```diff
@@ -25,7 +25,7 @@ const (
 	tenantSecretVolumeName    = "ag-tenant-secret"
 	authTokenSecretVolumeName = "ag-authtoken-secret"
 
-	annotationCustomPropsHash             = dynatracev1beta1.InternalFlagPrefix + "custom-properties-hash"
+	annotationActiveGateHash              = dynatracev1beta1.InternalFlagPrefix + "activegate-hash"
```
meik99: since the annotation changed, be aware that existing statefulsets of customers will be recreated

src/controllers/activegate/reconciler/statefulset/reconciler.go
```diff
@@ -236,3 +265,22 @@ func (r *Reconciler) getDataFromCustomProperty(customProperties *dynatracev1beta
 	}
 	return customProperties.Value, nil
 }
+
+func (r *Reconciler) getDataFromAuthTokenSecret() (string, error) {
```
mjgrzybek, line 269: Nit: make it generic, move to `kubeobjects`, like:
`getDataFromSecret(secretName string, namespace string) (string, error)`

src/controllers/dynakube/activegate/secrets/authtoken_reconciler.go
```diff
@@ -15,52 +16,66 @@ import (
...
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			log.Info("creating activeGateAuthToken secret")
+			return r.ensureAuthTokenSecret()
+		}
+		return nil, errors.WithStack(err)
+	}
+	if isSecretOutdated(&config) {
+		log.Info("activeGateAuthToken is outdated, creating new one")
+		if err := r.deleteSecret(&config); err != nil {
```
meik99, line 64: Instead of deleting the secret, it could be updated instead to save the delete permission

src/controllers/activegate/reconciler/statefulset/statefulset.go
```diff
@@ -51,7 +51,7 @@ const (
 type statefulSetProperties struct {
 	*dynatracev1beta1.DynaKube
 	*dynatracev1beta1.CapabilityProperties
-	customPropertiesHash    string
+	activeGateHash          string
```
mjgrzybek, line 54:
```suggestion
	activeGateConfigurationHash          string
```

src/dtclient/activegate_auth_token.go
```diff
@@ -50,6 +53,7 @@ func (dtc *dynatraceClient) createAuthTokenRequest(dynakubeName string) (*http.R
 		Name:           dynakubeName,
 		SeedToken:      false,
 		ActiveGateType: activeGateType,
+		ExpirationDate: getAuthTokenExpirationDate(),
```
0sewa0, line 56: I don't see this field being used currently

I remember us talking about if we should use the creation timestamp of the secret or store the expiration date in the secret. I don't 100% remember which one we agreed on to be the better option. (I think it was the secret creation timestamp as it was easier/simpler)

However the others have a point about updating the secret would be better than just deleting it, and then you would need to use this field to determine if the contents are outdated.

luhi-DT: Yes, I can also remember that discussion and we came to the conclusion, that we should just delete the secret and create a new one, because managing and storing the creationTimestamp ourselves is basically the same as letting kubernetes do it. But either way it makes sense to maintain the secret and to update it instead of recreating. However, the property is currently used in the POST request to the tenant, when creating the token, so that the newly created token is only valid for 60 days

src/controllers/dynakube/activegate/secrets/authtoken_reconciler.go
luhi-DT: yes, I had a discussion with @0sewa0 about that, the downside of updating the secret is that we need to save the timeStamp, when the authToken got updated in order to track, when we need to update the token again

meik99: makes sense

src/controllers/activegate/reconciler/statefulset/reconciler.go
luhi-DT: done

src/controllers/activegate/reconciler/statefulset/reconciler.go
luhi-DT: implemented a new function which, takes a secret name and namespace as a parameter and returns the data from the secret, thanks for the input :)

src/controllers/activegate/reconciler/statefulset/statefulset.go
luhi-DT: done

0sewa0 (approved):

mjgrzybek (dismissed):

src/controllers/activegate/reconciler/statefulset/reconciler.go
```diff
@@ -200,39 +202,64 @@ func (r *Reconciler) deleteStatefulSetIfOldLabelsAreUsed(desiredSts *appsv1.Stat
...
+	}
+
+	customPropertyData, err := r.getDataFromCustomProperty(r.capability.CustomProperties)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+	return customPropertyData, nil
+}
+
+func (r *Reconciler) getAuthTokenValueToHash() (string, error) {
```
meik99, line 240:
```suggestion
func (r *Reconciler) getAuthTokenValueAsHash() (string, error) {
```

src/controllers/activegate/reconciler/statefulset/reconciler.go
```diff
@@ -200,39 +202,64 @@ func (r *Reconciler) deleteStatefulSetIfOldLabelsAreUsed(desiredSts *appsv1.Stat
...
 	hash := fnv.New32()
-	if _, err = hash.Write([]byte(data)); err != nil {
+	if _, err := hash.Write([]byte(customPropertyData + authTokenData)); err != nil {
 		return "", errors.WithStack(err)
 	}
 
 	return strconv.FormatUint(uint64(hash.Sum32()), 10), nil
 }
 
+func (r *Reconciler) getCustomPropertyValueToHash() (string, error) {
```
meik99, line 228:
```suggestion
func (r *Reconciler) getCustomPropertyValueAsHash() (string, error) {
```

meik99 (changes requested):

0sewa0 (dismissed):

0sewa0 (approved):

PR #806, opened 2022-06-01
Store initial connect retry value in curl config file

meik99:
# Description

The value from the `feature.dynatrace.com/oneagent-initial-connect-retry-ms` feature flag should be passed to a file instead of an environment variable.
This file is now generated by the standalone runner and mounted by the webhook.

## How can this be tested?

* Create a test-namespace with test apps
* Deploy the operator 
* Apply a Dynakube with the mentioned flag set
* Restart the test-apps as soon as the operator deployment is ready
* The test app pods should have the `curl_options.conf` file mounted to `/var/lib/dynatrace/oneagent/agent/customkeys/curl_options.conf`
* Exec into one of the pods
* The log at `/opt/dynatrace/oneagent-paas/log/nginx/ruxitagent_nginx_myapp-__startup_1.0.log` should contain the line `Applied configuration from file: /var/lib/dynatrace/oneagent/agent/customkeys/curl_options.conf`

## Checklist


src/kubeobjects/query/dynakube.go
```diff
@@ -0,0 +1,83 @@
...
+	proxyKey     = "proxy"
+	tlsCertKey   = "server.crt"
+)
+
+type DynakubeQuery struct {
+	clt       client.Client
+	namespace string
+}
+
+func NewDynakubeQuery(clt client.Client, namespace string) *DynakubeQuery {
```
mjgrzybek, line 23: Queries shouldn't mutate object, so value should be returned instead of a pointer and used as method receivers.

src/kubeobjects/query/dynakube.go
```diff
@@ -0,0 +1,83 @@
...
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	trustedCAKey = "certs"
+	proxyKey     = "proxy"
+	tlsCertKey   = "server.crt"
```
0sewa0, line 15: could be in `dynatracev1beta1` so it can be reused in other places (test and such)

src/kubeobjects/query/dynakube.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func NewDynakubeQuery(clt client.Client, namespace string) *DynakubeQuery {
+	return &DynakubeQuery{
+		clt:       clt,
+		namespace: namespace,
+	}
+}
+
+func (query *DynakubeQuery) Proxy(dynakube *dynatracev1beta1.DynaKube) (string, error) {
```
0sewa0, line 30: nitpick: pointer receiver is not necessary

0sewa0: nitpick2: why not pass in the context ? 

I know that our current use of context is "not optimal" but it would make it future proof, and more easily reusable in other parts of the code

src/initgeneration/initgeneration.go
```diff
@@ -26,27 +26,29 @@ type InitGenerator struct {
...
+		client:        client,
+		apiReader:     apiReader,
+		namespace:     namespace,
+		dynakubeQuery: query.NewDynakubeQuery(client, namespace),
 	}
 }
 
 // GenerateForNamespace creates the init secret for namespace while only having the name of the corresponding dynakube
 // Used by the podInjection webhook in case the namespace lacks the init secret.
-func (g *InitGenerator) GenerateForNamespace(ctx context.Context, dk dynatracev1beta1.DynaKube, targetNs string) (bool, error) {
```
0sewa0, line 46: why remove `ctx`??

I see its not used, but I also see that it could be (here: https://github.com/Dynatrace/dynatrace-operator/pull/806/files#diff-23aafca6480c6cd32914b1b7b9691cbf3acfe3d7ec384c2926b56ebd698c347aR133)

src/initgeneration/initgeneration_test.go
```diff
@@ -6,17 +6,22 @@ import (
...
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
-var (
+const (
+	trustedCAKey = "certs"
+	tlsCertKey   = "server.crt"
```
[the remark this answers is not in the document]
meik99: moved the consts

src/kubeobjects/query/dynakube.go
meik99: moved the consts

src/kubeobjects/query/dynakube.go
meik99: Added option to add context

src/initgeneration/initgeneration.go
meik99: refactored code to hell and back

0sewa0 (dismissed):
I really like the refactors

meik99:
I'll wait for @mjgrzybek changes to be merged first before merging

src/kubeobjects/query.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type kubeQuery struct {
+	kubeClient client.Client
+	kubeReader client.Reader
```
mjgrzybek, line 12: I didn't find any usage of this member other than passing around.

mjgrzybek (changes requested):

src/kubeobjects/query.go
0sewa0: Its a leftover, the old `CreateOrUpdateSecret...` used to read the secret, for that it used the `kubeReader`, but the new logic tries to create it and if fails then it will just update it (which I think is simpler and therefore nicer)

but this `kubeQuery` should be reused for other `kubeobject`stuff (daemonset, deployment) and there they might need it

mjgrzybek: Agree but it's not tested neither.
To avoid pushing dead code, adding single usage of it would be sufficient.

meik99:
```golang
type SecretQuery struct {
	kubeQuery
}

func NewSecretQuery(ctx context.Context, kubeClient client.Client, kubeReader client.Reader, log logr.Logger) SecretQuery {
	return SecretQuery{
		newKubeQuery(ctx, kubeClient, kubeReader, log),
	}
}
```

meik99: I found something that could be quickly refactored, give me a minute

meik99: done, kubeReader is now used

src/kubeobjects/secret.go
```diff
@@ -25,6 +26,16 @@ func NewSecretQuery(ctx context.Context, kubeClient client.Client, kubeReader cl
 	}
 }
 
+func (query SecretQuery) Get(objectKey client.ObjectKey) (corev1.Secret, error) {
+	ctx := query.ctx
+	kubeReader := query.kubeReader
```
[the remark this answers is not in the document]
meik99: I just do that because otherwise the lines sometimes get long because of the call chains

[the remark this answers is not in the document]
mjgrzybek: Thus single letter receiver is a pattern.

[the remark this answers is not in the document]
meik99: I have been convinced

mjgrzybek (approved):

PR #809, opened 2022-06-02
Fix data-ingests's `endpoint.properties` secret handling

mjgrzybek:
# Description
1. Secret `dynatrace-data-ingest-endpoint` is now always generated nevertheless which ingestions are enabled.
If no ingestion is enabled, secret's data contains an empty string, as expected.

2. Fixed bug causing permanent reconciliation which was caused by comparing `nil` and an empty map, assuming it's equal.

# How can this be tested?
Check if `dynatrace-data-ingest-endpoint` secret in `deafult` namespace contains expected content.

Dynakubes to be used:
1. default one, with no annotation
2. with AG `statsd-ingest` capability enabled

Maniulate dynakubes with:
1. `k -ndynatrace annotate dynakubes.dynatrace.com dk --overwrite feature.dynatrace.com/disable-metadata-enrichment=false` - no change should be observed
2. `k -ndynatrace annotate dynakubes.dynatrace.com dk --overwrite feature.dynatrace.com/disable-metadata-enrichment=true` - metrics ingest URL and token should be removed from the secret data

## Checklist

### DIFF
--- a/src/ingestendpoint/secret.go
+++ b/src/ingestendpoint/secret.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"strings"
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	agcapability "github.com/Dynatrace/dynatrace-operator/src/controllers/activegate/capability"
@@ -19,10 +20,10 @@ import (
 )
 
 const (
-	UrlSecretField   = "DT_METRICS_INGEST_URL"
-	TokenSecretField = "DT_METRICS_INGEST_API_TOKEN"
-	StatsdIngestUrl  = "DT_STATSD_INGEST_URL"
-	configFile       = "endpoint.properties"
+	MetricsUrlSecretField   = "DT_METRICS_INGEST_URL"
+	MetricsTokenSecretField = "DT_METRICS_INGEST_API_TOKEN"
+	StatsdUrlSecretField    = "DT_STATSD_INGEST_URL"
+	configFile              = "endpoint.properties"
 )
 
 // EndpointSecretGenerator manages the mint endpoint secret generation for the user namespaces.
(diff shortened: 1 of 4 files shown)


src/ingestendpoint/secret_test.go
```diff
@@ -11,129 +11,131 @@ import (
...
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	testPaasToken              = "test-paas-token"
-	testAPIToken               = "test-api-token"
-	testDataIngestToken        = "test-data-ingest-token"
-	testUpdatedDataIngestToken = "updated-test-data-ingest-token"
+	paasToken              = "test-paas-token"
```
chrismuellner, line 19: why remove the `test` prefix?

mjgrzybek: Why keep it? This is test file by definition, no need to keep boilerplate in consts naming.

0sewa0: they can bleed into the actual `secret.go`, which is not cool

chrismuellner: can lead to confusing during testing. it's just easier to keep the prefix, also makes it clear it's a value only used for tests and not defined outside those.

0sewa0: I didn't know that 😮 
(I still like the prefix though 😅 )

mjgrzybek: Democracy won ;).

src/ingestendpoint/secret.go
```diff
@@ -159,21 +162,23 @@ func (g *EndpointSecretGenerator) PrepareFields(ctx context.Context, dk *dynatra
 		return nil, errors.WithMessage(err, "failed to query tokens")
 	}
 
-	if token, ok := tokens.Data[dtclient.DynatraceDataIngestToken]; ok {
-		fields[TokenSecretField] = string(token)
-	}
+	if !dk.FeatureDisableMetadataEnrichment() {
```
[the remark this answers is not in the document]
mjgrzybek: I prefer `dk` because it's easier to guess if method is called on object or is it function from `dynakube` package.

src/ingestendpoint/secret.go
```diff
@@ -159,21 +162,23 @@ func (g *EndpointSecretGenerator) PrepareFields(ctx context.Context, dk *dynatra
...
+		if token, ok := tokens.Data[dtclient.DynatraceDataIngestToken]; ok {
+			fields[MetricsTokenSecretField] = string(token)
+		}
 
-	if diUrl, err := dataIngestUrl(dk); err != nil {
-		return nil, err
-	} else {
-		fields[UrlSecretField] = diUrl
+		if diUrl, err := dataIngestUrl(dk); err != nil {
```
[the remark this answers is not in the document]
mjgrzybek: Done, but replaced suggestion of `parseDataIngestUrl` to `dataIngestUrlFor(dynakube)` because it's not parsing but rather calculating URL based on dynakube content.

src/ingestendpoint/secret.go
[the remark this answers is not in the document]
meik99: fair

meik99 (approved):

src/ingestendpoint/secret.go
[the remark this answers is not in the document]
meik99: fair

PR #810, opened 2022-06-02
Updates the questions.yml for rancher

0sewa0:
# Description
Updates the `questions.yml` to use all the relevant fields from the `values.yaml`

## How can this be tested?
TODO

## Checklist


config/helm/chart/default/questions.yml
```diff
@@ -3,29 +3,164 @@ categories:
...
+    description: "Installs the Custom Resource Definitions for the Dynakube. This is recommended if you haven't installed it manually yet. Default: true"
+    default: true
+    type: boolean
+    group: "Global Configuration"
+
   - variable: operator.image
-    label: "Set a custom image for operator"
-    description: "Optional: Set a custom image for operator. Defaults to docker.io/dynatrace/dynatrace-operator"
+    label: "Set a custom image for operator components"
+    description: "Set a custom image for operator. Defaults to docker.io/dynatrace/dynatrace-operator"
```
chrismuellner, line 16: defaults to tenant registry

chrismuellner, line 16: why remove the optional?

config/helm/chart/default/values.yaml
```diff
@@ -52,6 +52,4 @@ csidriver:
 
 securityContextConstraints:
   enabled: true # Only applicable for Openshift
-  
-activeGate:
```
chrismuellner, line 56: is this not used anymore?

chrismuellner (changes requested):

config/helm/chart/default/values.yaml
0sewa0: yep, it used to set a feature flag on the CR

config/helm/chart/default/questions.yml
0sewa0: everything is optional, felt redundant 
should I add `Optional` to every field ?

chrismuellner: you're right, not needed

0sewa0: I could update the `app-readme.md` so it say that the configuration fields are optional
or have the section names have the word `Optional`

0sewa0: there is an operator image in the tenant registry ?

chrismuellner: whoops

chrismuellner (approved):

PR #819, opened 2022-06-06
Helm development version is set to appropriate version.

aorcholski:
Set helm development version to:
- 0.0.0 if master is the current branch
- < ver > if release-< ver > is the current branch
- otherwise Chart.yaml version/appVersion fields are not changed

## How can this be tested?
Execute `make bundle` command on the following branches:
- master
- release-9.99.999
- feature/set-helm-version

Chart.yaml version/appVersion fields should be updated according to the requirements.

## Checklist

### DIFF
--- a/hack/make/manifests/crd.mk
+++ b/hack/make/manifests/crd.mk
@@ -1,14 +1,3 @@
--include config.mk
--include ../prerequisites.mk
-
-helm/version:
-ifneq ($(CHART_VERSION),)
-	sed "s/^version: .*/version: $(CHART_VERSION)/" $(HELM_CHART_DEFAULT_DIR)/Chart.yaml >  $(HELM_CHART_DEFAULT_DIR)/Chart.yaml.output
-	mv $(HELM_CHART_DEFAULT_DIR)/Chart.yaml.output $(HELM_CHART_DEFAULT_DIR)/Chart.yaml
-	sed "s/^appVersion: .*/appVersion: $(CHART_VERSION)/" $(HELM_CHART_DEFAULT_DIR)/Chart.yaml >  $(HELM_CHART_DEFAULT_DIR)/Chart.yaml.output
-	mv $(HELM_CHART_DEFAULT_DIR)/Chart.yaml.output $(HELM_CHART_DEFAULT_DIR)/Chart.yaml
-endif
-
 ## Generates a CRD in config/crd/bases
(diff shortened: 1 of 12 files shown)


hack/make/manifests/crd.mk
```diff
@@ -1,6 +1,12 @@
 -include config.mk
 -include ../prerequisites.mk
 
+manifests/version:
+ifneq ($(CHART_VERSION),)
+	sed -i "s/^version: .*/version: $(CHART_VERSION)/" $(HELM_CHART_DEFAULT_DIR)/Chart.yaml
+	sed -i "s/^appVersion: .*/appVersion: $(CHART_VERSION)/" $(HELM_CHART_DEFAULT_DIR)/Chart.yaml
```
0sewa0, line 7: `sed -i` doesn't work on Mac os 😬

0sewa0 (changes requested):

0sewa0:
tested it and got the following
on master:              `0.0.0` 👍 
on release-0.7:       `0.7` ❓ 
on feature branch:  nothing happened 👍 

shouldn't it be `0.7.0`  on release-0.7 ?

also something we should tackle after this PR: if we have a 0.7.1 release we will not have a separate branch for that, but the Chart.yaml should be updated to `0.7.1`, how should we go about that ?
@chrismuellner

hack/make/manifests/crd.mk
```diff
@@ -1,6 +1,14 @@
 -include config.mk
 -include ../prerequisites.mk
 
+manifests/version:
```
meik99, line 4: That target changes helm charts, it doesn't directly influence the manifests and should therefore have a different target

meik99 (changes requested):

hack/make/manifests/crd.mk
```diff
@@ -1,6 +1,14 @@
 -include config.mk
 -include ../prerequisites.mk
 
+helm/version:
```
meik99, line 4:
```suggestion
## Sets the Helm Charts version and appVersion
helm/version:
```

meik99: Also I meant putting this target in a different file or folder since it has nothing to do with the manifests, not just renaming it

meik99 (dismissed):

0sewa0 (approved):

PR #823, opened 2022-06-07
Rename `.Values.operator.image` to `.Values.image`

0sewa0:
# Description
Setting the `image` for the deployments when using helm now is less weird, as the while you already could set the `image` for the deployments, you had to do it with `operator.image` which is a bit confusing when it also updates the image used by the CSI and webhook.

## How can this be tested?
Set the `image` field when deploying(or generating yamls) via helm, then check that it was used in the deployments/daemonset.

Also try what is described in https://github.com/Dynatrace/dynatrace-operator/blob/master/config/helm/README.md.

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/_helpers.tpl
+++ b/config/helm/chart/default/templates/_helpers.tpl
@@ -113,8 +113,6 @@ Check if default image is used
 {{- define "dynatrace-operator.image" -}}
 {{- if .Values.image -}}
 	{{- printf "%s" .Values.image -}}
-{{- else if .Values.operator.image -}} # Left in for backwards compativility
-	{{- printf "%s" .Values.operator.image -}}
 {{- else -}}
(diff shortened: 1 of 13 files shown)


config/helm/chart/default/templates/_helpers.tpl
```diff
@@ -111,7 +111,9 @@ app.kubernetes.io/component: oneagent
 Check if default image is used
 */}}
 {{- define "dynatrace-operator.image" -}}
-{{- if .Values.operator.image -}}
+{{- if .Values.image -}}
+	{{- printf "%s" .Values.image -}}
+{{- else if .Values.operator.image -}} # Left in for backwards compativility
```
chrismuellner, line 116: not needed

0sewa0: 👍

meik99 (approved):

PR #824, opened 2022-06-08
Pushes release image to all registries

meik99:
# Description

Extends the GitHub Actions to include pushing the image of a release to RHCC, GCR and DockerHub as well as creating multiarch manifests for them.

## How can this be tested?

* Create a release tag in the form v[0-9]+.[0-9]+.[0-9]+
* Check if the actions complete successfully

## Checklist


.github/workflows/ci.yaml
```diff
@@ -168,6 +168,161 @@ jobs:
       docker_repo_username: ${{ secrets.QUAY_USERNAME }}
       docker_repo_password: ${{ secrets.QUAY_PASSWORD }}
 
+  push-rhcc:
```
chrismuellner, line 171: split these jobs into a separate workflow that is only triggered [on push of new tag](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs)

chrismuellner: since the logic for rhcc, gcr and dockerhub is mostly the same did you try running this using a [job matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-expanding-configurations)?
could pass in registry, registry username, registry password and reuse the same steps

chrismuellner (changes requested):

.github/workflows/ci.yaml
meik99: done

.github/workflows/publish-images.yaml
```diff
@@ -0,0 +1,83 @@
...
+        uses: ./.github/workflows/upload-dockerimage.yaml
+        with:
+          platform: ${{matrix.platform}}
+          labels: ${{ needs.prepare.outputs.labels }}
+          version: ${{ needs.prepare.outputs.version }}
+          registry: ${{matrix.url}}
+          repository: ${{matrix.repository}}
+          secrets.docker_repo_username: ${{matrix.username}}
+          secrets.docker_repo_password: ${{matrix.password}}
+      - name: Create manifest for on ${{matrix.registry}}
```
chrismuellner, line 79: typo

.github/workflows/publish-images.yaml
```diff
@@ -0,0 +1,83 @@
...
+        env:
+          IMAGE: ${{matrix.url}}/${{matrix.repository}}
+        run: |
+          hack/build/ci/create-manifest.sh "${IMAGE}" "${{ needs.prepare.outputs.version }}" true
```
chrismuellner, line 83: would this mean that manifest creation would fail for the first image push since it tries to create a manifest containing a non-existent image (other architecture)?

meik99:
Converting to a draft because I just noticed I cannot reuse any of the existing workflows as the tag for GCR is different

.github/workflows/publish-images.yaml
```diff
@@ -0,0 +1,118 @@
...
+          registry: ${{ matrix.url }}
+          username: ${{ matrix.username }}
+          password: ${{ matrix.password }}
+      - name: Create manifest for ${{matrix.repository}}
+        env:
+          IMAGE: ${{matrix.url}}/${{matrix.repository}}
+        # Simply pulls, otherwise the artifact would need to be re-downloaded and re-tagged correctly which would mean
+        # two more steps instead of two more lines
+        run: |          
+          docker pull ${{matrix.url}}/${{matrix.repository}}:${{ needs.prepare.outputs.version }}-amd64
```
chrismuellner, line 116:
```suggestion
          docker pull ${{IMAGE}}:${{ needs.prepare.outputs.version }}-amd64
```

chrismuellner: why the pull? does the image need to be in the local registry?

.github/workflows/publish-images.yaml
```diff
@@ -0,0 +1,118 @@
...
+          goargs: ${{ needs.prepare.outputs.goargs }}
+          labels: ${{ needs.prepare.outputs.labels }}
+          image-tag: ${{ needs.prepare.outputs.version }}
+  push:
+    name: Push images
+    environment: Release
+    needs: [prepare,build]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
chrismuellner, line 49: I'd love to not duplicate the matrix...
since we always want to do a multi arch build here can we simplify some logic by moving it into the `upload-dockerimage` action?

chrismuellner (changes requested):

.github/workflows/publish-images.yaml
meik99: can manifests be created without having the images locally?

chrismuellner (approved):

PR #825, opened 2022-06-08
Add `UsesImage` column to Dynakube metadata table

chrismuellner:
# Description
Add new `UsesImage` column to the Dynakube table to be used by CSI server and provisioner. Migration of existing databases to updated schema with new column should happen automatically.

## How can this be tested?
No change in functionality when using the CSI driver:
* existing data is migrated to new schema without data loss
* new value is currently not used, so no changes expected

## Checklist

### DIFF
--- a/src/controllers/csi/metadata/sqlite.go
+++ b/src/controllers/csi/metadata/sqlite.go
@@ -44,13 +44,13 @@ const (
 	// ALTER
 	dynakubesAlterStatementUsesImageColumn = `
 	ALTER TABLE dynakubes
-	ADD COLUMN uses_image BOOLEAN NOT NULL DEFAULT false;
+	ADD COLUMN UsesImage BOOLEAN NOT NULL DEFAULT false;
 	`
 
 	// INSERT
 	insertDynakubeStatement = `
-	INSERT INTO dynakubes (Name, TenantUUID, LatestVersion)
-	VALUES (?,?,?);
+	INSERT INTO dynakubes (Name, TenantUUID, LatestVersion, UsesImage)
+	VALUES (?,?,?,?);
 	`
 
 	insertVolumeStatement = `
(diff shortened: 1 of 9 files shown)


0sewa0 (approved):

src/controllers/csi/metadata/sqlite.go
```diff
@@ -177,6 +184,19 @@ func (a *SqliteAccess) createTables() error {
...
 
+// setupDynakubeTable creates the dynakubes table if it doesn't exist and tries to add additional columns
+func (a *SqliteAccess) setupDynakubeTable() error {
+	if _, err := a.conn.Exec(dynakubesCreateStatement); err != nil {
+		return fmt.Errorf("couldn't create the table %s, err: %s", dynakubesTableName, err)
+	}
+
+	if _, err := a.conn.Exec(dynakubesAlterStatementUsesImageColumn); err != nil {
+		// statement errors if column already exists, swallow error
+		log.Info("column UsesImage already exists")
```
meik99, line 195: what if there is an error for a different reason?

meik99 (changes requested):

src/controllers/csi/metadata/sqlite.go
chrismuellner: seems I missed that the [error code](http://www.sqlite.org/c3ref/c_abort.html) and [extended error code](http://www.sqlite.org/c3ref/c_abort_rollback.html) actually indicate a `SQLITE_ERROR` with `SQLITE_ERROR_MISSING_COLLSEQ`:
image

[error_missing_collseq](https://www.sqlite.org/rescode.html#error_missing_collseq) seems to be an issue specific to alter/create, so much safer to only skip in that case.

src/controllers/csi/metadata/sqlite.go
```diff
@@ -177,6 +184,22 @@ func (a *SqliteAccess) createTables() error {
...
+	if _, err := a.conn.Exec(dynakubesCreateStatement); err != nil {
+		return fmt.Errorf("couldn't create the table %s, err: %s", dynakubesTableName, err)
+	}
+
+	if _, err := a.conn.Exec(dynakubesAlterStatementUsesImageColumn); err != nil {
+		sqliteError := err.(sqlite3.Error)
+		if sqliteError.Code == sqlite3.ErrError {
+			// generic sql error, column already exists
+			log.Info("column UsesImage already exists")
+		}
```
0sewa0, line 198: shouldn't you return the error if its not the "ok" one ?

0sewa0 (approved):

PR #827, opened 2022-06-09
Added feature flag to run oneagent as privileged

waodim:
# Description

Added new feature flag which allows the oneagent container to run as privileged

## How can this be tested?
Apply this feature flag and check whether the oneagent container is ran in a privileged context.

## Checklist

### DIFF
--- a/src/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/src/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -198,7 +198,7 @@ func (dsInfo *builderInfo) podSpec() corev1.PodSpec {
 				TimeoutSeconds:      1,
 			},
 			Resources:       resources,
-			SecurityContext: dsInfo.unprivilegedSecurityContext(),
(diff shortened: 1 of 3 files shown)


src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -284,5 +284,9 @@ func (dsInfo *builderInfo) unprivilegedSecurityContext() *corev1.SecurityContext
 		securityContext.RunAsUser = &unprivilegedUser
 		securityContext.RunAsGroup = &unprivilegedGroup
 	}
+	runAsPrivileged := true
+	if dsInfo.instance.OneAgentRunsPrivileged() {
+		securityContext.Privileged = &runAsPrivileged
```
chrismuellner, line 289: not sure how this works with running as unprivileged user/group. maybe only providing a privileged securitycontext in this case and skipping the rest would be best

chrismuellner (changes requested):

chrismuellner (approved):

PR #834, opened 2022-06-14
Reworks imageInstaller to extract into a shared dir

0sewa0:
# Description

The current way we download/cache and unpack  `codeModulesImage` is not perfect when it comes to node storage usage, as it extracts the image for every tenant.

This PR changes how the provisioner prepares the node for the driver.
The image is downloaded to `/cache/<image-ingest>` and gets unpacked to `data/codemodules/<image-digest>`.
We only download/unpack if the digest changes (or its a new digest), we store the digest in the `dynakubes` table in the metadata database on the node.
After we unpack to `data/codemodules/<image-digest>` then `/cache/<image-ingest>` gets cleaned up as its no longer needed for anything.

Because of the shared nature of the agent "binary" we can't update the `ruxitagentproc.conf` at 
`data/codemodules/<image-digest>`, so we instead create a `data/<tenant>/config/...` where we update/merge the necessary config files.

The driver WILL use `data/codemodules/<image-digest>` + `data/<tenant>/config/...` as 2 of its lowerdirs during mounting the overlayfs volume for user pods.
Therefore `data/<tenant>/config/...` needs to follow the same folder structure as `data/codemodules/<image-digest>`.

## How can this be tested?
Deploy anything that uses the csidriver + `codeModulesImage`, 
then check make sure the csidriver works by: 
- seeing no error in the logs for the provisioner (the driver will not work which is expected)
- check the filesystem for:
  - `data/codemodules/<image-digest>` is a valid oneagent "install"
  - `data/codemodules/<image-digest>/agent/bin/current` is a valid symlink
  - `data/<tenant>/config/agent/conf/ruxitagentproc.conf` is a valid "tenantified" conf file
  - `/cache/<image-ingest>`is cleaned up 

## Checklist

### DIFF
--- a/src/controllers/csi/provisioner/agent.go
+++ b/src/controllers/csi/provisioner/agent.go
@@ -44,6 +44,7 @@ func newAgentUpdater(
 	var err error
 	if dk.CodeModulesImage() != "" {
 		agentInstaller, err = setupImageInstaller(ctx, fs, apiReader, certPath, dk)
+		targetDir = path.AgentSharedBinaryDirForVersion(targetVersion)
 		if err != nil {
 			return nil, err
 		}
--- a/src/controllers/csi/provisioner/agent.go
+++ b/src/controllers/csi/provisioner/agent.go
@@ -13,7 +13,6 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/src/installer/image"
 	"github.com/Dynatrace/dynatrace-operator/src/installer/url"
 	"github.com/spf13/afero"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -95,10 +94,10 @@ func setupImageInstaller(ctx context.Context, fs afero.Fs, apiReader client.Read
 	return imageInstaller, nil
 }
 
-func (updater *agentUpdater) updateAgent(installedVersion, tenantUUID string, latestProcessModuleConfigCache *processModuleConfigCache) (string, error) {
+func (updater *agentUpdater) updateAgent(installedVersion string, latestProcessModuleConfigCache *processModuleConfigCache) (string, error) {
 	defer updater.cleanCertsIfPresent()
 	var updatedVersion string
-	if updater.VersionDirNotPresent() || installedVersion == "" {
+	if updater.versionDirNotPresent() || installedVersion == "" {
 		log.Info("updating agent",
 			"target version", updater.targetVersion,
 			"installed version", installedVersion,
@@ -106,10 +105,10 @@ func (updater *agentUpdater) updateAgent(installedVersion, tenantUUID string, la
 		_ = updater.fs.MkdirAll(updater.targetDir, 0755)
 
 		if err := updater.installer.InstallAgent(updater.targetDir); err != nil {
-			updater.recorder.sendFailedInstallAgentVersionEvent(updater.targetVersion, tenantUUID)
+			updater.recorder.sendFailedInstallAgentVersionEvent(updater.targetVersion, updater.tenantUUID)
 			return "", err
 		}
-		updater.recorder.sendInstalledAgentVersionEvent(updater.targetVersion, tenantUUID)
+		updater.recorder.sendInstalledAgentVersionEvent(updater.targetVersion, updater.tenantUUID)
 		updatedVersion = updater.targetVersion
 	}
 	log.Info("updating ruxitagentproc.conf on latest installed version")
@@ -128,26 +127,7 @@ func (updater *agentUpdater) cleanCertsIfPresent() {
 	}
 }
 
-func (updater agentUpdater) VersionDirNotPresent() bool {
+func (updater agentUpdater) versionDirNotPresent() bool {
 	_, err := updater.fs.Stat(updater.targetDir)
 	return os.IsNotExist(err)
-}
-
-type updaterEventRecorder struct {
-	dynakube *dynatracev1beta1.DynaKube
-	recorder record.EventRecorder
-}
-
-func (event *updaterEventRecorder) sendFailedInstallAgentVersionEvent(version, tenantUUID string) {
-	event.recorder.Eventf(event.dynakube,
-		corev1.EventTypeWarning,
-		failedInstallAgentVersionEvent,
-		"Failed to install agent version: %s to tenant: %s", version, tenantUUID)
-}
-
-func (event *updaterEventRecorder) sendInstalledAgentVersionEvent(version, tenantUUID string) {
-	event.recorder.Eventf(event.dynakube,
-		corev1.EventTypeNormal,
-		installAgentVersionEvent,
-		"Installed agent version: %s to tenant: %s", version, tenantUUID)
 }
--- a/src/controllers/csi/provisioner/agent.go
+++ b/src/controllers/csi/provisioner/agent.go
@@ -99,7 +99,7 @@ func getUrlProperties(version string) *url.Properties {
 	}
 }
 
-func setupImageInstaller(ctx context.Context, fs afero.Fs, path metadata.PathResolver, apiReader client.Reader, certPath, digest string, dynakube *dynatracev1beta1.DynaKube) (installer.Installer, error) {
+func setupImageInstaller(ctx context.Context, fs afero.Fs, pathResolver metadata.PathResolver, apiReader client.Reader, certPath, digest string, dynakube *dynatracev1beta1.DynaKube) (installer.Installer, error) {
 	dockerConfig, err := dockerconfig.NewDockerConfig(ctx, apiReader, *dynakube)
 	if err != nil {
 		return nil, err
@@ -113,7 +113,7 @@ func setupImageInstaller(ctx context.Context, fs afero.Fs, path metadata.PathRes
 	imageInstaller := image.NewImageInstaller(fs, &image.Properties{
 		ImageUri:     dynakube.CodeModulesImage(),
 		ImageDigest:  digest,
-		PathResolver: path,
+		PathResolver: pathResolver,
 		DockerConfig: *dockerConfig})
 	return imageInstaller, nil
 }
--- a/src/controllers/csi/provisioner/agent.go
+++ b/src/controllers/csi/provisioner/agent.go
@@ -27,30 +27,18 @@ type agentUpdater struct {
 	recorder      updaterEventRecorder
 }
 
-func newAgentUpdater(
+func newAgentUrlUpdater(
 	ctx context.Context,
 	fs afero.Fs,
-	apiReader client.Reader,
 	dtc dtclient.Client,
 	path metadata.PathResolver,
 	recorder record.EventRecorder,
 	dk *dynatracev1beta1.DynaKube) (*agentUpdater, error) {
+
 	tenantUUID := dk.ConnectionInfo().TenantUUID
 	targetVersion := dk.CodeModulesVersion()
-	targetDir := path.AgentBinaryDirForVersion(tenantUUID, targetVersion)
-	certPath := path.ImageCertPath(tenantUUID)
 
-	var agentInstaller installer.Installer
-	var err error
-	if dk.CodeModulesImage() != "" {
-		agentInstaller, err = setupImageInstaller(ctx, fs, apiReader, certPath, dk)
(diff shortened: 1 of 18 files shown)


src/controllers/csi/provisioner/controller.go
```diff
@@ -156,28 +157,44 @@ func (provisioner *OneAgentProvisioner) updateAgentInstallation(ctx context.Cont
 	}
 	latestProcessModuleConfig = latestProcessModuleConfig.AddHostGroup(dk.HostGroup())
 
+	var agentUpdater *agentUpdater
 	if dk.CodeModulesImage() != "" {
+		// TODO: Maybe use the one in the Dynakube ?
```
mjgrzybek, line 162: I'd prefer connection info from Dynakube because if it changes it meantime and is not reonciled yet, it makes us use two states (currently stored in dynakube and to-be-updated-with) at the time.

src/controllers/csi/provisioner/agent.go
```diff
@@ -78,7 +99,7 @@ func getUrlProperties(version string) *url.Properties {
 	}
 }
 
-func setupImageInstaller(ctx context.Context, fs afero.Fs, apiReader client.Reader, certPath string, dynakube *dynatracev1beta1.DynaKube) (installer.Installer, error) {
+func setupImageInstaller(ctx context.Context, fs afero.Fs, path metadata.PathResolver, apiReader client.Reader, certPath, digest string, dynakube *dynatracev1beta1.DynaKube) (installer.Installer, error) {
```
mjgrzybek, line 102:
```suggestion
func setupImageInstaller(ctx context.Context, fs afero.Fs, pathResolver metadata.PathResolver, apiReader client.Reader, certPath, digest string, dynakube *dynatracev1beta1.DynaKube) (installer.Installer, error) {
```

src/installer/image/installer.go
```diff
@@ -17,35 +19,45 @@ import (
...
-func (installer *imageInstaller) UpdateProcessModuleConfig(targetDir string, processModuleConfig *dtypes.ProcessModuleConfig) error {
-	return processmoduleconfig.UpdateProcessModuleConfig(installer.fs, targetDir, processModuleConfig)
+func (installer *ImageInstaller) UpdateProcessModuleConfig(targetDir string, processModuleConfig *dtypes.ProcessModuleConfig) error {
+	sourceDir := installer.props.PathResolver.AgentSharedBinaryDirForDigest(installer.props.ImageDigest)
+	return processmoduleconfig.CreateAgentConfigDir(installer.fs, targetDir, sourceDir, processModuleConfig)
 }
 
-func (installer *imageInstaller) installAgentFromImage(targetDir string) error {
+func (installer *ImageInstaller) installAgentFromImage() error {
+	defer func() { _ = installer.fs.RemoveAll(CacheDir) }()
```
mjgrzybek, line 60: `error` is ignored anyway, so why not simply?
```suggestion
	defer installer.fs.RemoveAll(CacheDir)
```

mjgrzybek: Ah, probably linter is complaining...and it's right.
We should discuss how to not ignore errors in `defer`.

src/installer/image/installer.go
```diff
@@ -17,35 +19,45 @@ import (
...
 }
 
-func (installer *imageInstaller) InstallAgent(targetDir string) error {
-	log.Info("installing agent", "target dir", targetDir)
-	if err := installer.installAgentFromImage(targetDir); err != nil {
+func (installer *ImageInstaller) ImageDigest() string {
+	return installer.props.ImageDigest
+}
+
+func (installer *ImageInstaller) InstallAgent(targetDir string) error {
```
mjgrzybek, line 43:
```suggestion
func (installer ImageInstaller) InstallAgent(targetDir string) error {
```

src/installer/url/installer.go
```diff
@@ -25,35 +25,36 @@ func (props *Properties) fillEmptyWithDefaults() {
...
 		dtc:   dtc,
 		props: props,
 	}
 }
 
-func (installer *urlInstaller) InstallAgent(targetDir string) error {
+func (installer *UrlInstaller) InstallAgent(targetDir string) error {
 	log.Info("installing agent", "target dir", targetDir)
 	installer.props.fillEmptyWithDefaults()
+	_ = installer.fs.MkdirAll(targetDir, 0755)
```
mjgrzybek, line 45: Handle error.

mjgrzybek (dismissed):

src/installer/image/installer.go
0sewa0: it modifies the state (its not nice but must happen) 😞

src/controllers/csi/provisioner/agent_test.go
```diff
@@ -336,13 +330,25 @@ func updateOneagent(t *testing.T, alreadyInstalled bool) {
...
+
+	return updater
+}
+
+func createTestAgentImageUpdater(t *testing.T, dk *dynatracev1beta1.DynaKube, obj ...client.Object) *agentUpdater {
+	path := metadata.PathResolver{RootDir: "test"}
+	fs := afero.NewMemMapFs()
+	rec := record.NewFakeRecorder(10)
+
+	updater, err := newAgentImageUpdater(context.TODO(), fs, fake.NewClient(obj...), path, rec, dk, "TODO")
```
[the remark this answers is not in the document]
0sewa0: the problem here is that the `Installer` part is a mock, so the digest doesn't play any role here 😬 

what is even sadder is that the `ImageInstaller` is really hard (/impossible) to test, as there are no mocks or anything for the containers packages 😢

src/controllers/csi/provisioner/agent.go
```diff
@@ -78,7 +99,7 @@ func getUrlProperties(version string) *url.Properties {
 	}
 }
 
-func setupImageInstaller(ctx context.Context, fs afero.Fs, apiReader client.Reader, certPath string, dynakube *dynatracev1beta1.DynaKube) (installer.Installer, error) {
+func setupImageInstaller(ctx context.Context, fs afero.Fs, pathResolver metadata.PathResolver, apiReader client.Reader, certPath, digest string, dynakube *dynatracev1beta1.DynaKube) (installer.Installer, error) {
```
meik99, line 102: Those are way too many parameters and must be refactored. 
Maybe replace it using the builder pattern.

src/controllers/csi/provisioner/agent.go
```diff
@@ -91,25 +112,25 @@ func setupImageInstaller(ctx context.Context, fs afero.Fs, apiReader client.Read
...
 	var updatedVersion string
-	if updater.VersionDirNotPresent() || installedVersion == "" {
+	if updater.versionDirNotPresent() || installedVersion == "" {
 		log.Info("updating agent",
 			"target version", updater.targetVersion,
 			"installed version", installedVersion,
 			"target directory", updater.targetDir)
-		_ = updater.fs.MkdirAll(updater.targetDir, 0755)
-
 		if err := updater.installer.InstallAgent(updater.targetDir); err != nil {
```
meik99, line 129: I get that it's not part of your changes, but I just want to mention that the function is called `updateAgent`, but it also installs it. Then, further down, it updates the agents config, which is not really the agent itself. I'd argue that this function wants to be split up as it does two different things none of which really do what the functions name suggests

src/controllers/csi/provisioner/controller.go
```diff
@@ -156,28 +157,38 @@ func (provisioner *OneAgentProvisioner) updateAgentInstallation(ctx context.Cont
...
 		return nil, false, err
 	}
-	if updatedVersion, err := agentUpdater.updateAgent(dynakubeMetadata.LatestVersion, dynakubeMetadata.TenantUUID, latestProcessModuleConfigCache); err != nil {
+	if updatedVersion, err := agentUpdater.updateAgent(dynakubeMetadata.LatestVersion, latestProcessModuleConfigCache); err != nil {
 		log.Info("error when updating agent", "error", err.Error())
 		// reporting error but not returning it to avoid immediate requeue and subsequently calling the API every few seconds
 		return nil, true, nil
 	} else if updatedVersion != "" {
 		dynakubeMetadata.LatestVersion = updatedVersion
+		imageInstaller, ok := agentUpdater.installer.(*image.ImageInstaller)
```
meik99, line 188:
```suggestion
		imageInstaller, isImageInstaller := agentUpdater.installer.(*image.ImageInstaller)
```

src/controllers/csi/provisioner/controller.go
```diff
@@ -156,28 +157,38 @@ func (provisioner *OneAgentProvisioner) updateAgentInstallation(ctx context.Cont
...
 
 	latestProcessModuleConfigCache = newProcessModuleConfigCache(latestProcessModuleConfig)
 
-	agentUpdater, err := newAgentUpdater(ctx, provisioner.fs, provisioner.apiReader, dtc, provisioner.path, provisioner.recorder, dk)
 	if err != nil {
 		log.Info("error when setting up the agent updater", "error", err.Error())
 		return nil, false, err
 	}
-	if updatedVersion, err := agentUpdater.updateAgent(dynakubeMetadata.LatestVersion, dynakubeMetadata.TenantUUID, latestProcessModuleConfigCache); err != nil {
+	if updatedVersion, err := agentUpdater.updateAgent(dynakubeMetadata.LatestVersion, latestProcessModuleConfigCache); err != nil {
```
meik99, line 182: assignment and condition should be split up for better readability. I almost missed that there was an if when reviewing

src/controllers/csi/provisioner/events.go
```diff
@@ -0,0 +1,26 @@
+package csiprovisioner
+
+import (
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+)
+
+type updaterEventRecorder struct {
```
meik99, line 9: that's a good place for the events to be :+1:

src/installer/image/installer.go
```diff
@@ -2,82 +2,131 @@ package image
...
+func (installer ImageInstaller) UpdateProcessModuleConfig(targetDir string, processModuleConfig *dtypes.ProcessModuleConfig) error {
+	sourceDir := installer.props.PathResolver.AgentSharedBinaryDirForDigest(installer.props.ImageDigest)
+	return processmoduleconfig.CreateAgentConfigDir(installer.fs, targetDir, sourceDir, processModuleConfig)
 }
 
-func (installer *imageInstaller) installAgentFromImage(targetDir string) error {
-	_ = installer.fs.MkdirAll(CacheDir, 0755)
+func (installer *ImageInstaller) installAgentFromImage() error {
+	defer installer.fs.RemoveAll(CacheDir)
+	err := installer.fs.MkdirAll(CacheDir, 0755)
```
meik99, line 66: `0755` is used multiple times. Using a constant would make sense.

src/installer/image/installer.go
```diff
@@ -2,82 +2,131 @@ package image
...
+		log.Info("failed to create cache dir", "err", err)
+		return errors.WithStack(err)
+	}
 	image := installer.props.ImageUri
 
 	sourceCtx, sourceRef, err := getSourceInfo(CacheDir, *installer.props)
 	if err != nil {
 		log.Info("failed to get source information", "image", image)
-		return err
+		return errors.WithStack(err)
```
meik99, line 76: :heart:

src/installer/image/installer.go
```diff
@@ -2,82 +2,131 @@ package image
...
 
 	imageDigest, err := getImageDigest(sourceCtx, sourceRef)
 	if err != nil {
 		log.Info("failed to get image digest", "image", image)
-		return err
+		return errors.WithStack(err)
+	}
+
+	imageDigestEncoded := imageDigest.Encoded()
+	imageCacheDir := filepath.Join(CacheDir, imageDigestEncoded)
```
meik99, line 86: filepath joins should be wrapped in a function as their level of abstraction is different then the one of `installAgentFromImage`

meik99 (dismissed):

src/controllers/csi/provisioner/agent.go
0sewa0: so the agent updater as a whole needs to be refactored, we are planning to give it the webhook rework treatment that is why I didn't try to refactor the whole thing right now

chrismuellner (approved):

PR #843, opened 2022-06-21
Uses image for version label for deployed components (if set)

0sewa0:
# Description

When setting the version label for components managed by the operator (oneagent and activegate) the latest version on the cluster is used instead of the tag of the image in case an image is used.

Now when an image is used for activegate or oneagent, we use its tag as value for the `app.kubernetes.io/version` label.

## How can this be tested?
Deploy a dynakube where the oneagent/activegate image field is set, and see that the deployed daemonset/statefulset has the correct labels

## Checklist

### DIFF
--- a/src/controllers/activegate/reconciler/statefulset/statefulset.go
+++ b/src/controllers/activegate/reconciler/statefulset/statefulset.go
@@ -86,8 +86,13 @@ func NewStatefulSetProperties(instance *dynatracev1beta1.DynaKube, capabilityPro
 }
 
 func CreateStatefulSet(stsProperties *statefulSetProperties) (*appsv1.StatefulSet, error) {
+
+	versionLabelValue := stsProperties.Status.ActiveGate.Version
+	if stsProperties.DynaKube.Spec.ActiveGate.Image != "" {
+		versionLabelValue = stsProperties.DynaKube.Spec.ActiveGate.Image
+	}
 	appLabels := kubeobjects.NewAppLabels(kubeobjects.ActiveGateComponentLabel, stsProperties.DynaKube.Name,
-		stsProperties.feature, stsProperties.Status.ActiveGate.Version)
+		stsProperties.feature, versionLabelValue)
 
 	sts := &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
--- a/src/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/src/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -128,8 +128,12 @@ func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 	instance := dsInfo.instance
 	podSpec := dsInfo.podSpec()
 
+	versionLabelValue := instance.Status.OneAgent.Version
+	if instance.Image() != "" {
+		versionLabelValue = instance.Image()
+	}
 	appLabels := kubeobjects.NewAppLabels(kubeobjects.OneAgentComponentLabel, instance.Name,
-		dsInfo.deploymentType, instance.Status.OneAgent.Version)
+		dsInfo.deploymentType, versionLabelValue)
 	labels := kubeobjects.MergeLabels(
 		appLabels.BuildLabels(),
 		dsInfo.hostInjectSpec.Labels,
--- a/src/controllers/dynakube/oneagent/daemonset/daemonset_test.go
+++ b/src/controllers/dynakube/oneagent/daemonset/daemonset_test.go
@@ -1,9 +1,12 @@
 package daemonset
 
 import (
+	"strings"
 	"testing"
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
+	"github.com/Dynatrace/dynatrace-operator/src/version"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
@@ -11,7 +14,8 @@ import (
 )
 
 const (
-	testImage = "test-image"
+	testImage   = "test-image"
+	testVersion = "test-version"
 )
 
 func TestUseImmutableImage(t *testing.T) {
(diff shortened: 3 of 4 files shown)


src/controllers/activegate/reconciler/statefulset/statefulset.go
```diff
@@ -86,8 +86,13 @@ func NewStatefulSetProperties(instance *dynatracev1beta1.DynaKube, capabilityPro
 }
 
 func CreateStatefulSet(stsProperties *statefulSetProperties) (*appsv1.StatefulSet, error) {
+
+	versionLabelValue := stsProperties.Status.ActiveGate.Version
```
meik99, line 90:
```suggestion
func CreateStatefulSet(stsProperties *statefulSetProperties) (*appsv1.StatefulSet, error) {
	versionLabelValue := stsProperties.Status.ActiveGate.Version
```

src/controllers/activegate/reconciler/statefulset/statefulset.go
```diff
@@ -86,8 +86,13 @@ func NewStatefulSetProperties(instance *dynatracev1beta1.DynaKube, capabilityPro
...
+	if stsProperties.DynaKube.Spec.ActiveGate.Image != "" {
+		versionLabelValue = stsProperties.DynaKube.Spec.ActiveGate.Image
+	}
 	appLabels := kubeobjects.NewAppLabels(kubeobjects.ActiveGateComponentLabel, stsProperties.DynaKube.Name,
```
meik99, line 94:
```suggestion
	}
	
	appLabels := kubeobjects.NewAppLabels(kubeobjects.ActiveGateComponentLabel, stsProperties.DynaKube.Name,
```

src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -128,8 +128,12 @@ func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 	instance := dsInfo.instance
 	podSpec := dsInfo.podSpec()
 
+	versionLabelValue := instance.Status.OneAgent.Version
+	if instance.Image() != "" {
+		versionLabelValue = instance.Image()
+	}
 	appLabels := kubeobjects.NewAppLabels(kubeobjects.OneAgentComponentLabel, instance.Name,
```
meik99, line 135:
```suggestion
	}
	
	appLabels := kubeobjects.NewAppLabels(kubeobjects.OneAgentComponentLabel, instance.Name,
```

meik99 (commented):
I am just nitpicking, you can ignore my comments or commit the suggestions as they are only concerning white-space. I will now continue to test, but it lgtm

meik99 (changes requested):
An image is apparently not a valid label value
```
{"error":"StatefulSet.apps \"dynakube-activegate\" is invalid: [metadata.labels: Invalid value: \"asj34817.dev.dynatracelabs.com/linux/activegate:latest\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

src/api/v1beta1/properties.go
```diff
@@ -387,3 +399,10 @@ func splitArg(arg string) (key, value string) {
 	value = split[1]
 	return
 }
+
+func getImageTag(imageURI string) string {
+	if !strings.Contains(imageURI, ":") {
+		return ""
+	}
+	return strings.Split(imageURI, ":")[1]
```
meik99, line 407: If an image with a digest is supplied, that would also break the label.
Then the tag returned would be something like `sha256@<hash>` and an `@` is not allowed in a label value either.
I'd suggest normalizing the string.
E.g., any character not matching a regex is replaced with a dash or removed or something similar.

src/api/v1beta1/properties.go
```diff
@@ -387,3 +399,10 @@ func splitArg(arg string) (key, value string) {
 	value = split[1]
 	return
 }
+
+func getImageTag(imageURI string) string {
+	if !strings.Contains(imageURI, ":") {
+		return ""
```
meik99, line 405: If an image without a tag is supplied this would return nothing.
While technically correct, maybe we should instead go with the docker convention and assume that it's `latest`?

src/controllers/dynakube/oneagent/daemonset/daemonset_test.go
```diff
@@ -53,6 +58,92 @@ func TestUseImmutableImage(t *testing.T) {
...
+		ds, err := dsInfo.BuildDaemonSet()
+		require.NoError(t, err)
+
+		podSpecs := ds.Spec.Template.Spec
+		assert.NotNil(t, podSpecs)
+		assert.Equal(t, testImage, podSpecs.Containers[0].Image)
+		assert.Equal(t, expectedLabels, ds.Labels)
+		// 		assert.Equal(t, metav1.LabelSelector{
+		// 	MatchLabels: expectedMatchLabels,
+		// }, *sts.Spec.Selector)
```
meik99, line 140: remove unused code

src/controllers/dynakube/oneagent/daemonset/daemonset_test.go
```diff
@@ -53,6 +58,92 @@ func TestUseImmutableImage(t *testing.T) {
...
+		ds, err := dsInfo.BuildDaemonSet()
+		require.NoError(t, err)
+
+		podSpecs := ds.Spec.Template.Spec
+		assert.NotNil(t, podSpecs)
+		assert.Equal(t, instance.ImmutableOneAgentImage(), podSpecs.Containers[0].Image)
+		assert.Equal(t, expectedLabels, ds.Labels)
+		// 		assert.Equal(t, metav1.LabelSelector{
+		// 	MatchLabels: expectedMatchLabels,
+		// }, *sts.Spec.Selector)
```
meik99, line 101: remove unused code

meik99 (changes requested):

src/api/v1beta1/properties.go
```diff
@@ -387,3 +406,30 @@ func splitArg(arg string) (key, value string) {
...
+func normalizeImageTag(tag string) string {
+	notAllowedCharsRegex := regexp.MustCompile(`[^\d\w-_.]`)
+	safeTag := notAllowedCharsRegex.ReplaceAllString(tag, "-")
+	if len(safeTag) > maxLabelValueLength {
+		safeTag = safeTag[:maxLabelValueLength]
+	}
+	notAllowedEndingRegex := regexp.MustCompile(`[^\w\d]+$`)
+	safeTag = notAllowedEndingRegex.ReplaceAllString(safeTag, "")
+	return safeTag
+}
```
mjgrzybek, line 435: I think we can find a library to do it, e.g.: https://pkg.go.dev/github.com/novln/docker-parser#section-readme
Less code to maintain (prod and tests).

mjgrzybek (changes requested):

src/api/v1beta1/properties.go
mjgrzybek: Or actually this is what you need: https://github.com/moby/moby/blob/49bf474f9ed7ce7143a59d1964ff7b7fd9b52178/api/types/reference/image_reference_test.go
Moby is Docker's project :+1:.

meik99: First one is made by a dude with the only release being two years old, the second one would only retrieve the tag but not sanitize it. But since I just noticed it, those functions would be better kept in a separate package instead.

meik99, line 435: as mentioned before as a comment, those functions should be in a separate package

src/api/v1beta1/properties.go
```diff
@@ -387,3 +406,30 @@ func splitArg(arg string) (key, value string) {
...
+func getRawImageTag(imageURI string) string {
+	if !strings.Contains(imageURI, ":") {
+		return "latest"
+	}
+	splitURI := strings.Split(imageURI, ":")
+	return splitURI[len(splitURI)-1]
+}
+
+// getSafeImageTag extracts and normalizes the provided image's tag
+func getSafeImageTag(imageURI string) string {
```
meik99, line 419:
```suggestion
func getSafeImageTagForLabel(imageURI string) string {
```
so it is visible for what it has been made safe

src/api/v1beta1/properties.go
```diff
@@ -387,3 +406,30 @@ func splitArg(arg string) (key, value string) {
...
+	rawTag := getRawImageTag(imageURI)
+	return normalizeImageTag(rawTag)
+}
+
+// normalizeImageTag normalizes the tag
+// so it matches the requirements described here: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+func normalizeImageTag(tag string) string {
```
meik99, line 426:
```suggestion
func normalizeImageTagForLabel(tag string) string {
```

meik99 (changes requested):
just a few more minor things, otherwise looks good

src/api/v1beta1/properties.go
0sewa0: separate external library 👎
separate package  👍

mjgrzybek: Us maintaining regexes :-1: 
Names collision :-1: :smile: 
https://github.com/Dynatrace/dynatrace-operator/pull/843#discussion_r908179852

PR #847, opened 2022-06-22
Updates the image csi-gc to clean up unused shared images

0sewa0:
# Description
When the `codeModulesImage` is used then shared images are stored/unpacked under `/data/codemodules/<image-digest>`

To clean these folders up we check in the metadata database for the used image digests. (stored in the `dynakube table` + the versions in the `volumes` table)
We compare that to the folders we see, and if the folder's name is an image-digest that is not present in the database than we delete it.

~***Important sidenote***: I think the current implementation needs some changes, because if we delete a dynakube that has the digest that is used by a few volumes, we would delete the lowerdir from under them. In this case I don't know what will happen, I still have to test that.~ (fixed in: https://github.com/Dynatrace/dynatrace-operator/pull/847/commits/17c87ad2d3041aff9f8ad31f8b812056095fd54a)

## How can this be tested?
Have a set of dynakubes with `codeModulesImage` that point to images that have different digests.
Deploy them, wait for the downloads to finish, then delete one of the dynakubes.
Than check if `/data/codemodules/<image-digest>` was removed successfully.
**Important sidenote**:  The current GC logic is that it only runs if at least 1 dynakube is in the cluster. (which is not optimal IMO, but fixing that is not trivial and out of scope for this change)

## Checklist


chrismuellner (dismissed):

src/controllers/csi/gc/controller.go
```diff
@@ -22,6 +23,14 @@ func (set pinnedVersionSet) isNotPinned(version string) bool {
 	return !set[version]
 }
 
+// garbageCollectionInfo is a collection of tenant specific information
+// that is needed to safely delete unused files/directories connected to that tenant
```
meik99, line 27:
```suggestion
// garbageCollectionInfo stores tenant specific information
// used to delete unused files or directories connected to that tenant
```

src/controllers/csi/gc/controller.go
```diff
@@ -52,55 +61,89 @@ func (gc *CSIGarbageCollector) Reconcile(ctx context.Context, request reconcile.
...
+	log.Info("running shared images garbage collection")
+	if err := gc.runSharedImagesGarbageCollection(); err != nil {
+		log.Info("failed to garbage collect the shared images")
+		return reconcileResult, err
+	}
+
+	return reconcileResult, nil
+}
+
+func getDynakubeFromRequest(ctx context.Context, apiReader client.Reader, request reconcile.Request) (*dynatracev1beta1.DynaKube, error) {
```
meik99, line 95: context, apiReader and request are used together quite frequently. Would it make sense to combine them in a struct and call the functions as receiver functions instead?
Maybe something like `garbageCollectionRequest` that would result in `gcRequest.getDynakube()` or something like that

src/controllers/csi/gc/images.go
```diff
@@ -2,78 +2,93 @@ package csigc
...
+	usedImageDigests, err := gc.db.GetUsedImageDigests()
 	if err != nil {
-		return modificationTime, err
+		log.Info("failed to get the used image digests")
+		return nil, err
+	}
+
+	// the version of a Volume is the imageDigest if the shared image was used during mount
+	// this will still can contain versions that are not imageDigests,
+	// however this shouldn't cause issues as those versions doesn't matter in this context
```
meik99, line 72:
```suggestion
	// If a shared image was used during mount, the version of a Volume is the imageDigest.
	// A Volume can still reference versions that are not imageDigests.
	// However, this shouldn't cause issues as those versions don't matter in this context.
```

src/controllers/csi/gc/images.go
```diff
@@ -2,78 +2,93 @@ package csigc
...
+	usedVersions, err := gc.db.GetAllUsedVersions()
```
meik99, line 73: if only digests are needed, why isn't `GetUsedImageDigests` used instead?

meik99 (changes requested):

src/controllers/csi/gc/images.go
0sewa0: the version in the volume is the digest if the unpacked shared image is mounted, so to not kick the dir out of running user containers if the dynakube(s) was removed

meik99 (changes requested):

meik99 (dismissed):

meik99 (approved):

PR #859, opened 2022-06-23
Added flag in values.yaml and added check

waodim:
# Description

The new PVC monitoring feature for ActiveGate needs an additional permission to work. This permission can be enabled via a flag in the `values.yaml`.

## How can this be tested?

Set the flag to `true` in the `values.yaml`, apply the rendered helm chart and check the clusterrole `dynatrace-kubernetes-monitoring`. In the resources there should now be `nodes/metrics`.

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring.yaml
+++ b/config/helm/chart/default/templates/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring.yaml
@@ -32,6 +32,9 @@ rules:
       - pods/proxy
       - nodes/proxy
       - services
+      {{- if eq (default false .Values.additionalPermissions.pvcMonitoring) true}}
+      - nodes/metrics
+      {{- end }}
     verbs:
       - list
       - watch
--- a/config/helm/chart/default/tests/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring_test.yaml
+++ b/config/helm/chart/default/tests/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring_test.yaml
@@ -90,3 +90,95 @@ tests:
               - /livez
             verbs:
               - get
+
+  - it: should add additional permissions when flag is enabled
+    set:
+      additionalPermissions.pvcMonitoring: false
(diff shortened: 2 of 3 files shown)


config/helm/chart/default/templates/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring.yaml
```diff
@@ -32,6 +32,9 @@ rules:
       - pods/proxy
       - nodes/proxy
       - services
+      {{- if eq (default false .Values.additionalPermissions.pvcMonitoring) true}}
```
0sewa0, line 35: this is enough: 
```suggestion
      {{- if default false .Values.additionalPermissions.pvcMonitoring}}
```

0sewa0 (changes requested):

config/helm/chart/default/tests/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring_test.yaml
```diff
@@ -90,3 +90,95 @@ tests:
...
+              - events
+              - resourcequotas
+              - pods/proxy
+              - nodes/proxy
+              - services
+              - nodes/metrics
+            verbs:
+              - list
+              - watch
+              - get
```
meik99, line 126: only this assert is needed as the other ones are already checked in previous tests

meik99 (changes requested):

config/helm/chart/default/templates/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring.yaml
```diff
@@ -32,6 +32,9 @@ rules:
       - pods/proxy
       - nodes/proxy
       - services
+      {{- if default false .Values.additionalPermissions.pvcMonitoring}}
```
0sewa0, line 35:
```suggestion
      {{- if default false (.Values.additionalPermissions).pvcMonitoring}}
```
just to be safe

meik99 (approved):

PR #865, opened 2022-06-23
Code changes to support Power and Z

jsm84:
# Description

This PR omits the bundle/manifests/metadata portion of the changes necessary to support Power and Z on OpenShift, in favor of getting the code/build script changes merged first. This is to allow for multi-arch image infrastructure (registry / image hosting) changes to come online prior to merging any operator bundle/manifest updates.

Change Summary:
* Added support for Power and Z architectures to the newly refactored pod mutator webhook
* Created a new build script `./hack/build/deploy_rhel.sh` which is a RHEL-specific version of the older `deploy_local.sh` script (this is necessary as building/linking via CGO to alpine musl-libc does not work for Power or Z)
* Minor edit to `push_image.sh` to allow the use of other registries or quay namespaces
* Added workaround code (admittedly a hack) specifically for Z/s390x OneAgent installation:
   * Gracefully handle the condition where `ruxitagentproc.conf` is missing from the OneAgent zip installer archive
   * Additionally generate the static contents of `ruxitagentproc.conf` as they should be reflected for Z/s390x.
   * Inject tenant-specific connection info (UUID, token, communication site list) into the in-memory process module config to be merged with the statically generated "on disk" config file.

The motivation for this is of joint concern to Dynatrace, Red Hat and IBM to target existing customers running OpenShift on these IBM hardware platforms.

### Build Requirements
To build this yourself, you'd need the following:
* A RHEL 8 VM running natively on Power or Z in order to build the operator (native container-based builds are now possible but not provided with this PR) with the following installed
   * Go 1.18 installed from go.dev
   * The following additional RHEL packages: `git gcc gcc-c++ make device-mapper-devel golang podman podman-docker vim`
   * Operator SDK v1.16.0

### Build Steps:

With the above installed and running on the local RHEL 8 VM, the `deploy_rhel.sh` script is called as follows:

```
$ GOARCH=<s390x|ppc64le> LOCALBUILD=1 \
 IMG=quay.io/orgname/dynatrace-operator TAG=v0.6.0-<arch> \
./hack/build/deploy_rhel.sh
```
You should now have the operator image built and pushed to the quay repository specified, with the tag that was specified.

## Checklist
- [ ] I have read and understood the [contribution guidelines](/CONTRIBUTING.md)


src/processmoduleconfig/update_s390x.go
```diff
@@ -0,0 +1,141 @@
+//go:build s390x
+//+build s390x
+
+package processmoduleconfig
```
0sewa0, line 4: wouldn't it be enough to have 
`func createProcessModuleConfigFile(fs afero.Fs, destPath string) error {` in a separate files arch dependent files, so we minimize duplicate logic ?

src/standalone/run_s390x.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+	processModuleConfig, err := runner.dtclient.GetProcessModuleConfig(0)
+	if err != nil {
+		return err
+	}
+	connectionInfo, err := runner.dtclient.GetConnectionInfo()
+	if err != nil {
+		return err
+	}
```
0sewa0, line 123: this is the only difference from the original, the difference is minimal, so just a simple `if arch.Arch == arch.ArchS390` should be enough

src/webhook/mutation/pod_mutator/oneagent_mutation/annotations.go
```diff
@@ -26,7 +27,8 @@ func setInjectedAnnotation(pod *corev1.Pod) {
 
 func getInstallerInfo(pod *corev1.Pod) installerInfo {
 	return installerInfo{
-		flavor:        kubeobjects.GetField(pod.Annotations, dtwebhook.AnnotationFlavor, arch.FlavorMultidistro),
+		arch:          kubeobjects.GetField(pod.Annotations, dtwebhook.AnnotationArch, arch.Arch),
```
chrismuellner, line 30: this annotation has to be removed, since the architecture of the webhook is not necessarily the same as that of the injected pod

chrismuellner (changes requested):

src/webhook/mutation/pod_mutator/oneagent_mutation/annotations.go
jsm84: @chrismuellner thanks for the code review and quick fix. You had a solid point in the case of mixed node architecture environments. The best way is to have the standalone installer / init container handle it.

There is a holdover issue where `FlavorMultidistro` does not exist for s390x or ppc64le. The same issue occurs where we assume so in the webhook, and inject that into the pod initContainer.

So we should probably also set the install flavor to `arch.Flavor` in the standalone installer, as you did for `arch.Arch` in #880.

chrismuellner: makes sense, we'll do that in a separate PR as well to fix it for the next release already.
Thanks for pointing that out.

src/processmoduleconfig/update_s390x.go
jsm84: @0sewa0 I pushed new commit to correct this. I couldn't get around duplicating `createProcessModuleConfigFile()` for both s390x and other architectures, but the rest of the duplicate code was removed.

src/webhook/mutation/pod_mutator/oneagent_mutation/annotations.go
jsm84: @chrismuellner That sounds perfect!

I've restored the original source files for the pod mutator webhook from main, reverting any changes to those pkg's.

src/standalone/run_s390x.go
jsm84: @0sewa0 resolved this as requested.

jsm84:
Fixed an error with remnant `go:build` directives for s390x in `./src/standalone/run.go` and rebased against main branch.

src/processmoduleconfig/create_config_s390x.go
```diff
@@ -0,0 +1,73 @@
...
+	err = populateProcessModuleConfigFile(newProcessModuleConfigFile, destPath)
+	if err != nil {
+		return err
+	}
+
+	return newProcessModuleConfigFile.Close()
+}
+
+// populateProcessModuleConfigFile populates the empty config file on s390x
+func populateProcessModuleConfigFile(cfg afero.File, destPath string) error {
```
chrismuellner, line 32: this logic should be removed as it can't be kept in sync with the actual archive.

Passing the correct architectures to the api should be enough to support ppcle and s390 when they are available on the api.

chrismuellner (dismissed):

PR #881, opened 2022-06-27
Adds google-autopilot as a new platform to helm

0sewa0:
# Description

In gke autopilot version 1.23x the `nodeAffinity` specifying the arch for the node is very picky, and only supports a single value which is `amd64` so having it would just complicate things.
So on gke-autopilot we will not have them.

Added `google-autopilot` as a new possible platform, if used no `nodeAffinity` will be set for checking the arch.

## How can this be tested?
Deploy to a `rapid` channel autopilot cluster, while setting the platform to `google-autopilot`. (remember autopilot doesn't support CSI driver)

## Checklist

### DIFF
--- a/hack/make/manifests/config.mk
+++ b/hack/make/manifests/config.mk
@@ -11,6 +11,7 @@ HELM_CRD_DIR=$(HELM_TEMPLATES_DIR)/Common/crd/
 MANIFESTS_DIR=config/deploy/
 
 KUBERNETES_CORE_YAML=$(MANIFESTS_DIR)kubernetes/kubernetes.yaml
+KUBERNETES_AUTOPILOT_YAML=$(MANIFESTS_DIR)kubernetes/gke-autopilot.yaml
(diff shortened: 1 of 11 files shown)


hack/make/manifests/config.mk
```diff
@@ -11,6 +11,7 @@ HELM_CRD_DIR=$(HELM_TEMPLATES_DIR)/Common/crd/
 MANIFESTS_DIR=config/deploy/
 
 KUBERNETES_CORE_YAML=$(MANIFESTS_DIR)kubernetes/kubernetes.yaml
+KUBERNETES_AUTOPILOT_YAML=$(MANIFESTS_DIR)kubernetes/gke-autopilot.yaml
```
chrismuellner, line 14:
```suggestion
KUBERNETES_AUTOPILOT_YAML=$(MANIFESTS_DIR)kubernetes/google-autopilot.yaml
```

PR #888, opened 2022-06-28
Reworks dockerConfig auth setup

0sewa0:
# Description
The `NewDockerConfig` constructor will not setup the `Auths`, that is done in `SetupAuths` method.
This is a small optimisation so in case something needs the dockerConfig but requires no authentication, we will no longer force to use the default pull secret. (Also its nicer for testing)

## How can this be tested?
When you use `codeModulesImage` that points to a registry that doesn't need `customPullSecret`, it should work

## Checklist

### DIFF
--- a/src/controllers/csi/provisioner/agent.go
+++ b/src/controllers/csi/provisioner/agent.go
@@ -100,9 +100,12 @@ func getUrlProperties(version string) *url.Properties {
 }
 
 func setupImageInstaller(ctx context.Context, fs afero.Fs, pathResolver metadata.PathResolver, apiReader client.Reader, certPath, digest string, dynakube *dynatracev1beta1.DynaKube) (installer.Installer, error) {
-	dockerConfig, err := dockerconfig.NewDockerConfig(ctx, apiReader, *dynakube)
-	if err != nil {
-		return nil, err
+	dockerConfig := dockerconfig.NewDockerConfig(apiReader, *dynakube)
+	if dynakube.Spec.CustomPullSecret != "" {
+		err := dockerConfig.SetupAuths(ctx)
+		if err != nil {
+			return nil, err
+		}
 	}
 	if dynakube.Spec.TrustedCAs != "" {
 		err := dockerConfig.SaveCustomCAs(ctx, afero.Afero{Fs: fs}, certPath)
--- a/src/dockerconfig/docker_config.go
+++ b/src/dockerconfig/docker_config.go
@@ -3,10 +3,10 @@ package dockerconfig
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	"github.com/Dynatrace/dynatrace-operator/src/dtclient"
+	"github.com/pkg/errors"
 	"github.com/spf13/afero"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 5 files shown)


src/dockerconfig/docker_config.go
```diff
@@ -25,47 +25,30 @@ type DockerAuth struct {
...
+	if err != nil {
+		log.Info("failed to load pull secret", "dynakube", config.Dynakube.Name)
+		return errors.WithStack(err)
 	}
-	err := json.Unmarshal(config, &dockerConf)
+	dockerAuths, err := parseDockerAuthsFromSecret(&pullSecret)
 	if err != nil {
-		return nil, err
+		log.Info("failed to parse pull secret content", "dynakube", config.Dynakube.Name)
+		return err
```
luhi-DT, line 48:
```suggestion
		return errors.WithStack(err)
```

src/dockerconfig/docker_config.go
```diff
@@ -76,15 +59,36 @@ func (config *DockerConfig) SaveCustomCAs(
...
-		return err
+		return errors.WithStack(err)
 	}
 	config.TrustedCertsPath = path
 	return nil
 }
+
+func parseDockerAuthsFromSecret(secret *corev1.Secret) (map[string]DockerAuth, error) {
+	if secret == nil {
+		return nil, errors.New("given secret is nil")
```
luhi-DT, line 77: could this be a more specific error message ? could not generate docker auths from empty secret

src/controllers/csi/provisioner/agent.go
```diff
@@ -100,9 +100,12 @@ func getUrlProperties(version string) *url.Properties {
 }
 
 func setupImageInstaller(ctx context.Context, fs afero.Fs, pathResolver metadata.PathResolver, apiReader client.Reader, certPath, digest string, dynakube *dynatracev1beta1.DynaKube) (installer.Installer, error) {
-	dockerConfig, err := dockerconfig.NewDockerConfig(ctx, apiReader, *dynakube)
-	if err != nil {
-		return nil, err
+	dockerConfig := dockerconfig.NewDockerConfig(apiReader, *dynakube)
+	if dynakube.Spec.CustomPullSecret != "" {
```
luhi-DT, line 104: maybe refactor in a seperate function like `hasCustomPullSecret`

src/dockerconfig/docker_config.go
```diff
@@ -25,47 +25,30 @@ type DockerAuth struct {
...
-		return nil, err
-	}
-
+func NewDockerConfig(apiReader client.Reader, dynakube dynatracev1beta1.DynaKube) *DockerConfig {
 	dockerConfig := DockerConfig{
 		ApiReader:     apiReader,
+		Auths:         make(map[string]DockerAuth),
 		Dynakube:      &dynakube,
-		Auths:         dockerAuths,
 		SkipCertCheck: dynakube.Spec.SkipCertCheck,
```
luhi-DT, line 33: I dont think we would need SkipCertCheck to be part of the struct, as we have the dynakube as a whole anyway

luhi-DT (changes requested):

src/dockerconfig/docker_config.go
meik99: that error comes from an internal function, `err` is fine

src/controllers/csi/provisioner/agent.go
0sewa0: I don't think its necessary, its a higher level property

`dynakube.Spec.TrustedCAs != ""` should also need to be replaced to remain consistent

src/dockerconfig/docker_config.go
0sewa0: made it a method, so we don't have to dig for it 👍

src/dockerconfig/docker_config.go
luhi-DT: makes sense

luhi-DT (approved):

PR #898, opened 2022-06-28
Uses arch.Flavor when annotation is not set

0sewa0:
# Description
Webhook no longer sets a default value for flavor, and if not set the standalone init uses the `arch.Flavor`

## How can this be tested?
Deploy anything, not set the flavor annotation, then on a amd node multidistro should be used, while in an arm node default

## Checklist


src/standalone/env.go
```diff
@@ -108,7 +108,11 @@ func (env *environment) addInstallerFlavor() error {
 	if err != nil {
 		return err
 	}
-	env.InstallerFlavor = flavor
+	if flavor == "" {
```
luhi-DT, line 111: checkEnvVar() returns an error in case the envVar is not set, so this code is not reachable and flavor is always set to something

luhi-DT (changes requested):

src/standalone/env.go
0sewa0: fixed it by making it an "optional" env, which is actually true

luhi-DT (approved):

PR #905, opened 2022-06-29
Makes ruxitagentproc.conf read/writeable by all users

0sewa0:
# Description

Makes it so that the `ruxitagentproc.conf` has filemode 666 (read/write for all)

## How can this be tested?
Deploy the CSI driver (on a cluster whose nodes doesn't have any previous installs)
Deploy Dynakube with codeModules image or something that uses the CSI driver
Check if in the sample app everyone has read/write rights `/opt/dynatrace/oneagent-paas/agent/conf/ruxitagentproc.conf` file 

## Checklist


meik99 (dismissed):

luhi-DT:
What if the agent wants to add files to that deirectory, he would not be able to do that right? Is this a use case ?

src/installer/zip/zip.go
```diff
@@ -55,11 +55,6 @@ func extractFileFromZip(fs afero.Fs, targetDir string, file *zip.File) error {
 
 	mode := file.Mode()
 
-	// Mark all files inside ./agent/conf as group-writable
-	if file.Name != common.AgentConfDirPath && isAgentConfFile(file.Name) {
```
chrismuellner, line 59: still required, directory should be writable by everyone

chrismuellner (dismissed):

meik99 (approved):

PR #908, opened 2022-06-29
Add feature flag to opt-into injection per pod 

StefanHauth:
# Description
Added feature flag `feature.dynatrace.com/automatic-injection` to disable automatic injection for a dynakube. If set to `false`, pods nominated for injection need to be manually annotated with `oneagent.dynatrace.com/inject=true`. Default value for the feature flag is `true`.

## How can this be tested?
- Set the annotation `feature.dynatrace.com/automatic-injection: "false"` on a dynakube
- Create a pod or restart one in a namespace that is associated with the used dynakube
- Check that there is no injection happening on the pod.
- Add the annotation `oneagent.dynatrace.com/inject: "true"` to the pod spec and make sure the pod restarts
- Check that there is injection happening on the pod.

## Checklist

### DIFF
--- a/src/api/v1beta1/feature_flags.go
+++ b/src/api/v1beta1/feature_flags.go
@@ -86,6 +86,7 @@ const (
 
 	AnnotationFeatureIgnoreUnknownState = AnnotationFeaturePrefix + "ignore-unknown-state"
 	AnnotationFeatureIgnoredNamespaces  = AnnotationFeaturePrefix + "ignored-namespaces"
+	AnnotationFeatureAutomaticInjection = AnnotationFeaturePrefix + "automatic-injection"
 )
 
 var (
@@ -173,6 +174,13 @@ func (dk *DynaKube) FeatureAutomaticKubernetesApiMonitoringClusterName() string 
 // FeatureDisableMetadataEnrichment is a feature flag to disable metadata enrichment,
 func (dk *DynaKube) FeatureDisableMetadataEnrichment() bool {
 	return dk.getDisableFlagWithDeprecatedAnnotation(AnnotationFeatureMetadataEnrichment, AnnotationFeatureDisableMetadataEnrichment)
+}
+
+// FeatureAutomaticInjection controls OneAgent is injected to pods in selected namespaces automatically ("automatic-injection=true" or flag not set)
+// or if pods need to be opted-in one by one ("automatic-injection=false")
+func (dk *DynaKube) FeatureEnableAutomaticInjection() bool {
+	autoInjectionFlag := dk.getFeatureFlagRaw(AnnotationFeatureAutomaticInjection)
+	return autoInjectionFlag == "true" || len(autoInjectionFlag) == 0
 }
 
 // FeatureUseActiveGateImageForStatsd is a feature flag that makes the operator use ActiveGate image when initializing Extension Controller and Statsd containers
--- a/src/mapper/mapper.go
+++ b/src/mapper/mapper.go
@@ -79,52 +79,50 @@ func match(dk *dynatracev1beta1.DynaKube, namespace *corev1.Namespace) (bool, er
 // updateNamespace tries to match the namespace to every dynakube with codeModules
 // finds conflicting dynakubes(2 dynakube with codeModules on the same namespace)
 // adds/updates/removes labels from the namespace.
-func updateNamespace(namespace *corev1.Namespace, dkList *dynatracev1beta1.DynaKubeList) (bool, error) {
-	var updated bool
+func updateNamespace(namespace *corev1.Namespace, deployedDynakubes *dynatracev1beta1.DynaKubeList) (bool, error) {
+	namespaceUpdated := false
 	conflict := ConflictChecker{}
-	for i := range dkList.Items {
-		dynakube := &dkList.Items[i]
+	for i := range deployedDynakubes.Items {
+		dynakube := &deployedDynakubes.Items[i]
 		if isIgnoredNamespace(dynakube, namespace.Name) {
 			return false, nil
 		}
 		matches, err := match(dynakube, namespace)
+
 		if err != nil {
-			return updated, err
+			return namespaceUpdated, err
 		}
 		if matches {
 			if err := conflict.check(dynakube); err != nil {
-				return updated, err
+				return namespaceUpdated, err
 			}
 		}
 
-		upd, err := updateLabels(matches, dynakube, namespace)
-		if err != nil {
-			return updated, err
-		}
-		if upd {
-			updated = true
-		}
+		labelsUpdated := updateLabels(matches, dynakube, namespace)
+		namespaceUpdated = labelsUpdated || namespaceUpdated
 	}
-	return updated, nil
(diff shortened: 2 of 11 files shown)


src/api/v1beta1/feature_flags.go
```diff
@@ -175,6 +177,13 @@ func (dk *DynaKube) FeatureDisableMetadataEnrichment() bool {
 	return dk.getDisableFlagWithDeprecatedAnnotation(AnnotationFeatureMetadataEnrichment, AnnotationFeatureDisableMetadataEnrichment)
 }
 
+// FeatureAutomaticInjection controls OneAgent is injected to pods in selected namespaces automatically ("automatic-injection=true" or flag not set)
+// or if pods need to be opted-in one by one ("automatic-injection=false")
+func (dk *DynaKube) FeatureAutomaticInjection() bool {
+	autoInjectionFlag := dk.getFeatureFlagRaw(AnnotationFeatureAutomaticInjection)
+	return autoInjectionFlag == "true" || len(autoInjectionFlag) == 0
```
0sewa0, line 184:
```suggestion
func (dk *DynaKube) FeatureDisableAutomaticInjection() bool {
	return dk.getFeatureFlagRaw(AnnotationFeatureAutomaticInjection) == "false"
```

src/mapper/mapper.go
```diff
@@ -79,52 +79,50 @@ func match(dk *dynatracev1beta1.DynaKube, namespace *corev1.Namespace) (bool, er
 // updateNamespace tries to match the namespace to every dynakube with codeModules
 // finds conflicting dynakubes(2 dynakube with codeModules on the same namespace)
 // adds/updates/removes labels from the namespace.
-func updateNamespace(namespace *corev1.Namespace, dkList *dynatracev1beta1.DynaKubeList) (bool, error) {
-	var updated bool
+func updateNamespace(namespace *corev1.Namespace, installedDynaKubes *dynatracev1beta1.DynaKubeList) (bool, error) {
```
0sewa0, line 82: `deployedDynaKubes` (dynakubes are not "installed" 😄 )

src/mapper/mapper.go
```diff
@@ -79,52 +79,50 @@ func match(dk *dynatracev1beta1.DynaKube, namespace *corev1.Namespace) (bool, er
...
+	namespaceUpdated := false
```
0sewa0, line 83:
```suggestion
	updated := false
```
nitpick:
we are in updateNamespace, we don't have to be so verbose (also you didn't update the name here: `updateLabels` 😅 )

src/mapper/namespaces.go
```diff
@@ -32,12 +32,12 @@ func (nm NamespaceMapper) MapFromNamespace() (bool, error) {
 }
 
 func (nm NamespaceMapper) updateNamespace() (bool, error) {
-	dkList := &dynatracev1beta1.DynaKubeList{}
-	err := nm.client.List(nm.ctx, dkList)
+	installedDynakubes := &dynatracev1beta1.DynaKubeList{}
```
0sewa0, line 35: `deployedDynaKubes` (dynakubes are not "installed" 😄 )

src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -58,6 +58,6 @@ func getBasePodName(pod *corev1.Pod) string {
 	return basePodName
 }
 
-func addToInitContainers(pod *corev1.Pod, initContainer *corev1.Container) {
+func attachInitContainerToPod(pod *corev1.Pod, initContainer *corev1.Container) {
```
0sewa0, line 61: `attach` is not really what the function does, it just adds the init container to a list.

`inject` is what the webhook does, `attach` is mostly used when talking about doing something to something that is already running (attaching to a running container)

0sewa0:
```suggestion
func addInitContainerToPod(pod *corev1.Pod, initContainer *corev1.Container) {
```

src/webhook/mutation/pod_mutator/pod_mutator.go
```diff
@@ -96,6 +97,12 @@ func (webhook *podMutatorWebhook) isInjected(mutationRequest *dtwebhook.Mutation
 }
 
 func (webhook *podMutatorWebhook) handlePodMutation(mutationRequest *dtwebhook.MutationRequest) error {
+
+	if !kubeobjects.GetFieldBool(mutationRequest.BaseRequest.Pod.Annotations, dtwebhook.AnnotationOneAgentInject,
+		mutationRequest.BaseRequest.DynaKube.FeatureAutomaticInjection()) {
+		return nil
+	}
```
0sewa0, line 104: this should not be here, you should check it in the affected mutators, so at:
https://github.com/Dynatrace/dynatrace-operator/blob/master/src/webhook/mutation/pod_mutator/dataingest_mutation/mutator.go#L29-L33
https://github.com/Dynatrace/dynatrace-operator/blob/master/src/webhook/mutation/pod_mutator/oneagent_mutation/mutator.go#L32-L34

0sewa0: this is so that we don't start mixing things together (as it was previously before the refactor of the webhook)

src/mapper/mapper.go
StefanHauth: That's to clearly distinguish between the two update flags (`updated` and `upd`) uses here and increase readability. 'updateLabels' only uses one flag.

src/webhook/mutation/pod_mutator/pod_mutator.go
StefanHauth: If I do it just in the mutators (what I had before), it doesn't work, because the stuff that happens after the mutators have been iterated seems to  depend on what the mutators do. The `initContainer` requires an `injectionConfig` volume to exist, which is missing if only the mutators are skipped. Also the `initContainer` and other stuff should not be added to the pod if the injection is disabled.

0sewa0: AH! that totally makes sense. 🤦 

This is an actual bug, which needs fixing, because this would happen if the user disabled all mutations with the pod annotations.
So here is what I think should be done: We should only do anything with the pod IF at least 1 mutator was enabled.
The `handlePodMutation` should look something like: 
```
func (webhook *podMutatorWebhook) handlePodMutation(mutationRequest *dtwebhook.MutationRequest) error {
	...
	isMutated := false
	for _, mutator := range webhook.mutators {
		if !mutator.Enabled(mutationRequest.BaseRequest) {
			continue
		}
		if err := mutator.Mutate(mutationRequest); err != nil {
			return err
		}
		isMutated = true
	}
	if !isMutated {
		log.Info("no mutators where enabled")
		return nil
	}
    ...
}
```

StefanHauth: fixed in
https://github.com/Dynatrace/dynatrace-operator/pull/914/commits/b9ff99b4c70a7bc436d2f72b088973b07adfb428

0sewa0 (approved):

PR #914, opened 2022-06-30
Fixes standalone init and webhook integration

0sewa0:
# Description
The injection-info volume was not fully moved to the oneagent-mutator.
Right now the injection-info volume-mount for the init container is defined in the main pod-mutator, which is incorrect.

This causes an error when wanting to only use the dataingest-mutator, as the injection-info volume will not be present.

Furthermore the injection-info's clusterID field is used by the standalone init part of the dataingest mutation, which is not necessary, because that can be set as an env in the webhook.
Doing so the 2 mutators (dataingest and oneagent) can be independent as they were meant to be.

## How can this be tested?
Inject into a:
- pod that has data-ingest disabled
- pod that has oneagent disabled
- pod that has both disabled

## Checklist

### DIFF
--- a/src/standalone/env.go
+++ b/src/standalone/env.go
@@ -28,6 +28,7 @@ type environment struct {
 	K8NodeName    string `json:"k8NodeName"`
 	K8PodName     string `json:"k8PodName"`
 	K8PodUID      string `json:"k8BasePodUID"`
+	K8ClusterID   string `json:"k8ClusterID"`
 	K8BasePodName string `json:"k8BasePodName"`
 	K8Namespace   string `json:"k8Namespace"`
 
@@ -74,6 +75,7 @@ func (env *environment) setRequiredFields() error {
 		dataIngestFieldSetters := []func() error{
 			env.addWorkloadKind,
 			env.addWorkloadName,
+			env.addK8ClusterID,
 		}
 		requiredFieldSetters = append(requiredFieldSetters, dataIngestFieldSetters...)
 	}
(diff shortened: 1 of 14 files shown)


src/standalone/env_test.go
```diff
@@ -31,16 +31,84 @@ func TestNewEnv(t *testing.T) {
...
 		assert.NotEmpty(t, env.K8Namespace)
+		assert.NotEmpty(t, env.K8ClusterID)
 
 		assert.NotEmpty(t, env.WorkloadKind)
 		assert.NotEmpty(t, env.WorkloadName)
 
 		assert.True(t, env.OneAgentInjected)
 		assert.True(t, env.DataIngestInjected)
 	})
+	t.Run(`create new env for only data-ingest injection`, func(t *testing.T) {
```
StefanHauth, line 42: Should we have also a test with no injection at all?

0sewa0: not really, as this should only be run if at least 1 mutation was done.
Without injection there shouldn't be any init-container added so the standalone init shouldn't run

src/standalone/env.go
```diff
@@ -41,32 +42,45 @@ type environment struct {
...
+	}
+
+	if env.DataIngestInjected {
+		dataIngestFieldSetters := []func() error{
+			env.addWorkloadKind,
+			env.addWorkloadName,
+			env.addK8ClusterID,
+		}
+		requiredFieldSetters = append(requiredFieldSetters, dataIngestFieldSetters...)
+	}
```
meik99, line 81:
```suggestion
	if env.DataIngestInjected {
		requiredFieldSetters = append(requiredFieldSetters, getDataIngestFieldSetters())
	}
func getDataIngestFieldSetters() []func() error {	
  return []func() error{
			env.addWorkloadKind,
			env.addWorkloadName,
			env.addK8ClusterID,
		}
}
```

meik99 (changes requested):

PR #920, opened 2022-07-01
Makes standalone init env requirements injection specific

0sewa0:
# Description

The standalone init wants to read/setup envvars that are not necessary for data-ingest only injection, which the webhook will not provide ==> standalone init fails the pod

Now it only sets env fields to be required according to what injection is needed.

## How can this be tested?
Inject into a:
- pod that has data-ingest disabled
- pod that has oneagent disabled
- pod that has both disabled
all of these should work as expected

## Checklist


src/standalone/env.go
```diff
@@ -79,12 +76,33 @@ func (env *environment) setRequiredFields() error {
 }
 
 func (env *environment) setOptionalFields() {
```
chrismuellner, line 78: should be split into separate functions, e.g. injection/base and optional fields

src/standalone/env.go
```diff
@@ -206,37 +224,37 @@ func (env *environment) addK8Namespace() error {
...
 	env.InstallerUrl = url
 }
 
-func (env *environment) addOneAgentInjected() error {
-	oneAgentInjected, err := checkEnvVar(OneAgentInjectedEnv)
-	if err != nil {
-		return err
-	}
+func (env *environment) addOneAgentInjected() {
+	oneAgentInjected, _ := checkEnvVar(OneAgentInjectedEnv)
```
chrismuellner, line 251: inconsistent check for error, could also check it for optional fields

chrismuellner (changes requested):

meik99 (approved):

chrismuellner (approved):

PR #926, opened 2022-07-05
Use ARM libraries in ARM docker image

aorcholski:
## Description
Musl libraries for the right architecture are copied to the target docker image.
Linker parameters are surrounded by question marks.

## How can this be tested?
Docker images should be executable on appropriate platforms.

## Checklist
[x] PR is labeled accordingly
[x] I have read and understood the [contribution guidelines](https://github.com/CONTRIBUTING.md)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ WORKDIR /app
 # move previously cached go modules to gopath
 RUN if [ -d ./mod ]; then mkdir -p ${GOPATH}/pkg && [ -d mod ] && mv ./mod ${GOPATH}/pkg; fi;
 
-RUN CGO_ENABLED=1 go build "${GO_BUILD_ARGS}" -o ./build/_output/bin/dynatrace-operator ./src/cmd/operator/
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -11,7 +11,7 @@ WORKDIR /app
 # move previously cached go modules to gopath
 RUN if [ -d ./mod ]; then mkdir -p ${GOPATH}/pkg && [ -d mod ] && mv ./mod ${GOPATH}/pkg; fi;
 
-RUN CGO_ENABLED=1 go build "${GO_BUILD_ARGS}" -o ./build/_output/bin/dynatrace-operator ./src/cmd/operator/
+RUN CGO_ENABLED=1 go build -ldflags="${GO_BUILD_ARGS}" -o ./build/_output/bin/dynatrace-operator ./src/cmd/operator/
```
chrismuellner, line 14: this will break local builds on this branch, see `hack/build/push_image.sh`

aorcholski: right, I explicitly added "-ldflags" parameter to the "go build" command in push_image.sh script

chrismuellner (approved):

PR #932, opened 2022-07-06
Set unprivileged user/group for oneagent securitycontext

chrismuellner:
# Description

Oneagent installer needs to be run as unprivileged user.

## How can this be tested?
Deploy cloud native fullstack on Openshift.

## Checklist

### DIFF
--- a/src/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/src/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -259,6 +259,7 @@ func (dsInfo *builderInfo) securityContext() *corev1.SecurityContext {
 			Privileged:   address.Of(true),
 			RunAsNonRoot: address.Of(true),
 			RunAsUser:    address.Of(int64(1000)),
+			RunAsGroup:   address.Of(int64(1000)),
 		}
 		return securityContext
 	}
--- a/src/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/src/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -258,6 +258,7 @@ func (dsInfo *builderInfo) securityContext() *corev1.SecurityContext {
 		securityContext := &corev1.SecurityContext{
 			Privileged:   address.Of(true),
 			RunAsNonRoot: address.Of(true),
+			RunAsUser:    address.Of(int64(1000)),
 		}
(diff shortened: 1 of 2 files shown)


src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -254,16 +254,16 @@ func (dsInfo *builderInfo) imagePullSecrets() []corev1.LocalObjectReference {
...
-	if dsInfo.instance.IsOneAgentPrivileged() {
-		runAsPrivileged := true
-		securityContext := &corev1.SecurityContext{
-			Privileged: &runAsPrivileged,
-		}
-		return securityContext
+	securityContext := &corev1.SecurityContext{
+		RunAsNonRoot: address.Of(true),
+		RunAsUser:    address.Of(int64(1000)),
+		RunAsGroup:   address.Of(int64(1000)),
```
0sewa0, line 260: question: shouldn't it be the same as what we set for the operator, just to be consistent, 
or there is a reason for this user/group?

chrismuellner: I reused the user/group we previously only used in case of read only mode
since it's a different image than the operator image, the operator user wouldn't be available

0sewa0 (approved):

src/controllers/dynakube/oneagent/daemonset/daemonset_test.go
```diff
@@ -156,9 +158,71 @@ func TestInfraMon_SecurityContext(t *testing.T) {
...
+		dsInfo := NewHostMonitoring(&instance, testClusterID)
+		ds, err := dsInfo.BuildDaemonSet()
+		require.NoError(t, err)
+
+		assert.GreaterOrEqual(t, 1, len(ds.Spec.Template.Spec.Containers))
+
+		securityContext := ds.Spec.Template.Spec.Containers[0].SecurityContext
+
+		assert.NotNil(t, securityContext)
+		assert.Equal(t, address.Of(int64(1000)), securityContext.RunAsUser)
```
mjgrzybek, line 222: TIL about `assert.Equal`
```
// Pointer variable equality is determined based on the equality of the
// referenced values (as opposed to the memory addresses).
```

mjgrzybek (approved):

PR #957, opened 2022-07-12
Update `go-sqlite` dependency and adjust building process to it

mjgrzybek:
# Description
Bump `go-sqlite` version.
Hide compilation warning as described: https://github.com/mattn/go-sqlite3/issues/822

```
# github.com/mattn/go-sqlite3
sqlite3-binding.c: In function ‘sqlite3SelectNew’:
sqlite3-binding.c:128049:10: warning: function may return address of local variable [-Wreturn-local-addr]
128049 |   return pNew;
       |          ^~~~
sqlite3-binding.c:128009:10: note: declared here
128009 |   Select standin;
       |          ^~~~~~~
```

## How can this be tested?
- no regression on CI and e2e tests

## Checklist


.github/workflows/build-dockerimage.yaml
```diff
@@ -12,6 +12,9 @@ on:
       golinkerargs:
         required: true
         type: string
+      cgo_cflags:
+        required: true
```
meik99, line 16: Can we use a default instead of making it required?

mjgrzybek: We could, but it would be inconsistent with `golinkerflags`. I preferred to repeat the pattern.

mjgrzybek: There should be common logic reused in Makefiles and github actions instead of hardcoding it in each place. 
We should rework it in future and imho good starting point is having flags separated from github actions to sh scripts which can be also invoked from Makefile.

gkrenn (approved):

PR #969, opened 2022-07-14
Consolidates frequently used const into a separate package

0sewa0:
# Description
Our consts are all over the place, so I tried to come up with a package the holds the common constants.
Main reason is to avoid potential circular dependencies.

Currently there are ~3 packages that are obviously dependent on similar constant values. (and this PR targets them mainly)
1. webhook (its 3 packages here, podmutator, and each mutation)
2. init generation
3. standalone init

So I put almost all of the constants (so the ones that were shared) into a new `config` package,.
I tried to split the constants into different files depending on which feature they are connected to.
Also renamed most to have a prefix that matches this grouping so the `config` package doesn't become a big dumpster of random constants.

(There are a few constants that I didn't move just renamed)

## How can this be tested?
No change in functionality

## Checklist


src/config/enrichment_injection.go
```diff
@@ -0,0 +1,15 @@
...
+const (
+	EnrichmentEndpointSecretName = "dynatrace-data-ingest-endpoint"
+	EnrichmentFilenameTemplate   = "dt_metadata.%s"
+	EnrichmentInjectedEnv        = "DATA_INGEST_INJECTED"
+	EnrichmentWorkloadKindEnv    = "DT_WORKLOAD_KIND"
+	EnrichmentWorkloadNameEnv    = "DT_WORKLOAD_NAME"
+)
+
+var (
+	EnrichmentMountPath = filepath.Join("var", "lib", "dynatrace", "enrichment")
```
chrismuellner, line 14:
```suggestion
	EnrichmentMountPath = filepath.Join("/", "var", "lib", "dynatrace", "enrichment")
```

src/config/agent_injection.go
```diff
@@ -0,0 +1,38 @@
...
+	AgentContainerNameEnvTemplate  = "CONTAINER_%d_NAME"
+	AgentContainerImageEnvTemplate = "CONTAINER_%d_IMAGE"
+
+	AgentInjectedEnv = "ONEAGENT_INJECTED"
+)
+
+var (
+	AgentBinDirMount    = filepath.Join("mnt", "bin")
+	AgentShareDirMount  = filepath.Join("mnt", "share")
+	AgentConfigDirMount = filepath.Join("mnt", "config")
```
chrismuellner, line 37:
```suggestion
	AgentBinDirMount    = filepath.Join("/", "mnt", "bin")
	AgentShareDirMount  = filepath.Join("/", "mnt", "share")
	AgentConfigDirMount = filepath.Join("/", "mnt", "config")
```

chrismuellner: is there a reason this has `filepath.Join`? Could just be a constant

chrismuellner (changes requested):

chrismuellner (approved):

PR #981, opened 2022-07-18
Adds extra safety check for csi GC

0sewa0:
# Description

It can happen that the GC runs while a Dynakube is still downloading the codemodules image, and the GC just deletes the directory because it doesn't know according to the database if that codemodules image is used or not.
This is because we only set the digest/latestVersion in the database when said download is finished.

## How can this be tested?
Deploy >1 Dynakube at the same time, using the same `codeModulesImage`.

## Checklist


src/controllers/csi/gc/controller.go
```diff
@@ -132,17 +142,31 @@ func collectGCInfo(ctx context.Context, apiReader client.Reader, dynakube dynatr
 	}, nil
 }
 
+func isSafeToGC(access metadata.Access, dynakubeList *dynatracev1beta1.DynaKubeList) bool {
+	dkMetadataList, err := access.GetAllDynakubes()
+	if err != nil {
+		log.Info("failed to get dynakube metadata from database, err: %s")
```
mjgrzybek, line 148:
```suggestion
		log.Info("failed to get dynakube metadata from database", "err", err)
```

mjgrzybek (changes requested):

src/controllers/csi/gc/controller.go
```diff
@@ -132,17 +142,31 @@ func collectGCInfo(ctx context.Context, apiReader client.Reader, dynakube dynatr
...
+		return false
+	}
+	filteredDynakubes := filterCodeModulesImageDynakubes(dynakubeList)
+	for _, dkMetadata := range dkMetadataList {
+		if dkMetadata.LatestVersion == "" {
+			return false
+		}
+		dynakube, ok := filteredDynakubes[dkMetadata.Name]
```
meik99, line 156:
```suggestion
		dynakube, hasDynakube := filteredDynakubes[dkMetadata.Name]
```

src/controllers/csi/gc/controller.go
```diff
@@ -132,17 +142,31 @@ func collectGCInfo(ctx context.Context, apiReader client.Reader, dynakube dynatr
...
 
+func isSafeToGC(access metadata.Access, dynakubeList *dynatracev1beta1.DynaKubeList) bool {
+	dkMetadataList, err := access.GetAllDynakubes()
+	if err != nil {
+		log.Info("failed to get dynakube metadata from database, err: %s")
+		return false
+	}
+	filteredDynakubes := filterCodeModulesImageDynakubes(dynakubeList)
+	for _, dkMetadata := range dkMetadataList {
+		if dkMetadata.LatestVersion == "" {
```
meik99, line 153:
```suggestion
		if isDownloading(dkMetadata.LatestVersion) {
		
func isDownloading(latestVersion string) bool {
  return  latestVersion == ""
}
```

would make it clear why it is not safe to gc in that scenario

src/controllers/csi/gc/controller.go
```diff
@@ -132,17 +142,31 @@ func collectGCInfo(ctx context.Context, apiReader client.Reader, dynakube dynatr
...
+			return false
+		}
+		dynakube, ok := filteredDynakubes[dkMetadata.Name]
+		if ok && dynakube.CodeModulesVersion() != dkMetadata.LatestVersion {
```
meik99, line 157:
```suggestion
		if ok && isUpdating(dynakube.CodeModulesVersion(), dkMetadata.LatestVersion) {
		
func isUpdating(codeModulesVersion string, latestVersion string) bool {
  return dynakube.CodeModulesVersion() != dkMetadata.LatestVersion
}
```

src/controllers/csi/gc/controller.go
```diff
@@ -152,7 +176,29 @@ func getAllPinnedVersionsForTenantUUID(ctx context.Context, apiReader client.Rea
...
-		pinnedImages[dynakube.CodeModulesVersion()] = true
+		codeModuleVersion := dynakube.CodeModulesVersion()
+		if codeModuleVersion != "" {
+			pinnedVersions[codeModuleVersion] = true
+		}
+	}
+	return pinnedVersions, nil
+}
+
+func getAllDynakubes(ctx context.Context, apiReader client.Reader, namespace string) (*dynatracev1beta1.DynaKubeList, error) {
```
meik99, line 187: i might be wrong, but it feels like this function already exists.
maybe check if it does and refactor accordingly, or ignore this comment if it doesn't

meik99 (changes requested):

src/controllers/csi/gc/controller.go
mjgrzybek: Imho `ok` is better because it's standard way of checking it item exists in a map.

0sewa0: I moved this part into a separate (2 line) function and in that case `ok` is better IMO

src/controllers/csi/gc/controller.go
0sewa0: didn't find anything 😢

PR #991, opened 2022-07-20
In `csidriver` pod, split `driver` container to `server` and `provisioner`

mjgrzybek:
# Description

Currently, in `csidriver` pod, `driver` container contains logic of serving csi endpoint, as well as provisioning agent images.

The latter has spiky characteristics of resource consuming (cpu, mem) because it downloads agents from DT tenant, unpacks it and cleans it up if needed.

If `driver` container hits resource limits, it is killed and csi driver is broken for pods where it's mounted.

This PR splits `driver` container to `server` and `provisioner`.
`server` does lightweight work like answering grpc requests.
`provisioner` downloads images, unpacks them and executes `CsiGarbageCollection`.

Resources limits and requests for `server` set to: `100m`/`50Mi`.

## How can this be tested?
Set up dynakube with csidriver enabled.
Run custom pod to be injected.
Check if server is available if provisioner exceeded resources limits and is restarted.

## Checklist


src/cmd/csi-provisioner/builder.go
```diff
@@ -0,0 +1,147 @@
+package csi_server
```
meik99, line 1: imo file structure should be 
`src/cmd/csi/provisioner` and `src/cmd/csi/server` for the provisioner and server respectively.
The name of the package should reflact that, i.e. `package provisioner` and `package server`.
Having underscores in package names looks weird imo.

meik99 (changes requested):

config/helm/chart/default/templates/Common/csi/daemonset.yaml
```diff
@@ -87,11 +87,11 @@ spec:
...
           limits:
-            cpu: {{ default "300m" ((.Values.csi).limits).cpu }}
-            memory: {{ default "100Mi" ((.Values.csi).limits).memory }}
+            cpu: {{ default "100m" ((.Values.csi).limits).cpu }}
+            memory: {{ default "50Mi" ((.Values.csi).limits).memory }}
           requests:
-            cpu: {{ default "300m" ((.Values.csi).requests).cpu }}
-            memory: {{ default "100Mi" ((.Values.csi).requests).memory }}
+            cpu: {{ default "100m" ((.Values.csi).requests).cpu }}
+            memory: {{ default "50Mi" ((.Values.csi).requests).memory }}
```
0sewa0, line 94: we should either have a separate requests/limits section for these, or just static ones as the server part should have a stable memory/cpu usage on different environments

src/cmd/csi/provisioner/builder.go
```diff
@@ -0,0 +1,147 @@
...
+		if err != nil {
+			return err
+		}
+
+		err = createCsiDataPath(builder.getFilesystem())
+		if err != nil {
+			return err
+		}
+
+		// TODO: make the code below testable and test it, but in another ticket because otherwise adding the other commands will take a week
```
chrismuellner, line 122: this comment was copied over. wouldn't now be the perfect opportunity to do this and make it testable?

mjgrzybek: I think it should be addressed in separate story. 
Current story is not about refactoring, but feature implementation. Mixing things would lead to problems with e.g. reverting them if needed.

src/controllers/csi/config.go
```diff
@@ -37,8 +37,8 @@ const (
 
 var MetadataAccessPath = filepath.Join(DataPath, "csi.db")
 
-type CSIOptions struct {
-	NodeID   string
+type CsiOptions struct {
```
0sewa0, line 40: this rename make stuff a bit inconstant

we have `CSIDriverServer ` and `CSIGarbageCollector`

src/cmd/csi/provisioner/builder.go
0sewa0: this `TODO` is useless because its too generic, "make it testable" tells me nothing, the original writer of the comment must have understood it, but its only confusing at the moment.

also adding a `TODO` comment that says `in another ticket` is a big nono IMO, just create the said ticket, and be done with it.

src/controllers/csi/config.go
mjgrzybek: Reverted this change because k8s core uses same (wrong:slightly_smiling_face:) nanimg convention (e.g. `v1.CSIVolumeSource`)

mjgrzybek: Reverted this change because k8s core uses same (wrong:slightly_smiling_face:) nanimg convention (e.g. `v1.CSIVolumeSource`)

config/helm/chart/default/values.yaml
```diff
@@ -47,7 +47,7 @@ webhook:
     memory: 128Mi
   highAvailability: true
 
-csidriver:
+csi:
```
0sewa0, line 50: why change this?

it was recently fixed to be `csidriver` everywhere

0sewa0: it was fixed in: https://github.com/Dynatrace/dynatrace-operator/pull/1002

src/cmd/csi/provisioner/builder.go
0sewa0: pls remove useless TODO 😬

config/helm/chart/default/templates/Common/csi/daemonset.yaml
```diff
@@ -52,11 +52,11 @@ spec:
         # - Needs access to the csi socket, needs to read/write to it, needs root permissions to do so.
         # - Needs access to the filesystem of pods on the node, and mount stuff to it,needs to read/write to it, needs root permissions to do so
         # - Needs access to a dedicated folder on the node to persist data, needs to read/write to it.
-      - name: driver
+      - name: server
```
chrismuellner, line 55: annotations should be updated to replace `kubectl.kubernetes.io/default-logs-container: driver` with `kubectl.kubernetes.io/default-container: provisioner`

chrismuellner (changes requested):

chrismuellner (approved):

PR #1022, opened 2022-08-03
Add priorityClass to CSI Pods

luhi-DT:
# Description

Currently our CSI Pod cant be scheduled in certain scenarios, because of lacking resources on the cluster. However if the customer uses a deployment mode, which relies on the CSI Driver Pod to be there and handle volumes, it then leads to the situation that every pod gets the CSI Volume added by the webhook, but the CSI Pod can not handle those requests because he is simply not there. In consequence the application pod, which gets the CSI Volume cannot start and will fail.

Therefor I added a new priorityClass, which is set to a high priority by default, in order to make sure our CSI Pods can be scheduled

## How can this be tested?

- deploy our operator manifests
- check if the CSI Pod has set the correct priorityClass and correct priority

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/Common/csi/priority-class.yaml
+++ b/config/helm/chart/default/templates/Common/csi/priority-class.yaml
@@ -1,0 +1,24 @@
+{{- include "dynatrace-operator.platformRequired" . }}
+{{ if eq (include "dynatrace-operator.needCSI" .) "true" }}
+
+# Copyright 2021 Dynatrace LLC
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


config/helm/chart/default/templates/Common/csi/priority-class.yaml
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+kind: PriorityClass
+apiVersion: scheduling.k8s.io/v1
+metadata:
+  name: dynatrace-high-priority
+value: {{ int .Values.csidriver.priorityClassValue }}
+globalDefault: false
+description: "This priority class is used for Dynatrace Components with a higher priority"
```
chrismuellner, line 23: we could provide more reason why we need this, e.g. "... to make sure the pods are terminated in the correct order"

chrismuellner (changes requested):

config/helm/chart/default/templates/Common/csi/priority-class.yaml
luhi-DT: yes and not, what if customers want to reuse it on oneAgent pods? Or we reuse it in the future for operator/webhook pods aswell?

chrismuellner (approved):

PR #1031, opened 2022-08-05
Makes the csi extract archives into a tmp folder

0sewa0:
# Description
The csi-driver has problems determining if an agent zip was already SUCCESSFULLY downloaded if the Dynakube was deleted.

This causes that the csi-driver redownloads an already downloaded and mounted zip, which causes the already running agents to error, which leads to injected pods restarting.

So now we unzip the binaries into a `data/tmp_zip` and move it into place after unzip is complete.
Doing this we will be able to safely determine if a binary was downloaded/unziped successfully. 
(so the `isAlreadyDownloaded` checks are way simpler)

Because `os.Rename` is will not work if the destination already exists most `fs.MkdirAll(targetDir, common.MkDirFileMode)` was removed to avoid that. (also `afero`'s `MemFs` can't do `os.Rename` properly so some tests are funky)

## How can this be tested?
1. Deploy Dynakube with application monitoring or cloudNative.
1. Deploy sample apps that we inject into. Wait for them to startup
1. Delete Dynakube, wait a few seconds then deploy the Dynakube again.
    - This originally made the csi-driver redownload already used binaries, which caused the sample apps to restart (this shouldn't happen now)

## Checklist


mjgrzybek (changes requested):
Doesn't work for me (on kind and GKE):
```
{"level":"info","ts":"2022-08-12T07:43:15.343Z","logger":"oneagent-installer-zip","msg":"moving unpacked archive to target","targetDir":"/data/acc27517/bin/1.249.0.20220811-203908"}
{"level":"info","ts":"2022-08-12T07:43:15.343Z","logger":"oneagent-url-installer","msg":"failed to unzip OneAgent package","err":"rename /data/tmp_zip /data/acc27517/bin/1.249.0.20220811-203908: no such file or directory"}
{"level":"info","ts":"2022-08-12T07:43:15.346Z","logger":"oneagent-url-installer","msg":"failed to install agent","targetDir":"/data/acc27517/bin/1.249.0.20220811-203908"}
```

with `dynakube`:
```
kind: DynaKube
metadata:
  name: dk
  namespace: dynatrace
  annotations:
    feature.dynatrace.com/oneagent-privileged: "true"
spec:
  apiUrl: https://acc27517.dev.dynatracelabs.com/api
  oneAgent:
    cloudNativeFullStack:
      useCSIDriver: true
```

luhi-DT (changes requested):

luhi-DT (approved):

PR #1056, opened 2022-08-16
Adapt capabilities to use uppercase letters for ALL keyword

### LINKED ISSUE #1035: [BF] Fix the case of the securityContext
The current Helm Chart deploy the webhook and operator `securityContext` for [`drop capabilities`](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-capabilities-for-a-container) with this syntax:

```yaml
capabilities:
  drop: ["all"]
```
Ref:
- https://github.com/Dynatrace/dynatrace-operator/blob/master/config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml#L134
- https://github.com/Dynatrace/dynatrace-operator/blob/master/config/helm/chart/default/templates/Common/operator/deployment-operator.yaml#L105

This syntax is not recognized correctly by [Kyverno](https://kyverno.io/policies/best-practices/require_drop_all/require_drop_all/).

To solve this, the syntax and especially the case should be changed in this way.
```yaml
    capabilities:
      drop:
        - ALL # Uppercase
```

luhi-DT:
# Description

Please include the following:
Updates the securityContext to drop ALL capabilities in uppercase instead of lowercase which fixes #1035

## How can this be tested?
Should work as before, so applying our manifests should also propagate the `DROP: [ALL]` securityContext to webhook etc.

## Checklist


config/helm/chart/default/tests/Common/operator/deployment-operator_test.yaml
```diff
@@ -117,7 +117,7 @@ tests:
                   runAsUser: 1001
                   runAsGroup: 1001
                   capabilities:
-                    drop: ["all"]
+                    drop: ["ALL"]
```
chrismuellner, line 120:
```suggestion
                    drop: 
                      - ALL
```
we should be consistent and have it as an actual list everywhere

chrismuellner (changes requested):

config/helm/chart/default/tests/Common/operator/deployment-operator_test.yaml
luhi-DT: done

meik99 (approved):

chrismuellner (approved):

luhi-DT:
@pierluigilenoci will be released with the next release (v0.8.1)

PR #787, opened 2022-05-16
Support for IBM Power and fix proxy bug affecting cloud-native app-only monitoring

jsm84:
# Description

EDIT: Closed in favor of #865 and #866 

This PR brings application-only monitoring support to OpenShift on IBM Power and Z.

Z-support is affected by a bug where `ruxitagentproc.conf` is missing from the oneagent install.zip for the s390 platform.

The motivation behind this is of joint concern to Dynatrace, IBM and Red Hat, to target a growing customer audience that is requesting Dynatrace Operator on the IBM Power and Z platforms.

### New Build Scripts Provided
To get the operator code to build (with CGO enabled) for Power and Z, it was required to build locally against a RHEL 8 host. I would prefer a container/UBI-8 based build solution, but that was not possible due to the lack of a UBI-based go-toolset v1.17+ when I started this back in January. There is a new RHEL-specific container build script (a modified version of `./build/deploy_local.sh`) provided for this, as well as a script for building the bundle and index images, as well as the catalog source yaml file for deploying the index onto an OpenShift cluster, for installation using OperatorHub which is required for Red Hat certification of the operator.

### Requirements to Build and Test
To test this yourself, you'd need the following:
* A RHEL 8 VM running natively on Power or Z in order to build the operator (native container-based builds are now possible but not provided with this PR) with the following installed
   * Go 1.18 installed from go.dev
   * The following additional RHEL packages: `git gcc gcc-c++ make device-mapper-devel golang podman podman-docker gpgme-devel libassuan-devel`
   * Operator SDK v1.16.0
   * Operator Registry (`opm`) v1.19.5
   * The following go binaries:
      * sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0
      * sigs.k8s.io/kustomize/kustomize/v3
   * Helm v3
* Power or Z-based OpenShift cluster, to be able to test app-only monitoring on the respective platform

### Steps to Build:

With the above installed and running on the local RHEL 8 VM, the `deploy_rhel.sh` script is called as follows:

```
$ GOARCH=<s390x|ppc64le> LOCALBUILD=1 \
 IMG=quay.io/orgname/dynatrace-operator TAG=v0.6.0-<arch> \
./hack/build/deploy_rhel.sh
```
You should now have the operator image built and pushed to the quay repository specified, with the tag that was specified.

Next, prepare the bundle, index and catalog source for deployment onto OpenShift with the `bundle_deploy_ocp.sh` script:

```
[...]

PR #1071, opened 2022-08-19
Fixes and improvements to deploy_rhel.sh

### LINKED ISSUE #1051: RHEL deploy script no longer builds a working binary
Regardless of architecture (what `GOARCH` is set to), the `./hack/build/deploy_rhel.sh` script no longer builds a working binary.

This seems to be due to a relocation of the main pkg in the source tree, causing a partial go build to occur where the main pkg is never built, and thus a non-executable build artifact gets created in place of the usual binary in `./build/_output/bin/`.

The deploy_rhel.sh script just needs to be updated to reflect where the `main.go` file currently exists, and the script could use some feature enhancements as well, such as catching errors from `go build`, `go-licences save`, and `docker build` commands used in the script.

jsm84:
# Description

This PR fixes #1051 and incorporates the following changes:
* Fixed to track relocated main.go file
* Quay expiry is now optional (defaults to never)
* Added `-h` option for help
* Much improved error checking

## How can this be tested?

On a RHEL 8 or 9 host meeting the installation requirements (see #787 for the full list):
```
$ CC=/usr/bin/gcc LOCALBUILD=1 IMG=quay.io/<your-org>/dynatrace-operator TAG=v0.8.0-test ./hack/build/deploy_rhel.sh
```

## Checklist

### DIFF
--- a/hack/build/deploy_rhel.sh
+++ b/hack/build/deploy_rhel.sh
@@ -2,9 +2,59 @@
 
 if [[ ! "${TAG}" ]]; then
   echo "TAG variable not set"
-  echo "usage: 'make deploy-local TAG=\"<your-image-tag>\"' or 'make deploy-local-easy'"
+  echo "Usage: 'make deploy-local TAG=\"<your-image-tag>\"' or 'make deploy-local-easy'"
+  echo "See '-h' option for help"
   exit 5
 fi
+
+help_opts='^-h$'
+for o in "$@"; do
+	if [[ "$o" =~ $help_opts ]]; then
+		echo "Supported environment variables:"
+		echo "  IMG          Operator image pull spec override"
+		echo "  TAG          Operator image tag (required)"
+		echo "  LOCALBUILD   Do not build in a container (required for multi-arch)"
+		echo "  GOARCH       Desired CPU architecture (amd64|arm64|ppcle64|s390x)"
+		echo "  QUAY_EXPIRY  Days before deleting image from quay (defaults to never)"
+		exit 1
+	fi
+done
+
+function print_err() {
+	ERR_CMD=${ERR_CMD:-"UNKNOWN"}
+	if [[ "${ERR_CMD}" != "UNKNOWN" ]]; then
+		echo "ERROR: \"${ERR_CMD}\" exited abnormally. Aborting..."
+	else
+		echo "ERROR: An unknown error occurred. Aborting..."
+	fi
+}
+
+function check_err() {
+  if [[ "$?" != 0 ]]; then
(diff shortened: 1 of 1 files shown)


jsm84:
The CI failed a docker pull due to invalid username/password. Will rebase for another attempt.

chrismuellner:
@jsm84 repository secrets are not available on forks sadly. As long as only the push fails the CI is fine.

hack/build/deploy_rhel.sh
```diff
@@ -16,29 +66,32 @@ go_build_args=(
...
-
+  go build -ldflags "${args}" -tags exclude_graphdriver_btrfs -o ./build/_output/bin/dynatrace-operator ./src/cmd/
+  check_err GO_BUILD
+  
   go get github.com/google/go-licenses
   go-licenses save ./... --save_path third_party_licenses --force
+  check_err GO_LICENSES
 
-  docker build . -f ./Dockerfile-localbuild -t "${base_image}" --label "quay.expires-after=14d" --no-cache
+  docker build . -f ./Dockerfile-localbuild -t "${base_image}" ${expiry_args} --no-cache
```
chrismuellner, line 87: do you use the local build? the `Dockerfile-localbuild` is pretty outdated and I was planning on removing it.

hack/build/deploy_rhel.sh
```diff
@@ -2,10 +2,60 @@
```
chrismuellner, line 2: could we move this to a makefile step? as it is it contains various part of existing scripts, which could be better handled as make prerequisites.
That would also make it easier for us to keep it updated and break it less 😄

hack/build/deploy_rhel.sh
jsm84: TL;DR: do with Dockerfile-localbuild as you like.

Porting to power/z architectures is on hiatus, so I've sort of stopped working on it. I was using the localbuild Dockerfile, with intent to either fork or modify the existing Dockerfile to support s390x/ppc64le (with the build stage on UBI, since dynamic linking is required). 

The development VMs I was using were configured more as pets than cattle, per se, and the golang packages were too old to use ubi8/go-toolset for a UBI-based build.

hack/build/deploy_rhel.sh
jsm84: See my above comments. If things continue on porting DTO to power/Z then I have no problem moving this into the Makefile. I fear the whole thing has reached a dead end.

jsm84:
Closing this. Porting the operator to the power and z architecture is no longer in flight.

PR #1074, opened 2022-08-22
Set empty workload kind/name on unknown owner

chrismuellner:
# Description
In case of owner that is not well-known, the `kind` would be set to `Pod`, while the `name` couldn't be set because `.metadata.name` is empty at this point. The empty kind/name is correct in this case and should not block the init container.

## How can this be tested?
Inject into a pod that has owner reference set to something of unknown kind, i.e. not `Deployment`, `ReplicaSet`, ..., but a separate API.

## Checklist

### DIFF
--- a/src/config/enrichment_injection.go
+++ b/src/config/enrichment_injection.go
@@ -6,6 +6,7 @@ const (
 	EnrichmentInjectedEnv        = "DATA_INGEST_INJECTED"
 	EnrichmentWorkloadKindEnv    = "DT_WORKLOAD_KIND"
 	EnrichmentWorkloadNameEnv    = "DT_WORKLOAD_NAME"
+	UnknownWorkload              = "UNKNOWN"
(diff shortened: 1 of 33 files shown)


src/config/enrichment_injection.go
```diff
@@ -6,6 +6,7 @@ const (
 	EnrichmentInjectedEnv        = "DATA_INGEST_INJECTED"
 	EnrichmentWorkloadKindEnv    = "DT_WORKLOAD_KIND"
 	EnrichmentWorkloadNameEnv    = "DT_WORKLOAD_NAME"
+	UnknownWorkload              = "UNKNOWN"
```
0sewa0, line 9:
```suggestion
	EnrichmentUnknownWorkload              = "UNKNOWN"
```

0sewa0 (approved):

PR #1075, opened 2022-08-22
Adds columns for volume mount attempts

meik99:
# Description

Adds an alter statement for the volume and dynakube tables such that they have a column for mount attempts and maximal failed attempts respectively.

~Also updates the Docker base image to use ubi9 and adds `tar` as a dependency so that `kuebctl cp` can be used.~

### Case 1: new deployment

* Create a cluster and deploy the operator from this branch
* deploy cloudnative
* deploy a sample app
* make sure injection works
* copy the `/data/csi.db` to your local machine and inspect it with any sqlite editor
* make sure the new columns exist

### Case 2: existing deployment

* Create a cluster and deploy an older operator version, e.g. snapshot
* deploy cloudnative
* deploy a sample app
* make sure injection works
* use `cat /data/csi.db`  to inspect the database, it should contain create table statements without the columns since it is an old version
* deploy a new version of the operator
* redeploy sample apps
* make sure injection works
* copy the `/data/csi.db` to your local machine and inspect it with any sqlite editor
* make sure the new columns exist

## Checklist
- [x] I have read and understood the [contribution guidelines](CONTRIBUTING.md)


0sewa0 (dismissed):

src/controllers/csi/metadata/sqlite.go
```diff
@@ -188,33 +197,65 @@ func (access *SqliteAccess) connect(driver, path string) error {
...
+
+	return nil
+}
+
+func (access *SqliteAccess) executeAlterStatement(statement string) error {
+	if _, err := access.conn.Exec(statement); err != nil {
 		sqliteError := err.(sqlite3.Error)
 		if sqliteError.Code != sqlite3.ErrError {
-			return errors.WithStack(errors.WithMessage(err, "couldn't add ingest column"))
+			return errors.WithMessage(err, "could not add ingest column")
```
chrismuellner, line 253: error too specific

chrismuellner (changes requested):

Dockerfile
```diff
@@ -37,9 +53,8 @@ COPY --from=operator-build /usr/lib/libgpgme.so.* /usr/lib/
 COPY --from=k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.5.1 /csi-node-driver-registrar /usr/local/bin
 COPY --from=k8s.gcr.io/sig-storage/livenessprobe:v2.7.0 /livenessprobe /usr/local/bin
 
-# csi depdenencies
-COPY --from=dependency-src /bin/mount /bin/umount /bin/
-COPY --from=dependency-src /lib64/libmount.so.1 /lib64/libblkid.so.1 /lib64/libuuid.so.1 /lib64/
+# depdenencies
```
chrismuellner, line 56: we should keep "csi dependencies" and add another section with "general/core/debug dependencies"

Dockerfile
```diff
@@ -37,9 +53,8 @@ COPY --from=operator-build /usr/lib/libgpgme.so.* /usr/lib/
...
+COPY --from=dependency-src /mnt/installfs/ /
```
chrismuellner, line 57: can we try to copy files explicitly same as before instead of just copying the whole directory?

chrismuellner (changes requested):

Dockerfile
meik99: I'll answer this question as well as the one below:

By specifying the install root in the installation command, the files that are needed are installed in the specified directory.
This means, by default, only the needed binaries and its strong dependencies are in this folder.
There is no difference between copying all contents of this folder at once or copying the specific files one by one, but the additional leg work necessary to list all the files explicitly.
Therefore, I would opt to not list them explicitly which would also mean that there can't be two different sections.

The only thing I actually need to delete is the `os-release` file because that is copied in manually for `microdnf` to work. I'll fix that in the next commit.

src/controllers/csi/metadata/sqlite.go
meik99: done

Dockerfile
meik99: see above

chrismuellner (approved):

PR #1076, opened 2022-08-23
Adds csidriver.kubeletPath to values.yaml

0sewa0:
# Description
Adds `csidriver.kubeletPath` to values.yaml. 
In some kubernetes variants the path to kubelet is different so the csidriver needs to be configured to work

## How can this be tested?
Run `helm template ...` when `csidriver.kubeletPath` is set  to something and check the output

## Checklist


config/helm/chart/default/templates/Common/csi/daemonset.yaml
```diff
@@ -252,25 +255,26 @@ spec:
...
-          path: /var/lib/kubelet/pods
+      - name: mountpoint-dir
+        hostPath:
+          path: {{ include "dynatrace-operator.CSIMountPointDir" . }}
           type: DirectoryOrCreate
-        name: mountpoint-dir
         # This volume is where the driver persists data on the node
-      - hostPath:
-          path: /var/lib/kubelet/plugins/csi.oneagent.dynatrace.com/data
+      - name: dynatrace-oneagent-data-dir
```
chrismuellner, line 273: we could combine this volume with the `plugin-dir` volume and mount the subfolder via `mountPath`

chrismuellner (changes requested):

PR #1089, opened 2022-08-25
Adds feature flag to configure `MaxFailedMountAttempts`

meik99:
# Description

Adds a feature flag to configure the value for the `MaxFailedMountAttempts` column and modifies the provisioner so that it is correctly persisted.

## How can this be tested?

* Deploy the operator and cloudnative dynakube
* Copy the `/data/csi.db` from a csi driver pod
* Make sure the default value is used in the dynakube entry of the dynakubes table
* Add the `feature.dynatrace.com/max-csi-mount-attempts` annotation and set it to any value
* If you set it to an invalid value, e.g. a letter or a number < 0, "0" is persisted
* If no value is set, the default value is persisted
* Otherwise the specified value should is persisted

## Checklist


src/controllers/csi/metadata/sqlite.go
```diff
@@ -308,13 +309,18 @@ func (access *SqliteAccess) GetDynakube(dynakubeName string) (*Dynakube, error)
...
+	var maxFailedMountAttempts *int
 
 	err := access.querySimpleStatement(getDynakubeStatement, dynakubeName, &tenantUUID, &latestVersion, &imageDigest, &maxFailedMountAttempts)
 	if err != nil {
 		err = errors.WithMessagef(err, "couldn't get dynakube, name '%s'", dynakubeName)
 	}
-	return NewDynakube(dynakubeName, tenantUUID, latestVersion, imageDigest, maxFailedMountAttempts), err
+
+	if maxFailedMountAttempts == nil {
+		maxFailedMountAttempts = address.Of(-1)
```
0sewa0, line 320: shouldn't this be `defaultSqlMaxFailedMountAttempts`?
also if you have `ADD COLUMN MaxFailedMountAttempts INT DEFAULT ' + defaultSqlMaxFailedMountAttempts + ';` wouldn't it always return that default value when reading from the database ?

0sewa0: also, setting the `DEFAULT` in the database feels scary a bit.

what happens if the default changes between releases or something, will the `ALTER` statement be able to handle that ?

src/api/v1beta1/properties.go
```diff
@@ -400,3 +400,11 @@ func getRawImageTag(imageURI string) string {
 	splitURI := strings.Split(imageURI, ":")
 	return splitURI[len(splitURI)-1]
 }
+
+func (dk *DynaKube) MaxFailedCsiMountAttempts() int {
+	if dk.FeatureMaxFailedCsiMountAttempts() != nil {
+		return *dk.FeatureMaxFailedCsiMountAttempts()
+	}
+
+	return -1
```
0sewa0, line 409: wouldn't it be cleaner if we returned the default value here? (and not handled it in the `metadata.NewDynakube()`)

src/controllers/csi/metadata/sqlite.go
meik99: The ticket mentioned the column to be nullable.
The default value is only used if the value is not specified on an insert, if null is explicitly inserted the value will be null
If the value is null, it has been explicitly inserted, returning the default value in this case would make the column being nullable useless. I assume the intention is that if null, or -1, is returned, that there is effectively no maximum retry value and it should retry endlessly.
If the default is changed the alter command will handle that accordingly

meik99: after talking about it, it turned out that you were completely correct, I will make the changes accordingly

meik99: done

src/api/v1beta1/properties.go
meik99: done

src/api/v1beta1/feature_flags.go
```diff
@@ -276,3 +282,19 @@ func (dk *DynaKube) getFeatureFlagRaw(annotation string) string {
...
+	maxCsiMountAttemptsValue := dk.getFeatureFlagRaw(AnnotationFeatureMaxFailedCsiMountAttempts)
+
+	if maxCsiMountAttemptsValue == "" {
+		return defaultMaxFailedCsiMountAttempts
+	}
+
+	maxCsiMountAttempts, err := strconv.Atoi(maxCsiMountAttemptsValue)
+
+	if err != nil || maxCsiMountAttempts < 0 {
+		return 0
```
0sewa0, line 296: shouldn't this be also `defaultMaxFailedCsiMountAttempts`?

src/controllers/csi/metadata/metadata_test.go
```diff
@@ -42,7 +42,7 @@ func TestNewDynakube(t *testing.T) {
 		assert.Equal(t, testUUID, dynakube.TenantUUID)
 		assert.Equal(t, testVersion, dynakube.LatestVersion)
 		assert.Equal(t, testDigest, dynakube.ImageDigest)
-		assert.Equal(t, defaultMaxFailedMountAttempts, dynakube.MaxFailedMountAttempts)
+		assert.Equal(t, -1, dynakube.MaxFailedMountAttempts)
```
0sewa0, line 45: this should never happen

src/controllers/csi/metadata/sqlite.go
```diff
@@ -49,7 +50,7 @@ const (
 
 	dynakubesAlterStatementMaxFailedMountAttempts = `
 	ALTER TABLE dynakubes
-	ADD COLUMN MaxFailedMountAttempts INT DEFAULT ` + defaultSqlMaxFailedMountAttempts + `;`
+	ADD COLUMN MaxFailedMountAttempts INT NOT NULL DEFAULT 0;`  // "Not null"-columns need a default value set
```
0sewa0, line 53: spooky to have it at 0, but that should never happen

src/controllers/csi/metadata/sqlite.go
0sewa0, line 320: `maxFailedMountAttempts` should never be `nil` at this point
-1 is not a valid value

0sewa0 (changes requested):

src/api/v1beta1/feature_flags.go
meik99: done

src/controllers/csi/metadata/metadata_test.go
meik99: removed the test

src/controllers/csi/metadata/sqlite.go
meik99: should i set it to a different value?

src/controllers/csi/metadata/sqlite.go
meik99: changed from pointer to value, good catch

src/controllers/csi/metadata/sqlite.go
meik99: set it to 3

0sewa0 (dismissed):

0sewa0 (approved):

PR #1092, opened 2022-08-25
Adds validation that volume storage env setting isn't used in read-only mode

StefanHauth:
# Description
When the OneAgent is deployed with the `ONEAGENT_ENABLE_VOLUME_STORAGE` environment variable set to `true` while being used on a system with a read-only host filesystem, it fails its initialization procedure. Added a validation that this combination is rejected for `cloudNativeFullstack` and hostMonitoring setups. Also added a warning if the feature flag `oneagent-readonly-host-fs` is used in `classicFullstack` where it doesn't have any effects.

## How can this be tested?
- create a `cloudNativeFullstack` configuration with FF `oneagent-readonly-host-fs = true` and `ONEAGENT_ENABLE_VOLUME_STORAGE` env variable added in `oneAgent` spec.
- when applying the config, an error shall be displayed, that this combination is not accepted.

## Checklist

### DIFF
--- a/src/webhook/validation/oneagent.go
+++ b/src/webhook/validation/oneagent.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
 	"k8s.io/apimachinery/pkg/labels"
 )
 
@@ -16,6 +17,9 @@ const (
 	errorNodeSelectorConflict = `The DynaKube's specification tries to specify a nodeSelector conflicts with an another Dynakube's nodeSelector, which is not supported.
 The conflicting Dynakube: %s
 `
+	errorVolumeStorageReadOnlyModeConflict = `The DynaKube's specification specifies a read-only host file system and OneAgent has volume storage enabled.`
+
+	warningIneffectiveFeatureFlag = `Feature flag %s has no effect in classic full stack mode.`
 )
 
 func conflictingOneAgentConfiguration(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
@@ -89,4 +93,35 @@ func hasConflictingMatchLabels(labelMap, otherLabelMap map[string]string) bool {
 	labelSelectorLabels := labels.Set(labelMap)
 	otherLabelSelectorLabels := labels.Set(otherLabelMap)
 	return labelSelector.Matches(otherLabelSelectorLabels) || otherLabelSelector.Matches(labelSelectorLabels)
+}
+
+func hasOneAgentVolumeStorageEnabled(dynakube *dynatracev1beta1.DynaKube) (isEnabled bool, isSet bool) {
+	envVar := kubeobjects.FindEnvVar(dynakube.GetOneAgentEnvironment(), oneagentEnableVolumeStorageEnvVarName)
+	isSet = envVar != nil
+	isEnabled = isSet && envVar.Value == "true"
+	return
+}
+
+func conflictingOneAgentVolumeStorageSettings(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
+	volumeStorageEnabled, volumeStorageSet := hasOneAgentVolumeStorageEnabled(dynakube)
+	if dynakube.NeedsReadOnlyOneAgents() && volumeStorageSet && !volumeStorageEnabled {
+		return errorVolumeStorageReadOnlyModeConflict
+	}
+	return ""
+}
+
+func ineffectiveReadOnlyHostFsFeatureFlag(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
+	if dynakube.ClassicFullStackMode() {
+		if _, hasOneAgentReadOnlyFeatureFlag := dynakube.Annotations[dynatracev1beta1.AnnotationFeatureReadOnlyOneAgent]; hasOneAgentReadOnlyFeatureFlag {
+			return readonlyHostFsFlagWarning(dynatracev1beta1.AnnotationFeatureReadOnlyOneAgent)
+		}
+		if _, hasOneAgentReadOnlyFeatureFlag := dynakube.Annotations[dynatracev1beta1.AnnotationFeatureDisableReadOnlyOneAgent]; hasOneAgentReadOnlyFeatureFlag {
+			return readonlyHostFsFlagWarning(dynatracev1beta1.AnnotationFeatureDisableReadOnlyOneAgent)
+		}
+	}
+	return ""
+}
+
+func readonlyHostFsFlagWarning(featureFlag string) string {
+	return fmt.Sprintf(warningIneffectiveFeatureFlag, featureFlag)
 }
--- a/src/webhook/validation/oneagent_test.go
+++ b/src/webhook/validation/oneagent_test.go
@@ -2,9 +2,11 @@ package validation
 
 import (
 	"fmt"
+	"strconv"
 	"testing"
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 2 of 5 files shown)


src/webhook/validation/oneagent_test.go
```diff
@@ -290,3 +295,141 @@ func TestImageFieldSetWithoutCSIFlag(t *testing.T) {
 	})
 
 }
+
+func TestOneAgentVolumeStorageReadOnlyModeConflict(t *testing.T) {
```
meik99, line 299: please rework these tests so they align with the sample in the contribution guidelines

meik99 (changes requested):

src/webhook/validation/oneagent_test.go
StefanHauth: That just leads to lots of code duplication, long test methods, obfuscation of the tiny differences in the tests and doesn't provide any benefit to code quality whatsoever. One thing to rework is the test naming, Testnames should be written in the code (not generated as of now) to make them easier to find when failing.

StefanHauth: Also test maintenance get's potentially more cumbersome with lot's of duplicated code.

src/webhook/validation/oneagent.go
```diff
@@ -90,3 +93,46 @@ func hasConflictingMatchLabels(labelMap, otherLabelMap map[string]string) bool {
...
+
+func getOneAgentVolumeStorageEnvVar(env []corev1.EnvVar) bool {
+	envVar := kubeobjects.FindEnvVar(env, "ONEAGENT_ENABLE_VOLUME_STORAGE")
+	if envVar != nil {
+		return envVar.Value == "true"
+	}
+	return false
+}
+
+func hasOneAgentVolumeStorageEnabled(dynakube *dynatracev1beta1.DynaKube) bool {
```
chrismuellner, line 105: the `env` access could be moved to `properties.go` and tested there. That could also simplify test and only require the `HostInjectSpec` to be tested 🙂

src/webhook/validation/oneagent.go
```diff
@@ -90,3 +93,46 @@ func hasConflictingMatchLabels(labelMap, otherLabelMap map[string]string) bool {
...
+	case dynakube.ClassicFullStackMode():
+		return getOneAgentVolumeStorageEnvVar(dynakube.Spec.OneAgent.ClassicFullStack.Env)
+	case dynakube.HostMonitoringMode():
+		return getOneAgentVolumeStorageEnvVar(dynakube.Spec.OneAgent.HostMonitoring.Env)
+	}
+	return false
+}
+
+func conflictingOneAgentVolumeStorageSettings(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
+	if hasOneAgentVolumeStorageEnabled(dynakube) && !dynakube.FeatureDisableReadOnlyOneAgent() {
```
chrismuellner, line 118: the case I've seen the combination of env variable and read only mode fail before was only when readonly mode was enabled and the `ONEAGENT_ENABLE_VOLUME_STORAGE` env variable was set to `false`. The oneagent needs to be run with volume storage enabled in case of volumes being provided by the CSI driver, it just doesn't create the volume itself.

Not enabling the volume storage with or without the CSI driver seems like a valid option.

chrismuellner (changes requested):

src/webhook/validation/oneagent_test.go
```diff
@@ -290,3 +292,162 @@ func TestImageFieldSetWithoutCSIFlag(t *testing.T) {
...
+
+	for _, tc := range testcases {
+		t.Run(tc.testName, func(t *testing.T) {
+			assertDeniedResponse(t,
+				[]string{errorVolumeStorageReadOnlyModeConflict},
+				tc.creator(tc.featureFlag, tc.featureValue, true),
+				&defaultCSIDaemonSet)
+		})
+	}
+}
```
meik99, line 322:
```suggestion

func TestOneAgentVolumeStorageReadOnlyModeConflict(t *testing.T) {
	t.Run("enabling read-only mode with cloud-native is denied", func(t *testing.T) {
		assertDeniedResponse(t,
			[]string{errorVolumeStorageReadOnlyModeConflict},
			createCloudNativeFullstackDynaKube(dynatracev1beta1.AnnotationFeatureReadOnlyOneAgent, true, true),
			&defaultCSIDaemonSet)
	})
	
	t.Run("enabling read-only mode with deprecated flag and cloud-native is denied", func(t *testing.T) {
		assertDeniedResponse(t,
			[]string{errorVolumeStorageReadOnlyModeConflict},
			createCloudNativeFullstackDynaKube(dynatracev1beta1.AnnotationFeatureReadOnlyOneAgent, false, true),
			&defaultCSIDaemonSet)
	})
	
	t.Run("enabling read-only mode with host-monitoring is denied", func(t *testing.T) {
		assertDeniedResponse(t,
			[]string{errorVolumeStorageReadOnlyModeConflict},
			createHostMonitoringDynaKube(dynatracev1beta1.AnnotationFeatureReadOnlyOneAgent, true, true),
			&defaultCSIDaemonSet)
	})
	
	t.Run("enabling read-only mode with deprecated flag and host-monitoring is denied", func(t *testing.T) {
		assertDeniedResponse(t,
			[]string{errorVolumeStorageReadOnlyModeConflict},
			createHostMonitoringDynaKube(dynatracev1beta1.AnnotationFeatureDisableReadOnlyOneAgent, false, true),
			&defaultCSIDaemonSet)
	})
}
```

StefanHauth: How does that improve the code? Its just needless code duplication.

meik99: its more readable and there is barely any code duplication.
looping over test functions makes the code needlessly complex

StefanHauth: No it's not. The test cases are collected in a single table all at one place, where you can easily spot the the difference with one glimpse instead of having to compare 4 almost similar code snippets. The test execution logic is there once instead of 4 times, which means it's more maintainable. And the "complexity" of a simple loop is tolerable.

meik99: To understand a test-case, in your case one has to remember the method itself, e.g. `createHostMonitoringDynaKube`, the struct and what is inside `tc.creator`, then think about how `tc.creator` is called. After these steps it can be understood how the code works.

In the other case, how execution is performed is immediately obvious to a programmer.

meik99: If that does not convince you, then think about it this way: The way we write tests is standardized in the contribution guidelines. These guidelines may not always align with personal preferences, like in this case: is readability better then less code duplication? I would argue that in this case the answer is subjective. However, the guidelines exist so that the code base has a unified style. That would be another reason to implement the tests in the suggested way, because it aligns better with the majority of the other unit tests

StefanHauth: You're right, I'm not convinced ;-)

First of all, throwing the guidelines into the ring seems like a killer argument to me, so I'm going to ignore that for now. 

Second, to understand the test cases (what's tested) you merely have to have a look at the test table. How the test is performed, isn't actually that important at first. The table summarizes the covered test cases in a kind of abstracted way. Only if you want to go a bit deeper you can have a look at the execution logic. So you can approach the test code in a two-step manner - abstract to detail.
What aids the readability is, that the whole test method fits on a screen and it probably still would with double the number of tests. But I may rethink the naming of the "creator" member, though. It's probably to generic.

Generally speaking, table tests tend to need a bit of getting-used to in the first place (was my experience too), but they can really clean up test code and avoid lot's of tedious maintenance work when the tested code changes. Had this quite often in the past, that test code grew and grew (which mostly is a good thing as it means more tests) and once you made a tiny change to a used struct, a signature etc. you had to adopt tests in dozens of locations. So if guidelines don't allow them, it's maybe time to challenge the guidelines.

meik99: i see your points. In that case I'd like to see a well defined struct, maybe not anonymous but local to the test function with good member names. Maybe after this PR is done I'd suggest you create a PR to amend the guidelines so table tests are also standardized

src/webhook/validation/oneagent.go
StefanHauth: After gaining a deeper understanding of the dependencies of the flags, I've reworked the code and added more unit tests to cover various combinations.

src/webhook/validation/oneagent.go
StefanHauth: Not sure if I understood that correctly, but the tests in `validation` test the validation routines not the `HostInejctSpec`. And those routines work based on the feature-flag and the env var as an input. Thus, testing the validation routines always requires a full dynakube as input. So how would moving the env var access to properties simplify tests here?

src/webhook/validation/oneagent_test.go
StefanHauth: Baptized structs and added test cases.

0sewa0:
you have changes in packages under `src/controllers/csi`, your changes should touch these
there must have been a mix up in git. 😅

src/webhook/validation/oneagent_test.go
```diff
@@ -290,3 +292,242 @@ func TestImageFieldSetWithoutCSIFlag(t *testing.T) {
...
+				tc.acceptedWarningCount,
+				tc.dynakubeFactory(tc.featureFlag, tc.featureValue, tc.oaVolumeStorageVar, tc.oaVolumeStorageValue),
+				&defaultCSIDaemonSet)
+		})
+	}
+}
+
+func TestReadOnlyModeClassicFullstackWarning(t *testing.T) {
+	testcases := []validOneAgentVolumeStorageReadOnlyModeSettingsTestcase{
+		{"OneAgent volume storage disabled by env var and OneAgent read-only mode enabled by feature flag in classicFullstack",
```
chrismuellner, line 412: the env variable is irrelevant for classic fullstack. we provide a warning when the feature flag is used, no matter if the env variable is set or not.

src/webhook/validation/oneagent.go
```diff
@@ -90,3 +95,47 @@ func hasConflictingMatchLabels(labelMap, otherLabelMap map[string]string) bool {
 	otherLabelSelectorLabels := labels.Set(otherLabelMap)
 	return labelSelector.Matches(otherLabelSelectorLabels) || otherLabelSelector.Matches(labelSelectorLabels)
 }
+
+func getOneAgentVolumeStorageEnvVar(env []corev1.EnvVar) (bool, bool) {
```
0sewa0, line 99: why the 2 bools ???

we only care if its explicitly disabled (AFAIK)

src/webhook/validation/oneagent.go
```diff
@@ -90,3 +95,47 @@ func hasConflictingMatchLabels(labelMap, otherLabelMap map[string]string) bool {
...
+
+func conflictingOneAgentVolumeStorageSettings(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
+	volumeStorageEnabled, volumeStorageVarSet := hasOneAgentVolumeStorageEnabled(dynakube)
+	if dynakube.NeedsReadOnlyOneAgents() && volumeStorageVarSet && !volumeStorageEnabled {
+		return errorVolumeStorageReadOnlyModeConflict
+	}
+	return ""
+}
+
+func superfluousReadOnlyHostFsFeatureFlag(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
```
0sewa0, line 127:
```suggestion
func unnecessaryReadOnlyHostFsFeatureFlag(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
```

src/webhook/validation/oneagent.go
```diff
@@ -16,6 +18,9 @@ const (
 	errorNodeSelectorConflict = `The DynaKube's specification tries to specify a nodeSelector conflicts with an another Dynakube's nodeSelector, which is not supported.
 The conflicting Dynakube: %s
 `
+	errorVolumeStorageReadOnlyModeConflict = `The DynaKube's specification specifies a read-only host file system and OneAgent has volume storage enabled.`
+
+	warningSuperfluousFeatureFlag = `Feature flag %s has no effect in classic full stack mode.`
```
0sewa0, line 23:
```suggestion
	warningUnnecessaryFeatureFlag = `Feature flag %s has no effect in classic full stack mode.`
```

src/webhook/validation/oneagent_test.go
```diff
@@ -290,3 +292,242 @@ func TestImageFieldSetWithoutCSIFlag(t *testing.T) {
...
+			createCloudNativeFullstackDynaKube,
+			"", true,
+			"", true,
+			0},
+		{"OneAgent volume storage default with OneAgent read only mode enabled by default in hostMonitoring",
+			createHostMonitoringDynaKube,
+			"", true,
+			"", true,
+			0},
+	}
```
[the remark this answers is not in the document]
0sewa0: About the actual tests 😛 :
```
		oaVolumeStorageVar: "",
		oaVolumeStorageValue: true,
```
wouldn't just `oaVolumeStorageValue` be enough as we only care if its explicitly disabled

0sewa0 (approved):

src/webhook/validation/oneagent.go
StefanHauth: Agreed, naming is not optimal, but I will even go with "ineffective", that's even more to the point.

src/webhook/validation/oneagent.go
StefanHauth: For the warning in classicFullstack mode it's necessary to know if the flag was set. The value is rather irrelevant in this use-case.

src/webhook/validation/oneagent_test.go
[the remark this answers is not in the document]
StefanHauth: It's to cover all bases in the unit tests. These tests make sure that we do not report errors when everything is ok. From a testing perspective, having the variable not set is not the same thing as having the variable set to it's default value.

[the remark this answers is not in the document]
StefanHauth: > * I don't think this format can be used in every situation.
I never said, that it's for all situations, but IMHO it's the best for this situation
>   
>   * Unless you create rather convoluted factories and the like which IMO are not that easy to change (thankfully this is not the case here)
> * This is not what is in the guidelines
True, but a killer argument for this discussion.

>   
>   * The guidelines are there to make the code consistent. If we don't follow/enforce them then why do they even exist ?
In this case the guidelines rob us of a useful tool in the toolbox, so it's worth challenging them. Why would want to make code hard to read just for the sake of consistency?

>   * We had discussions in the team about testing structure before, and everyone agreed on that we are not fans of these kinds of tests.
Well, I have rather strong different opinion. When I first saw table tests, I was a bit set back as well. But they can make testing and - more important - test maintenance less painful. They are not the right tool for all situations, but in this case, where you deal with a large number of different combinations of flags that each is just a small change to rather convoluted test input they are perfect.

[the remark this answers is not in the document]
StefanHauth: However, adding the field names could really improve readabilty because it also cleans up formatting.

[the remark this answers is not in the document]
StefanHauth: Btw, while the guidelines provide a suggested test structure, nowhere they explicitly forbid table tests.

src/webhook/validation/oneagent.go
0sewa0:
```suggestion
func getOneAgentVolumeStorageEnvVar(env []corev1.EnvVar) (enabled bool,  set bool) {
```
maybe for better readability 🤔

0sewa0:
something funky is happening in your git-branch as these files shouldn't have been changed:
```
src/api/v1beta1/feature_flags.go
src/api/v1beta1/feature_flags_test.go
src/controllers/csi/metadata/sqlite.go
src/controllers/csi/metadata/sqlite_test.go
src/controllers/csi/provisioner/controller_test.go
```

src/webhook/validation/oneagent_test.go
StefanHauth: True, but this tests are to make sure that we only get warnings and not any of the newly introduced errors in classicFullstack. So they are not only to test the "warning-feature" but also to make sure the "validation-feature" doesn't run havoc.

0sewa0 (approved):

PR #1103, opened 2022-08-29
Creates the shared image dir so `os.Rename` can work

0sewa0:
# Description

The `/data/codemodules` directory is not created, and the `os.Rename` is not smart enough to move a dir from `/data/tmp_zip` to `/data/codemodules/<image-digest>`
This causes that the CSI driver can't finish the download if `codeModulesImage` is used

## How can this be tested?
Deploy the operator on a fresh cluster (where `/data/codemodules` doesn't exist)
and create a Dynakube that uses the CSI driver with `codeModulesImage`

## Checklist

### DIFF
--- a/src/installer/image/installer.go
+++ b/src/installer/image/installer.go
@@ -48,6 +48,11 @@ func (installer ImageInstaller) ImageDigest() string {
 func (installer *ImageInstaller) InstallAgent(targetDir string) (bool, error) {
 	log.Info("installing agent from image")
 
+	err := installer.fs.Mkdir(installer.props.PathResolver.AgentSharedBinaryDirBase(), common.MkDirFileMode)
+	if err != nil {
+		return false, errors.WithStack(err)
+	}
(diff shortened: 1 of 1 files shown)


src/installer/image/installer.go
```diff
@@ -48,6 +48,11 @@ func (installer ImageInstaller) ImageDigest() string {
 func (installer *ImageInstaller) InstallAgent(targetDir string) (bool, error) {
 	log.Info("installing agent from image")
 
+	err := installer.fs.Mkdir(installer.props.PathResolver.AgentSharedBinaryDirBase(), common.MkDirFileMode)
```
luhi-DT, line 51: will this return an error in case the Dir already exists?

0sewa0: `Mkdir` yes
`MkdirAll` no 😅

luhi-DT (approved):

PR #1106, opened 2022-08-29
E2E-Tests a simple cloud native installation

meik99:
# Description

Adds an e2e test for a default cloud native installation.
It
* installs the operator using the Kubernetes specific manifest
* waits for the components to be available
* applies a sample app
* applies a dynakube cr
* waits for the components, i.e. activegate, oneagents, to start up
* waits for the dynakubes phase to change to 'Running'
* restarts sample app pods and checks for the existence of the init containers

N.B.: waiting for a resource is testing the feature because there is a timeout after ten minutes which would fail the test if the resources wasn't created.

## How can this be tested?

* Get a fresh cluster
* Comment the check for the phase change, because it doesn't change the phase right now due to a bug
* Run the test
* Monitor the cluster for changes and that the test does the things it should do

## Checklist


0sewa0:
the file structure seems to be all over the place (some files/folders are general, some are operator specific)

IMO we should have a `/tests/kubeobjects` folder where the commonly used functions should be (like the `daemonset/wait.go`)

0sewa0 (approved):

PR #1109, opened 2022-08-30
Uses MountAttempts in csi-driver

0sewa0:
# Description

In case the CSI volume mount fails we store a `Volume` entry in the database (if not already present), and increment its `MountAttempts` field.

If the `MountAttempts` reaches the `MaxFailedMountAttempts` defined the corresponding Dynakube, the server will mount a "dummy volume" to the pod. (dummy volume as of now is actually nothing)
This volume will be removed from the database during unmount.
As we still need to unmount these "dummy volumes", if a `Volume` in the database has >0 `MountAttempts` then those are `dummy volumes`, so the server can handle them accordingly.

## How can this be tested?
Use the `feature.dynatrace.com/max-csi-mount-attempts` feature flag to reduce the default `MaxFailedMountAttempts` (which is 10) to a value that is low enough to cause the dummy volume mount. 

## Checklist


src/controllers/csi/driver/volumes/app/publisher.go
```diff
@@ -87,6 +97,11 @@ func (publisher *AppVolumePublisher) UnpublishVolume(ctx context.Context, volume
 	}
 	log.Info("loaded volume info", "id", volume.VolumeID, "pod name", volume.PodName, "version", volume.Version, "dynakube", volume.TenantUUID)
 
+	if volume.MountAttempts > 0 {
```
meik99, line 100: i don't quite understand that:
* the publisher tries to mount a volume for `bindCfg.MaxMountAttempts` times
* only if `volume.MountAttempts > bindCfg.MaxMountAttempts` a dummy volume is mounted
* so wouldn't it be technically possible for a real, non-dummy volume to be correctly mounted even if it has `> 0` mount attempts, as long as it is `<= bindCfg.MaxMountAttempts`?

0sewa0: if a Volume was mounted correctly we set the MountAttempts to `0`

https://github.com/Dynatrace/dynatrace-operator/pull/1109/files#diff-8f7ea714d1cb32f86ee9d8d9b5e11671a9f52de76970b12bb57dc203308ea537R83
https://github.com/Dynatrace/dynatrace-operator/pull/1109/files#diff-8f7ea714d1cb32f86ee9d8d9b5e11671a9f52de76970b12bb57dc203308ea537R226
https://github.com/Dynatrace/dynatrace-operator/pull/1109/files#diff-8f7ea714d1cb32f86ee9d8d9b5e11671a9f52de76970b12bb57dc203308ea537R244

src/controllers/csi/driver/volumes/app/publisher.go
```diff
@@ -209,3 +235,11 @@ func (publisher *AppVolumePublisher) loadVolume(ctx context.Context, volumeID st
 	}
 	return volume, nil
 }
+
+func createNewVolume(bindCfg *csivolumes.BindConfig, volumeCfg *csivolumes.VolumeConfig) *metadata.Volume {
```
chrismuellner, line 239: should this be an actual constructor not just a helper function?

src/controllers/csi/metadata/metadata.go
```diff
@@ -41,7 +41,7 @@ type Volume struct {
 
 // NewVolume returns a new Volume if all fields are set.
```
chrismuellner, line 42: comment should be updated: `version` can be empty when we create a volume that has no version yet because of failures

src/controllers/csi/metadata/sqlite.go
```diff
@@ -50,7 +50,7 @@ const (
 
 	dynakubesAlterStatementMaxFailedMountAttempts = `
 	ALTER TABLE dynakubes
-	ADD COLUMN MaxFailedMountAttempts INT NOT NULL DEFAULT 3;`
+	ADD COLUMN MaxFailedMountAttempts INT NOT NULL DEFAULT 10;`
```
chrismuellner, line 53: can we use the constant here?

chrismuellner (changes requested):

src/controllers/csi/metadata/sqlite.go
```diff
@@ -3,12 +3,21 @@ package metadata
...
 
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	"github.com/mattn/go-sqlite3"
 	"github.com/pkg/errors"
 )
 
+var (
+	dynakubesAlterStatementMaxFailedMountAttempts = `
+	ALTER TABLE dynakubes
+	ADD COLUMN MaxFailedMountAttempts INT NOT NULL DEFAULT ` + strconv.FormatInt(dynatracev1beta1.DefaultMaxFailedCsiMountAttempts, 10) + ";"
```
chrismuellner, line 17: I hope having the constant a 2nd time here won't cause issues if we update it 😅

chrismuellner (approved):

src/controllers/csi/metadata/sqlite.go
0sewa0: I tested that, set the default value several different numbers and re-built/deployed  without issue :)

PR #1122, opened 2022-09-05
Adds toleration matching node affinity to activegate

0sewa0:
# Description
On Google Autopilot we need a Toleration that matches our node affinity for the activegate statefulset to be deployed.

Initially I only added the toleration in case of Autopilot, however this is unnecessary complication and now I just always add it. (I kept the leftover refactored code)

Also added `make deploy/gke-autopilot` that matches the capability of `make deploy/kubernetes` but its actually compatible with Autopilot.
While I was there I refactored the `make deploy/*` commands so that they no longer generate files (except the CRD 😞 )  

## How can this be tested?
- Deploy an activegate on AutoPilot
- Deploy an activegate on Kubernetes/Openshift

## Checklist


chrismuellner (changes requested):

src/kubeobjects/affinity.go
```diff
@@ -15,6 +17,17 @@ func AffinityNodeRequirement() []corev1.NodeSelectorRequirement {
 	return affinityNodeRequirementsForArches(amd64)
 }
 
+func TolerationMatchingNodeAffinity() []corev1.Toleration {
```
meik99, line 20:
```suggestion
func TolerationForAmd() []corev1.Toleration {
```

Seems more intuative imo

src/kubeobjects/pod_test.go
```diff
@@ -0,0 +1,35 @@
...
+	testNamespace := "testNamespace"
+	fakeClient := fake.NewClient(
+		&corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      testPodName,
+				Namespace: testNamespace,
+			},
+		},
+	)
+	t.Run("happy path", func(t *testing.T) {
```
meik99, line 25:
```suggestion
	t.Run("get existing pod", func(t *testing.T) {
```

src/kubeobjects/pod_test.go
```diff
@@ -0,0 +1,35 @@
...
+		pod, err := GetPod(context.TODO(), fakeClient, testPodName, testNamespace)
+		require.NoError(t, err)
+		assert.NotNil(t, pod)
+	})
+	t.Run("sad path", func(t *testing.T) {
```
meik99, line 30:
```suggestion
	t.Run("return error if pod does not exist", func(t *testing.T) {
```

meik99 (changes requested):

chrismuellner (approved):

PR #1128, opened 2022-09-06
Adds toleration for arm and amd to operator components

meik99:
# Description

Adds tolerations to operator components so that they are deployed on ARM nodes as well

## How can this be tested?

* Acquire an ARM cluster
* Generate the manifest
* Deploy the manifest (an arm image with the snapshot-<branch> tag exists already)
* Observe

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/Common/csi/daemonset.yaml
+++ b/config/helm/chart/default/templates/Common/csi/daemonset.yaml
@@ -295,6 +295,9 @@ spec:
         - key: kubernetes.io/arch
           value: amd64
           effect: NoSchedule
+        - key: ToBeDeletedByClusterAutoscaler
+          operator: Exists
+          effect: NoSchedule
   updateStrategy:
     rollingUpdate:
       maxUnavailable: 1
--- a/config/helm/chart/default/templates/Common/csi/daemonset.yaml
+++ b/config/helm/chart/default/templates/Common/csi/daemonset.yaml
@@ -285,9 +285,16 @@ spec:
       {{- if .Values.csidriver.nodeSelector }}
       nodeSelector: {{- toYaml .Values.csidriver.nodeSelector | nindent 8 }}
       {{- end }}
-      {{- if .Values.csidriver.tolerations }}
-      tolerations: {{- toYaml .Values.csidriver.tolerations | nindent 8 }}
-      {{- end }}
+      tolerations:
+        {{- if .Values.csidriver.tolerations }}
+        {{- toYaml .Values.csidriver.tolerations | nindent 8 }}
+        {{- end }}
+        - key: kubernetes.io/arch
+          value: arm64
+          effect: NoSchedule
+        - key: kubernetes.io/arch
+          value: amd64
+          effect: NoSchedule
   updateStrategy:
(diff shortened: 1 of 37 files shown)


config/helm/chart/default/values.yaml
```diff
@@ -21,7 +21,13 @@ installCRD: false
 
 operator:
   nodeSelector: {}
-  tolerations: []
+  tolerations:
```
chrismuellner, line 24: these tolerations should be unconfigurable and directly set in the deployment/daemonset, so tolerations that are set here are just added to the default tolerations and don't overwrite them.

chrismuellner (changes requested):

config/helm/chart/default/values.yaml
meik99: fixed

config/helm/chart/default/templates/Common/csi/daemonset.yaml
```diff
@@ -285,9 +285,16 @@ spec:
...
       nodeSelector: {{- toYaml .Values.csidriver.nodeSelector | nindent 8 }}
       {{- end }}
-      {{- if .Values.csidriver.tolerations }}
-      tolerations: {{- toYaml .Values.csidriver.tolerations | nindent 8 }}
-      {{- end }}
+      tolerations:
+        {{- if .Values.csidriver.tolerations }}
+        {{- toYaml .Values.csidriver.tolerations | nindent 8 }}
+        {{- end }}
+        - key: kubernetes.io/arch
```
0sewa0, line 292:
```
    - effect: NoSchedule
      key: ToBeDeletedByClusterAutoscaler
      operator: Exists
```
maybe having this here as well would make sense ? (its in the values.yaml, the other 2 in the values.yaml should be configurable)

meik99: done

hack/make/manifests/manifests.mk
```diff
@@ -1,5 +1,9 @@
 manifests/prepare-directory:
 	find $(MANIFESTS_DIR) -type f -not -name 'kustomization.yaml' -delete
 
-# Generates manifests e.g. CRD, RBAC etc, for Kubernetes and OpenShift
+## Generates manifests e.g. CRD, RBAC etc, for Kubernetes and OpenShift
 manifests: manifests/prepare-directory manifests/kubernetes manifests/openshift
+
+## Generate manifests for the branch's image tag
+manifests/branch: export MASTER_IMAGE=quay.io/dynatrace/dynatrace-operator:snapshot-${SNAPSHOT_SUFFIX}
```
0sewa0, line 8: Just out of curiosity, why ?

0sewa0 (approved):

hack/make/manifests/manifests.mk
meik99: I am lazy

meik99: the cloudnative tests use the manifests to deploy the operator. The image change is done in the kustomize, by default, the manifests only have the `:snapshot` tag. I did this so I can run the e2e tests more easily

PR #1135, opened 2022-09-08
Adds annotations for generated daemonset pods

meik99:
# Description

This change adds a `Annotation` field to the `HostInjectSpec`.
It allows specifying annotations to be added to the pods of oneagent daemonset.

## How can this be tested?

* Deploy a dynakube with cloudnative, classic, or host monitoring enabled and the annotations field set
* When deployed, the pods must have the default apparmor annotation that is always applied and additionally the custom annotation

## Checklist

### DIFF
--- a/src/kubeobjects/labels.go
+++ b/src/kubeobjects/labels.go
@@ -110,9 +110,9 @@ func (labels *AppLabels) BuildMatchLabels() map[string]string {
 	}
 }
 
-func MergeMap(labels ...map[string]string) map[string]string {
+func MergeMap(maps ...map[string]string) map[string]string {
 	res := map[string]string{}
-	for _, m := range labels {
+	for _, m := range maps {
 		for k, v := range m {
 			res[k] = v
 		}
--- a/src/kubeobjects/labels.go
+++ b/src/kubeobjects/labels.go
@@ -110,17 +110,6 @@ func (labels *AppLabels) BuildMatchLabels() map[string]string {
 	}
 }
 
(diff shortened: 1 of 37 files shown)


src/kubeobjects/labels.go
```diff
@@ -110,7 +110,7 @@ func (labels *AppLabels) BuildMatchLabels() map[string]string {
 	}
 }
 
-func MergeLabels(labels ...map[string]string) map[string]string {
+func MergeMap(labels ...map[string]string) map[string]string {
```
0sewa0, line 113:
```suggestion
func MergeMap(maps ...map[string]string) map[string]string {
```

0sewa0: (it could also use generics 😛 )

meik99: done

src/kubeobjects/labels.go
```diff
@@ -110,9 +110,9 @@ func (labels *AppLabels) BuildMatchLabels() map[string]string {
 	}
 }
 
-func MergeLabels(labels ...map[string]string) map[string]string {
+func MergeMap(maps ...map[string]string) map[string]string {
```
0sewa0, line 113: there is a `map.go` file, this function would fit it perfectly 🥲

meik99: :smiling_face_with_tear:

0sewa0 (approved):

PR #1150, opened 2022-09-13
Refactors capabilities into modifiers

0sewa0:
# Description

Simplifies the `capability` package so it now only has the values necessary to support the current `multiCapability` and the deprecated separate capabilities.
Converted anything possible into modifiers.
Refactored some parts for consistency.

## How can this be tested?
No new feature/behaviour, the activegate creation should work the same as before

## Checklist

### DIFF
--- a/src/api/v1beta1/properties.go
+++ b/src/api/v1beta1/properties.go
@@ -90,11 +90,41 @@ func (dk *DynaKube) IsActiveGateMode(mode CapabilityDisplayName) bool {
 	return false
 }
 
-func (dk *DynaKube) KubernetesMonitoringMode() bool {
+func (dk *DynaKube) ActiveGateServiceAccountOwner() string {
+	if dk.IsKubernetesMonitoringCapabilityEnabled() {
+		return string(KubeMonCapability.DeepCopy().DisplayName)
+	} else {
+		return "activegate"
+	}
+}
+
+func (dk *DynaKube) ActiveGateServiceAccountName() string {
+	return "dynatrace-" + dk.ActiveGateServiceAccountOwner()
+}
+
+func (dk *DynaKube) IsKubernetesMonitoringCapabilityEnabled() bool {
 	return dk.IsActiveGateMode(KubeMonCapability.DisplayName) || dk.Spec.KubernetesMonitoring.Enabled
 }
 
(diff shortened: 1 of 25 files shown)


src/api/v1beta1/properties.go
```diff
@@ -90,11 +90,41 @@ func (dk *DynaKube) IsActiveGateMode(mode CapabilityDisplayName) bool {
...
+func (dk *DynaKube) ActiveGateServiceAccountName() string {
+	return "dynatrace-" + dk.ActiveGateServiceAccountOwner()
+}
+
+func (dk *DynaKube) IsKubernetesMonitoringCapabilityEnabled() bool {
 	return dk.IsActiveGateMode(KubeMonCapability.DisplayName) || dk.Spec.KubernetesMonitoring.Enabled
 }
 
-func (dk *DynaKube) NeedsStatsd() bool {
+func (dk *DynaKube) IsRoutingCapabilityEnabled() bool {
```
meik99, line 109: Could the names not have the workd `Capability`? E.g., `IsRoutingEnabled`? Including `Capability` everywhere seems a bit redundant

meik99 (changes requested):
looks nice

src/api/v1beta1/properties.go
0sewa0: That was my first draft, but because its on the Dynakube I wanted to make it explicit that `Routing` is an ActiveGate thing, because `Routing` doesn't really mean anything if you don't know the context.

Thought about calling it `IsRoutingActiveGateEnabled`, maybe that is better, because `Capability` also means nothing without the context?

meik99: `IsRoutingActiveGateEnabled` sounds good

meik99:
The deployed ActiveGate pod only adds `metrics_ingest` for every capability. Only for `dynatrace-api` `restInterface` is added.
For example:
When 
```
    capabilities:
      - routing
      - kubernetes-monitoring
      - metrics-ingest
      - dynatrace-api
```
is applied, the env var is set to `      DT_CAPABILITIES:            metrics_ingest,metrics_ingest,metrics_ingest,restInterface`

meik99:
The eec and statsd containers are missing

meik99 (approved):

PR #1152, opened 2022-09-14
Implements E2E test for using CodeModules

meik99:
# Description

Implements a test to check whether the CSI driver can correctly download a codemodules image and reuse its content for different tenants and dynakubes.

## How can this be tested?

* Create a cluster
* Run the test
* Observe what it does and make sure it does the right things

## Checklist


0sewa0 (changes requested):
I run the tests locally, and the size of the expected size of the volume was off by 20 bytes 😅 
I think that check should be not so exact 😉

0sewa0 (approved):

PR #1155, opened 2022-09-14
Limited security context of init container

waodim:
# Description
With this PR the security context of the init container is limited as much as possible.

## How can this be tested?
Run an application in a labeled namespace, such that the app pod gets injected and check the init containers security context.

## Checklist


src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -27,22 +28,42 @@ func createInstallInitContainerBase(webhookImage, clusterID string, pod *corev1.
...
 			{Name: config.K8sNodeNameEnv, ValueFrom: kubeobjects.NewEnvVarSourceForField("spec.nodeName")},
 		},
-		SecurityContext: copyUserContainerSecurityContext(pod),
+		SecurityContext: copyUserContainerSecurityContext(pod, &dynakube),
 		Resources:       *dynakube.InitResources(),
 	}
 }
 
-func copyUserContainerSecurityContext(pod *corev1.Pod) *corev1.SecurityContext {
+func copyUserContainerSecurityContext(pod *corev1.Pod, dynakube *dynatracev1beta1.DynaKube) *corev1.SecurityContext {
```
chrismuellner, line 36: we don't want to copy the container security context anymore, but only ever use the lowest possible privileges. user/group should be copied, but rest of the other security context can be ignored!

chrismuellner (changes requested):

src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -27,22 +28,39 @@ func createInstallInitContainerBase(webhookImage, clusterID string, pod *corev1.
...
 	}
 }
 
-func copyUserContainerSecurityContext(pod *corev1.Pod) *corev1.SecurityContext {
-	var securityContext *corev1.SecurityContext
-	if len(pod.Spec.Containers) == 0 {
-		return securityContext
-	}
+func securityContextForInitContainer(pod *corev1.Pod) *corev1.SecurityContext {
+	var securityContext = new(corev1.SecurityContext)
```
chrismuellner, line 37:
```suggestion
	var securityContext = &corev1.SecurityContext{
	  RunAsNonRoot = address.Of(true),
	  ...
	}
```

I'd just create the limited security context in the beginning since we always need it

src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -27,22 +28,39 @@ func createInstallInitContainerBase(webhookImage, clusterID string, pod *corev1.
...
+
 	if pod.Spec.Containers[0].SecurityContext != nil {
```
chrismuellner, line 39: this check could be improved:
* what if the container's security context is set but does not have user/group?
* could we store the `containerSecurityContext` in a variable, so we don't have `pod.Spec.Containers[0].SecurityContext` every time?

src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -27,22 +28,39 @@ func createInstallInitContainerBase(webhookImage, clusterID string, pod *corev1.
...
+		securityContext.RunAsGroup = address.Of(int64(1001))
+		securityContext.RunAsUser = address.Of(int64(1001))
 	}
+
+	limitSecurityContext(securityContext)
+
 	return securityContext
 }
 
+func limitSecurityContext(ctx *corev1.SecurityContext) {
```
chrismuellner, line 52: this should just be done in the beginning and return the limited security context. if user/group is set in the container it should overwrite those values.

src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -27,22 +28,39 @@ func createInstallInitContainerBase(webhookImage, clusterID string, pod *corev1.
...
+	ctx.RunAsNonRoot = address.Of(true)
```
chrismuellner, line 53:
```yaml
seccompProfile:
              type: RuntimeDefault
```
seccomp profile can also be limited, see [k8s docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-seccomp-profile-for-a-container)

chrismuellner (changes requested):

src/webhook/mutation/pod_mutator/init_container.go
waodim: done

src/webhook/mutation/pod_mutator/init_container.go
waodim: done

src/webhook/mutation/pod_mutator/init_container.go
waodim: done

src/webhook/mutation/pod_mutator/init_container.go
waodim: done

0sewa0 (changes requested):
If you use `applicationMonitoring` without csi-driver then the init fails, because the tmp file where we download is outside of any volume so read-only fs kills it

you could update the url installer to not use the `/tmp` for downloading, it is done here https://github.com/Dynatrace/dynatrace-operator/blob/master/src/installer/url/installer.go#L78

src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -27,19 +28,37 @@ func createInstallInitContainerBase(webhookImage, clusterID string, pod *corev1.
...
+			},
+		},
+		SeccompProfile: &corev1.SeccompProfile{
+			Type: corev1.SeccompProfileTypeRuntimeDefault,
+		},
 	}
-	if pod.Spec.Containers[0].SecurityContext != nil {
-		securityContext = pod.Spec.Containers[0].SecurityContext.DeepCopy()
+
+	var podSecurityContext = pod.Spec.Containers[0].SecurityContext
```
chrismuellner, line 52: when the container security context is empty, we have to also look at the pod security context and use the user/group from there if it is set.

chrismuellner (changes requested):

src/installer/url/installer.go
```diff
@@ -75,7 +76,14 @@ func (installer UrlInstaller) UpdateProcessModuleConfig(targetDir string, proces
...
 func (installer UrlInstaller) installAgentFromUrl(targetDir string) error {
 	fs := installer.fs
-	tmpFile, err := afero.TempFile(fs, "", "download")
+	var tmpFile afero.File
+	var err error
+	if installer.isInitContainerMode() {
+		tmpFile, err = afero.TempFile(fs, targetDir, "download")
+	} else {
+		tmpFile, err = afero.TempFile(fs, filepath.Dir(targetDir), "download")
+	}
```
chrismuellner, line 85:
```suggestion
	path := ""
	if installer.isInitContainerMode() {
		path := targetDir
	} else {
		path := filepath.Dir(targetDir)
	}
	tmpFile, err := afero.TempFile(fs, path, "download")
```

chrismuellner (changes requested):

src/installer/url/installer.go
chrismuellner: unsure if we want to change the base path from `""` to `targetDir` for the none init container as well

src/webhook/mutation/pod_mutator/init_container.go
waodim: done.

src/installer/url/installer.go
waodim: is there an outcome now, whether we want to use `"" ` for the non-init container way or `targetDir`?

0sewa0: having defined path is preferable to having `""` as it means nothing

src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -27,19 +28,41 @@ func createInstallInitContainerBase(webhookImage, clusterID string, pod *corev1.
...
+
+	if containerSecurityContext != nil && containerSecurityContext.RunAsUser != nil && containerSecurityContext.RunAsGroup != nil {
+		securityContext.RunAsGroup = containerSecurityContext.RunAsGroup
+		securityContext.RunAsUser = containerSecurityContext.RunAsUser
+	} else if podSecurityContext != nil && podSecurityContext.RunAsUser != nil && podSecurityContext.RunAsGroup != nil {
+		securityContext.RunAsGroup = containerSecurityContext.RunAsGroup
+		securityContext.RunAsUser = containerSecurityContext.RunAsUser
+	} else {
+		securityContext.RunAsGroup = address.Of(int64(1001))
+		securityContext.RunAsUser = address.Of(int64(1001))
```
0sewa0, line 63: could we abstract it into a separate function ?

abstracting the ifs might be enough, so to have something `if hasUserAndGroupSet(containerSecurityContext) ...`

0sewa0 (approved):

PR #1162, opened 2022-09-19
Fixes check for ignored namespaces

0sewa0:
# Description

In case you had 2 Dynakubes, 1 used the `ignored-namespaces` feature-flag, and 1 didn't, then depending on the order in which the dynakubes were applied the result was different. (sometimes the `ignored-namespaces` feature-flag was considered, while sometimes it was not)

## How can this be tested?
Have 2 Dynakubes, 1 use the `ignored-namespaces` feature-flag, 1 doesn't
No matter in what order the dynakubes are applied, the `ignored-namespaces` should always be considered.

## Checklist

### DIFF
--- a/src/mapper/mapper.go
+++ b/src/mapper/mapper.go
@@ -78,7 +78,7 @@ func updateNamespace(namespace *corev1.Namespace, deployedDynakubes *dynatracev1
 	for i := range deployedDynakubes.Items {
 		dynakube := &deployedDynakubes.Items[i]
 		if isIgnoredNamespace(dynakube, namespace.Name) {
-			return false, nil
+			continue
 		}
 		matches, err := match(dynakube, namespace)
 


src/mapper/mapper.go
```diff
@@ -78,7 +78,7 @@ func updateNamespace(namespace *corev1.Namespace, deployedDynakubes *dynatracev1
 	for i := range deployedDynakubes.Items {
 		dynakube := &deployedDynakubes.Items[i]
 		if isIgnoredNamespace(dynakube, namespace.Name) {
-			return false, nil
+			continue
```
meik99, line 81: this could be tested using a unit test

meik99 (changes requested):

src/mapper/mapper.go
0sewa0: done

src/mapper/mapper_test.go
```diff
@@ -38,3 +43,113 @@ func createNamespace(name string, labels map[string]string) *corev1.Namespace {
...
+		dk := createTestDynakubeWithMultipleFeatures("dk-test", labels, nil)
+		namespace := createNamespace("test-namespace", labels)
+
+		updated, err := updateNamespace(namespace, &dynatracev1beta1.DynaKubeList{Items: []dynatracev1beta1.DynaKube{*dk}})
+
+		require.NoError(t, err)
+		require.True(t, updated)
+		assert.Equal(t, 2, len(namespace.Labels))
+	})
+	t.Run("Overwrite stale entry in labels", func(t *testing.T) {
```
meik99, line 59: that doesn't look like it checks for the label to have actually changed

src/mapper/mapper_test.go
```diff
@@ -38,3 +43,113 @@ func createNamespace(name string, labels map[string]string) *corev1.Namespace {
...
+		dk := createTestDynakubeWithMultipleFeatures("appMonitoring", nil, nil)
+		namespace := createNamespace("openshift-something", nil)
+
+		updated, err := updateNamespace(namespace, &dynatracev1beta1.DynaKubeList{Items: []dynatracev1beta1.DynaKube{*dk}})
+
+		require.NoError(t, err)
+		require.False(t, updated)
+		assert.Equal(t, 0, len(namespace.Labels))
+	})
+	t.Run("Double dynakube, 1. ignores openshift namespaces, 2. doesn't", func(t *testing.T) {
```
meik99, line 123: this should check that the label has the correct dynakube assigned

src/mapper/mapper_test.go
```diff
@@ -38,3 +43,113 @@ func createNamespace(name string, labels map[string]string) *corev1.Namespace {
...
+		}
+		namespace := createNamespace("openshift-something", labels)
+
+		updated, err := updateNamespace(namespace, &dynatracev1beta1.DynaKubeList{Items: []dynatracev1beta1.DynaKube{*ignoreDk, *notIgnoreDk}})
+
+		require.NoError(t, err)
+		require.True(t, updated)
+		assert.Equal(t, 2, len(namespace.Labels))
+	})
+	t.Run("Double dynakube, 1. doesn't, 2. ignores openshift namespaces", func(t *testing.T) {
```
meik99, line 139: this should check that the label has the correct dynakube assigned

meik99 (changes requested):

src/mapper/mapper_test.go
0sewa0: done

src/mapper/mapper_test.go
0sewa0: done

meik99 (approved):

PR #1170, opened 2022-09-22
Adapt rotation interval for activeGate authToken

luhi-DT:
# Description

We should make sure that the ActiveGate Token deployed by Dynatrace Operator is going to be rotated at least 31 days before it actually expires. This is currently not the case because we rotate it after 30 days + dynakube reconciliation interval. This leads to an unexpected message in the UI

## How can this be tested?
 Create a Dynakube with an AG -> this will create a AuthToken -> Change the creationTimestamp -> Check if the token is rotated

## Checklist

### DIFF
--- a/src/controllers/dynakube/activegate/internal/authtoken/reconciler.go
+++ b/src/controllers/dynakube/activegate/internal/authtoken/reconciler.go
@@ -17,7 +17,7 @@ import (
 
 const (
 	ActiveGateAuthTokenName   = "auth-token"
-	AuthTokenRotationInterval = time.Hour * 24 * 30
+	AuthTokenRotationInterval = time.Hour * 24 * 29
 )
 
 var _ kubeobjects.Reconciler = &Reconciler{}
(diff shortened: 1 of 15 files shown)


chrismuellner (dismissed):

src/controllers/dynakube/activegate/internal/authtoken/reconciler.go
```diff
@@ -17,7 +17,10 @@ import (
 
 const (
 	ActiveGateAuthTokenName   = "auth-token"
-	AuthTokenRotationInterval = time.Hour * 24 * 30
+	AuthTokenRotationInterval = time.Hour*24*30 - AuthTokenBuffer
+
+	// Buffer to avoid warnings in the UI
+	AuthTokenBuffer = time.Hour * 24
```
meik99, line 23:
```suggestion

	// Buffer to avoid warnings in the UI
	AuthTokenBuffer = time.Hour * 24
	AuthTokenRotationInterval = time.Hour*24*30 - AuthTokenBuffer
```

meik99 (approved):

PR #1183, opened 2022-09-23
Makes the securityContext copying safer

0sewa0:
# Description

We didn't took enough precautions when determining the `SecurityContext` of the injected init-container.

Removing `SeccompProfile` as it causes problems on openshift with the SCCs
 `RunAsNonRoot ` is only set if the finalised `SecurityContext` is not using the root user/group 

## How can this be tested?
Deploy applicationMonitoring with sample apps that have a variety of `SecurityContext` set.
Priority of user/group to be used: (in order of least to max priority)
- default
- `PodSecurityContext`
- "1. container's" `SecurityContext`

## Checklist

### DIFF
--- a/src/webhook/mutation/pod_mutator/init_container.go
+++ b/src/webhook/mutation/pod_mutator/init_container.go
@@ -71,11 +71,10 @@ func copySecurityContextUserAndGroup(securityContext *corev1.SecurityContext, po
 		securityContext.RunAsGroup = containerSecurityContext.RunAsGroup
 	}
 
-	if notRoot(securityContext) {
+	if isNonRoot(securityContext) {
 		securityContext.RunAsNonRoot = address.Of(true)
 	}
 }
-
 
 func hasPodUserSet(ctx *corev1.PodSecurityContext) bool {
 	return ctx != nil && ctx.RunAsUser != nil
@@ -93,12 +92,11 @@ func hasContainerGroupSet(ctx *corev1.SecurityContext) bool {
 	return ctx != nil && ctx.RunAsGroup != nil
 }
 
-func notRoot(ctx *corev1.SecurityContext) bool {
+func isNonRoot(ctx *corev1.SecurityContext) bool {
 	return ctx != nil &&
-	 (ctx.RunAsUser != nil && *ctx.RunAsUser != rootUser) &&
-	 (ctx.RunAsGroup != nil && *ctx.RunAsGroup != rootUser)
+		(ctx.RunAsUser != nil && *ctx.RunAsUser != rootUser) &&
+		(ctx.RunAsGroup != nil && *ctx.RunAsGroup != rootUser)
 }
-
 
 func getBasePodName(pod *corev1.Pod) string {
 	basePodName := pod.GenerateName
--- a/src/webhook/mutation/pod_mutator/init_container.go
+++ b/src/webhook/mutation/pod_mutator/init_container.go
@@ -49,7 +49,7 @@ func securityContextForInitContainer(pod *corev1.Pod) *corev1.SecurityContext {
 }
 
 // combineSecurityContexts returns a SecurityContext that combines the provided SecurityContext
-// with the relevant parts of the provided Pod's SecurityContext and the Pod's 1. container's SecurityContext
+// with the user/group of the provided Pod's SecurityContext and the 1. container's SecurityContext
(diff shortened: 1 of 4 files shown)


src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -44,34 +43,59 @@ func securityContextForInitContainer(pod *corev1.Pod) *corev1.SecurityContext {
...
+func hasContainerGroupSet(ctx *corev1.SecurityContext) bool {
+	return ctx != nil && ctx.RunAsGroup != nil
 }
 
-func hasPodUserAndGroupSet(ctx *corev1.PodSecurityContext) bool {
-	return ctx != nil && ctx.RunAsUser != nil && ctx.RunAsGroup != nil
+func isNonRoot(ctx *corev1.SecurityContext) bool {
+	return ctx != nil &&
+		(ctx.RunAsUser != nil && *ctx.RunAsUser != rootUser) &&
+		(ctx.RunAsGroup != nil && *ctx.RunAsGroup != rootUser)
```
chrismuellner, line 98: `rootUser` constant is used for both "user" and "group", so it should be named to something that better reflects that

src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -44,34 +43,59 @@ func securityContextForInitContainer(pod *corev1.Pod) *corev1.SecurityContext {
...
 
-	if hasContainerUserAndGroupSet(containerSecurityContext) {
-		securityContext.RunAsGroup = containerSecurityContext.RunAsGroup
-		securityContext.RunAsUser = containerSecurityContext.RunAsUser
-	} else if hasPodUserAndGroupSet(podSecurityContext) {
-		securityContext.RunAsGroup = podSecurityContext.RunAsGroup
+	securityContext.RunAsUser = address.Of(defaultUser)
+	securityContext.RunAsGroup = address.Of(defaultGroup)
+
+	if hasPodUserSet(podSecurityContext) {
```
chrismuellner, line 60: should we even set user/group for the init container when the pod one would still be valid here?

src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -44,34 +43,59 @@ func securityContextForInitContainer(pod *corev1.Pod) *corev1.SecurityContext {
 				"ALL",
 			},
 		},
-		SeccompProfile: &corev1.SeccompProfile{
-			Type: corev1.SeccompProfileTypeRuntimeDefault,
-		},
 	}
 
+	copySecurityContextUserAndGroup(securityContext, *pod)
```
chrismuellner, line 48: it should be more explicit that the security context is modified in this function by either returning the modified security context or only returning user/group and modifying the security context outside the function

src/webhook/mutation/pod_mutator/pod_mutator_test.go
```diff
@@ -90,11 +90,18 @@ func TestHandlePodMutation(t *testing.T) {
 		require.NoError(t, err)
 		assert.NotNil(t, mutationRequest.InstallContainer)
 		assert.Len(t, mutationRequest.Pod.Spec.InitContainers, 2)
+
+		require.NotNil(t, *mutationRequest.Pod.Spec.InitContainers[1].SecurityContext.Privileged)
```
chrismuellner, line 94: isn't `NotNil` kind of implied if the value is `False`? Could just be `require.False`

chrismuellner (changes requested):

src/webhook/mutation/pod_mutator/pod_mutator_test.go
0sewa0: there is an error here but its that I should have written `require.NotNil(t, mutationRequest....`

you have to do this because if you just `require.False(t, *mutationRequest....` then it can fail with a panic, as you tried to get the value behind a pointer that is `nil`, so you should check that the pointer itself must not be nil and check its value after

src/webhook/mutation/pod_mutator/init_container.go
0sewa0: this is so the code is kept simple (IMO)

otherwise to determine if we can set the `RunAsNonRoot` we would have to check if we set anything in our securityContext, and if not then check the pod securityContext

so the current logic is to set the securityContext to default then overwrite it in revers order of the priority of the user's securityContext (Priority 1. Container SC; Priority 2. Pod SC; Priority 3. Default SC) , then check the resulting securityContext if its root or not

PR #1184, opened 2022-09-23
Include e2e build tag into linter.

StefanHauth:
# Description

Build tag e2e is missing for linter run in pre-commit hook, which leads to problems when committing. Added it to pre-commit hook.

## How can this be tested?
Commit changes to a branch.

## Checklist


.github/pre-commit
```diff
@@ -38,7 +38,7 @@ else
     done
 fi
 
-golangci-lint run --build-tags integration,containers_image_storage_stub
+golangci-lint run --build-tags integration,containers_image_storage_stub,e2e
```
chrismuellner, line 41: instead of duplicating the commands, we should use `make go/lint` and `make go/test` respectively

chrismuellner: this file is pretty out of date 🙂

chrismuellner (dismissed):
makefile commands

luhi-DT (approved):

.github/pre-commit
StefanHauth: we will do the rework in a separate PR but get this out of the door now as it causes some considerable friction in day-to-day dev work

PR #1187, opened 2022-09-27
Fix json format of data-ingest injected file

mjgrzybek:
# Description

Cherry-pick of https://github.com/Dynatrace/dynatrace-operator/pull/1186

## Checklist
~- [ ] Unit tests have been updated/added~


src/controllers/dynakube/dynakube_controller.go
```diff
@@ -237,6 +237,7 @@ func (controller *DynakubeController) reconcileDynaKube(ctx context.Context, dkS
 
 		err = endpointSecretGenerator.GenerateForDynakube(ctx, dkState.Instance)
 		if dkState.Error(err) {
+			log.Error(err, "failed to generate data-ingest secret for dynakube")
```
chrismuellner, line 240: we don't `log.Error`, please provide an information inf log and return the error

chrismuellner (changes requested):

src/controllers/dynakube/dynakube_controller.go
mjgrzybek: Actually we do. There are 6 other occurrences of`log.Error(err` in this file and 65 in the project.
This error is consistent with other if branches in current function.

chrismuellner (approved):

PR #1188, opened 2022-09-27
Added annotations for generated ActiveGate Pods

waodim:
# Description

With this change a Annotations field will be introduced in the ActiveGate section of the Dynakube.
With it, it is possible to add configurable annotations to the ActiveGate Pods.

## How can this be tested?
Apply a Dynakube object where there are annotations in the ActiveGate sections set. Check, whether the annotations are present in the ActiveGate Pods spec.

## Checklist

### DIFF
--- a/src/controllers/dynakube/activegate/internal/statefulset/statefulset.go
+++ b/src/controllers/dynakube/activegate/internal/statefulset/statefulset.go
@@ -51,6 +51,7 @@ func (statefulSetBuilder StatefulSetBuilder) getBase() appsv1.StatefulSet {
 	var sts appsv1.StatefulSet
 	sts.ObjectMeta = statefulSetBuilder.getBaseObjectMeta()
 	sts.Spec = statefulSetBuilder.getBaseSpec()
+	sts.Spec.Template.ObjectMeta.Annotations = kubeobjects.MergeMap(sts.Spec.Template.ObjectMeta.Annotations, statefulSetBuilder.dynakube.Spec.ActiveGate.Annotations)
 	statefulSetBuilder.addLabels(&sts)
 	statefulSetBuilder.addTemplateSpec(&sts)
 
--- a/src/controllers/dynakube/activegate/internal/statefulset/statefulset_test.go
+++ b/src/controllers/dynakube/activegate/internal/statefulset/statefulset_test.go
@@ -337,3 +337,33 @@ func TestBuildCommonEnvs(t *testing.T) {
 		assert.Equal(t, dynakube.Spec.NetworkZone, zoneEnv.Value)
 	})
 }
+func TestAnnotations(t *testing.T) {
+	t.Run("default annotations", func(t *testing.T) {
+		dynakube := getTestDynakube()
+		multiCapability := capability.NewMultiCapability(&dynakube)
+		builder := NewStatefulSetBuilder(testKubeUID, testConfigHash, dynakube, multiCapability)
+		sts, _ := builder.CreateStatefulSet(nil)
+		expectedTemplateAnnotations := map[string]string{
+			consts.AnnotationActiveGateConfigurationHash: testConfigHash,
+		}
+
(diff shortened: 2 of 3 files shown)


src/controllers/dynakube/activegate/internal/statefulset/statefulset.go
```diff
@@ -51,6 +51,7 @@ func (statefulSetBuilder StatefulSetBuilder) getBase() appsv1.StatefulSet {
 	var sts appsv1.StatefulSet
 	sts.ObjectMeta = statefulSetBuilder.getBaseObjectMeta()
 	sts.Spec = statefulSetBuilder.getBaseSpec()
+	sts.Spec.Template.ObjectMeta.Annotations = kubeobjects.MergeMap(sts.Spec.Template.ObjectMeta.Annotations, statefulSetBuilder.dynakube.Spec.ActiveGate.Annotations)
```
luhi-DT, line 54: This could be also a function like `statefulSetBuilder.addLabels(&sts)`

0sewa0, line 54: should be in `getBaseObjectMeta()` (line 68)

src/controllers/dynakube/activegate/internal/statefulset/statefulset_test.go
```diff
@@ -337,3 +337,33 @@ func TestBuildCommonEnvs(t *testing.T) {
 		assert.Equal(t, dynakube.Spec.NetworkZone, zoneEnv.Value)
 	})
 }
+func TestAnnotations(t *testing.T) {
```
0sewa0, line 340: could be part of `TestGetBaseObjectMeta`

src/controllers/dynakube/activegate/internal/statefulset/statefulset.go
waodim: As much as it makes sense to place it there, i encountered a problem. when placing it there the annotations from the dynakube seem to be overwritten and only the default annotations are there. This might be because we first get the `ObjectMeta` itself where the annotations are set to an empty map, and then we get the `Spec` where also again the `ObjectMeta` is changed and the annotations are set to the default ones.

So I think the approach from Lukas to just introduce a function for example `statefulSetBuilder.addAnnotations(&sts)` in line 54 is the best way here. What do you think?

src/controllers/dynakube/activegate/internal/statefulset/statefulset.go
```diff
@@ -94,6 +95,10 @@ func (statefulSetBuilder StatefulSetBuilder) addLabels(sts *appsv1.StatefulSet)
 	sts.Spec.Template.ObjectMeta.Labels = kubeobjects.MergeMap(statefulSetBuilder.capability.Properties().Labels, appLabels.BuildLabels())
 }
 
+func (statefulSetBuilder StatefulSetBuilder) addAnnotations(sts *appsv1.StatefulSet) {
+	sts.Spec.Template.ObjectMeta.Annotations = kubeobjects.MergeMap(sts.Spec.Template.ObjectMeta.Annotations, statefulSetBuilder.dynakube.Spec.ActiveGate.Annotations)
+}
+
```
0sewa0, line 101:
```suggestion
func (statefulSetBuilder StatefulSetBuilder) addUserAnnotations(sts *appsv1.StatefulSet) {
	sts.ObjectMeta.Annotations = kubeobjects.MergeMap(sts.ObjectMeta.Annotations, statefulSetBuilder.dynakube.Spec.ActiveGate.Annotations)
	sts.Spec.Template.ObjectMeta.Annotations = kubeobjects.MergeMap(sts.Spec.Template.ObjectMeta.Annotations, statefulSetBuilder.dynakube.Spec.ActiveGate.Annotations)
}
```
maybe we should add the annotations everywhere ? (on the statefulset and the pod)

0sewa0 (approved):

PR #1200, opened 2022-09-28
Extend troubleshoot command to check if code modules image can be pulled if set.

StefanHauth:
# Description

Add a check to the troubleshoot command so that it checks if a code modules image can be pulled if it is set. If no code modules image is set, no check is performed.

## How can this be tested?
Deploy operator to a cluster and run
`kubectl -n dynatrace exec pod/$(kubectl -n dynatrace get pods | grep dynatrace-operator | cut -f1 -d' ') -- /usr/local/bin/dynatrace-operator troubleshoot`

## Checklist


src/cmd/troubleshoot/image_pullable.go
```diff
@@ -70,14 +67,27 @@ func checkOneAgentImagePullable(troubleshootCtx *troubleshootContext) error {
...
+	pullSecret, err := getPullSecretToken(troubleshootCtx)
+	if err != nil {
 		return err
 	}
 
+	if dynakubeOneAgentCodeModulesImage := getOneAgentCodeModulesImageEndpoint(troubleshootCtx); dynakubeOneAgentCodeModulesImage != "" {
+		if err := checkCustomModuleImagePullable(troubleshootCtx.httpClient, "OneAgentCodeModules", pullSecret, dynakubeOneAgentCodeModulesImage); err != nil {
+			return err
+		}
+	}
```
0sewa0, line 90:
```suggestion
	dynakubeOneAgentCodeModulesImage := getOneAgentCodeModulesImageEndpoint(troubleshootCtx)
	err := checkCustomModuleImagePullable(troubleshootCtx.httpClient, "OneAgentCodeModules", pullSecret, dynakubeOneAgentCodeModulesImage)
	if err != nil {
		return err
	}
```

so that it would look the same as in line 72-75 in `checkOneAgentImagePullable`

src/cmd/troubleshoot/image_pullable.go
```diff
@@ -70,14 +67,27 @@ func checkOneAgentImagePullable(troubleshootCtx *troubleshootContext) error {
 	}
 
 	dynakubeOneAgentImage := getOneAgentImageEndpoint(troubleshootCtx)
-	if err != nil {
+
+	if err = checkComponentImagePullable(troubleshootCtx.httpClient, "OneAgent", pullSecret, dynakubeOneAgentImage); err != nil {
 		return err
 	}
```
0sewa0, line 73:
```suggestion
	err = checkComponentImagePullable(troubleshootCtx.httpClient, "OneAgent", pullSecret, dynakubeOneAgentImage);
	if  err != nil {
		return err
	}
```

IMO this is a bit more readable

src/cmd/troubleshoot/image_pullable.go
```diff
@@ -155,6 +152,64 @@ func checkComponentImagePullable(httpClient *http.Client, componentName string,
 	return nil
 }
 
+func checkCustomModuleImagePullable(httpClient *http.Client, componentName string, pullSecret string, componentImage string) error {
```
0sewa0, line 155: why is this needed if we have `checkComponentImagePullable` ?
what is the difference ?
AFAIK, the `codeModulesImage` is just as an image, just like the `image` field.
The content of the image is different, but the "is it pullable" check should be the same for both.

0sewa0 (changes requested):

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,141 @@
...
+var (
+	removeSchemaRegex      = regexp.MustCompile("^.*//(.*)$")
+	removeApiEndpointRegex = regexp.MustCompile("^(.*)/[^/]*$")
+	registryRegex          = regexp.MustCompile(`^(.*)/linux.*$`)
+	imageRegex             = regexp.MustCompile(`^.*/(linux.*)$`)
+)
+
+func splitImageName(imageName string) (registry string, image string, version string, err error) {
+	// some image path examples that work with this function
+	//   exq67461.dev.dynatracelabs.com/linux/oneagent
```
chrismuellner, line 18: remove mentions of specific dev tenants

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+var (
+	removeSchemaRegex      = regexp.MustCompile("^.*//(.*)$")
+	removeApiEndpointRegex = regexp.MustCompile("^(.*)/[^/]*$")
+	registryRegex          = regexp.MustCompile(`^(.*)/linux.*$`)
+	imageRegex             = regexp.MustCompile(`^.*/(linux.*)$`)
+)
+
+func splitImageName(imageName string) (registry string, image string, version string, err error) {
```
chrismuellner, line 16: registry, image, version should be a struct (e.g. `imageInfo`) and should be in the return, not implicitly returned

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,141 @@
...
+		image = fields[0]
+		version = fields[1]
+		logInfof("using custom image version: %s", version)
+	} else {
+		err = fmt.Errorf("invalid version of the image {\"image\": \"%s\"}", image)
+	}
+	return
+}
+
+func splitCustomImageName(imageURL string) (registry string, imagePath string, err error) {
```
chrismuellner, line 55: same here

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	logInfof("OneAgent image endpoint '%s'", imageEndpoint)
+	return imageEndpoint
+}
+
+func getOneAgentCodeModulesImageEndpoint(troubleshootCtx *troubleshootContext) string {
+	imageEndpoint := ""
+	switch {
+	case troubleshootCtx.dynakube.CloudNativeFullstackMode():
```
chrismuellner, line 112: should be ifs here, not a switch

src/cmd/troubleshoot/image_helper.go
chrismuellner: we already have `properties.go` and `image_paths.go` to check image fields. The logic here feels similar and we should combine it and move it to a location where it can be reused.

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,141 @@
...
+func getOneAgentImageEndpoint(troubleshootCtx *troubleshootContext) string {
+	customImage := ""
+	imageEndpoint := ""
+	version := ""
+
+	sr := removeSchemaRegex.FindStringSubmatch(troubleshootCtx.dynakube.Spec.APIURL)
+	er := removeApiEndpointRegex.FindStringSubmatch(sr[1])
+	imageEndpoint = er[1] + "/linux/oneagent"
+
+	if troubleshootCtx.dynakube.ClassicFullStackMode() {
```
chrismuellner, line 88: accessing all these images exists in `properties.go` (e.g. `CodeModulesImage()`), no need to duplicate this here.
all 3 of these functions (oneagent, codemoduels and activegate) can just use these getters to figure out the image and check it the same way

chrismuellner (changes requested):

src/cmd/troubleshoot/image_pullable.go
StefanHauth: The URL format of the component images are specific to DT, whereas custom images can have any URL. So they need different treatment.

src/cmd/troubleshoot/image_helper.go
StefanHauth: Switches are better readable.

src/cmd/troubleshoot/image_helper.go
StefanHauth: This function has only be moved and is not part of this PR per-se. So I'd rather not mix general code cleanups with the purpose of this PR. If necessary, this kind of cleanups shall be a separate PR.

src/cmd/troubleshoot/image_pullable.go
0sewa0: AFAIK, the `image` field can also have any kind of URL

src/cmd/troubleshoot/image_helper.go
StefanHauth: Again, function has just been moved, so IMHO not in the scope of this PR.

src/cmd/troubleshoot/image_helper.go
meik99: I'd have no problems with smaller, general changes being in the same PR. Just large overhauls should be their own ones imo

src/cmd/troubleshoot/image_pullable.go
luhi-DT: The function `checkOneAgentImagePullable` gets the right image which needs to be pulled from `getOneAgentImageEndpoint` this can return an image that is either from our own registry like tenant.dynatrace.com/v2/linux/oneagent:latest or an custom image if its set (see the if statements in that function) this could be like quay.io/myfanceregistry/repo/agent:something. This output is passed to `checkComponentImagePullable` which does not care where the image is (either own registry our Dynatrace one) The same can also be done with the codeModules image.

luhi-DT (changes requested):

src/cmd/troubleshoot/image_helper.go
luhi-DT: I agree here, if we already touch that code (does not matter if its changed or just moved) we can make it better in the same PR

StefanHauth: The whole `splitImage` code will be revisted in another PR because there seem to be other problems hidden. Will take this comment as input for that PR as well.

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
+package troubleshoot
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+var (
+	removeSchemaRegex      = regexp.MustCompile("^.*//(.*)$")
```
meik99, line 10: nitpick: with regexes I think a comment is necessary to describe what they are doing. maybe with an example, so they don't have to be manually parsed when read

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
...
+type imageInfo struct {
+	registry string
+	image    string
+	version  string
+}
+
+func splitImageName(imageName string) (imageInfo, error) {
+	// some image path examples that work with this function
+	//   aaa00000.dynatrace.com/linux/oneagent
+	//   aaa00000.dynatrace.com/linux/activegate:1.123
```
meik99, line 25: Comment could be a doc-string

```suggestion
// splitImageName creates an instance of imageInfo from an image uri.
// image uris can be in the form of
// * aaa00000.dynatrace.com/linux/oneagent
// * aaa00000.dynatrace.com/linux/activegate:1.123
func splitImageName(imageName string) (imageInfo, error) {
```

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
...
+
+	imgInfo := imageInfo{}
+
+	registryMatches := registryRegex.FindStringSubmatch(imageName)
+	if len(registryMatches) < 2 {
+		return imageInfo{}, fmt.Errorf("invalid image - registry not found (%s)", imageName)
+	}
+	imgInfo.registry = registryRegex.FindStringSubmatch(imageName)[1]
```
meik99, line 33:
```suggestion
	imgInfo.registry = registryMatches[1]
```

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
...
+
+	imageMatches := imageRegex.FindStringSubmatch(imageName)
+	if len(imageMatches) < 2 {
+		return imageInfo{}, fmt.Errorf("invalid image - endpoint not found (%s)", imageName)
+	}
+	imgInfo.image = imageRegex.FindStringSubmatch(imageName)[1]
```
meik99, line 39:
```suggestion
	imgInfo.image = imageMatches[1]
```
If you already use submatches, the registryRegex and imageRegex could be combined so FindStringSubmatch must only be called once.
I.e.

```
regex = regexp.MustCompile(`^(.*)/(linux.*)$`)

parsedUri := regex.FindStringSubmatch(imageName)
imgInfo.registry = parsedUri[1]
imgInfo.image = parsedUri[2]
```

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
...
+		// no version set, default to latest
+		imgInfo.version = "latest"
+		logInfof("using latest image version")
+	} else if len(fields) >= 2 {
+		imgInfo.image = fields[0]
+		imgInfo.version = fields[1]
+		logInfof("using custom image version: %s", imgInfo.version)
+	} else {
+		return imageInfo{}, fmt.Errorf("invalid version of the image {\"image\": \"%s\"}", imgInfo.image)
+	}
```
meik99, line 55: should be moved into its own function, e.g., `parseVersion(...)`

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
...
+	return imgInfo, nil
+}
+
+func splitCustomImageName(imageURL string) (imageInfo, error) {
+	imgInfo := imageInfo{}
+
+	customImageRegistryRegex := regexp.MustCompile(`^(.*?)/(.*)$`)
```
meik99, line 62: nitpick: with regexes I think a comment is necessary to describe what they are doing. maybe with an example, so they don't have to be manually parsed when read

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
...
+	if len(imgInfo.image) == 0 {
+		return imageInfo{}, fmt.Errorf("invalid image path (%s) - image path not found", imageURL)
+	}
+	return imgInfo, nil
+}
+
+func getOneAgentImageEndpoint(troubleshootCtx *troubleshootContext) string {
+	imageEndpoint := ""
+
+	sr := removeSchemaRegex.FindStringSubmatch(troubleshootCtx.dynakube.Spec.APIURL)
```
meik99, line 83: what does `sr` stand for?

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
...
+	er := removeApiEndpointRegex.FindStringSubmatch(sr[1])
```
meik99, line 84: what does `er` stand for?

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
...
+	imageEndpoint = er[1] + "/linux/oneagent"
+
+	customImage := troubleshootCtx.dynakube.CustomOneAgentImage()
+	version := troubleshootCtx.dynakube.Version()
+
+	if customImage != "" {
+		imageEndpoint = customImage
```
meik99, line 91: if custom image is set, it can be returned at the top of the function, as nothing must be parsed in that scenario

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
...
+	} else {
+		logInfof("OneAgent codeModules image endpoint is not used.")
+	}
+	return imageEndpoint
+}
+
+func getActiveGateImageEndpoint(troubleshootCtx *troubleshootContext) string {
+	imageEndpoint := ""
+
+	sr := removeSchemaRegex.FindStringSubmatch(troubleshootCtx.dynakube.Spec.APIURL)
```
meik99, line 114: what does `sr` stand for?

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
...
+	er := removeApiEndpointRegex.FindStringSubmatch(sr[1])
```
meik99, line 115: what does `er` stand for?

src/cmd/troubleshoot/image_pullable.go
```diff
@@ -90,67 +104,111 @@ func checkActiveGateImagePullable(troubleshootCtx *troubleshootContext) error {
...
+		return fmt.Errorf("invalid pull secret, could not unmarshal to JSON: %w", err)
+	}
+
+	codeModulesImageInfo, err := splitCustomImageName(codeModulesImage)
+	if err != nil {
+		return fmt.Errorf("invalid image URL: %w", err)
+	}
+	logInfof("using '%s' on '%s' as OneAgentCodeModules image", codeModulesImage, codeModulesImageInfo.registry)
+
+	endpoint, ok := result.Auths[codeModulesImageInfo.registry]
```
meik99, line 173:
```suggestion
	endpoint, hasEndpoint := result.Auths[codeModulesImageInfo.registry]
```

src/cmd/troubleshoot/image_pullable.go
```diff
@@ -90,67 +104,111 @@ func checkActiveGateImagePullable(troubleshootCtx *troubleshootContext) error {
...
-		logOkf("%s image '%s' found", componentName, componentRegistry+"/"+componentImage)
+		logOkf("%s image '%s' found", componentName, componentImageInfo.registry+"/"+componentImageInfo.image)
 	} else {
-		return fmt.Errorf("%s image '%s' missing", componentName, componentRegistry+"/"+componentImage)
+		return fmt.Errorf("%s image '%s' missing", componentName, componentImageInfo.registry+"/"+componentImageInfo.image)
+	}
+	return nil
+}
+
+func checkCustomModuleImagePullable(httpClient *http.Client, componentName string, pullSecret string, codeModulesImage string) error {
```
meik99, line 160: there is a docker library we already use in `installer/image` that could be used to do this. Would save us the need to re-implement this functionality if we can use that.

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,125 @@
+package troubleshoot
```
meik99, line 1: multiple files with the same prefix should be in a package imo.
I.e., `image_helper` and `image_pullable` should be `image/info` and `image/pull` (`image/check`) respectivly

meik99 (dismissed):

src/cmd/troubleshoot/image_helper.go
StefanHauth: dont't know, code was not created in this PR.

src/cmd/troubleshoot/image_helper.go
StefanHauth: dont't know, code was not created in this PR.

src/cmd/troubleshoot/image_pullable.go
StefanHauth: will note that down with all the other findings.

src/cmd/troubleshoot/image_helper.go
StefanHauth: Wouldn't that then end up with one file per package? IMHO a bit of an overkill.

meik99: would be one package with four files
```
image/
  info.go
  info_test.go
  pull.go
  pull_test.go
```

src/cmd/troubleshoot/image_helper.go
meik99: i guess renaming it wouldn't be that much of a change an could be done in this pr

src/cmd/troubleshoot/image_helper.go
meik99: i guess renaming it wouldn't be that much of a change an could be done in this pr

src/cmd/troubleshoot/image_helper.go
luhi-DT: we would not log the final imageEndpoint we use i guess its fine like that

src/cmd/troubleshoot/image_helper.go
StefanHauth: I tried and failed. Figuring out exactly what those regexes do is urgently necessary, but out of scope here. This is part of a piece of code I just moved. The regex stuff in this file generally needs some more love.

src/cmd/troubleshoot/image_helper.go
```diff
@@ -0,0 +1,138 @@
...
+	image    string
+	version  string
+}
+
+// Split image path in into its components.
+// Some image path examples that work with this function:
+//
+//	aaa00000.dynatrace.com/linux/oneagent
+//	aaa00000.dynatrace.com/linux/activegate:1.123
+func splitImageName(imageName string) (imageInfo, error) {
```
mjgrzybek, line 27: Parsing image could be moved to shared package so it can be reused.
We already have some kind of image parsing implemented here: https://github.com/Dynatrace/dynatrace-operator/blob/master/src/installer/image/source.go#L30

StefanHauth: Wil note it down with all other non PR related findings.

luhi-DT (approved):

PR #1232, opened 2022-10-11
Use version in application monitoring without CSI

0sewa0:
# Description

In case of applicationMonitoring the `version` field only worked with the CSI driver but didn't work in standalone mode.

## How can this be tested?
Set the version field in an `applicationMonitoring` dynakube and check if the deployed sample apps use the correct version

## Checklist


src/webhook/mutation/pod_mutator/oneagent_mutation/annotations.go
```diff
@@ -3,17 +3,18 @@ package oneagent_mutation
...
 	dtwebhook "github.com/Dynatrace/dynatrace-operator/src/webhook"
 	corev1 "k8s.io/api/core/v1"
 )
 
 type installerInfo struct {
-	flavor        string
-	technologies  string
-	installPath   string
-	installerURL  string
-	failurePolicy string
```
chrismuellner, line 16: what happened to the `failurePolicy`?

0sewa0: its redundant here, its not a oneagent specific thing so it still present but not in the `oneagent_mutation`

0sewa0: here: https://github.com/Dynatrace/dynatrace-operator/blob/master/src/webhook/mutation/pod_mutator/init_container.go#L23

PR #1233, opened 2022-10-12
Enhances troubleshoot error message

meik99:
# Description

If a Dynakube is invalid, the message now includes a hint that parameters can be used to specify different dynakube instances.
Also, the default namespace is now the one the pod is deployed in instead of just `dynatrace`

## How can this be tested?

* Deploy the Operator in a namespace different to `dynatrace`
* Do __not__ apply a Dynakube
* Run the troubleshoot script without any parameters

## Checklist
  - The changes are too small to warrant tests


src/cmd/troubleshoot/dynakube.go
```diff
@@ -40,14 +40,21 @@ func checkDynakube(troubleshootCtx *troubleshootContext) error {
 	for _, test := range tests {
 		if err := test(troubleshootCtx); err != nil {
 			logErrorf(err.Error())
-			return fmt.Errorf("'%s:%s' Dynakube isn't valid", troubleshootCtx.namespaceName, troubleshootCtx.dynakubeName)
+			return fmt.Errorf("'%s:%s' Dynakube isn't valid. %s",
```
chrismuellner, line 43: I'd remove the "not valid" part. The actual reason is that the Dynakube is missing

meik99: This loop iterates over multiple tests. It doesn't necessarily error at the "does not exist" part

StefanHauth (approved):

PR #1238, opened 2022-10-12
Propagate namespace label references as environment variables used for version detection

mjgrzybek:
# Description
Namespace can be annotated to add additional mapping:
```
annotations:
  mapping.release.dynatrace.com/version: "metadata.annotations['my-version']"
  mapping.release.dynatrace.com/product: "metadata.labels['app.kubernetes.io/name']"
  mapping.release.dynatrace.com/stage: "metadata.namespace"
  mapping.release.dynatrace.com/build-version: "metadata.labels['release.dynatrace.com/stage']"
```
This will configure the mapping used to get the value of the corresponding environment variable via reference in the injected pod.
References are not validated by operator.

Mapping for the different environment variables:
- `DT_RELEASE_VERSION`
  - mapping: `mapping.release.dynatrace.com/version`
  - fallback to default when mapping is missing
- `DT_RELEASE_PRODUCT`
  - mapping: `mapping.release.dynatrace.com/product`
  - fallback to default when mapping is missing
- `DT_RELEASE_STAGE`
  - mapping: `mapping.release.dynatrace.com/stage`
  - no default, no environment variable is missing
- `DT_RELEASE_BUILD_VERSION`
  - mapping: `mapping.release.dynatrace.com/build-version`
  - no default, no environment variable is missing

## How can this be tested?
Apply dynakube CR with `feature.dynatrace.com/label-version-detection: "true"`
Annotate monitored apps namespace with mappings described above, for example:
```
apiVersion: v1                                                                     
kind: Namespace                                                                    
metadata:                                                                          
  annotations:                                                                     
    mapping.release.dynatrace.com/build-version: metadata.annotations['my-build-version']
    mapping.release.dynatrace.com/product: metadata.annotations['my-product']      
    mapping.release.dynatrace.com/stage: metadata.annotations['my-stage']        
    mapping.release.dynatrace.com/version: metadata.annotations['my-version']      
```
Create pod with annotations set:
```
apiVersion: v1                                                 
kind: Pod                                                      
metadata:                                                      
  annotations:                                                 
    data-ingest.dynatrace.com/injected: "true"                 
    dynakube.dynatrace.com/injected: "true"                    
    my-build-version: my custom build-version from deployment-pod-template 
    my-product: ppp
    my-stage: sss
    my-version: bbb
```
Check if pod's env's env variables have values taken from pod's annotations

## Checklist


src/webhook/mutation/pod_mutator/oneagent_mutation/config.go
```diff
@@ -9,12 +9,14 @@ var (
...
-	networkZoneEnv       = "DT_NETWORK_ZONE"
-	proxyEnv             = "DT_PROXY"
-	dynatraceMetadataEnv = "DT_DEPLOYMENT_METADATA"
-	releaseVersionEnv    = "DT_RELEASE_VERSION"
-	releaseProductEnv    = "DT_RELEASE_PRODUCT"
+	preloadEnv             = "LD_PRELOAD"
+	networkZoneEnv         = "DT_NETWORK_ZONE"
+	proxyEnv               = "DT_PROXY"
+	dynatraceMetadataEnv   = "DT_DEPLOYMENT_METADATA"
+	releaseVersionEnv      = "DT_RELEASE_VERSION"
```
meik99, line 16: thats the most nitpicky thing, but I would prefer a newline here for better visbility
```suggestion
	dynatraceMetadataEnv   = "DT_DEPLOYMENT_METADATA"
	
	releaseVersionEnv      = "DT_RELEASE_VERSION"
```

src/webhook/mutation/pod_mutator/oneagent_mutation/versionlabel.go
```diff
@@ -18,11 +26,26 @@ func newVersionLabelMapping(namespace corev1.Namespace) VersionLabelMapping {
 }
 
 func getMappingFromNamespace(namespace corev1.Namespace) VersionLabelMapping {
-	// TODO: Implementation => Parse Namespace annotations into correct format
-	return VersionLabelMapping{}
+	mappingKey2versionLabelKey := map[string]string{
```
meik99, line 29:
```suggestion
	annotationLabelMap := map[string]string{
```

src/webhook/mutation/pod_mutator/oneagent_mutation/versionlabel.go
```diff
@@ -18,11 +26,26 @@ func newVersionLabelMapping(namespace corev1.Namespace) VersionLabelMapping {
...
+		versionMappingAnnotationName: releaseVersionEnv,
+		productMappingAnnotationName: releaseProductEnv,
+		stageMappingAnnotationName:   releaseStageEnv,
+		buildVersionAnnotationName:   releaseBuildVersionEnv,
+	}
+
+	versionLabelMapping := VersionLabelMapping{}
+	for mappingKey, versionLabelKey := range mappingKey2versionLabelKey {
```
meik99, line 37:
```suggestion
	for annotation, label := range mappingKey2versionLabelKey {
```

src/webhook/mutation/pod_mutator/oneagent_mutation/versionlabel.go
```diff
@@ -18,11 +26,26 @@ func newVersionLabelMapping(namespace corev1.Namespace) VersionLabelMapping {
...
+		if fieldRef, ok := namespace.Annotations[mappingKey]; ok {
```
meik99, line 38:
```suggestion
		if fieldRef, hasAnnotation := namespace.Annotations[annotation]; hasAnnotation {
```

meik99 (dismissed):

src/webhook/mutation/pod_mutator/oneagent_mutation/versionlabel.go
mjgrzybek: `ok` is idiomatic. I also think it's unfortunate but prefer to follow general rules.

src/webhook/mutation/pod_mutator/oneagent_mutation/versionlabel.go
mjgrzybek: Both annotation and label conceptually are key and value pairs.
In this case, we assign `...Key` as a value and I think keeping naming straightforward is better.

src/webhook/mutation/pod_mutator/oneagent_mutation/versionlabel.go
meik99: if we use golang conventions the line would look like so 
`if f, ok := n.Annotations[a]; ok {`
we already break a lot of conventions because most of them make code unreadable, so if we don't use `ok` for checking if a key is in a map I think thats okay as well

src/webhook/mutation/pod_mutator/oneagent_mutation/versionlabel.go
meik99: how about `annotationKey` and `labelKey`?

src/webhook/mutation/pod_mutator/oneagent_mutation/versionlabel.go
mjgrzybek: It was already discussed here: https://github.com/Dynatrace/dynatrace-operator/pull/981#discussion_r926447779

src/webhook/mutation/pod_mutator/oneagent_mutation/containers_test.go
```diff
@@ -150,10 +153,134 @@ func TestReinvokeUserContainers(t *testing.T) {
...
+}
+
+func assertDoesntContainMappings(t *testing.T, unexpectedMappingKeys []string, request *dtwebhook.MutationRequest) {
+	for _, key := range unexpectedMappingKeys {
+		idx := slices.IndexFunc(request.Pod.Spec.Containers[0].Env, func(envvar corev1.EnvVar) bool {
+			return envvar.Name == key
+		})
+		assert.Negative(t, idx, key+" found in container's envvars")
+	}
+}
```
0sewa0, line 286:
```suggestion
func assertNotContainsMappings(t *testing.T, unexpectedMappingKeys []string, request *dtwebhook.MutationRequest) {
	for _, env := range request.Pod.Spec.Containers[0].Env {
		assert.NotContains(t, unexpectedMappingKeys, env.Name)
	}
}
```

lets try not to use fancy experimental stuff, especially when there is a simpler solution

src/webhook/mutation/pod_mutator/oneagent_mutation/versionlabel.go
```diff
@@ -18,11 +26,26 @@ func newVersionLabelMapping(namespace corev1.Namespace) VersionLabelMapping {
...
+
+	return versionLabelMapping
 }
 
 func mergeMappingWithDefault(labelMapping VersionLabelMapping) VersionLabelMapping {
-	// TODO: Implementation => Combine the maps
-	return defaultVersionLabelMapping
+	result := VersionLabelMapping{}
+	maps.Copy(result, defaultVersionLabelMapping)
+	maps.Copy(result, labelMapping)
```
0sewa0, line 49: does it makes sense to introduce the experimental library, which promises 0 stability to avoid minimal code like
```
func (mapping VersionLabelMapping) Copy(source VersionLabelMapping) {
	for key, value := range source {
		mapping[key] = value
	}
}
```

then the function can be:
```suggestion
	result.Copy(defaultVersionLabelMapping)
	result.Copy(labelMapping)
```

we should move to these fancy functions when they are no longer experimental, but in the meanwhile I feel its less of a headache to just not use the experimental stuff

0sewa0 (changes requested):

src/webhook/mutation/pod_mutator/oneagent_mutation/versionlabel.go
0sewa0: having a special name for `ok` can make sense if its used in a special way, 
but this is a one liner, so having that single line be shorter is preferred

TLDR: `ok` is ok here 🙃

src/webhook/mutation/pod_mutator/oneagent_mutation/containers_test.go
mjgrzybek: Awesome, simpler == better!

0sewa0 (approved):

PR #1252, opened 2022-10-18
Fixes csi-driver stuck in Terminate after restart

0sewa0:
# Description
The CSI-driver can get stuck if it mounted volumes recently and then got restarted.

This happens due to some unexpected behaviour when using the same volume for 2 volumeMounts in a container while having `mountPropagation` on both volumeMounts.

Also fixes a misleading error that happens on restart.

## How can this be tested?
Deploy a Dynakube that uses the CSI-driver, deploy sample apps. 
After they are done deploying, restart the CSI-driver,  it should restart without problem.

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/_csidriver.tpl
+++ b/config/helm/chart/default/templates/_csidriver.tpl
@@ -28,6 +28,14 @@ CSI plugin-dir path
 	{{ printf "%s/plugins/csi.oneagent.dynatrace.com/" (trimSuffix "/" (default "/var/lib/kubelet" .Values.csidriver.kubeletPath)) }}
 {{- end -}}
 
+
+{{/*
+CSI data-dir path
+*/}}
+{{- define "dynatrace-operator.CSIDataDir" -}}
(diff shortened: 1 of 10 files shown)


config/helm/chart/default/templates/_csidriver.tpl
```diff
@@ -28,6 +28,14 @@ CSI plugin-dir path
 	{{ printf "%s/plugins/csi.oneagent.dynatrace.com/" (trimSuffix "/" (default "/var/lib/kubelet" .Values.csidriver.kubeletPath)) }}
 {{- end -}}
 
+
```
luhi-DT, line 31: Should we make that even configurable? Or is appending /data to the PluginDir enough?

luhi-DT: I am wrong - you are right

luhi-DT (approved):

chrismuellner (approved):

PR #1250, opened 2022-10-17
Add operator deployment test in cloudNativeFullStack mode with running istio

aorcholski:
# Description

New CloudNative istio test is based on existing CloudNative test. 

The following configuration changes are applied to enable Istio injection 
- _istio-injection_ label is added to the application namespace
- _istio-injection_ label is added to the _dynatrace_ namespace
- _enableIstio_ property is set to true

The following additional checks are performed
- Istio service mesh is up and running
- _istio-init_ container is injected into application pods
- _istio-init_ container is injected into operator pods
- _istio virtualservice_ related to ApiUrl is created
- _istio serviceentry_ related to ApiUrl is created

Additionally
- test/e2e/cloudnative/istio target has been added

## How can this be tested?
1) Setup k8s cluster with Istio service mesh enabled.
2) Create _testdata/secrets/multi-tenant.yaml_ (see test/testdata/secrets-samples).
3) Create _testdata/secrets/single-tenant.yaml_(see test/testdata/secrets-samples).
4) Use make command to start a test.
```
make test/e2e/cloudnative/istio 
make test/e2e/cloudnative
...
```

## Checklist
- [x] PR is labeled accordingly
- [x] I have read and understood the [contribution guidelines](https://github.com/Dynatrace/dynatrace-operator/blob/master/CONTRIBUTING.md)

PR #1260, opened 2022-10-19
Adds test for data ingest

meik99:
# Description

Adds a E2E test to check if data ingest enrichment works while oneagent injection is disabled.

## How can this be tested?

* Create a cluster and connect to it
* Create a file `test/testdata/secrets/application-monitoring-install.yaml` containing a tenant secret in the following format
```
tenantUid: <tenant-id>
apiUrl: <api-url>
apiToken: <apiToken>
```
* Run ` make test/e2e/appmon`

## Checklist


luhi-DT (dismissed):
Please add a better description :)

0sewa0 (changes requested):
maybe add a `make` target so people/robots without the fancy IDEA buttons can easily run the tests 🥺 
```
## Runs Application Monitoring e2e test only
test/e2e/appmon:
	go test -v -tags e2e -count=1 -failfast ./test/appmon

```

test/appmon/appmon_test.go
```diff
@@ -0,0 +1,174 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+const (
+	installSecretPath = "../testdata/secrets/application-monitoring-install.yaml"
```
0sewa0, line 36: why do we need a special file for every test, when its always the same ?

meik99: because the secrets are not checked in and it doesn't have to be the case that they are always the same. Someone might have different tokens or tenants for different tests which requires different files.

however, that example is pretty contrived so I could also use only one file.

0sewa0: we should decide on how we are doing this, because on the https://github.com/Dynatrace/dynatrace-operator/pull/1250 we have the single file way. 

we could have special files when its needed, and we could have samples for each type of file we need.
by doing this we would not bloat the repo with unnecessary amounts of files,

hack/make/tests/e2e.mk
```diff
@@ -8,3 +8,7 @@ test/e2e/activegate:
 ## Runs CloudNative e2e test only
 test/e2e/cloudnative:
 	go test -v -tags e2e -count=1 -failfast ./test/cloudnative
+
+## Runs Application Monitoring e2e test only
+test/e2e/appmon:
```
chrismuellner, line 13: nitpick: I don't think this is the right place to shorten `applicationmonitoring` to `appmon`

test/appmon/appmon_test.go
meik99: merged the files

hack/make/tests/e2e.mk
meik99: renamed the target

0sewa0 (approved):

PR #1264, opened 2022-10-20
Fixes image verification for custom images

StefanHauth:
# Description

Verifying that custom images for OneAgent or ActiveGate are available didn't work. This has been fixed now.  Instead of re-inventing the wheel we now use `github.com/containers/image`. The switch to this library caused a lot of code to be obsolete.

Sample troubleshoot output:
```
[imagepull ]    --- Verifying that OneAgent (custom image) image quay.io/sample/custom-oa-image can be pulled ...
[imagepull ]     √  OneAgent (custom image) image quay.io/sample/custom-oa-image can be successfully pulled
[imagepull ]    --- Verifying that OneAgentCodeModules image quay.io/sample/custom-oa-code-modules:1.253.0.20220929-235140 can be pulled ...
[imagepull ]     √  OneAgentCodeModules image quay.io/sample/custom-oa-code-modules:1.253.0.20220929-235140 can be successfully pulled
[imagepull ]    --- Verifying that ActiveGate (custom image) image quay.io/sample/custom-ag-image can be pulled ...
[imagepull ]     √  ActiveGate (custom image) image quay.io/sample/custom-ag-image can be successfully pulled
```

## How can this be tested?
- Deploy a Dynakube using custom images for OneAgent and/or ActiveGate
- Start the troubleshoot command
`kubectl -n dynatrace exec pod/$(kubectl -n dynatrace get pods | grep dynatrace-operator | cut -f1 -d' ') -- /usr/local/bin/dynatrace-operator troubleshoot`
- Verify that the tests for pulling your custom images pass (given that the images really exist).

## Checklist


0sewa0 (changes requested):
the refactor looks great as it makes the code lot simpler. 👍 

however I think this will not work if there is a `proxy` set.
this is because previously to check if the image is pullable we used an `httpClient` that we configured with a proxy (inside the `troubleshootContext `)

but when using the `containers` package we don't use this `httpClient` and the best part is that you can't specify the httpClient that should be used by the package. (still open issue https://github.com/containers/image/issues/1194)

we have the same issue for the `codeModulesImage` implementation and we have plans to resolve it.

StefanHauth:
Bug has been resolved with #1277

PR #1277, opened 2022-10-25
Refactors the troubleshoot command to cleanup code

meik99:
# Description

Refactors the troubleshoot command to:
1. Remove duplicated code
2. Refactor code to make it more readable
3. Remove code that is unused due to refactorings

Also introduces some minor improvements that where made as part of refactoring the code:
* Adds a `warning` log level for log messages signifying a potential error
* Changes the character used for success messages from a square root to a checkmark
* Changes the character used for errors to a bigger `x`
* Adds green, red, and yellow color to success, error, and warning messages respectively.

## How can this be tested?

* Create different Dynakube resources. E.g.:
  * a simple dynakube
  * one with a custom pullsecret
  * one with a proxy
  * using a codemodules image
  * covering different error cases
* Deploy the operator and run the troubleshoot command
* Check that the correct state of the dynakube is determined by the command

## Checklist


src/cmd/troubleshoot/context.go
```diff
@@ -0,0 +1,41 @@
...
+		if err != nil {
+			return errorWithMessagef(err, "could not parse proxy URL!")
+		}
+
+		if troubleshootCtx.httpClient.Transport == nil {
+			troubleshootCtx.httpClient.Transport = http.DefaultTransport
+		}
+
+		troubleshootCtx.httpClient.Transport.(*http.Transport).Proxy = http.ProxyURL(proxyUrl)
+		logInfof("using  '%s' proxy to connect to the registry", proxyUrl.Host)
```
chrismuellner, line 37:
```suggestion
		logInfof("using '%s' proxy to connect to the registry", proxyUrl.Host)
```

src/cmd/troubleshoot/endpoint.go
```diff
@@ -0,0 +1,48 @@
...
+	} else {
+		logInfof("OneAgent codeModules image endpoint is not used.")
+	}
+	return imageEndpoint
+}
+
+func getActiveGateImageEndpoint(troubleshootCtx *troubleshootContext) string {
+	imageEndpoint := ""
+
+	apiEndpoint := removeSchemaRegex.FindStringSubmatch(troubleshootCtx.dynakube.Spec.APIURL)
```
chrismuellner, line 37: we should have a property/getter for api url

src/cmd/troubleshoot/image.go
```diff
@@ -98,92 +89,105 @@ func checkOneAgentCodeModulesImagePullable(troubleshootCtx *troubleshootContext)
...
+		if err != nil {
+			// Only print as a warning since other credentials might still work
+			// At this point it is uncertain if there is an error with the credentials
+			logWarningf("cannot pull image '%s' with version '%s' from registry '%s': %v",
+				componentImageInfo.image, componentImageInfo.version, registry, err)
 		} else {
-			logInfof("image '%s' with version '%s' exists on registry '%s", componentImageInfo.image, componentImageInfo.version, registry)
+			logInfof("image '%s' with version '%s' exists on registry '%s",
+				componentImageInfo.image, componentImageInfo.version, registry)
+			imageWorks = true
```
chrismuellner, line 144: couldn't we just log success and return here? no need for the `imageWorks` variable then - if the loop is completed the image was not available/found

chrismuellner: actually, change in logic here: before we would check all credentials, now we stop when the first one works. In my opinion stopping when the image is pullable is a perfectly fine approach

src/cmd/troubleshoot/image.go
```diff
@@ -192,83 +196,73 @@ func checkCustomModuleImagePullable(httpClient *http.Client, componentName strin
...
-	if statusCode, err := connectToDockerRegistry(httpClient, "HEAD", imageUrl, "Basic", apiToken); err != nil {
+	statusCode, err := connectToDockerRegistry(httpClient, imageUrl, apiToken)
+
+	if err != nil {
 		return fmt.Errorf("registry unreachable: %w", err)
-	} else {
-		if statusCode != http.StatusOK {
-			return fmt.Errorf("image not found (status code = %d)", statusCode)
-		}
+	} else if statusCode != http.StatusOK {
```
chrismuellner, line 203: no need for else, when the first `if` returns

src/cmd/troubleshoot/image.go
```diff
@@ -192,83 +196,73 @@ func checkCustomModuleImagePullable(httpClient *http.Client, componentName strin
...
+		return fmt.Errorf("image not found (status code = %d)", statusCode)
```
chrismuellner, line 204: should these `fmt.Errorf` calls be replaced with `errors.Wrap` since we are replacing them in other parts of the code?

src/cmd/troubleshoot/image.go
```diff
@@ -192,83 +196,73 @@ func checkCustomModuleImagePullable(httpClient *http.Client, componentName strin
...
-		p, err := url.Parse(proxyUrl)
-		if err != nil {
-			return errorWithMessagef(err, "could not parse proxy URL!")
-		}
-		t := troubleshootCtx.httpClient.Transport.(*http.Transport)
-		t.Proxy = http.ProxyURL(p)
-		logInfof("using  '%s' proxy to connect to the registry", p.Host)
+	if err != nil {
+		return fmt.Errorf("registry '%s' unreachable: %v", registry, err)
+	} else if statusCode != http.StatusOK {
```
chrismuellner, line 215: unnecessary else

src/cmd/troubleshoot/endpoint_test.go
```diff
@@ -0,0 +1 @@
+package troubleshoot
```
chrismuellner, line 1: should this file contain tests?

src/cmd/troubleshoot/context.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"net/http"
+	"net/url"
+
+	"github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type troubleshootContext struct {
```
chrismuellner, line 12: can we pass in the actual `context.Context` here, so it can be used instead of `context.TODO()`?

chrismuellner (changes requested):

src/cmd/troubleshoot/image.go
meik99: if i had written a second if statement, there would be no need for it, as it could be concatenated with an else. I.e. what advantage does a second if-statement have to an else branch that I should change the code that way?

src/cmd/troubleshoot/image.go
meik99: if i had written a second if statement, there would be no need for it, as it could be concatenated with an else. I.e. what advantage does a second if-statement have to an else branch that I should change the code that way?

src/cmd/troubleshoot/endpoint_test.go
meik99: yes, I forgot to move them

src/cmd/troubleshoot/image.go
```diff
@@ -139,21 +139,12 @@ func checkComponentImagePullable(httpClient *http.Client, componentName string,
...
 			break
 		}
 	}
 
-	if imageWorks {
-		logOkf("%s image '%s' found", componentName, componentImageInfo.registry+"/"+componentImageInfo.image)
-	} else {
-		// The image could not be pulled with any of the credentials
-		// Return as an error
-		return fmt.Errorf("%s image '%s' missing", componentName, componentImageInfo.registry+"/"+componentImageInfo.image)
```
chrismuellner, line 155: this error is no longer returned if the image can't be pulled

src/cmd/troubleshoot/image.go
```diff
@@ -192,83 +187,73 @@ func checkCustomModuleImagePullable(httpClient *http.Client, componentName strin
...
+		return "", fmt.Errorf("token .dockerconfigjson does not exist in secret '%s:%s'",
+			troubleshootCtx.pullSecret.Namespace, troubleshootCtx.pullSecret.Name)
 	}
 
-	secretStr := string(secretBytes)
-	return secretStr, nil
+	return string(secretBytes), nil
+}
+
+func manifestUrl(registry string, componentImageInfo imageInfo) string {
```
chrismuellner, line 248: can these helper functions be combined with https://github.com/Dynatrace/dynatrace-operator/blob/master/src/api/v1beta1/image_paths.go?

chrismuellner (changes requested):

src/cmd/troubleshoot/image.go
meik99: these helper functions depend on imageInfo which is created by parsing the image uri. Could be combined, but only if the parsing logic is also moved and incorporated into `image_paths.go`. Seems like a lot of work for functions that just put a string together

src/cmd/troubleshoot/image.go
meik99: readded

src/cmd/troubleshoot/image.go
meik99: refactored

src/api/v1beta1/properties.go
```diff
@@ -186,21 +208,65 @@ func (dk *DynaKube) PullSecret() string {
...
+
+	if apiUrlHost == "" {
+		return ""
+	}
+
+	return apiUrlHost + defaultActiveGateImage
+}
+
+func (dk *DynaKube) deprecatedActiveGateImage() string {
+	if dk.Spec.KubernetesMonitoring.Image != "" {
```
chrismuellner, line 225: if different images are set in `kubernetesMonitoring` and `routing`, the `routing` image would never be returned

hack/make/tests/e2e.mk
```diff
@@ -1,5 +1,5 @@
 ## Runs e2e tests
-test/e2e: test/e2e/cloudnative test/e2e/applicationmonitoring test/e2e/activegate
+test/e2e: manifests/branch test/e2e/cloudnative test/e2e/applicationmonitoring test/e2e/activegate
```
chrismuellner, line 2: unrelated change

src/api/v1beta1/properties.go
```diff
@@ -323,8 +389,13 @@ func (dk *DynaKube) ImmutableOneAgentImage() string {
 		tag = truncatedVersion
 	}
```
chrismuellner, line 391: we should take the chance to rename this to just `OneAgentImage()` as it's not related to the immutable image at all - different naming used for activegate image as well, `CustomActiveGateImage()`

src/cmd/troubleshoot/endpoint.go
```diff
@@ -0,0 +1,40 @@
...
+	imageEndpoint := troubleshootCtx.dynakube.ApiUrlHost() + "/linux/oneagent"
+
+	customImage := troubleshootCtx.dynakube.CustomOneAgentImage()
+	version := troubleshootCtx.dynakube.Version()
+
+	if customImage != "" {
+		imageEndpoint = customImage
+	} else if version != "" {
+		imageEndpoint = imageEndpoint + ":" + version
+	}
```
chrismuellner, line 13:
```suggestion
	if customImage != "" {
		return customImage
	}
	return imageEndpoint + ":" + version
```
wouldn't this be more explicit? log could still be done outside of this function

src/cmd/troubleshoot/image.go
meik99: changed

src/cmd/troubleshoot/endpoint.go
```diff
@@ -0,0 +1,40 @@
...
+
+	logInfof("OneAgent image endpoint '%s'", imageEndpoint)
+	return imageEndpoint
+}
+
+func getOneAgentCodeModulesImageEndpoint(troubleshootCtx *troubleshootContext) string {
+	imageEndpoint := troubleshootCtx.dynakube.CodeModulesImage()
+
+	if imageEndpoint != "" {
+		logInfof("OneAgent codeModules image endpoint '%s'", imageEndpoint)
```
chrismuellner, line 23: should we log this in a getter?

chrismuellner (changes requested):

src/api/v1beta1/properties.go
meik99: it is the same code as before but inlined. The problem is, one of the images must take precedent since we only have one image now. The kubernetes monitoring and routing parts are deprecated since v1alpha1, so it is information that is just lost in conversion

hack/make/tests/e2e.mk
meik99: but better for testing purposes since most tests apply the generated manifests

src/cmd/troubleshoot/endpoint.go
meik99: if the version is empty this would concatenate a ":" and an empty string

src/api/v1beta1/properties.go
meik99: done

src/cmd/troubleshoot/endpoint.go
meik99: the getter is `troubleshootCtx.dynakube.CodeModulesImage()`, this function wraps the logging specifically

0sewa0 (approved):

PR #1298, opened 2022-11-04
Uses proper docker manifest URL for checking codeModules images.

StefanHauth:
# Description
Instead of using the URL configured for codeModules directly, the proper docker-complient manifest URL shall be used for checking if an image is available. E.g. instead of using https://quay.io/myrepo/myimage:1.234 use https://quay.io/v2/myrepo/myimage/manifests/1.234).

## How can this be tested?
Deploy a Dynakube using codeModules and run the troubleshoot command.

## Checklist


src/cmd/troubleshoot/image.go
```diff
@@ -210,7 +210,7 @@ func imageAvailable(httpClient *http.Client, imageUrl string, apiToken string) e
 }
 
 func registryAvailable(httpClient *http.Client, registry string, apiToken string) error {
-	statusCode, err := connectToDockerRegistry(httpClient, registryUrl(registry), apiToken)
+	statusCode, err := connectToDockerRegistry(httpClient, registryUrl(registry)+"/v2/", apiToken)
```
meik99, line 213: would it make sense to have the `/v2/` included in `registryUrl`? the same string is also used in `manifestUrl`

meik99: otherwise lgtm

meik99 (approved):

PR #1301, opened 2022-11-08
Add support for checking all dynakubes in troubleshoot subcommand

gkrenn:
# Description

The troubleshoot command only checks one dynakube of which it assumes the name if no name is specified.
The new default will be that all dynakubes will be checked. 

Additional, shorthands for the troubleshoot parameters were introduced.

## How can this be tested?

Run troubleshoot subcommand without specifying a dynakube. 

## Checklist

### DIFF
--- a/src/cmd/troubleshoot/builder.go
+++ b/src/cmd/troubleshoot/builder.go
@@ -15,9 +15,11 @@ import (
 )
 
 const (
-	use               = "troubleshoot"
-	dynakubeFlagName  = "dynakube"
-	namespaceFlagName = "namespace"
+	use                    = "troubleshoot"
+	dynakubeFlagName       = "dynakube"
+	dynakubeFlagShorthand  = "d"
+	namespaceFlagName      = "namespace"
+	namespaceFlagShorthand = "n"
 )
 
 var (
@@ -54,8 +56,8 @@ func (builder CommandBuilder) Build() *cobra.Command {
 }
 
 func addFlags(cmd *cobra.Command) {
-	cmd.PersistentFlags().StringVar(&dynakubeFlagValue, dynakubeFlagName, "", "Specify a different Dynakube name.")
-	cmd.PersistentFlags().StringVar(&namespaceFlagValue, namespaceFlagName, defaultNamespace(), "Specify a different Namespace.")
+	cmd.PersistentFlags().StringVarP(&dynakubeFlagValue, dynakubeFlagName, dynakubeFlagShorthand, "", "Specify a different Dynakube name.")
+	cmd.PersistentFlags().StringVarP(&namespaceFlagValue, namespaceFlagName, namespaceFlagShorthand, defaultNamespace(), "Specify a different Namespace.")
 }
 
 func defaultNamespace() string {
--- a/src/cmd/troubleshoot/builder.go
+++ b/src/cmd/troubleshoot/builder.go
@@ -106,6 +106,7 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
 				return nil
 			}
 		}
+		resetLog()
 
 		perDynakubeTests := []troubleshootFunc{
 			checkDynakube,
@@ -127,7 +128,7 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
 				dynakube:      dynakube,
 				dynakubeName:  dynakube.Name,
 			}
-
+			logNewDynakubef(troubleshootCtx.dynakubeName)
 			for _, test := range perDynakubeTests {
 				err = test(&troubleshootCtx)
 				if err != nil {
@@ -135,6 +136,8 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
 					return nil
 				}
 			}
+			resetLog()
+			logOkf("'%s' - all checks passed", troubleshootCtx.dynakubeName)
 		}
 		return nil
 	}
@@ -144,7 +147,7 @@ func getDynakubes(troubleshootCtx *troubleshootContext) ([]dynatracev1beta1.Dyna
 	var err error
 	var dynakubes []dynatracev1beta1.DynaKube
 	if dynakubeFlagValue == "" {
-		logInfof("No Dynakube specified. Checking all Dynakubes in namespace '%s'", namespaceFlagValue)
+		logInfof("no Dynakube specified - checking all Dynakubes in namespace '%s'", namespaceFlagValue)
 		dynakubes, err = getAllDynakubesInNamespace(troubleshootCtx)
 		if err != nil {
 			return nil, err
--- a/src/cmd/troubleshoot/builder.go
+++ b/src/cmd/troubleshoot/builder.go
@@ -2,6 +2,7 @@ package troubleshoot
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"os"
 
@@ -116,7 +117,8 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
 			checkImagePullable,
 		}
 
-		dynakubes, err := getDynakubes(&troubleshootCtx)
+		troubleshootCtx.dynakubeName = dynakubeFlagValue
+		dynakubes, err := getDynakubes(troubleshootCtx)
 		if err != nil {
 			return nil
 		}
@@ -145,10 +147,10 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
 	}
 }
 
-func getDynakubes(troubleshootCtx *troubleshootContext) ([]dynatracev1beta1.DynaKube, error) {
+func getDynakubes(troubleshootCtx troubleshootContext) ([]dynatracev1beta1.DynaKube, error) {
 	var err error
 	var dynakubes []dynatracev1beta1.DynaKube
-	if dynakubeFlagValue == "" {
+	if troubleshootCtx.dynakubeName == "" {
 		logInfof("no Dynakube specified - checking all Dynakubes in namespace '%s'", namespaceFlagValue)
 		dynakubes, err = getAllDynakubesInNamespace(troubleshootCtx)
 		if err != nil {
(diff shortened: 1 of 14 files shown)


src/cmd/troubleshoot/builder.go
```diff
@@ -88,27 +92,89 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
...
+		for _, test := range dynakubeUnrelatedTests {
 			err = test(&troubleshootCtx)
 			if err != nil {
 				logErrorf(err.Error())
 				return nil
 			}
 		}
+		resetLog()
+
+		perDynakubeTests := []troubleshootFunc{
```
meik99, line 114:
```suggestion
		dynakubeTests := []troubleshootFunc{
```

src/cmd/troubleshoot/builder.go
```diff
@@ -88,27 +92,89 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
...
+			checkDynakube,
+			checkDTClusterConnection,
+			checkImagePullable,
+		}
+
+		troubleshootCtx.dynakubeName = dynakubeFlagValue
```
meik99, line 120: i think it would be better if the flag value would be passed to `getDynakubes` as a parameter instead of part of the struct. setting it here just to check it in the function and then re-setting it in the loop seems a bit odd to me

src/cmd/troubleshoot/builder.go
```diff
@@ -88,27 +92,89 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
...
+			logOkf("'%s' - all checks passed", troubleshootCtx.dynakubeName)
+		}
 		return nil
 	}
 }
+
+func getDynakubes(troubleshootCtx troubleshootContext) ([]dynatracev1beta1.DynaKube, error) {
+	var err error
+	var dynakubes []dynatracev1beta1.DynaKube
+	if troubleshootCtx.dynakubeName == "" {
```
meik99, line 153:
```suggestion
	var dynakubes []dynatracev1beta1.DynaKube
	
	if troubleshootCtx.dynakubeName == "" {
```

src/cmd/troubleshoot/builder.go
```diff
@@ -88,27 +92,89 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
...
+			logNewDynakubef(troubleshootCtx.dynakubeName)
+			for _, test := range perDynakubeTests {
+				err = test(&troubleshootCtx)
+				if err != nil {
+					logErrorf(err.Error())
+					return nil
+				}
+			}
+			resetLog()
+			logOkf("'%s' - all checks passed", troubleshootCtx.dynakubeName)
```
meik99, line 144:
```suggestion
			}
			
			logNewDynakubef(troubleshootCtx.dynakubeName)
			
			for _, test := range perDynakubeTests {
				err = test(&troubleshootCtx)
				if err != nil {
					logErrorf(err.Error())
					return nil
				}
			}
			
			resetLog()
			logOkf("'%s' - all checks passed", troubleshootCtx.dynakubeName)
```
use more negative space to group different kinds of statements

src/cmd/troubleshoot/builder.go
```diff
@@ -88,27 +92,89 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
...
+	if err != nil {
+		logErrorf("failed to list Dynakubes: %v", err)
+		return nil, err
+	}
+	if len(dynakubes.Items) == 0 {
+		err = fmt.Errorf("no Dynakubes found in namespace '%s'", troubleshootContext.namespaceName)
+		logErrorf(err.Error())
+		return nil, err
+	}
+	return dynakubes.Items, nil
```
meik99, line 179:
```suggestion
	query := kubeobjects.NewDynakubeQuery(troubleshootContext.apiReader, troubleshootContext.namespaceName).WithContext(troubleshootContext.context)
	dynakubes, err := query.List()
	
	if err != nil {
		logErrorf("failed to list Dynakubes: %v", err)
		return nil, err
	}
	
	if len(dynakubes.Items) == 0 {
		err = fmt.Errorf("no Dynakubes found in namespace '%s'", troubleshootContext.namespaceName)
		logErrorf(err.Error())
		return nil, err
	}
	
	return dynakubes.Items, nil
```

src/cmd/troubleshoot/context.go
```diff
@@ -14,8 +14,8 @@ type troubleshootContext struct {
 	context            context.Context
 	apiReader          client.Reader
 	httpClient         *http.Client
-	namespaceName      string // the default namespace ("dynatrace") or provided in the command line
-	dynakubeName       string // the default name of dynakube ("dynakube") or provided in the command line
+	namespaceName      string // the default namespace ("dynatrace") or value provided in the command line
+	dynakubeName       string // all dynakubes in namespaceName or value provided in the command line
```
meik99, line 18:
```suggestion
```
if the dynakube flag is passed as a parameter, I think this field is not needed anymore. I would avoid having a dynakube instance and a dynakube name both in a struct when the dynakube instance contains the same information.

Thinking about it, the namespace field has the same problem, as the name of the namespace is also stored in the dynakube instance

src/cmd/troubleshoot/logger.go
```diff
@@ -23,26 +23,31 @@ const (
...
 )
 
 type troubleshootLogger struct {
-	logger logr.Logger
+	logger  logr.Logger
+	subTest bool
 }
 
-func newTroubleshootLogger(testName string) logr.Logger {
+func newTroubleshootLogger(testName string, subTest bool) logr.Logger {
```
meik99, line 38: I'd prefer having multiple constructors instead of passing a flag.
I.e.:
```
func newTroubleshootLogger(testName string) logr.Logger {
  return logger
}
func newSubtestTroubleshootLogger(testName string) logr.Logger {
  logger = newTroubleshootLogger(testName)
  logger.subTest = true
  return logger
}
```

meik99: instead of using a flag to change the behaviour, using inheritence, or "embedding" in go, would be better as it upholds the open-closed principle

src/cmd/troubleshoot/logger.go
```diff
@@ -75,20 +84,30 @@ func errorWithMessagef(err error, format string, v ...interface{}) error {
 func (dtl troubleshootLogger) Init(_ logr.RuntimeInfo) {}
 
 func (dtl troubleshootLogger) Info(level int, message string, keysAndValues ...interface{}) {
+	var msg string
+
 	switch level {
 	case levelNewTest:
-		dtl.logger.Info(prefixNewTest+message, keysAndValues...)
+		msg = prefixNewTest + message
```
meik99, line 91:
```suggestion
		message = prefixNewTest + message
```
message can be changed directly without the use of a variable

src/cmd/troubleshoot/builder.go
```diff
@@ -88,27 +92,89 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
...
+	if troubleshootCtx.dynakubeName == "" {
+		logNewDynakubef("no Dynakube specified - checking all Dynakubes in namespace '%s'", namespaceFlagValue)
+		dynakubes, err = getAllDynakubesInNamespace(troubleshootCtx)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		dynakube := dynatracev1beta1.DynaKube{}
+		dynakube.Name = troubleshootCtx.dynakubeName
+		dynakubes = append(dynakubes, dynakube)
```
meik99, line 162: the actual dynakube could be queried and used here since the list operation also queries the actual ones

meik99 (changes requested):

src/cmd/troubleshoot/context.go
gkrenn: The dynakube name is still used by the `checkDynakube` test, to get the actual dynakube. - Will move the `checkDynakubeCrdExists` from the "per dynakube" - tests to the one time tests. => then we can remove the dynakubename from the struct, agree.

src/cmd/troubleshoot/builder.go
gkrenn: Will do

src/cmd/troubleshoot/logger.go
StefanHauth: Although it doesn't have any side-effects here, I'm not a friend of re-using parameters as local variables. It changes the semantic from "input" to "variable" and if later on some decides to change the signature to using a pointer (admittedly not very likely here), it can lead to odd and hard to spot side-effects.

src/cmd/troubleshoot/builder.go
StefanHauth: And if its no longer used here, couldn't dynakubeName then be removed from the troubleshootContext? Once the troubleshootContext.dynakube is set, it's redundant anyway.

src/cmd/troubleshoot/logger.go
meik99: If a string is passed to a function as a pointer with the intention of propagating this mutation that would raise at least one red flag in my mind.

src/cmd/troubleshoot/builder.go
gkrenn: Do not agree - the list operation will not fail as we do not request a specific dynakube. 

In the case of querying one single dynakube it could happen that the dynakube does not exists, which is checked in the dynakube subtests

src/cmd/troubleshoot/logger.go
gkrenn: In general I agree with Stefan, would not re-use parameters. 

For the current lines it does not matter, as the msg will be returned directly in my new changes.

src/cmd/troubleshoot/builder.go
meik99: True, I'd still pass the dynakubeName flag as a parameter instead of having it in the context so the field can be removed

gkrenn: Already moved 👍

gkrenn:
Closed due to too much changes in master branch, will reopen a new one

Issue #1289, opened 2022-11-02
hostMonitoring is consuming too much licenses


olljanat:
It looks to be that `hostMonitoring` mode does not currently include `--set-infra-only=true` flag for pods so they consume licenses same way as full stack when only infra mode licenses should be used:
https://www.dynatrace.com/support/help/monitoring-consumption/application-and-infrastructure-monitoring

Parameters on pod deployed by this controller:
![image](https://user-images.githubusercontent.com/6213926/199426605-e3ad436c-452f-48fe-a223-1793096b2975.png)

same node in Dynatrace:
![image](https://user-images.githubusercontent.com/6213926/199426899-de9271e6-bd37-4791-85ae-28c4d4455036.png)

luhi-DT:
Hi, please reach out to the Dynatrace Support regarding your issue: https://support.dynatrace.com/ Thanks!

olljanat:
@luhi-DT Why? It is clearly operator which sets those parameters:
https://github.com/Dynatrace/dynatrace-operator/blob/225c6a61ff727a9712b15b203953e71e89b3c0bb/src/controllers/dynakube/oneagent/daemonset/arguments.go#L58
https://github.com/Dynatrace/dynatrace-operator/blob/225c6a61ff727a9712b15b203953e71e89b3c0bb/src/controllers/dynakube/oneagent/daemonset/daemonset.go#L39

so why `--set-infra-only` would be anyway different?

PR #1304, opened 2022-11-09
Update CONTRIBUTING.md

meik99:
# Description

These changes reflect the discussion of https://github.com/Dynatrace/dynatrace-operator/pull/1300#discussion_r1016528015 This PR should be place to openly discuss this proposal to formalize a common style.

## How can this be tested?

N.A.

## Checklist


meik99:
I am going to leave this open so everyone has a chance to read and discuss it. Especially @chrismuellner @luhi-DT @0sewa0 @StefanHauth

0sewa0 (approved):

olljanat:
> I am going to leave this open until everyone has had a chance to read and discuss it.

Not directly related to this update but to CONTRIBUTING.md in general that it should be clarified that how and which kind of issues are handled in here and which need to go through Dynatrace support? I have still hard time accept that #1289 was closes without proper handling even it clearly is bug/missing feature in this operator it selves. Not in any other parts of Dynatrace stack.

luhi-DT:
@olljanat 
The CONTRIBUTING.md is about contributing to the Operators repository and not about how the contributors handle external issues.
The problem with your issue is that this is not decided and covered by our department. We are developing the Dynatrace Operator. However that does not mean we are responsible also for licensing topics. As this "issue" is the way how it was designed, the only thing we can say from our end is, please go to the official Dynatrace support. 

Also the support is here to help our customers and to also be the first line of support, if we as Operator devs would be reachable by every customer directly and help the with all of their issues, we would have a hard time to implement new features.

I hope that is understandable.
Thanks!

CONTRIBUTING.md
```diff
@@ -16,6 +16,40 @@ For those just getting started, consult this  [guide](https://help.github.com/ar
 - Do not create methods with more than two parameters (in extremely rare occasions maybe three) except constructors and factory functions. Structs and interfaces exist for a reason.
 - Avoid returning responses (e.g., reconcile.Result, admission.Patched) in anything but Reconcile or Handle functions.
 
+### Cuddling of statements
+Statements must be cuddled, i.e., written as a single block, if an `if`-statement directly follows a single assignment and the condition is directly related to the assignment.
+This commonly occurs with error handling, but is not restricted to it.  
+Example:
+```
+err := assignment1()
```
StefanHauth, line 24: I'd actually even prefer
```suggestion
if err := assignment1(); err != {
}
```
where applicable.

CONTRIBUTING.md
```diff
@@ -16,6 +16,40 @@ For those just getting started, consult this  [guide](https://help.github.com/ar
 - Do not create methods with more than two parameters (in extremely rare occasions maybe three) except constructors and factory functions. Structs and interfaces exist for a reason.
 - Avoid returning responses (e.g., reconcile.Result, admission.Patched) in anything but Reconcile or Handle functions.
 
+### Cuddling of statements
+Statements must be cuddled, i.e., written as a single block, if an `if`-statement directly follows a single assignment and the condition is directly related to the assignment.
```
StefanHauth, line 20: I'd leave some freedom here and just put this as a suggestion.
```suggestion
Statements should be cuddled, i.e., written as a single block, if an `if`-statement directly follows a single assignment and the condition is directly related to the assignment.
```

luhi-DT (approved):

CONTRIBUTING.md
meik99: Personally I always feel like that makes conditions harder to read. Especially if a function name is longer, but even with shorter ones I feel like that is the case. Having the assignment be separated from the condition makes it clearer what is happening in the condition.

Going a bit more into extreme cases, if the assignment is merged with the condition and the condition has more parts than just checking the error it becomes even more unwieldy. Here an example from the production code:
```
err = dtpullsecret.
		NewReconciler(ctx, controller.client, controller.apiReader, controller.scheme, dynakube, tokens).
		Reconcile()
	if err != nil {
```
That would be an extreme case, but merged it would look like so:
```
if err := dtpullsecret.NewReconciler(ctx, controller.client, controller.apiReader, controller.scheme, dynakube, tokens).Reconcile(); err != nil {
```

PR #1311, opened 2022-11-10
Set --set-infra-only=true flag in hostMonitoring mode

### LINKED ISSUE #1289: hostMonitoring is consuming too much licenses

olljanat:
For me it sounds that there is some misunderstanding what is really need needed to fix #1289. Especially based on comment https://github.com/Dynatrace/dynatrace-operator/pull/1304#issuecomment-1309906205 There is no anything which needs decission but this operator need set OneAgent parameters like it is documented. Same way like it is done when deploying OneAgent without this operator to standalone Docker nodes, etc. So I decided to create pull request about.

# Description
Set `--set-infra-only=true` flag in hostMonitoring mode to correctly consume host units like they are explained in "Infrastructure" column in this table https://www.dynatrace.com/support/help/monitoring-consumption/application-and-infrastructure-monitoring#host-units

Fixes: #1289

## How can this be tested?
After this change it is visible in Dynatrace portal that host "Monitoring mode" is **Infrastructure only** instead of **Full stack**
image

## Checklist

### DIFF
--- a/src/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/src/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -38,6 +38,8 @@ const (
 
 	inframonHostIdSource = "--set-host-id-source=k8s-node-name"
 	classicHostIdSource  = "--set-host-id-source=auto"
+
+	inframonInfraOnly = "--set-infra-only=true"
 )
 
 type HostMonitoring struct {
@@ -102,6 +104,7 @@ func (dsInfo *HostMonitoring) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 
 	if len(result.Spec.Template.Spec.Containers) > 0 {
 		appendHostIdArgument(result, inframonHostIdSource)
+		appendHostIdArgument(result, inframonInfraOnly)
 		dsInfo.appendInfraMonEnvVars(result)
 	}
 


luhi-DT:
Hi @olljanat, thanks for the PR. However the operator team understood clearly what you wanted to address in your issue and comments.Nevertheless the licensing topic for host-monitoring is currently being discussed internally. This means for now the operator works as expected and does not set the `--set-infra-only=true` flag. So this works as expected and as designed. However there may be changes to this in the future, depending on the outcome of different discussion not related to the operator itself. Whenever there are changes to this behaviour our team will take care to make the necessary changes and change it to `--set-infra-only=true`.

Thanks!
Lukas

olljanat:
@StevenCane unfortunately there looks to be some weird policy inside of Dynatrace, etc which prevents this to be merged. So I recommend you to create ticket for them.

In mean time you also use my custom image `ollijanatuinen/dynatrace-operator:fix-infra-mode` in https://github.com/Dynatrace/dynatrace-operator/blob/69009ef8db7f4c31fea2febc55ed632647a31df4/config/helm/chart/default/templates/Common/operator/deployment-operator.yaml#L56 (probably it can be also provided with values file but I didn't tested that) which have this modification included.

PR #1314, opened 2022-11-10
Let troubleshooting command continue execution until unrecoverable error is found

mjgrzybek:
# Description
Currently, troubleshoot command executes series of `tests` on operator deployment. If one of them fails, `troubleshoot` exists early reporting first problem found.

Changes in scope of this PR:
- renames `test` to `check` to avoid confusion with testing code
- `checks` execution is moved to extracted function `runChecks` [1]
- `Check` struct is used to specify check including prerequisites required to run it
- `error`s wrapping is fixed (no double printing or pseudo-json structure)
- `runChecks` returns `AggregatedError`. It's implementation is troubleshoot-specific, shouldn't be used elsewhere in this form. Meanwhile, waiting for https://github.com/golang/go/issues/53435 🥺

[1]
`Check` struct contains prerequisites in form of a slice of strings containing pointers to other checks (directed acyclic graph).
If any of prerequisites failed or was skipped, check will be skipped.

## How can this be tested?
Run `k exec deploy/dynatrace-operator -n dynatrace -- dynatrace-operator troubleshoot` on your environment.
Try deleting e.g. some secrets to reproduce scenario with multiple problems existence.

Example output:
```
kubectl exec deploy/dynatrace-operator -n dynatrace -- dynatrace-operator troubleshoot --dynakube dk
{"level":"info","ts":"2022-11-15T14:20:56.184Z","logger":"version","msg":"dynatrace-operator","version":"snapshot-feature-continue-troubleshooting-on-non-crit-error","gitCommit":"a86866db8264690eb4694e9525675331f09bf67d","buildDate":"2022-11-15T14:15:27+00:00","goVersion":"go1.19.3","platform":"linux/amd64"}
[namespace ]    --- checking if namespace 'dynatrace' exists ...
[namespace ]     ✓  using namespace 'dynatrace'
[dynakube  ]    --- checking if 'dynatrace:dk' Dynakube is configured correctly
[dynakube  ]        CRD for Dynakube exists
[dynakube  ]        using 'dynatrace:dk' Dynakube
[dynakube  ]        checking if api url is valid
[dynakube  ]        api url is valid
[dynakube  ]     X  'dynatrace:dk' secret is missing: secrets "dk" not found
[dynakube  ]     ⚠  Skipped 'dynatraceApiSecretHasApiToken' check because prerequisites aren't met: [apiSecret]
[dynakube  ]     X  'dynatrace:dk-pull-secret' pull secret is missing: secrets "dk-pull-secret" not found
[dynakube  ]     ⚠  Skipped 'pullSecretHasRequiredTokens' check because prerequisites aren't met: [pullSecretExists]
[dynakube  ]        no proxy is configured
[dynakube  ]     X  'dynatrace:dk' Dynakube isn't valid. Target namespace and dynakube can be changed by providing '--namespace <namespace>' or '--dynakube <dynakube>' parameters.: 'dynatrace:dk' secret is missing: secrets "dk" not found
'dynatrace:dk-pull-secret' pull secret is missing: secrets "dk-pull-secret" not found
[dynakube  ]     ⚠  Skipped 'DynatraceClusterConnection' check because prerequisites aren't met: [dynakube]
[dynakube  ]     ⚠  Skipped 'imagePullable' check because prerequisites aren't met: [dynakube]
```

## Checklist


src/cmd/troubleshoot/context.go
```diff
@@ -24,11 +25,17 @@ type troubleshootContext struct {
 
 type troubleshootFunc func(troubleshootCtx *troubleshootContext) error
 
+type Check struct {
```
0sewa0, line 28: minor nitpick: This could be in the `check_executor.go`, which IMO is an overcomplicated name for a file. 
It could be just `checks.go` and have the `Check` struct near the where its run would be nicer to read

0sewa0 (changes requested):

src/cmd/troubleshoot/context.go
StefanHauth: I like `check_executor.go` because it's what this file is about. `checks.go` is a very generic name, where I would expect stuff that's doing checking of some sort. However, I agree that the `Check` struct should be in that file as well. They belong together.

src/cmd/troubleshoot/checks_executor.go
```diff
@@ -0,0 +1,48 @@
...
+
+const (
+	PASSED = iota + 1
+	FAILED
+	SKIPPED
+)
+
+var results = map[string]Result{}
+
+func runChecks(troubleshootCtx *troubleshootContext, checks []Check) error {
```
StefanHauth, line 17: This function should also take care of the execution order so that all pre-requisites are executed before their dependences.

src/cmd/troubleshoot/checks_executor.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	if errs.Empty() {
+		return nil
+	}
+
+	return errs
+}
+
+func shouldSkip(check Check) bool {
```
StefanHauth, line 41: Reverse invers logic and improve naming
```suggestion
func prerequisitesHavePassed(check Check) bool {
```

0sewa0: IMO `shouldSkip` is fine as there can be more reason to skip (in the future maybe?) 
also `prerequisitesPassed` is enough

src/cmd/troubleshoot/checks_executor.go
0sewa0: that would be overkill logic vs maintaining a reasonable list of `Checks` that are in the correct order.

having the checks in the correct order in code would be mandatory IMO to keep the code readable anyway.

mjgrzybek: Good catch, Stefan, It's directed acyclic graph, so traveling the graph can be done smarter.
I also think it's an overkill for the moment. We can improve it later.

src/cmd/troubleshoot/context.go
meik99: I agree with @0sewa0 here, just that it isn't a nitpick for me. `_executor` is redundant as mentioned, but also composite names are usually an inidication that files should be moved into a package. This is not the case here, therefore it should only be called `checks.go`

src/cmd/troubleshoot/dynakube.go
```diff
@@ -59,31 +96,37 @@ func checkDynakubeCrdExists(troubleshootCtx *troubleshootContext) error {
...
+		if runtime.IsNotRegisteredError(err) {
+			err = errors.Wrap(err, "CRD for Dynakube missing")
+		} else {
+			err = errors.Wrap(err, "could not list Dynakube")
+		}
+		return err
 	}
 
 	logInfof("CRD for Dynakube exists")
 	return nil
```
meik99, line 109: unwrapping `if`s for better readbility
```suggestion
    if err == nil {    
        logInfof("CRD for Dynakube exists")
    	return nil
    }
	
	if runtime.IsNotRegisteredError(err) {
		err = errors.Wrap(err, "CRD for Dynakube missing")
	} else {
		err = errors.Wrap(err, "could not list Dynakube")
	}
	
	return err
```

src/cmd/troubleshoot/dynakube.go
```diff
@@ -59,31 +96,37 @@ func checkDynakubeCrdExists(troubleshootCtx *troubleshootContext) error {
...
+			err = errors.Wrapf(err, "could not get Dynakube '%s:%s'",
+				troubleshootCtx.namespaceName, troubleshootCtx.dynakubeName)
+		}
+		return err
 	}
 
+	troubleshootCtx.dynakube = dynakube
+
 	logInfof("using '%s:%s' Dynakube", troubleshootCtx.namespaceName, troubleshootCtx.dynakubeName)
 	return nil
```
meik99, line 131:
```suggestion
    if err == nil {
 	    troubleshootCtx.dynakube = dynakube

    	logInfof("using '%s:%s' Dynakube", troubleshootCtx.namespaceName, troubleshootCtx.dynakubeName)
	    return nil
	}

	if k8serrors.IsNotFound(err) {
		err = errors.Wrapf(err,
	    	"selected '%s:%s' Dynakube does not exist",
			troubleshootCtx.namespaceName, troubleshootCtx.dynakubeName)
	} else {
		err = errors.Wrapf(err, "could not get Dynakube '%s:%s'",
			troubleshootCtx.namespaceName, troubleshootCtx.dynakubeName)
	}
		
	return err
```

src/cmd/troubleshoot/errors/aggregated.go
```diff
@@ -0,0 +1,31 @@
...
+func NewAggregatedError() AggregatedError {
+	return AggregatedError{Errs: []error{}}
+}
+
+func (e AggregatedError) Error() string {
+	sb := strings.Builder{}
+	for _, err := range e.Errs {
+		sb.WriteString(err.Error() + "\n")
+	}
+	return sb.String()
```
meik99, line 24: I'll allow `sb` as a name as long as there is some whitespace to make it more readable
```suggestion
	sb := strings.Builder{}
	
	for _, err := range e.Errs {
		sb.WriteString(err.Error() + "\n")
	}
	
	return sb.String()
```

src/cmd/troubleshoot/checks_executor.go
```diff
@@ -0,0 +1,48 @@
...
+
+type Result int
+
+const (
+	PASSED = iota + 1
+	FAILED
+	SKIPPED
+)
+
+var results = map[string]Result{}
```
meik99, line 15: global `map` is prime real estate for memory leaks, refactor either into parameter or a struct. Maybe a something like a `CheckSuit` struct is viable containing the array of checks as well as any result. `runChecks` could be a method of that.

src/cmd/troubleshoot/checks_executor.go
```diff
@@ -0,0 +1,48 @@
...
+
+func runChecks(troubleshootCtx *troubleshootContext, checks []Check) error {
+	errs := tserrors.NewAggregatedError()
+	for _, check := range checks {
+		if shouldSkip(check) {
+			results[check.Name] = SKIPPED
+			continue
+		}
+
+		if err := check.Do(troubleshootCtx); err != nil {
```
meik99, line 25:
```suggestion
       err := check.Do(troubleshootCtx)
		if err != nil {
```

src/cmd/troubleshoot/errors/aggregated.go
```diff
@@ -0,0 +1,31 @@
+package errors
+
+import (
+	"strings"
+)
+
+type AggregatedError struct {
```
meik99, line 7: nitpick: could be 
```suggestion
type AggregatedError []error
```
if the `Add` function is not absolutely necessary

meik99 (changes requested):

src/cmd/troubleshoot/checks_executor.go
meik99: In the calling function `shouldSkip` is better understandable.
`if shouldSkip(check)` is more straight forward compared to `if !prerequisitesPassed(check)`

src/cmd/troubleshoot/errors/aggregated.go
mjgrzybek: It won't work because `[]error` doesn't satisfy `error` interface thus cannot be returned as an `error`

src/cmd/troubleshoot/dynakube.go
mjgrzybek: Typically I'd agree but reverting idiomatic `if err != nil` to `if err == nil` may lead to problems because it's easy to overlook inverted logic.

src/cmd/troubleshoot/checks_executor.go
mjgrzybek: It's not an issue here because of `trobuleshoot` command nature - single run of sequential checks with single `results` instance in the end. 
I didn't want it over-engineered.

src/cmd/troubleshoot/context.go
StefanHauth: If I see a file called `checks.go` I'm wondering, what the checks in there are checking. My first thought would not be that it contains a framework for running checks in some kind of sophisticated manner. Also the plural suggest that there is a number of checks for whatever in this file. If at all it should be called `check.go`, because it contains the implementation of the type `Check`. Maybe a different name for the type would be better (`Checker`, `CheckerTask`, `Verifier`, ...)

mjgrzybek: `*er` names are generally used for interfaces names.

src/cmd/troubleshoot/checks_executor.go
StefanHauth: Do it right or not at all. The way how it's currently implemented it might happen that checks are not executed if they are configured in the wrong order, which is a potential pitfall for future devs that come new to the project. IMHO that's a bug which needs to be resolved.

src/cmd/troubleshoot/errors/aggregated.go
StefanHauth: I have to say, I find the wording ("I will allow") here quite pretentious. And I don't think that the empty lines here add much for readability. They just bloat up the code and disconnect parts from each other that belong together.

src/cmd/troubleshoot/checks_executor.go
meik99: It is done right at the moment, since building a graph and resolving it would be overkill and therefore wrong since a solution that is over-complicated for the intended purpose is wrong by definition. We can think about that when we have, like, 50 checks, but with the amount of checks we have now a list is more maintainable

src/cmd/troubleshoot/dynakube.go
meik99: then I'd suggest refactoring the inner code so its not that deep, e.g.:
```suggestion
	if err != nil {
		return determineDynakubeError(err)
	}

	logInfof("CRD for Dynakube exists")
	return nil
	...
}

func determineDynakubeError(err error) error {
       if runtime.IsNotRegisteredError(err) {
	    return errors.Wrap(err, "CRD for Dynakube missing")
        } else {
	    return errors.Wrap(err, "could not list Dynakube")
	}
}	
```

src/cmd/troubleshoot/errors/aggregated.go
meik99: it was meant as a joke since I am usually rather strict concerning variable naming, I guess that doesn't translate well when written.

src/cmd/troubleshoot/checks_executor.go
meik99: as far as I can tell it's populated in `runChecks` and read in `skipCheck`. defining and instantiating it in the former and passing it to the latter doesn't seem over-engineered to me

src/cmd/troubleshoot/errors/aggregated.go
mjgrzybek: I agree with Stefan here, this method does literally one thing and there is no need to separate its logic with empty lines.

src/cmd/troubleshoot/builder.go
```diff
@@ -88,27 +93,47 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
...
+		Do: func(troubleshootCtx *troubleshootContext) error {
+			return checkDtClusterConnection(results, troubleshootCtx)
+		},
+		Prerequisites: []*Check{namespaceCheck, dynakubeCheck},
+	}
+	imagePullableCheck := &Check{
+		checkImagePullable,
+		[]*Check{namespaceCheck, dynakubeCheck},
+		imagePullableCheckName,
+	}
```
meik99, line 135:
```suggestion
	imagePullableCheck := &Check{
		Name: imagePullableCheckName,
		Do: checkImagePullable,
		Prerequisites: []*Check{namespaceCheck, dynakubeCheck},
	}
```

src/cmd/troubleshoot/checks.go
```diff
@@ -0,0 +1,85 @@
...
+
+type Check struct {
+	Do            troubleshootFunc
+	Prerequisites []*Check
+	Name          string
+}
+
+type ChecksResults struct {
+	check2Result map[*Check]Result
+}
```
meik99, line 28: If it has only one field, could this be 
```suggestion
type ChecksResults map[*Check]Result
```
instead?

src/cmd/troubleshoot/checks.go
```diff
@@ -0,0 +1,85 @@
...
+
+func NewChecksResults() ChecksResults {
+	return ChecksResults{check2Result: map[*Check]Result{}}
+}
+
+func (cr ChecksResults) set(check *Check, result Result) {
```
meik99, line 34:
```suggestion
func (checkResults ChecksResults) set(check *Check, result Result) {
```

src/cmd/troubleshoot/checks.go
```diff
@@ -0,0 +1,85 @@
...
+	cr.check2Result[check] = result
+}
+
+func (cr ChecksResults) failedPrerequisites(check *Check) []*Check {
+	isFailed := func(c *Check) bool {
+		return cr.check2Result[c] == FAILED
```
meik99, line 40:
```suggestion
	isFailed := func(check *Check) bool {
		return cr.check2Result[check] == FAILED
```

src/cmd/troubleshoot/checks.go
```diff
@@ -0,0 +1,85 @@
...
+	failedPrerequisites := results.failedPrerequisites(check)
+
+	if len(failedPrerequisites) == 0 {
+		return false
+	}
+
+	getCheckName := func(c *Check) string {
+		return c.Name
+	}
+	prereqsNames := strings.Join(functional.Map(failedPrerequisites, getCheckName), ",")
```
meik99, line 81:
```suggestion
	prerequisiteNames := strings.Join(functional.Map(failedPrerequisites, getCheckName), ",")
```

src/cmd/troubleshoot/errors/aggregated.go
meik99: Just to clarify, I meant like leaving the receiver functions as is, i.e., `func (e AggregatedError) Error() string {` would still exist to fulfill the interface.
But again, only a nitpick, can also just stay as is

src/functional/map.go
```diff
@@ -0,0 +1,9 @@
+package functional
+
+func Map[In any, Out any](arr []In, transformFn func(it In) Out) []Out {
+	ret := make([]Out, len(arr))
+	for i, it := range arr {
+		ret[i] = transformFn(it)
+	}
+	return ret
+}
```
meik99, line 9:
```suggestion
func Map[In any, Out any](array []In, transformFn func(input In) Out) []Out {
	mappedResult := make([]Out, len(array))
	
	for i, element := range array {
		mappedResult[i] = transformFn(element)
	}
	
	return mappedResult
}
```

src/functional/filter.go
```diff
@@ -0,0 +1,11 @@
...
+
+func Filter[T any](arr []T, filterFn func(val T) bool) []T {
+	ret := make([]T, 0)
+	for _, val := range arr {
+		if filterFn(val) {
+			ret = append(ret, val)
+		}
+	}
+	return ret
+}
```
meik99, line 11:
```suggestion
func Filter[T any](array []T, predicate func(value T) bool) []T {
	filteredResult := make([]T, 0)
	
	for _, element := range array {
		if predicate(element) {
			filteredResult = append(filteredResult, element)
		}
	}
	
	return filteredResult
}
```

meik99 (changes requested):

src/functional/map.go
meik99: Somehow I f'ed up the intendation, please disregard the w i d e tabs :sweat_smile:

src/functional/filter.go
meik99: Somehow I f'ed up the intendation, please disregard the w i d e tabs :sweat_smile:

src/functional/map.go
```diff
@@ -0,0 +1,11 @@
+package functional
+
+func Map[In any, Out any](arr []In, transformFn func(it In) Out) []Out {
```
meik99, line 3: my changes for this and `Filter` were more about the parameter / variable names and less about the whitespace

meik99 (changes requested):

meik99 (approved):

PR #1319, opened 2022-11-11
Removes https proxy in samples

0sewa0:
# Description

Having the https in the proxy examples is probably a mistake and confusing, as 
HTTPS proxies are rare and not widely supported and don't really make sense for a kubernetes environment.

## How can this be tested?
Read 🙃 

## Checklist

### DIFF
--- a/assets/samples/applicationMonitoring.yaml
+++ b/assets/samples/applicationMonitoring.yaml
@@ -27,7 +27,7 @@ spec:
   # Optional: Set custom proxy settings either directly or from a secret with the field 'proxy'
   #
   # proxy:
-  #   value: https://my-proxy-url.com
+  #   value: http://my-proxy-url.com
   #   valueFrom: name-of-my-proxy-secret
 
   # Optional: Adds custom RootCAs from a configmap
--- a/assets/samples/applicationMonitoring.yaml
+++ b/assets/samples/applicationMonitoring.yaml
@@ -27,7 +27,7 @@ spec:
   # Optional: Set custom proxy settings either directly or from a secret with the field 'proxy'
   #
   # proxy:
-  #   value: http://my-proxy-url.com
+  #   value: my-proxy-url.com
   #   valueFrom: name-of-my-proxy-secret
 
   # Optional: Adds custom RootCAs from a configmap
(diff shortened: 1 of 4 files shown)


StefanHauth (approved):

assets/samples/applicationMonitoring.yaml
```diff
@@ -27,7 +27,7 @@ spec:
   # Optional: Set custom proxy settings either directly or from a secret with the field 'proxy'
   #
   # proxy:
-  #   value: https://my-proxy-url.com
+  #   value: http://my-proxy-url.com
```
mjgrzybek, line 30:
```suggestion
  #   value: my-proxy-url.com
```

assets/samples/classicFullStack.yaml
```diff
@@ -29,7 +29,7 @@ spec:
   # Optional: Set custom proxy settings either directly or from a secret with the field 'proxy'
   #
   # proxy:
-  #   value: https://my-proxy-url.com
+  #   value: http://my-proxy-url.com
```
mjgrzybek, line 32:
```suggestion
  #   value: my-proxy-url.com
```

assets/samples/cloudNativeFullStack.yaml
```diff
@@ -30,7 +30,7 @@ spec:
   # Optional: Set custom proxy settings either directly or from a secret with the field 'proxy'
   #
   # proxy:
-  #   value: https://my-proxy-url.com
+  #   value: http://my-proxy-url.com
```
mjgrzybek, line 33:
```suggestion
  #   value: my-proxy-url.com
```

assets/samples/hostMonitoring.yaml
```diff
@@ -25,7 +25,7 @@ spec:
   # Optional: Set custom proxy settings either directly or from a secret with the field 'proxy'
   #
   # proxy:
-  #   value: https://my-proxy-url.com
+  #   value: http://my-proxy-url.com
```
mjgrzybek, line 28:
```suggestion
  #   value: my-proxy-url.com
```

mjgrzybek (dismissed):
We support only `http` proxy, so we should just remove schema from example value not to give an impression that it's configurable.
Also some may think that "`http` == not safe" and replace it with `https` without understanding how proxy works.

luhi-DT (approved):

PR #1321, opened 2022-11-14
Adds logic for configurable unmounted volume gc

0sewa0:
# Description

The log gc in the csi driver is too complicated and not very effective.
As it only deletes folders IF there are (too many log files OR too big is the sum of the log files) AND more 2 weeks old.
This logic is not very helpful when it comes to short living pods, as the folders created by the CSI driver is counted towards the memory usage of the CSI driver.

So now we only care about the age of the folders.
Furthermore now the max age can be configured via a envvar on the provisioner container. (name of envvar: `MAX_UNMOUNTED_VOLUME_AGE`)
Or in the values.yaml incase of Helm.

Some refactor:
- Log garbage collection was misleading, as it did not gc logs per-say, it gc-ed unmounted volumes in case there were too many logs.
  -  Renamed to "unmounted volumes garbage collection", so we say what it actually does.

Why not a feature-flag?
There could be multiple dynakubes with the same tenantUUID affecting the same node. 
- To be able to handle the `MAX_UNMOUNTED_VOLUME_AGE` per dynakube would result in major code modifications. (+ extra confusion from the user side)

## How can this be tested?
Set the `MAX_UNMOUNTED_VOLUME_AGE` on the provisioner container to some value that suits you, check if unmounted volumes are removed on the node after the given time.

The value of `MAX_UNMOUNTED_VOLUME_AGE` can be whatever you `time.ParseDuration` can handle.

## Checklist


luhi-DT:
wouldnt it make more sense to make MAX_UNMOUNTED_VOLUME_AGE as a flag on the dynakube instead of an env_var?

Then it could be configurable for all the dynakubes differently

0sewa0:
@luhi-DT 
> Why not a feature-flag? There could be multiple dynakubes with the same tenantUUID affecting the same node.
> 
> * To be able to handle the `MAX_UNMOUNTED_VOLUME_AGE` per dynakube would result in major code modifications. (+ extra confusion from the user side)

its in the description 🙃 
> Why not a feature-flag?
There could be multiple dynakubes with the same tenantUUID affecting the same node.

>To be able to handle the MAX_UNMOUNTED_VOLUME_AGE per dynakube would result in major code modifications. (+ extra confusion from the user side)

luhi-DT:
Makes sense 👍 thanks and sorry for the oversight

config/helm/chart/default/values.yaml
```diff
@@ -52,6 +52,7 @@ csidriver:
   nodeSelector: {}
   kubeletPath: "/var/lib/kubelet"
   priorityClassValue: "1000000"
+  maxUnmountedVolumeAge: "" # if not set, default is 336h
```
meik99, line 55: nitpick
```suggestion
  maxUnmountedVolumeAge: "" # if not set, default is 336h or 14 days
```

src/controllers/csi/gc/unmounted.go
```diff
@@ -0,0 +1,76 @@
...
+	"context"
+	"os"
+	"time"
+
+	"github.com/spf13/afero"
+)
+
+const (
+	defaultMaxUnmountedCsiVolumeAge = 14 * 24 * time.Hour
+	MaxUnmountedCsiVolumeAgeEnv     = "MAX_UNMOUNTED_VOLUME_AGE"
```
meik99, line 13: doesn't appear to be used in another package
```suggestion
	maxUnmountedCsiVolumeAgeEnv     = "MAX_UNMOUNTED_VOLUME_AGE"
```

src/controllers/csi/gc/unmounted.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+func (gc *CSIGarbageCollector) runUnmountedVolumeGarbageCollection(ctx context.Context, tenantUUID string) {
+	unmountedVolumes, err := gc.getUnmountedVolumes(tenantUUID)
+	if err != nil {
+		log.Info("failed to get unmounted volume information")
+		return
+	}
+	gc.removeUnmountedVolumesIfNecessary(unmountedVolumes, tenantUUID)
```
meik99, line 22:
```suggestion
	}
	
	gc.removeUnmountedVolumesIfNecessary(unmountedVolumes, tenantUUID)
```

src/controllers/csi/gc/unmounted.go
```diff
@@ -0,0 +1,76 @@
...
+func (gc *CSIGarbageCollector) getUnmountedVolumes(tenantUUID string) ([]os.FileInfo, error) {
+	var unusedVolumeIDs []os.FileInfo
+
+	volumeIDs, err := afero.ReadDir(gc.fs, gc.path.AgentRunDir(tenantUUID))
+	if err != nil {
+		return nil, err
+	}
+
+	for _, volumeID := range volumeIDs {
+		unused, err := afero.IsEmpty(gc.fs, gc.path.OverlayMappedDir(tenantUUID, volumeID.Name()))
```
meik99, line 34:
```suggestion
		isUnused, err := afero.IsEmpty(gc.fs, gc.path.OverlayMappedDir(tenantUUID, volumeID.Name()))
```

src/controllers/csi/gc/unmounted.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+
+	return unusedVolumeIDs, nil
+}
+
+func (gc *CSIGarbageCollector) removeUnmountedVolumesIfNecessary(unusedVolumeIDs []os.FileInfo, tenantUUID string) {
+	for _, unusedVolumeID := range unusedVolumeIDs {
+		if gc.isUnmountedVolumeTooOld(unusedVolumeID.ModTime()) {
+			if err := gc.fs.RemoveAll(gc.path.AgentRunDirForVolume(tenantUUID, unusedVolumeID.Name())); err != nil {
```
meik99, line 50:
```suggestion
            err := gc.fs.RemoveAll(gc.path.AgentRunDirForVolume(tenantUUID, unusedVolumeID.Name()))
			if err != nil {
```

src/controllers/csi/gc/unmounted.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+}
+
+func (gc *CSIGarbageCollector) isUnmountedVolumeTooOld(t time.Time) bool {
+	return gc.maxUnmountedVolumeAge == 0 || time.Since(t) > gc.maxUnmountedVolumeAge
+}
+
+func determineMaxUnmountedVolumeAge() time.Duration {
+	maxAgeEnv := os.Getenv(MaxUnmountedCsiVolumeAgeEnv)
```
meik99, line 62: imo env var access should be abstracted to a function so `os.Getenv` is only used at one place, but then again, that is mitigated by using a const so consider it a nitpick

meik99 (changes requested):

luhi-DT (approved):

src/controllers/csi/gc/unmounted.go
```diff
@@ -32,12 +33,12 @@ func (gc *CSIGarbageCollector) getUnmountedVolumes(tenantUUID string) ([]os.File
 	}
 
 	for _, volumeID := range volumeIDs {
-		unused, err := afero.IsEmpty(gc.fs, gc.path.OverlayMappedDir(tenantUUID, volumeID.Name()))
+		isUnused, err := afero.IsEmpty(gc.fs, gc.path.OverlayMappedDir(tenantUUID, volumeID.Name()))
```
waodim, line 36: a little nitpick but do we want to use negated names for booleans? i'm thinking of the case !isUnused, in my opinion isUsed would be more readable. but as I mentioned only a thought.

meik99 (approved):

src/controllers/csi/gc/unmounted_test.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+	})
+
+}
+
+func TestDetermineMaxUnmountedVolumeAge(t *testing.T) {
+
+	t.Run("no env set ==> use default", func(t *testing.T) {
+		os.Setenv(maxUnmountedCsiVolumeAgeEnv, "")
+		defer os.Unsetenv(maxUnmountedCsiVolumeAgeEnv)
```
meik99, line 139: I've overlooked that in my earlier review, please refactor setting the env var in the constructor so it can be mocked. having to set and unset env vars in a unit test doesn't sit right with me

meik99, line 139: I've overlooked that in my earlier review, please refactor setting the env var in the constructor so it can be mocked. having to set and unset env vars in a unit test doesn't sit right with me

meik99 (changes requested):

src/controllers/csi/gc/unmounted.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+		}
+	}
+}
+
+func (gc *CSIGarbageCollector) isUnmountedVolumeTooOld(t time.Time) bool {
+	return gc.maxUnmountedVolumeAge == 0 || time.Since(t) > gc.maxUnmountedVolumeAge
+}
+
+func determineMaxUnmountedVolumeAge() time.Duration {
```
meik99, line 64:
```suggestion
func determineMaxUnmountedVolumeAge(maxUnmountedCsiVolumeAgeEnvValue string) time.Duration {
```

src/controllers/csi/gc/reconciler.go
```diff
@@ -35,17 +36,20 @@ type CSIGarbageCollector struct {
...
 	return &CSIGarbageCollector{
-		apiReader: apiReader,
-		fs:        afero.NewOsFs(),
-		db:        db,
-		path:      metadata.PathResolver{RootDir: opts.RootDir},
+		apiReader:             apiReader,
+		fs:                    afero.NewOsFs(),
+		db:                    db,
+		path:                  metadata.PathResolver{RootDir: opts.RootDir},
+		maxUnmountedVolumeAge: determineMaxUnmountedVolumeAge(),
```
meik99, line 52:
```suggestion
		maxUnmountedVolumeAge: determineMaxUnmountedVolumeAge(os.Getenv(maxUnmountedCsiVolumeAgeEnv)),
```

meik99 (approved):

PR #1331, opened 2022-11-17
Fixes an issue where the ActiveGate's statefulset is deleted and recreated if labeled manually

meik99:
# Description

The activegate's default labels where changed some time ago. The activegate reconciler checks the labels of the deployed activegate against the desired one to update the statefulset, i one with old labels is deployed. This caused the activegate statefulset to be deleted and re-created if it was labeled by a third party, e.g. some webhook that automatically labels new resources. These changes modify that logic so only the old labels are checked and the statefulset is only deleted if they specifically exist.

## How can this be tested?

* Deploy the DTO and dynakube with AG enabled
* Wait until the statefulset is deployed
* Label the statefulset manually
* Make sure the statefulset is not deleted
* Change the activegate section of the dynakube
* Make sure the statefulset is still recreated correctly
* Label the statefulset manually
* Make sure the statefulset is not deleted

## Checklist


src/controllers/dynakube/activegate/internal/statefulset/reconciler.go
```diff
@@ -171,6 +171,23 @@ func (r *Reconciler) deleteStatefulSetIfOldLabelsAreUsed(desiredSts *appsv1.Stat
 	return false, nil
 }
 
+func isUsingOldLabels(currentStatefulSet *appsv1.StatefulSet) bool {
```
0sewa0, line 174: shouldn't we check the version label if it changed?
we store the version of the activegate in the labels, and if there is a new version, label changes, get redeployed
was this even expected behaviour ?

meik99: if the version changed, that would affect the calculated hash value so we already detect that change.
this check specifically targets the old values of the statefulset since the labels are also used as match labels which are immutable. In that case we can't update the sts but have to delete it.

however, this also means that, if the version changes the label changes as well.
this has yet to be reflected by this PR, meaning the old as well as the new labels have to be checked, however for the new ones the value has to be checked as well

meik99: changed it to simply checking the selector instead of the labels

0sewa0 (approved):

PR #1343, opened 2022-11-18
Adds support-archive sub-command incl. tarball and operator version c…

StefanHauth:
# Description
Adds the basic functionality for collecting support-archives from the operator components. This first part consists of
* the resp. sub command
* functionality to create the support-archive tarball
* collecting of the operator version in a text file into the support-archive

The support archive can retrieved either by
* copying it from the pod, the necessary command is printed on screen
* using the `--stdout` commandline switch to stream the tarball directly to a local file via stdout

Collection of additional items (logs, K8S objects) will be added in future PRs in the upcoming weeks.

**Note**: logging of the operator version to stdout has been duplicated and moved away from the common main so it doesn't interfere with writing the tarball to stdout.

## How can this be tested?
execute the following command:
`kubectl -n dynatrace exec pod/$(kubectl -n dynatrace get pods | grep dynatrace-operator | cut -f1 -d' ') -- /usr/local/bin/dynatrace-operator support-archive`

## Checklist


0sewa0:
pls don't use `feature request` label on PRs as those are meant for Issues

src/cmd/support_archive/logger.go
```diff
@@ -0,0 +1,42 @@
...
+func registerSupportArchiveLogger(ctx *supportArchiveContext) {
+	if ctx.toStdout {
+		// using stderr because we use stdout to deliver the tarball
+		ctx.log = newSupportArchiveLoggerToWriter(supportArchiveLoggerName, os.Stderr)
+	} else {
+		ctx.log = newSupportArchiveLoggerToWriter(supportArchiveLoggerName, os.Stdout)
+	}
+}
+
+func newSupportArchiveLoggerToWriter(name string, out io.Writer) logr.Logger {
```
0sewa0, line 27: isn't the `name` param redundant as always the same value is passed into it?  (except in the tests)

0sewa0, line 27: `newSupportArchiveLoggerToWriter ` this name is reeally confusing, it kinda tells me that I should give it a Logger and I will get a Writer, however it will create a new Logger that is using the provided Writer.

IMO better options: 
`newSupportArchiveLoggerWithWriter`
`newSupportArchiveLogger` as the `Writer` is already part of the params so writing it into the name is a bit redundant.

(if you want more wrapping, but it would just overcomplicate things)
`newSupportArchiveLoggerToStdout`
`newSupportArchiveLoggerToStderr`

src/cmd/support_archive/operator_version.go
```diff
@@ -0,0 +1,25 @@
...
+func collectOperatorVersion(ctx *supportArchiveContext, tarball *tarball) error {
+	logInfof(ctx.log, "Storing operator version into %s", versionFileName)
+
+	versionString := fmt.Sprintf("version: %s\ngitCommit: %s\nbuildDate: %s\ngoVersion %s\nplatform %s/%s\n",
+		version.Version,
+		version.Commit,
+		version.BuildDate,
+		runtime.Version(),
+		runtime.GOOS, runtime.GOARCH)
+	tarball.addFile(versionFileName, strings.NewReader(versionString))
```
0sewa0, line 22: why are you not handling the error ?

src/cmd/support_archive/tarball.go
```diff
@@ -0,0 +1,96 @@
...
+
+	header := &tar.Header{
+		Name: fileName,
+		Size: int64(buffer.Len()),
+		Mode: int64(fs.ModePerm),
+	}
+
+	err := t.tarWriter.WriteHeader(header)
+	if err != nil {
+		return fmt.Errorf("could not write header for file '%s', got error '%w'", fileName, err)
```
0sewa0, line 59: pls use `errors` package (`errors.WithStack` or `errors.WithMessagef` or something like that :) )

src/cmd/support_archive/tarball.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	_, err = io.Copy(t.tarWriter, buffer)
+	if err != nil {
+		return fmt.Errorf("could not copy the file '%s' data to the tarball, got error '%w'", fileName, err)
```
0sewa0, line 64: pls use `errors` package (`errors.WithStack` or `errors.WithMessagef` or something like that :) )

src/cmd/support_archive/tarball.go
```diff
@@ -0,0 +1,96 @@
...
+		tarballFilePath = fmt.Sprintf(tarFileName, defaultTargetDir, time.Now().Format(time.RFC3339))
+	} else {
+		tarballFilePath = fmt.Sprintf(tarFileName, ctx.tarballTargetDir, time.Now().Format(time.RFC3339))
+	}
+
+	tarballFilePath = strings.Replace(tarballFilePath, ":", "_", -1)
+
+	tarFile, err := os.Create(tarballFilePath)
+	if err != nil {
+		return nil, err
```
0sewa0, line 93: pls use `errors.WithStack(err)`

src/cmd/support_archive/builder.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	for _, c := range collectors {
+		if err := c(ctx, supportArchive); err != nil {
+			logErrorf(ctx.log, err, "failed collector")
+		}
+	}
+}
+
+func printCopyCommand(ctx *supportArchiveContext, supportArchive *tarball) {
```
0sewa0, line 87: why not a method of `supportArchiveContext` ?

then you could "mock" the `os.Getenv` to some `ctx.GetPodNamespace` and `ctx.GetPodName` or something so it can be tested

src/cmd/support_archive/builder.go
```diff
@@ -0,0 +1,95 @@
...
+		defer supportArchive.close()
+
+		runCollectors(&ctx, supportArchive)
+		printCopyCommand(&ctx, supportArchive)
+
+		return nil
+	}
+}
+
+func runCollectors(ctx *supportArchiveContext, supportArchive *tarball) {
```
0sewa0, line 75: why not a method of `supportArchiveContext` ?

src/cmd/support_archive/tarball.go
```diff
@@ -0,0 +1,96 @@
...
+const tarFileName = "%s/operator-support-archive-%s.tgz"
+const defaultTargetDir = "/tmp/dynatrace-operator"
+
+type tarball struct {
+	tarWriter  *tar.Writer
+	gzipWriter *gzip.Writer
+	tarFile    *os.File
+}
+
+func newTarball(ctx *supportArchiveContext) (*tarball, error) {
```
0sewa0, line 24: you are just using 2 things from the context, a bool and a string,

IMO this struct could be independent, and not be bound to the `supportArchiveContext`, so just pass in only the things that are necessary, nicer read (if not more then 3 params are used),nicer to test

0sewa0 (changes requested):
I made several comments about the `supportArchiveContext` struct's and the lack of method usage + its (IMO) weird relation with the `tarball` struct 

if we have this "big" struct containing 90% of the state, but not the last 10% feels kinda weird. 
Wouldn't it make sense for the `supportArchiveContext` to "have a" `tarball`?

src/cmd/support_archive/tarball.go
```diff
@@ -0,0 +1,96 @@
...
+		t.gzipWriter.Close()
+	}
+	if t.tarFile != nil {
+		t.tarFile.Close()
+	}
+}
+
+func (t *tarball) addFile(fileName string, file io.Reader) error {
+	buffer := &bytes.Buffer{}
+	io.Copy(buffer, file)
```
luhi-DT, line 49: please add error handling here

src/version/version.go
```diff
@@ -26,6 +27,10 @@ var (
 
 // LogVersion logs metadata about the Operator.
 func LogVersion() {
+	LogVersionToLogger(log)
```
luhi-DT, line 30: Imho it would be better to leave the call of LogVersion in the main.go as it is called only once for all subcommands there and add an additional function call for the supportArchive command, with that we dont have to maintain the functionCall in 5 different places but in one.

src/cmd/support_archive/logger.go
StefanHauth: Exeactly, it's for testability.

src/cmd/support_archive/tarball.go
```diff
@@ -0,0 +1,96 @@
...
+	} else {
+		tarFile, err := createTarFile(ctx)
+		if err != nil {
+			return nil, err
+		}
+		return tarFile, nil
+	}
+}
+
+func createTarFile(ctx *supportArchiveContext) (*os.File, error) {
```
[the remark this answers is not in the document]
StefanHauth: Because it's not related to the context.

src/cmd/support_archive/logger.go
StefanHauth: Yeah, true it's a left-over from when `registerSupportArchiveLogger` still was called `newSupportArchiveLogger` and I needed a distinction. Now I would go with `newSupportArchiveLoggerWithWriter` except we don't need the addendum any more, so `newSupportArchiveLogger` it'll be.

0sewa0: but what are you testing ?
that you can pass in a string and use it in another constructor ? (a string which is always the same constant in real use)

this code is just a copy-paste from the troubleshooting package, where this param makes sense as there are different "things" we test in the troubleshoot script and the logs should reflect it.(crd, secret, dynakube, ect.) 

but here its always the same

src/cmd/support_archive/builder.go
StefanHauth: Again, printing this command isn't a task of the context, it just holds information. You can't put everything that uses the context into the context, because then you end up with a big mess of unrelated functions. That's the nature of a context, that it's used in different occasions, that doesn't imply it has to own all the stuff it is used for.

src/cmd/support_archive/builder.go
StefanHauth: Same as above.

src/cmd/support_archive/tarball.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	_, err = io.Copy(t.tarWriter, buffer)
+	if err != nil {
+		return fmt.Errorf("could not copy the file '%s' data to the tarball, got error '%w'", fileName, err)
+	}
+	return nil
+}
+
+func selectAndCreateTargetFile(ctx *supportArchiveContext) (*os.File, error) {
```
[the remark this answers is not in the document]
0sewa0: then the whole connection between the taball and supportArchiveContext struct is poorly designed

if its connected to the tarball then why its no where really mentioned ?

src/cmd/support_archive/tarball.go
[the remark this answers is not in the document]
0sewa0: then the whole connection between the taball and supportArchiveContext struct is poorly designed

wouldn't it make more sense that you have a `tarball` struct that has a `createFile` method ??

src/cmd/support_archive/tarball.go
StefanHauth: Agreed, although I find this 3-param rule a bit random. And in a way it prevents consistency of the function signatures.

src/cmd/support_archive/tarball.go
StefanHauth: Uupsi.

src/cmd/support_archive/logger.go
```diff
@@ -0,0 +1,42 @@
...
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	ctrlzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+const (
+	supportArchiveLoggerName = "[support-archive]"
+)
+
+func registerSupportArchiveLogger(ctx *supportArchiveContext) {
```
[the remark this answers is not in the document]
StefanHauth: Yeah, it's kind of related to the logger and to the context half/half. And as the method has some logging related decisions in it, I'd rather leave it in the logger file than moving it to the context file.

src/cmd/support_archive/tarball.go
meik99: constructors and factory functions are exempt from that rule

src/cmd/support_archive/builder.go
0sewa0: passing around a big struct that has a name that barley means anything into simple function can be reeally bad, because if the simple function uses 1 (maybe 2) fields out of the 6 fields of the struct then 
- its annoying to test (which 2 fields is needed again out of the 6 so I don't get a nil pointer?) 
- horrible to read the code (what is the function actually using and doing with the struct? this is further worsened by the fact you pass in the struct every where as a reference so anything can modify it which is misleading at best and dangerous at worst )

src/cmd/support_archive/context.go
```diff
@@ -0,0 +1,15 @@
+package support_archive
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+)
+
+type supportArchiveContext struct {
```
meik99, line 9: i was already skeptical with the `troubleshootContext`, but now seeing the `supportArchiveContext` as well as all the comments on this PR make me think that they are a symptom of bad architecture. 

They are, much like a `utils` package, just sinks for all the bits and bops of this functionality. That leads to them doing too much and being too interconnected with other structs. I think both should be refactored with single-responsibility in mind.

src/cmd/support_archive/logger.go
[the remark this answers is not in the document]
0sewa0: but don't you store the logger in the struct ? so to me that is veeery related to the context

src/cmd/support_archive/tarball.go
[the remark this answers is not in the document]
StefanHauth: It's in `tarball.go`, what more do you need?

src/cmd/support_archive/builder.go
StefanHauth: You're right about the pointers, will check that.
The one thing about the struct is also, that all collector functions (we know there will be a few more in the future) have a common signature. Each of those are then using what they need from the context (there will be a few more fields like API reader, http client, ...). So we have a few options:
* pass all fields to all collector functions. That's equally confusing because not all functions will use all fields AND it will break the 3 param max rule, e.g.
```
type collectorFunc func(ctx context.Context, log logr.Logger, apiReader client.Reader, clientSet kubernetes.Interface, namespace string, tb *tarball)
```
* don't use a common signature, which will de-loop `runCollectors` and quadruple the error handling code there.
```
err := collectOperatorVersion(log, tarball)
if err != nil {
   logErrorf(ctx.log, err, "failed collector")
}

err = collectOperatorLogs(ctx, log, clientSet, namespace, tarball)
if err != nil {
   logErrorf(ctx.log, err, "failed collector")
}

err = collectK8SObjects(ctx, log, apiReader, tarball)
if err != nil {
   logErrorf(ctx.log, err, "failed collector")
}

...
```



So for the collector functions, I think the context is still the best solution.

src/cmd/support_archive/tarball.go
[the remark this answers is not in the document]
StefanHauth: I chose to use a free function, because it doesn't use the tarball per-se, it's a utility function for the tarball.

src/cmd/support_archive/logger.go
StefanHauth: Yeah, might be overengineered... I liked to change the logger name in the test, but I guess it's not strictly necessary.

src/cmd/support_archive/builder.go
```diff
@@ -0,0 +1,114 @@
...
+		// avoid corrupting tarball
+		return os.Stderr
+	} else {
+		return os.Stdout
+	}
+}
+
+func runCollectors(log logr.Logger, supportArchive tarball) {
+	collectors := []collector{
+		operatorVersionCollector{
```
0sewa0, line 84: nitpick: Having a constructor for `operatorVersionCollector ` would make the code nicer (even if its a simple one)

0sewa0 (approved):

PR #1349, opened 2022-11-22
Combine changes of old branch with master

gkrenn:
# Description

The troubleshoot command only checks one dynakube of which it assumes the name if no name is specified.
The new default will be that all dynakubes will be checked. 

Additional, shorthands for the troubleshoot parameters were introduced.

## How can this be tested?

Run troubleshoot subcommand without specifying a dynakube. 

## Checklist

### DIFF
--- a/src/cmd/troubleshoot/builder.go
+++ b/src/cmd/troubleshoot/builder.go
@@ -2,23 +2,29 @@ package troubleshoot
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"os"
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	"github.com/Dynatrace/dynatrace-operator/src/cmd/config"
+	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
 	"github.com/Dynatrace/dynatrace-operator/src/scheme"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/cluster"
 )
 
 const (
-	use               = "troubleshoot"
-	dynakubeFlagName  = "dynakube"
-	namespaceFlagName = "namespace"
+	use                    = "troubleshoot"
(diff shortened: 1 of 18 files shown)


src/cmd/troubleshoot/builder.go
```diff
@@ -95,49 +101,122 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
...
+
+		_ = runChecks(results, &troubleshootCtx, getDynakubeSpecificChecks(results, checks)) // ignore error to avoid polluting pretty logs
+		resetLogger()
+		if !results.hasErrors() {
+			logOkf("'%s' - all checks passed", dynakube.Name)
+		}
+	}
+}
+
+func getDynakubeUnrelatedChecks() []*Check {
```
0sewa0, line 147:
```suggestion
func getPrerequisiteChecks() []*Check {
```

crd is very related to the Dynakubes 😅

src/cmd/troubleshoot/builder.go
```diff
@@ -95,49 +101,122 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
...
 		results := NewChecksResults()
+		_ = runChecks(results, &troubleshootCtx, getDynakubeUnrelatedChecks()) // ignore error to avoid polluting pretty logs
+		resetLogger()
+
+		dynakubes, err := getDynakubes(troubleshootCtx, dynakubeFlagValue)
+		if err != nil {
+			return nil
+		}
+
+		runChecksForAllDynakubes(results, getDynakubeSpecificChecks(results, getDynakubeUnrelatedChecks()), dynakubes, apiReader)
```
0sewa0, line 120: you are using the `getDynakubeUnrelatedChecks ` as `globalPrerequisites` in the `runChecksForAllDynakubes`

shouldn't we handle the error of `runChecks(results, &troubleshootCtx, getDynakubeUnrelatedChecks())` ? 
if something is wrong with the `globalPrerequisites`, then it does not make much sense to try to run any other tests

0sewa0 (changes requested):
I want to complain about the `troubleshootContext` a lot, but I did not, because its not in the scope of this PR 😅

0sewa0 (approved):

PR #1361, opened 2022-11-25
Adds cloudNative regression test with agentVersion set

waodim:
# Description

This test should verify that the operator is able to set the agentVersion which is given via the dynakube. Also an upgrade to a newer version should be possible

## How can this be tested?

Simply create a new cluster and run the cloudNative regression tests.

## Checklist


test/cloudnative/agentversion_set.go
```diff
@@ -0,0 +1,191 @@
...
+		time.Sleep(time.Minute)
+		resources := environmentConfig.Client().Resources()
+		pods := sampleapps.Get(t, ctx, resources)
+
+		for _, podItem := range pods.Items {
+			require.NotNil(t, podItem)
+			require.NotNil(t, podItem.Spec)
+			require.NotNil(t, podItem.Spec.InitContainers)
+
+			for _, container := range podItem.Spec.InitContainers {
```
mjgrzybek, line 176: There is at most one `"install-oneagent"`, so instead of looping, it should be found first (using e.g. `slices.Index`) and than processed.

mjgrzybek (changes requested):

test/cloudnative/agentversion_set.go
```diff
@@ -0,0 +1,191 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+const VERSION = "VERSION"
```
mjgrzybek, line 38:
```suggestion
const version = "VERSION"
```

test/cloudnative/agentversion_set.go
```diff
@@ -0,0 +1,191 @@
...
+	builder.Assess("start sample apps and injection", sampleapps.Restart)
+	builder.Assess("check init containers", checkInitContainers)
+	builder.Assess("check env vars of init container", checkVersionInSampleApp(version))
+}
+
+func assignVersions(t *testing.T, versions []string,
+	old version.SemanticVersion, new version.SemanticVersion) (version.SemanticVersion, version.SemanticVersion) {
+
+	lowerHalf := versions[0 : len(versions)/2]
+	upperHalf := versions[len(versions)/2+1 : len(versions)-1]
```
mjgrzybek, line 104: If I understand correctly, this algorithm wont work for `len(versions) < 2`

test/cloudnative/agentversion_set.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func assessVersionChecks(builder *features.FeatureBuilder, version version.SemanticVersion) {
+	builder.Assess("restart csi driver", restartCSIDriver)
+	builder.Assess("start sample apps and injection", sampleapps.Restart)
+	builder.Assess("check init containers", checkInitContainers)
+	builder.Assess("check env vars of init container", checkVersionInSampleApp(version))
+}
+
+func assignVersions(t *testing.T, versions []string,
```
mjgrzybek, line 100: Please comment on what `assignVersions` does

mjgrzybek (changes requested):

test/cloudnative/agentversion_set.go
waodim: you are right,thank you for pointing out! :)

test/cloudnative/agentversion_set.go
```diff
@@ -0,0 +1,183 @@
...
+	oldVersion, newVersion := assignVersions(t, versions, version.SemanticVersion{}, version.SemanticVersion{})
+
+	dk := dynakube.NewBuilder().
+		WithDefaultObjectMeta().
+		WithDynakubeNamespaceSelector().
+		ApiUrl(secretConfig.ApiUrl).
+		CloudNativeWithAgentVersion(&dynatracev1beta1.CloudNativeFullStackSpec{}, oldVersion).
+		Build()
+
+	agentVersionset := features.New("cloudnative with agent version set")
```
0sewa0, line 55: nitpick: `specificAgentVersion` maybe is clearer, because when I read `agentVersionset` I thought for a second that its a set of agent versions 😅 

```suggestion
	specificAgentVersion := features.New("cloudnative with specific agent version set")
```

test/cloudnative/agentversion_set.go
```diff
@@ -0,0 +1,183 @@
...
+		dk.Status.UpdatedTimestamp = metav1.Now()
+		dk.Spec.OneAgent.CloudNativeFullStack.Version = semanticVersion.String()
+		err = resources.Update(ctx, &dk)
+		require.NoError(t, err)
+
+		return ctx
+	}
+}
+
+func restartCSIDriver(ctx context.Context, t *testing.T, environmentConfig *envconf.Config) context.Context {
```
0sewa0, line 139: why restart the csi driver?

test/cloudnative/agentversion_set.go
```diff
@@ -0,0 +1,183 @@
...
+				require.NotNil(t, initContainer.Env)
+				require.True(t, kubeobjects.EnvVarIsIn(initContainer.Env, agentVersion))
+				for _, env := range initContainer.Env {
+					if env.Name == agentVersion {
+						require.NotNil(t, env.Value)
+						assert.Equal(t, semanticVersion.String(), env.Value)
+					}
+				}
+			}
+		}
```
0sewa0, line 180: well, this is deeep :D 

probably could be refactored to be more readable, by moving things to helper functions 😉

0sewa0 (changes requested):

test/cloudnative/agentversion_set.go
waodim: without this the injection does not take place. i am not sure why, but when i apply a dynakube with a specific version i see that the csi pods do restart (i thought maybe this is because they already have the latest version and when i apply a dynakube with a different version it comes to problems thats why i decided to restart it simply once after applying the dynakube and then it works fine)

waodim: if you have another, better way to solve this, i would happily change this

0sewa0: that sounds like something that should not happen 😅

waodim: alright i'll think of ways to do this better. thanks

mjgrzybek (approved):

PR #1363, opened 2022-11-25
Add support-archive E2E test.

StefanHauth:
# Description

Add E2E tests for the support-archive subcommand recently added in #1343.

## How can this be tested?
`make test/e2e/supportarchive`

## Checklist

### DIFF
--- a/test/bash/bash.go
+++ b/test/bash/bash.go
@@ -1,7 +1,6 @@
 package bash
 
 import (
-	"fmt"
 	"strings"
 )
 
@@ -11,7 +10,7 @@ func DiskUsageWithTotal(directory string) Command {
 	return Command{"du", "-c", directory}
 }
 
-func FilterLastLineOnly() []string {
+func FilterLastLineOnly() Command {
 	return Command{"tail", "-n", "1"}
 }
 
@@ -35,11 +34,7 @@ func Shell(command Command) Command {
 }
 
 func (c Command) String() string {
-	cmdString := ""
-	for _, str := range c {
-		cmdString = fmt.Sprintf("%s %s", cmdString, str)
-	}
-	return strings.Trim(cmdString, " ")
+	return strings.Join(c, " ")
 }
 
 func Echo(msg string) Command {
--- a/test/bash/bash.go
+++ b/test/bash/bash.go
@@ -41,3 +41,7 @@ func (c Command) String() string {
 	}
 	return strings.Trim(cmdString, " ")
 }
+
+func Echo(msg string) Command {
+	return Command{"echo", msg}
+}
--- a/test/bash/bash.go
+++ b/test/bash/bash.go
@@ -5,36 +5,36 @@ import (
 	"strings"
 )
 
-type BashCmd []string
+type Command []string
 
-func DiskUsageWithTotal(directory string) BashCmd {
-	return BashCmd{"du", "-c", directory}
+func DiskUsageWithTotal(directory string) Command {
+	return Command{"du", "-c", directory}
 }
 
 func FilterLastLineOnly() []string {
-	return BashCmd{"tail", "-n", "1"}
+	return Command{"tail", "-n", "1"}
 }
 
-func Pipe(source []string, sink []string) BashCmd {
+func Pipe(source []string, sink []string) Command {
 	piped := source
 	piped = append(piped, "|")
 	piped = append(piped, sink...)
 	return piped
 }
 
-func ReadFile(path string) BashCmd {
-	return BashCmd{"cat", path}
+func ReadFile(path string) Command {
+	return Command{"cat", path}
 }
 
-func ListDirectory(path string) BashCmd {
-	return BashCmd{"ls", path}
+func ListDirectory(path string) Command {
+	return Command{"ls", path}
 }
 
-func Shell(command BashCmd) BashCmd {
-	return BashCmd{"sh", "-c", command.String()}
+func Shell(command Command) Command {
+	return Command{"sh", "-c", command.String()}
 }
 
-func (c BashCmd) String() string {
+func (c Command) String() string {
 	cmdString := ""
 	for _, str := range c {
 		cmdString = fmt.Sprintf("%s %s", cmdString, str)
--- a/test/bash/bash.go
+++ b/test/bash/bash.go
@@ -1,23 +1,43 @@
 package bash
 
-import "fmt"
+import (
+	"fmt"
+	"strings"
+)
 
-func DiskUsageWithTotal(directory string) string {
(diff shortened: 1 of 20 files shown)


test/appmon/appmon_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/test/kubeobjects/environment"
 	"github.com/Dynatrace/dynatrace-operator/test/kubeobjects/namespace"
 	"github.com/Dynatrace/dynatrace-operator/test/sampleapps"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
mjgrzybek, line 12: It needs to be imported because of side effects (which is registering gcp auth plugin).
It should be done in single top-level place instead of in specific tests file.

test/kubeobjects/pod/execute.go
```diff
@@ -25,15 +26,31 @@ type ExecutionResult struct {
...
 		pod:       pod,
 		command:   command,
 		container: container,
+		arguments: make([]string, 0),
+		tty:       false,
 	}
+
+	for _, arg := range arguments {
+		parts := strings.Split(arg, " ")
+		query.arguments = append(query.arguments, parts...)
```
mjgrzybek, line 45: What is argument has (escaped) space in it?
Arguments should be passed as they were provided, splitting is not expected by caller imho.

mjgrzybek (changes requested):

test/support_archive/support_archive_test.go
```diff
@@ -0,0 +1,120 @@
...
+	supportArchiveExecution := features.New("support archive execution")
+	setup.InstallOperator(supportArchiveExecution, secretConfig)
+	setup.AssessOperatorDeployment(supportArchiveExecution)
+	supportArchiveExecution.Assess("support archive subcommand can be executed correctly", checkSupportArchiveExecution())
+	return supportArchiveExecution.Feature()
+}
+
+func checkSupportArchiveExecution() features.Func {
+	return func(ctx context.Context, t *testing.T, environmentConfig *envconf.Config) context.Context {
+		result := executeSupportArchive(context.TODO(), t, environmentConfig, "--stdout")
```
mjgrzybek, line 71:
```suggestion
		result := executeSupportArchive(context.Background(), t, environmentConfig, "--stdout")
```

test/support_archive/support_archive_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+func executeSupportArchive(ctx context.Context, t *testing.T, environmentConfig *envconf.Config, cmdLineArguments string) *pod.ExecutionResult {
+	resources := environmentConfig.Client().Resources()
+	pods := operator.Get(t, ctx, resources)
+	for _, podItem := range pods.Items {
+		require.NotNil(t, podItem)
+
+		if strings.Contains(podItem.Name, "dynatrace-operator") {
```
mjgrzybek, line 97: Nit: `functional.Filter(pod.Items, dynatraceOperatorMatch)` could be used before the loop.
As a result, loop's body is not if-nested.

test/support_archive/support_archive_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+var testEnvironment env.Environment
+
+func TestMain(m *testing.M) {
+	workingDir, err := os.Getwd()
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	//	require.NoError(m, err)
```
waodim, line 39: maybe you forgot this line to remove

waodim, line 39: i think you forgot to remove this line

test/support_archive/support_archive_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	fmt.Println(workingDir)
```
waodim, line 41: is this for testing purposes or is this line needed?

test/support_archive/support_archive_test.go
StefanHauth: the whole first paragraph is a left over :-)

test/bash/bash.go
```diff
@@ -1,23 +1,47 @@
 package bash
```
mjgrzybek, line 1: This package contains tools to run processes but not necessarily in bash (or any shell). Following unix standards, I'd name it `exec`.
```suggestion
package exec
```

StefanHauth: `exec` is kind of misleading, as non of the functions in there actually executes anything, it's just wrappers for string constructions for (shell) commands.

mjgrzybek: Ok, `shell` is better than `bash`.
But commands don't need to be executed in bash, subprocess can just fork from current process and there is no shell underneath. This is why I'd suggest non-shell package name. How about `commands`?

test/bash/bash.go
```diff
@@ -1,23 +1,47 @@
...
 
-func ReadFile(path string) string {
-	return fmt.Sprintf("cat %s", path)
+func (c Command) String() string {
+	cmdString := ""
+	for _, str := range c {
+		cmdString = fmt.Sprintf("%s %s", cmdString, str)
+	}
+	return strings.Trim(cmdString, " ")
 }
```
mjgrzybek, line 43: I think it gives same result
```suggestion
func (c Command) String() string {
	return strings.Join(c, " ")
}
```

test/kubeobjects/pod/execute.go
```diff
@@ -24,16 +24,25 @@ type ExecutionResult struct {
 
 type ExecutionQuery struct {
 	pod       v1.Pod
-	command   string
+	command   []string
```
mjgrzybek, line 27:
```suggestion
	command   shell.Command
```

test/support_archive/support_archive_test.go
```diff
@@ -0,0 +1,111 @@
...
+	operatorPods := functional.Filter(pods.Items, func(podItem v1.Pod) bool {
+		return strings.Contains(podItem.Name, "dynatrace-operator")
+	})
+
+	require.Len(t, operatorPods, 1)
+
+	executionQuery := pod.NewExecutionQuery(operatorPods[0],
+		"dynatrace-operator",
+		"/usr/local/bin/dynatrace-operator",
+		"support-archive",
```
mjgrzybek, line 96: Make it consts with meaningful names

StefanHauth: I kind of like it more this way, because you can see the command that is executed. With consts it get's a bit obfuscated.

mjgrzybek (approved):

PR #1381, opened 2022-12-02
Removes the oneagent immutable image feature-flag

0sewa0:
# Description

The customised oneagent image can handle the additional args/files provided by the operator when immutable image support is enabled. 
This means the feature flag is redundant and the behaviour can just default to immutable image support.

## How can this be tested?
Buy deploy not adding any feature flag the following is true
- `<dynakube-name>-oneagent-tenant-secret` is created in the operator's namespace
- "--set-tenant" "--set-server" args are provided the oneAgent daemonset
- "connection-info-secret" volume and volumeMount is present on the oneAgent daemonset
	- should mount the `<dynakube-name>-oneagent-tenant-secret` secret

## Checklist


src/controllers/dynakube/connectioninfo/reconciler.go
```diff
@@ -40,18 +40,15 @@ func (r *Reconciler) Reconcile() (err error) {
...
-		if err != nil {
-			log.Info("failed to get oneagent connection info")
-			return err
-		}
-
-		err = r.createOrUpdateSecret(r.dynakube.OneagentTenantSecret(), oneAgentConnectionInfo.ConnectionInfo)
-		if err != nil {
-			return err
-		}
+	err = r.createOrUpdateSecret(r.dynakube.OneagentTenantSecret(), oneAgentConnectionInfo.ConnectionInfo)
```
chrismuellner, line 49: we always create this secret now, which is not what we want. since we are splitting the secret into config map and secret, we should keep that in mind and not create/clean up these resources when they are not needed.

chrismuellner (approved):

chrismuellner (approved):

PR #1386, opened 2022-12-06
Adds operator pods log collection to support archive.

StefanHauth:
# Description

Adds collection of logs from all operator related pods into the support archive.

## How can this be tested?
* Deploy the operator
* run `kubectl -n dynatrace exec pod/$(kubectl -n dynatrace get pods | grep dynatrace-operator | cut -f1 -d' ') -- /usr/local/bin/dynatrace-operator support-archive`
* retrieve the support archive tarball (kubectl cmd will be printed on screen) and check it for log files

## Checklist


src/cmd/support_archive/logs.go
```diff
@@ -0,0 +1,121 @@
...
+func (lc logCollector) Name() string {
+	return logCollectorName
+}
+
+func (lc logCollector) getPodList() (*corev1.PodList, error) {
+	listOptions := metav1.ListOptions{
+		TypeMeta: metav1.TypeMeta{
+			Kind: "pod",
+		},
+	}
```
0sewa0, line 64: maybe overthinking here, BUT what if a customer adds some pod into our namespace?
then we would collect the logs for those containers, which I don't think we want

we should maybe use some label selector or something in our list options to only get what we want 

but also MAYBE if we have some random logs from some random container then it can maybe help us debug

0sewa0: if I sound conflicted, that is because I am 😅

0sewa0 (changes requested):

src/cmd/support_archive/logs.go
```diff
@@ -0,0 +1,121 @@
...
+
+type logCollector struct {
+	collectorCommon
+
+	context   context.Context
+	clientSet kubernetes.Interface
+	namespace string
+}
+
+func newLogCollector(log logr.Logger, supportArchive tarball, context context.Context, clientSet kubernetes.Interface, namespace string) collector { //nolint:revive // argument-limit doesn't apply to constructors
```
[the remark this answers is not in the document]
StefanHauth: All the other no-lint comments have the same format.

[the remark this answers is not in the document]
0sewa0: really?! then me sad, and its fine 😅 (but context should still be first argument 😛 )

src/cmd/support_archive/logs.go
StefanHauth: If in doubt, in this case more is better. Cause if there's an additional pod in our namespace, that's suspicious on its own and hence worth knowing.

chrismuellner: we should collect the information that other pods are running in our namespace (i.e. pod names), but should not collect the logs of these pods. I don't like collecting logs or random pods that are not owned by us because of privacy reasons.
We should have documentation how to collect these logs when/if we need them, but not collect them by default.

StefanHauth: Fair enough. Then we keep collecting the existence of enemy pods in our namespace to the next step where we collect K8S objects. So that we at least know something's fishy.

config/helm/chart/default/templates/Common/operator/clusterrole-operator.yaml
```diff
@@ -28,6 +28,12 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
+      - ""
+    resources:
+      - pods/log
+    verbs:
+      - get
```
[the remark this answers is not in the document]
StefanHauth: moved it to the operator role.

src/cmd/support_archive/logs.go
StefanHauth: I've restricted log collection to pods labeled with `app.kubernetes.io/name=dynatrace-operator`

src/cmd/support_archive/logs.go
```diff
@@ -0,0 +1,125 @@
...
+		logErrorf(collector.log, errors.Errorf("Invalid nil pointer returned instead of log stream for pod %s, container %s", pod.Name, container.Name), "")
+		return
+	}
+
+	podLogs, err := req.Stream(collector.context)
+
+	switch {
+	case logOptions.Previous && err != nil:
+		// Soften error message for previous pods a bit, so users don't get nervous. Previous pods often just didn't exist.
+		logInfof(collector.log, "logs for previous pod not found: %v", err)
```
0sewa0, line 99: this looks like 
```
[support-archive]	logs for previous pod not found: previous terminated container "webhook" in pod "dynatrace-webhook-598776b5df-xghxh" not found
```
a bit redundant

src/cmd/support_archive/logs.go
```diff
@@ -0,0 +1,125 @@
...
+func (collector logCollector) collectContainerLogs(pod *corev1.Pod, container corev1.Container, logOptions corev1.PodLogOptions) {
+	req := collector.pods.GetLogs(pod.Name, &logOptions)
+	if req == nil {
+		logErrorf(collector.log, errors.Errorf("Invalid nil pointer returned instead of log stream for pod %s, container %s", pod.Name, container.Name), "")
+		return
+	}
+
+	podLogs, err := req.Stream(collector.context)
+
+	switch {
```
0sewa0, line 96: very small nitpick: could be an `if {} else if {}` 😛

src/cmd/support_archive/mocks/pod_interface.go
```diff
@@ -0,0 +1,855 @@
+// Code generated by mockery v2.15.0. DO NOT EDIT.
+
+package mocks
```
0sewa0, line 3: this is cool, but I dunno if we will use mockery (its veeeeery likely)

couldn't this be in a more general location ?
so it can be reused by other packages if needed
- `PodInterface` from `k8s.io/client-go/kubernetes/typed/core/v1` sounds something that might be useful somewhere else as well 😄

0sewa0 (changes requested):

src/cmd/support_archive/logs.go
```diff
@@ -0,0 +1,125 @@
...
+func (collector logCollector) Name() string {
+	return logCollectorName
+}
+
+func (collector logCollector) getPodList() (*corev1.PodList, error) {
+	listOptions := metav1.ListOptions{
+		TypeMeta: metav1.TypeMeta{
+			Kind: "pod",
+		},
+		LabelSelector: "app.kubernetes.io/name=dynatrace-operator",
```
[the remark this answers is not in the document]
StefanHauth: I actually like to not re-use consts in test in these occasions, because it automatically makes sure, that the const isn't involuntarily changed. It is important that the label selector is kept correct. If it changes that test should fail and then force a thought process.

src/cmd/support_archive/logs.go
StefanHauth: I was waiting for this comment and very surprised it didn't come up earlier 😛

src/cmd/support_archive/mocks/pod_interface.go
StefanHauth: For this mock it isn't even necessary, that we use mockery as a general tool. I just used it to safe some tedious work of coding a mock manually and I had at hand (coincidently 😛).

But you are right, the location is not optimal.

src/cmd/support_archive/logs.go
```diff
@@ -0,0 +1,125 @@
...
+
+		podLogOpts.Previous = true
+		collector.collectContainerLogs(pod, container, podLogOpts)
+	}
+}
+
+func (collector logCollector) collectContainerLogs(pod *corev1.Pod, container corev1.Container, logOptions corev1.PodLogOptions) {
+	req := collector.pods.GetLogs(pod.Name, &logOptions)
+	if req == nil {
+		logErrorf(collector.log, errors.Errorf("Invalid nil pointer returned instead of log stream for pod %s, container %s", pod.Name, container.Name), "")
```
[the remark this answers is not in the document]
StefanHauth: I can make it less scary, but that will not be more informative 😛

src/cmd/support_archive/logs.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
```
chrismuellner, line 11: `v1` is ambiguous, maybe `clientgov1`

src/cmd/support_archive/logs.go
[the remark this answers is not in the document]
chrismuellner: we have these constants in the code already, please use them from `kubeobjects.labels.go`

src/cmd/support_archive/logs.go
```diff
@@ -0,0 +1,123 @@
...
+	req := collector.pods.GetLogs(pod.Name, &logOptions)
+	if req == nil {
+		logErrorf(collector.log, errors.Errorf("Unable to retrieve log stream for pod %s, container %s", pod.Name, container.Name), "")
+		return
+	}
+
+	podLogs, err := req.Stream(collector.context)
+
+	if logOptions.Previous && err != nil {
+		// Soften error message for previous pods a bit, so users don't get nervous. Previous pods often just didn't exist.
```
chrismuellner, line 97:
```suggestion
		// Soften error message for previous pods as they may just nott exist.
```

src/cmd/support_archive/logs.go
```diff
@@ -0,0 +1,123 @@
...
+		logInfof(collector.log, "%v", err)
+		return
+	} else if err != nil {
+		logErrorf(collector.log, err, "error in opening stream")
+		return
+	}
+
+	defer podLogs.Close()
+
+	fileName := buildLogFileName(pod, container, logOptions)
```
chrismuellner, line 107: how about a folder per pod with all the container logs (and previous) in it?
Not sure how many files we'll end up with, but seems like having them all in the top level may get confusing

src/mocks/k8s.io/client-go/kubernetes/typed/core/v1/pod_interface.go
```diff
@@ -0,0 +1,855 @@
+// Code generated by mockery v2.15.0. DO NOT EDIT.
```
chrismuellner, line 1: is this code supposed to be checked in?

config/helm/chart/default/templates/Common/operator/role-operator.yaml
```diff
@@ -127,6 +127,12 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
+      - "" # "" indicates the core API group
```
chrismuellner, line 131:
```suggestion
      - ""
```

chrismuellner (changes requested):

src/cmd/support_archive/logs.go
0sewa0: clientgo is a bit broad 
could be `corev1types` (which is a mouthful)

src/cmd/support_archive/logs.go
0sewa0: GOOD IDEA

I had 3 nodes that makes the following amount of files  => 3*4 (csi) + 2*1 (webhook) + 1 (operator) = 15
this is without any agents or activegate, and if I had 10 nodes this number would go up to 43 which is a bit much

StefanHauth: I'll have to check how easy it is to create folders in a tarball in memory on-the-fly.

src/mocks/k8s.io/client-go/kubernetes/typed/core/v1/pod_interface.go
StefanHauth: For now, yes, of course, it is needed by the tests.

src/cmd/support_archive/logs.go
StefanHauth: piece of cake, done!

0sewa0 (approved):

luhi-DT (approved):

PR #1405, opened 2022-12-13
Restructure and add description of e2e tests

aorcholski:
# Description

All tests moved to the `test/scenarios` folder.
All e2e tests documented in README.md files.

## Checklist


test/scenarios/cloudnative/basic/README.md
```diff
@@ -0,0 +1,58 @@
...
+- sample application restarted
+- check if oneagent init container is injected
+- check if oneagents are able to communicate with the cluster
+
+### CodeModules
+Verify that the storage in the CSI driver directory does not increase when there are multiple tenants and pods which are monitored.
+
+- DTO installed
+- sample application Deployment installed
+- check if oneagent started
```
StefanHauth, line 36:
```suggestion
- check if OneAgent started
```

StefanHauth: AFAIK OneAgent is a kind of brand name or something so it should be written in capital letters. This comment applies to all uses of `oneagent` in this PR.

StefanHauth (approved):
I'm a bit doubtful of separating test code and test description. In my experience the latter one has a hard time getting maintained / is constantly overlooked when tests change. I'd prefer a description of the basic idea behind each test as a comment close to the test code. And it shouldn't be a step-by-step description of what the test does, rather describing the intentions of the test.
Other than that, one step further into a better future of the E2E tests.

test/scenarios/activegate/README.md
```diff
@@ -0,0 +1,16 @@
+# Prerequisites
+k8s cluster
```
chrismuellner, line 2: `k8s cluster` is implied, I think only additional prerequisites are required

test/scenarios/applicationmonitoring/appmon_test.go
```diff
@@ -1,6 +1,6 @@
 //go:build e2e
 
-package appmon
+package applicationmonitoring
```
chrismuellner, line 3: file name should also be adjusted

test/scenarios/cloudnative/basic/README.md
chrismuellner: not official documentation, so unnecessary in my opinion

test/scenarios/applicationmonitoring/README.md
```diff
@@ -0,0 +1,26 @@
+# Prerequisites
+k8s cluster
+
+# Setup
+ApplicationMonitoring deployment without CSI driver
+
+# Goals
+### Verification of the data ingest part of the operator
```
chrismuellner, line 8: wrong section level

test/scenarios/activegate/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+# Setup
+CloudNative deployment with CSI driver
```
chrismuellner, line 5: this seems incorrect

test/scenarios/activegate/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+# Goals
+Verification if ActiveGate is rolled out successfully. All AG capabilities are enabled in Dynakube.
+
+Tests:
```
chrismuellner, line 10: we should split this into a numbered list of steps and then a sub list with a rough outline of checks done for each

example:
1. create Activegate
  - verify pod is running
  - ...
2. ...

chrismuellner (changes requested):

luhi-DT:
> I'm a bit doubtful of separating test code and test description. In my experience the latter one has a hard time getting maintained / is constantly overlooked when tests change. I'd prefer a description of the basic idea behind each test as a comment close to the test code. And it shouldn't be a step-by-step description of what the test does, rather describing the intentions of the test. Other than that, one step further into a better future of the E2E tests.

I also think that the readme's are not a good idea here, what we should have is a small seperate documentation like a wiki page and the rest (details etc.) should be taken from the code

chrismuellner (approved):

PR #1406, opened 2022-12-14
Moves deploymentmetadata to configmap

0sewa0:
# Description

Currently(previous this PR) the deployment-metadata is set in the following ways:
- as a set of arguments for the oneagent daemonset
	- this can cause an restart if the metadata changes, which should not happen
	- different way to the same thing (using args instead of envvars) for no reason (probably because of this part of the logic is quite old)
- as a static envvar for the activegate statefulset
   - this can cause an restart if the metadata changes, which should not happen
- as a static envvar for the injected user pods
   - this is fine

This changes it to the following:
1. We create a `configMap` for each `dynakube` with a name `<name-of-dynakube>-deployment-metadata`.
   - This `configMap` has 2 keys, `oneagent` and `activegate`, each containing a comma separated strings of deployment metadata
   - If only ActiveGate is configured then only the `activegate` key is present same for the OneAgent
2. We provide the deployment-metadata with an environment variable in all cases.
3. For the ActivGate statefulSet and OneAgent daemonSet the deployment-metadata envvar references the corresponding key in the `<name-of-dynakube>-deployment-metadata` configMap

Some misc:
Some consts were moved around to make more sense.
(using a const from the activegate package in the oneagent package also the other way around is a bit wrong)

## How can this be tested?
Deploy a dynakube, configured with OneAgent and/or ActiveGate.
Check if the configmap is present and is used by the statefulSet and daemonSet

## Checklist


src/controllers/dynakube/deploymentmetadata/reconciler.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (r *Reconciler) addOneAgentDeploymentMetadata(data map[string]string) {
+	if !r.dynakube.NeedsOneAgent() {
+		return
+	}
+	data[OneAgentMetadataKey] = NewDeploymentMetadata(r.clusterID, GetOneAgentDeploymentType(r.dynakube)).AsString()
+}
+
+func (r *Reconciler) addActiveGateDeploymentMetadata(data map[string]string) {
```
StefanHauth, line 46:
```suggestion
func (r *Reconciler) addActiveGateDeploymentMetadata(configMapData map[string]string) {
```

src/controllers/dynakube/deploymentmetadata/reconciler.go
```diff
@@ -0,0 +1,64 @@
...
+	if !r.dynakube.NeedsActiveGate() {
+		return
+	}
+	data[ActiveGateMetadataKey] = NewDeploymentMetadata(r.clusterID, ActiveGateMetadataKey).AsString()
+}
+
+func (r *Reconciler) maintainMetadataConfigMap(data map[string]string) error {
```
StefanHauth, line 53:
```suggestion
func (r *Reconciler) maintainMetadataConfigMap(configMapData map[string]string) error {
```

src/controllers/dynakube/deploymentmetadata/reconciler_test.go
```diff
@@ -0,0 +1,146 @@
+package deploymentmetadata
+
+import (
+	"context"
+	"testing"
+
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	// "github.com/stretchr/testify/assert"
```
StefanHauth, line 8: Remove, plz.

src/controllers/dynakube/deploymentmetadata/reconciler_test.go
```diff
@@ -0,0 +1,146 @@
...
+	testTenantToken     = "test-token"
+	testTenantUUID      = "test-uuid"
+	testTenantEndpoints = "test-endpoints"
+)
+
+func TestReconcile(t *testing.T) {
+	clusterID := "test"
+	t.Run(`don't create anything, if no mode is configured`, func(t *testing.T) {
+		dynakube := &dynatracev1beta1.DynaKube{
+			ObjectMeta: metav1.ObjectMeta{
```
StefanHauth, line 29:
```suggestion
			ObjectMeta: createObjectMeta(testNamespace, testName)
```

src/controllers/dynakube/deploymentmetadata/reconciler_test.go
```diff
@@ -0,0 +1,146 @@
...
+		err := r.Reconcile()
+		require.NoError(t, err)
+
+		var actualConfigMap corev1.ConfigMap
+		err = fakeClient.Get(context.TODO(), client.ObjectKey{Name: GetDeploymentMetadataConfigMapName(testName), Namespace: testNamespace}, &actualConfigMap)
+		require.Error(t, err)
+	})
+
+	t.Run(`create configmap with 1 key, if only oneagent is needed`, func(t *testing.T) {
+		dynakube := &dynatracev1beta1.DynaKube{
```
StefanHauth, line 66:
```suggestion
		dynakube := createDynakube(dynatracev1beta1.DynaKubeSpec{ ... })
```

StefanHauth (changes requested):

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -306,30 +318,17 @@ func (controller *Controller) removeAppInjection(ctx context.Context, dynakube *
 }
 
 func (controller *Controller) reconcileOneAgent(ctx context.Context, dynakube *dynatracev1beta1.DynaKube) error {
-	deploymentType := getDeploymentType(dynakube)
+	deploymentType := deploymentmetadata.GetOneAgentDeploymentType(*dynakube)
 
-	if deploymentType == "" {
+	if deploymentType == "" || deploymentType == deploymentmetadata.DeploymentTypeApplicationMonitoring {
```
chrismuellner, line 323: replace check with `dynakube.NeedsOneAgent()`

src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -62,32 +63,32 @@ type Builder interface {
 func NewHostMonitoring(instance *dynatracev1beta1.DynaKube, clusterId string) Builder {
 	return &HostMonitoring{
 		builderInfo{
-			instance:       instance,
+			dynakube:       instance,
 			hostInjectSpec: instance.Spec.OneAgent.HostMonitoring,
-			clusterId:      clusterId,
-			deploymentType: DeploymentTypeHostMonitoring,
+			clusterID:      clusterId,
+			deploymentType: deploymentmetadata.DeploymentTypeHostMonitoring,
```
chrismuellner, line 69: these constants are kind of redundant in combination with the package name, maybe `deploymentmetadata.HostMonitoringKey`

src/controllers/dynakube/deploymentmetadata/config.go
```diff
@@ -0,0 +1,28 @@
...
+
+const (
+	orchestrationTech = "Operator"
+
+	keyOperatorScriptVersion = "script_version"
+	keyOrchestratorID        = "orchestrator_id"
+	keyOrchestrationTech     = "orchestration_tech"
+
+	DeploymentTypeApplicationMonitoring = "application_monitoring"
+	DeploymentTypeFullStack             = "classic_fullstack"
```
chrismuellner, line 15: `ClassicFullStack`, or just `Classic`

src/controllers/dynakube/deploymentmetadata/config.go
```diff
@@ -0,0 +1,28 @@
...
+	DeploymentTypeCloudNative           = "cloud_native_fullstack"
+	DeploymentTypeHostMonitoring        = "host_monitoring"
+	DeploymentTypeActiveGate            = "active_gate"
+
+	EnvDtDeploymentMetadata = "DT_DEPLOYMENT_METADATA"
+
+	OneAgentMetadataKey   = "oneagent"
```
chrismuellner, line 22: `key...` (with `keyOrchestratorID`) as prefix or `...Key` (with `OneAgentMetadataKey`) as suffix should be consistent, both versions are used here

src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -128,7 +129,7 @@ func appendHostIdArgument(result *appsv1.DaemonSet, source string) {
 }
 
 func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
-	instance := dsInfo.instance
+	instance := dsInfo.dynakube
```
chrismuellner, line 132:
```suggestion
	dynakube := dsInfo.dynakube
```

src/controllers/dynakube/oneagent/daemonset/env_vars.go
```diff
@@ -83,20 +126,6 @@ func mapToArray(envVarMap map[string]corev1.EnvVar) []corev1.EnvVar {
...
-				Key:                  "proxy",
-			},
-		})
-	} else {
-		setDefaultValue(envVarMap, proxy, dsInfo.instance.Spec.Proxy.Value)
-	}
-	return envVarMap
-}
-
 func setDefaultValue(envVarMap map[string]corev1.EnvVar, name string, value string) map[string]corev1.EnvVar {
```
chrismuellner, line 129: should this (and the Source one) be in `kubeobjects.env.go`?

src/controllers/dynakube/oneagent/daemonset/env_vars_test.go
```diff
@@ -4,19 +4,244 @@ import (
 	"testing"
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/connectioninfo"
+	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/deploymentmetadata"
 	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
chrismuellner, line 13:
```suggestion
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

src/controllers/dynakube/oneagent/oneagent_reconciler.go
```diff
@@ -157,17 +146,17 @@ func (r *Reconciler) getOneagentPods(ctx context.Context, dynakube *dynatracev1b
...
 	var err error
 
-	switch {
-	case r.feature == daemonset.DeploymentTypeFullStack:
-		ds, err = daemonset.NewClassicFullStack(dynakube, clusterID).BuildDaemonSet()
-	case r.feature == daemonset.DeploymentTypeHostMonitoring:
-		ds, err = daemonset.NewHostMonitoring(dynakube, clusterID).BuildDaemonSet()
-	case r.feature == daemonset.DeploymentTypeCloudNative:
-		ds, err = daemonset.NewCloudNativeFullStack(dynakube, clusterID).BuildDaemonSet()
+	switch r.feature {
```
chrismuellner, line 153: I like the `switch` here, but shouldn't we use the matching properties from the dynakube directly here?

src/controllers/dynakube/oneagent/oneagent_reconciler.go
```diff
@@ -157,17 +146,17 @@ func (r *Reconciler) getOneagentPods(ctx context.Context, dynakube *dynatracev1b
 	return podList.Items, listOps, err
 }
 
-func (r *Reconciler) newDaemonSetForCR(dynakube *dynatracev1beta1.DynaKube, clusterID string) (*appsv1.DaemonSet, error) {
+func (r *Reconciler) getDesiredDaemonSet(dynakube *dynatracev1beta1.DynaKube) (*appsv1.DaemonSet, error) {
```
chrismuellner, line 149: we don't really "get" it, we "build" or "create" the daemonset here

src/kubeobjects/configmap_test.go
```diff
@@ -238,6 +239,30 @@ func testCreateConfigMapInTargetNamespace(t *testing.T) {
...
+			Namespace: testNamespace,
+			Labels:    labels,
+		},
+		Data: data,
+	})
+	configMap := createTestConfigMap(labels, data)
+	configMapQuery := NewConfigMapQuery(context.TODO(), fakeClient, fakeClient, log)
+
+	err := configMapQuery.Delete(*configMap)
+	assert.NoError(t, err)
```
chrismuellner, line 259:
```suggestion
	require.NoError(t, err)
```

chrismuellner (changes requested):

src/controllers/dynakube/deploymentmetadata/config.go
luhi-DT: we have to be careful here, changing this would lead to different values in cluster-data-hub

chrismuellner: I meant the constant name, not the value 😄

src/controllers/dynakube/oneagent/daemonset/daemonset.go
0sewa0: but its not a key :)
its actual the 2. part of the `OrchestrationTech`
the word deployment is the same but you cant remove it without causing more confusion IMO
- `deploymentmetadata.TypeHostMonitoring` kinda refers to the type of deploymentmetadata and not type of deployment
- `deploymentmetadata.HostMonitoringKey` key for what?

src/controllers/dynakube/oneagent/daemonset/env_vars.go
0sewa0: probably, but I think the PR is big enough already to move things around more

src/controllers/dynakube/activegate/internal/statefulset/statefulset.go
```diff
@@ -162,13 +162,19 @@ func (statefulSetBuilder Builder) buildBaseContainer() []corev1.Container {
...
-
 	envs := []corev1.EnvVar{
 		{Name: consts.EnvDtCapabilities, Value: statefulSetBuilder.capability.ArgName()},
 		{Name: consts.EnvDtIdSeedNamespace, Value: statefulSetBuilder.dynakube.Namespace},
 		{Name: consts.EnvDtIdSeedClusterId, Value: string(statefulSetBuilder.kubeUID)},
-		{Name: consts.EnvDtDeploymentMetadata, Value: deploymentMetadata.AsString()},
+		{Name: deploymentmetadata.EnvDtDeploymentMetadata, ValueFrom: &corev1.EnvVarSource{
+			ConfigMapKeyRef: &corev1.ConfigMapKeySelector{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: deploymentmetadata.GetDeploymentMetadataConfigMapName(statefulSetBuilder.dynakube.Name),
```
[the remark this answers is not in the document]
0sewa0: the properties are big enough already, and this is very specific IMO

src/controllers/dynakube/deploymentmetadata/config.go
```diff
@@ -0,0 +1,28 @@
...
+
+	operatorScriptVersionKey = "script_version"
+	orchestratorIDKey        = "orchestrator_id"
+	orchestrationTechKey     = "orchestration_tech"
+
+	ApplicationMonitoringDeploymentType = "application_monitoring"
+	ClassicFullStackDeploymentType      = "classic_fullstack"
+	CloudNativeDeploymentType           = "cloud_native_fullstack"
+	HostMonitoringDeploymentType        = "host_monitoring"
+	ActiveGateDeploymentType            = "active_gate"
```
mjgrzybek, line 18: These consts should come from `oneagent` and `activegate` to respect "open/closed principle".

src/controllers/dynakube/deploymentmetadata/deploymentmetadata.go
```diff
@@ -0,0 +1,51 @@
...
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	"github.com/Dynatrace/dynatrace-operator/src/version"
+)
+
+type DeploymentMetadata struct {
+	OrchestratorID string
+	DeploymentType string
+}
+
+func GetOneAgentDeploymentType(dynakube dynatracev1beta1.DynaKube) string {
```
mjgrzybek, line 16: It should be in `dynakube` package.

mjgrzybek (changes requested):

StefanHauth (dismissed):

src/controllers/dynakube/deploymentmetadata/deploymentmetadata.go
0sewa0: this kinda contradicts the previous comment
If I do both then I will get circular dependency as the `GetOneAgentDeploymentType` in the `dynakube` package would need to import the `oneagent` package to use the constants, but in the `oneagent` package we already import the `dynakube` package.

I could move both the function and constants to the `dynakube` package

chrismuellner (approved):

PR #1425, opened 2022-12-16
Adds K8S object collection to support archive.

StefanHauth:
# Description

Add collection of selected K8S objects from the operator namespace and namespaces where injection has happened to the support archive.

## How can this be tested?
Deploy the operator and run:
`kubectl -n dynatrace exec pod/$(kubectl -n dynatrace get pods | grep dynatrace-operator | cut -f1 -d' ') -- /usr/local/bin/dynatrace-operator support-archive`

The command to retrieve the support archive will be printed on screen. Get it, unpack it and check the content.

## Checklist


chrismuellner (changes requested):
Update PR title

src/cmd/support_archive/objects_query.go
```diff
@@ -0,0 +1,81 @@
...
+		},
+		{
+			groupVersionKind: schema.GroupVersionKind{
+				Group:   "",
+				Version: "v1",
+				Kind:    "NamespaceList",
+			},
+			listOptions: []client.ListOption{
+				client.MatchingFields{
+					"metadata.name": namespace,
```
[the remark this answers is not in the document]
StefanHauth: This file contains the description of how to query the objects we want. It is done this way to keep the query mechanism general and not implemented separately for each object type. I don't want special handling just because the list returns only one object.

src/cmd/support_archive/objects.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+	return nil
+}
+
+func (collector k8sCollector) Name() string {
+	return k8sCollectorName
+}
+
+func (collector k8sCollector) readObjectsList(manifest objectQuery) (*unstructured.UnstructuredList, error) {
```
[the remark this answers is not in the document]
StefanHauth: No, then we have to implement this for each object type separately, which is not necessary. The beauty of unstructured is that we can query all object types the same way. As the only thing we need is to write the manifests to files, we don't need any type-specific handling.

src/cmd/support_archive/objects.go
```diff
@@ -0,0 +1,90 @@
...
+	"bytes"
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const k8sCollectorName = "k8sCollector"
```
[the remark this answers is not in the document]
StefanHauth: IMHO `object` is even less specific. I'd go for `k8sObjects`.

src/cmd/support_archive/objects.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		logErrorf(collector.log, err, "Failed to add %s to support archive", fileName)
+		return
+	}
+
+	logInfof(collector.log, "Collected manifest for %s", fileName)
+}
+
+func createFileName(kind string, objectMeta unstructured.Unstructured) string {
+	if len(objectMeta.GetNamespace()) > 0 {
```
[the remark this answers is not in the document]
StefanHauth: In the future we might want to also collect stuff from other namespaces. If we start with a clean structure from the beginning, we can keep the support archive structure stable.

src/cmd/support_archive/objects_query.go
```diff
@@ -0,0 +1,81 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type objectQuery struct {
+	groupVersionKind schema.GroupVersionKind
+	listOptions      []client.ListOption
+}
+
+func getObjectQueries(namespace string) []objectQuery {
```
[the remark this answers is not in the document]
StefanHauth: This is the most maintainable form. The object queries define, which objects we are collecting. This is the input for the collecting algorithm. Here you a) have a nice overview of what we collect and b) can easily extend the list of objects by just adding a new query.

src/cmd/support_archive/objects_query.go
```diff
@@ -0,0 +1,81 @@
...
+				Kind:    "DaemonSetList",
+			},
+			listOptions: []client.ListOption{
+				client.InNamespace(namespace),
+			},
+		},
+		{
+			groupVersionKind: schema.GroupVersionKind{
+				Group:   "dynatrace.com",
+				Version: "v1beta1",
```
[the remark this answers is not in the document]
StefanHauth: Will check.

src/cmd/support_archive/objects.go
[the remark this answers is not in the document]
chrismuellner: why `k8s...`? We don't have `k8sLogs` either 😄

src/cmd/support_archive/objects_query.go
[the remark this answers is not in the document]
StefanHauth: And I wouldn't expect too much noise when collecting all objects in the operator namespace. On the contrary, if there are unexpected objects in this namespace, it is very interesting to know about them, because it is a strong indication for a faulty or at the very least non-standard operator deployment.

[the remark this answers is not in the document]
chrismuellner: I agree it's nice to collect all the queries in one place, but I don't think it applies here.

We have at least 2 different types of queries: resources in the operator namespace and other namespaces themselves. `listOptions` has 3 different values (1 of them not a list)...

src/cmd/support_archive/objects.go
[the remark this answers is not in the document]
StefanHauth: as discussed offline, we use `manifests`.

luhi-DT:
I'll add my thoughts here: I dont like the idea of maintaining the support archive manifest queries. We should opt for a more generic approach, in order to tackle this. What if we would add another ressource type deployed by the operator? We would have to adapt the queries and include this type aswell -> pitfall. I suggested to use something like `kubectl get all -n dynatrace` for that approach

StefanHauth:
> I'll add my thoughts here: I dont like the idea of maintaining the support archive manifest queries. We should opt for a more generic approach, in order to tackle this. What if we would add another ressource type deployed by the operator? We would have to adapt the queries and include this type aswell -> pitfall. I suggested to use something like `kubectl get all -n dynatrace` for that approach

Unfortunately, that's not doable. `kubectl get all` also just has an internal list of resources it pulls and it's actually far from `all`. A general `get *` like command isn't supported.

src/cmd/support_archive/objects_query.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
+	"github.com/Dynatrace/dynatrace-operator/src/webhook"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type resourceQueryGroup struct {
+	resources []string
```
luhi-DT, line 16: not needed

src/cmd/support_archive/objects_query.go
```diff
@@ -0,0 +1,101 @@
+package support_archive
```
luhi-DT, line 1: would rename this file to resource_query.go e.g

src/cmd/support_archive/objects_query.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/Dynatrace/dynatrace-operator/src/webhook"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type resourceQueryGroup struct {
+	resources []string
+	resourcex []schema.GroupVersionKind
```
luhi-DT, line 17: please rename

src/cmd/support_archive/objects_query.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, resource := range q.resourcex {
+		queries = append(queries, resourceQuery{
+			groupVersionKind: resource,
+			filters:          q.filters,
+		})
+	}
+	return queries
+}
+
+type resourceQuery struct {
```
luhi-DT, line 98: would move this struct to the top where resourceQueryGroup is defined

luhi-DT (dismissed):

src/cmd/support_archive/objects_query.go
StefanHauth: consequently all other `objects*.*` files have to be renamed too

src/cmd/support_archive/objects_query.go
StefanHauth: I don't like how the main purpose of this file, defining the queries get's pushed down in the file more and more.

waodim:
Tested it, seems to work and LGTM so apart from the comments of the others I thinks it is fine.

src/cmd/support_archive/resources_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+import (
+	"archive/tar"
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"testing"
+
+	"github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
```
0sewa0, line 11:
```suggestion
	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
```

src/cmd/support_archive/resource_query.go
```diff
@@ -0,0 +1,100 @@
...
+	resources []schema.GroupVersionKind
+	filters   []client.ListOption
+}
+
+type resourceQuery struct {
+	groupVersionKind schema.GroupVersionKind
+	filters          []client.ListOption
+}
+
+func getQueries(namespace string) []resourceQuery {
```
0sewa0, line 25: just to double check, we dont want to collect "everything" in the first iteration just to be safe ?

because there are still stuff that could be collected, like yamls of injected pods that would be really useful during debug

src/cmd/support_archive/resource_query.go
```diff
@@ -0,0 +1,100 @@
...
+			toGroupVersionKind(appsv1.SchemeGroupVersion, appsv1.ReplicaSet{}),
+			toGroupVersionKind(corev1.SchemeGroupVersion, corev1.Service{}),
+			toGroupVersionKind(corev1.SchemeGroupVersion, corev1.Pod{}),
+			toGroupVersionKind(v1beta1.GroupVersion, v1beta1.DynaKube{}),
+		},
+		filters: []client.ListOption{
+			client.MatchingLabels{
+				kubeobjects.AppNameLabel: "dynatrace-operator",
+			},
+			client.InNamespace(namespace),
```
0sewa0, line 74: so we collect only stuff that belongs directly to the operator?
I think it would be preferable to collect everything in the operator's namespace, as that will give us a fuller picture of what we are dealing with.

src/cmd/support_archive/resource_query.go
```diff
@@ -0,0 +1,100 @@
...
+func getOperatorComponentsQueryGroup(namespace string) resourceQueryGroup {
+	return resourceQueryGroup{
+		resources: []schema.GroupVersionKind{
+			toGroupVersionKind(appsv1.SchemeGroupVersion, appsv1.Deployment{}),
+			toGroupVersionKind(appsv1.SchemeGroupVersion, appsv1.StatefulSet{}),
+			toGroupVersionKind(appsv1.SchemeGroupVersion, appsv1.DaemonSet{}),
+			toGroupVersionKind(appsv1.SchemeGroupVersion, appsv1.ReplicaSet{}),
+			toGroupVersionKind(corev1.SchemeGroupVersion, corev1.Service{}),
+			toGroupVersionKind(corev1.SchemeGroupVersion, corev1.Pod{}),
+			toGroupVersionKind(v1beta1.GroupVersion, v1beta1.DynaKube{}),
```
0sewa0, line 68: nitpick:
this list feels like a limited selection of what we could collect.
probably every single component could be collected that comes with the install, but these are the most interesting ones so its fine for now.
stuff like `MutatingWebhookConfiguration`, `ValidatingWebhookConfiguration` could be useful, but they have certs, even if generated by us, it still feels wrong collecting them.

src/cmd/support_archive/resource_query.go
0sewa0: that said, there still could be some 3rd party stuff in the operator's namespace, that we will not know anything about because it can be a not-standard component so we couldn't really added to our list even if we wanted to.

(!! be warned, bad idea incoming !!)
we could get all the crds in the cluster and list every cr in our namespace.
would this blow up the resource usage of the support archive collection? most definitely 
should we do this? most definitely not :)

src/cmd/support_archive/resource_query.go
0sewa0: offline answer => maaaybe in the future

src/cmd/support_archive/resource_query.go
chrismuellner: > we could get all the crds in the cluster and list every cr in our namespace.

would require the operator to have permission to do this (non-namespaced CRDs and namespaced CRs), which sounds like a bad idea, just for the fact that it _may_ be useful in the support archive.

src/cmd/support_archive/resource_query.go
chrismuellner: goal for now is to collect information about direct operator components. let's get that archive into a good state, and then check how we can extend it to include more information, like injected pods, in the future.

src/cmd/support_archive/resource_query.go
0sewa0: > (!! be warned, bad idea incoming !!)
I did say that it was a bad idea 😂

0sewa0: TLDR: Its fine as it is now

0sewa0 (approved):

PR #1433, opened 2022-12-27
Fixes bug which shows unset codeModules image as error

waodim:
# Description
Previously the troubleshoot subcommand interpreted an unset codeModules image in the dynakube as an error, which is not correct. This should fix this issue.

## How can this be tested?
Simply run the troubleshoot subcommand on a cluster which has everything set up.

## Checklist


src/cmd/troubleshoot/image.go
```diff
@@ -40,7 +40,7 @@ func verifyAllImagesAvailable(troubleshootCtx *troubleshootContext) error {
 
 func verifyImageIsAvailable(troubleshootCtx *troubleshootContext, comp component, proxyWarning bool) {
 	image, isCustomImage := comp.getImage(&troubleshootCtx.dynakube)
-	if image == "" {
```
chrismuellner, line 43: would it make sense to move this check into `component`, e.g. `component.IsImageOptional()`?
Checking a specific component here is not extensible, if we were to add additional components

chrismuellner (changes requested):

src/cmd/troubleshoot/image.go
0sewa0: Good idea
also `component.SkipImageCheck()` sounds nicer IMO 😛 

However, the `getImage(...)` is a bit incorrect (maybe)
```go
func (c component) getImage(dynakube *v1beta1.DynaKube) (string, bool) {
        // ...
	switch c {
	case componentOneAgent:
		return dynakube.OneAgentImage(), dynakube.CustomOneAgentImage() != ""
	case componentCodeModules:
		return dynakube.CodeModulesImage(), false // <-- this should probably be true, as its always a custom-image
	case componentActiveGate:
		return dynakube.ActiveGateImage(), dynakube.ActiveGateImage() != "" // <-- should use local var to store `dynakube.ActiveGateImage()` return and reuse it
	}
	return "", false
}
```

0sewa0 (approved):

PR #1437, opened 2023-01-02
Improve performance of namespace secret generation

### LINKED ISSUE #1251: DT operator reconciles secrets too frequently
We like to keep a close tabs on the consumers of kube-api and found the dynatrace operator is the highest consumer of the secrets api in all of our clusters. Based on the operator logs its pretty apparent it's doing a constant reconcile of the secrets in all of the namespaces. This might only be true for app only injection as we don't use the CSI driver.

```
$ oc get apirequestcount secrets.v1 -o jsonpath='{range .status.currentHour..byUser[*]}{.requestCount}{","}{..byVerb[*].verb}{","}{.username}{","}{.userAgent}{"\n"}{end}' | sort -k 2 -t, -u | column -t -s, | sort -k1 -nr | head -4
26622  create delete get list update watch  system:serviceaccount:dynatrace:dynatrace-operator                                                             dynatrace-operator/v0.0.0
9236   get list watch                       system:serviceaccount:openshift-authentication-operator:authentication-operator                                Go-http-client/2.0
8705   get list watch                       system:serviceaccount:openshift-apiserver-operator:openshift-apiserver-operator                                Go-http-client/2.0
3486   get watch                            system:serviceaccount:openshift-machine-config-operator:default                                                machine-config-operator/v0.0.0
```

The secrets being reconciled in 400+ namespaces

1. dynatrace-dynakube-config
2. dynatrace-data-ingest-endpoint

Even though the secrets haven't changed, it will run the reconcile loop as soon as the previous one completes (avg 7 requests per second). Due to the number of calls to the api (2 per namespace) it would be preferable if there were more intelligence to determine if these secrets need to be reconciled. Example:

1. Perform a global watch and filter for the DT secrets using a fieldSelector and labelSelector to reconcile changes if necessary (e.g. inadvertently deleted or modified by user). 
2. Have the webhook configured for namespaces, to be notified when new namespaces are created and trigger the reconcile there (ie. create secrets)
3. A catch-all global reconcile like the current one that runs infrequently

An even better option would be to eliminate these secrets all together, it's a lot of duplicated data but I don't know whether that is feasible for a setup with app-only injection.

Cluster Info: OpenShift 4.10.36, Kubernetes v1.23.5
Dyntrace Operator Version: 0.9.0
DynaKube:
```
kind: DynaKube
metadata:
  annotations:
    feature.dynatrace.com/activegate-authtoken: "true"
    feature.dynatrace.com/automatic-injection: "false"
  name: dynakube
  namespace: dynatrace
spec:
  activeGate:
    capabilities:
    - routing
    - metrics-ingest
    - dynatrace-api
    group: lab_cluster
    replicas: 1
    resources:
      limits:
        cpu: 1000m
        memory: 1.5Gi
      requests:
        cpu: 100m
        memory: 512Mi
  apiUrl: **************
  namespaceSelector:
    matchExpressions:
    - key: dt-enable <--- 400+ namespaces labelled
      operator: Exists
  networkZone: lab.intranet
  oneAgent:
    applicationMonitoring:
      initResources:
        limits:
          cpu: 100m
          memory: 256Mi
        requests:
          cpu: 10m
          memory: 256Mi
      useCSIDriver: false
  skipCertCheck: true
  tokens: dynakube
```

gkrenn:
# Description

Handling secrets for multiple namespaces is currently done by checking each secret one by one. 

This leads to performance problems for the operator, lot's of requests to the kubernetes API and log spamming.

This PR addresses the problem by fetching all relevant namespace secrets at once and only updates/creates outdated ones.
Fixes #1251 

## How can this be tested?

- Deploy the operator as usual and a applicationOnly monitoring DynaKube, which monitores multiple namespaces
- In the logs one should see that reconciliation is not slowed down by secret updates

## Checklist


src/initgeneration/initgeneration.go
```diff
@@ -88,26 +88,22 @@ func (g *InitGenerator) GenerateForDynakube(ctx context.Context, dk *dynatracev1
...
-		if err != nil {
-			return errors.WithStack(err)
-		}
+	updateCnt, err := secretQuery.CreateOrUpdateForNamespacesList(secret, nsList)
+	if err != nil {
+		return err
 	}
 
-	log.Info("done updating init secrets")
+	log.Info("done updating init secrets", "update/creation count", updateCnt)
```
chrismuellner, line 106: log could be once in `CreateOrUpdateForNamespacesList` using the `secret.Name` instead of having it after every call of the function

src/kubeobjects/secret.go
```diff
@@ -70,6 +84,42 @@ func (query SecretQuery) CreateOrUpdate(secret corev1.Secret) error {
 	return nil
 }
 
+func (query SecretQuery) CreateOrUpdateForNamespacesList(newSecret corev1.Secret, namespaces []corev1.Namespace) (int, error) {
+	secretList, err := query.GetAllFromNamespaces(newSecret.Name)
+	if err != nil {
+		return 0, err
+	}
+
+	namespacesContainingSecret := make(map[string]corev1.Secret)
```
chrismuellner, line 93: since we know the length the map is going to have (`len(secretList)`) we could allocate that directly

hack/make/go.mk
```diff
@@ -29,7 +29,7 @@ go/test-all:
 go/run: export RUN_LOCAL=true
 go/run: export POD_NAMESPACE=dynatrace
 go/run: manifests/kubernetes manifests/openshift go/fmt go/vet
-	go run ./src/cmd/operator/
+	go run ./src/cmd/ operator
```
chrismuellner, line 32: 🤔

chrismuellner (changes requested):

src/initgeneration/initgeneration.go
gkrenn: The reason why I did not do that was that for me `SecretQuery` is just a utility and should not force its user to actually produce logs. This is already a pain when initially starting up on a multiple namespace cluster. 

It will give you lot's of log lines "created secret" because `SecretQuery` has a log coded into it. - In my opinion on the long run we should remove the logs from our kubernetes utility functions all together.

hack/make/go.mk
gkrenn: Looks like a typo but is actually correct. - We moved our main.go to ./src/cmd and added some subcommands. The one we need here is `operator`.

src/kubeobjects/secret.go
```diff
@@ -70,6 +84,42 @@ func (query SecretQuery) CreateOrUpdate(secret corev1.Secret) error {
 	return nil
 }
 
+func (query SecretQuery) CreateOrUpdateForNamespacesList(newSecret corev1.Secret, namespaces []corev1.Namespace) (int, error) {
+	secretList, err := query.GetAllFromNamespaces(newSecret.Name)
```
0sewa0, line 88: could add a log line here:
```
query.log.Info("creating/updating secret for multiple namespaces", "name", newSecret.Name, "len(namespaces)", len(namespaces))
```

src/kubeobjects/secret.go
```diff
@@ -70,6 +84,42 @@ func (query SecretQuery) CreateOrUpdate(secret corev1.Secret) error {
...
+			}
+		} else {
+			err = query.Create(newSecret)
+			if err != nil {
+				return updateAndCreationCount, err
+			}
+			updateAndCreationCount++
+		}
+	}
+
```
0sewa0, line 119: what would be kinda cool, is that you could delete the leftover/not-necessary secrets. (stuff in the `namespacesContainingSecret` that were not looked at during the previous loop)
but then the functions name is not that accurate 😅

src/initgeneration/initgeneration.go
0sewa0: having the `create` and `update` logs make sense most of the time, but I agree when you are bulk creating/updating then it gets a bit too much.

having reasonable logs handled by the query makes sense.

you could have a 
```
func (query SecretQuery) create(secret corev1.Secret) error {...)
```
that does not log, so it can be used by the other `query` functions in a way that it only logs what is necessary

src/kubeobjects/secret.go
gkrenn: This is done per dynakube, what if there are 2 application monitoring dynakubes on different namespaces? - One reconciliation would then delete the secrets of the other, wouldn't it?

src/initgeneration/initgeneration.go
gkrenn: Like the idea of the private function, added it

src/kubeobjects/secret.go
0sewa0: If the name of the secret is unique to a dynakube then no,
if the name of the secret is always the same then yes, (and I think this is the case, its always `dynatrace-dynakube-config`)

so its a possible future improvement 😉

src/kubeobjects/secret_test.go
```diff
@@ -0,0 +1,146 @@
...
+		secrets, err := secretQuery.GetAllFromNamespaces(testSecretName)
+		require.NoError(t, err)
+		assert.Len(t, secrets, 3)
+	})
+	// does not work due to the fake client
+	/*t.Run("get not yet existing secret from all namespaces", func(t *testing.T) {
+		secrets, err := secretQuery.GetAllFromNamespaces("not a secret")
+		require.NoError(t, err)
+		assert.Len(t, secrets, 0)
+	})*/
```
0sewa0, line 86: remove or fix

0sewa0: (also, unit tests are failing)

src/kubeobjects/secret.go
```diff
@@ -70,6 +94,49 @@ func (query SecretQuery) CreateOrUpdate(secret corev1.Secret) error {
 	return nil
 }
 
+func (query SecretQuery) CreateOrUpdateForNamespacesList(newSecret corev1.Secret, namespaces []corev1.Namespace) error {
+	secretList, err := query.GetAllFromNamespaces(newSecret.Name)
+	if err != nil {
+		return err
+	}
+
+	query.log.Info("creating/updating secret for multiple namespaces",
```
0sewa0, line 103:
```suggestion
	query.log.Info("reconciling secret for multiple namespaces",
```

src/kubeobjects/secret.go
```diff
@@ -70,6 +94,49 @@ func (query SecretQuery) CreateOrUpdate(secret corev1.Secret) error {
...
+		} else {
+			err = query.create(newSecret)
+			if err != nil {
+				return err
+			}
+			creationCount++
+		}
+	}
+
+	query.log.Info("created/updated secret for multiple namespaces",
```
0sewa0, line 134:
```suggestion
	query.log.Info("reconciled secret for multiple namespaces",
```

src/kubeobjects/secret.go
0sewa0: this log could possible be removed as the following looks a bit redundant
```
...
{"level":"info","ts":"2023-01-03T10:19:30.069Z","logger":"initgeneration","msg":"creating/updating secret for multiple namespaces","name":"dynatrace-dynakube-config","len(namespaces)":1 }
{"level":"info","ts":"2023-01-03T10:19:30.069Z","logger":"initgeneration","msg":"created/updated secret for multiple namespaces","name":"dynatrace-dynakube-config","creationCount":0,"updateCount":0}
...
```

gkrenn: Would not remove it otherwise it would print "querying secret from all namespaces" as last msg before the secrets are created. If there is a greater amount of namespaces this can take a while, which could lead to the false assumption that the querying takes its time.

0sewa0 (dismissed):

0sewa0 (approved):

chrismuellner (approved):

ctml91:
Any rough idea what release this will be included in? I don't see it in 0.10.2 which was cut a couple weeks back.

PR #1451, opened 2023-01-05
Fix sqlite library version

0sewa0:
# Description

the sqlite library's version was raised to 2.0.3 by accident by the creators of the package so its invalid.

## How can this be tested?
check that the csi-driver works as expected

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,6 +79,7 @@ tags
 .history
 # End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
 /.idea/
+/.fleet/
 /debug.sh
 /tmp/
 /reapply-all.sh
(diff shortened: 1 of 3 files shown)


chrismuellner (changes requested):
Can we take this opportunity to use tagged versions for a few more of our go packages?
* `golang.org/x/net:v0.5.0` https://pkg.go.dev/golang.org/x/net
* `golang.org/x/sys:v0.4.0` https://pkg.go.dev/golang.org/x/sys
* ...

Can we restrict dependabot from updating our dependencies to these dev versions? see [dependabot docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore)

chrismuellner (approved):
Let's update the other packages in another PR

PR #1469, opened 2023-01-12
Move to leases based leader election, + make it more configurable

0sewa0:
# Description

We used `configmapleases` type of leader election for 2 minor releases (0.9, 0.10), which is meant as a migration tool, for moving from `configmap` to `leases` based leader election. 
Furthermore I have tested upgrading from a `configmap` leader election operator (0.8.2) to this with `leases` leader election and noticed no issues.

Also made the leader election configurable, because in some cluster due to some timeout problems the operator gets restarted. The following envvars can be added to the operator container:
- `LEADER_ELECTION_LEASE_DURATION`
   - "LeaseDuration is the duration that non-leader candidates will wait to force acquire leadership. This is measured against time of last observed ack. Default is 15 seconds."
- `LEADER_ELECTION_RENEW_DEADLINE`
   -  "RenewDeadline is the duration that the acting controlplane will retry refreshing leadership before giving up. Default is 10 seconds."
- `LEADER_ELECTION_RETRY_PERIOD`
   -  "RetryPeriod is the duration the LeaderElector clients should wait between tries of actions. Default is 2 seconds."

Each envvar should be just a simple number that represents the amount of seconds.

The leader election lost due to timeout issue is only a "cosmetic" issue, because the operator gets back up and the leader election happens when there is not much happening.
This scenario is rather hard to reproduce, as you need a cluster with the problem(sometimes GKE cluster do it, but maybe 1 out of 5), + you have to wait hours/days for the problem to first appear.
So adding these config option will allow us to play around with config values easily when the problem shows up, and provide guidance in the future how to fix it.
At the end of the day we can't have 1 size fits all config for this, so letting the users change the config to their needs is inevitable.

## How can this be tested?
Try out the envvars, be careful you can configure it in a way that the operator will crash loop. (due to constantly losing the leader election)
Try if it "just works" with no config

## Checklist


luhi-DT (approved):

waodim (approved):
LGTM

PR #1493, opened 2023-01-20
Installs operator in e2e tests via make

waodim:
# Description
This should change the installation process in the e2e tests regarding the operator. With this change the make command will be used to install the operator.

## How can this be tested?
Simply run the e2e tests.

## Checklist

### DIFF
--- a/test/operator/installation.go
+++ b/test/operator/installation.go
@@ -3,7 +3,6 @@ package operator
 import (
 	"context"
 	"net/url"
-	"os"
 	"os/exec"
 	"path"
 	"strings"
@@ -33,7 +32,7 @@ func InstallFromSource(withCsi bool) features.Func {
 
 func InstallViaMake() features.Func {
 	return func(ctx context.Context, t *testing.T, environmentConfig *envconf.Config) context.Context {
-		rootDir := getSourceRootDir()
+		rootDir := project.RootDir()
 
 		runBuildAndManifests(rootDir, t)
 
@@ -71,18 +70,6 @@ func runBuildAndManifests(rootDir string, t *testing.T) {
 		t.Fatal("failed to install the operator via the make command", err)
 		return
 	}
-}
-
-func getSourceRootDir() string {
-	currentDir, err := os.Getwd()
-
-	if err != nil {
-		return ""
-	}
-
-	rootDir := strings.Split(currentDir, "test")
-
-	return rootDir[0]
 }
 
 func manifestsPaths(withCsi bool) []string {
--- a/test/operator/installation.go
+++ b/test/operator/installation.go
@@ -30,7 +30,7 @@ func InstallFromSource(withCsi bool) features.Func {
 	return manifests.InstallFromFiles(paths)
 }
 
-func InstallViaMake() features.Func {
+func InstallViaMake(withCSI bool) features.Func {
 	return func(ctx context.Context, t *testing.T, environmentConfig *envconf.Config) context.Context {
 		rootDir := project.RootDir()
 
@@ -46,6 +46,10 @@ func InstallViaMake() features.Func {
 		default:
 			t.Fatal("failed to install the operator via the make command as no correct platform was set")
 			return nil
+		}
+
+		if !withCSI {
+			makeTarget = strings.Join([]string{makeTarget, "no-csi"}, "-")
 		}
 
 		err := exec.Command("make", "-C", rootDir, makeTarget).Run()
--- a/test/operator/installation.go
+++ b/test/operator/installation.go
@@ -37,20 +37,7 @@ func InstallViaMake(withCSI bool) features.Func {
 		runBuildAndManifests(rootDir, t)
 
 		platform := kubeobjects.ResolvePlatformFromEnv()
-		makeTarget := "deploy"
-		switch platform {
-		case kubeobjects.Openshift:
-			makeTarget = strings.Join([]string{makeTarget, "openshift"}, "/")
-		case kubeobjects.Kubernetes:
-			makeTarget = strings.Join([]string{makeTarget, "kubernetes"}, "/")
-		default:
-			t.Fatal("failed to install the operator via the make command as no correct platform was set")
-			return nil
-		}
-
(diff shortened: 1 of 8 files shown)


test/operator/installation.go
```diff
@@ -25,6 +31,40 @@ func InstallFromSource(withCsi bool) features.Func {
...
+		if err != nil {
+			t.Fatal("failed to install the operator via the make command")
+			return nil
+		}
+
+		return ctx
+	}
+}
+
+func getCorrectDir() string {
```
StefanHauth, line 56: What qualifies a directory to be "correct"?
```suggestion
func getSourceRootDir() string {
```

test/operator/installation.go
```diff
@@ -25,6 +31,40 @@ func InstallFromSource(withCsi bool) features.Func {
...
+
+		platform := kubeobjects.ResolvePlatformFromEnv()
+		makeTarget := "deploy"
+		if platform == kubeobjects.Openshift {
+			makeTarget = strings.Join([]string{makeTarget, "openshift"}, "/")
+		} else if platform == kubeobjects.Kubernetes {
+			makeTarget = strings.Join([]string{makeTarget, "kubernetes"}, "/")
+		}
+
+		err := exec.Command("make", "-C", rootDir, makeTarget).Run()
```
StefanHauth, line 46: Doesn't that also need the "build" and "manifests/branch" targets, or is having built the operator from the command line a mandatory pre-requisite one has to know?

test/operator/installation.go
```diff
@@ -25,6 +31,40 @@ func InstallFromSource(withCsi bool) features.Func {
...
+	return func(ctx context.Context, t *testing.T, environmentConfig *envconf.Config) context.Context {
+		rootDir := getCorrectDir()
+
+		platform := kubeobjects.ResolvePlatformFromEnv()
+		makeTarget := "deploy"
+		if platform == kubeobjects.Openshift {
+			makeTarget = strings.Join([]string{makeTarget, "openshift"}, "/")
+		} else if platform == kubeobjects.Kubernetes {
+			makeTarget = strings.Join([]string{makeTarget, "kubernetes"}, "/")
+		}
```
StefanHauth, line 44: Add some handling of the `else` branch here. Otherwise you'll end up running `make deploy`, which doesn't work (`make: *** No rule to make target 'deploy'.  Stop.`) It's always better to prevent an error early than catch it later.

StefanHauth (changes requested):

test/operator/installation.go
waodim: Yes, you're right, thanks. I'll fix this.

test/operator/installation.go
```diff
@@ -25,6 +31,60 @@ func InstallFromSource(withCsi bool) features.Func {
...
+	}
+
+	err = exec.Command("make", "-C", rootDir, "manifests/branch").Run()
+	if err != nil {
+		t.Fatal("failed to install the operator via the make command", err)
+		return
+	}
+}
+
+func getSourceRootDir() string {
```
aorcholski, line 76: [project.RootDir](https://github.com/Dynatrace/dynatrace-operator/blob/c49dfc1f643b15bda9e71b0517a911a5b87911f1/test/project/project.go#L10) function could be used instead of new function doing the same job

waodim: thank you, did completely oversee this function.

test/scenarios/applicationmonitoring/labelversiondetection.go
```diff
@@ -124,17 +123,15 @@ func installDynakube(t *testing.T, name string, annotations map[string]string) f
...
+		ApplicationMonitoring(&v1beta1.ApplicationMonitoringSpec{}).Build()))
 	defaultInstallation.Assess("dynakube phase changes to 'Running'", dynakube.WaitForDynakubePhase(dynakube.NewBuilder().Name(name).Namespace(dynakube.Namespace).Build()))
 
 	return defaultInstallation.Feature()
 }
 
 func installSampleApplications() features.Feature {
 	defaultInstallation := features.New("sample applications installation")
-	defaultInstallation.Assess("sample applications applied", manifests.InstallFromFile("../testdata/application-monitoring/buildlabels-sample-apps.yaml"))
+	defaultInstallation.Assess("sample applications applied", manifests.InstallFromFile("../../testdata/application-monitoring/buildlabels-sample-apps.yaml"))
```
aorcholski, line 134: project.TestDataDir could be used here

aorcholski (changes requested):

test/operator/installation.go
```diff
@@ -25,6 +30,52 @@ func InstallFromSource(withCsi bool) features.Func {
...
+		case kubeobjects.Kubernetes:
+			makeTarget = strings.Join([]string{makeTarget, "kubernetes"}, "/")
+		default:
+			t.Fatal("failed to install the operator via the make command as no correct platform was set")
+			return nil
+		}
+
+		if !withCSI {
+			makeTarget = strings.Join([]string{makeTarget, "no-csi"}, "-")
+		}
```
0sewa0, line 53: nitpick: maybe a bit overkill but you could move this into a helper function AND unittest it, as its just simple string concats with easy inputs and outputs to check for

0sewa0: something like:
```
getDeployMakeTarget(platform string, withCSI bool) string {...}
```

test/scenarios/applicationmonitoring/labelversiondetection.go
```diff
@@ -102,7 +104,7 @@ func installOperator(t *testing.T) features.Feature {
 	defaultInstallation := features.New("default installation")
 
 	defaultInstallation.Setup(secrets.ApplyDefault(secretConfig))
-	defaultInstallation.Setup(operator.InstallFromSource(false))
+	defaultInstallation.Setup(operator.InstallViaMake(true))
```
aorcholski, line 107:
```suggestion
	defaultInstallation.Setup(operator.InstallViaMake(false))
```

aorcholski (approved):

PR #1495, opened 2023-01-20
Add runAsUser/runAsGroup fields to ActiveGate SecurityContext

aorcholski:
# Description

ActiveGate containers i.e.  `activegate` regular container and `certificate-loader` init container should have the same securityContext. Additionally the securityContext should have `runAsUser` and `runAsGroup` fields set to the same value as it is specified by ActiveGate Dockerfile (`USER`).

## How can this be tested?
Deploy DynaKube with `kubernetes-monitoring` capability enabled on:
- a Kubernetes cluster
- an OpenShift cluster.

## Checklist

### DIFF
--- a/src/controllers/dynakube/activegate/consts/consts.go
+++ b/src/controllers/dynakube/activegate/consts/consts.go
@@ -3,6 +3,8 @@ package consts
 import (
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/connectioninfo"
+	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects/address"
+	corev1 "k8s.io/api/core/v1"
 )
 
 const (
@@ -59,4 +61,26 @@ const (
 	GatewaySslMountPoint     = "/var/lib/dynatrace/gateway/ssl"
 	GatewayLogMountPoint     = "/var/log/dynatrace/gateway"
 	GatewayTmpMountPoint     = "/var/tmp/dynatrace/gateway"
+
+	DockerImageUser  int64 = 1001
+	DockerImageGroup int64 = 1001
+)
+
+var (
+	ContainerSecurityContext = corev1.SecurityContext{
+		Privileged:               address.Of(false),
+		AllowPrivilegeEscalation: address.Of(false),
+		RunAsNonRoot:             address.Of(true),
+		RunAsUser:                address.Of(DockerImageUser),
+		RunAsGroup:               address.Of(DockerImageGroup),
+		Capabilities: &corev1.Capabilities{
+			Drop: []corev1.Capability{
+				"ALL",
+			},
+		},
+		SeccompProfile: &corev1.SeccompProfile{
+			Type: corev1.SeccompProfileTypeRuntimeDefault,
+		},
+		ReadOnlyRootFilesystem: address.Of(false),
+	}
 )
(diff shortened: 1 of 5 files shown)


src/controllers/dynakube/activegate/consts/consts.go
```diff
@@ -59,4 +61,26 @@ const (
 	GatewaySslMountPoint     = "/var/lib/dynatrace/gateway/ssl"
 	GatewayLogMountPoint     = "/var/log/dynatrace/gateway"
 	GatewayTmpMountPoint     = "/var/tmp/dynatrace/gateway"
+
+	DockerImageUser  int64 = 1001
+	DockerImageGroup int64 = 1001
+)
+
+var (
+	ContainerSecurityContext = corev1.SecurityContext{
```
luhi-DT, line 70: I would put that into a method of the kubemon modifier like `getSecurityContext()`

src/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/kubemon.go
```diff
@@ -72,11 +73,11 @@ func (mod KubernetesMonitoringModifier) getInitContainers() []corev1.Container {
 			Args:            []string{"-c", k8scrt2jksPath},
 			VolumeMounts:    volumeMounts,
 			Resources:       mod.capability.Properties().Resources,
-			SecurityContext: &corev1.SecurityContext{
-				ReadOnlyRootFilesystem: &readOnlyRootFs,
-			},
+			SecurityContext: consts.ContainerSecurityContext.DeepCopy(),
```
luhi-DT, line 76: Did you test if it is fine using the same security context on the container as on the pod?

luhi-DT (changes requested):

src/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/kubemon.go
aorcholski: Only some of ContainerSecurityContext settings can be also set on the pod level
```
	PodSecurityContext = corev1.PodSecurityContext{
		// Privileged:               address.Of(false),
		// AllowPrivilegeEscalation: address.Of(false),
		RunAsNonRoot: address.Of(true),
		RunAsUser:    address.Of(DockerImageUser),
		RunAsGroup:   address.Of(DockerImageGroup),
		// Capabilities: &corev1.Capabilities{
		//	Drop: []corev1.Capability{
		//		"ALL",
		//	},
		// },
		SeccompProfile: &corev1.SeccompProfile{
			Type: corev1.SeccompProfileTypeRuntimeDefault,
		},
		//ReadOnlyRootFilesystem: address.Of(false),
	}
```
so I decided to keep them all in the shared ContainerSecurityContext structure.

I have tested a case with `PodSecurityContext` defined and istio enabled. Everything is fine (istio containers define own SecurityContexts with `runAsUser: 0` or `runAsUser: 1337`).

src/controllers/dynakube/activegate/consts/consts.go
```diff
@@ -59,4 +59,7 @@ const (
 	GatewaySslMountPoint     = "/var/lib/dynatrace/gateway/ssl"
 	GatewayLogMountPoint     = "/var/log/dynatrace/gateway"
 	GatewayTmpMountPoint     = "/var/tmp/dynatrace/gateway"
+
+	DockerImageUser  int64 = 1001
+	DockerImageGroup int64 = 1001
```
0sewa0, line 64: just curios, why is this 1001 while the agent is 1000 ?

aorcholski: no idea

src/controllers/dynakube/activegate/consts/consts.go
aorcholski: done

0sewa0 (approved):

luhi-DT (approved):

PR #1496, opened 2023-01-20
Adds a switch to control cleanup behavior after E2E test

StefanHauth:
# Description
It often is hard to analyze failing E2E tests when the environment they failed in is already cleaned up. This PR adds a switch that allows fine tuning cleanup behavior via an environment variable:

`TEST_ENV_CLEANUP`
* `always` - cleanup after each test
* `never` - never clean up
* `onSuccess` - only cleanup if test was successful

## How can this be tested?
* Prepare an environment to run E2E tests in
* set said environment variable to a value of you choice and run E2E tests
* observe if cleanup has happened the way you expected by checking resources on the test cluster

## Checklist

### DIFF
--- a/test/kubeobjects/environment/environment.go
+++ b/test/kubeobjects/environment/environment.go
@@ -3,27 +3,114 @@
 package environment
 
 import (
+	"context"
 	"os"
+	"testing"
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/e2e-framework/klient/conf"
 	"sigs.k8s.io/e2e-framework/pkg/env"
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 	"sigs.k8s.io/e2e-framework/pkg/envfuncs"
+	"sigs.k8s.io/e2e-framework/pkg/features"
 )
 
 const (
-	useKind = "TEST_ENV_USE_KIND"
+	useKindEnvVar = "TEST_ENV_USE_KIND"
+	cleanupEnvVar = "TEST_ENV_CLEANUP"
 )
 
-func Get() env.Environment {
-	if os.Getenv(useKind) == "true" {
+type CleanupMode string
+
+const (
+	cleanupAlways      CleanupMode = "always"
+	cleanupOnSuccess               = "onSuccess"
+	cleanupNever                   = "never"
+	cleanupModeDefault             = cleanupAlways
+)
+
+type Environment struct {
+	env.Environment
+	skipCleanup CleanupMode
+	t           *testing.T
+}
+
+type CleanupFunction func(ctx context.Context, envconf *envconf.Config, t *testing.T) (context.Context, error)
+
+func Get() *Environment {
+	if os.Getenv(useKindEnvVar) == "true" {
 		environment := env.New()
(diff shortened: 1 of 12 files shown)


test/kubeobjects/environment/environment.go
```diff
@@ -3,27 +3,126 @@
...
+	return &Environment{
+		Environment: env.NewWithConfig(envConfig),
+		skipCleanup: getSkipCleanupMode(),
+	}
+}
+
+type TestingT interface {
+	Failed() bool
+	Errorf(format string, args ...interface{})
+}
```
0sewa0, line 63: why is this needed?

0sewa0 (changes requested):
this looks like we are waaay overcomplicating things 😅 

just a question wouldn't the following be enough:
```
var testEnvironment env.Environment

func TestMain(m *testing.M) {
	testEnvironment = environment.Get()
	testEnvironment.BeforeEachTest(namespace.DeleteIfExists(sampleapps.Namespace))
	testEnvironment.BeforeEachTest(namespace.DeleteIfExists(dynakube.Namespace))
	testEnvironment.BeforeEachTest(oneagent.WaitForDaemonSetPodsDeletion())
	testEnvironment.BeforeEachTest(namespace.Recreate(namespace.NewBuilder(dynakube.Namespace).Build()))

	if os.Getenv("CLEANUP") == "true" {
		testEnvironment.AfterEachTest(namespace.DeleteIfExists(sampleapps.Namespace))
		testEnvironment.AfterEachTest(dynakube.DeleteIfExists(dynakube.NewBuilder().WithDefaultObjectMeta().Build()))
		testEnvironment.AfterEachTest(oneagent.WaitForDaemonSetPodsDeletion())
		testEnvironment.AfterEachTest(namespace.Delete(dynakube.Namespace))
	}
	testEnvironment.Run(m)
}
```

you could abstract the `os.Getenv("CLEANUP") == "true" ` part to be nicer, but that is it

the `onSuccess` state does not make sense (except 1 scenario), because:
Scenarios you would like to cover:
1. Run ALL tests in ONE pipeline, pointing to the SAME cluster (this is what is the current plan)
In this case you would like to clean up if:
- The cluster you get in the beginning is brand spanking new, no need to pre-clean
- The test passed, and want to run the next test
- The test failed,  and want to run the next test
- **SUMMERY: So always after a test, no ifs or buts (you could argue about the last test, but it doesn't make sense)**
2. Run EACH tests in DIFFERENT pipelines, pointing to DIFFERENT clusters (this is what is the "old" plan AFAIK)
In this case you would like to clean up if:
- The cluster you get in the beginning is brand spanking new, no need to pre-clean
- **SUMMERY:  Never, if test passed cluster goes bye bye, if not, we dont clean up**
3. Run locally, on your special cluster.
- It depends, here `onSuccess` would be a nice to have, but IMO this complex change does not worth the minor connivance it would bring

A would also question the necessity of the `testEnvironment.BeforeEachTest` calls, as they are not needed in the pipeline, and in a local scenario you CAN'T predict every kind of weird stuff that one might have installed a week or so ago in their environment, so it can easily occur that the pre-clean is not good enough, so it adds very little value.

StefanHauth:
Feature is no longer needed, at least in the intended way. Back to the drawing board

PR #1497, opened 2023-01-23
Fix failing RHCC image preflight scan upload

gkrenn:
# Description

Preflight for the RHCC image fails on upload due to the checks running before the upload.
Fixed by removing the artifacts after generation. 

## How can this be tested?

Run the publish image action.

## Checklist

### DIFF
--- a/hack/build/ci/preflight.sh
+++ b/hack/build/ci/preflight.sh
@@ -20,7 +20,7 @@ check_image() {
     1> "${PREFLIGHT_REPORT_NAME}" 2> "${PREFLIGHT_LOG}"
   echo "${PREFLIGHT_EXECUTABLE} returned ${?}"
   cat "${PREFLIGHT_LOG}"
-  rm -rf artifacts preflight.log
+  rm -rf artifacts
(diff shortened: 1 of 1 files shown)


luhi-DT (dismissed):

hack/build/ci/preflight.sh
```diff
@@ -20,6 +20,7 @@ check_image() {
     1> "${PREFLIGHT_REPORT_NAME}" 2> "${PREFLIGHT_LOG}"
   echo "${PREFLIGHT_EXECUTABLE} returned ${?}"
   cat "${PREFLIGHT_LOG}"
+  rm -rf artifacts preflight.log
```
chrismuellner, line 23:
```suggestion
  rm -rf artifacts "${PREFLIGHT_LOG}"
```

can we move this into a `cleanup()` function so the intention is clearer?

chrismuellner (dismissed):

hack/build/ci/preflight.sh
gkrenn: This line changed to only `rm -rf artifacts` which should be reasonable and does not require an own function.

luhi-DT (approved):

PR #1501, opened 2023-01-24
Makes image build faster for local envs

0sewa0:
# Description

The Dockerfile was restructured in a way the any downloading we do is cached locally so we don't have to redownload the dependencies every time. (only if something changes)

~Also made the build focused `make` targets to have a `nocache` variant, if somebody wants to use that.~
~`images/push/nocache`~
~`images/push/tagged/nocache`~

Removes the cache setup step in the CI as it does not work for the current Dockerfile.
Removes the `.dockerignore` file because we only use exact paths when we use the `COPY` command. (no `COPY .` is allowed)

### DISCLAIMER
Potential question: Why didn't you do something like: (AKA.: use less build stages)
```
...
FROM go-base AS operator-build
ARG GO_LINKER_ARGS
WORKDIR /app
COPY go.mod go.sum ./
RUN go mod download && go mod verify
COPY src ./src
RUN CGO_ENABLED=1 CGO_CFLAGS="-O2 -Wno-return-local-addr" \ ...
...
```
I tried, but in that case the `RUN go mod download && go mod verify` never got cached, I don't know why, according to the docker docs it should be cached, but it just never happened. (same for the `RUN apt-get ...` stage)

## How can this be tested?
Build the image locally (with cache) 2 times, and check if it caches all the downloading parts on the 2. build.

## Checklist


hack/build/ci/download-go-build-deps.sh
```diff
@@ -6,5 +6,5 @@ if ! [ -d ./third_party_licenses ]; then
 fi
 
 # fetch dependencies
-go get -d ./...
+go get -d ./...  # <-- WHY???
```
luhi-DT, line 9: on purpose? :)

chrismuellner: `./...` is used to get everything in sub directories as well, see [go docs](https://pkg.go.dev/cmd/go#hdr-Relative_import_paths), so it should be correct.

I don't think we need this script anymore at all, since we do `go mod download` as part of the `Dockerfile` -> `download-go-build-deps.sh` can be removed

Dockerfile
```diff
@@ -1,20 +1,28 @@
...
+COPY go.mod go.sum ./
+RUN go mod download && go mod verify
 
-RUN apt-get update && \
-    apt-get install -y libbtrfs-dev libdevmapper-dev
-
+# build operator binary
+FROM go-mod AS operator-build
+ARG GO_LINKER_ARGS
+COPY src ./src
```
chrismuellner, line 16: we don't copy the `third_party_licenses` folder anymore, which is created as part of `push_image.sh`.

We can't drop these licenses, so it should be installed and cached as part of the `Dockerfile` 🙂

Dockerfile
```diff
@@ -1,20 +1,28 @@
...
 RUN CGO_ENABLED=1 CGO_CFLAGS="-O2 -Wno-return-local-addr" \
     go build -tags "containers_image_openpgp,osusergo,netgo,sqlite_omit_load_extension,containers_image_storage_stub,containers_image_docker_daemon_stub" -trimpath -ldflags="${GO_LINKER_ARGS}" \
     -o ./build/_output/bin/dynatrace-operator ./src/cmd/
 
+# get extra dependencies
```
chrismuellner, line 21: not descriptive, suggest "download additional image dependencies"

Dockerfile
```diff
@@ -1,20 +1,28 @@
...
 FROM registry.access.redhat.com/ubi9-minimal:9.1.0 as dependency-src
-
-RUN microdnf install -y util-linux tar --nodocs
+RUN \
+    --mount=type=cache,target=/var/cache/dnf \
+    microdnf install -y util-linux tar --nodocs
 
 FROM registry.access.redhat.com/ubi9-micro:9.1.0
```
chrismuellner, line 28: line 32 has 2 `#` it seems, probably incorrect

hack/build/push_image.sh
```diff
@@ -13,10 +13,18 @@ out_image="${IMG:-quay.io/dynatrace/dynatrace-operator}:${TAG}"
 
 # directory required by docker copy command
 mkdir -p third_party_licenses
-docker build . -f ./Dockerfile -t "${base_image}" \
+
+if [[ "${NO_CACHE}" ]]; then
```
chrismuellner, line 17: is there ever a reason to run this without cache?
if yes, would it make sense to split `push_image` into `build_image` and `push_image`, with additional parameters possible for `push_image` (not just 1 very specific one)?

hack/make/images.mk
```diff
@@ -24,10 +24,16 @@ endif
 images/push:
 	./hack/build/push_image.sh
 
+images/push/nocache: export NO_CACHE=true
```
chrismuellner, line 27: does this need to be a dedicated make step?

chrismuellner (changes requested):

hack/build/ci/download-go-build-deps.sh
0sewa0: only the `go get` part as the licenses are still needed

hack/build/push_image.sh
0sewa0: undoing this change, and making this nicer in a different PR

hack/build/push_image.sh
```diff
@@ -14,9 +14,9 @@ out_image="${IMG:-quay.io/dynatrace/dynatrace-operator}:${TAG}"
 # directory required by docker copy command
 mkdir -p third_party_licenses
 docker build . -f ./Dockerfile -t "${base_image}" \
-  --build-arg "GO_LINKER_ARGS=${go_linker_args}" \
-  --label "quay.expires-after=14d" \
-  --no-cache
+--build-arg "GO_LINKER_ARGS=${go_linker_args}" \
+--label "quay.expires-after=14d" \
+--no-cache
```
chrismuellner, line 19: remove please, otherwise the layers will never be cached

chrismuellner (approved):

PR #1508, opened 2023-01-27
Refactor how the images work in the makefile and related scripts

0sewa0:
# Description
Simplifies a few things related to the makefile:
**I. The confusing way how we set the image/tag that is used during the makefile scripts**

No we have only 3(+1) things that can be set, or set is automatically:
  **+1/ `BRANCH`: The branch of the repo. (your current branch by default)**
         - pipelines don't have the actual git branches/history so we have to set this there manually
    **1/ `IMAGE`: The url to the image, without the tag (`quay...` by default)**
         - we tend to overwrite this in the release-pipeline
    **2/ `TAG`: The tag for the image  (`snapshot` or `snapshot-<sanitised version of $BRANCH>` by default )**
         - we tend to overwrite this in the release-pipeline
    **3/ `IMAGE_URI`: The combination of `$IMAGE`:`$TAG`**
         - just for easier use

**II. Removes unused kuttl scripts**
`kuttl` has not been used for a while

**III. Removes unused operator e2e installation code**
The code is related to installing the operator using the manifests, this is no longer used.

**IV. Removes generated `config/deploy` folder from git**
We still generate the stuff under `config/deploy` but we no longer track it in git.

**MISC**
Used dependencies in the makefiles in a way that we don't do the same thing multiple times, or do things that are not necessary. (sometimes we generated the manifests multiple times, or generated all the manifests when we only needed to generate the CRD) 

## How can this be tested?
Run all the `make` commands that you use, if anything breaks let me know. 😉 

## Checklist


hack/build/build_image.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+if [[ ! "${1}" ]]; then
+  echo "1 param is not set, should be the image without the tag"
```
waodim, line 4: nitpick: maybe rephrase to `"first param..." `instead of `"1 param..."`?

hack/build/build_image.sh
```diff
@@ -0,0 +1,28 @@
...
+  exit 5
+fi
+if [[ ! "${2}" ]]; then
+  echo "2 param is not set, should be the tag of the image"
```
waodim, line 8: same as above

hack/build/push_image.sh
```diff
@@ -1,23 +1,20 @@
 #!/bin/bash
 
-if [[ ! "${TAG}" ]]; then
-  echo "TAG variable not set"
-  echo "usage: 'make deploy-local TAG=\"<your-image-tag>\"' or 'make deploy-local-easy'"
+if [[ ! "${1}" ]]; then
+  echo "1 param is not set, should be the image without the tag"
```
waodim, line 4: same as above

hack/build/push_image.sh
```diff
@@ -1,23 +1,20 @@
...
   exit 5
 fi
+if [[ ! "${2}" ]]; then
+  echo "2 param is not set, should be the tag of the image"
```
waodim, line 8: same as above

hack/build/push_image.sh
```diff
@@ -1,23 +1,20 @@
...
---label "quay.expires-after=14d" \
-
-rm -rf third_party_licenses
-
-docker tag "${base_image}" "${out_image}"
+out_image="${IMAGE}${TAG}"
+is_image_present=$(docker images | grep "${out_image}")
+if [ "${is_image_present}" != "" ]; then
+   echo "image is not present, please build it first"
+   exit 5
```
chrismuellner, line 18: why `5`?

hack/build/push_image.sh
```diff
@@ -1,23 +1,20 @@
...
-mkdir -p third_party_licenses
-docker build . -f ./Dockerfile -t "${base_image}" \
---build-arg "GO_LINKER_ARGS=${go_linker_args}" \
---label "quay.expires-after=14d" \
-
-rm -rf third_party_licenses
-
-docker tag "${base_image}" "${out_image}"
+out_image="${IMAGE}${TAG}"
+is_image_present=$(docker images | grep "${out_image}")
```
chrismuellner, line 15: we could use`docker images --type=image <image>` and check for non-zero exit code instead of doing a string comparison.

hack/make/images.mk
```diff
@@ -1,33 +1,25 @@
-MAIN_IMAGE ?= quay.io/dynatrace/dynatrace-operator:snapshot
+IMAGE ?= quay.io/dynatrace/dynatrace-operator
+#Needed for the e2e pipeline to work
+BRANCH ?= $(shell git branch --show-current)
+SNAPSHOT_SUFFIX ?= $(shell echo "${BRANCH}" | sed "s/[^a-zA-Z0-9_-]/-/g")
+ifneq ($(BRANCH), main)
+	TAG ?= snapshot-${SNAPSHOT_SUFFIX}
+else
+	TAG ?= snapshot
+endif
```
chrismuellner, line 9: this is a lot nicer to read than before

hack/make/manifests/config.mk
```diff
@@ -27,7 +27,7 @@ ifneq ($(shell git branch --show-current | grep "^release-"),)
 	endif
 else ifeq ($(shell git branch --show-current), main)
 	# if the current branch is the main branch
-	CHART_VERSION=snapshot
+	CHART_VERSION=0.0.0-snapshot
```
chrismuellner, line 30:
```suggestion
	CHART_VERSION=snapshot
```

hack/make/images.mk
```diff
@@ -1,33 +1,25 @@
...
-endif
+## Pushes an ALREADY BUILT Operator image with a given IMAGE and TAG
+images/build:
+	./hack/build/build_image.sh "${IMAGE}" "${TAG}"
 
-## Builds and pushes an Operator image with a given TAG
+## Builds an Operator image with a given IMAGE and TAG
 images/push:
-	./hack/build/push_image.sh
+	./hack/build/push_image.sh "${IMAGE}" "${TAG}"
```
chrismuellner, line 19: why do we need to pass environment variables?

hack/make/bundle.mk
```diff
@@ -4,6 +4,8 @@ VERSION ?= 0.0.1
 PLATFORM ?= openshift
 # Needed variable for manifest generation
 OLM ?= false
+# Default bundle image tag
```
chrismuellner, line 7:
```suggestion
# Default bundle image
```

chrismuellner (changes requested):

hack/build/push_image.sh
aorcholski: --type ?
```
[] docker images --type <image>
unknown flag: --type
```

hack/make/images.mk
aorcholski: IMAGE and TAG are local variables (not seen out of the make script process).

hack/build/push_image.sh
chrismuellner: it's `docker inspect --type=image <image>`

hack/make/manifests/config.mk
0sewa0: this is needed for helm, otherwise helm doesn't recognise it as a version

hack/make/images.mk
0sewa0: its also cleaner IMO (it shows what env is used by what script)

hack/build/push_image.sh
0sewa0: it was like that , I can change it 😅

PR #1512, opened 2023-01-31
Skip unnecessary requests to tenant within 15 min window

gkrenn:
# Description

On dynakube reconciliation, request to the tenant are made that are unnecessary, as the data won't change that often. 
This PR adds a timeout, within these request will be skipped.

**Important:** For this improvement to work, the CRD has to be updated, as the lastAccess timestamps are new fields in the status object. - In case the CRDs are not updated, we skip back to the previous default, sending requests on each reconciliation.

## How can this be tested?

- Run multiple reconciliations and check if connections to the tenant are skipped.

## Checklist


src/api/v1beta1/dynakube_status.go
```diff
@@ -54,9 +56,29 @@ type DynaKubeStatus struct {
...
+	ActiveGate   ActiveGateStatus   `json:"activeGate,omitempty"`
+	OneAgent     OneAgentStatus     `json:"oneAgent,omitempty"`
+	Synthetic    SyntheticStatus    `json:"synthetic,omitempty"`
+	DynatraceApi DynatraceApiStatus `json:"dynatraceApi,omitempty"`
+}
+
+const MaxRequestInterval = 15 * time.Minute
+
+type DynatraceApiStatus struct {
+	LastTokenProbe                     metav1.Time `json:"lastTokenProbe,omitempty"`
```
waodim, line 69: suggestion: out of consistency maybe rename this to `LateTokenProbeTimestamp`?

src/api/v1beta1/dynakube_status.go
```diff
@@ -54,9 +57,38 @@ type DynaKubeStatus struct {
...
+	LastOneAgentConnectionInfoUpdate   metav1.Time `json:"lastOneAgentConnectionInfoUpdate,omitempty"`
+	LastActiveGateConnectionInfoUpdate metav1.Time `json:"lastActiveGateConnectionInfoUpdate,omitempty"`
+	LastAuthTokenSecretUpdate          metav1.Time `json:"lastAuthTokenSecretUpdate,omitempty"`
+	LastBasicStatusInformationUpdate   metav1.Time `json:"lastBasicStatusInformationUpdate,omitempty"`
```
0sewa0, line 73: not necessary, because that part of the status will be reworked, we will worry about this when we get there

src/api/v1beta1/dynakube_status.go
0sewa0: IMO `Timestamp` is redundant
but the rest of the names could be more consistent. (replace `...Update` with `Request` and then it could be called `LastTokenScopeRequest`)

src/api/v1beta1/dynakube_status.go
```diff
@@ -54,9 +57,38 @@ type DynaKubeStatus struct {
...
+}
+
+func (dynatraceApiStatus *DynatraceApiStatus) ResetCachedTimestamps() {
+	dynatraceApiStatus.LastTokenProbe = metav1.Time{}
+	dynatraceApiStatus.LastOneAgentConnectionInfoUpdate = metav1.Time{}
+	dynatraceApiStatus.LastActiveGateConnectionInfoUpdate = metav1.Time{}
+	dynatraceApiStatus.LastAuthTokenSecretUpdate = metav1.Time{}
+	dynatraceApiStatus.LastBasicStatusInformationUpdate = metav1.Time{}
+}
```
0sewa0, line 82: move to `..._test.go` file

src/api/v1beta1/dynakube_status.go
```diff
@@ -109,8 +141,7 @@ type OneAgentStatus struct {
 
 	Instances map[string]OneAgentInstance `json:"instances,omitempty"`
 
-	// LastHostsRequestTimestamp indicates the last timestamp the Operator queried for hosts
-	LastHostsRequestTimestamp *metav1.Time `json:"lastHostsRequestTimestamp,omitempty"`
+	LastInstanceStatusUpdateTimestamp *metav1.Time `json:"lastInstanceStatusUpdateTimestamp,omitempty"`
```
0sewa0, line 144:
```suggestion
	LastInstanceStatusUpdate *metav1.Time `json:"lastInstanceStatusUpdateTimestamp,omitempty"`
```

src/controllers/csi/provisioner/controller.go
```diff
@@ -178,6 +178,7 @@ func (provisioner *OneAgentProvisioner) updateAgentInstallation(ctx context.Cont
 	var agentUpdater *agentUpdater
 	if dk.CodeModulesImage() != "" {
 		connectionInfo, err := dtc.GetOneAgentConnectionInfo()
+
```
0sewa0, line 181:
```suggestion
```

src/controllers/dynakube/connectioninfo/reconciler.go
```diff
@@ -59,10 +65,17 @@ func (r *Reconciler) reconcileOneAgentConnectionInfo() error {
 	if err != nil {
 		return err
 	}
+
+	r.dynakube.Status.DynatraceApi.LastOneAgentConnectionInfoUpdate = metav1.Now()
 	return nil
 }
 
 func (r *Reconciler) reconcileActiveGateConnectionInfo() error {
+	if !dynatracev1beta1.IsRequestOutdated(r.dynakube.Status.DynatraceApi.LastActiveGateConnectionInfoUpdate) {
```
0sewa0, line 74: is this actually set 😅 ?

0sewa0 (changes requested):

src/api/v1beta1/dynakube_status.go
gkrenn: I agree that it would be more convenient to move this function to a test file, as is only used for testing. 
Sadly using functions declared in test files are not available to other packages test files, which is bad as we would need to have code duplication

src/controllers/dynakube/connectioninfo/reconciler.go
gkrenn: Nope 🙃

0sewa0 (approved):

PR #1515, opened 2023-02-02
Makes the e2e run_all.sh exit with none zero if a test failed

0sewa0:
# Description

The `hack/e2e/run_all.sh` now counts the amount of errors that happen and returns with exit code of the number errors that happened. This is needed because otherwise if the last test does not fail, then the command will exit with 0 which is a success in the pipeline, which we don't want

## How can this be tested?
Run the command 😄 

## Checklist


StefanHauth (approved):
LGTM I believe in you

PR #1517, opened 2023-02-02
Extends troubleshoot subcommand to check for OneAgentAPM objects

waodim:
# Description
With this the troubleshoot subcommand will check for possible existing `OneAgentAPM` objects. Also the existing method `kubeobjects.CheckIfOneAgentAPMExists` is refactored and takes now a client instead of a config as a parameter.

## How can this be tested?
Deploy the operator and simply run the troubleshoot subcommand

## Checklist


src/cmd/troubleshoot/oneagentapm.go
```diff
@@ -0,0 +1,26 @@
+package troubleshoot
+
+import (
+	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func checkOneAgentAPM(clt client.Client, ctx *troubleshootContext) error {
```
0sewa0, line 9: the `troubleshootContext` has a client in it

0sewa0 (changes requested):

src/kubeobjects/legacy.go
```diff
@@ -1,27 +1,27 @@
...
-	client, err := discovery.NewDiscoveryClientForConfig(cfg)
-	if err != nil {
-		return false, err
-	}
-	_, resourceList, err := client.ServerGroupsAndResources()
-	if err != nil {
+func CheckIfOneAgentAPMExists(clt client.Reader) (bool, error) {
+	var crd apiv1.CustomResourceDefinition
+
+	err := clt.Get(context.TODO(), client.ObjectKey{Name: crdName}, &crd)
```
chrismuellner, line 16: Does this work the same way now? It was a discovery client that checks for existence of specific APIs, but uses a normal client reader now. Check could also be done the same way as `checkCRD`

src/cmd/troubleshoot/oneagentapm.go
```diff
@@ -0,0 +1,25 @@
...
+	logNewCheckf("checking if OneAgentAPM object exists ...")
+
+	exists, err := kubeobjects.CheckIfOneAgentAPMExists(ctx.apiReader)
+
+	if err != nil {
+		return err
+	}
+
+	if exists {
+		return errors.New("OneAgentAPM object still exists")
```
chrismuellner, line 20: more information with some points how to fix the problem could be helpful here

chrismuellner (dismissed):

src/cmd/troubleshoot/oneagentapm.go
waodim: thanks, is there another way than to delete those objects?

chrismuellner: oneagent operator needs to be fully uninstalled by the customer

src/kubeobjects/legacy.go
waodim: I now adapted the change and made it similar to `checkCRD`, with the difference that I use `apiv1.CustomResourceDefinitionList` instead of `apiv1.DynakubeList.` So I think this should work as intended.

luhi-DT: Why not reuisng that function to be able to check any crd? Just passing the crd we want to check as parameter we could save one function and just have one that can check all the CRDs?

PR #1518, opened 2023-02-02
Adds latest available public registry image endpoints

waodim:
# Description
With this three new endpoints are added to the `dtClient`. These endpoints enables the operator getting `oneAgent/codeModules/activeGate` images which are compatible.

## Checklist


src/dtclient/image.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	latestImageInfo, err := dtc.readResponseForLatestImage(data)
+	if err != nil {
+		return nil, err
+	}
+	return latestImageInfo, err
+}
+
+func (dtc *dynatraceClient) createLatestAgentImageRequest() (*http.Request, error) {
```
0sewa0, line 99: you could generalise these `createLatest...ImageRequest`, because the only thing that changes is the URL, so you could pass that in.
```
func (dtc *dynatraceClient) createLatestImageRequest(url string) (*http.Request, error) {
```

0sewa0 (changes requested):

chrismuellner (dismissed):

src/dtclient/image.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if err != nil {
+		log.Info("failed to process latest image response")
+		return nil, err
+	}
+
+	return latestImageInfo, nil
+}
+
+func (dtc *dynatraceClient) processRequest(url string) (*LatestImageInfo, error) {
```
0sewa0, line 49: `processRequest` is a bit too generic, so maybe something like:
```suggestion
func (dtc *dynatraceClient) processLatestImageRequest(url string) (*LatestImageInfo, error) {
```

0sewa0 (approved):

PR #1542, opened 2023-02-14
Restructure e2e tests

0sewa0:
# Description

Reorgs the e2e tests of the codebase.
The main goal was:
- Remove hardcoded values
- Generalise the code
- Make the code more readable
- Make the structure less "flat"
- Make sample apps more reusable

Sidenotes:
- There are still things that need to be fixed(example: tests that need a service-mesh), but this PR is already waaaaaay too big
- Lots of things were just moved around, so not all stuff is new that git says is new

## How can this be tested?
Run any/all the e2e tests

## Checklist


test/helpers/components/dynakube/dynakube.go
```diff
@@ -140,24 +141,20 @@ func (dynakubeBuilder Builder) Build() dynatracev1beta1.DynaKube {
 	return dynakubeBuilder.dynakube
 }
 
-func Apply(dynakube dynatracev1beta1.DynaKube) features.Func {
+func Create(dynakube dynatracev1beta1.DynaKube) features.Func {
```
luhi-DT, line 144: Just a quick input: wouldn't it be better to just return a dynakube here and apply it where the Create() is called?

test/helpers/components/oneagent/daemonset.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func Get(ctx context.Context, resource *resources.Resources, dynakube dynatracev1beta1.DynaKube) (appsv1.DaemonSet, error) {
+	return daemonset.NewQuery(ctx, resource, client.ObjectKey{
+		Name:      dynakube.OneAgentDaemonsetName(),
+		Namespace: dynakube.Namespace,
+	}).Get()
+}
+
+func ForEachPod(ctx context.Context, resource *resources.Resources, dynakube dynatracev1beta1.DynaKube, consumer daemonset.PodConsumer) error {
```
luhi-DT, line 32: maybe renaming consumer to action would help to get an idea, that the passed PodConsumer is executed

test/helpers/components/operator/installation.go
```diff
@@ -70,6 +83,11 @@ func InstallFromGithub(releaseTag string, withCsi bool) features.Func {
 	return manifests.InstallFromUrls(manifestsUrls)
 }
 
+func UninstallFromGithub(releaseTag string, withCsi bool) features.Func {
```
luhi-DT, line 86: Function is not used, UninstallOperatorFromRelease() uses this function but is never called anywhere else

test/helpers/sampleapps/curl.go
```diff
@@ -0,0 +1,187 @@
...
+		logStream := getCurlPodLogStream(ctx, t, resources, testDynakube)
+
+		logs.AssertContains(t, logStream, "RUNNING")
+
+		return ctx
+	}
+}
+
+func InstallWebhookCurlProxyPod(testDynakube dynatracev1beta1.DynaKube) features.Func {
+	return func(ctx context.Context, t *testing.T, environmentConfig *envconf.Config) context.Context {
```
luhi-DT, line 101: Can this be combined with the same function from InstallActiveGaterCurlProxyPod() looks very much the same tbh

test/scenarios/activegate/installation.go
```diff
@@ -30,12 +29,9 @@ import (
 )
 
 const (
-	agNamespace = "dynatrace"
-	agPodName   = "dynakube-activegate-0"
+	agPodName = "dynakube-activegate-0"
```
luhi-DT, line 32: we remove this and check for the actual dynakube name, since the ag pod is named like the dynakube plus the -activegate-0 suffix

test/scenarios/applicationmonitoring/dataingest.go
```diff
@@ -43,42 +39,49 @@ type metadata struct {
...
 		WithDefaultObjectMeta().
-		ApiUrl(tenantSecret.ApiUrl).
-		ApplicationMonitoring(&v1beta1.ApplicationMonitoringSpec{
+		ApiUrl(secretConfig.ApiUrl).
+		ApplicationMonitoring(&dynatracev1beta1.ApplicationMonitoringSpec{
 			UseCSIDriver: address.Of(false),
 		}).Build()
 
-	require.NoError(t, err)
+	sampleDeployment := sampleapps.NewSampleDeployment(t, testDynakube)
```
luhi-DT, line 51: why do we create a deployment and a pod at the same time? Shouldnt the deployment handle this alone without the need to create a standalone pod? or is this intended to check for pods, that does not have owner references?

luhi-DT (changes requested):

test/scenarios/applicationmonitoring/dataingest.go
luhi-DT: thats what my 2nd thought was, thanks

test/helpers/sampleapps/curl.go
0sewa0: well there were more problems here, fixed now :D

luhi-DT (approved):

PR #1543, opened 2023-02-14
new examples for arm architecture, multi-architecture and multi-dynakube environments

kyledharrington:
# new examples for arm architecture, multi-architecture and multi-dynakube environments

1. hybrid-fullStack-hostMonitoring.yaml
    - Full stack & host monitoring environment - ideally used for on-prem environments with multiple node pools
2. classicFullStack-ARM.yaml
     - example reference for full stack in ARM architecture based environments. Active gate has been removed.
3.  classicFullStack-MultiArch.yaml
    - example reference for hybrid clusters running multiple architectures. 

- This provides documentation for arm based cluster deployments, which does not currently exist
- Many many Dynatrace users are moving workloads over to ARM based architecture for cloud cost savings

## How can this be tested?

-  Deploying the ARM dynakube samples can be tested by deploying to k8s cluster with an arm node pool
- Toleration for standard Kubernetes architecture labels have been added to account for ARM node taints
- node selectors have been added for standard Kubernetes architecture node annotations for ARM & AMD

- Deploying the hybrid dynakube samples requires separate node pools; one with labels for infra-only desired node deployments, and one with full stack desired node deployments. The labels in use are placeholder examples (full=stack & infra=only), as this yaml stands, these labels would need to be added to their respective node pools.

### DIFF
--- a/assets/samples/classicFullStack-ARM.yaml
+++ b/assets/samples/classicFullStack-ARM.yaml
@@ -1,0 +1,147 @@
+apiVersion: dynatrace.com/v1beta1
+kind: DynaKube
+metadata:
+  name: dynakube
+  namespace: dynatrace
+spec:
+  # Dynatrace apiUrl including the `/api` path at the end.
+  # For SaaS, set `ENVIRONMENTID` to your environment ID.
+  # For Managed, change the apiUrl address.
+  # For instructions on how to determine the environment ID and how to configure the apiUrl address, see https://www.dynatrace.com/support/help/reference/dynatrace-concepts/environment-id/.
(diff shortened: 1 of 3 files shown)


chrismuellner:
I'd prefer to not have samples documented that rely on the installer method from Dockerhub.

We will add (actual) support for multi-architecture environments with public registry support in our next feature release. This will not require separate Dynakubes to adjust the `INSTALLER_URL` for the architecture, but will use the immutable images that will be available for all Dynatrace components on Dockerhub soon. This will mean instead of configuring separate Dynakubes for each architecture, you'll just have to opt-in via a single feature flag.

Setting the `infra-only` argument is a completely separate topic that is currently being investigated and will be default for host monitoring if technically possible.

PR #1544, opened 2023-02-14
ReadOnly support for CSI Driver to make it work for Bottlerocket OS.

NatanAOliveira:
# Description

Adjustments to the mutating webhook that adds the CSI driver to the pod, giving the option to enable read only.

Using the current version, unable to enable the CSI driver due to the following error:
volumes/kubernetes.io~csi/oneagent-bin/mount/agent: permission denied
With this change it is possible to set readOnly: true for the CSI driver. The idea is not to make an interrupt switch, but to give the operator the ability to inflect read-only for the CSI driver in the Application pod.

We tested this change internally by creating a docker image with the changes and applied it to a large cluster and the pod startup time dropped dramatically.

## How can this be tested?
These changes are focused on giving effect to EKS using node groups with bottlerocket. So you need to test this change on a cluster with bottlerocket 1.11+

To enable this change, you need to build a new operator image using this branch and enable the following option in the helm chart values:
```
...
csidriver:
   enabled: true
   readOnlyFileSystem: true
...
```

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
+++ b/config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
@@ -94,6 +94,8 @@ spec:
           image: {{ include "dynatrace-operator.image" . }}
           imagePullPolicy: Always
           env:
+            - name: DT_MOUNT_READ_ONLY
+              value: "{{ .Values.csidriver.readOnlyFileSystem }}"
             - name: POD_NAMESPACE
               valueFrom:
                 fieldRef:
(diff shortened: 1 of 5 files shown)


luhi-DT:
Hi, thanks for your contribution. The topic of readOnly CSI Volumes is currently tackled internally by us, as this needs some more changes for making some writable paths for the OneAgent in order to allow him write access for logs, configs etc. I will close this PR for now and please stand by and wait for our changes to be done.

Regards,
Lukas

samoclay:
Will this be in the next available version of dynatrace-operator?

luhi-DT:
Hi, please find dates and more insights on that topic here: https://community.dynatrace.com/t5/Feedback-channel/Dynatrace-Observability-for-EKS-using-Bottlerocket-OS/m-p/204641/thread-id/2243

NatanAOliveira:
Big thanks for the explanation @luhi-DT.

PR #1546, opened 2023-02-15
Fixes internal activeGate capability reconciler tests

waodim:
# Description
With this the internal AG capability reconciler tests which where commented out are reintroduced, as there were changes in functionality.

## How can this be tested?
Run the respective tests.

## Checklist


src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,213 @@
...
+}
+
+type MockCustomPropertiesReconciler struct {
+	mock.Mock
+}
+
+func (m *MockCustomPropertiesReconciler) Reconcile() error {
+	args := m.Called()
+	return args.Error(0)
+}
```
0sewa0, line 43: you don't need to create 2 specific mocks, you could just create 1 general one, and in the test create 2 different instance of it
also you could have this general mock next (in the same package) to the actual interface definition so it can be reused in the future

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,213 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testDynakube,
+			Namespace: testNamespace,
+		}}
+
+	statefulSetReconcilerMock := MockStatefulSetReconciler{}
+	statefulSetReconcilerMock.On("Reconcile").Return(nil)
+
+	custompropertiesReconcilerMock := MockCustomPropertiesReconciler{}
+	custompropertiesReconcilerMock.On("Reconcile").Return(nil)
```
0sewa0, line 83: you could use the mocks to check if they were called or not with `.AssertCalled`
(you would need to be able to access these mocks in the actual tests, which in the current form not possible)

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,213 @@
...
+	r := NewReconciler(clt, capability.NewMultiCapability(instance), instance, &statefulSetReconcilerMock, &custompropertiesReconcilerMock)
+	require.NotNil(t, r)
+	require.NotNil(t, r.client)
+	require.NotNil(t, r)
+	require.NotNil(t, r.dynakube)
+
+	return r
+}
+
+func TestReconcile(t *testing.T) {
```
0sewa0, line 94: only 2 functions test the full `Reconcile` here, the rest should have their own `Test...(t *testing.T){...}`

you could also try testing the "sad path", so when something fails and how we handle it

0sewa0 (changes requested):

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
waodim: How could those "sad path" tests look like? I tried thinking of useful tests but couldn't come up with one.

0sewa0: what happens if a reconciler fails? (ie: returns an error)
you can have mock that will always return an error

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,316 @@
...
+
+		service := corev1.Service{}
+		err = r.client.Get(context.TODO(), client.ObjectKey{Name: r.dynakube.Name + "-" + r.capability.ShortName(), Namespace: r.dynakube.Namespace}, &service)
+
+		assert.Empty(t, service)
+		assert.Error(t, err)
+	})
+}
+
+func TestReconciler_CreateOrUpdateService(t *testing.T) {
```
0sewa0, line 152:
```suggestion
func TestCreateOrUpdateService(t *testing.T) {
```

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,316 @@
...
+
+		err = r.client.Get(context.TODO(), client.ObjectKey{Name: r.dynakube.Name + "-" + r.capability.ShortName(), Namespace: r.dynakube.Namespace}, tmpService)
+		require.NoError(t, err)
+		assert.NotNil(t, service)
+
+		require.NotEqual(t, tmpService, service)
+	})
+}
+
+func TestReconcile_PortsAreOutdate(t *testing.T) {
```
0sewa0, line 237:
```suggestion
func TestPortsAreOutdate(t *testing.T) {
```

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,316 @@
...
+
+		assert.False(t, r.portsAreOutdated(service, desiredService))
+
+		service.Spec.Ports = []corev1.ServicePort{}
+
+		assert.True(t, r.portsAreOutdated(service, desiredService))
+	})
+}
+
+func TestReconcile_LabelsAreOutdated(t *testing.T) {
```
0sewa0, line 265:
```suggestion
func TestLabelsAreOutdated(t *testing.T) {
```

src/controllers/dynakube/activegate/capability/capability_test.go
```diff
@@ -1,587 +1,86 @@
...
+
+const (
+	testNamespace     = "test-namespace"
+	testName          = "test-name"
+	testApiUrl        = "https://demo.dev.dynatracelabs.com/api"
+	expectedShortName = "activegate"
+	expectedArgName   = "MSGrouter,kubernetes_monitoring,metrics_ingest,restInterface,synthetic,beacon_forwarder,beacon_forwarder_synthetic"
+)
+
+func getTestInstance() *dynatracev1beta1.DynaKube {
```
chrismuellner, line 19: could this use `getTestInstanceWithoutCapabilities()`?
maybe change to `buildDynakube(includeCapabilities bool)`

src/controllers/dynakube/activegate/capability/capability_test.go
```diff
@@ -1,587 +1,86 @@
...
+			ActiveGate: dynatracev1beta1.ActiveGateSpec{},
+		},
+	}
+}
+
+func TestBuildProxySecretName(t *testing.T) {
+	t.Run(`test if build works`, func(t *testing.T) {
+		expectedProxySecretName := "dynatrace-activegate-internal-proxy"
+		actualProxySecretName := BuildProxySecretName()
+		assert.NotEmpty(t, actualProxySecretName)
```
chrismuellner, line 55:
```suggestion
		require.NotEmpty(t, actualProxySecretName)
```
`require` should be used when it doesn't make sense to proceed with the test, e.g. compare values/contents of a variable that we already know is empty

src/controllers/dynakube/activegate/capability/capability_test.go
```diff
@@ -1,587 +1,86 @@
...
-// 			}
-// 		})
-// 	}
-// }
+import (
+	"testing"
+
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
chrismuellner, line 8:
```suggestion
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

src/controllers/dynakube/activegate/capability/capability_test.go
```diff
@@ -1,587 +1,86 @@
...
+		},
+		Spec: dynatracev1beta1.DynaKubeSpec{
+			APIURL:     testApiUrl,
+			ActiveGate: dynatracev1beta1.ActiveGateSpec{},
+		},
+	}
+}
+
+func TestBuildProxySecretName(t *testing.T) {
+	t.Run(`test if build works`, func(t *testing.T) {
```
chrismuellner, line 52:
```suggestion
	t.Run(`correct name`, func(t *testing.T) {
```
"test ..." is kind of redundant here.

chrismuellner: similar for other tests

src/controllers/dynakube/activegate/capability/capability_test.go
chrismuellner: similar for other tests

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,360 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      dynatracev1beta1.AuthTokenSecretSuffix,
+				Namespace: testNamespace,
+			},
+			Data: map[string][]byte{authtoken.ActiveGateAuthTokenName: []byte(testToken)},
+		}).
+		Build()
+}
+
+func createInstance() *dynatracev1beta1.DynaKube {
```
chrismuellner, line 46: should be `create/buildDynakube`

chrismuellner: isn't this pretty similar to the one created in `capability_test.go`?

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,360 @@
...
+	require.NotNil(t, r)
+	require.NotNil(t, r.client)
+	require.NotNil(t, r)
+	require.NotNil(t, r.dynakube)
+}
+
+func TestReconcile(t *testing.T) {
+	t.Run(`Reconciler works with multiple capabilities`, func(t *testing.T) {
+		clt := createClient()
+		instance := createInstance()
```
chrismuellner, line 99: `instance` -> `dynakube`

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,360 @@
...
+		service := corev1.Service{}
+		err = r.client.Get(context.TODO(), client.ObjectKey{Name: r.dynakube.Name + "-" + r.capability.ShortName(), Namespace: r.dynakube.Namespace}, &service)
+
+		assert.Empty(t, service)
+		assert.Error(t, err)
+	})
+}
+
+func TestCreateOrUpdateService(t *testing.T) {
+	t.Run(`createOrUpdateService works`, func(t *testing.T) {
```
chrismuellner, line 197: this is testing the create part of `createOrUpdateService`, please update the test name here

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,360 @@
...
+		require.Error(t, err)
+		assert.NotNil(t, service)
+
+		err = r.createOrUpdateService()
+		require.NoError(t, err)
+
+		service = &corev1.Service{}
+		err = r.client.Get(context.TODO(), client.ObjectKey{Name: r.dynakube.Name + "-" + r.capability.ShortName(), Namespace: r.dynakube.Namespace}, service)
+		require.NoError(t, err)
+		assert.NotNil(t, service)
```
chrismuellner, line 217: should we be asserting that the right ports were set on the service?

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,360 @@
...
+		service.Spec.Ports = []corev1.ServicePort{}
+
+		err = r.createOrUpdateService()
+		require.NoError(t, err)
+
+		err = r.client.Get(context.TODO(), client.ObjectKey{Name: r.dynakube.Name + "-" + r.capability.ShortName(), Namespace: r.dynakube.Namespace}, tmpService)
+		require.NoError(t, err)
+		assert.NotNil(t, tmpService)
+
+		require.NotEqual(t, tmpService, service)
```
chrismuellner, line 247: no equal could also mean that it was updated to an incorrect value, please compare the actual fields

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,360 @@
...
+
+		err := r.createOrUpdateService()
+		require.NoError(t, err)
+
+		service := &corev1.Service{}
+		err = r.client.Get(context.TODO(), client.ObjectKey{Name: r.dynakube.Name + "-" + r.capability.ShortName(), Namespace: r.dynakube.Namespace}, service)
+		require.NoError(t, err)
+		assert.NotNil(t, service)
+
+		tmpService := &corev1.Service{}
```
chrismuellner, line 236:
```suggestion
		actualService := &corev1.Service{}
```
declare close to where it's used

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,360 @@
...
+		err = r.client.Get(context.TODO(), client.ObjectKey{Name: r.dynakube.Name + "-" + r.capability.ShortName(), Namespace: r.dynakube.Namespace}, tmpService)
+		require.NoError(t, err)
+		assert.NotNil(t, service)
+
+		require.NotEqual(t, tmpService, service)
+	})
+}
+
+func TestPortsAreOutdated(t *testing.T) {
+	t.Run(`portsAreOutdated works`, func(t *testing.T) {
```
chrismuellner, line 282: method name is redundant, this should describe what the test is doing, e.g. "ports are detected as outdated"

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,360 @@
...
+
+		err := r.Reconcile()
+		require.NoError(t, err)
+
+		service := &corev1.Service{}
+		err = r.client.Get(context.TODO(), client.ObjectKey{Name: r.dynakube.Name + "-" + r.capability.ShortName(), Namespace: r.dynakube.Namespace}, service)
+		require.NoError(t, err)
+		assert.NotNil(t, service)
+
+		assert.False(t, r.portsAreOutdated(service, desiredService))
```
chrismuellner, line 301: this should be split into a separate `t.Run`

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
```diff
@@ -1,517 +1,360 @@
...
+func verifyReconciler(t *testing.T, r *Reconciler) {
+	require.NotNil(t, r)
+	require.NotNil(t, r.client)
+	require.NotNil(t, r)
+	require.NotNil(t, r.dynakube)
+}
+
+func TestReconcile(t *testing.T) {
+	t.Run(`Reconciler works with multiple capabilities`, func(t *testing.T) {
+		clt := createClient()
```
chrismuellner, line 98: since most of these tests reuse the same 3/4 variables, is it possible to define them before the `t.Run` or does that cause issues with writing to the same variable?

chrismuellner (changes requested):

src/controllers/dynakube/activegate/internal/capability/reconciler_test.go
waodim: thank you, i did that everywhere where it is possible :+1:

src/controllers/dynakube/activegate/capability/capability_test.go
```diff
@@ -1,587 +1,78 @@
...
+		dk := buildDynakube(capabilities)
+		mc := NewMultiCapability(dk)
+		require.NotNil(t, mc)
+		assert.True(t, mc.Enabled())
+		assert.Equal(t, expectedShortName, mc.ShortName())
+		assert.Equal(t, expectedArgName, mc.ArgName())
+	})
+	t.Run(`creates new multicapability without capabilities set in dynakube`, func(t *testing.T) {
+		var emptyCapabilites []dynatracev1beta1.CapabilityDisplayName
+		dk := buildDynakube(emptyCapabilites)
```
luhi-DT, line 71: same here

luhi-DT (approved):

PR #1547, opened 2023-02-15
Improves crd generation

0sewa0:
# Description
Adds `make` target to generate only the CRD for the `dynatrace` namespace.
Meant to be used for releases

## How can this be tested?
Try `make manifests/crd/release` and check if the `config/deploy/dynatrace-operator-crd.yaml` is correct.
Try `make manifests` and see if it works as expected. (it should produce the same as before)

## Checklist

### DIFF
--- a/config/crd/bases/dynatrace.com_dynakubes.yaml
+++ b/config/crd/bases/dynatrace.com_dynakubes.yaml
@@ -601,8 +601,10 @@ spec:
                   or from a secret with the field ''proxy'''
                 properties:
                   value:
+                    description: nolint:unused
                     type: string
                   valueFrom:
+                    description: nolint:unused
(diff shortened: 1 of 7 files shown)


config/crd/bases/dynatrace.com_dynakubes.yaml
```diff
@@ -601,8 +601,10 @@ spec:
                   or from a secret with the field ''proxy'''
                 properties:
                   value:
+                    description: nolint:unused
```
luhi-DT, line 604: Please make sure the real description of the property is added here, i guess that the first comment besides property definition in the dynakube_types.go, and is added by controller-gen automatically

luhi-DT (approved):

PR #1552, opened 2023-02-16
Change order of e2e tests by adding prefix to classsicFullStack test

luhi-DT:
# Description

Currently our e2e tests fail on Openshift, because the classicFullStack tests has some leftovers, which need to be cleaned up properly in a seperate ticket. However for now, to make the tests work we should run the classicFullstack test as last one in order to limit side issues.

## How can this be tested?
Run the e2e tests on an openshift cluster and check if they are working

## Checklist

### DIFF
--- a/hack/make/tests/e2e.mk
+++ b/hack/make/tests/e2e.mk
@@ -15,7 +15,8 @@ test/e2e/cloudnative: manifests/crd/helm
 	go test -v -tags e2e -timeout 30m -count=1 ./test/scenarios/cloudnative/basic
 
 ## Runs CloudNative e2e test only
-test/e2e/classic: manifests/crd/helm
+## TODO: rename after proper implementation of cleanup step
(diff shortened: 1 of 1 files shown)


0sewa0 (approved):

hack/make/tests/e2e.mk
```diff
@@ -15,7 +15,8 @@ test/e2e/cloudnative: manifests/crd/helm
 	go test -v -tags e2e -timeout 30m -count=1 ./test/scenarios/cloudnative/basic
 
 ## Runs CloudNative e2e test only
```
waodim, line 17: The comment indicates that CloudNative tests are run, so maybe you could update the comments as well

luhi-DT: good catch

0sewa0 (approved):

waodim (approved):

PR #1560, opened 2023-02-23
Simplify usage of maxUnmountedVolumeAge in the values.yaml

luhi-DT:
# Description
Currently our helm chart fails if we set the maxUnmountedVolumeAge to a simple string in our helm chart with the following message:

`Error: INSTALLATION FAILED: release dynatrace-operator failed, and has been uninstalled due to atomic being set: DaemonSet in version "v1" cannot be handled as a DaemonSet: json: cannot unmarshal number into Go struct field EnvVar.spec.template.spec.containers.env.value of type string `

This change makes it easier to use a simple string in the helm values file.

## How can this be tested?
Deploy the operator using helm
Set a value for maxUnmountedVolumeAge in the helm chart other than "" e.g "6"
Check if the env variable is present on the CSI Pods
Also check if the provisioner logs the current value of the env var at the top of the logs

## Checklist

### DIFF
--- a/src/controllers/csi/gc/unmounted.go
+++ b/src/controllers/csi/gc/unmounted.go
@@ -72,7 +72,9 @@ func determineMaxUnmountedVolumeAge(maxAgeEnvValue string) time.Duration {
 		return defaultMaxUnmountedCsiVolumeAge
 	}
 	if maxAge < 0 {
-		return 0
+		maxAge = 0
 	}
+
+	log.Info("GarbageCollection is using", "MaxUnmountedCsiVolumeAge", maxAge)
 	return 24 * time.Duration(maxAge) * time.Hour
 }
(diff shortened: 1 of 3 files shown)


src/controllers/csi/gc/unmounted.go
```diff
@@ -72,7 +72,9 @@ func determineMaxUnmountedVolumeAge(maxAgeEnvValue string) time.Duration {
 		return defaultMaxUnmountedCsiVolumeAge
 	}
 	if maxAge < 0 {
-		return 0
+		maxAge = 0
 	}
+
+	log.Info("GarbageCollection is using", "MaxUnmountedCsiVolumeAge", maxAge)
```
chrismuellner, line 78:
```suggestion
	log.Info("max ummounted csi volume age used", "age in days", maxAge)
```

src/controllers/csi/gc/unmounted.go
```diff
@@ -72,7 +72,9 @@ func determineMaxUnmountedVolumeAge(maxAgeEnvValue string) time.Duration {
 		return defaultMaxUnmountedCsiVolumeAge
 	}
 	if maxAge < 0 {
-		return 0
+		maxAge = 0
```
chrismuellner, line 75: we could keep the return here and add a different log here stating that files will be deleted "instantly" (on next garbage collection), since stating a max age of 0 days could be up to interpretation

chrismuellner (changes requested):

chrismuellner (approved):

PR #1584, opened 2023-03-02
Adds NO_PROXY feature-flag to the dynakube

0sewa0:
# Description
Using the `feature.dynatrace.com/no-proxy` annotation on the `DynaKube` its possible to define the NO_PROXY of the dynatrace client.

**IMPORTANT:** This only effects the communications to the **dynatrace API** in the following components:
- operator pod
- csi-driver pod
- injected init-container (standalone init)

This PR also changes/generalises the way we get/read the proxy in the codebase (there was a lot of redundant code)
Also changed some names in an old tests. (dtclient tests)

## How can this be tested?
To test it properly is a bit cumbersome, so here is a tricky test:
 
- Get a cluster where you can restrict (via a NetworkPolicy) the dynatrace namespace, so its mandatory to use a proxy.
- Deploy a dynakube with a proxy. Make sure everything works as expected.
- Add your dyanatrace tenant to the NO_PROXY
- Now nothing should work :) (because the tenant is in the NO_PROXY, it wont be routed to the proxy, so the NetworkPolicy will not allow the request to go through -> reconcile fails)

## Checklist


src/api/v1beta1/properties.go
```diff
@@ -42,7 +42,6 @@ const (
 	defaultSyntheticImage  = "/linux/dynatrace-synthetic:latest"
 
 	TrustedCAKey = "certs"
```
luhi-DT, line 44: Would it make sense to move TrustedCAKey and TLSCertKey to certs.go aswell?

src/controllers/dynakube/oneagent/daemonset/arguments.go
```diff
@@ -45,12 +45,12 @@ func (dsInfo *builderInfo) appendNetworkZoneArg(args []string) []string {
...
-	if dsInfo.dynakube != nil && dsInfo.dynakube.NeedsOneAgentProxy() {
+	if dsInfo.hasProxy() {
 		return append(args, "--set-proxy=$(https_proxy)")
 	}
 	return args
 }
 
 func (dsInfo *builderInfo) hasProxy() bool {
-	return dsInfo.dynakube != nil && dsInfo.dynakube.Spec.Proxy != nil && (dsInfo.dynakube.Spec.Proxy.ValueFrom != "" || dsInfo.dynakube.Spec.Proxy.Value != "")
+	return dsInfo.dynakube != nil && dsInfo.dynakube.NeedsOneAgentProxy()
```
luhi-DT, line 55: can we remove that hasProxy() function and just use the two expressions directly?

src/dtclient/client.go
```diff
@@ -9,14 +9,14 @@ import (
...
 
 	"github.com/pkg/errors"
+	"golang.org/x/net/http/httpproxy"
 )
 
 const (
 	DynatracePaasToken             = "paasToken"
 	DynatraceApiToken              = "apiToken"
 	DynatraceDataIngestToken       = "dataIngestToken"
 	CustomCertificatesConfigMapKey = "certs"
```
luhi-DT, line 19: is this the same const as in properties.go? If yes can we unify them?

luhi-DT (changes requested):

src/controllers/dynakube/oneagent/daemonset/arguments.go
0sewa0: IMO its cleaner this way, because the null check makes it easier to test
(also its used multiple times)

0sewa0: the `NeedsOneAgentProxy` wraps the `HasProxy` with an extra check for a feature-flag that can turn off proxy just for the oneagent

luhi-DT (approved):

PR #1585, opened 2023-03-02
Collect `.yaml` instead of `.json` manifests in support archive

luhi-DT:
# Description

Currently the support archive collects kubernetes objects as json, but we should collect them as yaml manifests, as its easier to read and a more native approach

## How can this be tested?
1. Deploy the operator
2. Run support archive `kubectl exec -n dynatrace deployment/dynatrace-operator -- /usr/local/bin/dynatrace-operator support-archive`
3. Check if the generated archive has the correct manifests stored as .yaml and check if the content is right

## Checklist

### DIFF
--- a/src/cmd/support_archive/resources_test.go
+++ b/src/cmd/support_archive/resources_test.go
@@ -20,6 +20,7 @@ import (
 )
 
 const testOperatorNamespace = "dynatrace"
+const manifestExtension = ".yaml"
 
 func TestManifestCollector_Success(t *testing.T) {
 	logBuffer := bytes.Buffer{}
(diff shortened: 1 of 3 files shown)


src/cmd/support_archive/resources_test.go
```diff
@@ -20,6 +20,7 @@ import (
 )
 
 const testOperatorNamespace = "dynatrace"
+const manifestExtension = ".yaml"
```
0sewa0, line 23: could reuse the `const ManifestsFileExtension = ".yaml"` instead

0sewa0 (changes requested):

src/cmd/support_archive/resources_test.go
luhi-DT: I knew that this will come up, my opinion here is NO. Why? changing the const `ManifestsFileExtension` unintentionally will lead to passing unit tests, even though if its not intented

PR #1592, opened 2023-03-03
Simplify logging and get rid of log singleton

StefanHauth:
# Description
This PR is a preparation for running the troubleshoot command within the support archive command.
First, it gets rid of the hidden log singleton in troubleshoot. This is needed so that the support archive command can provide an own logger to grab all output into a file. With this, logging had to be reworked a bit to keep separate subtest loggers with dedicated names.

Furthermore, formatting log lines seems to be implemented overly complicated, this has been simplified along the way.

## How can this be tested?
Build the operator, deploy it and run the troubleshoot command. Everything should work as before.

## Checklist

### DIFF
--- a/src/cmd/troubleshoot/image_test.go
+++ b/src/cmd/troubleshoot/image_test.go
@@ -1,7 +1,6 @@
 package troubleshoot
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"net/http"
@@ -11,6 +10,7 @@ import (
 
 	"github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/dtpullsecret"
+	"github.com/go-logr/logr"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
@@ -65,17 +65,6 @@ func createSecret(auths Auths) (*corev1.Secret, error) {
 		build(), nil
 }
 
-func runWithTestLogger(testName string, function func()) string {
-	logBuffer := bytes.Buffer{}
-	logger := newTroubleshootLoggerToWriter(testName, &logBuffer)
-
-	oldLog := log
-	log = logger
-	function()
-	log = oldLog
-	return logBuffer.String()
-}
-
 func TestOneAgentImagePullable(t *testing.T) {
 	dockerServer, secret, _, err := setupDockerMocker(
 		[]string{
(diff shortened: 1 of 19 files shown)


src/cmd/troubleshoot/image_test.go
```diff
@@ -1,16 +1,17 @@
 package troubleshoot
```
0sewa0, line 1: this amount of tests feel a bit overkill 😅

0sewa0 (commented):
To be honest I reeeally dislike the "style" of this package, I could complain sooo much.

BUT,  the problem is the `troubleshootContext`, which is not new.

The new addition for logging is fine, and its a minimal approach to make it work, without reworking the whole thing.

TLDR: It is fine, but we should rework the whole package later. (pls fix linter 😄 )

StefanHauth:
> To be honest I reeeally dislike the "style" of this package, I could complain sooo much.
> 
> BUT, the problem is the `troubleshootContext`, which is not new.
> 
> The new addition for logging is fine, and its a minimal approach to make it work, without reworking the whole thing.
> 
> TLDR: It is fine, but we should rework the whole package later. (pls fix linter/test 😄 )

Full agree, I actually started this on another branch already before switching to the low-hanging-fruit approach. I'm planning on initiating the rework.

src/cmd/troubleshoot/testutils.go
```diff
@@ -0,0 +1,23 @@
+package troubleshoot
```
0sewa0, line 1: does GO know that this is only for testing? (so that its not built into the actual image)

I know that the `something_test.go` syntax does that, dunno about this

src/cmd/troubleshoot/image_test.go
StefanHauth: I didn't add any, just compressed them a wee bit.

src/cmd/troubleshoot/testutils.go
0sewa0: checked, if these functions are only used in tests, then rename the file to `utils_test.go`

StefanHauth: As long as those methods are not used in production code, they won't be part of it. The linker should take care of this, just checked it (`strings -6 operator-binary  | grep getNullLogger`).
`utils_test.go` IMHO is misleading as it suggests to contain "test for utils" not "utils for test".

0sewa0: could be in `logger_test.go` as it has to do with the logger (and then we dont use the forbidden `utils` word 😛 )

0sewa0 (approved):

PR #1593, opened 2023-03-03
Auto-detects platform for helm install

0sewa0:
# Description

Using the available API versions on the target cluster we can determine the platform so the user don't have to set it manually.

The "mapping" is the following:
- `security.openshift.io/v1` ==> `openshift`
- `auto.gke.io/v1` ==> `gke-autopilot` 
- ~`app.k8s.io/v1beta1` ==> `google-marketplace`~ (not necessary, and actually can be dangerous)
- fallback ==> `kubernetes`

The `.Values.platform` is still available to override if something goes wrong.
Furthermore, this "auto-detect" feature does not work for `helm template`, which we use for development. 
(this is perfectly fine, and makes all the sense)

## How can this be tested?
Run `helm install dynatrace-operator config/helm/chart/default ...` on all the platforms to see if everything is applied correctly. (make sure to set an image that actually exists)

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/_platform.tpl
+++ b/config/helm/chart/default/templates/_platform.tpl
@@ -1,0 +1,37 @@
+// Copyright 2020 Dynatrace LLC
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+//     http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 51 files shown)


config/helm/chart/default/templates/_platform.tpl
```diff
@@ -0,0 +1,37 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+{{/*
+Auto-detect the platform (if not set), according to the available APIVersions
+*/}}
+{{- define "dynatrace-operator.platform" -}}
+    {{- if .Values.platform}}
+        {{- printf .Values.platform -}}
```
chrismuellner, line 20: should keep same logic as previously, so `platform` has to be one of the allowed values

PR #1598, opened 2023-03-06
Concats LD_PRELOAD values if already exits on container

0sewa0:
# Description

Incase the `LD_PRELOAD` envvar is already set on a container that the webhook is injecting into:
- We add the oneagents preload path to the end
- We respect the delimiter used (`<space>` and `:` are supported), we default to `:` 
- If the oneagents preload path already in the envvar, then we do nothing

## How can this be tested?
The unittests are pretty extensive, since this is easily unittestable.

Otherwise inject into a pod that has a container with `LD_PRELOAD` already set.
(I just patched one of my sample apps 😅 )

## Checklist

### DIFF
--- a/src/webhook/mutation/pod_mutator/oneagent_mutation/env_test.go
+++ b/src/webhook/mutation/pod_mutator/oneagent_mutation/env_test.go
@@ -5,6 +5,7 @@ import (
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/deploymentmetadata"
+	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


src/webhook/mutation/pod_mutator/oneagent_mutation/env_test.go
```diff
@@ -11,16 +11,80 @@ import (
 )
 
 func TestAddPreloadEnv(t *testing.T) {
```
StefanHauth, line 13: Candidate for a table test, in the current state there's a lot of duplicate code.

0sewa0: IMO, table test would be to much hassle, and I am still not a fan of them. 😅 

But yeah, will look into minimising the code duplications 😉

luhi-DT (approved):

src/webhook/mutation/pod_mutator/oneagent_mutation/env_test.go
0sewa0: I tried the table tests and they actually add 3 lines more code 😄 (defining the struct for the testCases, is 4 lines so we would technically save 1 line)

and they feel way more brittle, so I will not do it :D
here is how it would have looked like:
```
type preloadTestCase struct {
	name         string
	existingPath string
	expectedPath string
}

func TestAddPreloadEnv(t *testing.T) {
	installPath := "path/test"
	testCases := []preloadTestCase{
		{
			name: "Add preload env",
			existingPath: "",
			expectedPath: installPath,
		},
		{
			name: "Concat preload env, default delimiter",
			existingPath: "path/user",
			expectedPath: "path/user" + ":" + installPath,
		},
		{
			name: "Concat preload env, respect delimiter",
			existingPath: "path1/user path2/user",
			expectedPath: "path1/user path2/user" + " " + installPath,
		},
		{
			name: "Ignore preload env, if value already present",
			existingPath: "path1/user path2/user" + " " + installPath,
			expectedPath: "path1/user path2/user" + " " + installPath,
		},
	}

	for _, test := range testCases {
		t.Run(test.name, func(t *testing.T) {
			container := createContainerWithPreloadEnv(test.existingPath)

			addPreloadEnv(container, installPath)

			verifyContainerWithPreloadEnv(t, container, test.expectedPath)
		})
	}
}
```

PR #1599, opened 2023-03-07
Run troubleshoot command and collect output into support archive.

StefanHauth:
# Description

The troubleshoot command is now automatically executed when a support archive is created. The output is added to `troubleshoot.txt` in the root of the support archive.

## How can this be tested?
Build and deploy the operator, run the `support-archive` subcommand and check the result for the newly added `troubleshoot.txt` file. It should contain a typical troubleshoot output. As a cross check you could run the `troubleshoot` command and compare results.

## Checklist

### DIFF
--- a/src/cmd/support_archive/troubleshoot.go
+++ b/src/cmd/support_archive/troubleshoot.go
@@ -1,0 +1,52 @@
+package support_archive
+
+import (
+	"bytes"
+	"context"
+
+	"github.com/Dynatrace/dynatrace-operator/src/cmd/troubleshoot"
+	"github.com/go-logr/logr"
+	"k8s.io/client-go/rest"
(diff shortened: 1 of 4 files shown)


src/cmd/support_archive/troubleshoot.go
```diff
@@ -0,0 +1,52 @@
...
+		kubeConfig: kubeConfig,
+		namespace:  namespace,
+	}
+}
+
+func (t troubleshootCollector) Name() string {
+	return troubleshootCollectorName
+}
+
+func (t troubleshootCollector) Do() error {
```
luhi-DT, line 42: If this function never returns an error, why do we need it then in the signature?

luhi-DT (changes requested):

luhi-DT (approved):

PR #1609, opened 2023-03-10
Support all custom pull secret auth methods

gkrenn:
# Description

Currently only custom pull secrets with a username and a password are supported. 
This PR changes logic to use the docker client library to handle pull secrets in order to support all `dockerconfigjson` formats.

## How can this be tested?

- Create a custom pull secret for a private registry that does not use username and password (e.g. gcr)
- Set pull secret and image in the dynakube 
- The operator should be able to pull the image (visible in the logs)

## Checklist


src/controllers/csi/provisioner/agent.go
```diff
@@ -98,17 +96,10 @@ func getUrlProperties(targetVersion, previousVersion string, pathResolver metada
...
+func newImageInstaller(ctx context.Context, fs afero.Fs, apiReader client.Reader, pathResolver metadata.PathResolver, db metadata.Access, dynakube *dynatracev1beta1.DynaKube) (installer.Installer, error) { //nolint:revive // argument-limit doesn't apply to constructors
 	dockerConfig := dockerconfig.NewDockerConfig(apiReader, *dynakube)
 	if dynakube.Spec.CustomPullSecret != "" {
-		err := dockerConfig.SetupAuths(ctx)
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	if dynakube.Spec.TrustedCAs != "" {
```
luhi-DT, line 110: You are now storing the trustedCAs only if there is a customPullSecret, that can lead to problems if a customer has trustedCAs but no customPullSecret

src/dockerconfig/docker_config.go
```diff
@@ -12,82 +12,120 @@ import (
...
-		log.Info("failed to parse pull secret content", "dynakube", config.Dynakube.Name)
+
+	if err := saveFile(registryCredentials, fs, config.RegistryAuthPath); err != nil {
+		log.Info("failed to store registry credentials", "dynakube", config.Dynakube.Name)
 		return err
 	}
-	config.Auths = dockerAuths
+
+	if config.Dynakube.Spec.TrustedCAs != "" {
+		config.TrustedCertsPath = path.Join(TmpPath, CADir, config.Dynakube.Name)
```
luhi-DT, line 55: a new function for the whole trustedCA part would make it way cleaner

luhi-DT (changes requested):

src/controllers/csi/provisioner/agent.go
gkrenn: You are right, this if should not exist here.

Added a unit test to check for this in the future.

src/dockerconfig/docker_config.go
gkrenn: Created 2 new functions for trustedCAs and auths to make it clear what the code does

0sewa0 (approved):

PR #1610, opened 2023-03-10
Adds feature flag to configure failure policy per dynakube

waodim:
# Description

Currently the failure policy environment variable is set via pod annotations. With this the logic should be changed to an approach where the environment variable can be set via a feature flag in the dynakube.

## How can this be tested?
Deploy the operator and inject into sample apps and check their respective environment variables.

## Checklist

### DIFF
--- a/src/webhook/mutation/pod_mutator/init_container.go
+++ b/src/webhook/mutation/pod_mutator/init_container.go
@@ -18,7 +18,7 @@ func createInstallInitContainerBase(webhookImage, clusterID string, pod *corev1.
 		ImagePullPolicy: corev1.PullIfNotPresent,
 		Args:            []string{"init"},
 		Env: []corev1.EnvVar{
-			{Name: config.InjectionFailurePolicyEnv, Value: kubeobjects.GetField(pod.Annotations, dtwebhook.AnnotationFailurePolicy, "silent")},
+			{Name: config.InjectionFailurePolicyEnv, Value: dynakube.FeatureInjectionFailurePolicy()},
 			{Name: config.K8sPodNameEnv, ValueFrom: kubeobjects.NewEnvVarSourceForField("metadata.name")},
 			{Name: config.K8sPodUIDEnv, ValueFrom: kubeobjects.NewEnvVarSourceForField("metadata.uid")},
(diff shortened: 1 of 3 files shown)


src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -18,7 +18,7 @@ func createInstallInitContainerBase(webhookImage, clusterID string, pod *corev1.
 		ImagePullPolicy: corev1.PullIfNotPresent,
 		Args:            []string{"init"},
 		Env: []corev1.EnvVar{
-			{Name: config.InjectionFailurePolicyEnv, Value: kubeobjects.GetField(pod.Annotations, dtwebhook.AnnotationFailurePolicy, "silent")},
```
luhi-DT, line 21: Changing that, we would ignore the pod.Annotations which are already present in customer environments, thats not what we should do, imho we should have an order, 1. if pod has annotation, take it, 2. if dynakube has annotation take that, 3. default (silent)

luhi-DT (dismissed):

src/webhook/mutation/pod_mutator/init_container.go
luhi-DT: Please check that comment :)

luhi-DT (changes requested):

src/webhook/mutation/pod_mutator/init_container.go
luhi-DT: `kubeobjects.GetField(pod.Annotations, dtwebhook.AnnotationFailurePolicy, FeatureInjectionFailurePolicy())}`

waodim: have a look at my way to implement it, as i think the default value of `kubeobjects.GetField` should be "silent" and we should only check whether `FeatureInjectionFailurePolicy() `is set explicitly to "fail". if not, we just take "silent".

luhi-DT:
Do we want to have unit tests for the cases, where a pod already has the annotation set, to check its not overwritten by the new logic?

luhi-DT (approved):

PR #1614, opened 2023-03-13
Always pull the connection info from remote

realtimetodie:
## Description

Fixes an issue where the endpoint connection info is never updated during the reconciliation loop, when the OneAgent code modules image option is not used (neither `spec.oneAgent.cloudNativeFullStack.codeModulesImage` nor `spec.oneAgent.applicationMonitoring.codeModulesImage`).

Ticket 121913

## How can this be tested?

1. deploy a dynakube that uses "spec.oneAgent.applicationMonitoring.useCSIDriver: true"
2. restart a workload in a targeted namespace so it is auto-injected
3. change the OneAgent connection endpoint confiiguration (e.g. by deploying an additional AG for the same network zone)
4. restart the same workload from step 2 once more
5. wait for the dynakube instance to be reconciled by the provisioner container (e.g. by checking the provisioner log for a message containing "reconciling DynaKube", usually happens within 3-5 minutes)
6. verify the property "serverAddress" in file /opt/dynatrace/oneagent-paas/agent/conf/ruxitagentproc.conf in the app container's file system contains the updated OneAgent connection endpoints reflecting the change made in step 3

## Checklist

* [x]  PR is labeled accordingly
* [x]  I have read and understood the [contribution guidelines](https://github.com/Dynatrace/dynatrace-operator/blob/main/CONTRIBUTING.md)


realtimetodie:
@anton-freyberg @baichinger

realtimetodie:
@0sewa0 @anton-freyberg @baichinger

anton-freyberg:
@baichinger could we take a look at this?

baichinger:
Thanks @realtimetodie for the contribution! This indeed solves the problem. We are going to merge this after the next release (0.11) is done. This is due to scaling concerns and a refactoring in that regards:

Scaling: Each node issues the API request against the server. A large infrastructure (in terms of nodes) would significantly increase the number of requests and therefore the load.

Refactoring: Connection info is currently stored in a ConfigMap (https://github.com/Dynatrace/dynatrace-operator/blob/v0.10.3/src/api/v1beta1/properties.go#L207-L209). The CSI provisioner would need to read the ConfigMap to omit the API request, but would introduce another request against the K8s API server. To overcome that, the connection info (endpoints) will also be stored in the DynaKube status field. This change is planned for 0.12.

realtimetodie:
@baichinger Thank you for giving us a preview of the next two releases. 

However, I want to emphasize that the provisionier controller requeues the reconcile key every five minutes. So personally, I don't see the aforementioned overhead

https://github.com/Dynatrace/dynatrace-operator/blob/cf50f3d1b0f60cff45f7cf40645731b571fb0102/src/controllers/csi/provisioner/controller.go#L44

Optimizations are fine, but it would be amazing if you could release a patch version v0.10.4 with the bugfix enabled. It's unclear to us why we're being held up for two entire minor release cycles to fix a major bug.

chrismuellner:
Thanks for your input, @realtimetodie !

> However, I want to emphasize that the provisionier controller requeues the reconcile key every five minutes. So personally, I don't see the aforementioned overhead

This reconciliation is done not _once_ per 5 minutes, but _once per node per Dynakube_ per 5 minutes. This would cause additional load on larger environments, e.g. 30k+ nodes, by changing the default behavior in the CSI driver.

With the refactor mentioned by @baichinger we will improve this by moving the request to the Operator and will cache it in the Dynakube. not once per node, but once per Dynakube, which will prevent overloading the Tenant with too many requests. This refactor is already planned for v0.12, but unfortunately won't make it into v0.11.

We understand that this solves your problem, however it introduces a potential, bigger problem which would affect all customers with bigger clusters.

d41k4n:
@chrismuellner 
> We understand that this solves your problem, however it introduces a potential, bigger problem which would affect all customers with bigger clusters.

I would like to kindly point out that the described "potential, bigger problem affecting all customers with bigger clusters" already applies today with *any* operator version whenever the Dynakube uses the OneAgent options `CSIdriver` and `codeModulesImage`.

Therefore, I can only conclude that either these options have so far never been used by customers with big clusters or the "tenant overloading problem" isn't really a problem as otherwise I doubt it would be acceptable for these customers having to wait several months for the issue getting resolved with 0.12...

d41k4n:
@realtimetodie

> ## How can this be tested?
> ...
> 4. restart the same workload from step 2 once more
> ..

I believe step 4 isn't really necessary to test this. If anything it might be required *after* the Dynakube has been reconciled so the workload is re-injected using the updated initial endpoint config.

luhi-DT:
Hi @d41k4n,

the codeModulesImage usage also reduces load of the tenant registry, as the image is downloaded and requested from the registry it is stored in. This reduced load compared to the default case makes it possible to call the tenant API to request the connectionInfo instead. 

Another thing, did you try to use a codeModulesImage for your problem? In that case we update the connectionInfo as you described already. Using it would solve your use case I guess right?

d41k4n:
Hi @luhi-DT 

> the codeModulesImage usage also reduces load on the tenant registry, as the image is downloaded and requested from the registry it is stored in. This reduced load compared to the default case makes it possible to call the tenant API to request the connectionInfo instead.

Understood - thanks for clarifying.

> Another thing, did you try to use a codeModulesImage for your problem? In that case we update the connectionInfo as you described already. Using it would solve your use case I guess right?

Yes, I guess it would, but so far we avoided it due to the additional complexity of having to manage the images. But as a temporary workaround (until 0.12 is released) it might be something to look into.

However, this approach currently fails because our private registry's certificate isn't trusted by the CSI driver:

```
{"level":"info","ts":"2023-03-24T17:37:42.850Z","logger":"oneagent-image","msg":"installing agent from image"}
{"level":"info","ts":"2023-03-24T17:37:42.850Z","logger":"oneagent-image","msg":"parsed image reference","imageRef":"image-registry.openshift-image-registry.svc:5000/dynatrace/oneagent-codemodules-stg:latest"}
{"level":"info","ts":"2023-03-24T17:37:42.850Z","logger":"oneagent-image","msg":"got source reference","image":"image-registry.openshift-image-registry.svc:5000/dynatrace/oneagent-codemodules-stg:latest"}
{"level":"info","ts":"2023-03-24T17:37:42.862Z","logger":"oneagent-image","msg":"failed to get image digest","image":"image-registry.openshift-image-registry.svc:5000/dynatrace/oneagent-codemodules-stg:latest"}
{"level":"info","ts":"2023-03-24T17:37:42.862Z","logger":"oneagent-image","msg":"failed to install agent from image","err":"pinging container registry image-registry.openshift-image-registry.svc:5000: Get \"https://image-registry.openshift-image-registry.svc:5000/v2/\": x509: certificate signed by unknown authority","errVerbose":"pinging container registry image-registry.openshift-image-registry.svc:5000: Get \"https://image-registry.openshift-image-registry.svc:5000/v2/\": x509: certificate signed by unknown authority\ngithub.com/Dynatrace/dynatrace-operator/src/installer/image.(*ImageInstaller).installAgentFromImage\n\t/app/src/installer/image/installer.go:96\ngithub.com/Dynatrace/dynatrace-operator/src/installer/image.(*ImageInstaller).InstallAgent\n\t/app/src/installer/image/installer.go:57\ngithub.com/Dynatrace/dynatrace-operator/src/controllers/csi/provisioner.(*agentUpdater).installAgent\n\t/app/src/controllers/csi/provisioner/agent.go:157\ngithub.com/Dynatrace/dynatrace-operator/src/controllers/csi...
{"level":"info","ts":"2023-03-24T17:37:42.862Z","logger":"csi-provisioner","msg":"error when updating agent","error":"pinging container registry image-registry.openshift-image-registry.svc:5000: Get \"https://image-registry.openshift-image-registry.svc:5000/v2/\": x509: certificate signed by unknown authority"}
```

Is there a way we can work around this?

0sewa0:
@d41k4n 
> Is there a way we can work around this?
did you set the `trustedCAs` in the dynakube: https://github.com/Dynatrace/dynatrace-operator/blob/main/assets/samples/cloudNativeFullStack.yaml#L38

d41k4n:
@0sewa0

> did you set the `trustedCAs` in the dynakube? samples: https://github.com/Dynatrace/dynatrace-operator/blob/main/assets/samples/cloudNativeFullStack.yaml#L38

No, because based on the [comment associated with that property in the CR sample](https://github.com/Dynatrace/dynatrace-operator/blob/v0.10.2/assets/samples/applicationMonitoring.yaml#L35) supposedly this "only affects certificates used to communicate with the Dynatrace API".... Confirm this also affects communication with private registries?

PR #1615, opened 2023-03-14
Feature flag for configuring minimum interval between requests to Dynatrace API

StefanHauth:
# Description

Provide a feature flag to define the minimum interval (in minutes) between requests to the Dynatrace API. Defaults to 15 minutes.

## How can this be tested?
Deploy operator, chose a value for the feature flag and deploy a resp. configured dynakube. Observe operator logs, you should see log messages like this stating that requests to Dynatrace API have been skipped:
```
{"level":"info","ts":"2023-03-15T08:19:40.046Z","logger":"dynatrace-client","msg":"skipping token verification, last request was made less than 20 minutes ago, 17 minutes remaining until next request"}
{"level":"info","ts":"2023-03-15T08:19:40.188Z","logger":"dynakube-connectioninfo","msg":"skipping activegate connection info update, last request was made less than 20 minutes ago, 17 minutes remaining until next request"}
{"level":"info","ts":"2023-03-15T08:19:40.188Z","logger":"dynakube-connectioninfo","msg":"skipping oneagent connection info update, last request was made less than 20 minutes ago, 17 minutes remaining until next request"}
```
Your configured limit should be reflected in the log lines, the remaining time should decrease.

## Checklist


src/api/v1beta1/properties.go
```diff
@@ -487,3 +488,27 @@ func (dk *DynaKube) GetOneAgentEnvironment() []corev1.EnvVar {
...
+	return timeutil.TimeoutReached(&dk.Status.DynatraceApi.LastActiveGateConnectionInfoRequest, timeProvider.Now(), dk.FeatureApiRequestThreshold())
+}
+
+func (dk *DynaKube) ShallVerifyTokenScope() bool {
+	return dk.shallVerifyTokenScope(timeutil.NewProvider())
+}
+
+func (dk *DynaKube) shallVerifyTokenScope(timeProvider *timeutil.Provider) bool {
+	return timeutil.TimeoutReached(&dk.Status.DynatraceApi.LastTokenScopeRequest, timeProvider.Now(), dk.FeatureApiRequestThreshold())
+}
```
0sewa0, line 514: the whole point of the `timeProvider` is that it makes time mockable, but you only allow that in for this package, 
so any package that relies on any of the functions can't mock time, which is not the intended outcome. Its really simple to create a `timeProvider` in the reconciler, you could just create it in the reconciler's constructor for now, and we can improve that later if we want to.

furthermore you could have used the `timeProvider.IsOutdated` instead of the `TimeoutReached` as it does the same, but cleaner. (if you are already using a `timeProvider`)

please dont use `Shall...` as a prefix for boolean functions, its really unorthodox and just weird to read 😅 


example:
```
func (dk *DynaKube) IsOutdatedTokenScope(timeProvider *timeutil.Provider) bool {
	return timeProvider.IsOutdated(&dk.Status.DynatraceApi.LastTokenScopeRequest, dk.FeatureApiRequestThreshold())
}
```

src/api/v1beta1/properties_test.go
```diff
@@ -417,3 +420,85 @@ func TestGetOneAgentEnvironment(t *testing.T) {
...
+				LastOneAgentConnectionInfoRequest:   metav1.Time{},
+				LastActiveGateConnectionInfoRequest: metav1.Time{},
+			},
+		},
+	}
+
+	timeProvider := timeutil.NewProvider()
+	tests := map[string]struct {
+		lastRequestTimeDeltaMinutes int
+		shallUpdate                 bool
```
0sewa0, line 438:
```suggestion
		update                 bool
```

0sewa0 (changes requested):

src/time/time.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// The purpose if the Provider is to have a non-moving resp. fakable now for testing.
+type Provider struct {
+	now *metav1.Time
+}
+
+// Warning: This method should be never called twice on the same instance in production code where you need an accurate "now" timestamp,
+// because it returns the same timestamp on each call, even if the stored "now" is hours in the past
+func (timeProvider *Provider) Now() *metav1.Time {
```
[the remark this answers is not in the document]
0sewa0: me like, does the same, but nicer

maybe need to update tests that might have utilised this "feature"

should maybe be done in a followup? 🤔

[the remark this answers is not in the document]
StefanHauth: Yes, followup.

src/controllers/dynakube/version/version.go
```diff
@@ -8,7 +8,7 @@ import (
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	"github.com/Dynatrace/dynatrace-operator/src/dockerconfig"
-	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
+	timeutil "github.com/Dynatrace/dynatrace-operator/src/time"
```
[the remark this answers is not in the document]
StefanHauth: If we have other time utils (there, I said it :-)) in the future, we need to make another package and have conversation again the same  again. It could be worse, e.g. `timehelper`

src/api/v1beta1/properties.go
StefanHauth: True, kind of got lost in details here, I guess.

src/controllers/dynakube/version/version.go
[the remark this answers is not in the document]
0sewa0: I don't see any reason to have other timeutils that couldn't be part of the `timeprovider` as a method

src/api/v1beta1/properties_test.go
StefanHauth: Here the `shall` is justified, because it reads "expect update" not "do that update"

src/api/v1beta1/properties.go
```diff
@@ -487,3 +488,15 @@ func (dk *DynaKube) GetOneAgentEnvironment() []corev1.EnvVar {
 	}
 	return []corev1.EnvVar{}
 }
+
+func (dk *DynaKube) ShallUpdateOneAgentConnectionInfo(timeProvider *timeprovider.Provider) bool {
```
0sewa0, line 492: pls, don't use "shall"", as its a not used anywhere in the codebase, and it shouldn't be anyway.

"Shall" is mainly used to express future events, or intentions (and not true/false questions). Furthermore (in my experience )its not really used in normal speech or text ("will" is waaay more common). Most common places I encountered "shall" were in, songs, videogames and fantasy novels. None of which fits our code 😄 .

There are way more generic, better and simpler options:
- `NeedsOneAgentConnectionInfoUpdate`
- `IsOutdatedOneAgentConnectionInfo`

src/api/v1beta1/properties_test.go
```diff
@@ -417,3 +420,85 @@ func TestGetOneAgentEnvironment(t *testing.T) {
...
+				LastOneAgentConnectionInfoRequest:   metav1.Time{},
+				LastActiveGateConnectionInfoRequest: metav1.Time{},
+			},
+		},
+	}
+
+	timeProvider := timeprovider.New()
+	tests := map[string]struct {
+		lastRequestTimeDeltaMinutes int
+		shallUpdate                 bool
```
0sewa0, line 438:
```suggestion
		shouldUpdate                 bool
```

0sewa0 (changes requested):

src/api/v1beta1/properties_test.go
StefanHauth: No. "should" means, "you can if you like, just saying", shall means "do this or die".

StefanHauth: And in the end this is a test specification where it is appropriate to use specification language.

src/api/v1beta1/properties.go
StefanHauth: I don't like this "is outdated" phrase, that sounds yukky to me. I probably can (and hence will) life with "Needs..."

src/api/v1beta1/properties_test.go
StefanHauth: But I can go with `expectUpdate`, that brings across the point as good.

0sewa0: `expectUpdate` or `updateExpected` are great 👍

src/api/v1beta1/properties.go
gkrenn: In my opinion I agree with @0sewa0. - Never saw "Shall" in any code and gives me a hard time reading it. Would rather vote for `IsOutdatedOneAgentConnectionInfo` to aline with the other functions in our code. - "Is" fits perfect here as it describes the state if the struct it refers to.

StefanHauth: On second thought all of suggestions so far are wrong, because what this property says, that it is now allowed to do an update wrt. the timeout restrictions. It doesn't say if the update has to be done, if it needs to be done or of the connection info is already outdated. The dynakube instance at hand just doesn't know that.

chrismuellner: let's go with one of marcell's suggestions. we should avoid "shall" if the meaning is ambiguous and there are other, well-known words with the same meaning already used in the code base.

0sewa0 (approved):

PR #1619, opened 2023-03-16
Adapt dockerfile to install dependencies correctly

luhi-DT:
# Description

### Why?
Currently we are copying some packages and libraries that our image needs, which are not included in ubi9-micro (our base image). The problem with that way of including additional packages is, that scanners are not detecting those dependencies properly as no index within package managers databases are created and therefore also SBOM creation is only partly possible.

### What?
The new implementation first installs properly a new package into the filesystem of ubi9-micro and copies the resulting filesystem then over to the last layer. This way all packages are correctly installed and scanners are able to detect all dependencies correctly

## How can this be tested?

Deploy the operator and CSI Pods, test if everything works correctly (mainly mounting)

## Checklist


gkrenn (approved):
Looks fine, one thing you could change, even thought it's not directly related to this change is to ensure that the "root" folder has root as owner and not "dynatrace-operator"

PR #1630, opened 2023-03-17
Removes unnecessary http port for activegates

0sewa0:
# Description

Removes http ports from the ActiveGate:
- `9998` on the ActiveGate container
- `80` on the ActiveGate service

These are not necessary, so we shouldn't keep them.

## How can this be tested?
Deployed ActiveGates work as expected

## Checklist

### DIFF
--- a/src/controllers/dynakube/activegate/internal/statefulset/statefulset.go
+++ b/src/controllers/dynakube/activegate/internal/statefulset/statefulset.go
@@ -135,7 +135,7 @@ func (statefulSetBuilder Builder) buildBaseContainer() []corev1.Container {
 			ProbeHandler: corev1.ProbeHandler{
 				HTTPGet: &corev1.HTTPGetAction{
 					Path:   "/rest/health",
-					Port:   intstr.IntOrString{IntVal: 9999},
+					Port:   intstr.IntOrString{IntVal: consts.HttpsContainerPort},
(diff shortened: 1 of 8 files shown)


src/controllers/dynakube/activegate/internal/statefulset/statefulset.go
```diff
@@ -135,7 +135,7 @@ func (statefulSetBuilder Builder) buildBaseContainer() []corev1.Container {
 			ProbeHandler: corev1.ProbeHandler{
 				HTTPGet: &corev1.HTTPGetAction{
 					Path:   "/rest/health",
-					Port:   intstr.IntOrString{IntVal: 9999},
```
luhi-DT, line 138: Why not using the const here? `HttpsContainerPort`

luhi-DT (changes requested):

src/controllers/dynakube/activegate/internal/statefulset/statefulset.go
0sewa0: that is the change here 😬 
(🔴 == old ; 🟢 == new)

luhi-DT: brain lag :D

luhi-DT (approved):

PR #1636, opened 2023-03-21
Uses the status for image info

0sewa0:
# Description

Follow up of the Status rework. Changes the properties and the logic in general, that the information about the image/version is always taken from the Status. (except when we generate the said status and the troubleshoot script)

Unexpected rework:
The Status fields `ImageRepository`, `ImageTag`, `ImageHash` have been moved into a single `ImageID` field, as having them separately just requires us to constantly combine them dynamically, as they are not used 1 by 1.
(also renames imageHash, to imageDigest as that is the correct name) 

## How can this be tested?
Deploy oneagents, activegates and codemodules in every possible way. 
- custom image
- custom version
- default
- public-registry( if available on your tenant)

## Checklist

### DIFF
--- a/src/controllers/dynakube/version/updater.go
+++ b/src/controllers/dynakube/version/updater.go
@@ -37,6 +37,7 @@ func (reconciler *Reconciler) run(ctx context.Context, updater versionStatusUpda
 
 	customImage := updater.CustomImage()
 	if customImage != "" {
+		log.Info("updating version status according to custom image", "updater", updater.Name())
 		err = updateVersionStatus(ctx, updater.Target(), customImage, reconciler.digestFunc, dockerCfg)
 		return err
 	}
@@ -68,6 +69,7 @@ func (reconciler *Reconciler) run(ctx context.Context, updater versionStatusUpda
 		return nil
 	}
 
+	log.Info("updating version status according to the tenant registry", "updater", updater.Name())
 	err = updater.UseDefaults(ctx, dockerCfg)
 	return err
 }
@@ -109,12 +111,12 @@ func updateVersionStatus(
 			target.ImageID = taggedRef.String()
 			log.Error(err, "failed to get image digest, falling back to tag")
 		} else {
-			cannonRef, err := reference.WithDigest(taggedRef, digest)
+			canonRef, err := reference.WithDigest(taggedRef, digest)
 			if err != nil {
 				target.ImageID = taggedRef.String()
 				log.Error(err, "failed to create canonical image reference, falling back to tag")
 			} else {
-				target.ImageID = cannonRef.String()
+				target.ImageID = canonRef.String()
 			}
 		}
 	} else {


src/controllers/csi/provisioner/controller.go
```diff
@@ -94,6 +95,11 @@ func (provisioner *OneAgentProvisioner) Reconcile(ctx context.Context, request r
 		return reconcile.Result{RequeueAfter: longRequeueDuration}, provisioner.db.DeleteDynakube(ctx, request.Name)
 	}
 
+	if dk.CodeModulesImage() != "" && dk.CodeModulesVersion() != "" {
+		log.Info("dynakube status is not yet ready, requeuing", "dynakube", dk.Name)
```
chrismuellner, line 99: wouldn't either the hash (for code modules image) or the version be an indication that the status is ready? i.e. depending on the "source"

src/controllers/dynakube/version/reconciler.go
```diff
@@ -84,12 +81,14 @@ func (reconciler *Reconciler) createDockerConfigWithCustomCAs(ctx context.Contex
...
 func (reconciler *Reconciler) needsReconcile(updaters []versionStatusUpdater) bool {
+	neededUpdaters := []versionStatusUpdater{}
 	for _, updater := range updaters {
 		if reconciler.needsUpdate(updater) {
-			return true
+			neededUpdaters = append(neededUpdaters, updater)
 		}
 	}
-	return false
+	updaters = neededUpdaters
```
chrismuellner, line 90: can you just return `neededUpdaters` and assign it to `updaters` outside this method? currently this has "hidden" side effects

src/controllers/dynakube/oneagent/oneagent_reconciler.go
```diff
@@ -127,10 +127,7 @@ func (r *Reconciler) reconcileRollout(ctx context.Context, dynakube *dynatracev1
 }
 
 func (r *Reconciler) getOneagentPods(ctx context.Context, dynakube *dynatracev1beta1.DynaKube, feature string) ([]corev1.Pod, []client.ListOption, error) {
-	agentVersion := dynakube.Status.OneAgent.ImageTag
-	if dynakube.CustomOneAgentImage() != "" {
-		agentVersion = kubeobjects.CustomImageLabelValue
-	}
+	agentVersion := dynakube.Status.OneAgent.Version
```
chrismuellner, line 130: property?

src/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -132,10 +132,7 @@ func (dsInfo *builderInfo) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 	dynakube := dsInfo.dynakube
 	podSpec := dsInfo.podSpec()
 
-	versionLabelValue := dynakube.Status.OneAgent.ImageTag
-	if dynakube.CustomOneAgentImage() != "" {
-		versionLabelValue = kubeobjects.CustomImageLabelValue
-	}
+	versionLabelValue := dynakube.Status.OneAgent.Version
```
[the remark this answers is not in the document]
chrismuellner: should we be using `ImageTag` instead of `version` here? version will only be set for code modules

chrismuellner (changes requested):

src/api/v1beta1/properties.go
```diff
@@ -206,13 +206,65 @@ func (dk *DynaKube) PullSecret() string {
...
+		return dk.Spec.OneAgent.CloudNativeFullStack.NodeSelector
+	}
+	return nil
+}
+
+func (dk *DynaKube) ActiveGateImage() string {
+	return dk.Status.ActiveGate.ImageURI()
 }
 
 func (dk *DynaKube) DefaultActiveGateImage() string {
```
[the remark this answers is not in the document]
0sewa0: yes, otherwise how would the `version` package access them ? (we could move that logic there, but that would be not nice, also we have to think about the troubleshoot script)

src/controllers/csi/gc/reconciler.go
```diff
@@ -75,11 +75,6 @@ func (gc *CSIGarbageCollector) Reconcile(ctx context.Context, request reconcile.
 		return defaultReconcileResult, err
 	}
 
-	if !isSafeToGC(ctx, gc.db, dynakubeList) {
```
0sewa0, line 79: unnecessary as the GC no longer runs at the same time as the provisioner

src/controllers/csi/provisioner/agent.go
```diff
@@ -76,7 +76,7 @@ func newAgentImageUpdater( //nolint:revive // argument-limit doesn't apply to co
 		fs:            fs,
 		path:          path,
 		targetDir:     path.AgentConfigDir(tenantUUID),
-		targetVersion: dk.CodeModulesImageTag(),
```
0sewa0: ImageTag my or may not be set, and in the `ImageInstaller`'s case, this is only used for logging, therefor its more helpful to have the image in the log, then potentially an empty string

src/controllers/csi/provisioner/controller.go
```diff
@@ -215,11 +215,12 @@ func (provisioner *OneAgentProvisioner) updateAgentInstallation(ctx context.Cont
...
 		return nil, true, nil
 	} else if updatedVersion != "" {
-		dynakubeMetadata.LatestVersion = updatedVersion
 		imageInstaller, isImageInstaller := agentUpdater.installer.(*image.Installer)
 		if isImageInstaller {
+			dynakubeMetadata.LatestVersion = ""
 			dynakubeMetadata.ImageDigest = imageInstaller.ImageDigest()
 		} else {
+			dynakubeMetadata.LatestVersion = updatedVersion
 			dynakubeMetadata.ImageDigest = ""
```
0sewa0, line 224: in the case of the `ImageInstaller` the `LatestVersion` field in the database is not necessary
in the case of the `UrlInstaller` the `ImageDigest` field in the database is not necessary

0sewa0 (commented):
Added some explanation to some of the questionable changes

src/controllers/csi/gc/reconciler.go
0sewa0: but why now?: it relied on the Version/Tag in the spec/status, so this logic would need update due to the Status changes, which is kinda useless work if its not needed in the first place

src/controllers/dynakube/version/activegate.go
```diff
@@ -11,18 +11,18 @@ import (
...
 	dtClient dtclient.Client
-	hashFunc ImageHashFunc
+	digestFunc ImageDigestFunc
 }
 
 func newActiveGateUpdater(
 	dynakube *dynatracev1beta1.DynaKube,
 	dtClient dtclient.Client,
-	hashFunc ImageHashFunc,
+	hashFunc ImageDigestFunc,
```
waodim, line 20:
```suggestion
	digestFunc ImageDigestFunc,
```

src/controllers/dynakube/version/activegate.go
```diff
@@ -11,18 +11,18 @@ import (
...
 ) *activeGateUpdater {
 	return &activeGateUpdater{
 		dynakube: dynakube,
 		dtClient: dtClient,
-		hashFunc: hashFunc,
+		digestFunc: hashFunc,
```
waodim, line 25:
```suggestion
		digestFunc: digestFunc,
```

src/controllers/dynakube/version/activegate.go
waodim: Same goes for the other updaters as well if you think about commiting this suggestion

src/controllers/dynakube/version/updater.go
```diff
@@ -88,37 +89,41 @@ func updateVersionStatus(
...
-		repo = taggedRef.Name()
-		tag = taggedRef.Tag()
-		hash, err = hashFunc(ctx, imageUri, dockerCfg)
+		digest, err := digestFunc(ctx, imageUri, dockerCfg)
 		if err != nil {
-			return errors.WithMessage(err, "failed to get image hash")
+			target.ImageID = taggedRef.String()
+			log.Error(err, "failed to get image digest, falling back to tag")
+		} else {
+			cannonRef, err := reference.WithDigest(taggedRef, digest)
```
chrismuellner, line 112:
```suggestion
			canonRef, err := reference.WithDigest(taggedRef, digest)
```

chrismuellner (changes requested):

chrismuellner (approved):

PR #1662, opened 2023-03-27
Add E2E test for synthetic capability 

romr-of-dt:
Goal
===
To offer an E2E test which confirms the synthetic location has acted for the browser monitor.

Means
---
The operator is to set up an **Active Gate** focused on the observability and a **synthetic location** apt to accomplish a browser visit.


test/scenarios/synthetic/monolocation/suite.go
```diff
@@ -0,0 +1,273 @@
+//go:build e2e
+
+package monolocation
```
0sewa0, line 3: what do you mean by `monolocation`?

test/scenarios/synthetic/monolocation/suite.go
```diff
@@ -0,0 +1,273 @@
...
+	context context.Context
+	t       *testing.T
+	config  *envconf.Config
+}
+
+type stepContext struct {
+	dynaKube         dynatracev1beta1.DynaKube
+	pod              corev1.Pod
+	containerLogSpec corev1.PodLogOptions
+}
```
0sewa0, line 38: what is the purpose of these?

looks like a hacky way to pass/store state between functions.
Wouldn't a struct with methods be better? like:
```
type SyntheticFeature struct {
....
}

func (feature *SyntheticFeature) build(t *testing.T) features.Feature {
....
}

....
```

0sewa0 (changes requested):

test/scenarios/synthetic/monolocation/suite.go
romr-of-dt: I was searching a neologism for the scenario which evaluates a location with a single VUC. Intended to create another for a location with a pair of VUCs (_stereoloc_). For now, postponed the latter.

test/scenarios/synthetic/monolocation/suite.go
romr-of-dt: The difficulty concerns the context for the subordinate steps. I need to keep the state for a series of `features.Func`. The `features.Feature` operates on the higher level.

Deducted from the given rule that the test scenario is a request which has to transit the state. The `context.Context` offers all the means.
> Use context Values only for request-scoped data that transits processes
> and APIs, not for passing optional parameters to functions.

Considered such signatures verbose. The function receiver `stepContext` looks superfluous when the `context.Context` is at hand.
```
func requireContainerLogToMatch(
    ctx context.Context,
    t *testing.T,
    cfg *envconf.Config,
    regexp *regexp.Regexp,
) string
    
func requireContainerLog(
    ctx context.Context,
    t *testing.T,
    cfg *envconf.Config,
) string

func requireVucBrowserLog(
    ctx context.Context,
    t *testing.T,
    cfg *envconf.Config,
) string
```

romr-of-dt: Following the given rule to dismiss the file-scoped `context.Context`.

> Do not store Contexts inside a struct type; instead, pass a Context
> explicitly to each function that needs it. The Context should be
> the first parameter, typically named ctx...

0sewa0: you are using a package global variable to store your state, that is a really big anti-pattern if you ask me

plus, storing the information in the context is possible yes, I would highly discourage that.
its highly unreadable as its basically an untyped map that you can put anything into.
you could use a struct with methods to pass/maintain state without resulting to global variables and type conversion from the context's values.

in other parts of the e2e tests we tend to use closures to inject state into a step (like you did in `requireStepContext` )
these closures are far more preferable to a function as `requireContainerLogToMatch`
according to the signature of that function it will use no state, as its just need a regex, so you are hiding what it needs (if it was a method of something you could say that it has access to the state of the struct)

and you don't have to store the context in your struct, mostly because the context is handled by the framework

romr-of-dt: Following the suggestion.

Collected hundreds of matches for the `context.WithValue(...)` in the imported golang modules. Could so many authors not bother the _type-safeness_? Presume they merely follow the `context.Context` contract. Which - unquestionably - the `e2e-framework` samples legitimate.

Not sure if the `func (*stepContext) requireOperableVuc(context.Context, *testing.T, *envconf.Config) context.Context` is unambiguous. It confines the `e2e-framework` step context and the custom one...
```
go/pkg/mod/go.mongodb.org/mongo-driver@v1.11.1/mongo/session.go:...:
// NewSessionContext creates a new SessionContext associated with the given Context and Session parameters.
func NewSessionContext(ctx context.Context, sess Session) SessionContext {
	return &sessionContext{
		Context: context.WithValue(ctx, sessionKey{}, sess),
		Session: sess,
	}
}

go/pkg/mod/k8s.io/apiserver@v0.20.0/pkg/warning/context.go:...:
func WithWarningRecorder(ctx context.Context, recorder Recorder) context.Context {
    return context.WithValue(ctx, warningRecorderKey, recorder)
}
func warningRecorderFrom(ctx context.Context) (Recorder, bool) {
    recorder, ok := ctx.Value(warningRecorderKey).(Recorder)
    return recorder, ok

go/pkg/mod/k8s.io/apiserver@v0.20.0/pkg/server/httplog/httplog.go:...:
            panic("multiple WithLogging calls!")
        }
        rl := newLogged(req, w).StacktraceWhen(pred)
        req = req.WithContext(context.WithValue(ctx, respLoggerContextKey, rl))

go/pkg/mod/k8s.io/apiserver@v0.20.0/pkg/server/filters/priority-and-fairness.go:...:
            noteExecutingDelta(1)
            defer noteExecutingDelta(-1)
            served = true
            innerCtx := context.WithValue(ctx, priorityAndFairnessKey, classification)

go/pkg/mod/sigs.k8s.io/e2e-framework@v0.0.7/examples/parallel_features/parallel_features_test.go:...:
func TestPodBringUp(t *testing.T) {
    featureOne := features.New("Feature One").
        Assess("Create Nginx Deployment 1", func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
            deployment := newDeployment(namespace, "deployment-1", 2)
            ...
            ctx = context.WithValue(ctx, "DEPLOYMENT", deployment)
            return ctx
        }).
```

test/scenarios/activegate/installation.go
```diff
@@ -3,18 +3,16 @@
...
-	"bytes"
 	"context"
 	"fmt"
-	"io"
 	"strings"
 	"testing"
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/activegate/consts"
+	"github.com/Dynatrace/dynatrace-operator/test/helpers/components/activegate"
```
0sewa0, line 13: this package does not exist in you branch (or you did not commit them)
I created that package in my example at https://github.com/Dynatrace/dynatrace-operator/tree/syntetic-e2e
so we don't repeat ourselves

romr-of-dt: I excluded it unintentionally. Correcting....

PR #1663, opened 2023-03-27
Avoid panic when generating certificates for conversion webhook

luhi-DT:
# Description
Currently if the CRD has no ClientConfig or the whole conversion section in the CRD was empty, the operator panicked, because of a nil pointer error. This should be solved with properly checking if the values, which are accessed inside `isBundleValid()` are set

## How can this be tested?
Deploy a CRD with an empty ClientConfig section
Start the operator and check if the cert-controller is still panicking

## Checklist


luhi-DT:
Adding a log is not necessary, as a log is written properly afterwards: https://github.com/Dynatrace/dynatrace-operator/blob/c2808df05b60b2f0c0696bac9d7a8c5faf37300a/src/controllers/certificates/webhook_cert_controller.go#L230

PR #1665, opened 2023-03-29
Fixes e2e networking related tests for openshift

0sewa0:
# Description
 Currently some e2e tests fail (catastrophically, so they kill other tests aswell) specifically on openshift for several reasons:
 - the proxy that is used/deployed does not have a SCC, so it get restricted too much + `CAP_NET_RAW` capability is not granted by default on openshift
 - the `NetworkPolicy` used to deny egress to the internet from the dynatrace namespace is way to strict (it works on kubernetes, because by default nothing enforces the policy, not even istio)
 - For istio to work on openshift a certain resource must be created in each namespace using istio. (`NetworkAttachmentDefinition`) 
    - In order to do this in a nice way (when we create a namespace that has the istio injection + we are on openshift) I had to do a small reorg of the `sampleapps` package to avoid a circular dependency.
 - The istio initcontainer is named differently on openshift. 🤷  

## How can this be tested?
run `make PLATFORM="openshift" test/e2e`

## Checklist

### DIFF
--- a/test/helpers/istio/install.go
+++ b/test/helpers/istio/install.go
@@ -5,14 +5,18 @@ package istio
 import (
 	"context"
 	"os"
+	"path"
 	"strings"
 	"testing"
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/istio"
 	"github.com/Dynatrace/dynatrace-operator/src/dtclient"
-	"github.com/Dynatrace/dynatrace-operator/test/helpers/sampleapps"
+	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
+	"github.com/Dynatrace/dynatrace-operator/test/helpers/kubeobjects/manifests"
+	"github.com/Dynatrace/dynatrace-operator/test/helpers/sampleapps/interface"
 	"github.com/Dynatrace/dynatrace-operator/test/helpers/tenant"
+	"github.com/Dynatrace/dynatrace-operator/test/project"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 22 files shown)


test/helpers/proxy/proxy.go
```diff
@@ -27,14 +28,18 @@ var (
...
 
 	ProxySpec = &dynatracev1beta1.DynaKubeProxy{
-		Value: "http://squid.proxy:3128",
+		Value: "http://squid.proxy.svc.cluster.local:3128",
 	}
 )
 
 func SetupProxyWithTeardown(builder *features.FeatureBuilder, testDynakube dynatracev1beta1.DynaKube) {
 	if testDynakube.Spec.Proxy != nil {
+		if kubeobjects.ResolvePlatformFromEnv() == kubeobjects.Openshift {
```
luhi-DT, line 40: Maybe a own function that is just called here would make sense to keep this method clean

test/helpers/istio/install.go
```diff
@@ -108,12 +134,12 @@ func assertIstioInitContainer(t *testing.T, pods corev1.PodList, testDynakube dy
 		istioInitFound := false
 
 		for _, initContainer := range podItem.Spec.InitContainers {
-			if initContainer.Name == istioInitContainerName {
+			if initContainer.Name == istioInitName {
```
luhi-DT, line 137: we could check with an or clause here for both platforms either openshift initContainerName || kubernetes initContainer name, so we can save the platform switch

luhi-DT (changes requested):

test/helpers/istio/install.go
0sewa0: `istioInitName` used somewhere else in the function, so I just moved the logic to decide which init name to look for into `determineIstioInitContainerName`

luhi-DT (approved):

PR #1667, opened 2023-03-30
Auto-detects platform for helm install

0sewa0:
# Description

Using the available API versions on the target cluster we can determine the platform so the user don't have to set it manually.

The "mapping" is the following:

security.openshift.io/v1 ==> openshift
auto.gke.io/v1 ==> gke-autopilot
fallback ==> kubernetes

The .Values.platform is still available to override if something goes wrong.
Furthermore, this "auto-detect" feature does not work for helm template, which we use for development.
(this is perfectly fine, and makes all the sense)

## How can this be tested?
Run helm install dynatrace-operator config/helm/chart/default ... on all the platforms to see if everything is applied correctly. (make sure to set an image that actually exists)

## Checklist


config/helm/chart/default/templates/_platform.tpl
```diff
@@ -0,0 +1,53 @@
...
+{{- if and (or (eq (include "dynatrace-operator.platform" .) "openshift") (.Values.olm)) (eq (include "dynatrace-operator.partial" .) "false") -}}
+    {{ default "true" }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Check if platform is set to a valid one
+*/}}
+{{- define "dynatrace-operator.platformValid" -}}
+{{- if or (eq (include "dynatrace-operator.platform" .) "kubernetes") (eq (include "dynatrace-operator.platform" .) "openshift") (eq (include "dynatrace-operator.platform" .) "google-marketplace") (eq (include "dynatrace-operator.platform" .) "gke-autopilot") -}}
```
chrismuellner, line 43: could you split this up to make it more readable?

chrismuellner (changes requested):

PR #1685, opened 2023-04-07
Reworks OneAgentCanConnect() check

waodim:
# Description

Previously we checked whether oneAgent can connect which is somewhat out of our scope, as we should focus on testing operator related things. That is why we from now simply check whether the oneAgent Pods are running/ready or not. FYI: I moved the function into `daemonset.go` as the name `connection.go` was misleading with the new check.

## How can this be tested?

Run `make test/e2e`

## Checklist


test/helpers/steps/assess/dynakube.go
```diff
@@ -36,7 +36,7 @@ func UpdateDynakube(builder *features.FeatureBuilder, testDynakube dynatracev1be
 func verifyDynakubeStartup(builder *features.FeatureBuilder, testDynakube dynatracev1beta1.DynaKube) {
 	if testDynakube.NeedsOneAgent() {
 		builder.Assess("oneagent started", oneagent.WaitForDaemonset(testDynakube))
-		builder.Assess("osAgent can connect", oneagent.OSAgentCanConnect(testDynakube))
+		builder.Assess("oneagent pods are ready", oneagent.PodsAreReady(testDynakube))
```
0sewa0, line 39: I think the `oneagent.WaitForDaemonset` already does what the `oneagent.PodsAreReady`
so I think `oneagent.PodsAreReady` is no necessary

0sewa0 (changes requested):

test/helpers/steps/assess/dynakube.go
0sewa0: just remove it

0sewa0 (approved):

PR #1689, opened 2023-04-11
Adds downgrade check for the public registry

0sewa0:
# Description

While the immutable oneagents from the public registry will not crash-loop due to downgrade, they will still not work, we should disallow downgrades of the oneagent in all cases where we know the version

## How can this be tested?
try using the public-registry mode (if its available for you)

## Checklist


src/controllers/dynakube/version/updater.go
```diff
@@ -61,6 +62,15 @@ func (reconciler *Reconciler) run(ctx context.Context, updater versionStatusUpda
 			log.Info("could not get public image", "updater", updater.Name())
 			return err
 		}
+		isDowngrade, err := updater.CheckForDowngrade(publicImage.Tag)
+		if err != nil {
+			log.Info("could not determine if is downgrade for public registry", "updater", updater.Name())
```
waodim:
```suggestion
			log.Info("could not determine if there is a downgrade for public registry", "updater", updater.Name())
```

luhi-DT (approved):

PR #1690, opened 2023-04-12
Run uninstall.sh script after classic e2e test

aorcholski:
# Description
The `classic` e2e test leaves some leftovers on the nodes which have to be removed before the `cloudnative` e2e test is started.
A daemonset is used to run the `uninstall.sh` script on each node. The `sleep 300` command keeps the pods running and set appropriate state of the deamonset. The actual command launching the `uninstall.sh` script is called from the e2e test. Standard output and errors reported by the scripts are printed to the test log if the script fails. 

Privileged `SecurityContext` is needed in case of deployment on an OpenShift cluster.

## How can this be tested?
Run the `classic` test and then the `cloudnative` test on a k8s cluster and openshift cluster.
```
make test/e2e/classic
make test/e2e/cloudnative
```

## Checklist


test/helpers/components/oneagent/daemonset.go
```diff
@@ -29,9 +45,36 @@ func Get(ctx context.Context, resource *resources.Resources, dynakube dynatracev
...
 
-func ForEachPod(ctx context.Context, resource *resources.Resources, dynakube dynatracev1beta1.DynaKube, actionFunc daemonset.PodConsumer) error {
-	return daemonset.NewQuery(ctx, resource, client.ObjectKey{
-		Name:      dynakube.OneAgentDaemonsetName(),
-		Namespace: dynakube.Namespace,
-	}).ForEachPod(actionFunc)
+func CreateUninstallDaemonSet(namespace string) features.Func {
+	return func(ctx context.Context, t *testing.T, environmentConfig *envconf.Config) context.Context {
+		oaDaemonSet := manifests.ObjectFromFile[*appsv1.DaemonSet](t, uninstallOneAgentDaemonSetPath)
+		oaDaemonSet.Namespace = namespace
```
0sewa0, line 51: should we also set the `Tolerations` for the daemonset according to the `Tolerations` set in the dynakube?

test/helpers/kubeobjects/daemonset/pods.go
```diff
@@ -0,0 +1,16 @@
+//go:build e2e
```
luhi-DT, line 1: I think we already have that in - test/helpers/kubeobjects/daemonset/get.go

test/helpers/components/oneagent/daemonset.go
luhi-DT: yes good idea, otherwise it could happen that agent runs on nodes where cleanup would not be run afterwards

0sewa0 (approved):

PR #1691, opened 2023-04-12
Cache OneAgent and ActiveGate connection info in Dynakube status

StefanHauth:
# Description
We want to avoid as many requests to the Dynatrace cluster and to the K8S control plane as possible. Therefore in addtion to the config map, we want to cache the connection info (tenant UUID, token, endpoints, communication hosts) in the Dynakube status. Reconcilers have hold of the DynaKube object anyway, so they don't need to do an extra request to read the required infos from the config map. We can't drop the config map as we need it to map the connection infos to environment variables in the OneAgent daemon set.

## How can this be tested?
Deploy operator and a danykube, check if the connection info config maps are created and if the connection info is part of the dynakube status.

## Checklist


src/controllers/dynakube/connectioninfo/reconciler.go
```diff
@@ -56,121 +55,80 @@ func (r *Reconciler) reconcileOneAgentConnectionInfo() error {
...
-		return err
-	}
-	return nil
+func (r *Reconciler) updateDynakubeActiveGateStatus(connectionInfo dtclient.ActiveGateConnectionInfo) {
+	r.dynakube.Status.ActiveGate.ConnectionInfoStatus.TenantUUID = connectionInfo.TenantUUID
+	r.dynakube.Status.ActiveGate.ConnectionInfoStatus.TenantToken = connectionInfo.TenantToken
+	r.dynakube.Status.ActiveGate.ConnectionInfoStatus.Endpoints = connectionInfo.Endpoints
 }
 
 func (r *Reconciler) createTenantTokenSecret(secretName string, connectionInfo dtclient.ConnectionInfo) error {
```
0sewa0, line 134: you moved out the config map creation to their respective reconcilers (to the activegate/oneagent reconcilers)
in order to make the connectioninfo reconciler only care about updating the status, but `createTenantTokenSecret` remained. Why ?

0sewa0 (commented):
only 1 tiny nitpick

0sewa0 (approved):

PR #1707, opened 2023-04-18
Adds default limits for the injected init-container

0sewa0:
# Description

If no init resource requests/limits is set and we use the CSI driver, then default resource requests/limits are set.

About the default limits:
- 30m cpu and 30Mi memory made the init container work without OOM after multiple times, however it was rather slow (took around 30 sec to get ready)
- 100m cpu and 60Mi memory made it faster by a lot (~5sec to get ready), and I increased the memory just to be safe
- so I chose:
    - requests: 30m cpu and 30Mi memory
    - limits: 100m cpu and 60Mi memory

updated the `InitResources` field to use a pointer, so we can tell if it was deliberately unset which we should respect, or was not set at all, in which case the defaults should be used  

## How can this be tested?
Deploy something that will be injected + use the csi driver, and check if the resource requests/limits are present for the init container

## Checklist


src/webhook/mutation/pod_mutator/init_container.go
```diff
@@ -27,7 +27,25 @@ func createInstallInitContainerBase(webhookImage, clusterID string, pod *corev1.
 			{Name: config.K8sNodeNameEnv, ValueFrom: kubeobjects.NewEnvVarSourceForField("spec.nodeName")},
 		},
 		SecurityContext: securityContextForInitContainer(pod),
-		Resources:       *dynakube.InitResources(),
+		Resources:       *initContainerResources(dynakube),
+	}
+}
+
+func initContainerResources(dynakube dynatracev1beta1.DynaKube) *corev1.ResourceRequirements {
+	initResources := dynakube.InitResources()
```
luhi-DT, line 35: can we call that `customInitResources` because othwise its hard to read

PR #1720, opened 2023-04-21
Add platform detection to makefile and e2e tests

gkrenn:
# Description

Recently the ability to automatically detect the platform was added to our helm release.

This PR updates the makefile and the e2e tests to use this ability.
Note that the e2e tests no longer use the GitHub releases but helm for installation of specific versions.

## How can this be tested?

- `make deploy` and `make undeploy` do not require a platform.
- `make deploy` is idempotent => `helm upgrade` is used for installation to achieve that

## Checklist


luhi-DT:
Ci fails because of:
`test/helpers/components/operator/installation.go:74:26: printf: fmt.Sprintf format %s has arg withCsi of wrong type bool, see also https://pkg.go.dev/fmt#hdr-Printing (govet)
  		helm.WithArgs("--set", fmt.Sprintf("csidriver.enabled=%s", withCsi)),`

hack/make/deploy/kubernetes.mk
```diff
@@ -1,24 +1,32 @@
...
+			--create-namespace \
+			--install \
+			--atomic \
 			--set installCRD=true \
-			--set platform="kubernetes" \
-			--set csidriver.enabled=true \
+			--set csidriver.enabled=$(ENABLE_CSI) \
 			--set manifests=true \
-			--set image="$(IMAGE_URI)" | kubectl apply -f -
+			--set image="$(IMAGE_URI)"
```
0sewa0, line 13: why did we keep the old `deploy/kubernetes` (and others) entrypoints ?
if we are going to keep them, shouldn't they act the same? (use helm upgrade so they are managed by helm)
also the openshift entrypoints were not updated, why? (they don't use `$(ENABLE_CSI)`)

0sewa0 (changes requested):

hack/make/deploy/kubernetes.mk
```diff
@@ -29,26 +37,25 @@ deploy/gke-autopilot: manifests/crd/helm
 			--set manifests=true \
 			--set image="$(IMAGE_URI)" | kubectl apply -f -
 
-
+## Undeploy the operator in a cluster configured in ~/.kube/config where platform and k8s version are autodetected
+undeploy:
+	-helm uninstall dynatrace-operator \
```
0sewa0, line 42:
```suggestion
	helm uninstall dynatrace-operator \
```

hack/make/deploy/kubernetes.mk
gkrenn: I kept the entries on purpose to give us the possibility to still deploy/undeploy the operator without helm from the makefile.

Will rename `deploy` to `deploy/helm` and move it in it's own file. 

Concerning the openshift entries, I will adapt them too

hack/make/deploy/kubernetes.mk
gkrenn: "-" would swallow the error. We agreed on removing it.

0sewa0 (approved):

PR #1722, opened 2023-04-24
Adds readonly csi feature-flag + related webhook changes

0sewa0:
# Description
Adds the `feature.dynatrace.com/injection-readonly-volume` feature-flag

Which causes the webhook oneagent injection change in a way that the csi-volume can be `readOnly:True`

The following additional volumes and volume-mounts are added in case the feature-flag is enabled:
```
# init container
volumeMounts:
- mountPath: /mnt/agent-conf
  name: oneagent-agent-conf
...
# app container
volumeMounts:
  - mountPath: /opt/dynatrace/oneagent-paas/agent/conf
    name: oneagent-agent-conf
...
volumes:
- emptyDir: {}
  name: oneagent-agent-conf
```

```
# app container
volumeMounts:
  - mountPath: /opt/dynatrace/oneagent-paas/datastorage
    name: oneagent-datastorage
...
volumes:
- emptyDir: {}
  name: oneagent-datastorage
```

```
# app container
volumeMounts:
  - mountPath: /opt/dynatrace/oneagent-paas/log
    name: oneagent-log
...
volumes:
- emptyDir: {}
  name: oneagent-log
```

+1 additional envvar is also added to the injected init-container: (this is always added, just not set to true by default)
```
- name: CSI_VOLUME_READONLY
   value: "true"
```

## How can this be tested?
When the feature-flag is set, the injected pod should be able to start, but the agent will not work properly, that will be added in a follow-up.
Checking the deployed Pod's injected yaml, to verify if mentioned changes are present.

## Checklist
- [x] Unit tests have been updated/added
- [x] PR is labeled accordingly
- [x] I have read and understood the [contribution guidelines](https://github.com/Dynatrace/dynatrace-operator/blob/main/CONTRIBUTING.md)

PR #1737, opened 2023-04-26
Copies agent configs from readonly csi volume to emptyDir using standalone init

0sewa0:
# Description

Follow up of https://github.com/Dynatrace/dynatrace-operator/pull/1722

In case `feature.dynatrace.com/injection-readonly-volume` feature flag is used the `install-oneagent` init-container will move the configs from the readonly csi-volume to an emptyDir that is mounted to the app container where these configs are expected.

## How can this be tested?
Use the `feature.dynatrace.com/injection-readonly-volume` feature-flag while using a that uses the CSI driver.
Deploy/Inject  sample-apps
Shell into a sample pod and check the `/opt/dynatrace/oneagent-paas/agent/conf` path, and it should have the config files that are originally from the CSI volume. (its a good indicator if the `_ruxitagentproc.conf` is present)

## Checklist


waodim (approved):
Tested it - LGTM

PR #1741, opened 2023-04-26
Revert digest usage for tenant-registry images

0sewa0:
As the tenant-registry is not made to be used with digest, this change goes back to the previous solution of using the build-label on the images of the tenant-registry in order to determine the version, so we can update the oneagent/activegate when necessary.

## How can this be tested?
deploy oneagents/activegates
check if the `image` field only uses the tag, and check that the version label on the statefulset and daemonset is set.

## Checklist

### DIFF
--- a/src/controllers/dynakube/dynakube_controller.go
+++ b/src/controllers/dynakube/dynakube_controller.go
@@ -67,7 +67,7 @@ func NewDynaKubeController(kubeClient client.Client, apiReader client.Reader, sc
 		config:                 config,
 		operatorNamespace:      os.Getenv(kubeobjects.EnvPodNamespace),
 		clusterID:              clusterID,
-		digestProvider:         version.GetImageDigest,
+		digestProvider:         version.GetImageVersion,
 	}
 }
 
@@ -91,7 +91,7 @@ type Controller struct {
 	config                 *rest.Config
 	operatorNamespace      string
 	clusterID              string
-	digestProvider         version.ImageDigestFunc
+	digestProvider         version.ImageVersionFunc
 }
 
 // Reconcile reads that state of the cluster for a DynaKube object and makes changes based on the state read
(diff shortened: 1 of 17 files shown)


src/controllers/dynakube/dynakube_controller.go
```diff
@@ -67,7 +67,7 @@ func NewDynaKubeController(kubeClient client.Client, apiReader client.Reader, sc
 		config:                 config,
 		operatorNamespace:      os.Getenv(kubeobjects.EnvPodNamespace),
 		clusterID:              clusterID,
-		digestProvider:         version.GetImageDigest,
+		digestProvider:         version.GetImageVersion,
```
luhi-DT, line 70: should be renamed to versionProvider i guess

src/controllers/dynakube/version/activegate.go
```diff
@@ -11,13 +11,13 @@ import (
 type activeGateUpdater struct {
 	dynakube   *dynatracev1beta1.DynaKube
 	dtClient   dtclient.Client
-	digestFunc ImageDigestFunc
+	digestFunc ImageVersionFunc
```
luhi-DT, line 14: should be renamed to versionFunc i guess

src/controllers/dynakube/version/updater.go
```diff
@@ -139,6 +140,39 @@ func updateVersionStatus(
...
+
+	log.Info("updating image version info for tenant registry image",
+		"image", imageUri,
+		"oldImageID", target.ImageID,
+		"oldVersion", target.Version)
+
+	if taggedRef, ok := ref.(reference.NamedTagged); ok {
+		imageVersion, err := imageVersionFunc(ctx, imageUri, dockerCfg)
+		if err != nil {
+			log.Error(err, "failed to get image digest, falling back to tag")
```
luhi-DT, line 163: This log is wrong as you actually fail to get the image version and not the image digest

luhi-DT (dismissed):
check comments from before

chrismuellner (approved):

PR #1757, opened 2023-05-04
General fixes for the e2e tests

0sewa0:
# Description

Commit-by-commit:
- `Fix sample namespace creation`
  - Sample namespace creation was removed from the `SampleApp` logic, which makes sense for cloudnative, as we should create the sample namespace before we create the dynakube to avoid the race condition for creating monitoringNodes in the namespace secret. However this broke application monitoring tests, where this race condition was not a thing. So I made the `SampleApp` logic handle both cases
- `Fix specific agent version test`
  - This was checking for the incorrect thing. In case of cloudnative the `version` field in the dynakube only affects the oneagent daemonset and has nothing to do with codemodules.
- `Make csi cleanup more reliable` / `Fix csi daemonset logic (again)`
   - Use code we already have to exec into pods
   - Now that we use helm to deploy the operator in the e2e tests, the logic for getting the pods of the daemonset was incorrect as the `daemonset.Labels` had an extra helm label that the pods didn't so we always got no pods when searching for them. `daemonSet.Spec.Selector.MatchLabels` is the way we should have always done it
-  `Fix classic agent test`
   - Had unnecessary complexity (deploy samples to check they don't get injected if no agent is there is kinda meh)
   - Had a bit of a race condition, as we didn't have a `wait.For` the condition to check if classic agent inject or not
-  `Fixes istio codemodules test`
   - Labeled the sample namespace incorrectly (`namespaceBuilder.WithLabels` does not append, but overwrites)
-  `Fixes upgrade test case`
   - Helm can detect the platform correctly for deploying the operator on the latest version, but this is not true for older version, so we still have to provide `platform` value for helm if we want to install older releases. 

## How can this be tested?
Run as many e2e tests as you can.

⚠️ This change will not make them bulletproof, there still can be timeouts if you cluster is slow or something,
⚠️  for the upgrade test make sure that your local helm repo for `dynatrace` does not exist for does not refer to `master`

## Checklist


test/scenarios/cloudnative/specific_agent_version/specific_agent_version.go
```diff
@@ -79,10 +66,13 @@ func getAvailableVersions(secret tenant.Secret, t *testing.T) []string {
...
 
-func assessVersionChecks(builder *features.FeatureBuilder, version version.SemanticVersion, sampleApp sample.App) {
-	builder.Assess("restart sample apps", sampleApp.Restart)
-	cloudnative.AssessSampleInitContainers(builder, sampleApp)
-	builder.Assess("check env vars of init container", checkVersionInSampleApp(version, sampleApp))
+func assessVersionChecks(testDynakube dynatracev1beta1.DynaKube) features.Func {
+	return func(ctx context.Context, t *testing.T, environmentConfig *envconf.Config) context.Context {
+		daemonset, err := oneagent.Get(ctx, environmentConfig.Client().Resources(), testDynakube)
+		require.NoError(t, err)
+		require.Contains(t, daemonset.Spec.Template.Spec.Containers[0].Image, testDynakube.OneAgentVersion())
```
aorcholski, line 73: does `testDynakube.OneAgentVersion()` return image name? (for instance `quay.io/dynatrace-operator:v0.10.4` or is it `v0.10.4`). The function name is misleading if it is image name.

aorcholski (changes requested):

test/scenarios/cloudnative/specific_agent_version/specific_agent_version.go
gkrenn: We check if the string **contains** the version, so it should be fine. 
The OneAgentVersion returns the version without the image path as you said.

gkrenn (approved):

aorcholski (approved):

PR #1762, opened 2023-05-04
Optimise Istio reconciliation to increase performance

gkrenn:
# Description

This PR updates the Istio reconciliation to run more stable and prevent issues caused by slow Istio reconciliation. 

Things that have been implemented to optimise Istio reconciliation:
- Dropped Istio detection on each DynaKube reconciliation
- Request specific Istio group version to detect Istio in the validation webhook
- Moved Istio reconcile to be the last step in the DynaKube reconciliation

## How can this be tested?

Run Istio e2e tests and check if they pass.

## Checklist


src/controllers/dynakube/istio/serviceentry.go
```diff
@@ -76,6 +76,9 @@ func handleIstioConfigurationForServiceEntry(istioConfig *configuration) (bool,
 	} else if probe == kubeobjects.ProbeUnknown {
 		log.Error(err, "istio: failed to query ServiceEntry")
 		return false, err
+	} else if probe == kubeobjects.ProbeTypeNotFound {
+		log.Error(err, "istio: service entry type not found, skipping creation")
```
waodim:
```suggestion
		log.Error(err, "istio: ServiceEntry type not found, skipping creation")
```

src/controllers/dynakube/istio/serviceentry.go
```diff
@@ -77,7 +77,7 @@ func handleIstioConfigurationForServiceEntry(istioConfig *configuration) (bool,
 		log.Error(err, "istio: failed to query ServiceEntry")
 		return false, err
 	} else if probe == kubeobjects.ProbeTypeNotFound {
-		log.Error(err, "istio: service entry type not found, skipping creation")
+		log.Info("istio: service entry type not found, skipping creation")
```
waodim, line 81:
```suggestion
		log.Info("istio: ServiceEntry type not found, skipping creation")
```

src/controllers/dynakube/istio/reconciler.go
```diff
@@ -60,15 +59,7 @@ func (reconciler *Reconciler) initializeIstioClient(config *rest.Config) (istioc
...
-	enabled, err := CheckIstioEnabled(reconciler.config)
-	if err != nil {
-		return false, fmt.Errorf("istio: failed to verify Istio availability: %w", err)
-	}
-	log.Info("istio: status", "enabled", enabled)
-
-	if !enabled {
-		return false, nil
-	}
+	log.Info("istio: reconciling")
```
chrismuellner, line 62: isn't this `istio: ` prefix outdated and should be replaced with a logger name?

src/controllers/dynakube/istio/serviceentry.go
```diff
@@ -71,11 +71,15 @@ func buildServiceEntryIP(meta metav1.ObjectMeta, host string, port uint32) *isti
...
 	probe, err := kubeobjects.KubernetesObjectProbe(ServiceEntryGVK, istioConfig.instance.GetNamespace(), istioConfig.name, istioConfig.reconciler.config)
-	if probe == kubeobjects.ProbeObjectFound {
+	switch probe {
+	case kubeobjects.ProbeObjectFound:
 		return false, nil
-	} else if probe == kubeobjects.ProbeUnknown {
+	case kubeobjects.ProbeUnknown:
 		log.Error(err, "istio: failed to query ServiceEntry")
 		return false, err
+	case kubeobjects.ProbeTypeNotFound:
```
chrismuellner, line 80: should this just be the `default` case?

src/webhook/validation/istio.go
```diff
@@ -11,11 +14,28 @@ const (
...
+		enabled, err := CheckIstioInstalled(dv.cfg)
 		if !enabled || err != nil {
 			return errorNoResources
 		}
 	}
 
 	return ""
 }
+
+func CheckIstioInstalled(cfg *rest.Config) (bool, error) {
```
chrismuellner, line 26: this should be in the `istio` package

src/webhook/validation/istio_test.go
```diff
@@ -16,3 +21,70 @@ func TestNoResourcesAvailable(t *testing.T) {
...
+			}
+
+			resources = &metav1.APIGroup{
+				Name: "networking.istio.io",
+				Versions: []metav1.GroupVersionForDiscovery{
+					{GroupVersion: "v1alpha3", Version: "v1alpha3"},
+				},
+			}
+		default:
+			// t.Logf("unexpected request: %s", req.URL.Path)
```
chrismuellner, line 75: 🤔

src/webhook/validation/istio.go
```diff
@@ -11,11 +14,28 @@ const (
...
+
+	_, err = discoveryclient.ServerResourcesForGroupVersion(istio.IstioGVR)
+	if errors.IsNotFound(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
```
chrismuellner, line 40:
```suggestion
	return err == nil, err
```

chrismuellner (changes requested):

src/webhook/validation/istio.go
0sewa0: unless you want to check if istio exists anywhere else, but it could be private if not used anywhere else

src/controllers/dynakube/istio/serviceentry.go
gkrenn: No, this should be correct, per default everything should be fine and nothing has to be done.

The probe type not found check is there to skip creation if istio CRDs were not installed.

src/webhook/validation/istio.go
gkrenn: Made the function private

src/controllers/dynakube/istio/reconciler.go
gkrenn: You are right, will remove the "istio: " prefix

src/controllers/dynakube/istio/reconciler_test.go
```diff
@@ -72,7 +72,46 @@ func TestIstioClient_BuildDynatraceVirtualService(t *testing.T) {
...
+		},
+	}
+
+	updated, err := reconciler.Reconcile(instance, []dtclient.CommunicationHost{})
+
+	assert.NoError(t, err)
+	assert.False(t, updated)
+}
+
+func TestController_ReconcileIstio2(t *testing.T) {
```
0sewa0, line 113: why?
what is the difference between `TestController_ReconcileIstio` and `TestController_ReconcileIstio2` 😂 ?

src/controllers/dynakube/istio/serviceentry.go
```diff
@@ -71,20 +71,24 @@ func buildServiceEntryIP(meta metav1.ObjectMeta, host string, port uint32) *isti
...
+	case kubeobjects.ProbeObjectFound:
 		return false, nil
-	} else if probe == kubeobjects.ProbeUnknown {
-		log.Error(err, "istio: failed to query ServiceEntry")
+	case kubeobjects.ProbeUnknown:
+		log.Error(err, "failed to query ServiceEntry")
 		return false, err
+	case kubeobjects.ProbeTypeNotFound:
+		log.Info("ServiceEntry type not found, skipping creation")
+		return false, nil
```
0sewa0, line 82: this logic is the same (except the log) for the VirtualService and ServiceEntry logic.

wouldn't it be nicer to wrap the `kubeobjects.KubernetesObjectProbe` call with this logic, instead of calling it twice

0sewa0 (changes requested):

src/controllers/dynakube/istio/reconciler_test.go
gkrenn: Ups 😅

PR #1771, opened 2023-05-08
Remove injected check for classicFullstack e2e test

luhi-DT:
# Description

Our e2e test for classicFullStack runs flaky, as the oneAgent needs more time to startup properly, than our sample apps come up. The problem is that the agent reports its ready state as ready, even though its not yet 100% ready to inject.

## How can this be tested?
run the classic e2e test (`make test/e2e/classic`)

## Checklist


waodim (approved):
Tested it several times - works for me.

PR #1723, opened 2023-04-24
Adds applicationMonitoring without CSIDriver e2e test 

waodim:
# Description
This is a scenario where the operator is deployed with applicationMonitoring mode with the CSIDriver disabled. NOTE: a timeout was added to the make target for the appOnly e2e tests because of timeout issues when running the appOnly e2e tests.

## How can this be tested?
Run the e2e tests.

## Checklist
- [x] PR is labeled accordingly
- [x] I have read and understood the [contribution guidelines](https://github.com/Dynatrace/dynatrace-operator/blob/main/CONTRIBUTING.md)

PR #1774, opened 2023-05-09
Adds e2e scenario which tests appOnly with readOnly CSI volume

waodim:
# Description
With this scenario application monitoring is tested with the read only csi volume feature flag enabled. NOTE: added a timeout to the make target, as otherwise the appMon tests would fail due to timeout.

## How can this be tested?
run `make test/e2e/applicationmonitoring`

## Checklist

### DIFF
--- a/hack/make/tests/e2e.mk
+++ b/hack/make/tests/e2e.mk
@@ -58,7 +58,7 @@ test/e2e/applicationmonitoring/default: manifests/crd/helm
 test/e2e/applicationmonitoring/labelversion: manifests/crd/helm
 	go test -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)"  -count=1 ./test/scenarios/applicationmonitoring  -run ^TestLabelVersionDetection$  $(SKIPCLEANUP)
 
-## Runs Application Monitoring e2e test only
+## Runs Application Monitoring readonly csi-volume e2e test only
 test/e2e/applicationmonitoring/readonlycsivolume: manifests/crd/helm
 	go test -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)"  -count=1 ./test/scenarios/applicationmonitoring  -run ^TestReadOnlyCSIVolume$  $(SKIPCLEANUP)
 


hack/make/tests/e2e.mk
```diff
@@ -52,7 +52,7 @@ test/e2e/cloudnative/upgrade: manifests/crd/helm
 
 ## Runs Application Monitoring e2e test only
 test/e2e/applicationmonitoring: manifests/crd/helm
-	go test -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -count=1 ./test/scenarios/applicationmonitoring $(SKIPCLEANUP)
+	go test -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 20m -count=1 ./test/scenarios/applicationmonitoring $(SKIPCLEANUP)
```
0sewa0, line 55: same as in https://github.com/Dynatrace/dynatrace-operator/pull/1723#discussion_r1192016092 😉

test/scenarios/applicationmonitoring/read_only_csi_volume.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+func checkMountedVolumes(sampleApp sample.App) features.Func {
+	return func(ctx context.Context, t *testing.T, environmentConfig *envconf.Config) context.Context {
+		resources := environmentConfig.Client().Resources()
+
+		err := deployment.NewQuery(ctx, resources, client.ObjectKey{
+			Name:      sampleApp.Name(),
+			Namespace: "sample-dynakube",
```
0sewa0, line 86:
```suggestion
			Namespace: sampleApp.Namespace().Name,
```

0sewa0, line 86:
```suggestion
			Namespace: sampleApp.Namespace().Name,
```

test/scenarios/applicationmonitoring/read_only_csi_volume.go
```diff
@@ -0,0 +1,97 @@
...
+		}).ForEachPod(func(podItem corev1.Pod) {
+			listCommand := shell.ListDirectory(configPath)
+			result, err := pod.Exec(ctx, resources, podItem, "sample-dynakube", listCommand...)
```
0sewa0, line 89:
```suggestion
			result, err := pod.Exec(ctx, resources, podItem, sampleApp.Namespace().Name, listCommand...)
```

test/scenarios/applicationmonitoring/read_only_csi_volume.go
```diff
@@ -0,0 +1,97 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+const (
+	csiReadOnlyVolumeEnvVar = "CSI_VOLUME_READONLY"
+	initContainerName       = "install-oneagent"
+	configPath              = "/opt/dynatrace/oneagent-paas/agent/conf"
+	ruxitAgentProcFile      = "_ruxitagentproc.conf"
```
0sewa0, line 33: don't we have already consts for (some of) these somewhere in the codebase? 🤔

0sewa0 (changes requested):

test/scenarios/applicationmonitoring/read_only_csi_volume.go
waodim: Yes for the initContainerName most probably. I can check for the others. If so, should I simply export them in order to use them here?

0sewa0: if its not already exported then yes I think its ok to export them to use them

waodim: To be honest I seem only to find a constant for the default install path under the webhook package. Can you confirm, that the others are not elsewhere defined?

0sewa0: https://github.com/Dynatrace/dynatrace-operator/blob/6cdc2990cd437273436b60988e0dada225efcdb4/src/config/agent_injection.go#L31

https://github.com/Dynatrace/dynatrace-operator/blob/6cdc2990cd437273436b60988e0dada225efcdb4/src/webhook/config.go#L55

https://github.com/Dynatrace/dynatrace-operator/blob/6cdc2990cd437273436b60988e0dada225efcdb4/src/webhook/mutation/pod_mutator/oneagent_mutation/config.go#L34-L35 (this is not exported)

`_ruxitagentproc.conf` is the only that is missing (so that should be fine)

waodim: Thank you very much, I had a brain lag don't know how I could not find those :sweat_smile:

hack/make/tests/e2e.mk
0sewa0: should rebase on `main` so you don't overwrite it

hack/make/tests/e2e.mk
```diff
@@ -58,11 +58,14 @@ test/e2e/applicationmonitoring/default: manifests/crd/helm
 test/e2e/applicationmonitoring/labelversion: manifests/crd/helm
 	go test -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)"  -count=1 ./test/scenarios/applicationmonitoring  -run ^TestLabelVersionDetection$  $(SKIPCLEANUP)
 
+## Runs Application Monitoring e2e test only
```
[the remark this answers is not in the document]
waodim: That's what copy+paste does to one :sweat_smile: thanks for pointing out :)

0sewa0 (approved):

PR #1779, opened 2023-05-12
Improves error for unexpected type for image ref in image installer

0sewa0:
Improves the error message when the provided image reference for the image installer does not have a digest.

This happens in case the image for the code-modules is in a registry that is not accessible by the operator, to set the status of the dynakube correctly (by getting and storing the digest), which will cause the operator to fall back to the "raw" text version of the provided image, which tends to be specified with a tag instead of a digest.

## How can this be tested?
Try to reproduce the above mentioned scenario (with a proxy, or something).
And now the error message should be more specific

## Checklist

### DIFF
--- a/src/installer/image/installer.go
+++ b/src/installer/image/installer.go
@@ -30,7 +30,7 @@ func NewImageInstaller(fs afero.Fs, props *Properties) (*Installer, error) {
 	}
 	canonRef, ok := ref.(reference.Canonical)
 	if !ok {
-		return nil, errors.New("unexpected type of image reference provided to image installer")
+		return nil, errors.Errorf("unexpected type of image reference provided to image installer, expected reference with digest but received %s", props.ImageUri)
 	}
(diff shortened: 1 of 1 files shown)


src/installer/image/installer.go
```diff
@@ -30,7 +30,7 @@ func NewImageInstaller(fs afero.Fs, props *Properties) (*Installer, error) {
 	}
 	canonRef, ok := ref.(reference.Canonical)
 	if !ok {
-		return nil, errors.New("unexpected type of image reference provided to image installer")
+		return nil, errors.Errorf("unexpected type of image reference provided to image installer, expected reference with digest but received %s", props.ImageUri)
```
luhi-DT, line 33: we should add the imageUri to the err in line 29 aswell

luhi-DT (changes requested):

luhi-DT (approved):

PR #1782, opened 2023-05-15
fixed helm unit tests to work with updated plugin

wepudt:
# Description

updated the unit tests to work with new plugin version

## How can this be tested?
`make test/helm/unit`

## Checklist

### DIFF
--- a/hack/helm/install-unittest-plugin.sh
+++ b/hack/helm/install-unittest-plugin.sh
@@ -1,33 +1,4 @@
 #!/usr/bin/env bash
 
-HELM_PLUGINS="$(helm env HELM_PLUGINS)"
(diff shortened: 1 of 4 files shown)


wepudt:
need to upated environemnt first

luhi-DT (dismissed):

hack/helm/install-unittest-plugin.sh
```diff
@@ -26,7 +30,7 @@ fi
```
luhi-DT, line 30: Maybe we should add a comment here that we use curl because of the api.github request limit to keep it for future knowledge :)

luhi-DT (approved):

PR #1809, opened 2023-05-26
Optimize renovate dependency parsing in makefiles and third party licenses

wepudt:
# Description

reworked how third party licenses and prerequisites are integrated into renovate to simplify the regex used.

## How can this be tested?
1. create a fork of the repo, 
2. set up renovate and 
3. merge the pull-request onto main then 
4. choose a preceding version number for all packages used so renovate finds an updated version and reports it

## Checklist
- [ ] ~~Unit tests have been updated/added~~


hack/make/prerequisites.mk
```diff
@@ -1,3 +1,14 @@
+#renovate depName=sigs.k8s.io/kustomize/kustomize/v5
+kustomize_version=v5.0.3
+#renovate depName=sigs.k8s.io/controller-tools/cmd
+controller_gen_version=v0.12.0
+# renovate depName=github.com/golangci/golangci-lint
+golang_ci_cmd_version=v1.52.2
+# renovate depName=github.com/daixiang0/gci
+gci_version=v0.10.1
+# renovate depName=golang.org/x/tools
```
luhi-DT, line 9: why is this not like the original package name? `golang.org/x/tools/cmd/goimports`

luhi-DT (changes requested):

hack/make/prerequisites.mk
StefanHauth: I think that's because with the added sub-packages, renovate is not able to find new versions. It has to look at the root of a package.

StefanHauth: That's what caused the painful regexes prior to this change.

StefanHauth (approved):

PR #1827, opened 2023-06-01
Simplify dtclient timeout logic

0sewa0:
# Description

Undoes the previous commit about using the `NewRequestWithContext` with a quick work around timeout, as it cause unexplainable problems (1 specific request always got its context immediately canceled)

So now we just set a default timeout on the client itself

## How can this be tested?
Run e2e test, and they should all pass.

## Checklist
- [ x] Unit tests have been updated/added
- [ x] PR is labeled accordingly
- [ x] I have read and understood the [contribution guidelines](https://github.com/Dynatrace/dynatrace-operator/blob/main/CONTRIBUTING.md)

### DIFF
--- a/hack/make/prerequisites.mk
+++ b/hack/make/prerequisites.mk
@@ -28,7 +28,7 @@ CONTROLLER_GEN=$(shell hack/build/command.sh controller-gen)
 
 prerequisites/setup-pre-commit:
 	go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(golang_ci_cmd_version)
-	go install github.com/daixiang0/gci@v$(gci_version)
+	go install github.com/daixiang0/gci@$(gci_version)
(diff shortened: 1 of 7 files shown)


hack/make/prerequisites.mk
```diff
@@ -28,7 +28,7 @@ CONTROLLER_GEN=$(shell hack/build/command.sh controller-gen)
 
 prerequisites/setup-pre-commit:
 	go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(golang_ci_cmd_version)
-	go install github.com/daixiang0/gci@$(gci_version)
```
luhi-DT, line 31: Please keep it without the `v`

luhi-DT (changes requested):

waodim (commented):
Tested it and can confirm it works.

Besides the minor remark by @luhi-DT I would approve it.

luhi-DT (approved):

PR #1832, opened 2023-06-02
Update e2e tests to use automatic platform detection

gkrenn:
# Description

Removes the dependency of the e2e tests on the platform env var.
Openshift is automatically detected by checking if SCCs group version is available in the cluster.

## How can this be tested?

Run e2e tests in openshift and kubernetes.

## Checklist


src/kubeobjects/platform.go
```diff
@@ -15,22 +20,54 @@ const (
...
 
-func ResolvePlatformFromEnv() Platform {
-	switch os.Getenv(platformEnvName) {
-	case openshiftPlatformEnvValue:
-		return Openshift
-	case kubernetesPlatformEnvValue:
-		fallthrough
-	default:
-		return Kubernetes
+const SccGVR = "security.openshift.io/v1"
```
0sewa0, line 23: 😭 🤮  (I don't have a better name, `OpenshiftSecurityGVR`, but its not much better)

src/kubeobjects/platform.go
```diff
@@ -15,22 +20,54 @@ const (
...
+
+type discoveryClientCreation interface {
```
0sewa0, line 25:
```suggestion
type discoveryClientProvider interface {
```

src/kubeobjects/platform.go
```diff
@@ -15,22 +20,54 @@ const (
...
+	getDiscoveryClient() (discovery.DiscoveryInterface, error)
+}
+
+type DiscoveryClientResolver struct{}
+
+func (p *DiscoveryClientResolver) getDiscoveryClient() (discovery.DiscoveryInterface, error) {
```
0sewa0, line 31: THERE IS NOT EVEN THE LETTER `p` IN THE TYPE NAME 😂 
```suggestion
func (resolver *DiscoveryClientResolver) getDiscoveryClient() (discovery.DiscoveryInterface, error) {
```

0sewa0 (changes requested):

src/kubeobjects/platform.go
StefanHauth, line 23: Does this have to be public?

test/scenarios/cloudnative/proxy/proxy.go
```diff
@@ -60,7 +60,7 @@ func withProxy(t *testing.T, proxySpec *dynatracev1beta1.DynaKubeProxy) features
 	assess.InstallOperatorFromSourceWithCustomNamespace(builder, operatorNamespaceBuilder.Build(), testDynakube)
 
 	// Register proxy create and delete
-	proxy.SetupProxyWithTeardown(builder, testDynakube)
+	proxy.SetupProxyWithTeardown(builder, t, testDynakube)
```
StefanHauth, line 63: Matter of taste: testing.T should always be the first paramter.

src/kubeobjects/platform.go
```diff
@@ -15,22 +20,54 @@ const (
...
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
+}
+
+type PlatformResolver struct {
+	discoveryClientCreation
+}
```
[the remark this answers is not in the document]
gkrenn: Will do

src/kubeobjects/platform.go
```diff
@@ -1,9 +1,14 @@
 package kubeobjects
```
0sewa0, line 1: there is a `kubeobjects` under `test` (for example: https://github.com/Dynatrace/dynatrace-operator/blob/03695c1d95fe519ea50492906016410452ed8926/test/helpers/kubeobjects/environment/environment.go#L14)

as these functions are only used for testing we should have it under `/test`

0sewa0 (changes requested):

0sewa0 (approved):

PR #1836, opened 2023-06-05
Add extra check for secret creation during pod mutation

0sewa0:
# Description

Adds an extra check during the `ensure...` secret creation during the webhook mutation for the super unlucky scenario when the operator and the webhook are creating secrets at the same time.

Also fixes error handling that we don't log the error/stacktrace multiple times.

## How can this be tested?
Try to recreate the unlucky scenario. (you shouldn't be able to)

## Checklist


src/webhook/mutation/pod_mutator/oneagent_mutation/mutator.go
```diff
@@ -68,13 +68,13 @@ func (mutator *OneAgentPodMutator) ensureInitSecret(request *dtwebhook.MutationR
...
+		if err != nil && !k8serrors.IsAlreadyExists(err) {
+			log.Info("failed to create the init secret before oneagent pod injection")
+			return err
 		}
-		log.Info("created the init secret before oneagent pod injection")
+		log.Info("ensured that the init secret is present before oneagent pod injection")
 	} else if err != nil {
-		log.Error(err, "failed to query the init secret before oneagent pod injection")
+		log.Info("failed to query the init secret before oneagent pod injection")
 		return errors.WithStack(err)
```
chrismuellner, line 78: since both these init secret checks (data ingest and oneagent) are pretty similiar, should this also just return the `err`?

src/webhook/mutation/pod_mutator/dataingest_mutation/mutator.go
```diff
@@ -77,14 +76,14 @@ func (mutator *DataIngestPodMutator) ensureDataIngestSecret(request *dtwebhook.M
 		&endpointSecret)
 	if k8serrors.IsNotFound(err) {
 		err := endpointGenerator.GenerateForNamespace(request.Context, request.DynaKube.Name, request.Namespace.Name)
-		if err != nil {
-			log.Error(err, "failed to create the data-ingest endpoint secret before pod injection")
-			return errors.WithStack(err)
```
chrismuellner, line 82: since we log the error anyway, we could also wrap it with the info log messages.
could be harder to read though, but has the advantage of it being all in a single place

src/webhook/mutation/pod_mutator/oneagent_mutation/mutator.go
0sewa0: `mutator.apiReader.Get` is not our code  so we want to use `errors.WithStack`

src/webhook/mutation/pod_mutator/dataingest_mutation/mutator.go
0sewa0: the log lines shouldn't be far from each other, and logging the exact error twice is just spam.

I though about using `errors.WithMessage` but the problem with that is that there can only be 1 message for an error, so if multiple error handling `if`s add a message only the last (most hight level) one can be seen

src/webhook/mutation/pod_mutator/oneagent_mutation/mutator.go
0sewa0: (fixed it in the other place though 😅 )

chrismuellner (approved):

PR #1884, opened 2023-06-21
Adds automated sbom creation and upload to publish-image step

luhi-DT:
# Description

This PR adds logic in order to generate a SBOM using trify and attaches it to each container registry (only Dockerhub and GCR support it) using cosign.

There are two ways to attach an SBOM to an image using cosign:

- cosign attest
- cosign attach

The big benefit of attest is that this already signs the attestation (SBOM), however puts it into a different format (IN-TOTO). Whereas cosign attach leaves the sbom as it is, but would need an extra step to sign it, thats why we should go with cosign attest. 

## How can this be tested?

Push a release Tag

.ATT Image should appear in dockerhub and GCR

**Test the sbom by downloading it and scanning it**

`cosign verify-attestation <key> --type cyclonedx <operator-image> > yourSbom.jsonl` 
`trivy sbom ./yourSbom.jsonl` in order to scan it 

## Checklist


.github/actions/upload-sbom/action.yaml
```diff
@@ -0,0 +1,31 @@
+name: Upload SBOM
+description: uploads a sbom for a given image to a registry
+inputs:
+  image:
+    description: full image tag that the sbom will be generated for
+    required: true
+  sbom:
+    description: the sbom to attach
```
chrismuellner, line 8:
```suggestion
    description: sbom of type cyclonedx to attach to image
```

.github/actions/upload-sbom/action.yaml
```diff
@@ -0,0 +1,31 @@
+name: Upload SBOM
+description: uploads a sbom for a given image to a registry
+inputs:
+  image:
+    description: full image tag that the sbom will be generated for
```
chrismuellner, line 5:
```suggestion
    description: full image to generate and attach sbom
```

.github/actions/upload-sbom/action.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Install Cosign
+      uses: sigstore/cosign-installer@dd6b2e2b610a11fd73dd187a43d57cc1394e35f9 # v3.0.5
+    - name: Attach sbom to image
+      shell: bash
+      run: |
+        cosign attach sbom --sbom "${{ inputs.sbom }}" "${{ inputs.image }}" --input-format json --type cyclonedx
+    - name: Attach sbom attestation to image
+      shell: bash
+      run: |
+        cosign attest -y --key env://COSIGN_PRIVATE_KEY  --type cyclonedx --predicate ${{ inputs.sbom }} ${{ inputs.image }}
```
chrismuellner, line 28: does this mean the sbom is only signed with the private key and not keyless in this case?

hack/build/ci/get-image-digest.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
```
chrismuellner, line 1: IMO we should store the variables in env variables in this script and then return them from the step using the `outputs` field to make it more explicit where these values come from.

.github/workflows/publish-images.yaml
```diff
@@ -176,9 +176,78 @@ jobs:
...
+          image: ${{ matrix.url }}/${{ secrets[matrix.repository] }}:${{ needs.prepare.outputs.version }}@${{steps.operator-digest.outputs.digest}}
+          sbom: 'result.json'
+          signing-key: ${{ secrets.COSIGN_PRIVATE_KEY }}
+          signing-password: ${{ secrets.COSIGN_PASSWORD }}
+      - name: Get sbom-image digest
+        id: sbom-digest
+        env:
+          IMAGE: ${{ matrix.url }}/${{ secrets[matrix.repository] }}:${{steps.operator-digest.outputs.digest_formatted}}.sbom
+        run: |
+          hack/build/ci/get-image-digest.sh
```
chrismuellner, line 247: would it make sense to move this to a dedicated action since we use it a bunch of times?

.github/actions/upload-sbom/action.yaml
luhi-DT: yes, i just tested it if i attest it both ways (with key and keyless) it works aswell, will add keyless way aswell

hack/build/ci/get-image-digest.sh
luhi-DT: I agree that the get-image-digest is not pretty and should be a seperate action, however i would tackle that in a seperate PR

.github/workflows/publish-images.yaml
luhi-DT: yes see https://github.com/Dynatrace/dynatrace-operator/pull/1884#discussion_r1237545522

0sewa0 (approved):

PR #1896, opened 2023-06-27
Use raw images instead of pre-configured latest or versioned images.

StefanHauth:
# Description

Images retrieved from tenant with the `:latest` tag come pre-configured. We don't need that as the operator is configuring the agent anyway. Now we use `:raw` instead of `:latest` and e.g. `1.271.0-raw` instead of `1.271.0`.

## How can this be tested?
Deploy the operator and a dynakube and check if oneagent pods startup alright and use a `raw` image.

## Checklist


src/api/v1alpha1/helpers.go
```diff
@@ -54,7 +54,7 @@ func (dk *DynaKube) ActiveGateImage() string {
 	}
 
 	registry := buildImageRegistry(dk.Spec.APIURL)
-	return fmt.Sprintf("%s/linux/activegate:latest", registry)
+	return fmt.Sprintf("%s/linux/activegate:raw", registry)
```
luhi-DT, line 57: you defined a const for `raw` why not use it here aswell?

luhi-DT (changes requested):

src/api/v1alpha1/helpers.go
chrismuellner: why are we updating `v1alpha` helpers? these should not be used anymore

src/api/v1beta1/properties.go
```diff
@@ -38,7 +38,7 @@ const (
 	AuthTokenSecretSuffix                   = "-activegate-authtoken-secret"
 	PodNameOsAgent                          = "oneagent"
 
-	defaultActiveGateImage = "/linux/activegate:latest"
+	defaultActiveGateImage = "/linux/activegate:raw"
```
chrismuellner, line 41: should the default image include the tag or can we add that when building the final image?

src/api/v1beta1/properties.go
```diff
@@ -389,10 +389,10 @@ func (dk *DynaKube) DefaultOneAgentImage() string {
 		return ""
 	}
 
-	tag := api.LatestTag
+	tag := api.RawTag
 	if version := dk.CustomOneAgentVersion(); version != "" {
 		truncatedVersion := truncateBuildDate(version)
-		tag = truncatedVersion
+		tag = truncatedVersion + "-" + api.RawTag
```
chrismuellner, line 395:
```suggestion
		tag += "-" + truncatedVersion
```
the `tag` variable already has a value of `raw`

chrismuellner (changes requested):

chrismuellner:
Please limit yourself to a single label

src/api/v1beta1/properties.go
StefanHauth: yes, but the tag should be "1.234.0-raw", not "raw-1.234.0". One could write `tag = truncatedVersion + "-" + tag` but that's more confusing than using the raw tag a second time.

src/api/consts.go
```diff
@@ -2,4 +2,5 @@ package api
 
 const (
 	LatestTag = "latest"
```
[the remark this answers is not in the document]
StefanHauth: Its still used for synthetic images.

src/api/v1alpha1/helpers.go
StefanHauth: As long as the code exists it should be maintained. If its dead code, it should be removed.

src/api/v1alpha1/helpers.go
```diff
@@ -54,7 +54,7 @@ func (dk *DynaKube) ActiveGateImage() string {
 	}
 
 	registry := buildImageRegistry(dk.Spec.APIURL)
-	return fmt.Sprintf("%s/linux/activegate:latest", registry)
+	return fmt.Sprintf("%s/linux/activegate:%s", registry, api.RawTag)
```
chrismuellner, line 57: this while file is not used and should not be updated at all.

we should probably remove these in a follow up

src/api/v1beta1/properties.go
chrismuellner: I prefer having the version at the end, since it's the optional part.

we would always have raw at the start and then an optional version after it

chrismuellner (changes requested):

StefanHauth:
Closing this PR for now as it's not done by just using `raw` tags. There seems to be more to it, which we need to clarify first.

src/api/v1beta1/properties.go
StefanHauth: I didn't realize that both tags are actually supported by the tenant registry.

PR #1900, opened 2023-06-28
Adds correct labels to topologySpreadConstraints of webhook

luhi-DT:
# Description

Currently the matchLabels for the webhook are hardcoded and static. We should use the same label placeholder as the webhook pod has.

## How can this be tested?
Deploy the operator via helm -> check if topologySpreadConstraints have correct matchLabels and match with the Pods labels

## Checklist


config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
```diff
@@ -56,15 +56,13 @@ spec:
...
-              app.kubernetes.io/name: dynatrace-operator
+              {{- include "dynatrace-operator.webhookLabels" . | nindent 14 }}
         - maxSkew: 1
           topologyKey: "kubernetes.io/hostname"
           whenUnsatisfiable: DoNotSchedule
           labelSelector:
             matchLabels:
-              app.kubernetes.io/component: webhook
-              app.kubernetes.io/name: dynatrace-operator
+              {{- include "dynatrace-operator.webhookLabels" . | nindent 14 }}
```
0sewa0, line 65:
```suggestion
             {{- include "dynatrace-operator.webhookSelectorLabels" . | nindent 14 }}
```

so we don't have the version in it, maybe could cause problem during upgrades 
- the old version's pods will not match ==> topologySpreadConstraints has no affect (probably)

0sewa0 (approved):

PR #1903, opened 2023-06-29
Use pod label to filter for operator components in support archive

StefanHauth:
# Description

Replace the hardcoded filter `app.kubernetes.io/name=dynatrace-operator` for operator components by using the label value from the pod the support archive sub command is running in.

## How can this be tested?
Install the operator via helm using a custom name, run the support archive subcommand and see if operator pod logs and operator manifests are collected correctly.

## Checklist

### DIFF
--- a/src/cmd/support_archive/builder.go
+++ b/src/cmd/support_archive/builder.go
@@ -14,7 +14,9 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


src/cmd/support_archive/builder.go
```diff
@@ -116,6 +119,19 @@ func getLogOutput(tarballToStdout bool, logBuffer *bytes.Buffer) io.Writer {
 	}
 }
 
+func getAppNameLabel(ctx context.Context, pods clientgocorev1.PodInterface) string {
```
luhi-DT, line 122: would it make sense to move this function to a common location, in case we need it later somewhere else?

luhi-DT (approved):

PR #1905, opened 2023-06-29
Updates pull request template

luhi-DT:
# Description

As we have some rules applied that are not written down in the PR template, this adds them.
E.g the guidelines for the PR titles and also that only a single label should be added to the PR

## Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,7 @@
 # Description
 
 Please include the following:
+- a meaningful PR title [guidelines](https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#use-imperative-mood-in-your-commit-message-subject)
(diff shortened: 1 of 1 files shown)


chrismuellner (approved):

.github/pull_request_template.md
```diff
@@ -13,6 +14,6 @@ Please include some guiding steps on how to test this change during a review.
 
 ## Checklist
 - [ ] Unit tests have been updated/added
-- [ ] PR is labeled accordingly
+- [ ] PR is labeled accordingly with a single label
```
chrismuellner, line 17: do we want a checklist item that the commit message documentation was read? 😄

luhi-DT: let's see we still can add more to it afterwards but for now we can keep it like that i think :)

PR #1920, opened 2023-07-05
Always set 'RunAsNonRoot' explicitly on init container

### LINKED ISSUE #1894: install-oneagent container breaks non-root policy
**Describe the bug**
We're deploying [istio-cni](https://istio.io/latest/docs/setup/additional-setup/cni/) that comes with the following pod security context:
```yaml
securityContext:
  fsGroup: 2000
  runAsGroup: 3000
  runAsNonRoot: true
  runAsUser: 1000
  seccompProfile:
    type: RuntimeDefault
```
Dynatrace injects the install-oneagent container with the following container security context:
```yaml
securityContext:
  capabilities:
    drop:
    - ALL
  privileged: false
  readOnlyRootFilesystem: true
  runAsGroup: 0
  runAsUser: 0
```

It should also have `runAsNonRoot: false` so that we don't see the following error:
```bash
message: 'container''s runAsUser breaks non-root policy (pod: "istio-cni-node-ghjgn_istio-system(c7fa6210-a40f-40d9-b838-67aae08fa039)", container: install-oneagent)'
```

**To Reproduce**
1. Create a pod with above pod and security context

**Expected behavior**
Dynatrace injection should work with runAsNonRoot pods. Maybe the injected security context could simply be customized via the helm chart.

**Additional context**
K8s: v1.24.13
Istio: 1.17.2

albertogdd:
Fix #1894 

## Description

Always set `RunAsNonRoot` explicitly on webhook's init container. Before it was only set when its value was `true`.
This caused `Init:CreateContainerConfigError` status on some cases.

## How to reproduce this
- Deploy the operator in cloudNativeFullStack or applicationMonitoring
- Deploy the deployment sample below
- Check the pods status (kubectl get pod php-musl-.. -n sample -o yaml
- See the error message

<details>
<summary>Deployment sample</summary>

```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: php-musl
  name: 02-php-musl
  namespace: sample
spec:
  replicas: 2
  selector:
    matchLabels:
      app: php-musl
  strategy:
    type: Recreate
  template:
    metadata:
      annotations:
        oneagent.dynatrace.com/technologies: php
        oneagent.dynatrace.com/flavor: musl
        oneagent.dynatrace.com/failure-policy: fail
      labels:
        app: php-musl
    spec:
      securityContext:
        runAsNonRoot: true
      containers:
        - image: docker.io/php:fpm-alpine
          securityContext:
            runAsUser: 0
            runAsGroup: 0
          imagePullPolicy: Always
          livenessProbe:
            failureThreshold: 3
            tcpSocket:
              port: 9000
            initialDelaySeconds: 600
            periodSeconds: 30
            successThreshold: 1
            timeoutSeconds: 2
          name: app
          env:
            - name: DT_DEBUGFLAGS
              value: debugBootstrapNative=true
          ports:
            - containerPort: 9000
              protocol: TCP
          readinessProbe:
            failureThreshold: 3
            tcpSocket:
              port: 9000
            periodSeconds: 10
            successThreshold: 1
            timeoutSeconds: 1
          resources:
            requests:
              memory: 64Mi
            limits:
              memory: 128Mi
      restartPolicy: Always
      terminationGracePeriodSeconds: 30
```

</details>

## Checklist

### DIFF
--- a/src/webhook/mutation/pod_mutator/pod_mutator_test.go
+++ b/src/webhook/mutation/pod_mutator/pod_mutator_test.go
@@ -155,7 +155,8 @@ func TestHandlePodMutation(t *testing.T) {
 		require.NotNil(t, initSecurityContext.ReadOnlyRootFilesystem)
 		assert.True(t, *initSecurityContext.ReadOnlyRootFilesystem)
 
-		assert.Nil(t, initSecurityContext.RunAsNonRoot)
+		assert.NotNil(t, initSecurityContext.RunAsNonRoot)
+		assert.False(t, *initSecurityContext.RunAsNonRoot)
 
 		assert.Equal(t, mutationRequest.Pod.Spec.InitContainers[1].Resources, testResourceRequirements)
 		assert.Equal(t, "true", mutationRequest.Pod.Annotations[dtwebhook.AnnotationDynatraceInjected])
--- a/src/webhook/mutation/pod_mutator/pod_mutator_test.go
+++ b/src/webhook/mutation/pod_mutator/pod_mutator_test.go
@@ -156,7 +156,7 @@ func TestHandlePodMutation(t *testing.T) {
 		assert.True(t, *initSecurityContext.ReadOnlyRootFilesystem)
 
 		assert.NotNil(t, initSecurityContext.RunAsNonRoot)
-		assert.False(t, *initSecurityContext.RunAsNonRoot)
+		assert.True(t, *initSecurityContext.RunAsNonRoot)
 
 		assert.Equal(t, mutationRequest.Pod.Spec.InitContainers[1].Resources, testResourceRequirements)
 		assert.Equal(t, "true", mutationRequest.Pod.Annotations[dtwebhook.AnnotationDynatraceInjected])
--- a/src/webhook/mutation/pod_mutator/init_container.go
+++ b/src/webhook/mutation/pod_mutator/init_container.go
@@ -89,9 +89,7 @@ func combineSecurityContexts(baseSecurityCtx corev1.SecurityContext, pod corev1.
 		baseSecurityCtx.RunAsGroup = containerSecurityCtx.RunAsGroup
 	}
 
-	if isNonRoot(&baseSecurityCtx) {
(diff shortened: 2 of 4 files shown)


albertogdd:
WIP: Fixing unit tests

src/webhook/mutation/pod_mutator/pod_mutator_test.go
```diff
@@ -155,7 +155,7 @@ func TestHandlePodMutation(t *testing.T) {
 		require.NotNil(t, initSecurityContext.ReadOnlyRootFilesystem)
 		assert.True(t, *initSecurityContext.ReadOnlyRootFilesystem)
 
-		assert.Nil(t, initSecurityContext.RunAsNonRoot)
+		assert.NotNil(t, initSecurityContext.RunAsNonRoot)
```
albertogdd, line 159: As mentioned in the daily, when doing assert.True here it would pass when running the single unit test but fail when running all the package ones :clown_face:

albertogdd: Fixed

0sewa0: yes its veeery weird,

I think I found the problem, find 
```
func getTestPod() *corev1.Pod {
```

and make it 
```
func getTestPod() corev1.Pod {
```

0sewa0: don't ask me why this is the problem 😅 

(also bunch of things will be red after the change, but fixing them is simple)

albertogdd: Yes I thought about doing something similar!
I will refactor it in a new commit, thanks for checking :+1:

albertogdd: @0sewa0 I tried it out and the end behavior is the same... Which makes sense because in getTestPod() the struct gets instantiated anyways whether you return it as reference or value.

0sewa0 (approved):

PR #1930, opened 2023-07-10
Get http proxy in OneAgent from mounted secret instead of env variable

albertogdd:
## Description

Remove http proxy env variable in os-OneAgent and mount a secret with its value instead (in the same fashion as ActiveGate does, actually using the same secret for both now).

## How can this be tested?

Apply a Dynakube with proxy (either with `value` or `valueFrom`) and check that OneAgent gets a mounted volume on the expected path (`/var/lib/dynatrace/secrets/internal-proxy`). Also, you can check its value in the container shell. (see pictures)

image
image

Check also that the reconciliator creating the proxy secret works, it should create `<dynakubeName>-internal-proxy`.

image

## Checklist


src/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/proxy.go
```diff
@@ -52,27 +53,27 @@ func (mod ProxyModifier) getVolumeMounts() []corev1.VolumeMount {
 	return []corev1.VolumeMount{
```
0sewa0, line 53: nitpick: not related to your change really, but this is the same as 
```
return []corev1.VolumeMount{
		{
			Name:      config.ProxySecretVolumeName,
			MountPath: config.ProxySecretMountPath,
		},
	}
```
the same way you did it in the oneagent daemonset

src/kubeobjects/labels.go
```diff
@@ -18,6 +18,7 @@ const (
 	ActiveGateComponentLabel = "activegate"
 	WebhookComponentLabel    = "webhook"
 	SyntheticComponentLabel  = "synthetic"
+	ProxyComponentLabel      = "proxy"
```
0sewa0, line 21: not needed

src/controllers/dynakube/oneagent/reconciler.go
```diff
@@ -35,22 +37,27 @@ func NewOneAgentReconciler( //nolint:revive // maximum number of return results
...
 	// that reads objects from the cache and writes to the apiserver
-	client    client.Client
-	apiReader client.Reader
-	scheme    *runtime.Scheme
-	clusterID string
+	client          client.Client
+	apiReader       client.Reader
+	scheme          *runtime.Scheme
+	clusterID       string
+	proxyReconciler controllers.Reconciler
```
0sewa0, line 60: you should do this somewhere in the [dynakube controller](https://github.com/Dynatrace/dynatrace-operator/blob/2328a307c407168f5b319c8bd278a92409bab305/src/controllers/dynakube/dynakube_controller.go#L248), because both the activegate and oneagent can need it, so it should be a more central location

0sewa0 (changes requested):

src/controllers/dynakube/oneagent/daemonset/volumes.go
```diff
@@ -75,6 +81,13 @@ func getCSIStorageMount() corev1.VolumeMount {
 	}
 }
 
+func getHttpProxyMount() corev1.VolumeMount {
```
andriisoldatenko, line 84:
```suggestion
func prepareHttpProxyMount() corev1.VolumeMount {

or

func buildHttpProxyMount() corev1.VolumeMount {

```

albertogdd: I followed the naming convention from the functions above. We can refactor all of them in a different PR if using `prepare` or `build` makes more sense than `get`.

src/controllers/dynakube/oneagent/reconciler.go
albertogdd: Great point, I didn't recalled about this :+1::+1::+1:

src/controllers/dynakube/oneagent/daemonset/volumes.go
albertogdd: As discussed: I will use `prepare` for the new function and leave the old package with the `get` naming. We will refactor afterwards

albertogdd:
PR is on hold until OneAgent mount path bug is solved.

PR #1931, opened 2023-07-10
Adapt K8s connection settings creation to newest schema version

annazaionchkovska:
# Description

After splitting k8s setting to Connection and Monitoring setting creation of k8s should be also adopted.
Description of split you can find at:  [Info](https://www.dynatrace.com/support/help/platform-modules/infrastructure-monitoring/container-platform-monitoring/kubernetes-monitoring/default-monitoring-settings)

## Approach 1 (removed after introducing approach 2)
Depending on schema "builtin:cloud.kubernetes" version we use  "1.0.27" or "3.0.0" version for creating configuration 
For all <"3.0.0" we use "1.0.27", for all which are after "3.0.0"  we use "3.0.0" 

so for "1.0.27" monitoring setting should be send together on connection creation
for "3.0.1" only connection settings are send

## Approach 2

- it uses the newest Kubernetes connection settings schema version, which is 3.0.0
- in case the request for the newest schema version fails (e.g. if the Operator connects to an older cluster version, which does not have the newest schema version yet) ---> fallback to the original request for schema version 1.0.27

## How can this be tested?

- Dy cluster should be up and running
- new k8s connection should be created with Automatic connection
- for "builtin:cloud.kubernetes"   version "3.0.0" msg 'These settings are overriding Environment settings' must NOT be shown on Monitoring settings

## Checklist

### DIFF
--- a/.github/actions/build-image/action.yaml
+++ b/.github/actions/build-image/action.yaml
@@ -16,7 +16,7 @@ runs:
     - name: Set up QEMU
       uses: docker/setup-qemu-action@2b82ce82d56a2a04d2637cd93a637ae1b359c0a7 # v2.2.0
     - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@16c0bc4a6e6ada2cfd8afd41d22d95379cf7c32a # v2.8.0
+      uses: docker/setup-buildx-action@4c0219f9ac95b02789c1075625400b2acbff50b1 # v2.9.1
     - name: Set up Golang
       uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 # v4.0.1
       with:
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,19 +1,21 @@
-# Description
+## Description
 
 Please include the following:
-- a meaningful PR title [guidelines](https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#use-imperative-mood-in-your-commit-message-subject)
-- a summary of the change
-- which issue is fixed (if there is one)
-- relevant motivation and context
+
+- A meaningful PR title [guidelines](https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#use-imperative-mood-in-your-commit-message-subject)
+- A summary of the change
+- Which issue is fixed (if there is one)
+- Relevant motivation and context
 
 ## How can this be tested?
+
 Please include some guiding steps on how to test this change during a review.
+
 - What environment is necessary for the change to be noticeable ?
 - What needs to be enabled/applied for the change to be noticeable ?
 
-
 ## Checklist
+
 - [ ] Unit tests have been updated/added
 - [ ] PR is labeled accordingly with a single label
 - [ ] I have read and understood the [contribution guidelines](https://github.com/Dynatrace/dynatrace-operator/blob/main/CONTRIBUTING.md)
-
--- a/.github/workflows/codeql-analysis.yaml
+++ b/.github/workflows/codeql-analysis.yaml
@@ -34,12 +34,12 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@004c5de30b6423267685b897a3d595e944f7fed5 # v2.20.2
+      uses: github/codeql-action/init@46ed16ded91731b2df79a2893d3aea8e9f03b5c4 # v2.20.3
       with:
         languages: ${{ matrix.language }}
 
     - name: Autobuild
-      uses: github/codeql-action/autobuild@004c5de30b6423267685b897a3d595e944f7fed5 # v2.20.2
+      uses: github/codeql-action/autobuild@46ed16ded91731b2df79a2893d3aea8e9f03b5c4 # v2.20.3
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@004c5de30b6423267685b897a3d595e944f7fed5 # v2.20.2
+      uses: github/codeql-action/analyze@46ed16ded91731b2df79a2893d3aea8e9f03b5c4 # v2.20.3
--- a/.github/workflows/openssf-scorecards.yaml
+++ b/.github/workflows/openssf-scorecards.yaml
@@ -56,6 +56,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@004c5de30b6423267685b897a3d595e944f7fed5 # v2.20.2
+        uses: github/codeql-action/upload-sarif@46ed16ded91731b2df79a2893d3aea8e9f03b5c4 # v2.20.3
         with:
           sarif_file: results.sarif
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # setup build image
-FROM golang:1.20.5@sha256:4b1fc02d16fca272e5e6e6adc98396219b43ef663a377eef4a97e881d364393f AS go-base
+FROM golang:1.20.6@sha256:8e5a0067e6b387263a01d06b91ef1a983f90e9638564f6e25392fd2695f7ab6c AS go-base
 RUN \
     --mount=type=cache,target=/var/cache/apt \
     apt-get update && apt-get install -y libbtrfs-dev libdevmapper-dev
--- a/assets/calico/activegate-policy.yaml
+++ b/assets/calico/activegate-policy.yaml
@@ -20,7 +20,11 @@ spec:
     - protocol: TCP
       port: 9999
     - protocol: TCP
+      port: 9998
+    - protocol: TCP
       port: 443
+    - protocol: TCP
+      port: 80
   egress:
   # Allow DNS lookup
   - to:
--- a/assets/samples/applicationMonitoring.yaml
+++ b/assets/samples/applicationMonitoring.yaml
@@ -160,7 +160,7 @@ spec:
     #
     # env: []
 
-    # Optional: Sets the activation group for ActiveGate instances
+    # Recommended: Sets the activation group for ActiveGate instances
     #
     # group: ""
 
--- a/assets/samples/classicFullStack.yaml
+++ b/assets/samples/classicFullStack.yaml
@@ -198,7 +198,7 @@ spec:
     #
     # env: []
 
-    # Optional: Sets the activation group for ActiveGate instances
+    # Recommended: Sets the activation group for ActiveGate instances
     #
     # group: ""
 
--- a/assets/samples/cloudNativeFullStack.yaml
+++ b/assets/samples/cloudNativeFullStack.yaml
@@ -192,7 +192,7 @@ spec:
     #
     # image: ""
 
-    # Optional: Sets the activation group for ActiveGate instances
+    # Recommended: Sets the activation group for ActiveGate instances
     #
     # group: ""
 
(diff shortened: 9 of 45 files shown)


src/dtclient/kubernetes_settings.go
```diff
@@ -6,21 +6,30 @@ import (
...
+
+type monitoringSettings struct {
+	CloudApplicationPipelineEnabled bool `json:"cloudApplicationPipelineEnabled"`
+	OpenMetricsPipelineEnabled      bool `json:"openMetricsPipelineEnabled"`
+	EventProcessingActive           bool `json:"eventProcessingActive"`
+	EventProcessingV2Active         bool `json:"eventProcessingV2Active"`
+	FilterEvents                    bool `json:"filterEvents"`
+}
+
+func (d *postKubernetesSettings) UnmarshalJSON(b []byte) error {
```
chrismuellner, line 31: does this need to be its own method?

src/dtclient/kubernetes_settings.go
```diff
@@ -67,33 +80,50 @@ type constraintViolations []struct {
...
-func (dtc *dynatraceClient) CreateOrUpdateKubernetesSetting(clusterLabel, kubeSystemUUID, scope string) (string, error) {
-	if kubeSystemUUID == "" {
-		return "", errors.New("no kube-system namespace UUID given")
+const (
+	schemaId                                    = "builtin:cloud.kubernetes"
+	defaultSchemaVersion                        = "1.0.27"
+	hierarchicalMonitoringSettingsSchemaVersion = "3.0.0"
+)
+
+func (dtc *dynatraceClient) GetSchemasVersion(schemaId string) (string, error) {
```
chrismuellner, line 89: I don't know if things changed, but didn't we want to skip the schema version check and just attempt to first post schema v3 and if it fails fall back to v1?

this would mean we have the same number of requests if the v3 schema is missing and only a single request if it's present/available.

chrismuellner (changes requested):

src/dtclient/kubernetes_settings.go
annazaionchkovska: Thanks for this point, I was able to remove it after renaming of field  monitoringSettings->MonitoringSettings
Actually the problem was in tests on 

var parsedBody []postKubernetesSettingsBody
err = json.Unmarshal(body, &parsedBody)

as monitoringSettings should be parsed as part of postKubernetesSettings, but I got an error so U added additional UnmarshalJSON. But after renaming monitoringSettings->MonitoringSettings it did work as exoected.

src/dtclient/kubernetes_settings.go
0sewa0: is there a nice a reliable way to notice that its a fail because the schema v3 is not present, but otherwise everything is OK? 
I would prefer to avoid handling weird Status codes, and also always swallowing the error for the creation using v3 is not a nice thing aswell.

annazaionchkovska: I have change code according request of Christian and description in original issue, although to my opinion previous solution with version check was better. Please let me know in case we need to change it back.

src/dtclient/kubernetes_settings.go
```diff
@@ -67,33 +72,13 @@ type constraintViolations []struct {
...
-		},
-	}
-
-	if scope != "" {
-		body[0].Scope = scope
-	}
+const (
+	schemaId                                    = "builtin:cloud.kubernetes"
+	defaultSchemaVersion                        = "1.0.27"
+	hierarchicalMonitoringSettingsSchemaVersion = "3.0.0"
```
chrismuellner, line 78: is this now the `defaultSchemaVersion`?

src/dtclient/kubernetes_settings.go
```diff
@@ -133,6 +118,57 @@ func (dtc *dynatraceClient) CreateOrUpdateKubernetesSetting(clusterLabel, kubeSy
 	return resDataJson[0].ObjectId, nil
 }
 
+func (dtc *dynatraceClient) getKubernetesSettingBody(clusterLabel, kubeSystemUUID, scope string, schemaId string, isK8sHierarchicalMonitoringSettings bool) []postKubernetesSettingsBody {
```
chrismuellner, line 121: does not need to be a method

src/dtclient/kubernetes_settings.go
```diff
@@ -133,6 +118,57 @@ func (dtc *dynatraceClient) CreateOrUpdateKubernetesSetting(clusterLabel, kubeSy
...
+			SchemaVersion: currentSchemaVersion,
+			Value: postKubernetesSettings{
+				Enabled:          true,
+				Label:            clusterLabel,
+				ClusterIdEnabled: true,
+				ClusterId:        kubeSystemUUID,
+			},
+		},
+	}
+	if scope != "" {
```
chrismuellner, line 138: can we always set this for the `body` as the default value is also an empty string?

src/dtclient/kubernetes_settings_test.go
```diff
@@ -201,9 +204,29 @@ func TestDynatraceClient_GetSettingsForMonitoredEntities(t *testing.T) {
...
+		skipCert := SkipCertificateValidation(true)
+		dtc, err := NewClient(dynatraceServer.URL, apiToken, paasToken, skipCert)
+		require.NoError(t, err)
+		require.NotNil(t, dtc)
+
+		// act
+		actual, err := dtc.(*dynatraceClient).CreateOrUpdateKubernetesSetting(testName, testUID, testScope)
+
+		// assert
+		assert.NotNil(t, actual)
```
chrismuellner, line 221: use `require` to quick fail tests and not pollute output with follow up errors (see above, makes most sense for `NotNil`/`NoError` checks)

src/dtclient/kubernetes_settings.go
annazaionchkovska: so we creating new setting with 2 schemas: "1.0.27" and  "3.0.0"
in case correct schema version > "3.0.0" but we try to use  "1.0.27" it still work and would not bring to error, so I suppose it is ok to have "1.0.27" as defaultSchemaVersion

src/dtclient/kubernetes_settings.go
```diff
@@ -133,6 +118,57 @@ func (dtc *dynatraceClient) CreateOrUpdateKubernetesSetting(clusterLabel, kubeSy
...
+			OpenMetricsPipelineEnabled:      false,
+			EventProcessingActive:           false,
+			FilterEvents:                    false,
+			EventProcessingV2Active:         false,
+		}
+
+		body[0].Value.MonitoringSettings = &ms
+	}
+	return body
+}
```
0sewa0, line 153: I would split it this up a bit to be nicer:

We should only work with a list in the end so we don't do stuff like `body[0]...`

So would have a function for creating the base settings body: 
```
func createBaseKubernetesSettings(... ) postKubernetesSettingsBody {
	base := postKubernetesSettingsBody{
			SchemaId:      schemaId,
			Value: postKubernetesSettings{
				Enabled:          true,
				Label:            clusterLabel,
				ClusterIdEnabled: true,
				ClusterId:        kubeSystemUUID,
			},
		}
	if scope != "" {
		base.Scope = scope
	}
}
```

then have separate functions for the new Schema and the old Schema:
```
func createV3KubernetesSettingsBody(...) []postKubernetesSettingsBody {
	settings := createBaseKubernetesSettings(...)
	settings.Schema = hierarchicalMonitoringSettingsSchemaVersion
	ms := MonitoringSettings{
			CloudApplicationPipelineEnabled: true,
			OpenMetricsPipelineEnabled:      false,
			EventProcessingActive:           false,
			FilterEvents:                    false,
			EventProcessingV2Active:         false,
	}
	settings.Value.MonitoringSettings = &ms
	return []postKubernetesSettingsBody{settings}
}

func createV1KubernetesSettingsBody(...) []postKubernetesSettingsBody {
	settings := createBaseKubernetesSettings(...)
	settings.Schema = defaultSchemaVersion
	return []postKubernetesSettingsBody{settings}
}
```
by doing this you remove the weird ifs + bool logic, so you can make the code more readable and clear what it does

and the following reads a bit nicer:
```
func (dtc *dynatraceClient) CreateOrUpdateKubernetesSetting(clusterLabel, kubeSystemUUID, scope string) (string, error) {
	if kubeSystemUUID == "" {
		return "", errors.New("no kube-system namespace UUID given")
	}
	body := createV3KubernetesSettingsBody(clusterLabel, kubeSystemUUID, scope, schemaId)
	objectId, err := dtc.performCreateOrUpdateKubernetesSetting(body)
	if err != nil {
		if strings.Contains(err.Error(), strconv.Itoa(http.StatusNotFound)) {
			body = createV1KubernetesSettingsBody(clusterLabel, kubeSystemUUID, scope, schemaId)
			return dtc.performCreateOrUpdateKubernetesSetting(body)
		} else {
			return "", err
		}
	}
	return objectId, nil
}
```

src/dtclient/kubernetes_settings.go
annazaionchkovska: for me it is question of taste, additionally I can test response body separately as json 
src/dtclient/kubernetes_settings_test.go:303
please let me know if we should really remove this method, thanks

src/dtclient/kubernetes_settings.go
chrismuellner: we use v3 as default now and only fallback to v1 later though. We can also just rename the constant from `defaultSchemaVersion` to `schemaVersionV1`

src/dtclient/kubernetes_settings.go
chrismuellner:
```suggestion
func getKubernetesSettingBody(clusterLabel, kubeSystemUUID, scope string, schemaId string, isK8sHierarchicalMonitoringSettings bool) []postKubernetesSettingsBody {
```
I mean to change it to a function, not completely remove it.

src/dtclient/kubernetes_settings_test.go
annazaionchkovska: thanks for input, done

src/dtclient/kubernetes_settings.go
annazaionchkovska: thanks, moved it to code

src/dtclient/kubernetes_settings.go
annazaionchkovska: done

src/dtclient/kubernetes_settings.go
```diff
@@ -133,6 +118,57 @@ func (dtc *dynatraceClient) CreateOrUpdateKubernetesSetting(clusterLabel, kubeSy
...
+
+func (dtc *dynatraceClient) CreateOrUpdateKubernetesSetting(clusterLabel, kubeSystemUUID, scope string) (string, error) {
+	if kubeSystemUUID == "" {
+		return "", errors.New("no kube-system namespace UUID given")
+	}
+	body := dtc.getKubernetesSettingBody(clusterLabel, kubeSystemUUID, scope, schemaId, true)
+	objectId, err := dtc.performCreateOrUpdateKubernetesSetting(body)
+	if err != nil {
+		if strings.Contains(err.Error(), strconv.Itoa(http.StatusNotFound)) {
```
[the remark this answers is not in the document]
0sewa0: `k8serrors.IsNotFound` is meant to be used when trying to get a kubernetes object, and not when talking to the communicating with the dynatrace API.
(while both kubernetes and dynatrace API return the StatusCode 404, `k8serrors.IsNotFound` does some extra parsing that is kubernetes specific so I would avoid using it here)

TLDR:
let's leave it as it is now, its ain't perfect but using `k8serrors.IsNotFound` is not way here

chrismuellner (approved):

PR #1934, opened 2023-07-11
Refactor PR markdown template following Markdownlint rules

albertogdd:
##  Description

Refactor PR markdown template and comply with the [Markdownlint project](https://github.com/DavidAnson/markdownlint/tree/main) rules.

It's available as an extension on VSCode marketplace (https://marketplace.visualstudio.com/items\?itemName\=DavidAnson.vscode-markdownlint)

## How can this be tested?
No tests are needed.

## Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,19 +1,21 @@
-# Description
+## Description
 
 Please include the following:
-- a meaningful PR title [guidelines](https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#use-imperative-mood-in-your-commit-message-subject)
-- a summary of the change
-- which issue is fixed (if there is one)
-- relevant motivation and context
+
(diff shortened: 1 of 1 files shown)


andriisoldatenko:
Thanks! It's a really good idea, 
is it possible to add it as GH action? so we can automatically check it across all markdown files?

albertogdd:
> Thanks! It's a really good idea, is it possible to add it as GH action? so we can automatically check it across all markdown files?



> Thanks! It's a really good idea, is it possible to add it as GH action? so we can automatically check it across all markdown files?

@andriisoldatenko I will write a user story about it! It would be possible to have a GH action (I just tried it [with this PR](https://github.com/Dynatrace/dynatrace-operator/pull/1935)) but it will require some time to update the existing markdown files to comply with the rules etc.

IMO we can merge this PR, and I'll create the follow-up user story.

andriisoldatenko (approved):

PR #1949, opened 2023-07-14
Change SupportArchive file format from tar to zip

wepudt:
## Description
replace the compression in the supportARchive with ZIP as the tarball may lead to an OOM kill.Due to the nature of tar the whole file has to be loaded into memory while using zip we can stream the log files.

## How can this be tested?

- e2e tests: `make test/e2e/supportarchive`
- create a support archive with loadtest: `kubectl exec -n dynatrace deployment/dynatrace-operator -- dynatrace-operator support-archive --loadsim-file-size 1000 --loadsim-files 1 --stdout > operator-support-archive.zip`

## Checklist


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/Dynatrace/dynatrace-operator
 go 1.20
 
 require (
+	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137
```
andriisoldatenko, line 6: I'm a bit concerned about introducing new package for 1 single usage.
Also after checikng package, it doesn't have proper tests. 
Do it make sense to remove it and a little bit redo it on our own? 

Go proverb “a little copying is better than a little dependency” is to be careful when bringing new dependencies to our programs.

andriisoldatenko: cc @0sewa0

andriisoldatenko (changes requested):
would be nice to have benchmark tests IMO:
here is an example:
```
package main

import (
    "testing"
)

var num = 1000

func BenchmarkPrimeNumbers(b *testing.B) {
    for i := 0; i < b.N; i++ {
        primeNumbers(num)
    }
}
```

go.mod
wepudt: true- I just used it as it was already there and I wanted to change as little as possible...

wepudt: read through the usage again and replaced it with 1024 * 1024- not that much of a change and as it is in test code I think worth it to remove a dependency

andriisoldatenko: @wepudt thanks!

src/cmd/support_archive/tarball_test.go
```diff
@@ -1,13 +1,11 @@
 package support_archive
```
andriisoldatenko, line 1: since you removing tarball.go, does it make sense to rename this file to archive_test ?

test/scenarios/support_archive/install.go
```diff
@@ -120,16 +114,16 @@ func executeSupportArchiveCommand(ctx context.Context, t *testing.T, environment
 }
 
 func collectRequiredFiles(t *testing.T, ctx context.Context, resources *resources.Resources, testDynakube dynatracev1beta1.DynaKube) []string {
-	namespace := testDynakube.Namespace
+	n := testDynakube.Namespace
```
andriisoldatenko, line 117:
```suggestion
	ns := testDynakube.Namespace
```
IMO is more readable name.

andriisoldatenko: I think similar rename you did with `context` -> `ctx`.

src/cmd/support_archive/tarball_test.go
wepudt: sure- just wanted to keep some memories :D

src/cmd/support_archive/resources_test.go
```diff
@@ -125,19 +120,17 @@ func TestManifestCollector_NoManifestsAvailable(t *testing.T) {
...
 	err := newK8sObjectCollector(ctx, log, supportArchive, testOperatorNamespace, defaultOperatorAppName, clt).Do()
 	require.NoError(t, err)
-
-	tarReader := tar.NewReader(&tarBuffer)
-	_, err = tarReader.Next()
-	require.ErrorIs(t, err, io.EOF)
+	supportArchive.Close()
+	zipReader, err := zip.NewReader(bytes.NewReader(buffer.Bytes()), int64(buffer.Len()))
+	assert.NoError(t, err)
+	assert.Equal(t, 0, len(zipReader.File))
```
StefanHauth, line 133:
```suggestion
	assert.Len(t, zipReader.File, 0)
```

src/cmd/support_archive/resources_test.go
```diff
@@ -169,30 +162,20 @@ func TestManifestCollector_PartialCollectionOnMissingResources(t *testing.T) {
...
-	}
+	buffer := bytes.Buffer{}
+	supportArchive := newZipArchive(bufio.NewWriter(&buffer))
 
 	collector := newK8sObjectCollector(context, log, supportArchive, testOperatorNamespace, defaultOperatorAppName, clt)
 	require.NoError(t, collector.Do())
+	supportArchive.Close()
+	zipReader, err := zip.NewReader(bytes.NewReader(buffer.Bytes()), int64(buffer.Len()))
+	assert.NoError(t, err)
+	assert.Equal(t, 3, len(zipReader.File))
```
StefanHauth, line 173:
```suggestion
	require.Equal(t, 3, len(zipReader.File))
```

StefanHauth (changes requested):

src/cmd/support_archive/loadsim.go
```diff
@@ -0,0 +1,88 @@
...
+		fileName := buildLoadsimFileName(i)
+
+		lg := loadGenerator{
+			fileSize: collector.fileSize,
+		}
+
+		err := collector.supportArchive.addFile(fileName, &lg)
+
+		if err != nil {
+			logErrorf(collector.log, err, "error writing simulated load to tarball")
```
waodim, line 61:
```suggestion
			logErrorf(collector.log, err, "error writing simulated load to zip")
```

src/cmd/support_archive/archiver.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (z zipArchive) addFile(fileName string, reader io.Reader) error {
+	w, err := z.writer.Create(fileName)
+	if err != nil {
+		return errors.WithMessagef(err, "could not write header for file '%s'", fileName)
+	}
+
+	_, err = io.Copy(w, reader)
+	if err != nil {
+		return errors.WithMessagef(err, "could not copy the file '%s' data to the zip archive", fileName)
```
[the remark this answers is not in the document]
wepudt: isn't that just a rewording?

test/scenarios/support_archive/install.go
```diff
@@ -75,30 +74,25 @@ func testSupportArchiveCommand(testDynakube dynatracev1beta1.DynaKube) features.
...
-
 		assert.Emptyf(t, requiredFiles, "Support archive does not contain all expected files.")
 		logMissingFiles(t, requiredFiles)
 		return ctx
 	}
 }
 
 func executeSupportArchiveCommand(ctx context.Context, t *testing.T, environmentConfig *envconf.Config, cmdLineArguments, namespace string) *pod.ExecutionResult { //nolint:revive
-	resources := environmentConfig.Client().Resources()
+	r := environmentConfig.Client().Resources()
```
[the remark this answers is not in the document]
0sewa0: you could call it `envResources` if you don't want to shadow imports

src/cmd/support_archive/archiver.go
```diff
@@ -0,0 +1,84 @@
...
+	"strings"
+	"time"
+
+	"github.com/klauspost/compress/zip"
+	"github.com/pkg/errors"
+)
+
+const zipArchiveFileName = "%s/operator-support-archive-%s.zip"
+
+type archiver interface {
```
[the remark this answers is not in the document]
wepudt: I don't want to do this as
* it is not support archive specific
* not named to go standards
* this is just a package internal interface

src/cmd/support_archive/archiver.go
```diff
@@ -0,0 +1,84 @@
...
+		archiveFile, err := createZipArchiveFile(targetDir)
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		return archiveFile, nil
+	}
+}
+
+func createZipArchiveFile(targetDir string) (*os.File, error) {
+	tarballFilePath := fmt.Sprintf(zipArchiveFileName, targetDir, time.Now().Format(time.RFC3339))
```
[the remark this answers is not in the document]
wepudt: removed all tar mentioning in an updated commit

src/cmd/support_archive/logs_test.go
```diff
@@ -268,20 +244,12 @@ func TestLogCollectorNoAbortOnError(t *testing.T) {
...
-
-	tarHeader, err = tarReader.Next()
-	require.NoError(t, err)
-	assert.Equal(t, "logs/pod2/container2.log", tarHeader.Name)
+	supportArchive.Close()
 
-	_, err = tarReader.Next()
-	require.ErrorIs(t, err, io.EOF)
+	zipReader, err := zip.NewReader(bytes.NewReader(buffer.Bytes()), int64(buffer.Len()))
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
wepudt: done in followup commit

test/scenarios/support_archive/install.go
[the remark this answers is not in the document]
wepudt: the problem I have with resources is, that I get a warning in my IDE. And normally I do not have warnings as they might point to real problems. Having to look though all warnings an tell which ones are bad and which not creates much too much mental load (for me).
If we want to have meaningful names use some that do not produce warnings (give it some more thought). I went with environmentResources in this case. Otherwise I would stick to r, ns, ctx or the other commonly used abbrevs.

src/cmd/support_archive/loadsim.go
```diff
@@ -0,0 +1,88 @@
...
+		pods:      pods,
+	}
+}
+
+func (collector loadSimCollector) Do() error {
+	if collector.fileCount <= 0 {
+		return nil
+	}
+
+	logInfof(collector.log, "Starting load simulation")
```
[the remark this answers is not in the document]
wepudt: I'm actually fine with it. It actually does create a loadtest so I would just keep it...

src/cmd/support_archive/loadsim.go
```diff
@@ -0,0 +1,88 @@
...
+	"io"
+
+	"github.com/go-logr/logr"
+	"golang.org/x/exp/rand"
+	clientgocorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+const loadSimCollectorName = "loadSimCollector"
+
+type loadSimCollector struct {
```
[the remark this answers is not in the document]
wepudt: I think it's OK- sums up the purpose so I'll keep it...

src/cmd/support_archive/loadsim.go
```diff
@@ -0,0 +1,88 @@
...
+	collectorCommon
+	context   context.Context
+	fileSize  int
+	fileCount int
+	pods      clientgocorev1.PodInterface
+}
+
+func newLoadSimCollector(context context.Context, log logr.Logger, supportArchive archiver, fileSize int, fileCount int, pods clientgocorev1.PodInterface) collector { //nolint:revive // argument-limit doesn't apply to constructors
```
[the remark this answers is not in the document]
wepudt: fixed

luhi-DT (dismissed):

luhi-DT (approved):

PR #1953, opened 2023-07-14
Make the time provider return the current timestamp by default

aorcholski:
## Description

TimeProvider returns the current timestamp by default. Method chaining (`SetNow`) allows this function to be used in initialization lists (temporary variable to call `Set`/`SetNow` is not needed). 

## How can this be tested?
Dynatrace Operator works properly - dynakube status...

## Checklist

### DIFF
--- a/src/api/v1beta1/properties_test.go
+++ b/src/api/v1beta1/properties_test.go
@@ -441,7 +441,7 @@ func TestDynaKube_ShallUpdateActiveGateConnectionInfo(t *testing.T) {
 		},
 	}
 
-	timeProvider := timeprovider.New()
+	timeProvider := timeprovider.New().SetToNow()
 	tests := map[string]struct {
 		lastRequestTimeDeltaMinutes int
 		updateExpected              bool
--- a/src/api/v1beta1/properties_test.go
+++ b/src/api/v1beta1/properties_test.go
@@ -441,7 +441,7 @@ func TestDynaKube_ShallUpdateActiveGateConnectionInfo(t *testing.T) {
 		},
 	}
 
-	timeProvider := timeprovider.New().SetToNow()
+	timeProvider := timeprovider.New().Freeze()
 	tests := map[string]struct {
 		lastRequestTimeDeltaMinutes int
 		updateExpected              bool
--- a/src/controllers/dynakube/connectioninfo/reconciler.go
+++ b/src/controllers/dynakube/connectioninfo/reconciler.go
@@ -33,7 +33,7 @@ func NewReconciler(ctx context.Context, clt client.Client, apiReader client.Read
 		dynakube:     dynakube,
 		scheme:       scheme,
 		dtc:          dtc,
-		timeProvider: timeprovider.New(),
+		timeProvider: timeprovider.New().SetToNow(),
 	}
 }
 
--- a/src/controllers/dynakube/connectioninfo/reconciler.go
+++ b/src/controllers/dynakube/connectioninfo/reconciler.go
@@ -33,7 +33,7 @@ func NewReconciler(ctx context.Context, clt client.Client, apiReader client.Read
 		dynakube:     dynakube,
 		scheme:       scheme,
 		dtc:          dtc,
-		timeProvider: timeprovider.New().SetToNow(),
+		timeProvider: timeprovider.New(),
 	}
 }
 
--- a/src/controllers/dynakube/dynakube_controller.go
+++ b/src/controllers/dynakube/dynakube_controller.go
@@ -238,7 +238,7 @@ func (controller *Controller) reconcileDynaKube(ctx context.Context, dynakube *d
 		dynatraceClient,
 		controller.fs,
 		controller.versionProvider,
-		timeprovider.New(),
+		timeprovider.New().SetToNow(),
 	)
 	err = versionReconciler.Reconcile(ctx)
 	if err != nil {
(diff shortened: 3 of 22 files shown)


src/timeprovider/provider_test.go
```diff
@@ -58,3 +58,20 @@ func TestTimeProvider(t *testing.T) {
 		assert.True(t, provider.IsOutdated(nil, testTimeout))
 	})
 }
+
+func TestTimeProvider(t *testing.T) {
+	t.Run("timestamp is not set", func(t *testing.T) {
+		provider := New()
+		earlier := provider.Now()
+		later := provider.Now()
+		assert.False(t, earlier.Equal(later))
```
StefanHauth, line 67:
```suggestion
		assert.NotEqual(t, earlier, later)
```

src/timeprovider/provider_test.go
```diff
@@ -58,3 +58,20 @@ func TestTimeProvider(t *testing.T) {
...
+		assert.True(t, earlier.Before(later))
+	})
+
+	t.Run("timestamp is set", func(t *testing.T) {
+		provider := New().SetToNow()
+		earlier := provider.Now()
+		later := provider.Now()
+		assert.True(t, earlier.Equal(later))
```
StefanHauth, line 75:
```suggestion
		assert.Equal(t, earlier, later)
```

src/api/v1beta1/properties_test.go
```diff
@@ -441,7 +441,7 @@ func TestDynaKube_ShallUpdateActiveGateConnectionInfo(t *testing.T) {
 		},
 	}
 
-	timeProvider := timeprovider.New()
+	timeProvider := timeprovider.New().SetToNow()
```
StefanHauth, line 444: I'm somehow not entirely happy with the naming. OTOH I do not have an superior idea. Maybe
```suggestion
	timeProvider := timeprovider.New().Freeze()
	timeProvider := timeprovider.New().Lock()
```
IDK...

src/controllers/dynakube/connectioninfo/reconciler.go
```diff
@@ -33,7 +33,7 @@ func NewReconciler(ctx context.Context, clt client.Client, apiReader client.Read
 		dynakube:     dynakube,
 		scheme:       scheme,
 		dtc:          dtc,
-		timeProvider: timeprovider.New(),
+		timeProvider: timeprovider.New().SetToNow(),
```
StefanHauth, line 36: I don't think you need to "lock-in" the timestamp here. The time provider is used to check the current time against the timestamp of the last connection info request. An accurate "now" should be used.

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -238,7 +238,7 @@ func (controller *Controller) reconcileDynaKube(ctx context.Context, dynakube *d
 		dynatraceClient,
 		controller.fs,
 		controller.versionProvider,
-		timeprovider.New(),
+		timeprovider.New().SetToNow(),
```
StefanHauth, line 241: Same here. All instances were the timeprovider is used, IMHO benefits of an more accurate now.

src/controllers/dynakube/dynatraceclient/builder.go
```diff
@@ -119,7 +119,7 @@ func (dynatraceClientBuilder builder) BuildWithTokenVerification(dynaKubeStatus
 func (dynatraceClientBuilder builder) verifyTokenScopes(dynatraceClient dtclient.Client, dynaKubeStatus *dynatracev1beta1.DynaKubeStatus) error {
 	var err error
 
-	if dynatraceClientBuilder.dynakube.IsTokenScopeVerificationAllowed(timeprovider.New()) {
+	if dynatraceClientBuilder.dynakube.IsTokenScopeVerificationAllowed(timeprovider.New().SetToNow()) {
```
StefanHauth, line 122: And also here IMHO no timestamp freeze is needed.

StefanHauth (changes requested):

src/timeprovider/provider_test.go
aorcholski: see comment above

src/controllers/dynakube/connectioninfo/reconciler.go
aorcholski: SetTo removed

src/controllers/dynakube/dynakube_controller.go
aorcholski: `versionReconciler` calls 4 `versionStatusUpdaters`. IMHO it's better to call all of them or none of them (`timeProvider` is `Set`)  than call 1 updater out of 4 this time and 3 updaters out of 4 next time because the timeout expires at different points in time during the execution of the [loop](https://github.com/Dynatrace/dynatrace-operator/blob/main/src/controllers/dynakube/version/reconciler.go#L90)

src/controllers/dynakube/dynatraceclient/builder.go
aorcholski: `SetTo` removed

src/api/v1beta1/properties_test.go
aorcholski: `SetToNow` renamed to `Freeze`

StefanHauth (approved):
LGTM

PR #1954, opened 2023-07-17
Treat zip and image codemodules as the same in csi server

0sewa0:
## Description

Currently there are two places where the csi-driver puts the codemodules on the filesystem
- `/data/<tenant>/bin/...` : Here are the codemodules that were unpacked from a **customized** zip from the Dynatrace Environment. 
- `/data/codemodules/...`: Here are the codemodules that were unpacked from a **immutable** image from a Image-Registry

What this PR changes:
- Now every codemodules is put into `/data/codemodules/...`
- A new init container was added to the csi-driver that does the following:
    - "Moves" codemodules from `/data/<tenant>/bin/...` to `/data/codemodules/...` using symlinks.
    - Does the same logic that was previously done before the startup of the csi-server ("correctness check", that cleans up the local database a bit before startup. )

## How can this be tested?

Scenarios to test:
- **Fresh cluster**, use csi-driver for any mode that uses ZIP codemodules (example: default cloudNativeFullstack)
   - To test if the download works, puts it into the new/correct place
   - Install and inject sample apps and see if they get deep monitored
- Cluster with previous version of operator that has a dynakube that uses the csi-driver with ZIP codemodules (example: default cloudNativeFullstack), **upgrade to "this" version**
   - To check that the "move" from the old location to the new location works.
   - Make sure no new download is triggered in the provisioner. 
   - Check that newly injected apps also get monitored.
- Check that default `applicationMonitoring` (so no csi-driver) still works.
   - Minor things were changed. Just to be sure. 

## Checklist


src/cmd/csi/init/builder.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/pkg/errors"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"golang.org/x/sys/unix"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+const use = "csi-init"
+
+var (
```
andriisoldatenko, line 17: just nit:
```
var nodeId, endpoint string
```
0 value for string also "".

src/cmd/csi/init/builder.go
```diff
@@ -0,0 +1,92 @@
...
+		Use:  use,
+		RunE: builder.buildRun(),
+	}
+
+	return cmd
+}
+
+func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		unix.Umask(0000)
```
andriisoldatenko, line 52: could you please use named constant here?

src/controllers/csi/gc/images.go
```diff
@@ -42,47 +42,40 @@ func (gc *CSIGarbageCollector) getSharedImageDirs() ([]os.FileInfo, error) {
...
-		if !usedImageDigests[imageDigest] {
-			toDelete = append(toDelete, gc.path.AgentSharedBinaryDirForImage(imageDigest))
-		}
-	}
-	return toDelete, nil
-}
-
-func (gc *CSIGarbageCollector) getUsedImageDigests(ctx context.Context) (map[string]bool, error) {
-	usedImageDigests, err := gc.db.GetUsedImageDigests(ctx)
+	setAgentImages, err := gc.db.GetUsedImageDigests(ctx)
```
andriisoldatenko, line 52: nit: didn't get why variable has `set` prefix. Could you please rename it?

src/standalone/run_test.go
```diff
@@ -118,23 +118,21 @@ func TestSetHostTenant(t *testing.T) {
...
-			On("UpdateProcessModuleConfig", config.AgentBinDirMount, &testProcessModuleConfig).
-			Return(nil)
 		runner.installer.(*installer.Mock).
 			On("InstallAgent", config.AgentBinDirMount).
 			Return(true, nil)
 
 		err := runner.installOneAgent()
 
-		require.NoError(t, err)
+		require.NoError(t, err) //this
```
andriisoldatenko, line 133:
```suggestion
		require.NoError(t, err)
```

andriisoldatenko (approved):
code LGTM, tested on my cluster also practically LGTM

PR #1960, opened 2023-07-19
Make seccomp field configurable for components

0sewa0:
## Description

Makes the seccomp field in the security context of our components configurable via helm

## How can this be tested?

Play with the seccomp fields in the values.yaml. 

## Checklist

### DIFF
--- a/config/helm/chart/default/values.yaml
+++ b/config/helm/chart/default/values.yaml
@@ -95,7 +95,7 @@ csidriver:
       operator: Exists
   labels: []
   annotations: []
-  csi_init:
+  csiInit:
     securityContext:
       runAsUser: 0
       privileged: false
--- a/config/helm/chart/default/values.yaml
+++ b/config/helm/chart/default/values.yaml
@@ -34,6 +34,15 @@ operator:
   annotations: []
   apparmor: false
   securityContext:
+    privileged: false
+    allowPrivilegeEscalation: false
+    readOnlyRootFilesystem: true
+    runAsNonRoot: true
+    runAsUser: 1001
+    runAsGroup: 1001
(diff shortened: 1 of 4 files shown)


config/helm/chart/default/values.yaml
```diff
@@ -71,7 +77,14 @@ csidriver:
       operator: Exists
   labels: []
   annotations: []
+  csi_init:
```
chrismuellner, line 80:
```suggestion
  csiInit:
```
let's stick with camelCase like in other places of the `values.yaml`, maybe it even makes sense to change the init container name

config/helm/chart/default/values.yaml
```diff
@@ -47,6 +50,9 @@ webhook:
   labels: []
   annotations: []
   apparmor: false
+  securityContext:
```
chrismuellner, line 53: I think we should set the whole security context, instead of just the seccomp profile - similar to `resources.requests/limits`

chrismuellner (changes requested):

config/helm/chart/default/values.yaml
0sewa0: the only problem with that is you can really easily mess it up for the csi-driver

not really for the rest

chrismuellner: You can already set the `resources` field to a value that will now allow the operator to work correctly. The only problem I see is that it's a bit different for the `securityContext`, since you may only want to change a single value, but would have to define all other values as well + figure out what the default values even are...

waodim (approved):
LGTM

chrismuellner (approved):

PR #1962, opened 2023-07-19
Add implementation of Keychain interface able to read docker config file from any path

aorcholski:
## Description

Single config file (`$HOME/.docker/config.json`) may be a bottleneck in case of multiple reconcilers using the go-container library in a multi-threaded way.  Added `DockerKeychain` is able to read docker config file from any path. 

## How can this be tested?

DockerKeychain will be used by `GetImageVersion` and  `installAgentFromImage`.

## Checklist


0sewa0 (approved):

src/containerregistry/docker_keychain.go
```diff
@@ -0,0 +1,82 @@
...
+func NewDockerKeychain(dockerconfigFile string, filesystem afero.Fs) authn.Keychain {
+	return &dockerKeychain{
+		dockerConfigFile: dockerconfigFile,
+		filesystem:       filesystem,
+	}
+}
+
+// Resolve implements Keychain interface by interpreting the docker config file.
+// It is based on the 'defaultKeychain' type from the go-gontainerregistry library
+// https://github.com/google/go-containerregistry/blob/main/pkg/authn/keychain.go
```
andriisoldatenko, line 29: just nit: for future readers it's better to replace main to current commit hash, otherwise if they change structure this link won't work.
```suggestion
// https://github.com/google/go-containerregistry/blob/27a6ad6/pkg/authn/keychain.go
```

src/containerregistry/docker_keychain.go
```diff
@@ -0,0 +1,82 @@
+package containerregistry
```
andriisoldatenko, line 1: May be I missed something, but it's not clear to me why this package has the name `containerregistry` ?

For example:
>Another convention is that the package name is the base name of its source directory; the package in src/encoding/base64 is imported as "encoding/base64" but has name base64, not encoding_base64 and not encodingBase64.

https://go.dev/doc/effective_go#package-names

cc @0sewa0

0sewa0: it package could be called `docker` and the file could be called `keychain.go`

docker is basically the same thing as `container registry` in our context, but shorter

andriisoldatenko: Correct, but also docker is not only a container registry it's also the name of the company, client, server, etc. :) 
I'm ok with your suggestion, at least it fits with the type inside keychain.go

aorcholski: `containerregistry` because the package defines a component specifically to be used with the `go-containerregistry` library

src/containerregistry/docker_keychain.go
```diff
@@ -0,0 +1,82 @@
...
+		dockerConfigFile: dockerconfigFile,
+		filesystem:       filesystem,
+	}
+}
+
+// Resolve implements Keychain interface by interpreting the docker config file.
+// It is based on the 'defaultKeychain' type from the go-gontainerregistry library
+// https://github.com/google/go-containerregistry/blob/27a6ad6/pkg/authn/keychain.go
+// dockerKeychain implementation can read a docker config file of any name and from any directory.
+func (dk *dockerKeychain) Resolve(target authn.Resource) (authn.Authenticator, error) {
```
0sewa0, line 31: `dk` is still sometimes used as a short name for the `dynakube` so I would avoid using it here, might cause confusion 😅 

you could used `keychain` or `key` if you want to be short

andriisoldatenko:
test

.golangci.yml
```diff
@@ -127,6 +127,8 @@ linters-settings:
           - "github.com/evanphx/json-patch"
           - "github.com/go-logr/logr"
           - "github.com/stretchr/testify"
+          - "github.com/google/go-containerregistry"
```
andriisoldatenko, line 130: test

src/containerregistry/docker_keychain.go
aorcholski: package name changed to more meaningful `dockerkeychain`

andriisoldatenko (approved):

src/containerregistry/docker_keychain.go
aorcholski: fixed

0sewa0 (approved):

PR #1965, opened 2023-07-21
Deprecate feature flags as they are switched to enabled by default

waodim:
## Description

With this the following feature flags will be marked as deprecated in the validation webhook:

- ActiveGate Auth Token
- ActiveGate ReadOnly FS
- ActiveGate Raw Image
- ActiveGate Updates
- ActiveGate Automatic K8S Monitoring
- Hosts Request
- OneAgent ReadOnly FS
- Webhook Reinvocation Policy
- Metadata Enrichment
- Automatic Injection
- Label Version Detection
- Injection Failure Policy

Some of those features were already enabled by default, some are now switched to enabled by default.

## How can this be tested?

There are unit tests added, but otherwise deploy a dynakube with no feature flags set and check whether those features that are listed above are enabled.

## Checklist


src/webhook/validation/deprecation.go
```diff
@@ -56,6 +104,19 @@ func warnIfDeprecatedIsUsed(dynakube *dynatracev1beta1.DynaKube, newAnnotation s
...
+
+	return ""
+}
+
 func deprecatedAnnotationWarning(newAnnotation string, deprecatedAnnotation string) string {
 	return fmt.Sprintf("annotation '%s' is deprecated, use '%s' instead", deprecatedAnnotation, newAnnotation)
 }
+
+func deprecationWarning(annotation string) string {
+	return fmt.Sprintf("feature flag '%s' is deprecated and will be removed in the future as the feature will be enabled by default", annotation)
```
chrismuellner, line 121: this message would also appear if the ff is used and already enabled, leading to a confusing warning.

I think it's enough to just warn that the ff will be removed in the future and we'll communicate the change in default value in the release notes

src/webhook/validation/deprecation.go
```diff
@@ -47,6 +47,54 @@ func deprecatedFeatureFlagDisableMetadataEnrichment(_ *dynakubeValidator, dynaku
 	return warnIfDeprecatedIsUsed(dynakube, dynatracev1beta1.AnnotationFeatureMetadataEnrichment, dynatracev1beta1.AnnotationFeatureDisableMetadataEnrichment)
 }
 
+func deprecatedFeatureFlagActiveGateReadOnlyFilesystem(_ *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
+	return warnIfDeprecatedFFisUsed(dynakube, dynatracev1beta1.AnnotationFeatureActiveGateReadOnlyFilesystem)
```
chrismuellner, line 51: could we move this into a single validator that just checks if any of the deprecated feature flags is used and returns a warning with all of them? assuming message max length is allows that

chrismuellner (changes requested):

src/webhook/validation/deprecation.go
```diff
@@ -56,6 +104,19 @@ func warnIfDeprecatedIsUsed(dynakube *dynatracev1beta1.DynaKube, newAnnotation s
 	return ""
 }
 
+func warnIfDeprecatedFFisUsed(dynakube *dynatracev1beta1.DynaKube, annotation string) string {
+	_, hasDeprecatedFlag := dynakube.Annotations[annotation]
+	if hasDeprecatedFlag {
+		return deprecationWarning(annotation)
+	}
```
andriisoldatenko, line 111:
```suggestion
	if _, ok := dynakube.Annotations[annotation]; ok {
	         return deprecationWarning(annotation)
	}
```
you can simplify it

src/webhook/validation/deprecation.go
andriisoldatenko: ☝️ agree with @chrismuellner I would prefer to refactor it using a single validator instead of lots of similar functions.

src/webhook/validation/deprecation.go
```diff
@@ -47,6 +64,14 @@ func deprecatedFeatureFlagDisableMetadataEnrichment(_ *dynakubeValidator, dynaku
 	return warnIfDeprecatedIsUsed(dynakube, dynatracev1beta1.AnnotationFeatureMetadataEnrichment, dynatracev1beta1.AnnotationFeatureDisableMetadataEnrichment)
 }
 
+func deprecatedFeatureFlag(_ *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
+	warnings := ""
+	for _, ff := range getDeprecatedFeatureFlags() {
+		warnings += warnIfDeprecatedFFisUsed(dynakube, ff)
```
chrismuellner, line 70: did you make sure this shows up correctly, if multiple feature flags are deprecated?

Individual warning messages can't be above 256 characters until they are truncated, see [kubernetes documentation on webhook response](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#response). This could be optimized a bit, by not writing the whole warning message every time, but only the actual feature flag that changed: "Some feature flags are deprecated and will be removed in the future: feature-flag1, feature-flag2, ...". If the truncation is a problem, we could change the message to "feature-flag1, feature-flag2 + 3 more" so it always fits into the message.

waodim: I did it now in a way that follows your suggestion as I think you are right regarding optimization. Please have a look and give me feedback whether you like this or not.

waodim: BTW: it worked before but the warning messages were bloated if you used various deprecated feature flags, which is why I tend to your proposed way.

waodim: FYI: with the approach I introduced in commit https://github.com/Dynatrace/dynatrace-operator/pull/1965/commits/f3475d56283ffb89cc6286b6c3a8c5d48d5e3534 there is no warning at all unfortunately. Working on it

waodim: With [29279eb](https://github.com/Dynatrace/dynatrace-operator/pull/1965/commits/29279ebd2314cbe33deb61f620f7cde02d863f56) I managed to output the warning message and there seems to be no problem with truncation: 
Screenshot from 2023-07-24 17-09-59

src/webhook/validation/deprecation.go
```diff
@@ -65,11 +65,22 @@ func deprecatedFeatureFlagDisableMetadataEnrichment(_ *dynakubeValidator, dynaku
 }
 
 func deprecatedFeatureFlag(_ *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
-	warnings := ""
+	warningMessage := ""
 	for _, ff := range getDeprecatedFeatureFlags() {
-		warnings += checkDeprecatedFeatureFlag(dynakube, ff)
+		if checkDeprecatedFeatureFlag(dynakube, ff) {
+			warningMessage += fmt.Sprintf("'%s'", ff) + " "
```
chrismuellner, line 71: you can collect the feature flags in an array, and then combine them at the end using `strings.join(ffs, ", ")`

chrismuellner (approved):

PR #1968, opened 2023-07-24
Get http proxy in OneAgent from mounted secret instead of env variable

albertogdd:
## Description

Remove http proxy env variable in os-OneAgent and mount a secret with its value instead (in the same fashion as ActiveGate does, actually using the same secret for both now).

## How can this be tested?

Apply a Dynakube with proxy (either with `value` or `valueFrom`) and check that OneAgent gets a mounted volume on the expected path (`/var/lib/dynatrace/secrets/internal-proxy`). Also, you can check its value in the container shell. (see pictures)

image
image

Check also that the reconciliator creating the proxy secret works, it should create `<dynakubeName>-internal-proxy`.

image

## Checklist


albertogdd:
Based on the closed PR #1930

The bug is now solved on OneAgent side.

PR #1977, opened 2023-07-25
Replicate connectionInfo to injected namespaces

0sewa0:
## Description

We forgot to provide the information needed for the codemodules to communicate (tenant token + endpoints) in case no CSI driver is used.

To be able to do this, we have replicate this information to all the injected namespaces, in the already existing `dynatrace-dynakube-config` secret.

As a side effect, this change also fixes the issue with the tenant alias, as the connectionInfo contains the correct tenant. (this part needs further testing, to check if we need to do anything extra)

## How can this be tested?

Deploy application monitoring without csi and inject sample apps.
Tipp: If you use a proxy, in the log of the proxy it will show up how the calls to "localhost" will fail, as the codemodule will try to use the proxy to reach "localhost", which is its default communication endpoint if none is provided.

## Checklist

### DIFF
--- a/src/initgeneration/initgeneration.go
+++ b/src/initgeneration/initgeneration.go
@@ -6,6 +6,7 @@ import (
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
 	"github.com/Dynatrace/dynatrace-operator/src/config"
+	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/connectioninfo"
 	"github.com/Dynatrace/dynatrace-operator/src/dtclient"
 	"github.com/Dynatrace/dynatrace-operator/src/kubeobjects"
 	"github.com/Dynatrace/dynatrace-operator/src/kubesystem"
@@ -155,10 +156,17 @@ func (g *InitGenerator) createSecretConfigForDynaKube(ctx context.Context, dynak
 		return nil, errors.WithStack(err)
 	}
 
+	tenantToken, err := g.getTenantToken(ctx, dynakube)
+	if err != nil {
+		return nil, err
+	}
+
 	return &standalone.SecretConfig{
 		ApiUrl:              dynakube.Spec.APIURL,
 		ApiToken:            getAPIToken(tokens),
 		PaasToken:           getPaasToken(tokens),
+		TenantToken:         tenantToken,
+		ConnectionInfo:      dynakube.Status.OneAgent.ConnectionInfoStatus,
 		Proxy:               proxy,
 		NoProxy:             dynakube.FeatureNoProxy(),
 		NetworkZone:         dynakube.Spec.NetworkZone,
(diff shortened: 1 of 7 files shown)


0sewa0:
NOTE TO SELF: I should introduce a conditional check for this replication, as its only needed when no csi-driver is used, and it costs an extra API call (get the tenant token from secret)

src/standalone/run.go
```diff
@@ -145,8 +145,8 @@ func (runner *Runner) getProcessModuleConfig() (*dtclient.ProcessModuleConfig, e
 	if runner.config.Proxy != "" {
 		processModuleConfig.AddProxy(runner.config.Proxy)
 	}
-
-	return processModuleConfig.AddTenantUUID(runner.config.TenantUUID), nil
+	processModuleConfig = processModuleConfig.AddConnectionInfo(runner.config.ConnectionInfo, runner.config.TenantToken)
```
luhi-DT, line 148: why not returning this immediately instead of caching in unused variable?

src/initgeneration/initgeneration.go
```diff
@@ -150,21 +151,22 @@ func (g *InitGenerator) createSecretConfigForDynaKube(ctx context.Context, dynak
...
-		return nil, errors.WithStack(err)
+		return nil, err
 	}
 
 	return &standalone.SecretConfig{
 		ApiUrl:              dynakube.Spec.APIURL,
 		ApiToken:            getAPIToken(tokens),
 		PaasToken:           getPaasToken(tokens),
+		TenantToken:         tenantToken,
+		ConnectionInfo:      dynakube.Status.OneAgent.ConnectionInfoStatus,
```
luhi-DT, line 164: what if this is not fast enough reconciled in the dynakube? e.g Dynakube is created and immediatly afterwards apps are started?

luhi-DT (changes requested):

src/initgeneration/initgeneration.go
0sewa0: the webhook would error, as it won't be able to get the `tenantToken` secret

src/standalone/run.go
0sewa0: done

aorcholski:
value of  `tenantToken` field  shoud be  masked with `***`  in the log

```
{"apiUrl":"https://ABC.dynatracelabs.com/api","apiToken":"***","paasToken":"***","proxy":"","noProxy":"","networkZone":"","trustedCAs":"","skipCertCheck":false,"tenantToken":"XYZ"
```

src/initgeneration/initgeneration.go
```diff
@@ -174,6 +179,15 @@ func (g *InitGenerator) createSecretConfigForDynaKube(ctx context.Context, dynak
 	}, nil
 }
 
+func (g *InitGenerator) getTenantToken(ctx context.Context, dynakube *dynatracev1beta1.DynaKube) (string, error) {
+	secretQuery := kubeobjects.NewSecretQuery(ctx, g.client, g.apiReader, log)
```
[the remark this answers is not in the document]
0sewa0: nice, done

luhi-DT (approved):

aorcholski (approved):

andriisoldatenko (approved):

0sewa0:
thx for the reviews, I will close this PR as it solves the problem in an overcomplicated way.
I will split it into 2 smaller PRs for better understanding:
1. A PR for fixing the use of the tenantUUID, in the provisioner and initcontainer
2. A PR for not using {{skipMetadata=true}} in case of applicationMonitoring without CSI-driver, as here we are not gaining much or any benefit from using a "raw" codemodule per pod, because each pod still has to download a whole codemodule each.

PR #1981, opened 2023-07-26
Use correct tenantUUID

0sewa0:
## Description
The `connectionInfo` we get from the cluster already contains the tenantUUID, and its the actual correct one, and not the one that we parse out of the apiURL, which caused problem if the apiURL was an alias of a tenant.

(Sidenote: most of the changes are related to rewriting the 22 months old tests 😅  )

## How can this be tested?
When using an apiURL that points to a tenant using an alias, setup cloudNativeFullStack (or application monitoring with CSI driver), check that the agents can communicate. (if alias is used in as the tenant, then the agents will complain about not being able to connect, this should not happen)

## Checklist

### DIFF
--- a/src/controllers/csi/provisioner/controller.go
+++ b/src/controllers/csi/provisioner/controller.go
@@ -214,9 +214,7 @@ func (provisioner *OneAgentProvisioner) updateAgentInstallation(ctx context.Cont
 
 	latestProcessModuleConfig = latestProcessModuleConfig.
 		AddHostGroup(dk.HostGroup()).
-		AddConnectionInfo(dk.Status.OneAgent.ConnectionInfoStatus, tenantToken).
-		AddTenantUUID(dynakubeMetadata.TenantUUID)
-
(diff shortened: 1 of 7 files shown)


src/controllers/csi/provisioner/controller.go
```diff
@@ -214,9 +214,7 @@ func (provisioner *OneAgentProvisioner) updateAgentInstallation(ctx context.Cont
 
 	latestProcessModuleConfig = latestProcessModuleConfig.
 		AddHostGroup(dk.HostGroup()).
-		AddConnectionInfo(dk.Status.OneAgent.ConnectionInfoStatus, tenantToken).
-		AddTenantUUID(dynakubeMetadata.TenantUUID)
```
aorcholski, line 218: `AddTenantUUID` is not used , should be removed from `dtclient.ProcessModuleConfig`

0sewa0: done

luhi-DT (approved):

PR #1982, opened 2023-07-26
Only use skipMetadata=true with csiDriver

0sewa0:
## Description
Incase of applicationMonitoring without CSI driver we shouldn't use the "raw" codemodules as they don't provide much if any efficiency improvements, and we would need to do connectionInfo duplication (see https://github.com/Dynatrace/dynatrace-operator/pull/1977) in each injected namespace which would just add complexity and not much more.

## How can this be tested?
Scenario 1.: Test applicationMonitoring without CSI driver, check that the codemodules can connect when injected into sample apps.
Scenario 2.: Doable check that I didn't break what I did in: https://github.com/Dynatrace/dynatrace-operator/pull/1963

## Checklist


andriisoldatenko (approved):
code LGTM,
I ran `make test/e2e/activegate` and it works.

PR #1988, opened 2023-07-27
Clean up better in switchmode e2e tests

0sewa0:
## Description

We currently have the problem that the switchmode e2e test cases mess with the outcome of other tests ran after them.
In the switchmode e2e test cases we have to be extra careful when it comes to clean up.
As we are using both classic and cloudnative we have to clean up:
- the csi driver's filesystem due to the cloudnative part
    - not doing this might cause unpredictable behaviour in tests coming after this one 
- the run the uninstall.sh because of the classic part
   - this is extra important on CRIO (so mainly openshift) because not running the uninstall.sh here will cause other tests just to fail straight out.

Sidenote: The current "helpers" for generalizing the steps were not really designed for this, so I had to do some quick modifications there, but now they feel a bit like "abstraction for the sake of abstraction" which I am not a fan of.
So we should probably rethink them in a follow up, to be more explicit in each test.

## How can this be tested?
run (⚠️ on Openshift aswell)
 `make test/e2e/cloudnative/switchmodes`
 ` make test/e2e/classic/switchmodes`

## Checklist

### DIFF
--- a/test/helpers/steps/assess/dynakube.go
+++ b/test/helpers/steps/assess/dynakube.go
@@ -37,6 +37,13 @@ func UpdateDynakube(builder *features.FeatureBuilder, testDynakube dynatracev1be
 	builder.Assess("dynakube updated", dynakube.Update(testDynakube))
 }
 
+func DeleteDynakube(builder *features.FeatureBuilder, testDynakube dynatracev1beta1.DynaKube) {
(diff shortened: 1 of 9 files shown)


andriisoldatenko:
@0sewa0 pre-commit hook :)

luhi-DT (approved):

PR #1997, opened 2023-08-01
Adds EdgeConnect controller

StefanHauth:
## Description

Adds a controller for the new EdgeConnect CRD and an accompanying basic E2E test.

## How can this be tested?

* Deploy the operator
* create a simple EdgeConnect CR

```
kind: EdgeConnect
metadata:
  name: myEdgeConnect
  namespace: dynatrace
spec:
  apiServer: "<tenant>.dynatrace.com"

  oauth:
    clientSecret: "secret"
    endpoint: "endpoint"
    resource: "urn"
```

* deploy this EdgeConnect CR
* check operator logs for "reconcling edgeconnct" log entries
* check that the "UpdatedTimestamp" filed in the EC status gets updated
* run E2E test with : `make test/e2e/edgeconnect`

## Checklist


src/controllers/edgeconnect/controller.go
```diff
@@ -0,0 +1,124 @@
...
+// Controller reconciles an EdgeConnect object
+type Controller struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the api-server
+	client            client.Client
+	apiReader         client.Reader
+	scheme            *runtime.Scheme
+	config            *rest.Config
+	operatorNamespace string
+	clusterID         string
```
0sewa0, line 37: `clusterID` field is just there and not used (we set it but that is it)

src/controllers/edgeconnect/controller.go
```diff
@@ -0,0 +1,124 @@
...
+	versionProvider   version.ImageVersionFunc
```
0sewa0, line 38: putting this here is premature
- in the followup we want to extract the reusable logic from the `version` package so this 100% will need to be refactored 
- we shouldn't follow the pattern we have in the DynaKube controller, where we only mock 1 callback that will be used in a totally different package. Instead we should make it possible to mock the whole logic of said package, by making the (for example) `VersionReconciler` configurable in the controller, and not just 1 call (out of 20) that this sub package would make

src/controllers/edgeconnect/controller.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func Add(mgr manager.Manager, _ string) error {
+	kubeSysUID, err := kubesystem.GetUID(mgr.GetAPIReader())
+	if err != nil {
+		return errors.WithStack(err)
+	}
```
0sewa0, line 45: we don't need this as its connected to `clusterID`

src/controllers/edgeconnect/controller.go
```diff
@@ -0,0 +1,124 @@
...
+
+// Controller reconciles an EdgeConnect object
+type Controller struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the api-server
+	client            client.Client
+	apiReader         client.Reader
+	scheme            *runtime.Scheme
+	config            *rest.Config
+	operatorNamespace string
```
0sewa0, line 36: `operatorNamespace` is not used, we don't need to add stuff prematurely

src/controllers/edgeconnect/controller.go
```diff
@@ -0,0 +1,124 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	defaultUpdateInterval = 30 * time.Minute
```
0sewa0, line 25: where did the others go?
```
	errorUpdateInterval   = 1 * time.Minute
	changesUpdateInterval = 5 * time.Minute
```

src/controllers/edgeconnect/controller_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+func TestEdgeConnectController_Reconcile(t *testing.T) {
+	t.Run(`Create works with minimal setup`, func(t *testing.T) {
+		controller := &Controller{
+			client:    fake.NewClient(),
+			apiReader: fake.NewClient(),
+		}
+		result, err := controller.Reconcile(context.TODO(), reconcile.Request{})
+
+		assert.NoError(t, err)
```
0sewa0, line 40:
```suggestion
		require.NoError(t, err)
```

src/controllers/edgeconnect/controller_test.go
```diff
@@ -0,0 +1,89 @@
...
+	testNamespace = "test-namespace"
+
+	testUID = "testUID"
+)
+
+func TestIsError(t *testing.T) {
+	assert.False(t, k8serrors.IsConflict(nil))
+}
+
+func TestEdgeConnectController_Reconcile(t *testing.T) {
```
0sewa0, line 32:
```suggestion
func TestReconcile(t *testing.T) {
```

src/controllers/edgeconnect/controller.go
```diff
@@ -0,0 +1,124 @@
...
+func (controller *Controller) updateEdgeConnectStatus(ctx context.Context, edgeConnect *edgeconnectv1alpha1.EdgeConnect) error {
+	log.Info("updateEdgeConnectStatus", "name", edgeConnect.Name, "namespace", edgeConnect.Namespace)
+	edgeConnect.Status.UpdatedTimestamp = metav1.Now()
+
+	err := controller.client.Status().Update(ctx, edgeConnect)
+	if k8serrors.IsConflict(err) {
+		log.Info("could not update edgeconnect due to conflict", "name", edgeConnect.Name)
+		return errors.WithStack(err)
+	} else if err != nil {
+		log.Error(err, "updateEdgeConnectStatus", "name", edgeConnect.Name, "namespace", edgeConnect.Namespace)
```
0sewa0, line 119: weird error message

src/scheme/scheme.go
```diff
@@ -18,6 +18,9 @@ package scheme
 
 import (
 	dynatracev1alpha1 "github.com/Dynatrace/dynatrace-operator/src/api/v1alpha1"
+	// needed to ensure proper package initialization order
+	_ "github.com/Dynatrace/dynatrace-operator/src/api/v1alpha1/dynakube"
+	_ "github.com/Dynatrace/dynatrace-operator/src/api/v1alpha1/edgeconnect"
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1/dynakube"
 	istiov1alpha3 "istio.io/client-go/pkg/apis/networking/v1alpha3"
 	corev1 "k8s.io/api/core/v1"
```
0sewa0, line 26: not your change that has the problem here but
this package is a bit messed up, because the `v1beta1` package is a bit out of line

did a quick fix [here](https://github.com/Dynatrace/dynatrace-operator/commit/d7646960ea1e9d9e6f727258e2be0ab2d7340f29)

if you want to add it, then do it
otherwise I can create a follow up

test/helpers/components/edgeconnect/edgeconnect.go
```diff
@@ -0,0 +1,68 @@
+//go:build e2e
```
0sewa0, line 1: AFAIK, the e2e tests are out of scope for this change

0sewa0 (changes requested):

src/controllers/edgeconnect/controller.go
```diff
@@ -0,0 +1,124 @@
...
+
+func Add(mgr manager.Manager, _ string) error {
+	kubeSysUID, err := kubesystem.GetUID(mgr.GetAPIReader())
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	return NewController(mgr, string(kubeSysUID)).SetupWithManager(mgr)
+}
+
+// NewController returns a new ReconcileDynaKube
```
waodim, line 49: Does this really return a new `ReconcileDynakube`?

src/controllers/edgeconnect/controller.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (controller *Controller) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	requeueAfter := defaultUpdateInterval
+
+	log.Info("reconciling EdgeConnect", "name", request.Name, "namespace", request.Namespace)
+
+	edgeConnect, err := controller.getEdgeConnect(ctx, request.Name, request.Namespace)
+	if err != nil {
+		log.Error(errors.WithStack(err), "reconciliation of edgeconnect failed", "name", request.Name, "namespace", request.Namespace)
```
waodim, line 80:
```suggestion
		log.Error(errors.WithStack(err), "reconciliation of EdgeConnect failed", "name", request.Name, "namespace", request.Namespace)
``

Comment (waodim) [reply src/controllers/edgeconnect/controller.go:80]: minor nitpick: Out of consistency I would suggest to change it like that (that goes also for every other appearance where edgeconnect is written in lower case)

File: src/controllers/edgeconnect/controller.go (hunk 1 above)
Comment (StefanHauth) [reply src/controllers/edgeconnect/controller.go:37]: removed

File: src/controllers/edgeconnect/controller.go (hunk 2 above)
Comment (StefanHauth) [reply src/controllers/edgeconnect/controller.go:38]: removed

File: src/controllers/edgeconnect/controller.go (hunk 4 above)
Comment (StefanHauth) [reply src/controllers/edgeconnect/controller.go:36]: removed

File: src/controllers/edgeconnect/controller.go (hunk 3 above)
Comment (StefanHauth) [reply src/controllers/edgeconnect/controller.go:45]: removed

File: src/controllers/edgeconnect/controller.go (hunk 5 above)
Comment (StefanHauth) [reply src/controllers/edgeconnect/controller.go:25]: Added `errorUpdateInterval`, `changesUpdateInterval` is not yet useful, except if we use it after updating the status timestamp which is done basically everytime now, so it would replace `defaultUpdateInterval`

File: src/controllers/edgeconnect/controller_test.go (hunk 6 above)
Comment (StefanHauth) [reply src/controllers/edgeconnect/controller_test.go:40]: changed

File: src/controllers/edgeconnect/controller_test.go (hunk 7 above)
Comment (StefanHauth) [reply src/controllers/edgeconnect/controller_test.go:32]: changed

File: src/controllers/edgeconnect/controller.go (hunk 8 above)
Comment (StefanHauth) [reply src/controllers/edgeconnect/controller.go:119]: upsi, seemed to be some leftover

File: test/helpers/components/edgeconnect/edgeconnect.go (hunk 10 above)
Comment (StefanHauth) [reply test/helpers/components/edgeconnect/edgeconnect.go:1]: Having proper tests from the beginning is a good thing, this includes E2E tests. They also provide an reproducible way beyond "works on my machine" of showing, that the controller works. Or quickly detect it doesn't. So while they were not mentioned in the ticket, I've taken the freedom to add them as I see them as a mandatory part of a new feature.

File: src/controllers/edgeconnect/controller.go (hunk 11 above)
Comment (StefanHauth) [reply src/controllers/edgeconnect/controller.go:49]: removed the whole comment, as it's kind of redundant and useless anyway

File: src/controllers/edgeconnect/controller.go (hunk 12 above)
Comment (StefanHauth) [reply src/controllers/edgeconnect/controller.go:80]: fixed

File: src/controllers/edgeconnect/controller.go [hunk 13]
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return NewController(mgr, string(kubeSysUID)).SetupWithManager(mgr)
+}
+
+// NewController returns a new ReconcileDynaKube
+func NewController(mgr manager.Manager, clusterID string) *Controller {
+	return NewEdgeConnectController(mgr.GetClient(), mgr.GetAPIReader(), mgr.GetScheme(), mgr.GetConfig(), clusterID)
+}
+
+func NewEdgeConnectController(kubeClient client.Client, apiReader client.Reader, scheme *runtime.Scheme, config *rest.Config, clusterID string) *Controller { //nolint:revive
```
aorcholski, line 54: `NewEdgeConnectController` could be a private function

StefanHauth: fixed

src/scheme/scheme.go
StefanHauth: I've cherry-picked your commit.

test/helpers/components/edgeconnect/edgeconnect.go
0sewa0: the thing is that we have other stuff that depends on having the base reconciler in the code but at the same time you did some refactoring in the e2e tests, which touches almost all scenarios there.
If the e2e tests were just a drop in thing, no major changes, then I would have no problem with it.
But the current state for the e2e tests and the helpers we have there are veeeeery dynakube/ single CR focused, which needs rethinking/refactoring, and I would prefer to do that rethinking not in this PR.

So if you really want e2e tests to be part of this change, then pls don't refactor the majority of the e2e logic, or we will be here forever, discussing how the e2e helpers should be refactored 😅 

I would prefer to tackle e2e tests in a separate PR, like we always did, furthermore we only add e2e tests when a feature is actually complete, while this PR only introduces a very simple part of the feature, not much to test.

StefanHauth: Fair point.

aorcholski (approved):

test/helpers/components/dynakube/dynakube.go
```diff
@@ -9,7 +9,8 @@ import (
 	"testing"
 
 	"github.com/Dynatrace/dynatrace-operator/src/api/status"
-	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
+	v1betadynakube "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1/dynakube"
```
0sewa0, line 13:
```suggestion
	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
	dynakubev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1/dynakube"
```

test/scenarios/cloudnative/public_registry/public_registry.go
```diff
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	"github.com/Dynatrace/dynatrace-operator/src/api/status"
+	"github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
```
0sewa0, line 11:
```suggestion
	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1"
```

0sewa0 (changes requested):
we will need to fix the rest of the code to use 
```
dynakubev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1/dynakube"
```
instead of 
```
```

but right now, just use it correctly where we must need to make a change

test/scenarios/cloudnative/public_registry/public_registry.go
StefanHauth: fixed

0sewa0 (approved):

PR #2010, opened 2023-08-04
Adds volume-profile label in openshift-csi manifest

luhi-DT:
## Description

Currently the CSI manifests on OCP are missing the "csi-ephemeral-volume-profile" label. The check is wrong to add it, when the manifests are generated, as the check expects "partial" to be false, but it actually is "csi" in case of openshift-csi.yaml.

## How can this be tested?

make manifests -> should generate an openshift-csi.yaml that has the CSIDriver resource and the label on it

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/Common/csi/csidriver.yaml
+++ b/config/helm/chart/default/templates/Common/csi/csidriver.yaml
@@ -18,7 +18,7 @@ kind: CSIDriver
 metadata:
   name: csi.oneagent.dynatrace.com
   labels:
-    {{- if (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
+    {{- if eq (include "dynatrace-operator.platform" .) "openshift" }}
     security.openshift.io/csi-ephemeral-volume-profile: "restricted"
     {{- end }}
     {{- include "dynatrace-operator.csiLabels" . | nindent 4 }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -270,8 +270,8 @@ jobs:
           hack/build/ci/generate-release-notes.sh
       - name: Generate manifests
         run: |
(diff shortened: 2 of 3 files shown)


config/helm/chart/default/templates/Common/csi/csidriver.yaml
```diff
@@ -18,7 +18,7 @@ kind: CSIDriver
 metadata:
   name: csi.oneagent.dynatrace.com
   labels:
-    {{- if (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
+    {{- if eq (include "dynatrace-operator.platform" .) "openshift" }}
```
andriisoldatenko, line 21: can you please update or add new helm test?

aorcholski: The code depends on 3 variables 
- `platform`
- needCSI (`or`): 
  - `csidriver.enabled`
  - `partial`

The test should also set all of them and cover at least 2 basic cases
```
- it: should contain correct labels for openshift with CSI enabled
    set:
      platform: openshift
      csidriver.enabled: true
      partial: false

  - it: should contain correct labels for openshift manifest
    set:
      platform: openshift
      csidriver.enabled: false
      partial: csi
```

aorcholski: please add `partial`  to the first case 

```
  - it: should contain correct labels for openshift with CSI enabled
    set:
      platform: openshift
      csidriver.enabled: true
      partial:false                     <----
    asserts:
```

aorcholski: thanks!

aorcholski (approved):

andriisoldatenko (approved):

PR #2026, opened 2023-08-07
Fixes pre-release action by generating correct manifest

waodim:
## Description

Previously the pre-release action created wrong manifests as:

1. the image was wrongly set (this should already be fixed)
2. the version label which is set via the `Chart.yaml`

The problem was that the corresponding make targets used commands that (`git branch --show-current`, etc.) did not work as intended inside the action. That is why I set the version in the `Chart.yaml` inside the action as it should normally be done anyway. This is done with instead of using `git branch`, `github.ref_name` inside the action.

## How can this be tested?

Probably the best way is to create a fork and trigger the action.

Have a look at this [release](https://github.com/waodim/dynatrace-operator/releases/tag/untagged-8c5346c573ceb8e9e841) from my fork where version label is set correctly as well as the image.

## Checklist


.github/workflows/release.yaml
```diff
@@ -272,6 +272,11 @@ jobs:
         run: |
           export "IMAGE=${{ matrix.url }}/${{ secrets[matrix.repository] }}"
           export "TAG=$(echo "${{ github.ref }}" | awk -F"/" '{print $3}')"
+          export "CHART_VERSION=$(echo "${{ github.ref_name }}" | sed 's/^v//')"
```
luhi-DT, line 275: instead of copying the content of the version.mk we should try to adapt the config.mk to properly support a tag imho

luhi-DT (changes requested):

.github/workflows/release.yaml
luhi-DT: You could make CHART_VERSION configurable e.g and set it via an env var

waodim: I am having a look into that :+1:

luhi-DT (approved):

PR #2041, opened 2023-08-10
Not fail csi-init on network problems

0sewa0:
## Description

The init container for the csi driver (`csi-init`) is correcting the filesystem and the database.

This logic uses the kubernetes api (ei.: we make some network calls) which can't happen if the istio-proxy container is not running.
- This is because istio injected an init container that basically messes up the pods networking so it only works with the istio-proxy running. This init container runs before ours.

Solution: The `csi-init` will not fail, it will produce some errors and logs: (the first big error is printed by the framework, can't really do much about it)
```
{"level":"info","ts":"2023-08-10T14:37:22.579Z","logger":"version","msg":"dynatrace-operator","version":"snapshot-fix-csi-init-ocp-istio","gitCommit":"41ad0f0f2e817b0d487c9e12cc81a611ac93371d","buildDate":"2023-08-10T10:38:07+00:00","goVersion":"go1.20.7","platform":"linux/amd64"}
{"error":"Get \"https://172.30.0.1:443/api?timeout=32s\": dial tcp 172.30.0.1:443: connect: connection refused","level":"error","logger":"main","msg":"Failed to get API Group-Resources","stacktrace":"github.com/Dynatrace/dynatrace-operator/src/logger.logSink.Error
    github.com/Dynatrace/dynatrace-operator/src/logger/logger.go:40
sigs.k8s.io/controller-runtime/pkg/log.(*DelegatingLogSink).Error
    sigs.k8s.io/controller-runtime@v0.14.6/pkg/log/deleg.go:138
github.com/go-logr/logr.Logger.Error
    github.com/go-logr/logr@v1.2.4/logr.go:299
sigs.k8s.io/controller-runtime/pkg/cluster.New
    sigs.k8s.io/controller-runtime@v0.14.6/pkg/cluster/cluster.go:161
sigs.k8s.io/controller-runtime/pkg/manager.New
    sigs.k8s.io/controller-runtime@v0.14.6/pkg/manager/manager.go:351
github.com/Dynatrace/dynatrace-operator/src/cmd/csi/init.createManager
    github.com/Dynatrace/dynatrace-operator/src/cmd/csi/init/manager.go:12
github.com/Dynatrace/dynatrace-operator/src/cmd/csi/init.CommandBuilder.buildRun.func1
    github.com/Dynatrace/dynatrace-operator/src/cmd/csi/init/builder.go:58
github.com/spf13/cobra.(*Command).execute
    github.com/spf13/cobra@v1.7.0/command.go:940
github.com/spf13/cobra.(*Command).ExecuteC
    github.com/spf13/cobra@v1.7.0/command.go:1068
github.com/spf13/cobra.(*Command).Execute
    github.com/spf13/cobra@v1.7.0/command.go:992
main.main
    github.com/Dynatrace/dynatrace-operator/src/cmd/main.go:115
runtime.main
    runtime/proc.go:250","ts":"2023-08-10T14:37:22.581Z"}{"level":"info","ts":"2023-08-10T14:37:22.581Z","logger":"csi-init","msg":"failed to create manager, will only run non-network related corrections and checks","err":"Get \"https://172.30.0.1:443/api?timeout=32s\": dial tcp 172.30.0.1:443: connect: connection refused"}
{"level":"info","ts":"2023-08-10T14:37:22.582Z","logger":"csi-metadata","msg":"no kubernetes client configured, skipping orphaned volume metadata cleanup"}
{"level":"info","ts":"2023-08-10T14:37:22.582Z","logger":"csi-metadata","msg":"no kubernetes client configured, skipping orphaned dynakube metadata cleanup"}
{"level":"info","ts":"2023-08-10T14:37:22.582Z","logger":"csi-metadata","msg":"CSI filesystem corrected, linked deprecated agent binary to current location (tenant|version-bin)","movedBins":[]}
{"level":"info","ts":"2023-08-10T14:37:22.582Z","logger":"csi-metadata","msg":"The current overview of the csi metadata","overview":{"volumes":[],"dynakubes":[],"osAgentVolumes":[]}}
```
So it will only do the symlink logic, which is needs to run
While the database cleanup won't run, which is not the most optimal, but its manly for edge cases:
- If some metadata is "orphaned", then some agent binaries and/or folders won't get cleaned up.
    - The volume metadata cleanup matters if the customer would **force delete** pods while the csi-driver is not available.
    - The dynakube metadata cleanup matters if the customer deletes the dynakube from the cluster while the csi-driver is not available, and later the user will use a dynakube with a different name.

Sidenote: Updates the SCC we use for the sample apps, as isito needs volumes I added during injection. The rename matters because in our e2e pipelines the previous was present by default, so to not overwrite it. (also our logic for adding the SCC is a `create` and not an `apply`, so it wouldn't even work if it was already there)

## How can this be tested?

Run `make test/e2e/cloudnative/istio` on an openshift cluster that has istio.

## Checklist


src/controllers/csi/metadata/correctness.go
```diff
@@ -49,14 +49,18 @@ func (checker *CorrectnessChecker) CorrectCSI(ctx context.Context) error {
 
 // Removes volume entries if their pod is no longer exists
 func (checker *CorrectnessChecker) removeVolumesForMissingPods(ctx context.Context) error {
+	if checker.apiReader == nil {
```
luhi-DT, line 52: Why not moving this check to CorrectCSI() and check it only once for both removeVolumes and removeDynakubes ?

src/cmd/csi/init/builder.go
```diff
@@ -56,7 +57,7 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
 
 		csiManager, err := createManager(builder.namespace, kubeConfig)
 		if err != nil {
```
luhi-DT, line 59: Can't we make that error check more explicit than just allowing any error?

luhi-DT (changes requested):

src/controllers/csi/metadata/correctness.go
0sewa0: is cleaner IMO, keeps each check independent

src/cmd/csi/init/builder.go
0sewa0: not really there is not special error that is returned, and I would like to avoid checking the message and/or StatusCode as they can be fickle.

luhi-DT (approved):

PR #2042, opened 2023-08-11
Not fail csi-init on network problems

0sewa0:
## Description
 
cherry pick of  #2041 

## How can this be tested?

same as  #2041 

## Checklist

### DIFF
--- a/src/controllers/csi/metadata/correctness.go
+++ b/src/controllers/csi/metadata/correctness.go
@@ -15,18 +15,18 @@ import (
 )
 
 type CorrectnessChecker struct {
-	cl     client.Client
-	fs     afero.Fs
-	path   PathResolver
-	access Access
+	apiReader client.Reader
+	fs        afero.Fs
+	path      PathResolver
+	access    Access
 }
 
-func NewCorrectnessChecker(cl client.Client, access Access, opts dtcsi.CSIOptions) *CorrectnessChecker {
+func NewCorrectnessChecker(cl client.Reader, access Access, opts dtcsi.CSIOptions) *CorrectnessChecker {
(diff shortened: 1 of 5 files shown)


src/controllers/csi/metadata/correctness.go
```diff
@@ -15,18 +15,18 @@ import (
...
 }
 
-func NewCorrectnessChecker(cl client.Client, access Access, opts dtcsi.CSIOptions) *CorrectnessChecker {
+func NewCorrectnessChecker(cl client.Reader, access Access, opts dtcsi.CSIOptions) *CorrectnessChecker {
 	return &CorrectnessChecker{
-		cl:     cl,
-		fs:     afero.NewOsFs(),
-		path:   PathResolver{RootDir: opts.RootDir},
-		access: access,
+		apiReader: cl,
```
andriisoldatenko, line 26: not a blocker, but would be nice to rename to `r` or `ar` :)

andriisoldatenko (approved):

src/controllers/csi/metadata/correctness.go
0sewa0: if I did that I would start a naming war 😅 

we use `apiReader` often in the codebase, and we try to avoid the "super-short-names-that-go-likes" because it makes the code kinda unreadable, and think its even in the [CONTRIBUTION.md](https://github.com/Dynatrace/dynatrace-operator/blob/main/CONTRIBUTING.md#general)

andriisoldatenko: okay :) I'm totally fine with that :) but cl is a bit odd.

PR #2054, opened 2023-08-22
Bugfix/always pass proxy to oa even when empty

StefanHauth:
## Description

When a proxy has been configured in a DynaKube previously and that proxy settings has been reomoved, OneAgents would not take into account that change, as they would still use their cached settings. We need to not only remove the proxy setting we pass to the OneAgent but explicityl set it as "empty".

## How can this be tested?
- deploy the operator
- setup a proxy in your cluster
- apply a DynaKube that uses that proxy
- deploy a test application to also include code modules in your test
- check if OneAgents and code modules are connecting to your tenant
  * Note: OneAgents can'r resolve a service address, use proxy IP
- check if code modules can connect (use agent registry in debug UI)
- remove proxy and see if agents loose the connection
- remove proxy from Dynakube config and re-apply
- restart app pods
- wait a bit and see if OneAgents and code modules are able to connect again

## Checklist


andriisoldatenko:
code LGTM. 
tested using a modified version of the e2e test for proxy in my cluster when removed proxy from dynokube:
```
│ Containers:                                                                                                                                                                                                                                                                                │
│   dynatrace-oneagent:                                                                                                                                                                                                                                                                      │
│     Container ID:  containerd://d95ed6c789a14bafae12117b7a8db09ef6f840d95c72db95db2b891b37b4d16c                                                                                                                                                                                           │
│     Image:         mts86258.dev.dynatracelabs.com/linux/oneagent:latest                                                                                                                                                                                                                    │
│     Image ID:      mts86258.dev.dynatracelabs.com/linux/oneagent@sha256:bde6d8a28f894163631f3b364cfc08487cd8b80d32a1ac58b12db9898a94b5de                                                                                                                                                   │
│     Port:          <none>                                                                                                                                                                                                                                                                  │
│     Host Port:     <none>                                                                                                                                                                                                                                                                  │
│     Args:                                                                                                                                                                                                                                                                                  │
│       --set-proxy=                                                                                                                                                                                                                                                                         │
│       --set-host-property=OperatorVersion=$(DT_OPERATOR_VERSION)                                                                                                                                                                                                                           │
│       --set-tenant=$(DT_TENANT)                                                                                                                                                                                                                                                            │
│       --set-server={$(DT_SERVER)}                                                                                                                                                                                                                                                          │
│       --set-host-id-source=k8s-node-name  
```

andriisoldatenko (approved):

PR #2058, opened 2023-08-24
Add code-style section about func param ordering

0sewa0:
## Description

As it came up multiple times recently, I added to the code style guide how we should order our parameters of our functions.

## How can this be tested?

Read it 😅 

## Checklist

### DIFF
--- a/doc/coding-style-guide.md
+++ b/doc/coding-style-guide.md
@@ -38,7 +38,7 @@ Ordering of **function parameters** should be:
 4. Simple/base types
    - example: `data string`
 
-Ordering of **return values** is simpler, the `err error` should always be the last, and AVOID returning more than 2 values. If more than 2 return values are needed try splitting the logic up or collecting the return values in a `struct`.
+Ordering of **return values** is more straightforward; the `err error` should always be the last, and AVOID returning more than two values. If more than two return values are needed, try splitting the logic or collecting the return values in a `struct`.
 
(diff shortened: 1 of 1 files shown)


andriisoldatenko (dismissed):

doc/coding-style-guide.md
```diff
@@ -26,6 +27,21 @@
...
+   - example: `kubeClient client.Client`
+3. Data Structs
+   - example: `pod corev1.Pod`
+4. Simple/base types
+   - example: `data string`
+
+Ordering of **return values** is simpler, the `err error` should always be the last, and AVOID returning more than 2 values. If more than 2 return values are needed try splitting the logic up or collecting the return values in a `struct`.
+
+So a full example: `func ExampleFunc(ctx context.Context, kubeClient client.Client, pod corev1.Pod, data string) (corev1.Pod, error) {...}`
+
```
andriisoldatenko, line 44:
```suggestion
## Function Parameter and Return-Value Order
Ordering of **function parameters** should be:
1. Convention
   - example: `ctx context.Context`
2. Interfaces
   - example: `kubeClient client.Client`
3. Data Structs
   - example: `pod corev1.Pod`
4. Simple/base types
   - example: `data string`

Ordering of **return values** is more straightforward; the `err error` should always be the last, and AVOID returning more than two values. If more than two return values are needed, try splitting the logic or collecting the return values in a `struct`.

So a full example: `func ExampleFunc(ctx context.Context, kubeClient client.Client, pod corev1.Pod, data string) (corev1.Pod, error) {...}`

```

andriisoldatenko (approved):

PR #2066, opened 2023-08-28
Change check for adding scc permission to csi clusterrole

luhi-DT:
## Description

Currently our CSI Driver clusterrole lacks access to the privileged SCC, which blocks creation of CSI pods. The problem is that the check for adding the privileged SCC access is wrong an excludes the case where partial is set to csi.

## How can this be tested?

Run make manifests and check openshift-csi.yaml if it contains the correct permissions for accessing the privileged SCC

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/Common/csi/clusterrole-csi.yaml
+++ b/config/helm/chart/default/templates/Common/csi/clusterrole-csi.yaml
@@ -62,7 +62,7 @@ rules:
       - get
       - list
       - watch
-  {{- if (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
+  {{- if (eq (include "dynatrace-operator.platform" .) "openshift") }}
   - apiGroups:
       - security.openshift.io
     resourceNames:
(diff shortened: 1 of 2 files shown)


config/helm/chart/default/templates/Common/csi/clusterrole-csi.yaml
```diff
@@ -62,7 +62,7 @@ rules:
       - get
       - list
       - watch
-  {{- if (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
+  {{- if (eq (include "dynatrace-operator.platform" .) "openshift") }}
```
0sewa0, line 65: Sidenote: IMO, we should "rework" `openshiftOrOlm`, so that it does not have the `(include "dynatrace-operator.partial" .) "false")` part as the name in no way describes that behaviour, and that is the reason this bug snuck in.

```
{{- define "dynatrace-operator.openshiftOrOlm" -}}
{{- if and (or (eq (include "dynatrace-operator.platform" .) "openshift") (.Values.olm)) (eq (include "dynatrace-operator.partial" .) "false") -}}
    {{ default "true" }}
{{- end -}}
{{- end -}}

```

0sewa0 (approved):

config/helm/chart/default/templates/Common/csi/clusterrole-csi.yaml
luhi-DT: will create a follow up for that :)

PR #2068, opened 2023-08-29
WIP research logic to ensure secrets - single controller

aorcholski:
## Description

Logic to ensure namespace secrets moved from pod mutator to dynakube controller. Dynakube controller listens to namespace events to make sure ingestendpoint and init secrets are created ASAP in monitored namespace.

The following scenarios do not trigger dynakube reconcile event:
- namespace CREATE event w/o matching selector label
- namespace UPDATE#1, UPDATE#2, DELETE sequence
- namespace UPDATE w/o `dynakube/instance` label
- namespace UPDATE because of controller doing mapping

Approach using [delaying event queue](https://github.com/Dynatrace/dynatrace-operator/blob/86c62db462e2d4aa7c8b7744de7a6a03d2ea5623/src/controllers/dynakube/requestmapper/mapper.go#L151) seems to be most promising so far:
- in case of incoming stream of events - we can limit the number of dynakube reconciliation calls to a specific number

Approach using [rate limited queue](https://github.com/Dynatrace/dynatrace-operator/blob/86c62db462e2d4aa7c8b7744de7a6a03d2ea5623/src/controllers/dynakube/requestmapper/mapper.go#L150):
- when incoming stream of namespace events starts the dynakube controller is called too often
From DefaultControllerRateLimiter src:
```
// DefaultControllerRateLimiter is a no-arg constructor for a default rate limiter for a workqueue.  It has
// both overall and per-item rate limiting.  The overall is a token bucket and the per-item is exponential
```

## How can this be tested?
Script to generate big amount of namespace events:
```
/bin/bash namespace.sh create
/bin/bash namespace.sh delete
/bin/bash namespace.sh verify
```
In case of my machine ~600 is still OK, 1000 running parallel processes -> other random processes die.

```
#!/bin/bash
if [ "${1}" == "create" ]; then
    
  for i in {1..500} ; do
    cat <<EOF | kubectl create -f - &
apiVersion: v1
kind: Namespace
metadata:
  labels:
    monitor: star
  name: sample-${i}
EOF
  done

elif [ "${1}" == "delete" ]; then

  for i in {1..500} ; do
    cat <<EOF | kubectl delete -f - &
apiVersion: v1
kind: Namespace
metadata:
  labels:
    monitor: star
  name: sample-${i}
EOF
  done

elif [ "${1}" == "verify" ]; then

  for i in {1..500} ; do
    /bin/bash ${0} verifybg sample-${i} &
  done

elif [ "${1}" == "verifybg" ]; then
    if ! kubectl -n ${2} get secrets/dynatrace-data-ingest-endpoint secrets/dynatrace-dynakube-config 2>/dev/null 1>&2 ; then
      echo "secret(s) not found in ${2}"
    else
      echo ${2}
    fi
fi
```

## Checklist

### DIFF
--- a/config/helm/chart/default/templates/Common/webhook/clusterrole-webhook.yaml
+++ b/config/helm/chart/default/templates/Common/webhook/clusterrole-webhook.yaml
@@ -42,18 +42,18 @@ rules:
       - secrets
     verbs:
       - create
-  - apiGroups:
-      - ""
-    resources:
-      - secrets
-    resourceNames:
-      - dynatrace-dynakube-config
-      - dynatrace-data-ingest-endpoint
-    verbs:
-      - get
-      - list
-      - watch
-      - update
+  #- apiGroups:
+  #    - ""
+  #  resources:
+  #    - secrets
+  #  resourceNames:
+  #    - dynatrace-dynakube-config
+  #    - dynatrace-data-ingest-endpoint
+  #  verbs:
+  #    - get
+  #    - list
+  #    - watch
+  #    - update
   # data-ingest workload owner lookup
   - apiGroups:
       - ""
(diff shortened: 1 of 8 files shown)


aorcholski:
the solution was presented to colleagues

PR #2070, opened 2023-08-30
Set default values when Healthcheck time vars are not set

albertogdd:
## Description

Set default values when Interval, Timeout, StartPeriod and Retries are not set in the Docker Image config.

## How can this be tested?

- Run unit tests

## Checklist

### DIFF
--- a/src/controllers/dynakube/version/healthcheck.go
+++ b/src/controllers/dynakube/version/healthcheck.go
@@ -1,0 +1,79 @@
+package version
+
+import (
+	"context"
+	"time"
+
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/src/registry"
+	containerv1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	DefaultHealthConfigInterval    = time.Second * 10
+	DefaultHealthConfigStartPeriod = time.Second * 1200
+	DefaultHealthConfigTimeout     = time.Second * 30
+	DefaultHealthConfigRetries     = 3
+)
+
+// Constructor setting default values for docker image HealthConfig
+func newHealthConfig() *containerv1.HealthConfig {
+	healthConfig := &containerv1.HealthConfig{}
+	healthConfig.Test = []string{}
+	healthConfig.Interval = DefaultHealthConfigInterval
+	healthConfig.StartPeriod = DefaultHealthConfigStartPeriod
+	healthConfig.Timeout = DefaultHealthConfigTimeout
+	healthConfig.Retries = DefaultHealthConfigRetries
+
+	return healthConfig
+}
+
+func SetOneAgentHealthcheck(ctx context.Context, apiReader client.Reader, registryClient registry.ImageGetter, dynakube *dynatracev1beta1.DynaKube, imageUri string) error {
+	imageInfo, err := PullImageInfo(ctx, apiReader, registryClient, dynakube, imageUri)
+	if err != nil {
+		return errors.WithMessage(err, "error pulling image info")
+	}
+
+	configFile, err := (*imageInfo).ConfigFile()
+	if err != nil {
+		return errors.WithMessage(err, "error reading image config file")
+	}
+
+	// Healthcheck.Test values from go-containerregistry documentation:
+	// {} : inherit healthcheck
+	// {"NONE"} : disable healthcheck
+	// {"CMD", args...} : exec arguments directly
+	// {"CMD-SHELL", command} : run command with system's default shell
+	if configFile.Config.Healthcheck != nil && len(configFile.Config.Healthcheck.Test) > 0 {
+		var testCommand []string
+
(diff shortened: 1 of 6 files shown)


andriisoldatenko (approved):

src/controllers/dynakube/version/healthcheck.go
```diff
@@ -0,0 +1,79 @@
...
+
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/src/registry"
+	containerv1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	DefaultHealthConfigInterval    = time.Second * 10
```
andriisoldatenko, line 15:
```suggestion
	DefaultHealthConfigInterval    = 10 * time.Second
```
Usually we use <num> than seconds.

andriisoldatenko (changes requested):

src/controllers/dynakube/version/healthcheck.go
```diff
@@ -0,0 +1,79 @@
...
+	DefaultHealthConfigStartPeriod = time.Second * 1200
+	DefaultHealthConfigTimeout     = time.Second * 30
+	DefaultHealthConfigRetries     = 3
+)
+
+// Constructor setting default values for docker image HealthConfig
+func newHealthConfig() *containerv1.HealthConfig {
+	healthConfig := &containerv1.HealthConfig{}
```
andriisoldatenko, line 23:
```suggestion
	return &containerv1.HealthConfig{
	  Test: []string{}
	Interval: DefaultHealthConfigInterval
	... etc
}
```
just simplify

src/controllers/dynakube/version/healthcheck.go
```diff
@@ -0,0 +1,79 @@
...
+			testCommand = []string{"/bin/sh", "-c"}
+			testCommand = append(testCommand, configFile.Config.Healthcheck.Test[1:]...)
+		case "CMD":
+			testCommand = configFile.Config.Healthcheck.Test[1:]
+		}
+
+		if len(testCommand) > 0 {
+			healthConfig := newHealthConfig()
+			healthConfig.Test = testCommand
+			if configFile.Config.Healthcheck.Interval != 0 {
```
andriisoldatenko, line 63: I think since we set defaults early on line 61, we should compare with defailts:
E.g.
```
 configFile.Config.Healthcheck.Interval != DefaultHealthConfigInterval
```

andriisoldatenko: nvm I misread it;

src/controllers/dynakube/version/healthcheck.go
```diff
@@ -0,0 +1,79 @@
+package version
```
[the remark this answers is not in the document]
albertogdd: Mmm, maybe we can move it to the dynakube/status package? as it sets the Dynakube OneAgent status, what do you think?
I would try to not create even more packages in the codebase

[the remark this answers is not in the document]
andriisoldatenko: IMO you mix 2 ideas: 
a) healthconfig related code
b) setStatusOneAgent 

So i would better create new package for healthconfig and use it in reconcile and move SetStatus to where it was.

src/controllers/dynakube/version/healthcheck.go
```diff
@@ -0,0 +1,79 @@
...
+	healthConfig.Test = []string{}
+	healthConfig.Interval = DefaultHealthConfigInterval
+	healthConfig.StartPeriod = DefaultHealthConfigStartPeriod
+	healthConfig.Timeout = DefaultHealthConfigTimeout
+	healthConfig.Retries = DefaultHealthConfigRetries
+
+	return healthConfig
+}
+
+func SetOneAgentHealthcheck(ctx context.Context, apiReader client.Reader, registryClient registry.ImageGetter, dynakube *dynatracev1beta1.DynaKube, imageUri string) error {
```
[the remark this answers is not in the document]
andriisoldatenko: So health check code can be independent of set status and we ca reuse it later to set status for another healthcheck

src/controllers/dynakube/version/healthcheck.go
albertogdd: Fixed

src/controllers/dynakube/version/healthcheck.go
albertogdd: Fixed

src/controllers/dynakube/version/healthcheck.go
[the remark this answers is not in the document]
albertogdd: That makes total sense. Fixed!

src/controllers/dynakube/version/healthcheck.go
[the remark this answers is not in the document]
albertogdd: I tried to, but we end up with some import loops (because PullImageInfo is also in `version` package...)
I would say, let's leave it like this, it's not a big deal.

andriisoldatenko (approved):
LGTM, looks much better now 🌮

PR #2074, opened 2023-09-01
Set version label based on .Status.Version.Version from EdgeConnect

andriisoldatenko:
## Description

This PR use `.Status.Version.Version` as version label ~otherwise fallback to "latest".~

## How can this be tested?

Unit tests has been added.

## Checklist


luhi-DT (approved):

luhi-DT (changes requested):
As discussed needs rework

0sewa0:
I wouldn't call this a hotfix, and we don't really have hotfixes as that would require us to release it quick which we haven't done so far.
This is just a normal PR, as the thing we are fixing was never released and caused no actual issues

andriisoldatenko:
> I wouldn't call this a hotfix, and we don't really have hotfixes as that would require us to release it quick which we haven't done so far. This is just a normal PR, as the thing we are fixing was never released and caused no actual issues

@0sewa0 fixed description.

src/controllers/edgeconnect/deployment/deployment.go
```diff
@@ -88,11 +89,18 @@ func prepareContainerEnvVars(instance *edgeconnectv1alpha1.EdgeConnect) []corev1
 }
 
 func buildAppLabels(instance *edgeconnectv1alpha1.EdgeConnect) *kubeobjects.AppLabels {
+	version := "latest"
+	statusVersion := instance.Status.Version.Version
+
+	if statusVersion != "" && len(validation.IsQualifiedName(statusVersion)) == 0 {
```
chrismuellner, line 95: We should just take whatever is in the status, and not validate it again. Similar with the empty check, we should default the version status to `latest` and use it as tag verbatim, without introducing duplicate logic here.

andriisoldatenko: But it's not implemented, I created a ticket for that, we don't have code which reads LABEL.

0sewa0 (approved):

luhi-DT (approved):

PR #2078, opened 2023-09-01
Add helm package generation to GitHub actions

gkrenn:
## Description

This PR add the ability to generate helm packages directly in the release action. 

In addition the manifest generation was refactored such that we don't use a matrix anymore but rather a for loop.

## How can this be tested?

Create a tag, run these changes in GitHub actions and check if new files were appended to draft release

## Checklist


.github/workflows/release.yaml
```diff
@@ -267,31 +261,66 @@ jobs:
...
-          TAG: ${{ needs.prepare.outputs.version }}
-          CHART_VERSION: ${{ needs.prepare.outputs.version }}
-        run: make manifests
-      - name: Rename manifests
+          version: ${{ needs.prepare.outputs.version }}
+          platforms: |
+            kubernetes
+            openshift
+          images: |
+            docker.io/dynatrace/dynatrace-operator
```
luhi-DT, line 276: These should be taken from the Secrets

.github/workflows/release.yaml
```diff
@@ -267,31 +261,66 @@ jobs:
...
+        uses: ./.github/actions/build-helm
+        with:
+          version_without_prefix: ${{ needs.prepare.outputs.version_without_prefix }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          secring: ${{ secrets.HELM_SECRING }}
+          passphrase: ${{ secrets.HELM_PASSPHRASE }}
+          output-dir: "./helm-pkg"
+      - name: Create pre-release
         uses: softprops/action-gh-release@v1
+        if: ${{ contains(github.ref, '-rc.') }}
```
luhi-DT, line 295: to me it seems splitting this part up into 2 separate would maybe duplicate some logic but would make it easier to read and distinguishable, not really a problem imho but a suggestion from my side. What is you take on that?

luhi-DT (changes requested):

.github/workflows/release.yaml
```diff
@@ -267,31 +261,66 @@ jobs:
...
       - name: Generate manifests
         env:
-          IMAGE: ${{ matrix.url }}/${{ secrets[matrix.repository] }}
-          TAG: ${{ needs.prepare.outputs.version }}
-          CHART_VERSION: ${{ needs.prepare.outputs.version }}
-        run: make manifests
-      - name: Rename manifests
+          version: ${{ needs.prepare.outputs.version }}
+          platforms: |
+            kubernetes
```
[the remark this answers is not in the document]
chrismuellner: it would make sense to define platforms in a matrix with each of the platform specific fields (image, generated file, ...) specified there and not spread around in different parts of the script. This could simplify the `generate-release-manifests` script and not require a loop there.

The helm package is only required once and should be done in a separate job, so we'd have the following jobs:
1. create manifests
    - kubernets
    - openshift
    - gke-autopilot 	
2. create helm package
3. create pre-release

chrismuellner (dismissed):

.github/workflows/release.yaml
gkrenn: I wrote it here directly, as I do not see the need to put it in a secret, as this information is publicly available.

Where do you see the advantage in using a secret?

luhi-DT (approved):

PR #2098, opened 2023-09-08
Add gke-autopilot namespaces to default ignored namespaces

waodim:
## Description

This is a change to make gke-autopilot run, because previously when the operator was deployed things would not work as there would be an error message that certain namespaces (with prefix gke- and gmp-) cannot be updated or changed in any way. That is why those are added now to the ignored namespaces.

Also, in order to make the e2e tests run I added a timeout in wait.go because when I ran the e2e tests on gke-autopilot it took a lot of time and it timed out almost everytime. When this was fixed I had problems with the sample pods which where randomly stuck in a ContainerStatusUnknown state. This is fixed by adding ephemeral storage of 2Gi to the sample app manifest.

## How can this be tested?

Create a gke-autopilot cluster and run the tests that should be possible (application monitoring default for example)

## Checklist


test/helpers/kubeobjects/deployment/wait.go
```diff
@@ -34,5 +35,5 @@ func WaitFor(name string, namespace string) features.Func {
 }
 
 func WaitUntilReady(resource *resources.Resources, deployment *appsv1.Deployment) error {
-	return wait.For(conditions.New(resource).DeploymentConditionMatch(deployment, appsv1.DeploymentAvailable, corev1.ConditionTrue))
+	return wait.For(conditions.New(resource).DeploymentConditionMatch(deployment, appsv1.DeploymentAvailable, corev1.ConditionTrue), wait.WithTimeout(time.Minute*15))
```
andriisoldatenko, line 38:
```suggestion
	return wait.For(conditions.New(resource).DeploymentConditionMatch(deployment, appsv1.DeploymentAvailable, corev1.ConditionTrue), wait.WithTimeout(time.Minute*15))
```
Better to extract to constant and use more natural format `<Number> * time.Minute`

andriisoldatenko: Perfect example:

https://github.com/dynatrace/dynatrace-operator/blob/2bf21151d19357a0575de5d1bb1a2b42f5f4ff6d/src/controllers/dynakube/version/healthconfig.go#L14-L15

test/helpers/kubeobjects/deployment/wait.go
```diff
@@ -15,8 +13,12 @@ import (
 	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 	"sigs.k8s.io/e2e-framework/pkg/features"
+	"testing"
+	"time"
 )
 
+const timeout = time.Minute * 15
```
andriisoldatenko, line 20:
```suggestion
const DeploymentAvailableTimeout = 15 * time.Minute
```

test/helpers/kubeobjects/deployment/wait.go
```diff
@@ -34,5 +36,5 @@ func WaitFor(name string, namespace string) features.Func {
 }
 
 func WaitUntilReady(resource *resources.Resources, deployment *appsv1.Deployment) error {
-	return wait.For(conditions.New(resource).DeploymentConditionMatch(deployment, appsv1.DeploymentAvailable, corev1.ConditionTrue))
+	return wait.For(conditions.New(resource).DeploymentConditionMatch(deployment, appsv1.DeploymentAvailable, corev1.ConditionTrue), wait.WithTimeout(timeout))
```
andriisoldatenko, line 39:
```suggestion
	return wait.For(conditions.New(resource).DeploymentConditionMatch(deployment, appsv1.DeploymentAvailable, corev1.ConditionTrue), wait.WithTimeout(DeploymentAvailableTimeout))
```

andriisoldatenko (approved):

PR #2099, opened 2023-09-08
Add make target for running e2e tests on gke-autopilot

waodim:
## Description

With this the make targets for e2e tests are added. This works only for the following tests:
* applicationMonitoring/default
* appMon/withoutcsi
* appMon/labelversion
* appMon/readonlycsi **(NOTE: this test will not work with the current version of the operator, as there were problems with the whitelisting of the image and csi Driver initContainer, this will be fixed)**

## How can this be tested?

Create a gke-autopilot cluster and run the added make targets. But please be aware that the tests will only work after https://github.com/Dynatrace/dynatrace-operator/pull/2098 is merged.

## Checklist


andriisoldatenko (approved):
IMO, we can merge the change, because `make targets` are correct. 

Cc @luhi-DT, issues with tests themselves we can improve in follow-up PRs.

waodim:
> IMO, we can merge the change because `make targets` are correct.
> 
> Cc @luhi-DT, issues with tests themselves we can improve in follow-up PRs.

I would agree and I can btw confirm that the tests run as before when using an older agent version.

PR #2105, opened 2023-09-11
Rework how istio is handled

0sewa0:
## Description
So the "old" way (what is currently in main) how we handle istio is described in this image: (press the `<-->` looking icon to view it in big 😅  )
```mermaid
---
title: Old
---
flowchart LR
    dynakube[Dynakube\nName: test\nEnableIstio: true]
    subgraph operator
        direction TB
        reconcile-everything...
        reconcile-istio
    end
    subgraph for-api-url
        direction TB
        se1[ServiceEntry\nName: some-hash \nPurpose: For the api-url]
        vs1[VirtualService\nName: some-hash \nPurpose: For the api-url]
        se1 -.- vs1
    end
    subgraph for-communication-hosts
        direction TB
        se2[ServiceEntry\nName: some-hash \nPurpose: For oneagent IP-based communication-hosts]
        se3[ServiceEntry\nName: some-hash \nPurpose: For oneagent FQDN-based communication-hosts]
        vs2[VirtualService\nName: some-hash \nPurpose: For oneagent FQDN-based communication-hosts]
        se3 -.- vs2
        se2 ~~~ se3
    end


    exp1([Redundant, as for the operator to get to this point, this should already exist]):::bad-red
    exp2([+there is a overcomplicated delete logic, as the names constantly change, because it's hash based on the contents\n+If any error happens, we just log it, Which can silently break functionality]):::bad-red

    dynakube-->operator
    reconcile-istio-->for-api-url
    reconcile-istio-->for-communication-hosts
    for-api-url --> exp1
    for-communication-hosts --> exp2
    exp1-->exp2

    classDef bad-red stroke:#f00
```
According to this PR how it works:
```mermaid
---
title: New
---
flowchart LR
    dynakube[Dynakube\nName: test\nEnableIstio: true]
    subgraph operator
        direction TB
        reconcile-istio-for-api-url
        reconcile-connectionInfo...
        reconcile-istio-communication-hosts
        reconcile-everything-else...
    end
    subgraph for-api-url
        direction TB
        se1[ServiceEntry\nName: test-dk-fqdn-operator \nPurpose: For the api-url]
        vs1[VirtualService\nName: test-dk-fqdn-operator \nPurpose: For the api-url]
        se1 -.- vs1
    end
    subgraph for-communication-hosts
        direction TB
        se2[ServiceEntry\nName: test-dk-ip-oneagent \nPurpose: For oneagent IP-based communication-hosts]
        se3[ServiceEntry\nName: test-dk-fqdn-oneagent \nPurpose: For oneagent FQDN-based communication-hosts]
        vs2[VirtualService\nName: test-dk-fqdn-oneagent \nPurpose: For oneagent FQDN-based communication-hosts]
        se3 -.- vs2
        se2 ~~~ se3
    end


    exp1([No longer redundant, user no longer need to configure istio for the API Url by hand]):::good-green
    exp2([Update logic is simple as the names are static + names mean something\nincase of error, we actually stop what we are doing]):::good-green

    dynakube-->operator
    reconcile-istio-for-api-url-->for-api-url
    reconcile-istio-communication-hosts-->for-communication-hosts
    for-api-url --> exp1
    for-api-url --> exp2
    for-communication-hosts --> exp2

    classDef good-green stroke:#0f0
```

TLDR:
How we used to handle the istio integration was less then ideal:
- The user had to do extra steps for the operator to even get to the isito reconcile part
- The resources we created had a hash for a name, that only complicated things and confused users
- The resources we created had weird labels (that we were afraid to change as they were used as labels selector due to the names being unpredictable)
- The code of the istio reconciler was "refactored" multiple times, but was never re-thinked, so the code itself was overly convoluted for what it was doing.

~Still TODO:~
~- fix e2e tests...~
The fix for the e2e test was rather simple as its not the best and that should be fixed separately in a follow up. (also there are multiple e2e PRs in the work so I don't want to add to the pile right now)

## How can this be tested?

1. have a cluster with istio with `REGISTRY_ONLY` mode
2. have istio inject the operator's namespace
3. deploy a dynakube with `EnableIstio: true`
4. it should just work
    - previously you would need to create a `ServiceEntry` + `VirtualService` for your APIURL to get to the istio reconcile part, you DON'T have to do this anymore
    -  The created `ServiceEntries` + `VirtualServices` should be named nicely and have nice labels.

## Checklist


src/controllers/dynakube/istio/client.go
```diff
@@ -0,0 +1,180 @@
...
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+type ClientBuilder func(config *rest.Config, scheme *runtime.Scheme, namespace string) (*Client, error)
+
+// Client - an adapter for the external istioclientset library
+type Client struct {
+	IstioClient istioclientset.Interface
```
andriisoldatenko, line 22:
```suggestion
	IstioClientset istioclientset.Interface
```
istioClient = istio.Client
istioClient.IstioClient -- sounds wired, wdyt?

src/controllers/dynakube/istio/client.go
```diff
@@ -0,0 +1,180 @@
...
+	if k8serrors.IsNotFound(err) {
+		return nil, nil
+	} else if err != nil {
+		log.Info("failed to get current virtual service", "name", name, "error", err.Error())
+		return nil, errors.WithStack(err)
+	}
+	return virtualService, nil
+}
+
+func (cl *Client) ApplyVirtualService(ctx context.Context, owner metav1.Object, newVirtualService *istiov1alpha3.VirtualService) error {
```
andriisoldatenko, line 59:
```suggestion
func (cl *Client) CreateVirtualService(ctx context.Context, owner metav1.Object, newVirtualService *istiov1alpha3.VirtualService) error {
```

andriisoldatenko: jsut to respect other naming convention: Get/Create/Delete/Update (CRUD like ;)

src/controllers/dynakube/istio/serviceentry.go
```diff
@@ -1,20 +1,13 @@
 package istio
```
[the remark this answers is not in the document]
andriisoldatenko: nvm, it's named as in client https://github.com/istio/client-go/blob/master/pkg/clientset/versioned/typed/networking/v1alpha3/serviceentry.gen.go.

src/controllers/dynakube/istio/client.go
0sewa0: "create" is bit misleading for me, as it also updates it if its already there
also "apply" is also not 100% true, because that would mean I use a "patch" which I don't 😅 

will go with `CreateOrUpdate...` as we already use that naming, and I can't come up with anything better

src/controllers/dynakube/istio/client.go
0sewa0: done

src/controllers/dynakube/istio/client.go
0sewa0: done

src/controllers/dynakube/istio/istio.go
```diff
@@ -19,41 +14,21 @@ var (
...
-	}
-	sum := sha256.Sum256([]byte(strings.Join(result, "\n")))
-	return hex.EncodeToString(sum[:])
-}
-
+// TODO: Maybe move to Client
 // CheckIstioInstalled run discovery query for server resource for group version
-func CheckIstioInstalled(discoveryclient discovery.DiscoveryInterface) (bool, error) {
-	_, err := discoveryclient.ServerResourcesForGroupVersion(IstioGVR)
+func CheckIstioInstalled(discoveryClient discovery.DiscoveryInterface) (bool, error) {
```
andriisoldatenko, line 19: I think it's a good idea to move it to a new istio client. Should be easy refactoring, since we use it only in 2 places:
- https://github.com/dynatrace/dynatrace-operator/blob/11d19c8cc31d8ae1bf12d771a3114ccc6ab9ad56/src/webhook/validation/dynakube/istio.go#L21
and in `setupIstio`. And we always have instance of istio client :)

andriisoldatenko:
practically LGTM
```
PASS
ok      github.com/Dynatrace/dynatrace-operator/test/scenarios/cloudnative/istio        509.196s
```

andriisoldatenko (approved):

src/controllers/dynakube/istio/istio.go
0sewa0: fixed

andriisoldatenko (dismissed):

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -173,29 +169,37 @@ func (controller *Controller) createDynakubeMapper(ctx context.Context, dynakube
...
-		updated, err = istio.NewReconciler(controller.config, controller.scheme, ic).Reconcile(dynakube, communicationHosts)
-		if err != nil {
-			// If there are errors log them, but move on.
-			log.Info("istio failed to reconcile objects", "error", err)
-		}
+	isInstalled, err := istioClient.CheckIstioInstalled()
+	if err != nil {
+		return nil, errors.WithMessage(err, "failed to initialize istio client")
+	} else if !isInstalled {
+		return nil, errors.New("istio not installed yet is enabled, aborting reconciliation, check configuration")
```
luhi-DT, line 185:
```suggestion
		return nil, errors.New("istio not installed, yet it is enabled, aborting reconciliation, check configuration")
```

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -173,29 +169,37 @@ func (controller *Controller) createDynakubeMapper(ctx context.Context, dynakube
...
 	}
-	return updated
+	istioReconciler := istio.NewReconciler(istioClient)
+	err = istioReconciler.ReconcileAPIUrl(ctx, dynakube)
+	if err != nil {
+		return nil, errors.WithMessage(err, "failed to reconcile istio objects for API url")
+	}
+	log.Info("reconciled istio objects for API url")
```
luhi-DT, line 192: why not putting this log into the reconciler itself?

src/controllers/dynakube/istio/config.go
```diff
@@ -1,9 +1,23 @@
...
 var (
 	log = logger.Factory.GetLogger("dynakube-istio")
 )
+
+const (
+	OperatorComponent = "operator"
+	OneAgentComponent = "oneagent"
+
+	IstioGVRName    = "networking.istio.io"
+	IstioGVRVersion = "v1alpha3"
```
luhi-DT, line 18: do we need to document with which istio version we are compatible?

0sewa0: The official docu has examples for both `v1alpha3` and `v1beta1`, so we are safe as far as support, but we should move to v1beta1, but we should look into that separately

src/controllers/dynakube/dynakube_controller.go
0sewa0: done

src/controllers/dynakube/dynakube_controller.go
0sewa0: done

PR #2129, opened 2023-09-15
Block OA rollout until communication hosts are available.

StefanHauth:
## Description

If a network zone is configured in a DynaKube, OneAgent rollout and code modules injection shall be postponed until an ActiveGate is available so that proper communication with DT server is possible.
OneAgents rollout shall be blocked until then. Code modules injection shall be prevented with the annotation `oneagent.dynatrace.com/reason` set to `EmptyConnectionInfo`.

## How can this be tested?
* Create a network zone (https://www.dynatrace.com/support/help/dynatrace-api/environment-api/network-zones/put-network-zone) 
  * configure it with fallback mode "None"
* Deploy operator
* Configure a cloud native Dynakube using the network zone and an ActiveGate
* Deploy dynakube and watch ActiveGate and OneAgent pods startup
  * first only ActiveGate shall come up
  * OA pods shall come up roughly a minute after AG has started
  * Operator logs shall contain log lines containing `postponing OneAgent deployment`
* Deploy a sample app while AG is still starting, Code Modules shall not be injected, annotation shall be set
  * could be more easily checked by deploying the Dynakube without an AG, so that agent rollout would be blocked forever
  
## Checklist


src/controllers/dynakube/connectioninfo/reconciler.go
```diff
@@ -15,6 +15,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var NoOneAgentCommunicationHostsError = errors.New("no communication hosts for OneAgent are available")
```
0sewa0, line 18: This is not the way of creating a special error and checking for it in another place,
this way any other package could just overwrite this `var` and totally break the whole logic. (Also the moment this error is wrapped in another error the `errors.Is` will break)

The proper way is to create a `type` that implements the `Error` interface ([see more](https://go.dev/blog/error-handling-and-go)), for which you can check for with `errors.As(err, MySpecialError)`. Which will return `true` or `false` and do the conversion aswell.

src/controllers/dynakube/connectioninfo/reconciler.go
```diff
@@ -92,7 +94,19 @@ func (r *Reconciler) reconcileOneAgentConnectionInfo() error {
...
+	log.Info("OneAgent connection info updated")
+
+	if len(connectionInfo.Endpoints) == 0 {
+		log.Info("tenant has no endpoints", "tenant", connectionInfo.TenantUUID)
+	}
+
+	if len(connectionInfo.CommunicationHosts) == 0 {
+		log.Info("no OneAgent communication hosts received, tenant API requests not yet throttled")
+		return NoOneAgentCommunicationHostsError
+	}
```
0sewa0, line 106: weird check, is it possible that `Endpoints` are empty but the `CommunicationHosts` are not ?
AFAIK the `CommunicationHosts` are just the `Endpoints` but formatted.

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -108,44 +110,58 @@ type Controller struct {
...
+func (controller *Controller) reconcile(ctx context.Context, dynaKube *dynatracev1beta1.DynaKube) (reconcile.Result, error) {
+	oldStatus := *dynaKube.Status.DeepCopy()
+
+	controller.requeueAfter = defaultUpdateInterval
+
+	err := controller.reconcileDynaKube(ctx, dynaKube)
+
+	var serverErr dtclient.ServerError
+	switch {
+	case errors.As(err, &serverErr) && (serverErr.Code == http.StatusTooManyRequests || serverErr.Code == http.StatusServiceUnavailable):
```
0sewa0, line 139: this is proper custom error checking 👍

0sewa0: nitpick: we could have a little abstraction for `errors.As(err, &serverErr) && (serverErr.Code == http.StatusTooManyRequests || serverErr.Code == http.StatusServiceUnavailable)`, like `isDynatraceAPIUnavailable(err)` or something like that

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -269,6 +285,18 @@ func (controller *Controller) reconcileDynaKube(ctx context.Context, dynakube *d
 	return controller.reconcileComponents(ctx, dynatraceClient, dynakube)
 }
 
+func (controller *Controller) reconcileConnectionInfo(ctx context.Context, dynakube *dynatracev1beta1.DynaKube, dynatraceClient dtclient.Client) error {
+	err := connectioninfo.NewReconciler(ctx, controller.client, controller.apiReader, controller.scheme, dynakube, dynatraceClient).Reconcile()
+	if errors.Is(err, connectioninfo.NoOneAgentCommunicationHostsError) {
+		// missing communication hosts is not an error per se and shall not stop reconciliation, just make sure next reconciliation is happening ASAP
+		// this situation will clear itself after AG has been started
+		controller.setRequeueAfterIfNewIsShorter(errorUpdateInterval)
```
0sewa0, line 293: > ... is not an error ...

but still using the `errorUpdateInterval` reads a bit backwards 😄 
we could rename `errorUpdateInterval` to something better

src/dtclient/connection_info.go
```diff
@@ -6,6 +6,13 @@ import (
 	"github.com/pkg/errors"
 )
 
+// CommunicationHost => struct of connection endpoint
+type CommunicationHost struct {
+	Protocol string // nolint:unused
+	Host     string // nolint:unused
+	Port     uint32 // nolint:unused
+}
```
0sewa0, line 14: shouldn't this be in the `oneagent_connection_info.go` as its only relevant for the oneagent connection info

src/dtclient/connection_info.go
```diff
@@ -6,6 +6,13 @@ import (
 	"github.com/pkg/errors"
```
0sewa0, line 6: file should be renamed to `activegate_connection_info.go` if it only has activegate relevant functionality

0sewa0 (changes requested):

src/controllers/dynakube/connectioninfo/reconciler.go
StefanHauth: TBH IDK, that's how the checks have been done previously as well, so I just kept them as good as possible. But for this issue the `CommunicationHosts` is what counts.

src/dtclient/connection_info.go
StefanHauth: true

src/controllers/dynakube/connectioninfo/reconciler.go
StefanHauth: Actually the doc of Is says: 
``` // Is reports whether any error in err's chain matches target. ```
So wrapping should not be a problem. But I can make it an own type. Wanted to keep things as simple as possible.

StefanHauth: I honestly think, it makes error type checking cumbersome because you need an extra var for no other reason. Probably will wrap this in an extra helper.

0sewa0: discussed offline, outcome: this is fine

0sewa0 (approved):

PR #2132, opened 2023-09-18
Add validator to check dynakube name for leading digits

waodim:
## Description

With this a new validator is added that checks. whether a `dynakube` is applied that starts with a digit. This should be prohibited, as then also services etc. would be created with a leading digit. There are rules for service names that need to conform to DNS-1035 labels and cannot start with numbers.

## How can this be tested?

Apply a `dynakube` with a name that starts with a digit.

## Checklist

### DIFF
--- a/src/webhook/validation/dynakube/dynakube_name.go
+++ b/src/webhook/validation/dynakube/dynakube_name.go
@@ -13,7 +13,10 @@ const (
 
 func nameStartsWithDigit(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
 	dynakubeName := dynakube.Name
-	firstChar := rune(dynakubeName[0])
+	var firstChar rune
+	if dynakubeName != "" {
+		firstChar = rune(dynakubeName[0])
+	}
 
 	if unicode.IsDigit(firstChar) {
 		return errorDigitInName
--- a/src/webhook/validation/dynakube/dynakube_name.go
+++ b/src/webhook/validation/dynakube/dynakube_name.go
@@ -1,9 +1,8 @@
 package dynakube
 
 import (
-	"unicode"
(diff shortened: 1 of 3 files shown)


src/webhook/validation/dynakube/dynakube_name.go
```diff
@@ -0,0 +1,26 @@
...
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1/dynakube"
+)
+
+const (
+	errorDigitInName = `The DynaKube's specification has an invalid name: It starts with a digit.
+	`
+)
+
+func nameStartsWithDigit(dv *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
+	dynakubeName := dynakube.Name
```
andriisoldatenko, line 15: just curious, why don't we validate in same way as k8s does?

https://github.com/kubernetes/kubernetes/blob/f7dacb63cc0b0e79450b2f6d97b1c6e6c46070bd/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L236-L245C2
```
func IsDNS1035Label(value string) []string {
        var errs []string
        if len(value) > DNS1035LabelMaxLength {
                errs = append(errs, MaxLenError(DNS1035LabelMaxLength))
        }
        if !dns1035LabelRegexp.MatchString(value) {
                errs = append(errs, RegexError(dns1035LabelErrMsg, dns1035LabelFmt, "my-name", "abc-123"))
        }
        return errs
}
```

waodim: tbh, i did not know that k8s does it that way, I will use it, thank you!

andriisoldatenko (approved):

PR #2135, opened 2023-09-19
Skipping to send mark for termination event if host not found

andriisoldatenko:
## Description

Skipping to send mark for termination event if host not found

## How can this be tested?

1. Create cluster with 2 nodes
2. Edit configMap `dynatrace-node-cache` and change IP address (to simulate host not found from dynatrace) and seen/mark timestamp to past 2h.
3. Delete node from cluster (NOT cordon, delete)

Expected results:
```
{"level":"info","ts":"2023-09-19T15:58:52.195+0200","logger":"nodes","msg":"sending mark for termination event to dynatrace server","dynakube":"dynakube","ip":"10.128.0.223","node":"gke-andrii-test-default-pool-cf2748c9-pt7f"}
{"level":"info","ts":"2023-09-19T15:58:52.734+0200","logger":"dtclient","msg":"hosts cache: ignoring inactive hosts","ids":[""]}
{"level":"info","ts":"2023-09-19T15:58:52.736+0200","logger":"nodes","msg":"skipping to send mark for termination event","dynakube":"dynakube","nodeIP":"10.128.0.223","reason":"host not found","reasonVerbose":"host not found\ngithub.com/Dynatrace/dynatrace-operator/src/dtclient.init\n\t/Users/andrii.soldatenko/work/dynatrace-operator/src/dtclient/config.go:8\nruntime.doInit1\n\t/opt/homebrew/Cellar/go/1.21.1/libexec/src/runtime/proc.go:6740\nruntime.doInit\n\t/opt/homebrew/Cellar/go/1.21.1/libexec/src/runtime/proc.go:6707\nruntime.main\n\t/opt/homebrew/Cellar/go/1.21.1/libexec/src/runtime/proc.go:249\nruntime.goexit\n\t/opt/homebrew/Cellar/go/1.21.1/libexec/src/runtime/asm_arm64.s:1197"}
```

and check that broken node is not exists in configMap

## Checklist


src/controllers/nodes/nodes_controller.go
```diff
@@ -54,6 +54,7 @@ func (controller *Controller) SetupWithManager(mgr ctrl.Manager) error {
 func nodeDeletionPredicate(controller *Controller) predicate.Predicate {
 	return predicate.Funcs{
 		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
+			log.Info("nodeDeletionPredicate")
```
0sewa0, line 57: a debug log I think 😅

src/dtclient/dynatrace_client.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+var ErrHostNotFound = errors.New("host not found")
```
0sewa0, line 16: isn't this incorrect? shouldn't we create a `type` and use `errors.As` to check it, isn't that safer/nicer ?

I complained about it in @StefanHauth 's PR [here](https://github.com/Dynatrace/dynatrace-operator/pull/2129#discussion_r1329957567), sooooo I wanna know, if its me who does not know something 😅

0sewa0 (changes requested):

src/dtclient/dynatrace_client.go
andriisoldatenko: fixed

src/controllers/nodes/nodes_controller.go
andriisoldatenko: fixed in 6bd00cc6

src/dtclient/dynatrace_client.go
andriisoldatenko: fixed in 6bd00cc6

StefanHauth: Its probably safer, but checking for a special error later on isn't nicer because it always requires creating an extra var. For simple things, I'm also more in favor if this style. While it's ofc not impossible, I'm not sure how likely it is, that some overwrites this var.

0sewa0 (approved):

PR #2141, opened 2023-09-21
Always write cluster id to container.conf

StefanHauth:
## Description
Always write the cluster id to the container.conf file independent of any host monitoring OneAgent being present on the node  ore not.

## How can this be tested?

* Deploy operator
* Deploy application monitoring DynaKube
* Deploy sample app
* Shell into app pod
* cat /var/lib/dynatrace/oneagent/agent/config/container.conf
  * make sure that `k8s_cluster_id` and `k8s_node_name` entries are present

## Checklist


waodim (approved):
Tested it - works

PR #2146, opened 2023-09-22
Add context parameter to Reconcile function of the Reconciler interface

aorcholski:
## Description

The `Reconcile(ctx)` function of `Reconciler` interface requires context parameter to enforce the use of the context.

ActiveGate Reconcilers used context.TODO() instead of top-level context. Now the context is passed along the call stack:
- authtoken
- capability
- customproperties
- proxy
- statefulset

Additionally the following Reconcilers also pass the context to functions that need it instead of keeping it in the Reconciler struct: 
- connectioninfo
- deploymentmetadata
- dtpullsecret
- webhook_cert_controller

## How can this be tested?
All e2e tests should PASS.

## Checklist

### DIFF
--- a/src/controllers/dynakube/activegate/internal/proxy/reconciler_test.go
+++ b/src/controllers/dynakube/activegate/internal/proxy/reconciler_test.go
@@ -45,7 +45,7 @@ func newTestReconcilerWithInstance(client client.Client) *Reconciler {
 func TestReconcileWithoutProxy(t *testing.T) {
 	t.Run(`reconcile dynakube without proxy`, func(t *testing.T) {
 		r := newTestReconcilerWithInstance(fake.NewClientBuilder().Build())
-		err := r.Reconcile()
+		err := r.Reconcile(context.TODO())
 
 		require.NoError(t, err)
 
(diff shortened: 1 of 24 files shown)


waodim (approved):
LGTM

src/controllers/dynakube/activegate/internal/proxy/reconciler_test.go
```diff
@@ -45,7 +45,7 @@ func newTestReconcilerWithInstance(client client.Client) *Reconciler {
 func TestReconcileWithoutProxy(t *testing.T) {
 	t.Run(`reconcile dynakube without proxy`, func(t *testing.T) {
 		r := newTestReconcilerWithInstance(fake.NewClientBuilder().Build())
-		err := r.Reconcile()
+		err := r.Reconcile(context.TODO())
```
0sewa0, line 48: Nitpick: I think we should start using `context.Background` in tests. So we can use `context.TODO` as an actual TODO. (I might have been for using `context.TODO` in tests, and told people to do this, but I was wrong 😬 )

also I should probably add this in the styleguide or somewhere 😅

0sewa0: you don't have to update anything you haven't touched! but if we are touching it right now, lets do it properly

andriisoldatenko: agree with @0sewa0 ☝️

aorcholski: agree with you :) fixed

0sewa0 (approved):

PR #2151, opened 2023-09-27
Fix `container.conf` contents for cloud native

waodim:
## Description

This unittest should cover the cloudNative test case. For more info: https://github.com/Dynatrace/dynatrace-operator/pull/2150 

NOTE: I now also readded content to the `container.conf` file, because it was removed in https://github.com/Dynatrace/dynatrace-operator/pull/2150 by accident.

## How can this be tested?

Run the unit tests.

## Checklist


src/standalone/files.go
```diff
@@ -24,7 +24,10 @@ k8s_basepodname %s
 k8s_namespace %s
 `
 
-	k8NodeNameFormatString = `k8s_node_name %s
+	k8HostInfoFormatString = `k8s_node_name %s
+[host]
+tenant
```
chrismuellner, line 29:
```suggestion
tenant %s
```
should also be adjusted in other locations

chrismuellner (changes requested):

src/standalone/files.go
waodim: fixed in https://github.com/Dynatrace/dynatrace-operator/pull/2151/commits/b4add95502ea42ed738533dcbf9c97b686be4571

chrismuellner (approved):

PR #2152, opened 2023-09-27
Update oneagent readiness and liveness probe defaults

chrismuellner:
## Description

Default oneagent liveness probe was too short to reliable start the container:
- Removed default initial delay used for both readiness and liveness probe
- Use "healthcheck start period" as is for liveness probe
	- Higher values are not an issue, as it just delays the first health check
	- Pod is still marked as healthy until the first probe is run
- Limit "healthcheck start period" to 90s, but allow lower values, for readiness probe
	- Value should be limited to a sensible value, as probe needs to run once for pod to be marked as ready
- Add success threshold value of 1 for both probes (same as default, but more explicit)

## How can this be tested?

Apply a Dynakube that deploys oneagent pods:
- readiness/liveness probe should both be set
- liveness probe
	- values are used as is, no limit
- readiness probe
	- tenant registry images should be limited to initial delay of 90s (instead of 1200s)
	- immutable images should have a lower initial delay of 60s

## Checklist

### DIFF
--- a/src/controllers/dynakube/oneagent/daemonset/daemonset_test.go
+++ b/src/controllers/dynakube/oneagent/daemonset/daemonset_test.go
@@ -503,9 +503,9 @@ func TestPodSpecServiceAccountName(t *testing.T) {
 func TestPodSpecProbes(t *testing.T) {
 	expectedHealthcheck := containerv1.HealthConfig{
 		Test:        []string{"echo", "super pod"},
-		Interval:    time.Second * 10,
-		Timeout:     time.Second * 30,
-		StartPeriod: time.Second * 60,
+		Interval:    10 * time.Second,
+		Timeout:     30 * time.Second,
+		StartPeriod: 60 * time.Second,
 		Retries:     3,
 	}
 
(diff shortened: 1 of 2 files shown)


src/controllers/dynakube/oneagent/daemonset/daemonset_test.go
```diff
@@ -500,85 +500,76 @@ func TestPodSpecServiceAccountName(t *testing.T) {
...
-						},
-					},
-				},
-			},
-		}
-		podSpec := builder.podSpec()
+func TestPodSpecProbes(t *testing.T) {
+	expectedHealthcheck := containerv1.HealthConfig{
+		Test:        []string{"echo", "super pod"},
+		Interval:    time.Second * 10,
```
andriisoldatenko, line 506: yeah I know it was there before you changed, but anyway better to use `<NUM> * time.Second` style, it's more readable IMO.

Example https://pkg.go.dev/time#example-After

cc @0sewa0 wdyt if we add it to our guide?

chrismuellner: wasn't able to find an existing rule, most similar seems to be for [order of operands in ifs](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#optimize-operands-order)

waodim (commented):
I tested it and can confirm that it works, so I would approve if the conversation is finished by all sides @chrismuellner @0sewa0 @andriisoldatenko

waodim (approved):

PR #2156, opened 2023-09-28
Improve oneAgent image validation using manifest labels

waodim:
## Description

With this validation of the `oneAgent` image is improved. Previously the only verification that was done was by determining if there was a downgrade for the `oneAgent` or not (downgrades are not allowed). Now a new label is taken into consideration (`com.dynatrace.type`). With this we can check whether an image is immutable or not.

If the agent is immutable:
- `classicFullStack` is not allowed
- `feature.dynatrace.com/oneagent-readonly-host-fs` is not allowed

If the `com.dynatrace.type` label is missing:
- for now no error is thrown

If the `com.dynatrace.build-version` label is missing:
- throw an error

## How can this be tested?

~Deploy the operator and apply a `dynakube` (`cloudNative` or `classicFullstack`):~
~- when using the standard image everything should work normally as it a mutable image.~
~- when using the immutable image either use the feature flag (`feature.dynatrace.com/oneagent-readonly-host-fs`) or use `classicFullStack` mode: the `oneAgents` should be in an error state.~

~NOTE: For now this is not testable as there are problems with public-registry. Will be updated as soon as there is a solution.~

It should be testable in the internal e2e pipelines 😉 

## Checklist


waodim:
Still work in progress

src/controllers/dynakube/version/updater.go
```diff
@@ -179,6 +190,10 @@ func updateVersionStatusForTenantRegistry( //nolint:revive
 	return nil
 }
 
+func isOneAgentImage(imageID string) bool {
+	return strings.Contains(imageID, "oneagent")
+}
```
0sewa0, line 195: that is rather a weak check also the "design" of this was meant have these specific checks in each `Updaters` (the oneagent `src/controllers/dynakube/version/oneagent.go`)

0sewa0 (changes requested):

src/controllers/dynakube/version/updater.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/2156/commits/72e5f15c7b6a88a9ca8a6cc78498154bd11d4244

src/controllers/dynakube/version/updater.go
```diff
@@ -24,6 +24,7 @@ type versionStatusUpdater interface {
 	IsAutoUpdateEnabled() bool
 	IsPublicRegistryEnabled() bool
 	CheckForDowngrade(latestVersion string) (bool, error)
+	CheckLabels() error
```
0sewa0, line 27: There are few problems:
- I don't see you calling it anywhere (other then tests)
- Name is not the greatest as it could be confused for checking Kubernetes labels
     - Furthermore you are not reaally checking the image-labels, you are checking the status fields that are based on the image label, which makes it extra confusing
        - This also brings up an interesting problem, should we let the status have "incorrect" data in it, AFAIK we currently only write stuff to the status when its ok 🤷  
- `updater.Target().Version` is ONLY set in case of tenant registry, however we would like to enforce it in all cases for the OneAgent (and probably the ActiveGate aswell)
    - especially when your check is mainly for the public-registry or custom-image use-case

0sewa0 (dismissed):

src/controllers/dynakube/version/updater.go
waodim: Thank you for the feedback, I will adapt my PR to fulfill your points! :slightly_smiling_face:

waodim: Regarding:
> This also brings up an interesting problem, should we let the status have "incorrect" data in it, AFAIK we currently only write stuff to the status when its ok 🤷

I mean if we can guarantee that errors/incorrect data are caught beforehand (because they will only be written to the status if it is correct) I do not really see the point to do another round of error handling/checking for incorrect data. What is your take on that? If I understand that correctly than I do not need to check if the `Version` field is empty

0sewa0: you need to check IF the version label is missing from the image

the problem with `updater.Target().Version` is that it not necessarily from the image label
1. in case of codemodules + tenant-registry (so default case): its what the Dynatrace API gives you for the latest version
2. in case of public-registry and custom-image it's not even set,  [see](https://github.com/Dynatrace/dynatrace-operator/blob/dd188937761c2d422694f3cd99b7fb926e0f872d/src/controllers/dynakube/version/updater.go#L142)
3. in case of oneagent/activegate + tenant-registry (so default case): **it actually uses the image-label**
4. in case of codemodules + public-registry: I do not know if those images even have these version-labels, so that would require more looking into
And we (currently) only want to enforce the version-label on the image in case of 2. and 3., as we know that does should have it

also about " AFAIK we currently only write stuff to the status when its ok", I do not know if this is the case, pls check if I am right or not. 
But it could be a valid (and simpler) solution to write stuff into the status, then verify it if its correct or not separately.
- The reason for this is on the "lower-level", so when we are actually checking for the version label, we do not know if its supposed to be there or not 😢 
   - SO, our main problem is that there is almost no consistency in how each case need to be handled, yet we have an interface that tries its best to enforce consistency which results in this mess 😬

waodim: I get your point, so basically in the scope of the referring ticket I need to improve image validation of the `oneagent` image.

which means:
1. if we are in the case of tenant-registry: I ***can*** use `updater.Target()`
2. if we ware in the case of public-registry: I ***can not*** use `updater.Target()` and have to find another way to check the label
3. if we ware in the case of custom image: same as 2.

am I right here?

0sewa0: I might have overcomplicate things, lets not worry about where the version comes from, as long as we have a version we can verify, and we always want  to have a version for certain things.

I had to play with it locally to get my head around the thing, here is a patch of my changes, (linting is complaining 😅 )
```
diff --git a/src/api/status/version.go b/src/api/status/version.go
index c511a3c4..286ebad5 100644
--- a/src/api/status/version.go
+++ b/src/api/status/version.go
@@ -11,6 +11,8 @@ const (
 	CustomImageVersionSource    VersionSource = "custom-image"
 	CustomVersionVersionSource  VersionSource = "custom-version"
 	PublicRegistryVersionSource VersionSource = "public-registry"
+
+	ImmutableImageType = "immutable"
 )
 
 type VersionStatus struct {
diff --git a/src/controllers/dynakube/version/activegate.go b/src/controllers/dynakube/version/activegate.go
index c7a93b98..37f8f033 100644
--- a/src/controllers/dynakube/version/activegate.go
+++ b/src/controllers/dynakube/version/activegate.go
@@ -6,6 +6,7 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/src/api/status"
 	"github.com/Dynatrace/dynatrace-operator/src/dtclient"
+	"github.com/pkg/errors"
 )
 
@@ -71,6 +72,10 @@ func (updater *activeGateUpdater) UseTenantRegistry(ctx context.Context) error {
 	return updateVersionStatusForTenantRegistry(ctx, updater.apiReader, updater.dynakube, updater.Target(), updater.versionFunc, defaultImage)
 }
 
-func (updater activeGateUpdater) CheckLabels() error {
+func (updater activeGateUpdater) ValidateStatus() error {
+	imageVersion := updater.Target().Version
+	if imageVersion == "" {
+		return errors.New("build version of ActiveGate image is not set")
+	}
 	return nil
 }
diff --git a/src/controllers/dynakube/version/codemodules.go b/src/controllers/dynakube/version/codemodules.go
index 3d8510a6..83ec9880 100644
--- a/src/controllers/dynakube/version/codemodules.go
+++ b/src/controllers/dynakube/version/codemodules.go
@@ -82,6 +82,6 @@ func (updater *codeModulesUpdater) UseTenantRegistry(_ context.Context) error {
 	return nil
 }
 
-func (updater codeModulesUpdater) CheckLabels() error {
+func (updater codeModulesUpdater) ValidateStatus() error {
 	return nil
 }
diff --git a/src/controllers/dynakube/version/oneagent.go b/src/controllers/dynakube/version/oneagent.go
index ee4b694a..901366f1 100644
--- a/src/controllers/dynakube/version/oneagent.go
+++ b/src/controllers/dynakube/version/oneagent.go
@@ -102,18 +102,18 @@ func (updater *oneAgentUpdater) CheckForDowngrade(latestVersion string) (bool, e
 	return isDowngrade(updater.Name(), previousVersion, latestVersion)
 }
 
-func (updater oneAgentUpdater) CheckLabels() error {
+func (updater oneAgentUpdater) ValidateStatus() error {
 	imageVersion := updater.Target().Version
 	imageType := updater.Target().Type
 
 	if imageVersion == "" {
-		return errors.New("build version label of oneAgent image not set")
+		return errors.New("build version of OneAgent image is not set")
 	}
 
-	if imageType == "immutable" && (updater.dynakube.FeatureDisableReadOnlyOneAgent() || updater.dynakube.ClassicFullStackMode()) {
-		return errors.New("immutable oneAgent image in combination with classicFullStack mode or readOnly oneAgent filesystem is not possible")
+	if imageType == status.ImmutableImageType && (updater.dynakube.FeatureDisableReadOnlyOneAgent() || updater.dynakube.ClassicFullStackMode()) {
+		return errors.New("immutable OneAgent image in combination with classicFullStack mode or readOnly OneAgent filesystem is not possible")
 	}
-	log.Info("successfully validated oneAgent image")
+	log.Info("successfully validated OneAgent image")
 
 	return nil
 }
diff --git a/src/controllers/dynakube/version/oneagent_test.go b/src/controllers/dynakube/version/oneagent_test.go
index a2697cfb..dba21c17 100644
--- a/src/controllers/dynakube/version/oneagent_test.go
+++ b/src/controllers/dynakube/version/oneagent_test.go
@@ -240,7 +240,7 @@ func TestCheckLabels(t *testing.T) {
 		dynakube := newDynakubeForCheckLabelTest(versionStatus)
 		dynakube.Spec.OneAgent.CloudNativeFullStack = &dynatracev1beta1.CloudNativeFullStackSpec{}
 		updater := newOneAgentUpdater(dynakube, fake.NewClient(), nil, nil)
-		require.NoError(t, updater.CheckLabels())
+		require.NoError(t, updater.ValidateStatus())
 	})
 	t.Run("Validate immutable oneAgent image with cloudNative and read only oneAgent feature flag", func(t *testing.T) {
 		dynakube := newDynakubeForCheckLabelTest(versionStatus)
@@ -249,20 +249,20 @@ func TestCheckLabels(t *testing.T) {
 			dynatracev1beta1.AnnotationFeatureReadOnlyOneAgent: "false",
 		}
 		updater := newOneAgentUpdater(dynakube, fake.NewClient(), nil, nil)
-		require.Error(t, updater.CheckLabels())
+		require.Error(t, updater.ValidateStatus())
 	})
 	t.Run("Validate immutable oneAgent image with classicFullStack", func(t *testing.T) {
 		dynakube := newDynakubeForCheckLabelTest(versionStatus)
 		dynakube.Spec.OneAgent.ClassicFullStack = &dynatracev1beta1.HostInjectSpec{}
 		updater := newOneAgentUpdater(dynakube, fake.NewClient(), nil, nil)
-		require.Error(t, updater.CheckLabels())
+		require.Error(t, updater.ValidateStatus())
 	})
 	t.Run("Validate immutable oneAgent image when image version is not set", func(t *testing.T) {
 		dynakube := newDynakubeForCheckLabelTest(versionStatus)
 		dynakube.Spec.OneAgent.CloudNativeFullStack = &dynatracev1beta1.CloudNativeFullStackSpec{}
 		dynakube.Status.OneAgent.VersionStatus.Version = ""
 		updater := newOneAgentUpdater(dynakube, fake.NewClient(), nil, nil)
-		require.Error(t, updater.CheckLabels())
+		require.Error(t, updater.ValidateStatus())
 	})
 	t.Run("Validate mutable oneAgent image with cloudNative and read only oneAgent feature flag", func(t *testing.T) {
 		dynakube := newDynakubeForCheckLabelTest(versionStatus)
@@ -272,13 +272,13 @@ func TestCheckLabels(t *testing.T) {
 		}
 		dynakube.Status.OneAgent.VersionStatus.Type = "mutable"
 		updater := newOneAgentUpdater(dynakube, fake.NewClient(), nil, nil)
-		require.NoError(t, updater.CheckLabels())
+		require.NoError(t, updater.ValidateStatus())
 	})
 	t.Run("Validate mutable oneAgent image with classicFullStack", func(t *testing.T) {
 		dynakube := newDynakubeForCheckLabelTest(versionStatus)
 		dynakube.Spec.OneAgent.ClassicFullStack = &dynatracev1beta1.HostInjectSpec{}
 		dynakube.Status.OneAgent.VersionStatus.Type = "mutable"
 		updater := newOneAgentUpdater(dynakube, fake.NewClient(), nil, nil)
-		require.NoError(t, updater.CheckLabels())
+		require.NoError(t, updater.ValidateStatus())
 	})
 }
diff --git a/src/controllers/dynakube/version/synthetic.go b/src/controllers/dynakube/version/synthetic.go
index f562d57a..a143a310 100644
--- a/src/controllers/dynakube/version/synthetic.go
+++ b/src/controllers/dynakube/version/synthetic.go
@@ -72,6 +72,6 @@ func (updater *syntheticUpdater) UseTenantRegistry(ctx context.Context) error {
 	return updateVersionStatusForTenantRegistry(ctx, updater.apiReader, updater.dynakube, updater.Target(), updater.versionFunc, defaultImage)
 }
 
-func (updater syntheticUpdater) CheckLabels() error {
+func (updater syntheticUpdater) ValidateStatus() error {
 	return nil
 }
diff --git a/src/controllers/dynakube/version/updater.go b/src/controllers/dynakube/version/updater.go
index c776c796..7d7fa10a 100644
--- a/src/controllers/dynakube/version/updater.go
+++ b/src/controllers/dynakube/version/updater.go
@@ -24,7 +24,7 @@ type versionStatusUpdater interface {
 	IsAutoUpdateEnabled() bool
 	IsPublicRegistryEnabled() bool
 	CheckForDowngrade(latestVersion string) (bool, error)
-	CheckLabels() error
+	ValidateStatus() error
 	LatestImageInfo() (*dtclient.LatestImageInfo, error)
 
 	UseTenantRegistry(context.Context) error
@@ -81,7 +81,10 @@ func (reconciler *Reconciler) run(ctx context.Context, updater versionStatusUpda
 
 	log.Info("updating version status according to the tenant registry", "updater", updater.Name())
 	err = updater.UseTenantRegistry(ctx)
-	return err
+	if err != nil {
+		return err
+	}
+	return updater.ValidateStatus()
 }
 
 func determineSource(updater versionStatusUpdater) status.VersionSource {
diff --git a/src/controllers/dynakube/version/updater_test.go b/src/controllers/dynakube/version/updater_test.go
index 8f1a6a8a..af1cd70c 100644
--- a/src/controllers/dynakube/version/updater_test.go
+++ b/src/controllers/dynakube/version/updater_test.go
@@ -64,7 +64,7 @@ func (m *mockUpdater) CheckForDowngrade(latestVersion string) (bool, error) {
 	return args.Bool(0), args.Error(1)
 }
 
-func (m *mockUpdater) CheckLabels() error {
+func (m *mockUpdater) ValidateStatus() error {
 	args := m.Called()
 	return args.Error(0)
 }

```

waodim: Thanks I see it as my fault, I lost some time by wrapping my head around the whole concept, but now I'm getting it to a certain extend :+1:. I am honestly glad that my changes where similar to yours after understanding the concept of the updater and what exactly i need to change. :slightly_smiling_face:

waodim: updated in https://github.com/Dynatrace/dynatrace-operator/pull/2156/commits/487df27538ca33c8f018dbd03d6d6d7c63713446

0sewa0: > Thanks I see it as my fault, I lost some time by wrapping my head around the whole concept

Well, I had to also check it out and poke at it to understand what the hell is happening, eventhough it was me who wrote this spaghetti 😓 
Also my explanation was a biiiit confusing and focused on totally the wrong thing, now that I read it back after a few days. 😅

src/controllers/dynakube/version/updater.go
```diff
@@ -130,16 +146,14 @@ func setImageIDWithDigest( //nolint:revive
 				target.ImageID = canonRef.String()
 			}
 		}
+		target.Version = imageVersion.Version
```
0sewa0, line 149: if you don't set the `target.Version` in the `canonRef, ok := ref.(reference.Canonical); ok` case,  then you will error (this can happen in the `customImage` case)

you can just set the version to the digest in that case, which is a bit scary as `CheckForDowngrade` would go boom as a digest is not a semantic version, but we don't do that in the case of `customImage` so it should be fine 😬

0sewa0 (changes requested):

src/controllers/dynakube/version/updater.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/2156/commits/5c6a5fff8d2dfd2c794118618de21306aaf2f9e7

0sewa0 (dismissed):

andriisoldatenko:
@waodim BTW when I run it locally I see some missing changes from you PR:
```
        modified:   config/crd/bases/dynatrace.com_dynakubes.yaml
        modified:   config/crd/bases/dynatrace.com_edgeconnects.yaml
```

andriisoldatenko:
> @waodim BTW when I run it locally I see some missing changes from you PR:
> 
> ```
>         modified:   config/crd/bases/dynatrace.com_dynakubes.yaml
>         modified:   config/crd/bases/dynatrace.com_edgeconnects.yaml
>         modified:   config/helm/chart/default/templates/Common/crd/dynatrace-operator-crd.yaml
> ```

confirmed!

andriisoldatenko (dismissed):

pkg/controllers/dynakube/version/oneagent.go
```diff
@@ -101,3 +102,19 @@ func (updater *oneAgentUpdater) CheckForDowngrade(latestVersion string) (bool, e
...
+	imageType := updater.Target().Type
+
+	if imageVersion == "" {
+		return errors.New("build version of OneAgent image not set")
+	}
+
+	if imageType == status.ImmutableImageType && (updater.dynakube.FeatureDisableReadOnlyOneAgent() || updater.dynakube.ClassicFullStackMode()) {
+		return errors.New("immutable OneAgent image in combination with classicFullStack mode or readOnly OneAgent filesystem is not possible")
+	}
+	log.Info("successfully validated OneAgent image")
```
chrismuellner, line 117:
```suggestion
	log.Info("OneAgent metadata present")
```
or some other way to inform the user on what was validated

pkg/controllers/dynakube/version/oneagent.go
```diff
@@ -101,3 +102,19 @@ func (updater *oneAgentUpdater) CheckForDowngrade(latestVersion string) (bool, e
...
+func (updater oneAgentUpdater) ValidateStatus() error {
+	imageVersion := updater.Target().Version
+	imageType := updater.Target().Type
+
+	if imageVersion == "" {
+		return errors.New("build version of OneAgent image not set")
+	}
+
+	if imageType == status.ImmutableImageType && (updater.dynakube.FeatureDisableReadOnlyOneAgent() || updater.dynakube.ClassicFullStackMode()) {
+		return errors.New("immutable OneAgent image in combination with classicFullStack mode or readOnly OneAgent filesystem is not possible")
```
chrismuellner, line 115: this should be separate errors, combining them could be confusing

chrismuellner (changes requested):

pkg/controllers/dynakube/version/oneagent.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/2156/commits/27899211fb4112ad1045ccc88b8e491f28168227

pkg/controllers/dynakube/version/oneagent.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/2156/commits/27899211fb4112ad1045ccc88b8e491f28168227

andriisoldatenko (approved):
Code LGTM, tested in my cluster without labels and it works ( basic cloudNativeStack).

PR #2158, opened 2023-09-29
Fix operator behavior related to `503 ServiceUnavailable` DynatraceApi error

aorcholski:
## Description
The Operator doesn't handle `ServiceUnavailable` and `TooManyRequests` DynatraceApi errors properly thus Dynakube Reconciler in not called as often as it should be in such a case (after a minute).

The root cause is  `concatErrors(errs []error)` function which concatenates error messages into a single string error so the information that `ServerError` has been encountered is lost.

The fixed logic returns an instance of ServerError if any of concatenated errors was ServerError. The disadvantage of this solution is that part of the final message is duplicated:
`dynatrace server error 503: error 1\n\tdynatrace server error 503: ServiceUnavailable`
It would be better to report:
`error 1\n\tdynatrace server error 503: ServiceUnavailable`

Finally `isDynatraceAPIUnreachable(err error)` function is able to recognize DynatraceApi errors and handle them properly.

## How can this be tested?
We can't force the server to return 503 so I tested the fix using unit tests and debugged a small application consisting of investigated code and basic http server returning 503 response.

## Checklist

### DIFF
--- a/src/controllers/dynakube/dynakube_controller.go
+++ b/src/controllers/dynakube/dynakube_controller.go
@@ -127,16 +127,6 @@ func (controller *Controller) setRequeueAfterIfNewIsShorter(requeueAfter time.Du
 	}
 }
 
-func isDynatraceAPIUnreachable(err error) bool {
-	var serverErr dtclient.ServerError
-	if dynatraceapi.IsUnreachable(err) {
-		log.Info("dynaTrace API server is unavailable or request limit reached! trying again in one minute",
-			"errorCode", serverErr.Code, "errorMessage", serverErr.Message)
-		return true
-	}
-	return false
-}
-
 func (controller *Controller) reconcile(ctx context.Context, dynaKube *dynatracev1beta1.DynaKube) (reconcile.Result, error) {
 	oldStatus := *dynaKube.Status.DeepCopy()
 
@@ -145,7 +135,9 @@ func (controller *Controller) reconcile(ctx context.Context, dynaKube *dynatrace
 	err := controller.reconcileDynaKube(ctx, dynaKube)
 
 	switch {
-	case isDynatraceAPIUnreachable(err):
+	case dynatraceapi.IsUnreachable(err):
+		log.Info("dynaTrace API server is unavailable or request limit reached! trying again in one minute",
+			"errorCode", dynatraceapi.StatusCode(err), "errorMessage", dynatraceapi.Message(err))
 		// should we set the phase to error ?
 		return reconcile.Result{RequeueAfter: fastUpdateInterval}, nil
 
--- a/src/controllers/dynakube/dynakube_controller.go
+++ b/src/controllers/dynakube/dynakube_controller.go
@@ -2,7 +2,6 @@ package dynakube
 
 import (
 	"context"
-	"net/http"
 	"os"
 	"time"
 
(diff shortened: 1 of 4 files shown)


src/controllers/dynakube/token/tokens.go
```diff
@@ -98,14 +99,25 @@ func (tokens Tokens) VerifyValues() error {
 
 func concatErrors(errs []error) error {
 	concatenatedError := ""
+	apiError := 0
```
StefanHauth, line 102: Call it `apiStatus`, wrap 0 in a const like `NoApiError` to give it a meaningful name and reuse it later for checks

src/controllers/dynakube/token/tokens_test.go
```diff
@@ -161,3 +162,59 @@ func testVerifyTokenValues(t *testing.T) {
 	assert.NoError(t, validTokens.VerifyValues())
 	assert.EqualError(t, invalidTokens.VerifyValues(), "value of token 'whitespaces' contains whitespaces at the beginning or end of the value")
 }
+
+func TestConcatErrors(t *testing.T) {
```
StefanHauth, line 166: Nitpick: looks like it could be a table test.

StefanHauth (changes requested):

src/controllers/dynakube/token/tokens.go
aorcholski: done

src/controllers/dynakube/token/tokens_test.go
aorcholski: agree, done

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -129,7 +129,7 @@ func (controller *Controller) setRequeueAfterIfNewIsShorter(requeueAfter time.Du
 
 func isDynatraceAPIUnreachable(err error) bool {
 	var serverErr dtclient.ServerError
-	if errors.As(err, &serverErr) && (serverErr.Code == http.StatusTooManyRequests || serverErr.Code == http.StatusServiceUnavailable) {
+	if dynatraceapi.IsUnreachable(err) {
```
andriisoldatenko, line 132: any reason why we need to warp the online function into another one? Should we move the log to case and make only 1 function instead of two?

andriisoldatenko: i mean:
```
switch {
	case dynatraceapi.IsUnreachable(err) (err):
		log.Info("dynaTrace API server is unavailable or request limit reached! trying again in one minute",
			"errorCode", serverErr.Code, "errorMessage", serverErr.Message)
		return reconcile.Result{RequeueAfter: fastUpdateInterval}, nil
```

aorcholski: Just to keep the (condition) logic in one place.

andriisoldatenko (approved):

src/controllers/dynakube/dynakube_controller.go
```diff
@@ -145,7 +135,9 @@ func (controller *Controller) reconcile(ctx context.Context, dynaKube *dynatrace
 	err := controller.reconcileDynaKube(ctx, dynaKube)
```
chrismuellner, line 135: line 163 (sorry can't comment on it directly) should never return both a result and an error

chrismuellner (changes requested):

src/controllers/dynakube/dynakube_controller.go
andriisoldatenko: @chrismuellner you mean this one 

https://github.com/Dynatrace/dynatrace-operator/blob/1ad499de870525707d6b2f4e1acdae6a003f8e0a/src/controllers/dynakube/dynakube_controller.go#L173

StefanHauth (approved):

src/controllers/dynakube/dynakube_controller.go
aorcholski: either a result or an error is returned

aorcholski: sorry, I mean either non-empty result + nil error  or empty result  + error

chrismuellner (approved):

andriisoldatenko (approved):

StefanHauth (approved):

PR #2159, opened 2023-09-29
Replace containers/image library for image reference parsing.

StefanHauth:
## Description

We currently use the containers/image library just for reference parsing. We can do this with the new image lib resp. on our own. This PR is about evicting the containers/image library completely

## How can this be tested?

Run unit tests

## Checklist


src/controllers/dynakube/version/updater.go
```diff
@@ -143,6 +147,11 @@ func setImageIDWithDigest( //nolint:revive
 	return nil
 }
 
+// TODO: this should be in a more general place
```
StefanHauth, line 150: I'm open for suggestions where to put this function.

src/controllers/dynakube/version/image_hash_test.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/src/controllers/dynakube/version/testdata"
 	"github.com/Dynatrace/dynatrace-operator/src/registry"
 	"github.com/Dynatrace/dynatrace-operator/src/registry/mocks"
-	"github.com/containers/image/v5/docker/reference"
```
andriisoldatenko, line 13: @StefanHauth could you please remove it from go.mod too?

 https://github.com/Dynatrace/dynatrace-operator/blob/5422bcfc70d52e33872070b7c8cb8dc4ecbd541e/go.mod#L7

src/controllers/dynakube/version/updater.go
```diff
@@ -113,23 +119,21 @@ func setImageIDWithDigest( //nolint:revive
...
 		"oldImageID", target.ImageID)
 
-	if canonRef, ok := ref.(reference.Canonical); ok {
-		target.ImageID = canonRef.String()
-	} else if taggedRef, ok := ref.(reference.NamedTagged); ok {
+	if digestRef, ok := ref.(name.Digest); ok {
+		target.ImageID = digestRef.String()
+	} else if taggedRef, ok := ref.(name.Tag); ok {
+		if taggedRef.TagStr() == name.DefaultTag {
+			return errors.New(fmt.Sprintf("unsupported image reference: %s", imageUri))
```
andriisoldatenko, line 126:
```suggestion
			return errors.Errorf("unsupported image reference: %s", imageUri)
```

src/controllers/dynakube/version/updater.go
andriisoldatenko: I would like to rename it, cause append is something when we add to exiting, here we join or concat strings.

src/controllers/dynakube/version/updater.go
```diff
@@ -180,15 +189,24 @@ func updateVersionStatusForTenantRegistry( //nolint:revive
...
-	ref, err := reference.Parse(imageID)
+	ref, err := name.ParseReference(imageID)
 	if err != nil {
 		return "", err
 	}
-	taggedRef, ok := ref.(reference.NamedTagged)
-	if !ok {
-		return "", errors.New("no tag found to check for downgrade")
+
+	var taggedRef name.Tag
```
andriisoldatenko, line 197: I think you can remove this declaration since we're not using it anymore in return.

andriisoldatenko (dismissed):

src/controllers/dynakube/version/updater.go
0sewa0: > I'm open for suggestions where to put this function.
maaybe `src/registry` ?

src/controllers/dynakube/version/updater.go
```diff
@@ -180,15 +189,21 @@ func updateVersionStatusForTenantRegistry( //nolint:revive
...
+
+	if taggedRef, ok := ref.(name.Tag); ok {
+		return taggedRef.TagStr(), nil
+	} else if digestRef, ok := ref.(name.Digest); ok {
+		taggedStr := strings.TrimSuffix(digestRef.String(), digestDelimiter+digestRef.DigestStr())
+
+		if taggedRef, err = name.NewTag(taggedStr); err == nil {
+			return taggedRef.TagStr(), nil
+		}
 	}
```
0sewa0, line 205: what are we doing here ? and why? (we didn't use to care about digest here)

0sewa0 (changes requested):

src/controllers/dynakube/version/updater.go
StefanHauth: Removing the digest from input strings like `my.reg/rep:1.2.3@sha256:abcdcoffee`

0sewa0: so in case `my.reg/rep:1.2.3@sha256:abcdcoffee` then  `_, ok := ref.(name.Tag)` results in `ok == false`?
if that is the case, then I am not angry, just disappointed in the library 😓 

otherwise it does not make sense

0sewa0: just debuged the tests, and yes it does that 😭

0sewa0: also during the debugging the tests (nice tests btw ⭐ ), in the case of `some.registry.com@sha256:7ece13a07a20c77a31cc36906a10ebc90bd47970905ee61e8ed491b7f4c5d62f` (so when only digest given), then we should return the error, as we don't know the tag. It shouldn't return `latest`, as we couldn't know if a given digest IS or not latest.

we only use this function when dealing with the PublicRegsitry feature, where we will always get a tag and not a digest, so we should error if that behaviour changes.

StefanHauth: That's another caveat of the new lib, that it adds `latest` as a default fallback without asking. I'll have a closer look how to make this happen.

StefanHauth: Fixed this and updated unit tests accordingly.

0sewa0 (approved):

PR #2167, opened 2023-10-03
Prevent duplicates from oneAgent connection info resp

andriisoldatenko:
## Description

Please include the following:

- A meaningful PR title [guidelines](https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#use-imperative-mood-in-your-commit-message-subject)
- A summary of the change
- Which issue is fixed (if there is one)
- Relevant motivation and context

## How can this be tested?

Please include some guiding steps on how to test this change during a review.

- What environment is necessary for the change to be noticeable ?
- What needs to be enabled/applied for the change to be noticeable ?

## Checklist

- [ ] Unit tests have been updated/added
- [ ] PR is labeled accordingly with a single label
- [ ] I have read and understood the [contribution guidelines](https://github.com/Dynatrace/dynatrace-operator/blob/main/CONTRIBUTING.md)

PR #2168, opened 2023-10-03
Adds ActiveGate endpoints as Istio ServiceEntries

wepudt:
## Description

- adds ActiveGate endpoints as well as OneAgent communication hosts as Istio service entries

## How can this be tested?

- install istio
- deploy dynakube
- check serviceentries: should contain oneagent and activegate endpoints

## Checklist


src/controllers/dynakube/connectioninfo/communication_hosts.go
```diff
@@ -16,3 +21,59 @@ func GetOneAgentCommunicationHosts(dynakube *dynatracev1beta1.DynaKube) []dtclie
...
+
+func parseCommunicationHostFromActiveGateEndpoints(activegateEndpointsString string) []dtclient.CommunicationHost {
+	endpointStrings := strings.Split(activegateEndpointsString, ";")
+	endpointCommunicationHosts := make([]dtclient.CommunicationHost, 0)
+	for _, endpointString := range endpointStrings {
+		endpoint, err := parseEndpointToCommunicationHost(endpointString)
+		if err == nil {
+			endpointCommunicationHosts = append(endpointCommunicationHosts, endpoint)
+		}
+	}
```
0sewa0, line 41: be aware that the endpoints might have duplicates in them, which is not allowed in `VirtualServices`
here is the recent bugfix for it: https://github.com/Dynatrace/dynatrace-operator/pull/2167
- that fix was aimed at the OneAgent Status as we store the OneAgent `CommunicationHosts` and just use them in the istio reconciler
- we don't do such a thing for the ActiveGate Status, so you have to do it here, and in the future we can consider to also store the `CommunicationHosts` of the ActiveGate in the Status

0sewa0 (changes requested):

src/controllers/dynakube/connectioninfo/communication_hosts.go
wepudt: done

src/controllers/dynakube/istio/reconciler.go
```diff
@@ -41,22 +41,60 @@ func (r *Reconciler) ReconcileAPIUrl(ctx context.Context, dynakube *dynatracev1b
...
+func (r *Reconciler) reconcileCommunicationHostsForComponent(ctx context.Context, dynakube *dynatracev1beta1.DynaKube, comHosts []dtclient.CommunicationHost, componentName string) error {
+	err := r.reconcileCommunicationHosts(ctx, dynakube, comHosts, componentName)
+	if err != nil {
+		return errors.WithMessage(err, "error reconciling config for Dynatrace communication hosts")
+	}
+	log.Info("reconciled istio objects for communication hosts", "component", componentName)
 	return nil
 }
 
+func mergeCommunicationHosts(oneAgentCommunicationHosts, activeGateEndpoints []dtclient.CommunicationHost) []dtclient.CommunicationHost {
```
0sewa0, line 73: this is not used anymore, as we create istio resources separately for each component

0sewa0 (changes requested):
tested it and something weird is happening for me
I have 4 endpoints in the `dynakube.Status.ActiveGate.ConnectionInfoStatus`
But only 1 of them are in the activegate `ServiceEntry`

src/controllers/dynakube/istio/reconciler.go
wepudt: true- done

src/controllers/dynakube/connectioninfo/communication_hosts.go
```diff
@@ -16,3 +21,64 @@ func GetOneAgentCommunicationHosts(dynakube *dynatracev1beta1.DynaKube) []dtclie
...
+	activegateEndpointsString := dynakube.Status.ActiveGate.ConnectionInfoStatus.Endpoints
+	if activegateEndpointsString == "" {
+		return []dtclient.CommunicationHost{}
+	}
+	return parseCommunicationHostFromActiveGateEndpoints(activegateEndpointsString)
+}
+
+func parseCommunicationHostFromActiveGateEndpoints(activegateEndpointsString string) []dtclient.CommunicationHost {
+	endpointStrings := strings.Split(activegateEndpointsString, ";")
+	endpointCommunicationHosts := make(map[dtclient.CommunicationHost]bool)
```
andriisoldatenko, line 35:
```suggestion
	endpointCommunicationHosts := make(map[dtclient.CommunicationHost]bool, len(endpointStrings))
```

src/controllers/dynakube/connectioninfo/communication_hosts.go
```diff
@@ -16,3 +21,64 @@ func GetOneAgentCommunicationHosts(dynakube *dynatracev1beta1.DynaKube) []dtclie
...
+	for _, endpointString := range endpointStrings {
+		endpoint, err := parseEndpointToCommunicationHost(endpointString)
+		if err == nil {
+			endpointCommunicationHosts[endpoint] = true
+		}
```
andriisoldatenko, line 40:
```suggestion
		if err != nil {
			continue
		}
		endpointCommunicationHosts[endpoint] = true
```

andriisoldatenko: or even in one line:
```
if endpoint, err := parseEndpointToCommunicationHost(endpointString); err != nil {
    endpointCommunicationHosts[endpoint] = true
}
```

src/controllers/dynakube/connectioninfo/communication_hosts.go
```diff
@@ -16,3 +21,64 @@ func GetOneAgentCommunicationHosts(dynakube *dynatracev1beta1.DynaKube) []dtclie
...
+
+func parsePort(hostParts []string) (uint32, error) {
+	if len(hostParts) > 1 {
+		intPort, err := strconv.ParseUint(hostParts[1], 10, 32)
+		if err != nil {
+			return 0, err
+		}
+		return uint32(intPort), nil
+	} else {
+		return uint32(80), nil
```
andriisoldatenko, line 82: use namedConstant here please

src/controllers/dynakube/connectioninfo/communication_hosts.go
```diff
@@ -1,8 +1,17 @@
...
+	"strconv"
+	"strings"
+
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/src/dtclient"
+	"github.com/pkg/errors"
+)
+
+const (
+	DefaultHostPort = uint32(80)
```
andriisoldatenko, line 14:
```suggestion
	DefaultHttpPort = uint32(80)
```

src/controllers/dynakube/connectioninfo/communication_hosts.go
```diff
@@ -16,3 +25,63 @@ func GetOneAgentCommunicationHosts(dynakube *dynatracev1beta1.DynaKube) []dtclie
...
+	}
+	parsedEndpoint, err := url.Parse(endpointString)
+
+	if err != nil {
+		return dtclient.CommunicationHost{}, err
+	}
+	host := parsedEndpoint.Host
+	parts := strings.Split(host, ":")
+
+	port, err := parsePort(parts)
```
andriisoldatenko, line 66: I think you can reuse standart library: `net.SplitHostPort` https://pkg.go.dev/net#SplitHostPort and remove parsePort function.

wepudt: that looks rather nice

src/controllers/dynakube/connectioninfo/communication_hosts.go
```diff
@@ -16,3 +22,56 @@ func GetOneAgentCommunicationHosts(dynakube *dynatracev1beta1.DynaKube) []dtclie
...
+	return parseCommunicationHostFromActiveGateEndpoints(activegateEndpointsString)
+}
+
+func parseCommunicationHostFromActiveGateEndpoints(activegateEndpointsString string) []dtclient.CommunicationHost {
+	endpointStrings := strings.Split(activegateEndpointsString, ",")
+	endpointCommunicationHosts := make(map[dtclient.CommunicationHost]bool, len(endpointStrings))
+	for _, endpointString := range endpointStrings {
+		if endpoint, err := parseEndpointToCommunicationHost(endpointString); err == nil {
+			endpointCommunicationHosts[endpoint] = true
+		}
```
0sewa0, line 40: you are secretly not handling the error here, and that is why its not working

`hostPart, portPart, err := net.SplitHostPort(host)` this call will fail if the endpoint has no port and cause an error

andriisoldatenko: yeah good catch ^

andriisoldatenko: @0sewa0  / @wepudt wdyt about such version?

Instead of parsing (as I suggested wrongly) , we can use `func (*URL) Port` and func (*URL) Hostname`

- Hostname returns u.Host, stripping any valid port number if present.
- Port returns the port part of u.Host, without the leading colon.

```
func parseEndpointToCommunicationHost(endpointString string) (dtclient.CommunicationHost, error) {
	if endpointString == "" {
		return dtclient.CommunicationHost{}, errors.New("empty endpoint string not allowed")
	}
	parsedEndpoint, err := url.Parse(endpointString)

	if err != nil {
		return dtclient.CommunicationHost{}, err
	}
	port := parsedEndpoint.Port()
	
	if port == "" {
		port = "80"
	}

	if err != nil {
		return dtclient.CommunicationHost{}, err
	}

	intPort, err := strconv.ParseUint(port, 10, 32)
	if err != nil {
		return dtclient.CommunicationHost{}, err
	}

	return dtclient.CommunicationHost{
		Protocol: parsedEndpoint.Scheme,
		Host:     parsedEndpoint.Hostname(),
		Port:     intPort,
	}, nil
}
```

wepudt: good input, thanks you two- submitted a solution including your suggestions

0sewa0 (approved):

andriisoldatenko (approved):

PR #2172, opened 2023-10-04
E2E tests for network zone support

StefanHauth:
## Description

The new E2E test
* creates a network zone with "FallbackMode = None"
* creates a DynaKube without ActiveGate in that network zone
* checks if OneAgent rollout and code modules injection is blocked
* enables ActiveGate in the Dynakube
* checks if OneAgents start
* checks if code modules get injected into app pods (after restart)
* deletes network zone again

## How can this be tested?

Run `make test/e2e/cloudnative/networkzones`
Note: the API token used for running the test needs to have the following scopes:
* `networkZones.read`
* `networkZones.write`

## Checklist


test/helpers/tenant/network_zone.go
```diff
@@ -0,0 +1,108 @@
...
+
+		client := &http.Client{}
+		req, err := http.NewRequest(http.MethodPut, nzApiUrl, bytes.NewReader(body))
+		require.NoError(t, err)
+
+		req.Header.Add("Authorization", "Api-Token "+secret.ApiToken)
+		req.Header.Add("Content-Type", "application/json")
+
+		resp, err := client.Do(req)
+		defer func() { _ = resp.Body.Close() }()
```
aorcholski, line 58: `Do` may return an error (`nil` response) so we should swap lines 58 and 59 in place or check if `resp` is not `nil` in the `defer` statement

waodim (approved):
Code LGTM - ran test several times and can confirm it worked :+1:

PR #2186, opened 2023-10-09
Use image index digest instead of plain image digest

StefanHauth:
## Description

Currently the image digest for AMD64 images is used in OneAgent daemonsets. This causes errors if OneAgents should be rolled out on nodes with architectures other than amd64.

Instead of using the digest of the amd64 image the operator shall use the digest of the image index, that points to an image manifest containing images for all supported architectures.

## How can this be tested?

- setup a K8s cluster with arm and amd nodes
- build and deploy a multi-arch operator
- create a dynakube with proper tolerations for one agent pods:
```
       tolerations:
        - effect: NoSchedule
          key: kubernetes.io/arch
          value: arm64
          operator: Equal
```
- deploy DynaKube
- check that OneAgents can properly start on arm nodes

## Checklist

### DIFF
--- a/pkg/oci/registry/client.go
+++ b/pkg/oci/registry/client.go
@@ -95,7 +95,7 @@ var _ ClientBuilder = NewClient
 func (c *Client) GetImageVersion(ctx context.Context, imageName string) (ImageVersion, error) {
 	ref, err := name.ParseReference(imageName)
 	if err != nil {
-		return ImageVersion{}, fmt.Errorf("parsing reference %q: %w", imageName, err)
+		return ImageVersion{}, errors.WithMessagef(err, "parsing reference %q", imageName)
 	}
 
 	options := []remote.Option{
@@ -108,13 +108,13 @@ func (c *Client) GetImageVersion(ctx context.Context, imageName string) (ImageVe
 
 	descriptor, err := remote.Get(ref, options...)
 	if err != nil {
-		return ImageVersion{}, fmt.Errorf("getting reference %q: %w", ref, err)
+		return ImageVersion{}, errors.WithMessagef(err, "getting reference %q", ref)
 	}
 
 	// TODO: does not work for indexes which contain schema v1 manifests
 	img, err := descriptor.Image()
 	if err != nil {
-		return ImageVersion{}, fmt.Errorf("descriptor.Image(): %w", err)
+		return ImageVersion{}, errors.WithMessagef(err, "descriptor.Image()")
 	}
 
 	// use image digest as a fallback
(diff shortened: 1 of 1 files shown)


pkg/oci/registry/client.go
```diff
@@ -116,10 +116,21 @@ func (c *Client) GetImageVersion(ctx context.Context, imageName string) (ImageVe
 	if err != nil {
 		return ImageVersion{}, fmt.Errorf("descriptor.Image(): %w", err)
 	}
-	dig, err := img.Digest()
+
+	// use image digest as a fallback
+	digestFn := img.Digest
```
luhi-DT, line 121: was it on purpose to not use the Digest() function?

pkg/oci/registry/client.go
```diff
@@ -116,10 +116,21 @@ func (c *Client) GetImageVersion(ctx context.Context, imageName string) (ImageVe
...
+
+	// try to get image manifest to cover multi arch images
+	digestSrc, err := descriptor.ImageIndex()
```
luhi-DT, line 124: you are re-assigning the error variable here that means the error check in 130 does not make sense anymore or am I missing something

pkg/oci/registry/client.go
StefanHauth: I'm using a function pointer here that is later on used to call the Digest() function. Depending on the outcome of line 124, the function pointer might be switched to a different Digest() function. I tried to avoid lengthy if... else constructs here. Alternative implementation:
```
	var dig containerv1.Hash
	// try to get image manifest to cover multi arch images
	if imgIdx, err := descriptor.ImageIndex(); err == nil {
		dig, err = imgIdx.Digest()
		if err != nil {
			return ImageVersion{}, fmt.Errorf("img.Digest(): %w", err)
		}
	} else {
		dig, err = img.Digest()
		if err != nil {
			return ImageVersion{}, fmt.Errorf("img.Digest(): %w", err)
		}
	}
```

pkg/oci/registry/client.go
StefanHauth: The error check in 130 is for the outcome of 129, Or am I understanding something wrong?

0sewa0: we do not care about the error from `descriptor.ImageIndex()`, as that errors (in theory) when not a image-index was given but an actual image (like `snapshot-amd64` instead of just `snapshot`)

the actual error that we care about, and should be propagated,  is what the `digestFn` would return, as that shouldn't error

TLDR: This is fine, BUT pls rename `digestSrc` to `imageIndex` as `digestSrc` is veeery generic 😅

0sewa0 (approved):

PR #2187, opened 2023-10-09
Reorg and rename then `src` package

0sewa0:
## Description
Basically reorgs the whole top level of the codebase, mainly focusing on the `src` folder.

Reasons:
- There are waaaay too many top level packages
- Some of the names are just not intuitive 
	- (😬 giving a very specific package a name that makes sense on the top level is not really possible most of the time)
- We didn't follow some of the conventions. 
	- (🥴 because we thought we knew better )

Changes:
- renames `src` to `pkg`
   -  to conform to golang standarts
- move `cmd` to top level
   -  to conform to golang standarts
- move `src/dockerkeychain` and `src/registry` under `pkg/oci`
   -  as both packages handle dealing with OCI images/registries
- move `src/ingestendpoint`, `src/initgeneration` and `src/mapper` under `pkg/injection/namespace`
   - as all of them handle a certain part of the namespaces for injection.
- move `src/installer` and `src/processmoduleconfig` under `pkg/injection/codemodule`
   - as both handle the the codemodule relevant parts of the injection
- move(rename) `src/standalone` to `pkg/injection/startup`
   -  "standalone" is not really a name that means something, its the logic for the injected init container AND "init" can't be a package name in go (as the `init` function has special meaning)
- rename `src/config` to `pkg/consts`
	- as its a package full of consts (the consts should be further separated into different packages)
- move `src/scheme` under `pkg/api`
	- it just makes sense IMO
- move(rename) `src/dtclient` to `pkg/clients/dynatrace`
	- `dtclient` is just a weird name for the dynatrace client, and we will probably have different clients in the future
- move `src/timeprovider`, `src/builder`, `src/eventfilter`, `src/functional`, `src/kubeobjects`, `src/kubesystem`, `src/logger` and `src/testing` under `pkg/util`
	- as all of these are just some "random" util and does not really belong anywhere
		- there could be deeper reorgs here (in follow up), for example the contents of`kubeobjects` should be further grouped into subpackages, or having an `pkg/internal/types` package for the `builder` package

Disclaimer:
- This PR ONLY touches the  top level packages, and moves packages to the "correct" place. 
There are probably other packages that are deeper in the folder structure that should be also moved/renamed, BUT those should be handled in a follow-up.

## How can this be tested?
DO NOT LOOK AT IT IN THE BROWSER(its not that bad), because it will be SLOOOW.
Check it out locally and test it / look at it there/

## Checklist


wepudt:
lgtm.

ran the unit-tests; looked at your structure and the reasoning 👍

pkg/controllers/dynakube/connectioninfo/communication_hosts.go
```diff
@@ -5,19 +5,19 @@ import (
 	"strconv"
 	"strings"
 
-	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/src/api/v1beta1/dynakube"
-	"github.com/Dynatrace/dynatrace-operator/src/dtclient"
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta1/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
```
aorcholski, line 9: could `dtclient` alias be used here for `.../clients/dynatrace` import (as in many other places) ? The alias gives better context than `dynatrace`

andriisoldatenko: yeah agree `dtClient` alias is okey.

andriisoldatenko (approved):
practically LGTM, tested in my cluster via installing edgeConnect and cloudFullStack samples.

aorcholski (approved):

PR #2190, opened 2023-10-09
Document the use of mockery in contributing.md

wepudt:
## Description

Adds a short overview of mockery, as well as make targets and `.mockery.yaml`.

## How can this be tested?

use mockery according to description in contributing.md

## Checklist


.mockery.yaml
```diff
@@ -0,0 +1,15 @@
+quiet: False
+disable-version-string: True
+with-expecter: True
+mockname: "{{.InterfaceName}}"
+filename: "{{.MockName}}.go"
+outpkg: mocks
+dir: "{{.InterfaceDir}}/mocks"
```
andriisoldatenko, line 7: I prefer to have one global mocks folder, currently we have:
image

.mockery.yaml
```diff
@@ -0,0 +1,15 @@
...
+packages:
+  github.com/Dynatrace/dynatrace-operator/src/builder:
```
andriisoldatenko, line 9: Also, it's not clear why we add only 1 mock to the config, but not others for instance registry mocks.

.mockery.yaml
andriisoldatenko: @0sewa0 it's related to you global rename folders PR ^

0sewa0: in some way yes, but I don't have any major againts having it in 1 or multiple places.
- I am leaning a bit towards the "have it in the same folder as the interface", because its nicer to use IMO
	- However when it comes to external dependencies then its kinda given that those have to be in a dedicated folder

If we want to have it in a single folder, then I would have it in `src/testing/mocks`.

.mockery.yaml
0sewa0: isn't that a followup to do the migration of all the mocks ?

andriisoldatenko: I'm okay of the current state; my problem is that now we have config, but we don't list all existing mocks based on mockery. For instance, registries/mocks - are based on mockery. Also `repo/mocks/*` based on mocker too AFAIK.

andriisoldatenko:
okey may be I miss understand the purpose of the ticket, but I played around locally with this branch and my assumption we should have existing mocks (which we already generated using mockery in config)
```diff
diff --git .mockery.yaml .mockery.yaml
index 79c71c14..4ac95621 100644
--- .mockery.yaml
+++ .mockery.yaml
@@ -6,6 +6,14 @@ filename: "{{.MockName}}.go"
 outpkg: mocks
 dir: "{{.InterfaceDir}}/mocks"
 packages:
+  github.com/Dynatrace/dynatrace-operator/src/registry:
+    interfaces:
+      ImageGetter:
+          config:
+            mockname: "Mock{{.InterfaceName}}"
+            # TODO: refactor rename it to registry.go
+            filename: "client.go"
+            dir: "{{.InterfaceDir}}/mocks"
   github.com/Dynatrace/dynatrace-operator/src/builder:
     config:
       recursive: true

```

hack/make/go.mk
```diff
@@ -32,3 +32,6 @@ go/lint: go/format go/vet go/golangci
 go/test:
 	go test ./... -coverprofile=coverage.txt -covermode=atomic -tags "$(shell ./hack/build/create_go_build_tags.sh false)"
 
+## creates mocks from .mockery.yaml
+go/mock:
```
andriisoldatenko, line 36:
```suggestion
go/gen_mocks:
```

.mockery.yaml
wepudt: not pkg/test/mocks?

andriisoldatenko:
@wepudt looks like you should do rebase 🤔

wepudt:
will redo changes as something messed up my sources

PR #2193, opened 2023-10-10
Adds missing nodeAffinity settings for arm and ppcle support to the activeGate statefulset

luhi-DT:
## Description

Currently the operator fails to roll out ActiveGate instances on ARM nodes due to missing node affinity settings. This PR adds those to the statefulset.

## How can this be tested?

Get a cluster with a different architecture (i.e arm64)
Install the operator
Apply a Dynakube that rolls out an AG
Check if the AG pod is actually scheduled (may fail since the tenant registry only supports amd64 -> thats fine)

## Checklist

### DIFF
--- a/pkg/util/kubeobjects/affinity.go
+++ b/pkg/util/kubeobjects/affinity.go
@@ -23,7 +23,7 @@ func AffinityNodeRequirementForSupportedArches() []corev1.NodeSelectorRequiremen
 }
 
 func tolerationsForArches(arches ...string) []corev1.Toleration {
-	tolerations := make([]corev1.Toleration, len(arches))
+	tolerations := make([]corev1.Toleration, 0)
 	for _, arch := range arches {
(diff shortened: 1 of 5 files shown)


pkg/util/kubeobjects/affinity.go
```diff
@@ -14,23 +14,25 @@ const (
 	linux   = "linux"
 )
 
-func AffinityNodeRequirement() []corev1.NodeSelectorRequirement {
-	return affinityNodeRequirementsForArches(amd64)
+func TolerationForAllArches() []corev1.Toleration {
```
andriisoldatenko, line 17:
```suggestion
func TolerationForSupportedArches() []corev1.Toleration {
```
better to name it more explicitly.

andriisoldatenko (approved):

andriisoldatenko (approved):

PR #2230, opened 2023-10-16
Common sanitization logic for all env var and arguments

StefanHauth:
## Description

Goal of this PR is to introduce a common logic for all env var and argument handling for containers. Basically duplication shall be avoided and arguments/env vars passed by the user always shall take precedence.

## How can this be tested?

Deploy a Dynakube with Active Gate and One Agent, provide custom env var and argument settings and check the daemonset/statefulset.

## Checklist


pkg/util/parametermap/map.go
```diff
@@ -0,0 +1,145 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"golang.org/x/exp/slices"
+	corev1 "k8s.io/api/core/v1"
+)
+
+type argument struct {
```
andriisoldatenko, line 11: is it something like Entry or Element? It's not clear what's argument, more like function argument for me.

pkg/util/parametermap/map.go
```diff
@@ -0,0 +1,145 @@
...
+	keys := m.getSortedKeys()
+	valStrings := make([]string, 0)
+	for _, key := range keys {
+		val := m.arguments[key]
+		valStrings = append(valStrings, fmt.Sprintf("%s%s%v", key, val.delimiter, val.value))
+	}
+	return valStrings
+}
+
+func (m Map) getSortedKeys() []string {
```
andriisoldatenko, line 137: Why not just implement Sort interface instead?

pkg/util/parametermap/map.go
```diff
@@ -0,0 +1,145 @@
+package parametermap
```
andriisoldatenko, line 1: also i don't understand this name, do we want to implement ordered dict/map?

StefanHauth: What this map shall provide is adding new values with a prio, when the key for a value already exists a new value shall not be added if it is less important than the already existing value. Adding values to the map can happen distributed over a codebase while keeping precedence of values decoupled from execution order.

andriisoldatenko: Priority map sounds better than parameter map :) as you described.

pkg/util/parametermap/map.go
StefanHauth: renamed.

pkg/util/parametermap/map.go
StefanHauth: renamed.

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/rawimage_test.go
```diff
@@ -40,7 +41,7 @@ func TestRawImageModify(t *testing.T) {
 		dynakube := getBaseDynakube()
 		enableKubeMonCapability(&dynakube)
 		setRawImageUsage(&dynakube, true)
-		mod := NewRawImageModifier(dynakube)
+		mod := NewRawImageModifier(dynakube, prioritymap.NewMap())
```
andriisoldatenko, line 44:
```suggestion
		mod := NewRawImageModifier(dynakube, prioritymap.NewMap())
```
btw you can just use `prioritymap.New()` because it's already in package name.

pkg/controllers/dynakube/activegate/internal/statefulset/statefulset.go
```diff
@@ -19,11 +20,15 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const defaultEnvPriority = 1
```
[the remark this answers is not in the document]
andriisoldatenko: 100% agree with @aorcholski ☝️ good point

pkg/util/parametermap/map.go
andriisoldatenko: resolved in slack. no needed

pkg/controllers/dynakube/activegate/internal/statefulset/statefulset.go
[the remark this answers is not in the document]
andriisoldatenko: Instead of numbers, it's okay with me to use constants and readable names.

pkg/util/prioritymap/map.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+type entry struct {
+	delimiter string
+	value     any
+	priority  int
+}
+
+const DefaultPriority = LowPriority
+const LowPriority = 1
```
andriisoldatenko, line 16:
```suggestion
const (
    LowPriority = 1
    MediumPriority = 5
    HighPriority = 10
)
```

andriisoldatenko (approved):

PR #2236, opened 2023-10-17
Add context parameter to webhook functions 

aorcholski:
## Description

Context is passed through webhook call-stack:
- CR validators have context parameter because some of them make requests to the K8S API
- `kubesystem.GetUID(..)` has context parameter because is called by validators and reconcilers and always should use correct context 
- `registerInjectEndpoint()`  uses `context.Background()` because no context is available
- in case of unit tests `context.Background()` is used

## How can this be tested?
- all unit tests PASS
- CN deployment + sample app is injected

## Checklist


StefanHauth (approved):
LGTM

PR #2277, opened 2023-10-24
Enable new linter: perfsprint

andriisoldatenko:
## Description

Enable new linter: perfsprint

Some context: 
>Sprint is the slowest way to convert small and big int to string, and on the other hand strconv. FormatInt is the fastest way to convert a int to string , strconv,Itoa is the second fastest way almost near to strconv. FormatInt. So hence prove that for either small or big always use strconv.

From benchmarks in https://dev.to/mavensingh/strconvitoa-vs-strconvformatint-vs-fmtsprintf-which-is-good-to-convert-int-to-string-in-golang-3iol#:~:text=Sprint%20is%20the%20slowest%20way,or%20big%20always%20use%20strconv.

## How can this be tested?

unit-tests

## Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -183,9 +183,10 @@ linters:
   - whitespace
   - unused
   - ineffassign
+  - perfsprint
 
 service:
-  golangci-lint-version: 1.51.x # use the fixed version to not introduce new linters unexpectedly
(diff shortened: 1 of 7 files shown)


waodim (approved):
Tested it - works and LGTM

PR #2281, opened 2023-10-24
Add k8s app enabled feature flag

annazaionchkovska:
## Description

To automatically enable Kubernetes cluster for the Kubernetes 3gen app during Dynatrace deployment process with (kubectl apply -f dynakube.yaml) newly provided feature flag on operator (dynakube.yaml) can be used: **feature.dynatrace.com/k8s-app-enabled**

In case FF is set to true (feature.dynatrace.com/k8s-app-enabled: "true"), then setting object is created accordingly ("schemaId": "builtin:app-transition.kubernetes").

Creation of schema object ("schemaId": "builtin:app-transition.kubernetes") is done with scope of newly created cluster ME, which was created with creation of schema object with "schemaId": "builtin:cloud.kubernetes".

**Expected behavior for feature flags in the operator is:**
FF true --> explicitly set enableApp to true
FF false --> don't do anything
FF not there --> don't do anything.

## How can this be tested?

The result can be validated on cluster settings page 'Kuberneted App' on new cluster creation, toggle 'Enable Kubernetes App' should be enabled
image

*Important*
settings schema version: it will be 1.0.0 with future release of Kubernetes App settings on cluster, but you should be able to use the current version 0.1.0, as regarding properties, there should be no difference between the versions.

## Checklist


pkg/controllers/dynakube/apimonitoring/reconciler.go
```diff
@@ -19,8 +20,13 @@ func NewReconciler(dtc dtclient.Client, clusterLabel, kubeSystemUUID string) *Re
 	}
 }
 
-func (r *Reconciler) Reconcile() error {
-	objectID, err := r.createObjectIdIfNotExists()
+const (
+	settingsSchemaId      = "builtin:cloud.kubernetes"
+	appTransitionSchemaId = "builtin:app-transition.kubernetes"
```
aorcholski, line 25: do we want them to be duplicated? https://github.com/Dynatrace/dynatrace-operator/pull/2281/files#diff-2d6b21fa20ddac7f3c9773d874e19e38ed9706d5aafb9c0e01a7aee578e459b3R85

annazaionchkovska: really good point, how can I reference them from another file ?  actually I was trying to do it, but then I have thought maybe different package is an issue?

0sewa0: export them by starting them with a capital letter

annazaionchkovska: thanks for explanations, issue is fixed

PR #2272, opened 2023-10-24
Add groundwork for OpenTelemetry instrumentation in operator components

StefanHauth:
## Description

Here the groundwork for instrumenting operator components using OpenTelemetry is laid.
* handling of OTel config secret
* OTLP exporters for traces and metrics
* creation of TracerProvider and MetricsProvider
* startup auto instrumentation for golang runtime and host (just as an example)
* simple manually instrumented metric in pod mutator webhook
* simple manual span creation in pod mutator webhook

## How can this be tested?
* deploy the operator
* create the OTel configuration secret according to `olm.md` provided with this PR
* in your tenant and use data explorer to search for the metrics
* in your tenant search for the service "dynatrace-webhook" and have a look at distributed traces
Note: the last two steps might differ if your tenant is configure differently. In case of doubt consult the Dyntrace support pages.

## Checklist

- [x] Unit tests have been updated/added
- [x] PR is labeled accordingly with a single label
- [x] I have read and understood the [contribution guidelines](https://github.com/Dynatrace/dynatrace-operator/blob/main/CONTRIBUTING.md)

PR #2289, opened 2023-10-25
Add more instrumentation for OpenTelemetry traces in webhook.

StefanHauth:
## Description

This is a follow up of   #2272 Added some more exemplary trace instrumentation to the webhook and some auto-instrumentation for "net/http" used in the managers.

## How can this be tested?
Same as #2272.

## Checklist


pkg/injection/namespace/mapper/namespaces.go
```diff
@@ -19,21 +21,27 @@ type NamespaceMapper struct {
...
 }
 
-func (nm NamespaceMapper) updateNamespace() (bool, error) {
+func (nm NamespaceMapper) updateNamespace(ctx context.Context) (bool, error) {
+	ctx, span := injectionotel.StartSpan(ctx, "NamespaceMapper.updateNamespace")
+	defer span.End()
+
 	deployedDynakubes := &dynatracev1beta1.DynaKubeList{}
-	err := nm.client.List(nm.ctx, deployedDynakubes)
+	err := nm.client.List(ctx, deployedDynakubes)
```
aorcholski, line 44: line 47 - maybe we should report an error
```
  span.ReportError(err)
  return
```

pkg/webhook/mutation/namespace_mutator/otel.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func startSpan(ctx context.Context, title string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
+	options := make([]trace.SpanStartOption, 0)
+	options = append(options, opts...)
+	options = append(options, trace.WithAttributes(
+		attribute.String(webhookotel.WebhookPodNameKey, webhookotel.GetWebhookPodName()),
+
+		// TODO: this is just for showcasing now, should be removed in the future
+		attribute.String("debug.info", "foobar")))
```
0sewa0, line 30: we want this in?

StefanHauth: As this PR is kind of a demonstrator, I wouldn't mind keeping this for now along with the comment, knowingly ignoring that keeping TODOs in the code isn't good style.

pkg/injection/namespace/mapper/namespaces.go
StefanHauth: good point, added error recording.

pkg/injection/namespace/mapper/namespaces.go
```diff
@@ -19,21 +21,27 @@ type NamespaceMapper struct {
 }
 
 func NewNamespaceMapper(ctx context.Context, clt client.Client, apiReader client.Reader, operatorNs string, targetNs *corev1.Namespace) NamespaceMapper { //nolint:revive // argument-limit doesn't apply to constructors
-	return NamespaceMapper{ctx, clt, apiReader, operatorNs, targetNs}
+	return NamespaceMapper{ctx, controller_runtime.NewClient(clt), apiReader, operatorNs, targetNs}
```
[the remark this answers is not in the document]
StefanHauth: done.

pkg/injection/namespace/mapper/namespaces.go
```diff
@@ -4,38 +4,46 @@ import (
...
+	updatedNamespace, err := nm.updateNamespace(ctx)
 	if err != nil {
 		return false, err
 	}
 	return updatedNamespace, nil
 }
 
-func (nm NamespaceMapper) updateNamespace() (bool, error) {
+func (nm NamespaceMapper) updateNamespace(ctx context.Context) (bool, error) {
+	ctx, span := injectionotel.StartSpan(ctx, "NamespaceMapper.updateNamespace")
```
0sewa0, line 39: I am not an otel span expert, but I don't want to believe that we have to do it like this (to start a span for each function and close it), this feels unmaintainable and messy.

I would expect to start a span in the entry point, and end it there. (in the case of the webhook, when we get the request)
and maybe pass the span around aswell, to add something to it.

I looked at the keptn operator, their way makes more sense to me: https://github.com/keptn/lifecycle-toolkit/blob/main/lifecycle-operator/controllers/lifecycle/keptnworkloadversion/controller.go#L116

if you follow how they use that span, I hope it gets my point across.

0sewa0 (changes requested):

pkg/injection/namespace/mapper/namespaces.go
StefanHauth: I'll check that, thx!

StefanHauth: It seems that they built a lot more framework/glue layer around OTel than I did in this PR. I tried to keep it as simple as possible on purpose to not completely overstep the focus of it. This PR has an demonstrator character, I think it's also good if the mechanics of OTel are still kept visible.

Passing around spans etc. additionally to a "tracer" context IMHO messes up the code and the function signatures much more than having a two liner to start a new span and attaching function relevant information there directly. In my mind, having one span per function also ties the resp. bits and pieces of information more tightly to the function that produces them. That being said, we still need to learn how we can actually use this information and if it really helps us.

Furthermore, they also start spans in lower parts of the code, just not on every level. It could be a feasible approach to only measure the "more interesting" or heavier/more expensive parts of the code and leave out smaller functions in between.

While I believe that there are good reasons for their way to do it I see a few drawbacks:
* to me it looks like they do not always end a span properly, which is a bit weird to me, not sure what negative side effects that could have on timings
* they do not get full call stack coverage which could reduce the insights, but that's something we need to consider
* they pass around additional contexts and objects, which really clutters the function signatures. From a quick glance I don't really understand, why they do it like this.

albertogdd:
:warning: Small nitpick, label is missing in the PR

StefanHauth:
> ⚠️ Small nitpick, label is missing in the PR

I actually didn't find a fitting one. Suggestions?

luhi-DT:
> > ⚠️ Small nitpick, label is missing in the PR
> 
> I actually didn't find a fitting one. Suggestions?

Core

cmd/csi/provisioner/builder.go
```diff
@@ -113,7 +115,7 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
 		}
 
 		signalHandler := ctrl.SetupSignalHandler()
-		otelShutdownFn := otel.Start(signalHandler, "dynatrace-csi-provisioner", csiManager.GetAPIReader(), builder.namespace)
+		otelShutdownFn := dtotel.Start(context.Background(), "dynatrace-csi-provisioner", csiManager.GetAPIReader(), builder.namespace)
```
0sewa0, line 118: why are you not using the `signalHandler`?
we should have 1 context that is used by everything, no?

the `signalHandler` is just a "fancy" context that gets cancelled in case of a sigterm or sigkill.

cmd/csi/server/builder.go
```diff
@@ -117,7 +119,7 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
 		}
 
 		signalHandler := ctrl.SetupSignalHandler()
-		otelShutdownFn := otel.Start(signalHandler, "dynatrace-csi-server", csiManager.GetAPIReader(), builder.namespace)
+		otelShutdownFn := dtotel.Start(context.Background(), "dynatrace-csi-server", csiManager.GetAPIReader(), builder.namespace)
```
0sewa0, line 122: same as above:
> why are you not using the signalHandler?

cmd/operator/builder.go
```diff
@@ -178,7 +178,7 @@ func (builder CommandBuilder) runOperatorManager(kubeCfg *rest.Config, isDeploye
 		return err
 	}
 
-	otelShutdownFn := otel.Start(builder.getSignalHandler(), "dynatrace-operator", operatorManager.GetAPIReader(), builder.namespace)
+	otelShutdownFn := dtotel.Start(builder.getSignalHandler(), "dynatrace-operator", operatorManager.GetAPIReader(), builder.namespace)
```
0sewa0, line 181: same as above:
> why are you not using the signalHandler?

pkg/injection/otel/otel.go
```diff
@@ -0,0 +1,15 @@
...
+	"go.opentelemetry.io/otel/trace"
+)
+
+func SpanOptions(opts ...trace.SpanStartOption) []trace.SpanStartOption {
+	options := make([]trace.SpanStartOption, 0)
+	options = append(options, opts...)
+	options = append(options, trace.WithAttributes(
+		attribute.String(webhookotel.WebhookPodNameKey, webhookotel.GetWebhookPodName())))
+	return options
+}
```
0sewa0, line 15: is this the correct place for this?

the `injection` directory is for packages that are related to the injection of the user workloads, this is more of just `instrumentation` of otel for our webhook

or am I missing something?

pkg/util/dtotel/controller_runtime/client.go
```diff
@@ -0,0 +1,134 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ client.Client = wrappedClient{}
+
+// Reader knows how to read and list Kubernetes objects.
+type wrappedClient struct {
```
0sewa0, line 16: maybe dumb question: 
instrumenting an external client, wont this cause a bunch of noise for us?
I would want the spans to "start" from code that is actually in my control.

wdyt?

pkg/webhook/mutation/namespace_mutator/otel.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func spanOptions(opts ...trace.SpanStartOption) []trace.SpanStartOption {
+	options := make([]trace.SpanStartOption, 0)
+	options = append(options, opts...)
+	options = append(options, trace.WithAttributes(
+		attribute.String(webhookotel.WebhookPodNameKey, webhookotel.GetWebhookPodName()),
+
+		// TODO: this is just for showcasing now, should be removed in the future
+		attribute.String("debug.info", "foobar")))
```
0sewa0, line 30: I don't see what is this showcasing
there is already an attribute, so we don't need a second one that is just noise

pkg/webhook/mutation/pod_mutator/pod_mutator.go
```diff
@@ -196,17 +201,20 @@ func setDynatraceInjectedAnnotation(mutationRequest *dtwebhook.MutationRequest)
...
+		return silentErrorResponse(ctx, pod, err)
 	}
 	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
 }
 
-func silentErrorResponse(pod *corev1.Pod, err error) admission.Response {
+func silentErrorResponse(ctx context.Context, pod *corev1.Pod, err error) admission.Response {
+	_, span := dtotel.StartSpan(ctx, webhookotel.Tracer(), spanOptions()...)
+	defer span.End()
+
```
0sewa0, line 217: why open a span for a function that does not really use the context?

I don't want to comment for every function where it occurs but my thoughts/questions are:
- why open spans for functions that are not even using the context (because they don't do anything that would really need it)?
- aren't we supposed to "record" errors to the span? which would be made hell if every function creates its on span, no ? to which span should the error belong to?

pkg/webhook/mutation/namespace_mutator/otel.go
StefanHauth: It should show, that one can attach arbitrary attributes. But you might be right, it's worth is probably very limited.

pkg/webhook/mutation/pod_mutator/pod_mutator.go
StefanHauth: If a function uses the context or not has no relevance for the question if it needs a span. The context merely carries the parent span. The new span is written to the context that is returned by `StartSpan`, and in this case it's never used as a parent span for other functions or to record errors. But if this function would have some potentially expensive calculations it would still make sense to have a span to measure the time it needs.
Now, with that being said, it's for sure debatable  if this function worth a span, but we also discussed to go overboard with spans in some code parts just to learn how this feels and looks like.

StefanHauth: And IMHO the error should always be recorded to a span that's opened in the function where it first appears. That way we have the best mapping to the error location. I'm not sure how this would look like, if all errors are recorded to say one global span. That's also something we have to learn.

pkg/webhook/otel/config.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"sync"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const (
+	OpenTelemetryServiceName = "DynatraceWebhook"
```
waodim, line 12: Wouldn't it be better to name the service with lower case letters or am I missing something here?

pkg/util/dtotel/controller_runtime/client.go
StefanHauth: The client is doing potentially expensive I/O. So with creating spans for the client's functions, one can see if time is spent in the calling function (our code) or in the external function. It makes for better visibility of runtime.

cmd/csi/provisioner/builder.go
StefanHauth: fixed

cmd/operator/builder.go
StefanHauth: It is using `builder.getSignalHandler()`?!?

pkg/injection/otel/otel.go
StefanHauth: The idea is to have package specific otel related stuff close to the resp. package but nevertheless separated. Previously I've implemented package specific `StartSpan` functions, but that didn't fly with the automatic span naming.

cmd/webhook/builder.go
```diff
@@ -123,8 +124,11 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
 			return err
 		}
 
+		// instrument webhook manager HTTP client with OpenTelemetry
+		webhookManager.GetHTTPClient().Transport = otelhttp.NewTransport(webhookManager.GetHTTPClient().Transport)
+
 		signalHandler := ctrl.SetupSignalHandler()
-		otelShutdownFn := otel.Start(signalHandler, "dynatrace-webhook", webhookManager.GetAPIReader(), builder.namespace)
+		otelShutdownFn := dtotel.Start(signalHandler, "dynatrace-webhook", webhookManager.GetAPIReader(), builder.namespace)
```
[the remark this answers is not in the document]
StefanHauth: yes, `singalHandler` is the way to go.

pkg/webhook/otel/config.go
StefanHauth: fixed [a139345](https://github.com/Dynatrace/dynatrace-operator/pull/2289/commits/a139345ed45618b919f9dd4663df34270447a785)

pkg/injection/otel/otel.go
0sewa0: we could put it into `pkg/injection/internal/otel`, so it limits the public stuff

pkg/webhook/mutation/namespace_mutator/otel.go
albertogdd: I would remove it too

cmd/manager/mock.go
```diff
@@ -0,0 +1,73 @@
+package manager
```
aorcholski, line 1: This implementation is not needed. We have mock for Manager interface built by [mockery](https://github.com/Dynatrace/dynatrace-operator/blob/main/test/mocks/sigs.k8s.io/controller-runtime/pkg/manager/manager.go)

aorcholski (changes requested):

cmd/manager/mock.go
StefanHauth: removed in [8d47858](https://github.com/Dynatrace/dynatrace-operator/pull/2289/commits/8d4785872b1570310f4f1f3e084a692d52e52202)

pkg/webhook/mutation/namespace_mutator/otel.go
StefanHauth: removed [97b95f6](https://github.com/Dynatrace/dynatrace-operator/pull/2289/commits/97b95f6dc9dc947241d2d154667b93effda24027)

pkg/injection/otel/otel.go
StefanHauth: [8d47858](https://github.com/Dynatrace/dynatrace-operator/pull/2289/commits/8d4785872b1570310f4f1f3e084a692d52e52202)

pkg/webhook/mutation/namespace_mutator/otel.go
0sewa0: well I still see it 😄 

https://github.com/Dynatrace/dynatrace-operator/pull/2289/files#diff-4cf48530dbad519d66501942ecbbceccc2d0ec368545789a52b4458a7da029e1R30

0sewa0 (approved):

PR #2295, opened 2023-10-27
Feature/k8s app enablement ff app transition schema version

annazaionchkovska:
## Description

This PR is follow up to PR [Add k8s app enabled feature flag](https://github.com/Dynatrace/dynatrace-operator/pull/2281)

The version of schema *"builtin:app-transition.kubernetes"* is set to *"0.1.0"*

as mentioned in previous PR and issue description:
<em>
settings schema version: it will be 1.0.0 with future cluster update, but you should be able to use the current version 0.1.0, as regarding properties, there should be no difference between the versions.
</em>

It was discussed again with PO and there is a suggestion to use version  *"0.1.0"* for testing purpouse.

## How can this be tested?

same to original PR [Add k8s app enabled feature flag](https://github.com/Dynatrace/dynatrace-operator/pull/2281)

## Checklist

- [X] Unit tests have been updated/added -> same to original PR [Add k8s app enabled feature flag](https://github.com/Dynatrace/dynatrace-operator/pull/2281)


annazaionchkovska:
It was agreed that we should continue using version 1.0.0 for operator, as version of transition schema on cluster was set to 1.0.0 also

annazaionchkovska:
It was agreed that we should continue using version 1.0.0 for operator, as version of transition schema on cluster was set to 1.0.0 also

PR #2297, opened 2023-10-27
Fix self deleting owner reference in created istio CRs

gkrenn:
## Description

Fixes a problem where owner references of created istio CRs were deleted after they were updated.

In addition refactors the istio client implementation to keep a reference to the owner directly, as one instance is anyway always using the same owner.

## How can this be tested?

- create dynakube without active gate section and istio enabled
- check if service entries were created and have owner references
- update dynakube to have an active gate section 
- service entries should be updated and not loose the owner reference
- do the same for virtual services

## Checklist


wepudt (approved):
lgtm

PR #2298, opened 2023-10-30
Add edgeconnect  provisioner mode

aorcholski:
## Description

The operator supports EdgeConnect `Provisioner` mode.  EdgeConnect CR has new flag `spec.oauth.provisioner`. 
If the flag is set to `true` then the operator 
- creates the EdgeConnect entry on the tenant
- updates the EdgeConnect entry on the tenant if `HostPatterns` field is changed in the CR
- removes the EdgeConnect entry on the tenant when the CR is being deleted
- handles the OAuth credentials, provides spec.oauth.clientSecret to the EC deployment in `Regular` mode and an additional secret with credentials from the EC API (CREATE) in `Provisioner` mode

TODO comments point out to locations where error recovery logic should be added (based on `managedByDynatraceOperator` attribute). For example:
- EdgeConnect entry on the tenant created by the operator exists but has to be re-created
- EdgeConnect entry on the tenant having the same name as CR already exist 
- EdgeConnect entry on the tenant has the same name as CR but different ID

OAuth credentials provided by the EC API are not rotated because there is no way to detect the problem.

## How can this be tested?
Make sure your tenant OAuth client has required scopes.
1) Deploy EdgeConnect CR
```
apiVersion: v1
kind: Secret
metadata:
  name: ec-provisioner
  namespace: dynatrace
data:
  oauth-client-id: <client-id>
  oauth-client-secret: <client-secret>
---                                                                                                                                                                                                                                                                                                                                                                          
kind: EdgeConnect
metadata:
  name: ec-provisioner
  namespace: dynatrace
spec:
  apiServer: <tenant>.dev.apps.dynatracelabs.com
  oauth:
    clientSecret: ec-provisioner                                                                                                                                                                                                                                                                                                                                                     
    endpoint: <sso-endpoint>
    resource: <resource>
    provisioner: true
  hostPatterns:
    - "*.internal.org"
```
After a moment 
- the EdgeConnect CR status is `RUNNING`
- `ec-provisioner` deployment is created
- `ec-provisioner-client` secret is created
- `ec-provisioner` entry on the tenant is created
2) Modify CR `hostPatterns` field and apply changes
- `ec-provisioner` entry on the tenant is updated accordingly
3) Delete EdgeConnect CR
- `ec-provisioner` entry on the tenant is also deleted


## Checklist


pkg/clients/edgeconnect/client.go
```diff
@@ -149,25 +153,25 @@ func (c *client) getServerResponseData(response *http.Response) ([]byte, error)
 }
 
 // GetEdgeConnect returns edge connect if it exists
-func (c *client) GetEdgeConnect(edgeConnectId string) (GetResponse, error) {
+func (c *client) GetEdgeConnect(edgeConnectId string) (EdgeConnectResponse, error) {
```
andriisoldatenko, line 156: I intentionally named it short. If you check the package name:
Now we have `edgeconnect.EdgeConnectResponse` instead of `edgeconnect.GetResponse`

pkg/controllers/edgeconnect/controller.go
```diff
@@ -152,8 +229,244 @@ func (controller *Controller) updateEdgeConnectStatus(ctx context.Context, edgeC
...
+	oauthClientId, err := kubeobjects.ExtractToken(&secret, consts.KeyEdgeConnectOauthClientID)
+	if err != nil {
+		return "", "", errors.WithStack(err)
+	}
+
+	oauthClientSecret, err := kubeobjects.ExtractToken(&secret, consts.KeyEdgeConnectOauthClientSecret)
+	if err != nil {
+		return "", "", errors.WithStack(err)
+	}
+	return oauthClientId, oauthClientSecret, nil
```
andriisoldatenko, line 333: Since we need these two as pairs, we can create a new type to hold both; it will simplify the return type and reduce the amount of func arguments later from 2 to 1.

pkg/controllers/edgeconnect/controller.go
```diff
@@ -152,8 +229,244 @@ func (controller *Controller) updateEdgeConnectStatus(ctx context.Context, edgeC
...
+}
+
+func newEdgeConnectClient() func(ctx context.Context, edgeConnect *edgeconnectv1alpha1.EdgeConnect, oauthClientId string, oauthClientSecret string) (edgeconnect.Client, error) {
+	return func(ctx context.Context, edgeConnect *edgeconnectv1alpha1.EdgeConnect, oauthClientId string, oauthClientSecret string) (edgeconnect.Client, error) {
+		edgeConnectClient, err := edgeconnect.NewClient(
```
andriisoldatenko, line 338: same here, we can create new type
type OauthCredentials {
ClientId string
ClientSecret string
}

pkg/clients/edgeconnect/client.go
```diff
@@ -276,3 +280,37 @@ func (c *client) CreateEdgeConnect(name string, hostPatterns []string, oauthClie
 
 	return response, nil
 }
+
+// GetEdgeConnects returns list of edge connects
+func (c *client) GetEdgeConnects(name string) (GetEdgeConnectsResponse, error) {
```
0sewa0, line 285:
```suggestion
func (c *client) List(name string) (ListResponse, error) {
```
if we want to follow Andrii's pattern, which I also would prefer

pkg/clients/edgeconnect/config.go
```diff
@@ -5,5 +5,5 @@ import (
 )
 
 var (
-	log = logger.Factory.GetLogger("edgeConnectClient")
+	log = logger.Factory.GetLogger("edgeConnectClient") //nolint: unused
```
0sewa0, line 8: why not using it ? (then just remove it)

pkg/clients/edgeconnect/edgeconnect.go
```diff
@@ -28,12 +28,20 @@ type GetResponse struct {
 	Metadata         Metadata         `json:"metadata"`
 }
 
+type GetEdgeConnectsResponse struct {
```
0sewa0, line 31:
```suggestion
type ListResponse struct {
```
(related to my other comment)

pkg/controllers/edgeconnect/controller.go
```diff
@@ -26,20 +30,23 @@ import (
 )
 
 const (
-	errorUpdateInterval   = 1 * time.Minute
-	defaultUpdateInterval = 30 * time.Minute
+	fastUpdateInterval    = 1 * time.Minute
+	defaultUpdateInterval = 3 * time.Minute // 30
```
0sewa0, line 34: what is with the `//30`? (probably a mistake)

pkg/controllers/edgeconnect/controller.go
```diff
@@ -75,52 +83,87 @@ func (controller *Controller) Reconcile(ctx context.Context, request reconcile.R
...
-	if err = versionReconciler.Reconcile(ctx); err != nil {
-		log.Error(err, "reconciliation of EdgeConnect failed", "name", request.Name, "namespace", request.Namespace)
-		return reconcile.Result{RequeueAfter: errorUpdateInterval}, nil
+	edgeConnectId, err := controller.getEdgeConnectIdFromClientSecret(ctx, edgeConnect)
+	if err != nil {
+		return err
 	}
 
+	if edgeConnectId == "" {
+		// TODO: managedByDynatraceOperator attribute needed
```
0sewa0, line 112: will we need `managedByDynatraceOperator` during deletion ?

pkg/controllers/edgeconnect/controller.go
```diff
@@ -138,6 +181,40 @@ func (controller *Controller) getEdgeConnect(ctx context.Context, name, namespac
 	return edgeConnect, nil
 }
 
+func (controller *Controller) updateFinalizers(ctx context.Context, edgeConnect *edgeconnectv1alpha1.EdgeConnect) error {
+	if edgeConnect.Spec.OAuth.Provisioner && len(edgeConnect.ObjectMeta.Finalizers) == 0 {
+		log.Info("updating finalizers", "name", edgeConnect.Name, "namespace", edgeConnect.Namespace)
+		edgeConnect.ObjectMeta.Finalizers = []string{"server"}
```
0sewa0, line 187: could we have the finalizer string in a `const` ?

pkg/clients/edgeconnect/client_test.go
```diff
@@ -188,7 +188,7 @@ func edgeConnectGetServerHandler() http.HandlerFunc {
 			writeOauthTokenResponse(writer)
 		case fmt.Sprintf("/edge-connects/%s", EdgeConnectID):
 			writer.WriteHeader(http.StatusOK)
-			resp := GetResponse{
+			resp := EdgeConnectResponse{
```
waodim, line 191: If we follow @andriisoldatenko pattern I think GetResponse would be better. What do you think?

pkg/controllers/edgeconnect/controller.go
aorcholski: 3 is a debug value, fixed

pkg/clients/edgeconnect/config.go
aorcholski: logger is unused, removed

pkg/controllers/edgeconnect/controller.go
aorcholski: done but private name and fields

pkg/clients/edgeconnect/client.go
aorcholski: I wanted to have 
- EdgeConnectResponse
- EdgeConnectsResponse

but IMO
- GetResponse
- ListResponse

you suggest are better because there is no duplication of `EdgeConnect` in names
thx

pkg/controllers/edgeconnect/controller.go
```diff
@@ -27,20 +32,30 @@ import (
...
 const (
-	errorUpdateInterval   = 1 * time.Minute
+	fastUpdateInterval    = 1 * time.Minute
 	defaultUpdateInterval = 30 * time.Minute
+
+	finalizerName = "server"
 )
 
+type oauthCredentialsType struct {
+	oauthClientId     string
```
andriisoldatenko, line 42: nit pick: 
```suggestion
	clientId     string
```

pkg/controllers/edgeconnect/controller.go
```diff
@@ -27,20 +32,30 @@ import (
...
+	oauthClientSecret string
```
andriisoldatenko, line 43:
```suggestion
	clientSecret string
```

pkg/controllers/edgeconnect/controller.go
andriisoldatenko: to use it like `oauthCredentialsType.clientId`

waodim (commented):
From my point of view my review finished, as I think it is okay. Maybe @0sewa0 and @andriisoldatenko have more things to add, so let us wait for their review.

andriisoldatenko:
> From my point of view my review finished, as I think it is okay. Maybe @0sewa0 and @andriisoldatenko have more things to add, so let us wait for their review.

yeah it's in my list :) will do

pkg/controllers/edgeconnect/controller.go
```diff
@@ -174,3 +259,243 @@ func (controller *Controller) reconcileEdgeConnect(edgeConnect *edgeconnectv1alp
...
+			edgeConnectId = ""
+		} else if edgeConnectId != edgeConnectIdFromSecret {
+			// TODO: managedByDynatraceOperator attribute needed
+			log.Info("EdgeConnect has to be recreated due to invalid Id")
+			if err := edgeConnectClient.DeleteEdgeConnect(edgeConnectId); err != nil {
+				return err
+			}
+			edgeConnectId = ""
+		}
+	}
```
0sewa0, line 297: is this possible, considering we have the `managedByDynatraceOperator` ?

if not, then we should make it clear that this logic can be removed when we start using `managedByDynatraceOperator`

pkg/controllers/edgeconnect/controller.go
0sewa0: but we are not making a request here, so we would add a request, to delete it by name if it is managed by the operator, right

0sewa0 (commented):
practically LGTM, I tested it on my cluster

I just raised a few questions about the TODOs

pkg/controllers/edgeconnect/controller.go
0sewa0: offline conversation summary: there should be an extra if there when we have the new property

0sewa0 (approved):

andriisoldatenko (approved):

PR #2301, opened 2023-11-02
Add infra-only mode to host monitoring sample

waodim:
## Description

Host monitoring sample is updated such that it has infra-only mode enabled by default.

## Checklist


albertogdd (approved):
Tried the sample locally and works fine :+1:

0sewa0 (approved):

PR #2302, opened 2023-11-03
Speed up retrying token fixing

wepudt:
## Description

In case of a problem with the token we did only retry every 15 minutes to fix the token. By only setting the timestamp if a token could be validated we retry much more often until succeeded.

## How can this be tested?

Deploy a dynakube with a proxy pointing nowhere- if described the dynakube is in error state because it cannot get the token. Reapply the dynakube without the proxy and it should get immediately healed.

## Checklist

### DIFF
--- a/pkg/controllers/dynakube/dynatraceclient/builder.go
+++ b/pkg/controllers/dynakube/dynatraceclient/builder.go
@@ -120,7 +120,6 @@ func (dynatraceClientBuilder builder) verifyTokenScopes(dynatraceClient dtclient
 	var err error
 
 	if dynatraceClientBuilder.dynakube.IsTokenScopeVerificationAllowed(timeprovider.New()) {
-		dynaKubeStatus.DynatraceApi.LastTokenScopeRequest = metav1.Now()
 		err = dynatraceClientBuilder.tokens.VerifyScopes(dynatraceClient)
 		log.Info("token verified")
 	} else {
@@ -132,6 +131,7 @@ func (dynatraceClientBuilder builder) verifyTokenScopes(dynatraceClient dtclient
 	}
 
 	if err != nil {
+		dynaKubeStatus.DynatraceApi.LastTokenScopeRequest = metav1.Now()
 		return err
 	}
 
--- a/pkg/controllers/dynakube/dynatraceclient/builder.go
+++ b/pkg/controllers/dynakube/dynatraceclient/builder.go
@@ -119,21 +119,20 @@ func (dynatraceClientBuilder builder) BuildWithTokenVerification(dynaKubeStatus 
 func (dynatraceClientBuilder builder) verifyTokenScopes(dynatraceClient dtclient.Client, dynaKubeStatus *dynatracev1beta1.DynaKubeStatus) error {
 	var err error
 
-	if dynatraceClientBuilder.dynakube.IsTokenScopeVerificationAllowed(timeprovider.New()) {
-		err = dynatraceClientBuilder.tokens.VerifyScopes(dynatraceClient)
-		log.Info("token verified")
-	} else {
+	if !dynatraceClientBuilder.dynakube.IsTokenScopeVerificationAllowed(timeprovider.New()) {
 		log.Info(dynatracev1beta1.GetCacheValidMessage(
 			"token verification",
 			dynaKubeStatus.DynatraceApi.LastTokenScopeRequest,
 			dynatraceClientBuilder.dynakube.FeatureApiRequestThreshold()))
-		err = lastErrorFromCondition(dynaKubeStatus)
+		return lastErrorFromCondition(dynaKubeStatus)
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/controllers/dynakube/dynatraceclient/builder.go
```diff
@@ -120,7 +120,6 @@ func (dynatraceClientBuilder builder) verifyTokenScopes(dynatraceClient dtclient
 	var err error
 
 	if dynatraceClientBuilder.dynakube.IsTokenScopeVerificationAllowed(timeprovider.New()) {
-		dynaKubeStatus.DynatraceApi.LastTokenScopeRequest = metav1.Now()
 		err = dynatraceClientBuilder.tokens.VerifyScopes(dynatraceClient)
 		log.Info("token verified")
 	} else {
```
0sewa0, line 125: the problem here is that timestamp now is always reset if there is no error
- it should only be reset IF `VerifyScopes` was called without any error

this code is a bit confusing because it tries to save 1 `if err != nil` check, but goes about it the wrong way, the whole structure is a bit over complicated

it could be simplified: (super quick github-based code copy-pasta, soo may not work :D)
```
func (dynatraceClientBuilder builder) verifyTokenScopes(dynatraceClient dtclient.Client, dynaKubeStatus *dynatracev1beta1.DynaKubeStatus) error {
	if dynatraceClientBuilder.dynakube.IsTokenScopeVerificationAllowed(timeprovider.New()) {
		dynaKubeStatus.DynatraceApi.LastTokenScopeRequest = metav1.Now()
		err := dynatraceClientBuilder.tokens.VerifyScopes(dynatraceClient)
		if err != nil { 
		     return err
		}
		log.Info("token verified")
		dynaKubeStatus.DynatraceApi.LastTokenScopeRequest = metav1.Now()
		return nil
	}
	log.Info(dynatracev1beta1.GetCacheValidMessage(
			"token verification",
			dynaKubeStatus.DynatraceApi.LastTokenScopeRequest,
			dynatraceClientBuilder.dynakube.FeatureApiRequestThreshold()))
	return lastErrorFromCondition(dynaKubeStatus)
}
```

wdyt?

0sewa0 (changes requested):

pkg/controllers/dynakube/dynatraceclient/builder.go
0sewa0: now that I reread it, it may make sense to have the "simple" block of code in the `if statement` and the "complicated" block outside the `if statement` 🤔 

wdyt?

pkg/controllers/dynakube/dynatraceclient/builder.go
```diff
@@ -119,22 +119,21 @@ func (dynatraceClientBuilder builder) BuildWithTokenVerification(dynaKubeStatus
...
+	if !dynatraceClientBuilder.dynakube.IsTokenScopeVerificationAllowed(timeprovider.New()) {
 		log.Info(dynatracev1beta1.GetCacheValidMessage(
 			"token verification",
 			dynaKubeStatus.DynatraceApi.LastTokenScopeRequest,
 			dynatraceClientBuilder.dynakube.FeatureApiRequestThreshold()))
-		err = lastErrorFromCondition(dynaKubeStatus)
+		return lastErrorFromCondition(dynaKubeStatus)
 	}
 
+	err = dynatraceClientBuilder.tokens.VerifyScopes(dynatraceClient)
```
andriisoldatenko, line 130:
```suggestion
	err := dynatraceClientBuilder.tokens.VerifyScopes(dynatraceClient)
```

pkg/controllers/dynakube/dynatraceclient/builder.go
```diff
@@ -119,22 +119,21 @@ func (dynatraceClientBuilder builder) BuildWithTokenVerification(dynaKubeStatus
 func (dynatraceClientBuilder builder) verifyTokenScopes(dynatraceClient dtclient.Client, dynaKubeStatus *dynatracev1beta1.DynaKubeStatus) error {
 	var err error
```
andriisoldatenko, line 120:
```suggestion
```

pkg/controllers/dynakube/dynatraceclient/builder.go
```diff
@@ -119,22 +119,21 @@ func (dynatraceClientBuilder builder) BuildWithTokenVerification(dynaKubeStatus
...
-		err = lastErrorFromCondition(dynaKubeStatus)
+		return lastErrorFromCondition(dynaKubeStatus)
 	}
 
+	err = dynatraceClientBuilder.tokens.VerifyScopes(dynatraceClient)
 	if err != nil {
 		return err
 	}
 
+	log.Info("token verified")
```
andriisoldatenko, line 135:
```suggestion
	log.Info("token has been verified")
```

PR #2313, opened 2023-11-06
 Validate NamespaceSelector Labels

wepudt:
## Description

Unit now we did not validate if the labels provided in the NamespaceSelector was a valid K8s Label. This remedies the issue and only allows labels of the correct format  

## How can this be tested?

Apply a Dynakube with an incorrect label (over 63 chars, special characters other than '-', '_', '.'). There should be a corresponding error message.

## Checklist


pkg/webhook/validation/dynakube/namespace_selector.go
```diff
@@ -14,6 +15,7 @@ Make sure the namespaceSelector doesn't conflict with other Dynakubes namespaceS
 	errorConflictingNamespaceSelectorNoSelector = `The DynaKube does not specificy namespaces where it should inject into while another Dynakube already injects into namespaces, which is not supported.
 Make sure you have a namespaceSelector doesn't conflict with other Dynakubes namespaceSelector
 `
+	errorNamespaceSelectorMatchLabelsViolateLabelSpec = "The DynaKube's namespaceSelector contains matchLabels that are not conform to spec. A valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character."
```
0sewa0, line 18: doesn't the validation give back some output that we could just use?
Something like:
`The DynaKube's namespaceSelector contains matchLabels that are not conform to spec: <What the validation gives back>`

pkg/webhook/validation/dynakube/namespace_selector.go
```diff
@@ -33,3 +35,17 @@ func conflictingNamespaceSelector(ctx context.Context, dv *dynakubeValidator, dy
...
 }
+
+func namespaceSelectorMatchLabelsViolateLabelSpec(ctx context.Context, _ *dynakubeValidator, dynakube *dynatracev1beta1.DynaKube) string {
+	var errs []string
+	matchLabels := dynakube.NamespaceSelector().MatchLabels
+	for _, v := range matchLabels {
+		if v != "" {
+			errs = validation.IsValidLabelValue(v)
+		}
+	}
```
0sewa0, line 46: why not use `validation.ValidateLabels` ?

it needs a field path, that is only used for creating error message. (based on example I found: `field.NewPath("spec", "namespaceSelector", "matchLabels")`)

0sewa0 (changes requested):

pkg/webhook/validation/dynakube/namespace_selector.go
wepudt: I actually thought about that too but wanted not to deviate from the pattern I saw, but will happily do so :)

0sewa0 (approved):
LGTM, tested it and got ⭐️fancy⭐️ output: (it found and printed both problems)
```
Error from server (Forbidden): error when creating "./.vscode/deploy/dynakube_zib50933.yaml": admission webhook "webhook.dynatrace.com" denied the request:
1 error(s) found in the Dynakube
 1. The DynaKube's namespaceSelector contains matchLabels that are not conform to spec. ([spec.namespaceSelector.matchLabels: Invalid value: "AMuchTooLongLabelThatGoesOverSixtyThreeCharactersAndSoIsInvalidAccordingToSpec%": must be no more than 63 characters spec.namespaceSelector.matchLabels: Invalid value: "AMuchTooLongLabelThatGoesOverSixtyThreeCharactersAndSoIsInvalidAccordingToSpec%": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')])
```

PR #2316, opened 2023-11-07
Rework e2e tests

0sewa0:
## Description
The structure of our e2e tests have been redone, to make them less flaky. (also many fixes were introduced)

Now instead of redeploying the operator for each "feature", instead we group the features into "scenarios".
Each scenarios is a type of install of the operator.
- `standard`: installing the latest operator, according to the branch we are running the tests
- `istio`:  installing `standard` + istio is injected into the dynatrace namespace, features that require network restrictions can be found here.
- `release`:   installing a previously released operator version, this is only used for the "upgrade" test, and probably it will stay that way.

By doing this, we are also speeding up the test by a lot, also switching between modes/custom-resources should be something that the operator can handle.

Changes to the make targets:
- now we are using labels to run a single feature test, these labels need to be defined when implementing the feature test
- no more `run_all.sh` 🎉
- limitation of the `/debug` targets:
     - it will **never** clean up the operator, even if everything passed, keep this in mind. (there was no reasonable way to get over this)
     - this is why there is no `test/e2e/debug`, because switching between the scenarios can/will cause problems duo to lack of cleanup

Some refactors:
- Removed all the `Builder` patterns, and replaced them with the "functional options pattern"
   - Why?: Its a more go-like pattern, and the `Builder` pattern enabled usage patterns that were convoluted, messy and prone to error. (example: reusing a builder)
- Some functions were changed to be `env.Func` from `feature.Func`, so they can be used during the `Setup` and `Finish`
    - Its easy to make an `feature.Func` from an `env.Func` on the fly, wrote a simple helper for that. (the other way is not that easy) 

## How can this be tested?

`make test/e2e/....` every command
MAKE SURE YOUR CLUSTER IS CLEAN, otherwise it might get flaky

## Checklist


hack/make/tests/e2e.mk
```diff
@@ -1,107 +1,108 @@
 ## Start a test and skip TEARDOWN steps if it fails
```
waodim, line 1: Please add a make target for edgeconnect test

0sewa0: done

test/features/edgeconnect/edgeconnect.go
```diff
@@ -0,0 +1,35 @@
...
+	secretConfig := tenant.GetEdgeConnectTenantSecret(t)
+
+	testEdgeConnect := *edgeconnect.New(
+		// this name should match with tenant edge connect name
+		edgeconnect.WithName(secretConfig.Name),
+		edgeconnect.WithApiServer(secretConfig.ApiServer),
+		edgeconnect.WithOAuthClientSecret(fmt.Sprintf("%s-client-secret", secretConfig.Name)),
+		edgeconnect.WithOAuthEndpoint("https://sso-dev.dynatracelabs.com/sso/oauth2/token"),
+		edgeconnect.WithOAuthResource(fmt.Sprintf("urn:dtenvironment:%s", secretConfig.TenantUid)),
+		edgeconnect.WithCustomPullSecret(fmt.Sprintf("%s-docker-pull-secret", secretConfig.Name)),
```
waodim, line 28: I am pretty sure that this line can be removed as we no longer need the custom pull secret to run edgeconnect (because of public registry). Also the secret sample for edgeconnect can be adapted and the fields according fields can be removed.

0sewa0: done 😉

wepudt:
ran the e2e tests yesterday and had two insights:
* our EdgeConnect docs are missing information- could not get it to work properly which was a good thing as it allowed me to notice...
* that even though tests failed my cluster was in a clean state, which is very nice.

andriisoldatenko (approved):

PR #2320, opened 2023-11-07
Always Add ClusterId EnvVar To Pods

wepudt:
## Description

When the DynaKube feature flag `feature.dynatrace.com/automatic-injection` was set to false to use `oneagent.dynatrace.com/inject: "true"` on a pod, the pod did not get the env var `K8S_CLUSTER_ID` injected. This resulted in the OneAgent not being able to generate a correct host identifier which then lead to a pod not being visible in the UI.
This PR fixes that issue.

## How can this be tested?

create a DynaKube with `feature.dynatrace.com/automatic-injection: "false"` and a sample app with `oneagent.dynatrace.com/inject: "true"`.
Then `/var/lib/dynatrace/oneagent/agent/config/container.conf` should contain a valid cluster ID.

## Checklist

### DIFF
--- a/pkg/injection/startup/env.go
+++ b/pkg/injection/startup/env.go
@@ -92,6 +92,7 @@ func (env *environment) getCommonFieldSetters() []func() error {
 		env.addK8PodName,
 		env.addK8PodUID,
 		env.addK8Namespace,
+		env.addK8ClusterID,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


andriisoldatenko (approved):

0sewa0 (approved):

0sewa0:
nitpick: the PR title is not that accurate, as we even currently always add the "ClusterId EnvVar To Pods", but we not always look at it / consider it

PR #2321, opened 2023-11-07
Replace custom modifier test code with mockery generated one

gkrenn:
## Description

This PR replaces the mock structs we used with mockery generated ones.

## How can this be tested?

- Run unit tests and check if everything is still working

## Checklist


0sewa0 (approved):

test/mocks/pkg/util/builder/Modifier.go
```diff
@@ -0,0 +1,115 @@
+// Code generated by mockery. DO NOT EDIT.
```
andriisoldatenko, line 1: why not lowercase?

PR #2339, opened 2023-11-10
Generate mock for cmd.config.Provider interface using vectra/mockery

andriisoldatenko:
## Description

This PR migrates existing mock for cmd.config.Provider interface using vectra/mockery

>Note!: removed `goimports` from `.golangci.yml` because we use gci instead of goimports and they are conflict. 
E.g.
```
go fmt ./...
gci write .
go vet -copylocks=false ./...
golangci-lint run --build-tags "containers_image_openpgp,osusergo,netgo,sqlite_omit_load_extension,containers_image_storage_stub,containers_imag
e_docker_daemon_stub,e2e," --timeout 300s
pkg/util/builder/builder_test.go:6: File is not `goimports`-ed with -local github.com/Dynatrace/dynatrace-operator (goimports)
        "github.com/Dynatrace/dynatrace-operator/pkg/util/builder/mocks"
        modifiermock "github.com/Dynatrace/dynatrace-operator/test/mocks/pkg/util/builder"
pkg/version/version.go:7: File is not `goimports`-ed with -local github.com/Dynatrace/dynatrace-operator (goimports)
        "github.com/Dynatrace/dynatrace-operator/pkg/util/logger"
pkg/clients/edgeconnect/client.go:12: File is not `goimports`-ed with -local github.com/Dynatrace/dynatrace-operator (goimports)
        "github.com/Dynatrace/dynatrace-operator/pkg/clients/utils"
make: *** [go/golangci] Error 1
```

## How can this be tested?

unittests

## Checklist


.mockery.yaml
```diff
@@ -6,6 +6,11 @@ filename: "{{.InterfaceName | snakecase}}.go"
 outpkg: mocks
 dir: "test/mocks/{{.InterfaceDirRelative}}"
 packages:
+  github.com/Dynatrace/dynatrace-operator/cmd/config:
+    interfaces:
+      Provider:
+        config:
+          mockname: "Mock{{.InterfaceName}}"
```
aorcholski, line 13: would it be possible to unify `mockname` ? I see so far we prefer `Mock{{InterfaceName}}` so maybe it could be the default value for `mockname` ?

andriisoldatenko: Yeah after migrating, we can do a round of refactoring, I just reuse the same name we used before.

On another the side: package name already is: `mocks.InterfaceName` so it's better to refactor the code and use default style to not dup mock 2 times like in current implementation: `mocks.MockInterfaceName`.

andriisoldatenko: @aorcholski done in e1b4002d

aorcholski (approved):

albertogdd (approved):

PR #2340, opened 2023-11-10
Generate mock for webhook.PodMutator interface using vectra/mockery 

albertogdd:
## Description

Removed old manual mock impl and added the auto generated one via Mockery.

## How can this be tested?

`make go/gen_mocks` to generate mocks
`make go/test` to see everything is working fine

## Checklist


andriisoldatenko (approved):

pkg/webhook/mutation/pod_mutator/pod_mutator_test.go
```diff
@@ -284,7 +285,7 @@ func createTestWebhook(mutators []dtwebhook.PodMutator, objects []client.Object)
 }
 
 func createSimplePodMutatorMock() dtwebhook.PodMutator {
-	mutator := dtwebhook.PodMutatorMock{}
+	mutator := mocks.PodMutator{}
```
aorcholski, line 288: `mocks.NewProvider(t)` should be used in create*Mock functions -> line [240](https://github.com/Dynatrace/dynatrace-operator/pull/2340/files#diff-130a5f5e79e744f15bb582d9ac4e4d97c3b8cd6722caeb0cc71054fb81e86ddbR240)

pkg/webhook/mutation/pod_mutator/pod_mutator_test.go
```diff
@@ -236,13 +237,13 @@ func TestHandlePodReinvocation(t *testing.T) {
 
 		updated := podWebhook.handlePodReinvocation(context.Background(), mutationRequest)
 		require.False(t, updated)
-		failingMutator.(*dtwebhook.PodMutatorMock).AssertCalled(t, "Enabled", mutationRequest.BaseRequest)
-		failingMutator.(*dtwebhook.PodMutatorMock).AssertCalled(t, "Reinvoke", mutationRequest.ToReinvocationRequest())
+		failingMutator.(*mocks.PodMutator).AssertCalled(t, "Enabled", mutationRequest.BaseRequest)
```
aorcholski, line 240: without `AssertExpectations()` (see [288](https://github.com/Dynatrace/dynatrace-operator/pull/2340/files#diff-130a5f5e79e744f15bb582d9ac4e4d97c3b8cd6722caeb0cc71054fb81e86ddbR288)) we don't know if `Injected` and `Mutate` functions declared by `createFailPodMutatorMock` are called / not called in this scenario.

We can add `Maybe()` to `On()` declarations if want to reuse `create*Mock` function in different scenarios. Then we have to clearly define our expectations using for example `AssertCalled`/`AssertNotCalled`.

albertogdd: After having a look into AssertExpectations, what's your opinion on simply using
```go
		failingMutator.(*mocks.PodMutator).AssertNotCalled(t, "Injected", mock.Anything)
		failingMutator.(*mocks.PodMutator).AssertNotCalled(t, "Mutated", mock.Anything)
```
Because we use the same function in all tests to create the mock with `On()`, `Return()` and so on, hence `AssertExpectations()` can't pass because Injected and Mutated are not even called.

pkg/webhook/mutation/pod_mutator/pod_mutator_test.go
albertogdd: I think we are mixing terms here! `NewProvider()` is an auto-generated function based on the config command interface, this doesn't exist on the webhook package I'm testing (AFAIK). IMO `mocks.PodMutator{}` is still the correct way, what do you think?

albertogdd: Clarified in the call :+1: I just added the mock constructor call

pkg/webhook/mutation/pod_mutator/pod_mutator_test.go
albertogdd: Clarified in the call too :+1:

aorcholski (approved):

PR #2350, opened 2023-11-14
Fix race condition where code modules are not injected into ActiveGate pods

StefanHauth:
## Description

Injection of code modules into ActiveGate for self-monitoring purposes doesn't work without an ActiveGate pod restart because of a race condition and a wrong reconciliation loop ordering.

The connection info is not always written to the DynaKube status in timely enough that the code modules injection can pick it up when ActiveGate pods starts. Consequently, injection is postponed.
Additionally the app monitoring reconciliation was happening after ActiveGate deployment, so that it again was not ready when the ActiveGate pods start.

## How can this be tested?

* Deploy a cloud native DynaKube with the following feature flag applied:
```
feature.dynatrace.com/ignored-namespaces: '["^kube-.*", "openshift(-.*)?", "^gke-.*", "^gmp-.*" ]'
```
* Check if the code modules got injected into the ActiveGate pods
* Retry a few times and see if it works everytime

## Checklist


pkg/webhook/mutation/pod_mutator/pod_mutator.go
```diff
@@ -55,16 +55,32 @@ type podMutatorWebhook struct {
...
+		log.Error(err, "building mutation request base encountered an error")
+		return nil, emptyPatch, false
+	}
+	if mutationRequest == nil {
+		emptyPatch.Result.Message = "unable to inject into pod (no error reported)"
+		log.Error(err, "building mutation request base delivered nil result")
+		return nil, emptyPatch, false
+	}
+	return mutationRequest, emptyPatch, true
+}
```
0sewa0, line 71: this function is a doozy:
- has 3 return values, already a bad sign 😬
-  feels very much like the `silentErrorResponse`, but different ?

so AFAIK`mutationRequest` is only `nil` in case there was an error, so `mutationRequest == nil` will never happen, which makes this function rather unnecessary

is there something I am not seeing ? (if so then I think its more important to make sure that `mutationRequest` is only `nil` in case of error, because it wouldn't make sense otherwise)

pkg/controllers/dynakube/dynakube_controller.go
```diff
@@ -266,6 +284,7 @@ func (controller *Controller) reconcileDynaKube(ctx context.Context, dynakube *d
 		return err
 	}
 
+	log.Info("start reconciling connection info")
 	err = controller.reconcileConnectionInfo(ctx, dynakube, dynatraceClient)
```
0sewa0, line 288: **Main Suggestion** 👇 

I would just add a `controller.updateDynakubeStatus(ctx, dynaKube)` here, instead of all this convoluted logic, if we want to further relay on timely status refreshes, then we should start updating the status as soon as we can,

restarting the whole `Reconcile` does the same, but with many more steps

pkg/controllers/dynakube/dynakube_controller.go
```diff
@@ -329,23 +352,29 @@ func (controller *Controller) reconcileConnectionInfo(ctx context.Context, dynak
 }
 
 func (controller *Controller) reconcileComponents(ctx context.Context, dynatraceClient dtclient.Client, dynakube *dynatracev1beta1.DynaKube) error {
-	err := controller.reconcileActiveGate(ctx, dynakube, dynatraceClient)
+	// it's important to setup app injection before AG so that it is already working when AG pods start, in case code modules shall get
```
0sewa0, line 355: order change makes sense for now, but I would like to repeat my earlier question:
> would this kind of logic work in case we had different CR for activegate?

(considering "different CR" == "different Controller")

pkg/webhook/mutation/pod_mutator/request.go
```diff
@@ -26,15 +26,18 @@ func (webhook *podMutatorWebhook) createMutationRequestBase(ctx context.Context,
...
+	if err != nil {
+		if webhook.deployedViaOLM {
+			// in case of olm deployment, all pods are sent to us
+			// but not all of them need to be mutated,
+			// therefore their namespace might not have a dynakube assigned
+			// in which case we don't need to do anything
+			return nil, nil // nolint:nilerr
+		} else {
+			return nil, err
+		}
```
0sewa0, line 38: ❤️

0sewa0 (changes requested):

pkg/controllers/dynakube/dynakube_controller.go
StefanHauth: There is a dependency between AG and App monitoring in case AG shall be monitored with code modules. This has to be taken into account, when we split the controller up. At the moment I don't have an idea how to solve this, because when deploying AG one would have to check that app monitoring is operational and that connection infos are available at whatever location they are placed by then. And one would need to know if AG should be monitored with code modules in the first place (aka. read the related feature flag of a different CR).
TL;DR; it looks like that's gonna be a challenge, but one that's for sure out of scope in this PR.

pkg/controllers/dynakube/dynakube_controller.go
StefanHauth: Actually, I was working under the assumption, that a Reconile restart is the proper way to do this. If just writing back the updated status is sufficient, I'm more than happy to do it like that.

pkg/controllers/dynakube/connectioninfo/reconciler.go
```diff
@@ -36,19 +36,32 @@ func NewReconciler(clt client.Client, apiReader client.Reader, scheme *runtime.S
 	}
 }
 
+func (r Reconciler) updateDynakubeStatus(ctx context.Context) {
+	r.dynakube.Status.UpdatedTimestamp = metav1.Now()
+	err := r.client.Status().Update(ctx, r.dynakube)
+	if err != nil {
+		log.Error(err, "could not update dynakube status", "name", r.dynakube.Name)
+	}
+}
```
0sewa0, line 45: there is already a function like this ([its also private](https://github.com/Dynatrace/dynatrace-operator/blob/141ae33217b78bf6bdcd054a57cb94777f96313c/pkg/controllers/dynakube/dynakube_controller.go#L451-L451)), could we somehow reuse it 🤔 ?

also, shouldn't we return the error?

pkg/controllers/dynakube/connectioninfo/reconciler.go
```diff
@@ -36,19 +36,32 @@ func NewReconciler(clt client.Client, apiReader client.Reader, scheme *runtime.S
...
 
-	err := r.reconcileOneAgentConnectionInfo(ctx)
+	oneAgentConnectionInfoUpdated, err := r.reconcileOneAgentConnectionInfo(ctx)
 	if err != nil {
 		return err
 	}
 
+	if oneAgentConnectionInfoUpdated || activeGateConnectionInfoUpdated {
+		r.updateDynakubeStatus(ctx)
+	}
```
0sewa0, line 64: these `bool` values are always `true` if there was no error, so they are not that useful. 
- if there was no change and no error we would still update 

we could either:
- update the status always
- actually check if there was a change in the connectionInfo status
   - [how we do it in the dynakube-controller](https://github.com/Dynatrace/dynatrace-operator/blob/141ae33217b78bf6bdcd054a57cb94777f96313c/pkg/controllers/dynakube/dynakube_controller.go#L158-L158)

0sewa0 (changes requested):

pkg/webhook/mutation/pod_mutator/pod_mutator.go
```diff
@@ -66,7 +66,14 @@ func (webhook *podMutatorWebhook) Handle(ctx context.Context, request admission.
 	emptyPatch := admission.Patched("")
 	mutationRequest, err := webhook.createMutationRequestBase(ctx, request)
 	if err != nil {
-		return silentErrorResponse(mutationRequest.Pod, err)
+		emptyPatch.Result.Message = fmt.Sprintf("unable to inject into pod (err=%s)", err.Error())
+		log.Error(err, "building mutation request base encountered an error")
+		span.RecordError(err)
+		return emptyPatch
+	}
+	if mutationRequest == nil {
```
0sewa0, line 74: This is not a thing that can happen, if the `mutationRequest` is nil then we hit an error(so the if above will take over)

If we want to make sure, then write tests for it, but this is just dead code to me
```

File: pkg/controllers/dynakube/connectioninfo/reconciler.go [hunk 8]
```diff
@@ -49,7 +62,23 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
...
+
+	needStatusUpdate, err = hasher.IsDifferent(oldStatus.ActiveGate.ConnectionInfoStatus, r.dynakube.Status.ActiveGate.ConnectionInfoStatus)
+	if err != nil {
+		log.Error(err, "failed to compare ActiveGate connection info status hashes")
+		return err
+	} else if needStatusUpdate {
+		err = r.updateDynakubeStatus(ctx)
+	}
+	return err
```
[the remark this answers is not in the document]
StefanHauth: No, if the first status is different, it would trigger an update, that implicitly also updates the second one and than does a return, so the second status wouldn't even be checked.
But I agree to your suggestion anyway, it's cleaner.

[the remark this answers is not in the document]
0sewa0: > No, if the first status is different, it would trigger an update, that implicitly also updates the second one and than does a return,

aaaah, true 😵‍💫

pkg/webhook/mutation/pod_mutator/pod_mutator.go
StefanHauth: No, `createMutationRequestBase` has an path that explicitly does a `return nil, nil` in case of OLM stuff (https://github.com/Dynatrace/dynatrace-operator/blob/0202ee4c9736ab59e0a740f1072223f357b5d428/pkg/webhook/mutation/pod_mutator/request.go#L36)

0sewa0: oh, its that OLM special scenario 🤦 
would it make sense to have a more specific message in this case ? (the current message may be spooky)

also a sidenote: you added the error to the `span` there, wouldn't it be possible to update/upgrade the `silentErrorResponse` to add the span also, so we could reuse it ? 🤔 (maybe out of scope)

pkg/controllers/dynakube/dynakube_controller.go
```diff
@@ -118,12 +119,20 @@ func (controller *Controller) Reconcile(ctx context.Context, request reconcile.R
 
 	dynaKube, err := controller.getDynakubeOrUnmap(ctx, request.Name, request.Namespace)
 	if err != nil {
+		log.Info("reconciling DynaKube finished, could not get dynakube", "namespace", request.Namespace, "name", request.Name, "result", "empty", "err", err.Error())
```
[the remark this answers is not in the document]
StefanHauth: I added some logs to clearly see which part of the reconciliation has been started and finshed when. This can be really helpful when analysing timing related issues between processes (e.g. operator vs. webhook).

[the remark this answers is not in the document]
0sewa0: my main concern is/was the possible duplicated logging of errors, the rest are fine by me

go.sum
```diff
@@ -314,8 +314,6 @@ go.opencensus.io v0.22.4/go.mod h1:yxeiOL68Rb0Xd1ddK5vPZ/oVn4vY4Ynel7k9FzqtOIw=
 go.opencensus.io v0.22.5/go.mod h1:5pWMHQbX5EPX2/62yrJeAkowc+lfs/XD7Uxpq3pI6kk=
 go.opentelemetry.io/contrib/instrumentation/host v0.45.0 h1:1uzNKJDqZ6y6F5J6aKWgJjRREpKiGhBvKHlWon/bqB4=
 go.opentelemetry.io/contrib/instrumentation/host v0.45.0/go.mod h1:vlqPvzDsmB4+jlERxBRXsdLCD6Q0LoBzxHqNXp3qvG4=
-go.opentelemetry.io/contrib/instrumentation/host v0.46.0 h1:n2JE5k4ZtefmqXseW11OpvmYW0ouawMX2vjWemTGfCA=
-go.opentelemetry.io/contrib/instrumentation/host v0.46.0/go.mod h1:wJV0E+WkEumP6ph4Q3F+H6YOy2P/hwmKu4MbO+qj57I=
```
andriisoldatenko, line 318: i don't understand this change, why don't we need all 3 versions? 

https://github.com/Dynatrace/dynatrace-operator/blob/main/go.sum#L315-L320

🤔

andriisoldatenko: i fixed it in seperate PR https://github.com/Dynatrace/dynatrace-operator/pull/2386, so you can rebase after merging.

andriisoldatenko: I didn't get your point 🤷 what did I do manually?

pkg/controllers/dynakube/connectioninfo/reconciler.go
```diff
@@ -36,7 +37,19 @@ func NewReconciler(clt client.Client, apiReader client.Reader, scheme *runtime.S
 	}
 }
 
+func (r Reconciler) updateDynakubeStatus(ctx context.Context) error {
+	r.dynakube.Status.UpdatedTimestamp = metav1.Now()
+	err := r.client.Status().Update(ctx, r.dynakube)
+	if err != nil {
+		log.Error(err, "could not update dynakube status", "name", r.dynakube.Name)
+		return err
```
0sewa0, line 45: pls don't log the `err` if you are returning it because you will cause a double log:

you can either:
- do a `return errors.WithMessage(err, <message>)`
- use `log.Info` without using the `err`


[I update the style-guide PR with an example](https://github.com/Dynatrace/dynatrace-operator/pull/2383/files#diff-7c8612f5428f9d88cdd38edbaa31cd4d455a4d9736a3122efb3132444c180fc8R126-R149)

pkg/controllers/dynakube/connectioninfo/reconciler.go
```diff
@@ -49,7 +62,15 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 		return err
 	}
 
-	return nil
+	needStatusUpdate, err := hasher.IsDifferent(oldStatus, r.dynakube.Status)
+	if err != nil {
+		log.Error(err, "failed to compare connection info status hashes")
+		return err
```
0sewa0, line 68: same as above

pkg/controllers/dynakube/connectioninfo/reconciler.go
StefanHauth: fixed

pkg/controllers/dynakube/connectioninfo/reconciler.go
StefanHauth: fixed

andriisoldatenko:
practically LGTM, tested in my cluster 3 times in row.

0sewa0 (approved):

PR #2351, opened 2023-11-14
Add no-proxy to image library transport

luhi-DT:
## Description

Currently we have the problem that the no-proxy feature flag is not used for the image library traffic. Therefore its not possible to add NO_Proxy values for this kind of communication. Especially when having customImages it can lead to problems.

## How can this be tested?

Create a Dynakube with a Proxy and set NoProxy value to the target of the customImage, check if the image manifest can be accessed without accessing the proxy

## Checklist

### DIFF
--- a/pkg/oci/registry/client.go
+++ b/pkg/oci/registry/client.go
@@ -5,7 +5,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
-	"golang.org/x/net/http/httpproxy"
 	"net/http"
 	"net/url"
 
@@ -17,6 +16,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
+	"golang.org/x/net/http/httpproxy"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
--- a/pkg/oci/registry/client.go
+++ b/pkg/oci/registry/client.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"golang.org/x/net/http/httpproxy"
 	"net/http"
 	"net/url"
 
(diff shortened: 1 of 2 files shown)


pkg/oci/registry/client.go
```diff
@@ -164,18 +165,27 @@ func BuildImageIDWithTagAndDigest(taggedRef name.Tag, digest digest.Digest) stri
...
 	proxyUrl, err := url.Parse(proxy)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
 
-	transport.Proxy = func(req *http.Request) (*url.URL, error) {
-		return proxyUrl, nil
+	proxyConfig := httpproxy.Config{
+		HTTPProxy:  proxyUrl.String(),
+		HTTPSProxy: proxyUrl.String(),
```
chrismuellner, line 176: does this work correctly if we always set both http and https proxy independant of yhe actual protocol used?

chrismuellner (changes requested):

andriisoldatenko (approved):

pkg/oci/registry/client.go
andriisoldatenko: I think it's only about of incoming request, and it should work
https://cs.opensource.google/go/x/net/+/refs/tags/v0.18.0:http/httpproxy/proxy.go;l=127-135
```
func (cfg *config) proxyForURL(reqURL *url.URL) (*url.URL, error) {
	var proxy *url.URL
	if reqURL.Scheme == "https" {
		proxy = cfg.httpsProxy
	} else if reqURL.Scheme == "http" {
		proxy = cfg.httpProxy
		if proxy != nil && cfg.CGI {
			return nil, errors.New("refusing to use HTTP_PROXY value in CGI environment; see golang.org/s/cgihttpproxy")
		}
	}
	if proxy == nil {
		return nil, nil
	}
	if !cfg.useProxy(canonicalAddr(reqURL)) {
		return nil, nil
	}

	return proxy, nil
}

```

no matter if your proxy string has http/or https.

luhi-DT: yes, the proxyFunc() uses https incase the proxy passed to it has https:// prefix and uses http if it has http:// prefix

chrismuellner (approved):

PR #2360, opened 2023-11-15
new PR

andriisoldatenko:
## Description

Please include the following:

- A summary of the change

## How can this be tested?

## Checklist

### DIFF
--- a/pkg/util/kubeobjects/object_test.go
+++ b/pkg/util/kubeobjects/object_test.go
@@ -1,0 +1,25 @@
+package kubeobjects
+
+import (
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
+)
+
+func TestKey(t *testing.T) {
+	type args struct {
+		object client.Object
+	}
+	tests := []struct {
+		name string
+		args args
+		want client.ObjectKey
+	}{
+		{name: "tttttt", args: args{object: nil}, want: client.ObjectKey{}},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equalf(t, tt.want, Key(tt.args.object), "Key(%v)", tt.args.object)
+		})
+	}
+}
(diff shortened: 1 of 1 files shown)


pkg/util/kubeobjects/object_test.go
```diff
@@ -0,0 +1,25 @@
...
+func TestKey(t *testing.T) {
+	type args struct {
+		object client.Object
+	}
+	tests := []struct {
+		name string
+		args args
+		want client.ObjectKey
+	}{
+		{name: "tttttt", args: args{object: nil}, want: client.ObjectKey{}},
```
0sewa0, line 18:
```suggestion
		{name: "handle nil value", args: args{object: nil}, want: client.ObjectKey{}},
```

I don't really understand the test, but it kinda looks like a nil check test, would this name not be better?

0sewa0 (changes requested):

0sewa0 (commented):
This is a test

pkg/util/kubeobjects/object_test.go
```diff
@@ -0,0 +1,25 @@
+package kubeobjects
```
0sewa0, line 1: This is a test

pkg/util/kubeobjects/object_test.go
andriisoldatenko: fixed in f984e317

pkg/util/kubeobjects/object_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
+)
+
+func TestKey(t *testing.T) {
+	type args struct {
+		object client.Object
+	}
+	tests := []struct {
+		name string
```
0sewa0, line 14:
```suggestion
		title string
```

pkg/util/kubeobjects/object_test.go
andriisoldatenko: reolied.

pkg/util/kubeobjects/object_test.go
andriisoldatenko: ldkfajsdlfkjalkdsfj

0sewa0 (approved):

PR #2375, opened 2023-11-16
Add validation for CR name length

0sewa0:
## Description

Please include the following:

It can be a problem when the name of the CR is too long.
This is because we use the name of the CR as a base for other resources we create that are related to it, this is further complicated by the fact that some resource names are used as label values, which have a limit of 63 characters.
- So name-of-CR + suffix + hash-given-by-kubernetes should be shorter then 63 characters

## How can this be tested?
Try to create a CR (dynakube or edgeconnect) with a long name (longer then 40 chars)
(the main thing that should be checked is the message it self 😅 )

## Checklist

- [x] Unit tests have been updated/added
- [x] PR is labeled accordingly with a single label
- [x] I have read and understood the [contribution guidelines](https://github.com/Dynatrace/dynatrace-operator/blob/main/CONTRIBUTING.md)

PR #2387, opened 2023-11-20
Shorten test dynakube name to pass validation

0sewa0:
## Description

[In a recent PR](https://github.com/Dynatrace/dynatrace-operator/pull/2375) we added validation for the length of dynakube names, and as it turns out 1 of our e2e tests had a dynakube that was rejected because of it.

Also added some missing e2e make targets

## How can this be tested?
Check if the new e2e targets work. (only 2. one had problem with name length validation)
```
make test/e2e/cloudnative/codemodules-with-proxy
make test/e2e/cloudnative/codemodules-with-custom-ca
```

## Checklist


hack/make/tests/e2e.mk
albertogdd, line 1: Let's also remove `test/e2e/cloudnative/proxy` from the make commands, or?

0sewa0: they are different tests, but it might be worth to look what is the difference, and just have 1 🤔 

`test/e2e/cloudnative/proxy` -> [this](https://github.com/Dynatrace/dynatrace-operator/blob/518088d78ff36b3710da11348f6d628fbdd72cab/test/features/cloudnative/proxy/proxy.go#L31-L31)

`test/e2e/cloudnative/codemodules-with-proxy` ->[that](https://github.com/Dynatrace/dynatrace-operator/blob/518088d78ff36b3710da11348f6d628fbdd72cab/test/features/cloudnative/codemodules/codemodules.go#L113-L113)

albertogdd: I was just wondering about it! but no need to do it in this PR

albertogdd (approved):
E2E passed locally

aorcholski (approved):

PR #2402, opened 2023-11-24
Only check the actual activegate container for mounts in e2e test

0sewa0:
## Description

For some reason this used to work, don't really know how.
But in this test case it can be that we inject the activegate with istio, which adds a sidecar, so directly accessing the `Container` array is risky 😅 

## How can this be tested?

`make test/e2e/activegate/proxy` passes

## Checklist


chrismuellner (approved):

luhi-DT (approved):

StefanHauth (approved):
LGTM

PR #2413, opened 2023-11-28
WIP: Add script for automatic e2e test docs generation

albertogdd:
## Description

Please include the following:

- A summary of the change

## How can this be tested?

## Checklist


waodim (commented):
Some thoughts:

- I am not sure about the placement of the file inside hack. Didn't we say, we want a dedicated doc(s) package one level above? I may be wrong there. I think @andriisoldatenko knows more there.
- I think you forgot to add the feature flags to the documentation, as far as i can see there are only e2e tests docu
- Would it be a good idea to maybe add a step where the tool gomarkdoc is downloaded and installed? For the case that one wants to run the make targets but does not have the tool installed?

.github/pre-commit
```diff
@@ -8,6 +8,10 @@
```
waodim, line 8: > Some thoughts:

- I am not sure about the placement of the file inside hack. Didn't we say, we want a dedicated doc(s) package one level above? I may be wrong there. I think @andriisoldatenko knows more there.
- I think you forgot to add the feature flags to the documentation, as far as i can see there are only e2e tests docu
- Would it be a good idea to maybe add a step where the tool gomarkdoc is downloaded and installed? For the case that one wants to run the make targets but does not have the tool installed?

I am quoting this here such that you guys can also comment :sweat_smile:

albertogdd: - Moved scripts to ./doc/script as mentioned by you guys :+1: 
- Added script for feature_flags.go documentation efedfe13542ff8b80dd9dc815609578100e206c4
- Make command for installing gomarkdoc is already in place! `make prerequisites/gomarkdoc`

.github/pre-commit
```diff
@@ -8,6 +8,10 @@
 
 echo Shell: "${SHELL}"
 
+echo Updating auto-generated markdown files...
+make doc/gen-gomarkdoc
+git add ./doc
```
andriisoldatenko, line 13: I'm not sure that makes lots of sense ....  running `make...` is okay, but git add is redundant. Git will force you to add some uncommitted changes anyway. 

Also from another side not sure that we want to run it every commit.

doc/e2e-features.md
```diff
@@ -0,0 +1,156 @@
+<!-- Code generated by doc/script/gen_e2e_features.sh - DO NOT EDIT -->
+
+# activegate
```
andriisoldatenko, line 3: it should be `e2e` subfolder

.github/pre-commit
waodim: Yeah I mean it depends if we want to run it every commit. What about if we run it every release, then we could add it to the github actions pipeline and it would run for every release

albertogdd: This, I just discussed with andrii about it and I will update the CI validation :+1:

albertogdd: Done!

andriisoldatenko (approved):

waodim (dismissed):
LGTM

andriisoldatenko (approved):

PR #2419, opened 2023-11-29
Make cloud native network zone E2E test less fragile

StefanHauth:
## Description

The E2E test for extended network zone support has been unstable and error prone in case of parallel test runs. Make it more stable by 
* using random network zone names for each test run
* make sure that the network zone is correctly created before doing any other test steps

## How can this be tested?

* Enable network zone support on your tenant
* run `make test/e2e/cloudnative/network-zone` multiple times (min. 3x)

## Checklist


test/features/cloudnative/network_zones/name.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"crypto/rand"
+)
+
+const prefix = "op-e2e-"
+const defaultName = "test-network-zone"
+
+// Use a randomized network zone names, to avoid problems with improper or slow cleanup of network zones on DT cluster side.
+// With randomized names a fresh start is guaranteed for each test run.
+func getNetworkZoneName() string {
```
andriisoldatenko, line 12: small nitpick: i was wondering if we need something similar in another tests, wdyt to move this code to helpers, and import there like getRandomBytes(length int). and reuse inside getNetworkZoneNames?

test/features/cloudnative/network_zones/name.go
```diff
@@ -0,0 +1,31 @@
...
+
+	b := make([]byte, length)
+	_, err := rand.Read(b)
+	if err != nil {
+		return prefix + defaultName
+	}
+	return prefix + encode(b)
+}
+
+func encode(randBytes []byte) string {
```
andriisoldatenko, line 23: same here

test/features/cloudnative/network_zones/network_zones.go
```diff
@@ -53,13 +52,17 @@ func Feature(t *testing.T) features.Feature {
 	builder.WithLabel("name", "cloudnative-network-zone")
 	secretConfig := tenant.GetSingleTenantSecret(t)
 
+	networkZone := getNetworkZoneName()
```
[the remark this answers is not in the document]
andriisoldatenko: Okay we need a third opinion here cc @0sewa0 ?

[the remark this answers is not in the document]
0sewa0: but we are talking implementation details here 😄 

if you give me a deterministic Network zone name, based on some factors, which can be easily reproduced, then you wouldn't need to name it `Random`, because its not that.
- Example: For a given dynakube, your function would return a networkzone name, lets say based on the name of the dynakube and what mode it is in. then `getNetworkZoneName(dk)` is totally fine.

but, when you functions use the `rand` package, then it is kinda important for the user to know, that it will be random.
+1: We are not talking about an external API used as a library somewhere else, where it might be actually irrelevant, but an internal function, used for e2e tests, I wanna know what exactly the e2e tests are doing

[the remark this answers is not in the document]
andriisoldatenko: ☝️

andriisoldatenko:
Practically LGTM. tested in my cluster.

andriisoldatenko (approved):

test/helpers/rand/name.go
```diff
@@ -0,0 +1,55 @@
+package rand
```
0sewa0, line 1: don't we want to have `//go:build e2e` here aswell, like in all e2e testing packages 🤔

StefanHauth: true, fixed it

PR #2426, opened 2023-11-30
Use local digest for signing the docker images

gkrenn:
## Description

This PR changes the release pipeline to reuse the digests that is stored locally during build-time, instead of using the one from dockerhub.

## How can this be tested?

Run the release pipeline, should work as before

## Checklist


.github/workflows/release.yaml
```diff
@@ -223,8 +225,9 @@ jobs:
         id: operator-digest
         env:
           IMAGE: ${{ matrix.url }}/${{ secrets[matrix.repository] }}:${{ needs.prepare.outputs.version }}
+          ONLINE: true
```
andriisoldatenko, line 228: can we use more clear variable name to highlight that we use `docker://"` or `"docker-daemon:"`?

andriisoldatenko (changes requested):

.github/workflows/release.yaml
andriisoldatenko: https://github.com/containers/skopeo/blob/main/docs/skopeo.1.md
>docker://docker-reference An image in a registry implementing the "Docker Registry HTTP API V2". By default, uses the authorization state in either $XDG_RUNTIME_DIR/containers/auth.json, which is set using (skopeo login). If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using (docker login).

docker-archive:path[:docker-reference] An image is stored in the docker save formatted file. docker-reference is only used when creating such a file, and it must not contain a digest.

docker-daemon:docker-reference An image docker-reference stored in the docker daemon internal storage. docker-reference must contain either a tag or a digest. Alternatively, when reading images, the format can be docker-daemon:algo:digest (an image ID).

andriisoldatenko: `IS_DOCKER_REGISTRY or `IS_DOCKER_DAEMON`

andriisoldatenko (approved):

PR #2438, opened 2023-12-04
Move test/features/README.md doc to code

albertogdd:
## Description

E2E doc is now automatically generated by ./hack/doc/gen_e2e_features.sh

Hence, with this PR, I removed the old README file and moved its content to the actual code.

## How can this be tested?

`make doc` runs the script that generates the docs.

CI checks that there is no diff uncommitted.

## Checklist


test/features/applicationmonitoring/label_version_detection.go
```diff
@@ -90,6 +90,14 @@ var (
 	}
 )
 
+/*
```
andriisoldatenko, line 93: The go way to comment multi lines is:
```
// Package path implements utility routines for manipulating slash-separated
// paths.
//
// The path package should only be used for paths separated by forward
// slashes, such as the paths in URLs. This package does not deal with
// Windows paths with drive letters or backslashes; to manipulate
// operating system paths, use the [path/filepath] package.
package path
```


could you please change it?

andriisoldatenko (changes requested):

test/features/applicationmonitoring/label_version_detection.go
albertogdd: Fixed in b969217d173736eb1cb514aa49225a78e49edb5c

andriisoldatenko (approved):

PR #2468, opened 2023-12-11
Version Reconciler Refactoring Approach

wepudt:
## Description

we want to do a refactoring to simplify the image version reconciliation of the OneAgent, ActiveGate and CodeModule images. To do so I explored the code to come up with a draft solution to achieve these goals.

## How can this be tested?

It's a draft and not intended to work with tests

## Checklist

~- [ ] Unit tests have been updated/added~


pkg/controllers/dynakube/version/oneAgentVersionReconciler.go
```diff
@@ -0,0 +1,212 @@
...
+
+	downgrade, err := checkForDowngrade(previous, latestVersion)
+	if err != nil || downgrade {
+		return previous, err
+	}
+	imageUri, err := getImageUri(dynakube, reconciler.dtClient.GetLatestOneAgentImage, dynakube.DefaultOneAgentImage())
+	if err != nil {
+		return previous, err
+	}
+	return updateVersionWithImage(ctx, reconciler.registryClient, current, imageUri)
```
0sewa0, line 164: `updateVersionWithImage` uses the digest in the `ImageID` which should not be used with the tenant registry (the digest changes there overtime)

pkg/controllers/dynakube/version/activeGateVersionReconciler.go
```diff
@@ -0,0 +1,96 @@
...
+			log.Info("status updated skipped, due to no auto update", "updater", "activegate")
+			return previous, nil
+		}
+	}
+
+	imageUri, err := getImageUri(dynakube, reconciler.dtClient.GetLatestActiveGateImage, dynakube.DefaultActiveGateImage())
+	if err != nil {
+		return previous, err
+	}
+	return updateVersionWithImage(ctx, reconciler.registryClient, current, imageUri)
```
0sewa0, line 95: you are not checking here about the public-registry, which would be nice if it worked but

`updateVersionWithImage` uses the digest in the ImageID which should not be used with the tenant registry (the digest changes there overtime)

pkg/controllers/dynakube/version/oneAgentVersionReconciler.go
```diff
@@ -0,0 +1,212 @@
...
+		} else if dynakube.FeatureDisableReadOnlyOneAgent() {
+			return errors.New("immutable OneAgent image in combination with readOnly OneAgent filesystem is not possible")
+		}
+	}
+	log.Info("OneAgent metadata present, image type and image version validated")
+	return nil
+}
+
+func (reconciler *OneAgentImageVersionReconciler) SetOneAgentHealthConfig(ctx context.Context, dynakube *dynatracev1beta1.DynaKube) {
+	healthConfig, err := GetOneAgentHealthConfig(ctx, reconciler.apiReader, reconciler.registryClient, dynakube, dynakube.OneAgentImage())
```
0sewa0, line 79: this is more of a sidenote and out of scope probably: To get the `HealthConfig` we need to query the image registry, the "same way" as when you call `updateVersionWithImage`, so it would be nice that we do not make the "same" call twice **if possible**

0sewa0 (commented):
I like the general ideas, like moving some checks to the `VersionStatus` struct, and init-ing the "current" `VersionStatus` at the beginning and doing checks against it.

The main problem is still the requirements themself, the weird differences between sources (the difference between the tenant registry and normal image registries).
Most weirdness/confusion stems from this.

While I believe that this would be an improvement compared to our current implementation, however I do not know if its worth to invest more time into it, to implement it for every "version reconciler". 😬
And hopefully we can change the requirements themself, so the logic can be properly simplified/refactored.

wepudt:
> I like the general ideas, like moving some checks to the `VersionStatus` struct, and init-ing the "current" `VersionStatus` at the beginning and doing checks against it.
> 
> The main problem is still the requirements themself, the weird differences between sources (the difference between the tenant registry and normal image registries). Most weirdness/confusion stems from this.
> 
> While I believe that this would be an improvement compared to our current implementation, however I do not know if its worth to invest more time into it, to implement it for every "version reconciler". 😬 And hopefully we can change the requirements themself, so the logic can be properly simplified/refactored.

What cought my eye was that we determine the source type and have multiple different differences downstream. One improvement might be to store the old settings, but that might complicate changes down the line.

one other thing: I was not always certain that reordering the sequences changes a lot. so perhaps have some final checks for OneAgent (preventing downgrades, ImageType) might streamline the code to a point where we just have to deal with some minor adjustments.

pkg/controllers/dynakube/version/oneAgentVersionReconciler.go
wepudt: I am sorry- that was a letdown on my part: I did not rename the function! IN the function i check for tenant registry and handle the digest. that way kind a do not have to split the tracks beforehand...

pkg/controllers/dynakube/version/oneAgentVersionReconciler.go
andriisoldatenko: good point ☝️

pkg/controllers/dynakube/version/oneAgentVersionReconciler.go
```diff
@@ -0,0 +1,212 @@
...
+
+func checkForDowngrade(previous status.VersionStatus, latestVersion string) (bool, error) {
+	imageID := previous.ImageID
+	if imageID == "" {
+		return false, nil
+	}
+
+	var previousVersion string
+	var err error
+	switch previous.Source {
```
andriisoldatenko, line 202: I think the `switch case` here is a bit redundant since you're not using default, and you have only 2 branches.
Just simple if and else if should work better IMO.

pkg/controllers/dynakube/version/oneAgentVersionReconciler.go
```diff
@@ -0,0 +1,212 @@
...
+	latestVersion := dynakube.CustomOneAgentVersion()
+	var err error
+	if latestVersion == "" {
+		latestVersion, err = reconciler.dtClient.GetLatestAgentVersion(dtclient.OsUnix, dtclient.InstallerTypeDefault)
+		if err != nil {
+			return previous, err
+		}
+	}
+
+	downgrade, err := checkForDowngrade(previous, latestVersion)
```
andriisoldatenko, line 156: can we name it in same manner? E.g. `previous`and `latest` or with `Version` suffix both variables?

andriisoldatenko (changes requested):
General note: 
IMO we should start with simple design interface, then we can also create some types which implements it and we can share some common logic or at least have same structure.

PR #2470, opened 2023-12-11
Make each section independent for error handling 

0sewa0:
## Description

The idea of this change is to deploy everything we can for each section.
So if you configure OneAgent + ActiveGate, but there is a problem in the OneAgent section we still deploy what is configured in the ActiveGate section, and vice versa.

To do this some reconcilers had to be modified a bit so they do not have a single `Reconcile` function, but several.
- There could have been a "nicer" way to go about it, but that would require to totally refactor ~4 packages, which I would like to avoid doing in a single go.

Things that still needs to be done or discussed:
- ~Is my approche acceptable, or do we want 1 huge refactor 😅 ?~
- ~I added a bunch of TODOs to places where I saw inconsistencies, we should discuss them and probably handle them separately~ Will create follow ups for the TODOs I added
- ~I only updated the tests so they pass, but not really want to make them "nicer" before I know which way we are going~ Making the tests nicer is not really possible at this scope, added the most relevant ones
- ~my names are not great, but good luck finding much better ones 😅~
- ~split the `dynakube_controller` to different files, as it was a bit of a mess (still kinda is)~

## How can this be tested?
One way:
Deploy a CloudNative + ActiveGate dynakube where the image field for the ActiveGate is pointing to an inaccessible image.
No ActiveGate should show up but OneAgent will be deployed.

## Checklist


pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -41,7 +41,7 @@ func (r *Reconciler) ReconcileAPIUrl(ctx context.Context, dynakube *dynatracev1b
 	return nil
 }
 
-func (r *Reconciler) ReconcileCommunicationHosts(ctx context.Context, dynakube *dynatracev1beta1.DynaKube) error {
+func (r *Reconciler) ReconcileCMCommunicationHosts(ctx context.Context, dynakube *dynatracev1beta1.DynaKube) error {
```
andriisoldatenko, line 44: It's not clear what's `CM`?

pkg/controllers/dynakube/version/reconciler.go
```diff
@@ -34,12 +34,33 @@ func NewReconciler(dynakube *dynatracev1beta1.DynaKube, apiReader client.Reader,
...
+	}
+	return nil
+}
+
+func (reconciler *Reconciler) ReconcileOA(ctx context.Context) error {
+	updaters := []StatusUpdater{
+		newOneAgentUpdater(reconciler.dynakube, reconciler.apiReader, reconciler.dtClient, reconciler.registryClient),
+	}
+
+	neededUpdaters := reconciler.needsReconcile(updaters)
```
andriisoldatenko, line 54: The wording is confusing to me :) from the right side `needs` in the left `needed`?

pkg/controllers/dynakube/version/reconciler.go
```diff
@@ -34,12 +34,33 @@ func NewReconciler(dynakube *dynatracev1beta1.DynaKube, apiReader client.Reader,
...
+	if len(neededUpdaters) > 0 {
+		return reconciler.updateVersionStatuses(ctx, neededUpdaters)
+	}
+	return nil
+}
+
+func (reconciler *Reconciler) ReconcileAG(ctx context.Context) error {
```
andriisoldatenko, line 61: I prefer to name it without abbreviations: `ReconcileActiveGate`?

pkg/controllers/dynakube/connectioninfo/reconciler.go
```diff
@@ -47,16 +47,25 @@ func (r Reconciler) updateDynakubeStatus(ctx context.Context) error {
...
+	needStatusUpdate, err := hasher.IsDifferent(oldStatus, r.dynakube.Status)
+	if err != nil {
+		return errors.WithMessage(err, "failed to compare connection info status hashes")
+	} else if needStatusUpdate {
+		err = r.updateDynakubeStatus(ctx)
+	}
+	return err
+}
 
+func (r *Reconciler) ReconcileOA(ctx context.Context) error {
```
andriisoldatenko, line 67:
```suggestion
func (r *Reconciler) ReconcileOoneAgent(ctx context.Context) error {
```
🙏 🙏 🙏

andriisoldatenko (changes requested):

pkg/controllers/dynakube/connectioninfo/reconciler.go
wepudt: just updated the comment to remove a small typo

wepudt: I would prefer several small steps instead of a big-bang change, but only if it is feasible and we do not get side-effects from incrementally changing the code...

pkg/controllers/dynakube/activegate.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/apimonitoring"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/connectioninfo"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/istio"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/version"
+	"github.com/pkg/errors"
+)
+
+func (controller *Controller) reconcileActiveGate(ctx context.Context, dynakube *dynakube.DynaKube, dtc dynatrace.Client, istioReconciler *istio.Reconciler, connectionReconciler *connectioninfo.Reconciler, versionReconciler *version.Reconciler) error { // nolint: revive
+	if dynakube.NeedsActiveGate() { // TODO: this is not optimal, because this check is in the activegate reconciler as well (to do the cleanup)
+		err := connectionReconciler.ReconcileAG(ctx)
```
waodim, line 18:
```suggestion
		err := connectionReconciler.ReconcileActiveGate(ctx)
```

waodim: Out of consistency because the suggestion was made to rename `ReconcileOA` to `ReconcileOneAgent`

pkg/controllers/dynakube/version/reconciler_test.go
```diff
@@ -107,21 +107,25 @@ func TestReconcile(t *testing.T) {
...
 		previousProbe := *dkStatus.CodeModules.VersionStatus.LastProbeTimestamp
-		err = versionReconciler.Reconcile(ctx)
+		err = versionReconciler.ReconcileCM(ctx)
 		require.NoError(t, err)
 		assert.Equal(t, previousProbe, *dkStatus.CodeModules.VersionStatus.LastProbeTimestamp)
 
 		// change if probe old enough
 		changeTime(timeProvider, 15*time.Minute+1*time.Second)
-		err = versionReconciler.Reconcile(ctx)
+		err = versionReconciler.ReconcileCM(ctx)
```
waodim, line 128: same here out of consistency we might want to write out the name instead of abbreviating

waodim (dismissed):

pkg/controllers/dynakube/istio/reconciler.go
0sewa0: CM == CodeModules (but I mentioned that the names are still not the best 😅 )

pkg/controllers/dynakube/version/reconciler.go
0sewa0: this is not "new new", I just moved things around here, but I agreed 😅

0sewa0: trying to come up with a different name, but I am struggling 😅 

`needsReconcile(updaters)`  gives back those `updaters` that can/need-to be run.
- this is not really the nice way to do it considering I split up each updater (except for 1 instance) into its separate `Reconcile` function, but I want to minimise the refactoring, so we can tackle that later.
    - so we could probably just call [needsUpdate](https://github.com/Dynatrace/dynatrace-operator/blob/2c8bb29f220888601e0a82e1c4da5dc14a8fc8c3/pkg/controllers/dynakube/version/reconciler.go#L105-L105) on the 1 updater per `Reconcile` function

pkg/controllers/dynakube/istio/reconciler.go
0sewa0: done

pkg/controllers/dynakube/version/reconciler.go
0sewa0: done

pkg/controllers/dynakube/version/reconciler.go
0sewa0: done

pkg/controllers/dynakube/connectioninfo/reconciler.go
0sewa0: done

pkg/controllers/dynakube/activegate.go
0sewa0: done

pkg/controllers/dynakube/version/reconciler_test.go
0sewa0: done

andriisoldatenko (commented):
let me know when it's ready for final review

pkg/controllers/dynakube/status/status.go
```diff
@@ -8,7 +8,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func SetDynakubeStatus(ctx context.Context, dynakube *dynatracev1beta1.DynaKube, apiReader client.Reader) error {
+func SetKubeSystemUUIDInStatus(ctx context.Context, dynakube *dynatracev1beta1.DynaKube, apiReader client.Reader) error {
```
wepudt, line 11: imho we should have the same identifier everywhere- I would prefer SetDynakubeSystemUUIDInStatus here

0sewa0: I think its redundant to say `Dynakube`, also its the `KubeSystemUUID` we want to add to the `Dynakube`'s status, so if we want to be super explicit then we would need to say something like `SetKubeSystemUUIDInDynakubeStatus` which is a mouthful 😅

wepudt: true from that perspective this makes sense.

pkg/controllers/dynakube/activegate.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/activegate"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/apimonitoring"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/connectioninfo"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/istio"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/version"
+	"github.com/pkg/errors"
+)
+
+func (controller *Controller) reconcileActiveGate(ctx context.Context, dynakube *dynakube.DynaKube, dtc dynatrace.Client, istioReconciler istio.Reconciler, connectionReconciler connectioninfo.Reconciler, versionReconciler version.Reconciler) error { // nolint: revive
+	if dynakube.NeedsActiveGate() { // TODO: this is not optimal, because this check is in the activegate reconciler as well (to do the cleanup)
```
andriisoldatenko, line 17:
```suggestion
	if dynakube.NeedsActiveGate() { // TODO: this is not optimal because this check is in the ActiveGate reconciler as well (to do the cleanup)
```

andriisoldatenko:
practically LGTM:

I see error for ActiveGate image:
```
{"level":"info","ts":"2024-01-03T09:22:46.750Z","logger":"dynakube","msg":"reconciling DynaKube finished","namespace":"dynatrace","name":"dy │
│ nakube","result":{"Requeue":false,"RequeueAfter":0}}                                                                                         │
│     sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:227","ts":"2024-01-03T09:22:27.087Z"}{"error":"getting refe │
│ rence \"dummy\": GET https://index.docker.io/v2/library/dummy/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Clas │
│ s: Name:library/dummy Type:repository]]","level":"error","logger":"dynakube","msg":"error reconciling DynaKube","name":"dynakube","namespace │
│ ":"dynatrace","stacktrace":"github.com/Dynatrace/dynatrace-operator/pkg/util/logger.logSink.Error                                            │
│     github.com/Dynatrace/dynatrace-operator/pkg/util/logger/logger.go:40                                                                     │
│ github.com/go-logr/logr.Logger.Error                                                                                                         │
│     github.com/go-logr/logr@v1.3.0/logr.go:305    
```

but I see oneAgent:

image

andriisoldatenko (approved):
LGTM, i left one no blocking nit-pick.

PR #2481, opened 2023-12-13
Enhance support archive to collect CRD and WebhookConfig manifests

waodim:
## Description

Support archive is enhanced in a way such that it now also collects manifests of `CRDs` (`edgeconnect` and `dynakube`), `MutatingWebhookConfiguration` and `ValidatingWebhookConfiguration`. 

Some things to consider: The other resources are all queried and listed by an api call. This is unfortunately not possible with `CRDs` and `WebhookConfigs` (they are cluster wide objects), because of missing permission in the role of the operator. This is why I had to rethink my approach and decided to just get them with an api call (because for that we have necessary permissions). That is why I added `edgeconnect` to the operator role. So in summary, I simply get those objects and convert them to an `unstructured.Unstructured` type and store them.

## How can this be tested?

- Unit tests
- make test/e2e/supportarchive
- deploy operator and run support archive subcommand 

## Checklist


cmd/support_archive/resources.go
```diff
@@ -7,13 +7,21 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 )
 
-const k8sResourceCollectorName = "k8sResourceCollector"
+const (
+	k8sResourceCollectorName = "k8sResourceCollector"
+	webhookValidatorName     = "dynatrace-webhook"
+	dynakubeCRDName          = "dynakubes.dynatrace.com"
+	edgeconnectCRDName       = "edgeconnects.dynatrace.com"
```
andriisoldatenko, line 23: just random thoughts: is it possible to collect all CRDS by api-group?
```
kubectl api-resources --api-group=dynatrace.com
NAME           SHORTNAMES   APIVERSION               NAMESPACED   KIND
dynakubes                   dynatrace.com/v1beta1    true         DynaKube
edgeconnects                dynatrace.com/v1alpha1   true         EdgeConnect
```

andriisoldatenko: why is it important, as soon as we split dynakube resource, we need to add each new resource here, instead of get all our CRDS.

cc @0sewa0 wdyt?

waodim: thanks for the input, i will need to pass clientset to the k8scollector struct in order to do this. I will try because I think this is a nice solution.

test/features/support_archive/files.go
```diff
@@ -5,6 +5,7 @@ package support_archive
 import (
```
albertogdd, line 5: When trying to run the e2e tests
```sh
❯ make test/e2e/supportarchive --debug
GNU Make 4.3
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Reading makefiles...
Updating makefiles....
Updating goal targets....
 File 'test/e2e/supportarchive' does not exist.
   File 'manifests/crd/helm' does not exist.
     File 'prerequisites/kustomize' does not exist.
    Must remake target 'prerequisites/kustomize'.
go install "sigs.k8s.io/kustomize/kustomize/v5@v5.3.0"
    Successfully remade target file 'prerequisites/kustomize'.
     File 'helm/version' does not exist.
    Must remake target 'helm/version'.
    Successfully remade target file 'helm/version'.
     File 'manifests/crd/generate' does not exist.
       File 'prerequisites/controller-gen' does not exist.
      Must remake target 'prerequisites/controller-gen'.
go install "sigs.k8s.io/controller-tools/cmd/controller-gen@v0.13.0"
      Successfully remade target file 'prerequisites/controller-gen'.
    Must remake target 'manifests/crd/generate'.
/home/alberto/go/bin/controller-gen "crd:crdVersions=v1" paths="./..." output:crd:artifacts:config=config/crd/bases
    Successfully remade target file 'manifests/crd/generate'.
  Must remake target 'manifests/crd/helm'.
./hack/helm/generate-crd.sh /home/alberto/go/bin/kustomize config/helm/chart/default//templates//Common/crd/ config/deploy/
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
  Successfully remade target file 'manifests/crd/helm'.
Must remake target 'test/e2e/supportarchive'.
go test -v -tags "osusergo,netgo,sqlite_omit_load_extension,e2e," -timeout 20m -count=1  ./test/scenarios/standard -args --labels "name=support-archive"
F1214 14:46:34.322841  168239 env.go:375] Setup failure: exit status 2
FAIL	github.com/Dynatrace/dynatrace-operator/test/scenarios/standard	3.784s
FAIL
make: *** [hack/make/tests/e2e.mk:101: test/e2e/supportarchive] Error 1
```

aorcholski: remove `dynatrace` namespace before running the test

albertogdd: Passing now! Thanks

cmd/support_archive/resources.go
waodim: I did it now that way that I list the CRDs by api-group. Everything works, expect for the unit tests. There we have the problem that we probably need to mock this call to the api-group, which I am currently trying to do. Do you guys have a better solution there?

For example: https://github.com/Dynatrace/dynatrace-operator/blob/main/cmd/support_archive/resources_test.go#L134

Here I am adding the kube config, and this results then in the test failing because of refused connection to localhost:80. Should I try to fix these tests or should I stick to my other approach where I queried every CRD with .get() and not via api-group? 

The benefit would be that we would not need to rework the unit tests to such an extent, the downside would be that we would add a new api call for every new CRD that we add. 

What do you guys think?

andriisoldatenko: if you show us the problem we can help you to mock it :)

waodim: pushed the changes in https://github.com/Dynatrace/dynatrace-operator/pull/2481/commits/c0fe2a95ffbb8275c0ed9a28818a7525c3e59ef9

waodim: fixed in https://github.com/Dynatrace/dynatrace-operator/pull/2481/commits/7424836288a062b611295c5c7781b13cce58c5f6

andriisoldatenko (approved):

PR #2491, opened 2023-12-14
Remove deprecated feature flags

gkrenn:
## Description

When switching to version 1.0 deprecated feature flags will be removed.
This PR removes them.

Removed feature flags: 
```
dynatracev1beta1.AnnotationFeatureWebhookReinvocationPolicy,
dynatracev1beta1.AnnotationFeatureReadOnlyOneAgent,
dynatracev1beta1.AnnotationFeatureHostsRequests,
dynatracev1beta1.AnnotationFeatureActiveGateAuthToken,
dynatracev1beta1.AnnotationFeatureActiveGateRawImage,
dynatracev1beta1.AnnotationFeatureActiveGateReadOnlyFilesystem,
```

## How can this be tested?

- Run operator => nothing should have changed. 
- Removed feature flags should no longer be taken into account

## Checklist


albertogdd:
Question: Are we then completely removing the use cases of these feature flags? 
I'm asking because I'm working on the new Activegate CRD spec, and if so, I shouldn't include these.

pkg/api/v1beta1/dynakube/feature_flags.go
```diff
@@ -38,15 +38,10 @@ const (
```
albertogdd, line 38: Question: Are we then completely removing the use cases of these feature flags? 
I'm asking because I'm working on the new Activegate CRD spec, and if so, I shouldn't include these.

gkrenn: They should be removed completely.

gkrenn: [Here](https://github.com/Dynatrace/dynatrace-operator/blob/010623f6bdc6601856fffa71b44944a7dd2ee925/pkg/webhook/validation/dynakube/deprecation.go#L27) you can see what should be moved

pkg/api/v1beta1/dynakube/properties.go
```diff
@@ -228,18 +228,15 @@ func (dk *DynaKube) PullSecretWithoutData() corev1.Secret {
 
 func (dk *DynaKube) NeedsReadOnlyOneAgents() bool {
 	inSupportedMode := dk.HostMonitoringMode() || dk.CloudNativeFullstackMode()
-	return inSupportedMode && !dk.FeatureDisableReadOnlyOneAgent()
+	return inSupportedMode
```
0sewa0, line 231:
```suggestion
	return dk.HostMonitoringMode() || dk.CloudNativeFullstackMode()
```
would be enough

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/authtoken.go
```diff
@@ -25,7 +25,7 @@ type AuthTokenModifier struct {
 }
 
 func (mod AuthTokenModifier) Enabled() bool {
-	return mod.dynakube.UseActiveGateAuthToken()
+	return mod.dynakube.NeedsActiveGate()
```
0sewa0, line 28:
```suggestion
	return true // TODO: Investigate moving this package to the default statefulset
```
in this context (an internal activegate package), I think using `NeedsActiveGate` is a bit weird 😅

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/kubemon.go
```diff
@@ -73,7 +72,7 @@ func (mod KubernetesMonitoringModifier) getInitContainers() []corev1.Container {
 			Args:            []string{"-c", k8scrt2jksPath},
 			VolumeMounts:    volumeMounts,
 			Resources:       mod.capability.Properties().Resources,
-			SecurityContext: GetSecurityContext(readOnlyRootFs),
+			SecurityContext: GetSecurityContext(true),
```
0sewa0, line 75: do we ever use `GetSecurityContext(false)` ?
if not, we don't need to pass in anything

gkrenn: Yes we do, will leave it as is

PR #2508, opened 2023-12-19
Avoid duplication of `--set-host-id-source` OneAgent argument

StefanHauth:
## Description

If the OneAgent argument `--set-host-id-source` is provided in the DynaKube, it causes a duplication for classic fullstack and host monitoring modes. It's then unclear which argument counts. 

This PR fixes the handling of this argument, so that it can be overridden in the DynaKube while still maintaining the applied default values otherwise.

## How can this be tested?

* deploy an ClassicFull stack Dynakube with and without `--set-host-id-source=fqdn` in it
* check the OneAgent pod specs, they should contain it only once (with value `fqdn` resp. `auto` as default)

## Checklist


pkg/controllers/dynakube/oneagent/daemonset/arguments.go
```diff
@@ -57,3 +64,7 @@ func (dsInfo *builderInfo) appendProxyArg(argMap *prioritymap.Map) {
 func (dsInfo *builderInfo) hasProxy() bool {
 	return dsInfo.dynakube != nil && dsInfo.dynakube.NeedsOneAgentProxy()
 }
+
+func appendHostIdSource(argMap *prioritymap.Map, hostIdSource string) {
```
andriisoldatenko, line 68: I don't see any benefits of using function for 1 line to replace one line .

andriisoldatenko:
practically LGTM.

aorcholski (changes requested):

pkg/controllers/dynakube/oneagent/daemonset/arguments_test.go
```diff
@@ -58,6 +58,29 @@ func TestArguments(t *testing.T) {
 		assert.NotEmpty(t, podSpecs.Containers)
 		assert.Contains(t, podSpecs.Containers[0].Args, testValue)
 	})
+	t.Run("classic fullstack", func(t *testing.T) {
```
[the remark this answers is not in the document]
StefanHauth: fixed in [37f5af7](https://github.com/Dynatrace/dynatrace-operator/pull/2508/commits/37f5af762bc2c0bdef231ee7781663376055dbed)

pkg/controllers/dynakube/oneagent/daemonset/arguments.go
StefanHauth: fixed in [37f5af7](https://github.com/Dynatrace/dynatrace-operator/pull/2508/commits/37f5af762bc2c0bdef231ee7781663376055dbed)

andriisoldatenko (approved):

PR #2509, opened 2023-12-19
Pass ActiveGate host as noProxy for CodeModules

andriisoldatenko:
## Description

Pass ActiveGate host as noProxy for CodeModules

## How can this be tested?

1.  Deploy dynakube with AG and with proxy
2. Deploy sample-app
3. make sure `ruxitagentproc.conf` has `noProxy` with value like: `xxx-activegate.dynatrace`.
4. make sure oneAget log has proper records:
```
root@php-glibc-all-inject-848c9c9f99-7n7l5:/var/www/html# cat /opt/dynatrace/oneagent-paas/log/php/ruxitagent_PHP-FPM_php-glibc-all-inject-__1.0.log
...
2024-01-09 10:04:09.121721 UTC [1/00000001] info    [native] Proxy ....................... http://squid.proxy.svc.cluster.local:3128/
2024-01-09 10:04:09.121729 UTC [1/00000001] info    [native] No-Proxy Hosts .............. xxx-activegate.dynatrace
...
2024-01-09 10:04:09.220309 UTC [10/0000000a] info    [comm  ] Connected to [https://xxxx-activegate.dynatrace:443/communication](https://xxx-activegate.dynatrace/communication)
```

## Checklist


pkg/controllers/csi/provisioner/controller.go
```diff
@@ -225,6 +226,11 @@ func (provisioner *OneAgentProvisioner) updateAgentInstallation(ctx context.Cont
 			return nil, false, err
 		}
 		latestProcessModuleConfig.AddProxy(proxy)
+
+		if dk.NeedsActiveGate() {
+			multiCap := capability.NewMultiCapability(dk)
+			latestProcessModuleConfig.AddNoProxy(capability.BuildDNSEntryPoint(dk.Name, dk.Namespace, multiCap))
+		}
```
0sewa0, line 233: so far it looks good, but what about the `applicationMonitoring without CSI-Driver` scenario?
do we support that?
if so then you have to do some magic from the init-container side

andriisoldatenko: do you mean when we have ActiveGate and without CSI ?

0sewa0: no, ActiveGate + applicationMonitoring without CSIDriver (so only injection, where each pod downloads the agent on its own)

the problem there:
- the init-container does not directly know about the Dynakube
- you could try [creating the NoProxy thing directly in the ini-secret](https://github.com/Dynatrace/dynatrace-operator/blob/010623f6bdc6601856fffa71b44944a7dd2ee925/pkg/injection/namespace/initgeneration/initgeneration.go#L164-L164)

andriisoldatenko: fixed in 9985b747

pkg/controllers/dynakube/activegate/capability/capability.go
```diff
@@ -206,3 +207,34 @@ func BuildProxySecretName(dynakubeName string) string {
...
+}
+
+// BuildDNSEntryPoint for give capability
+func BuildDNSEntryPoint(dynakubeName, dynakubeNamespace string, capability Capability) string {
+	if capability.ShortName() == consts.MultiActiveGateName && strings.Contains(capability.ArgName(), dynatracev1beta1.RoutingCapability.ArgumentName) ||
+		capability.ShortName() == dynatracev1beta1.RoutingCapability.ShortName {
+		return fmt.Sprintf("https://%s/communication,https://%s/communication", BuildServiceHostName(dynakubeName, capability.ShortName()), BuildServiceDomainName(dynakubeName, dynakubeNamespace, capability.ShortName()))
+	}
+	return fmt.Sprintf("https://%s/communication", BuildServiceHostName(dynakubeName, capability.ShortName()))
+}
```
0sewa0, line 240: Using this logic for the codemodules is a major flaw, and it will not work

So the PR works as far as it updates the ruxitagentproc conf and all:
```
bash-5.1# cat data/zib50933/config/agent/conf/ruxitagentproc.conf  | grep noProxy
noProxy https://$(ZIB50933_ACTIVEGATE_SERVICE_HOST):$(ZIB50933_ACTIVEGATE_SERVICE_PORT)/communication,https://zib50933-activegate.dynatrace:$(ZIB50933_ACTIVEGATE_SERVICE_PORT)/communication
```

however these environment variables:
- ZIB50933_ACTIVEGATE_SERVICE_HOST 
- ZIB50933_ACTIVEGATE_SERVICE_PORT
- ...
will NOT be present in the user's namespace, where the codemodules will run as the service is not in there namespace so kubernetes will not add these envvars automatically.
What is even worse, is that we couldn't even just evaluate these envvars in the CSI-driver/Operator as the Service is created dynamically, and these are envvars are only added to the container if the Service already existed when the pod was scheduled.
[official docs for service envvars](https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables)

0sewa0 (changes requested):

0sewa0 (approved):
maybe the `How to test` section in the description could use some love, for future reference
also the title of the PR is a bit off 😅 (Pass ActiveGate host as noProxy for CodeModules)

PR #2534, opened 2024-01-03
Research OpenFeature API for OneAgent multi version support (non-working PoC)

wepudt:
## Description

a possible solution on how to use K8s ConfigMaps to model differences in OneAgent multi version support.
One possibility would be to mount one property file for each supported version and determine supported versions by listing all mounted files.

Solution includes a quick mockup of a ConfigMap, some addition to the operator pod to mount the CM. A very drafty FeatureProvider implementation showcases how flags are made available from the CM via the OpenFeature API.

## How can this be tested?

not intended to be run- just for discussion.

## Checklist

### DIFF
--- a/pkg/util/feature/OneAgnetConfigMapSample.yaml
+++ b/pkg/util/feature/OneAgnetConfigMapSample.yaml
@@ -1,0 +1,35 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: operator-oneagent-features
+data:
+
+#EvaluationContext: per OneAgent version a file mount?
+#     |
+#<----+----->
+  oneagent-v1.properties: |
+    param1=1
+    param2=sampletext
+  oneagent-v2.properties: |
+    param1=2
+    param2=text
+
+
+---
+#perator pod snippet to mount ConfigMap
+volumeMounts:
+  - name: oneagent-features
+    mountPath: "/supported-oneagent-versions"
+    readOnly: true
+volumes:
+  # You set volumes at the Pod level, then mount them into containers inside that Pod
+  - name: config
+    configMap:
+      # Provide the name of the ConfigMap you want to mount.
+      name: oneagent-features
+      # An array of keys from the ConfigMap to create as files
+      items:
+        - key: "oneagent-v1.properties"
+          path: "oneagent-v1.properties"
+        - key: "oneagent-v2.properties"
+          path: "oneagent-v2.properties"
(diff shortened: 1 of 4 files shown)


pkg/util/feature/OneAgnetConfigMapSample.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: operator-oneagent-features
+data:
+
+#EvaluationContext: per OneAgent version a file mount?
```
andriisoldatenko, line 7: The original requirement is to have one configMap for everything. Also we don't want to update configMap each time when oneAgent team will release. So better to have something similar we have inside node_controller and `CacheEntry`. To review it you can describe `dynatrace-node-cache`.

andriisoldatenko: - Also, it would be nice to have an example of integrating it with the current code base.?
- It would be nice to understand how we want to reconcile changes. Mount file will work, and k8s will update automatically. But how will the operator react to changes? It's unclear.

wepudt: to the first requirement: it might be simpler to separate different things and have an own entity for OneAgent configurations. So changes from one do not interfere with the other.

Will look into node_controller `CacheEntry`

currently I would try to create an object upon reconciliation with a lazy reading of available versions (via file listing) and reading a config if accessed.

I am uncertain what the downsides of this approach are (performance, some K8s behaviour I am ignorant of) but it is the simplest approach that came to my mind...

pkg/util/feature/config_map_feature_provider.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/open-feature/go-sdk/openfeature"
+	"google.golang.org/appengine/log"
+	"strconv"
+)
+
+// ConfigMapFeatureProvider implements the OpenFeatureProvider interface
+type ConfigMapFeatureProvider struct {
```
andriisoldatenko, line 16: 1. NewConfigMapFeatureProvider(kubeAPi Reader)
2. basic evalCOntext, flag name -> you can show
3. Fresh version feature flag value.
3b. configmap we will add help release.

wepudt:
as requested https://github.com/Dynatrace/dynatrace-operator/pull/2574 for discussion if required

wepudt:
repleaced by followup

PR #2535, opened 2024-01-04
Remove check before pull-secret generation

0sewa0:
## Description
Remove a check that is meant to save a bit on requests by only creating the pull-secret when necessary, because in case of applicationMonitoring we don't really need it (unless codeModulesImage is used).

However other parts of the code requires this pull secret to be there, even its not used.
Fixing that would be take longer and should be done with lots of care.

So I remove this very minor "improvement" so that we don't break things

## How can this be tested?

Deploy `applicationMonitoring` dynakube

## Checklist

### DIFF
--- a/pkg/controllers/dynakube/dtpullsecret/reconciler.go
+++ b/pkg/controllers/dynakube/dtpullsecret/reconciler.go
@@ -37,9 +37,9 @@ func NewReconciler(clt client.Client, apiReader client.Reader, scheme *runtime.S
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context) error {
-	/* TODO: Introduce cleanup, and also this is would not be needed every case,
-	if we don't want to do anything with images, then its not necessary,
-	but other parts of the code would also need to be updated to handle this scenario.
+	/* TODO: Introduce cleanup, and also, this would not be needed in every case,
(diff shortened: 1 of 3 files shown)


pkg/controllers/dynakube/dtpullsecret/reconciler.go
```diff
@@ -37,10 +37,10 @@ func NewReconciler(clt client.Client, apiReader client.Reader, scheme *runtime.S
...
 
 func (r *Reconciler) Reconcile(ctx context.Context) error {
-	if !r.dynakube.NeedsOneAgent() && !r.dynakube.NeedsActiveGate() {
-		return nil // TODO: Introduce cleanup
-	}
-
+	/* TODO: Introduce cleanup, and also this is would not be needed every case,
+	if we don't want to do anything with images, then its not necessary,
+	but other parts of the code would also need to be updated to handle this scenario.
+	*/
```
andriisoldatenko, line 43:
```suggestion
	/* TODO: Introduce cleanup, and also, this would not be needed in every case,
	if we don't want to do anything with images, then it's not necessary,
	but other parts of the code must also be updated to handle this scenario.
	*/
```

andriisoldatenko (approved):

PR #2545, opened 2024-01-09
Streamline init container config

luhi-DT:
## Description

We should only have 1 way to configure each thing, so the same thing can't be in env and the secret.
The general rule is that the env should be used for things that are only available at the time of injection.

What was moved?
- InstallerMode
- CSIMode
- ReadOnlyCSI

## How can this be tested?

Everything should work as before, deploy a Dynakube with applicationMonitoring and test -> readOnly CSI, set a fixed version and a dynakube without readOnlyCSI but CSI still enabled

## Checklist

### DIFF
--- a/pkg/injection/startup/secret.go
+++ b/pkg/injection/startup/secret.go
@@ -29,6 +29,9 @@ type SecretConfig struct {
 	HostGroup           string            `json:"hostGroup"`
 	InitialConnectRetry int               `json:"initialConnectRetry"`
 	EnforcementMode     bool              `json:"enforcementMode"`
+	OneAgentVersion     string            `json:"oneAgentVersion"`
+	CSIMode             bool              `json:"CSIMode"`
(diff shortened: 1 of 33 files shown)


andriisoldatenko (approved):
code LGTM.

pkg/injection/startup/secret.go
```diff
@@ -29,6 +29,8 @@ type SecretConfig struct {
 	HostGroup           string            `json:"hostGroup"`
 	InitialConnectRetry int               `json:"initialConnectRetry"`
 	EnforcementMode     bool              `json:"enforcementMode"`
+	CSIMode             bool              `json:"CSIMode"`
```
waodim, line 32:
```suggestion
	CSIMode             bool              `json:"csiMode"`
```

waodim: apart from this suggestion it also LGTM.

luhi-DT: good catch

waodim (approved):
LGTM

PR #2566, opened 2024-01-11
Merge csi driver manifest to general manifest in release workflow

aorcholski:
## Description

Release kubenetes/openshift manifest contains also csi driver manifest as in case of pre-release. Needed to enable csi driver by default.

## How can this be tested?

I use forked project. See an [example](https://github.com/abcxyzdef123654/dynatrace-operator/releases/tag/v3.2.1).

## Checklist


andriisoldatenko (approved):

albertogdd (approved):

albertogdd:
> will this not just remove the csi manifests? because for example the [kubernetes.yaml in release 0.15.0](https://github.com/Dynatrace/dynatrace-operator/releases/download/v0.15.0/kubernetes.yaml) does not have csi stuff 😅

If you check in the pre-releases, we already have one single manifest including the CSI stuff.
With this PR, the logic for pre-releases is also used for normal releases (check that the `if` clause is removed).

PR #2574, opened 2024-01-12
Research OpenFeature API for OneAgent multi version support updated

wepudt:
## Description

Please include the following:

a possible solution on how to use K8s ConfigMaps to model differences in OneAgent multi version support.

a data model for mapping OneAgent versions to flag values was out-of-scope for this research.

Solution shows how the OneAgent Reconciler creates a component doing the version magic (VersionManager for lack of better name) implementing a basic OpenFeature FeatureProvider that uses the K89s API to access a ConfigMap. 
The VersionManager is then passed through the DaemonSet to the argument component to illustrate the use of the mechanism with the proxy parameter.

## How can this be tested?

not intended to run.

## Checklist


albertogdd:
I'd say we can close this PR; what do you think @wepudt ?

PR #2579, opened 2024-01-15
Reintroduce clusterid as env var instead of secret value

luhi-DT:
## Description

As we would have to restructure the whole process of generating and creating the injection secret in every case of injection we should reintroduce the cluster id as an env var instead of reading it from the secret, as it is needed in both cases (oneagent and datingest). This makes the process way easier instead of always creating the initSecret.

## How can this be tested?

Create a Dynakube with AppOnly -> Check all cases (oneagent+dataingest, oneagent only, dataingest only)

## Checklist


waodim (approved):
lgtm

PR #2583, opened 2024-01-16
Store ruxitagentproc file in node fs by dynakube name

albertogdd:
## Description

Change the file path on which the ruxitagentproc.conf file is stored in the node fs. Adding dynakube name to allow having multiple dynakubes per tenant.

Old:
`rootDir/tenantUUID/config/ruxitagentproc.conf`
New:
`rootDir/tenantUUID/dynakubeName/config/ruxitagentproc.conf`

'Old' files are left behind in the old file path and will be cleaned up when old pods are terminated.

## How can this be tested?

Unit tests + run operator locally with no errors shown.

## Checklist

### DIFF
--- a/pkg/controllers/csi/driver/volumes/app/publisher.go
+++ b/pkg/controllers/csi/driver/volumes/app/publisher.go
@@ -147,7 +147,7 @@ func (publisher *AppVolumePublisher) buildLowerDir(bindCfg *csivolumes.BindConfi
 		binFolderName = bindCfg.ImageDigest
 	}
 	directories := []string{
-		publisher.path.AgentConfigDir(bindCfg.TenantUUID),
+		publisher.path.AgentConfigDir(bindCfg.TenantUUID, bindCfg.DynakubeName),
 		publisher.path.AgentSharedBinaryDirForAgent(binFolderName),
 	}
 	return strings.Join(directories, ":")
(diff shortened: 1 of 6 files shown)


pkg/controllers/csi/driver/volumes/app/publisher.go
```diff
@@ -139,15 +139,15 @@ func (publisher *AppVolumePublisher) fireVolumeUnpublishedMetric(volume metadata
 	}
 }
 
-func (publisher *AppVolumePublisher) buildLowerDir(bindCfg *csivolumes.BindConfig) string {
+func (publisher *AppVolumePublisher) buildLowerDir(bindCfg *csivolumes.BindConfig, dynakubeName string) string {
```
luhi-DT, line 142: I would rather include the dynakube name into the bindCfg instead, as we are in control of that struct and it already has a lot of dynakube specific attributes :)

luhi-DT (changes requested):

pkg/controllers/csi/driver/volumes/app/publisher.go
albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/2583/commits/2433ba5e9238f667b34fd699125d0bc4fc313be3)

PR #2587, opened 2024-01-17
Add annotation to exclude containers from injection.

StefanHauth:
## Description

Allow to exclude containers from injection based on their names with annotations on Dynakube or pod level.

## How can this be tested?

Deploy a Dynakube and an application.
Add annotation `container.inject.dynatrace.com/<containername> = false` to exclude app containers from injection.
Check that code modules are not injected into containers.

## Checklist


pkg/webhook/mutation/pod_mutator/oneagent_mutation/containers.go
```diff
@@ -20,12 +21,19 @@ func (mutator *OneAgentPodMutator) setContainerCount(initContainer *corev1.Conta
...
-		mutator.addOneAgentToContainer(request.ToReinvocationRequest(), container)
+
+		if !ContainerIsExcluded(request.BaseRequest, container.Name) {
+			addContainerInfoInitEnv(request.InstallContainer, i+1, container.Name, container.Image)
+			mutator.addOneAgentToContainer(request.ToReinvocationRequest(), container)
+			injecteContainers++
+		}
 	}
+
+	return injecteContainers
```
andriisoldatenko, line 36:
```suggestion
	return injectedContainers
```

pkg/webhook/mutation/pod_mutator/pod_mutator.go
```diff
@@ -150,8 +151,19 @@ func (webhook *podMutatorWebhook) handlePodMutation(ctx context.Context, mutatio
 	ctx, span := dtotel.StartSpan(ctx, webhookotel.Tracer(), spanOptions()...)
 	defer span.End()
 
+	needsInjection := false
```
luhi-DT, line 154: I would extract this logic and use a new function instead of making handePodMutation even bigger

luhi-DT (changes requested):

pkg/webhook/config.go
```diff
@@ -43,6 +43,8 @@ const (
 	// "fail", the init container will exit with error code 1. Defaults to "silent".
```
waodim, line 43: Because I am currently testing it: What do you think about placing some logs when a container is excluded? for example: a container spins up which should be excluded and we log something along the line: "application container <name>  is being excluded"

waodim (commented):
See comment below

luhi-DT (changes requested):

luhi-DT (approved):

PR #2589, opened 2024-01-17
Enable missing linters

andriisoldatenko:
## Description

This PR adds missing golang linters, which doesn't require to
fixing any code.

Also as TODO: I left list of linters we need to enable seperatly, 
ideally one by one, so 1 linter 1 PR whoch must includes all nessesary changes
to pass linting checks.

## How can this be tested?

make go/lint

## Checklist


.golangci.yml
```diff
@@ -159,30 +159,111 @@ linters-settings:
 linters:
   disable-all: true
   enable:
-  - gci
```
andriisoldatenko, line 162: I've sorted linters so it's easy to catch duplicates and do some analysis.

.golangci.yml
```diff
@@ -159,30 +159,110 @@ linters-settings:
...
+  - unconvert                 # Remove unnecessary type conversions
+  - unparam                   # Reports unused function parameters
+  - unused                    # Checks Go code for unused constants, variables, functions and types
+  - whitespace                # Tool for detection of leading and trailing whitespace.
+  - zerologlint               # Detects the wrong usage of zerolog that a user forgets to dispatch with Send or Msg.
+
+    # TODO: Requires some code changes before enabling:
+
+# - bodyclose         # Checks whether HTTP response body is closed successfully.
+# - contextcheck      # Check whether the function uses a non-inherited context.
```
wepudt, line 212: would that help us with proper context handling, or would the changes be too numerous?

.golangci.yml
```diff
@@ -159,30 +159,110 @@ linters-settings:
...
+# - cyclop            # Checks function and package cyclomatic complexity.
+# - dupl              # Tool for code clone detection.
```
wepudt, line 214: problem I have with code duplication: not all duplication is an issue. Only if the same concept is duplicated but not simply duplication of boilerplate code...

.golangci.yml
```diff
@@ -159,30 +159,110 @@ linters-settings:
...
+# - usestdlibvars     # A linter that detect the possibility to use variables/constants from the Go standard library.
+# - wastedassign      # wastedassign finds wasted assignment statements.
+# - wrapcheck         # Checks that errors returned from external packages are wrapped.
+# - wsl               # Add or remove empty lines.
+# TODO: Need to discuss/decide/remove
+# - exhaustive     # Check exhaustiveness of enum switch statements.
+# - godox          # Tool for detection of FIXME, TODO and other comment keywords.
+# - inamedparam    # reports interfaces with unnamed method parameters.
+# - interfacebloat # A linter that checks the number of methods inside an interface.
+# - lll            # Reports long lines.
```
wepudt, line 261: I am a bit torn here: on one hand I do not like those long lines- and we have them, but on the other hand having to read very long functions also does not help. I especially detest that with long signature lines as in golang you have to scroll to the end to see the return type. But putting each param on a new line also needs getting used to as it is not in line with golang formatter 😞

0sewa0 (approved):

.golangci.yml
andriisoldatenko: When I ran it, it was a bit useless, needs more investigation.

.golangci.yml
andriisoldatenko: Correct, this one only complains about our tests, and it;s okey to copy paste some stuff during testing phase, so I think we need may be configure it to ignore tests and enable.

.golangci.yml
andriisoldatenko: long lines are always signal that you are doing something wrong. Always remember of https://en.wikipedia.org/wiki/The_Magical_Number_Seven,_Plus_or_Minus_Two

PR #2614, opened 2024-01-25
Add integration test to check build variables

waodim:
## Description

With this a new integration (no e2e) test is added to check if the build variables (commitHash, version) are set correctly, as we had issues in the past with the corresponding log line.

## How can this be tested?

run the test but do not forget the necessary flags.
example: `go test -ldflags="-X 'github.com/Dynatrace/dynatrace-operator/pkg/version.Commit=$(git rev-parse HEAD)' -X 'github.com/Dynatrace/dynatrace-operator/pkg/version.Version=$(git branch)'" **./cmd/integration/*`

## Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,10 +61,11 @@ jobs:
         id: depdownload
         run: |
           hack/build/ci/install-cgo-dependencies.sh
-      - name: Run Unit tests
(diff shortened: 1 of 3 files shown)


andriisoldatenko:
@waodim I think what's missing here, is to tweaking CI scripts.

andriisoldatenko (approved):

Issue #2623, opened 2024-01-25
Configure Port via Environment Variable WEBHOOK_PORT


ChuckQuinnIV:
It would be nice to be able to configure the webhook to listen on a different port than `8443` via an environment variable called `WEBHOOK_PORT`.

Webhooks in some EKS cluster configurations using overlays need to use `hostNetwork`, where ports between different applications can conflict.

Happy to create this pull request if this is something the project would want to support!

PR #2624, opened 2024-01-25
Add WEBHOOK_PORT environment variable

ChuckQuinnIV:
## Description

Allows users to configure port webhook listens on via environment variable WEBHOOK_PORT. Default of 8443 is kept if environment variable is unset or invalid.

Resolves https://github.com/Dynatrace/dynatrace-operator/issues/2623

This change will allow users in EKS clusters using overlays to configure the port the webhook listens on to avoid collisions with other webhooks.

## How can this be tested?

Running the application with `WEBHOOK_PORT` set to a valid integer should cause the webhook to listen on the provided port.

Running the application without `WEBHOOK_PORT` set, or set to an invalid value, should cause the webhook to continue listening on `8443`.

## Checklist


chrismuellner:
Thanks for your contribution! This will allow users to override the webhook port, assuming they also update the corresponding webhook resources via another tool. We'll look into simplifying that in the future, but for now, it will give some more flexibility, as outlined in your description.

luhi-DT:
@ChuckQuinnIV please checkout our CI and check the failing linting step

luhi-DT (approved):

PR #2654, opened 2024-02-01
Remove registryClient from version reconciler

0sewa0:
## Description

We no longer use the registry client for getting the version info for any of the scenarios, we just use endpoints, or just not care.

This PR handles ~2~ 3 things, see commits: (it was faster to do it in one, as one of them was forgotten 😅 )
1. Make the OneAgent HealthConfig be dependent on a version threshold instead of the image manifests
   - We do this so we don't have to ping the registry, which might be problematic for users that need a custom-pullsecret
   - I avoided using our `semantic.go` implementation: https://github.com/Dynatrace/dynatrace-operator/blob/40fd5e3795a4e5f02cfcb8e5c0a72f12226f0227/pkg/version/semantic.go#L30-L30
     - I did this so I introduce the least amount of breaking change, as that code is a bit old and strict on what it accepts.
     - (+) I think that code should be removed and we should use the official `"golang.org/x/mod/semver"` package, like in my change. (unfortunately the version format of the OneAgent is not exactly semver, but can be "converted" into)
2. Removes the registryClient from the public registry scenario, this just simplifies the whole logic a bunch, as we will relay on what the DynatraceAPI gives us. There the expected behaviour is that we receive the version as the `Tag` from the API.
3. Fixes the CodeModulesImage scenario:
    - The CodeModulesImage (both custom and public-registry)NEEDS and EXPECTS the digest to be in the status for it to work, (or to not to introduce major changes in behaviour), so I moved the logic for getting the digest in the provisioner
         - Downside: we ping the registry on each node, instead of only in the operator
         - Happy thought: It looks like, that we don't need the digest here since the change to the "new" image library, so we could remove it from here as well, but that would affect folder names and such which could cause problems.  

## How can this be tested?

1. Test that the readiness-probe on the OneAgent DaemonSet correctly changes if the version is low/high enough
   - The old `bin/sh ...` readiness-probe should be used for versions older then 1.276
   - For 1.276 and newer version the binary based probe should be used

2. In case when the public-registry feature-flag is used, we should only relay on the information we get from the API.
   - We no longer have the digest in the `ImageID` of the Dynakube status, and therefore in any of our components
      - Don't forget to check the CodeModulesImage scenario!!

## Checklist


andriisoldatenko:
@0sewa0 please resolve conflicts.

pkg/controllers/dynakube/version/healthconfig_test.go
```diff
@@ -1,154 +1,100 @@
...
-	dynakube := &dynatracev1beta1.DynaKube{}
-	apiReader := fakek8s.NewClientBuilder().Build()
-	pullSecret := &corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: dynakube.PullSecretName(),
+	type test struct {
+		title           string
+		inputVersion    string
+		expectedCommand []string
+		errors          bool
```
chrismuellner, line 15:
```suggestion
		expectError          bool
```

luhi-DT (approved):

pkg/controllers/dynakube/version/healthconfig_test.go
0sewa0: done 😉

luhi-DT (approved):

PR #2668, opened 2024-02-05
Copy ruxitagentproc.conf config into upper-dir

0sewa0:
## Description

Instead of having multiple lower-dirs that will be merged, the config is just copied to the upper-dir.

We do this to avoid undefined behaviour incase we change the config overtime.
- We regularly update the config, which is problematic if that is a lower dir, because:
   - "Changes to the underlying filesystems while part of a mounted overlay filesystem are not allowed." (Ref https://docs.kernel.org/filesystems/overlayfs.html#changes-to-underlying-filesystems)
   - It "works", but the behaviour is undefined and can actually cause problems.

## How can this be tested?

If you shell into the CSI driver's pod, and check the upper-dir `ruxitagentproc.conf` it should be the correct config:
Example for upper-dir: 
`data/zib50933/run/csi-82962f01ee59384a8105bf60be0fab48aa030f6c9d5eeb604e3ab88e5ff8daa6/var/agent/conf/ruxitagentproc.conf`

to check that there is no more config lower-dir look at `/proc/mounts`:
```shell
> tail -n 1 /proc/mounts
overlay /data/zib50933/run/csi-82962f01ee59384a8105bf60be0fab48aa030f6c9d5eeb604e3ab88e5ff8daa6/mapped overlay rw,relatime,lowerdir=/data/codemodules/1.286.0.20240131-150854,upperdir=/data/zib50933/run/csi-82962f01ee59384a8105bf60be0fab48aa030f6c9d5eeb604e3ab88e5ff8daa6/var,workdir=/data/zib50933/run/csi-82962f01ee59384a8105bf60be0fab48aa030f6c9d5eeb604e3ab88e5ff8daa6/work 0 0
```

For a more real life scenario, the following **should no longer happen**:
Step 1: Deploy DynaKube, builtin:networkzones is disabled and .spec.networkZone is unset
Step 2: Deploy app with agent injection
Step 3: Update DynaKube, set .spec.networkZone
- The content of the ruxitagentproc.conf changes in the container, because the content of the ruxitagentproc.conf  in the lower-dir gets updated (THIS IS WHAT DOES NOT HAPPEN ANYMORE)
Step 4: Let app container with agent crash and restart
- Agent reads updated but invalid ruxitagentproc.conf and fails to start

## Checklist

Instead of having multiple lower-dirs that will be merged, the config is just copied to the upper-dir

### DIFF
--- a/pkg/controllers/csi/driver/volumes/app/publisher.go
+++ b/pkg/controllers/csi/driver/volumes/app/publisher.go
@@ -19,6 +19,7 @@ package appvolumes
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -27,6 +28,7 @@ import (
 	csivolumes "github.com/Dynatrace/dynatrace-operator/pkg/controllers/csi/driver/volumes"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/csi/metadata"
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/pkg/errors"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/spf13/afero"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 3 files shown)


pkg/controllers/csi/driver/volumes/app/publisher.go
```diff
@@ -147,18 +149,57 @@ func (publisher *AppVolumePublisher) buildLowerDir(bindCfg *csivolumes.BindConfi
...
+	upperDir := publisher.path.OverlayVarDir(bindCfg.TenantUUID, volumeCfg.VolumeID)
+	err := publisher.fs.MkdirAll(upperDir, os.ModePerm)
+	if err != nil {
+		return "", errors.WithMessagef(err, "failed create overlay upper directory structure, path: %s", upperDir)
+	}
+
+	destAgentConfPath := publisher.path.OverlayVarRuxitAgentProcConf(bindCfg.TenantUUID, volumeCfg.VolumeID)
+	err = publisher.fs.MkdirAll(filepath.Dir(destAgentConfPath), os.ModePerm)
+	if err != nil {
+		return "", errors.WithMessagef(err, "failed create overlay upper directory agen config directory structure, path: %s", upperDir)
```
andriisoldatenko, line 167: what is `agen`?

pkg/controllers/csi/driver/volumes/app/publisher.go
```diff
@@ -147,18 +149,57 @@ func (publisher *AppVolumePublisher) buildLowerDir(bindCfg *csivolumes.BindConfi
...
+	}
+
+	srcAgentConfPath := publisher.path.AgentSharedRuxitAgentProcConf(bindCfg.TenantUUID, volumeCfg.DynakubeName)
+	srcFile, err := publisher.fs.Open(srcAgentConfPath)
+	if err != nil {
+		return "", errors.WithMessagef(err, "failed to open source ruxitagentproc.conf file, path: %s", srcAgentConfPath)
```
andriisoldatenko, line 173:
```suggestion
		return "", errors.WithMessagef(err, "failed to open ruxitagentproc.conf file, path: %s", srcAgentConfPath)
```

andriisoldatenko (approved):
LGTM, except few typos

pkg/controllers/csi/driver/volumes/app/publisher.go
0sewa0: done

pkg/controllers/csi/driver/volumes/app/publisher.go
0sewa0: fixed 😅

waodim (approved):
tested it as far as i could and code looks good to me :)

PR #2676, opened 2024-02-06
Handle custom-image in case of version checks

0sewa0:
## Description

Incase a custom image is used for the OneAgent DaemonSet, then we set the `Version` and `Source` in the Dynakube's `Status` to the `custom-image` constant. 

Our version checks didn't handle this, now they do.

Also flipped the proxy env version check, so if we **don't** know the version then we **don't** use the deprecated proxy envvar.
- This was decided to be consistent with the HealthConfig logic.
   - Furthermore the HealthConfig version check is checking for a newer version.
- If we don't want to do this, then IMO we need to make the HealthConfig logic consistent with it as well. 

## How can this be tested?

In case the `oneAgent.image` field is set to something, everything works

## Checklist


waodim (approved):
tested it - works and code LGTM

PR #2705, opened 2024-02-08
Add context to dynatrace client

0sewa0:
## Description

The dynatrace client was written a while ago and back then (for some reason) the `context.Context` was not used and passed properly.

Now we do it for every call where the dynatrace-client can use it. (Better late than never 😅 )

Its a rather "big" PR, but most of the changes are trivial, there are just many of them. 😬 
I tried to split it into several logical commits (retrospectively 😅) to maybe help with the reviews.

## How can this be tested?

No change in behaviour. (nothing should break)
- In some cases the lack of context usage caused the client to not timeout, like downloading an agent zip, but reproducing this is not really easy. 😅 

## Checklist


waodim (dismissed):
code LGTM and tests run through

andriisoldatenko (changes requested):
please upgrade mockery locally, and re-run `make go/gen_mocks`

andriisoldatenko (approved):

PR #2710, opened 2024-02-08
Add additional linters: nolintlint, predeclared, promlinter, protogetter

waodim:
## Description

With this new linters a added:
- nolintlint: ca52f9df63ccdba8cbbafbbb821ae9a85cefc628 I had to remove several nolint statements because their directives where unused, also removed whitespace between // and nolint
- predeclared: d287e6bf96d382b9dd38edd5d380619bb37f3cb3 I had to change names when variables where named after internal predeclared identifiers
~- promlinter: 1debc524cc526bd49ff7bfd7f071d0119db8e8aa I had to change the parameters create a dedicated variable with a suffix "_total", I decided to not do that. Let me know if you do not agree and want me to do it.~

## How can this be tested?

No changes in functionality

## Checklist


pkg/controllers/csi/provisioner/controller_test.go
```diff
@@ -42,7 +42,7 @@ func (fs *mkDirAllErrorFs) MkdirAll(_ string, _ os.FileMode) error {
 	return fmt.Errorf(errorMsg)
 }
 
-func TestOneAgentProvisioner_Reconcile(t *testing.T) { //nolint:revive
+func TestOneAgentProvisioner_Reconcile(t *testing.T) {
 	ctx := context.TODO()
```
wepudt, line 46: think in unit-tests context.Background() is used

wepudt (changes requested):
when the linter is satisfied I am ok with merging :)

pkg/controllers/csi/provisioner/controller_test.go
waodim: Fixed in https://github.com/Dynatrace/dynatrace-operator/pull/2710/commits/560727953ee8b7e6a577db07967421ac7ee56b5f

pkg/controllers/csi/gc/config.go
```diff
@@ -9,21 +9,21 @@ import (
 var (
 	log = logger.Get().WithName("csi-gc")
 
-	reclaimedMemoryMetric = prometheus.NewCounter(prometheus.CounterOpts{
+	reclaimedMemoryMetric = prometheus.NewCounter(prometheus.CounterOpts{ //nolint:promlinter
```
andriisoldatenko, line 12: I'm thinking of not enabling this linter at all. Because sounds like we enable it and only code we use prom related code we ignore it :)

waodim: Yes i thought so as well. We can enable it but in my opinion it is unnecessary... and it would make the code ugly IMO. what do you think?

waodim: I removed promlinter because as you already mentioned it is currently not necessary
https://github.com/Dynatrace/dynatrace-operator/pull/2710/commits/e48ccddd39950f46fa5cdcbb5a023832539d1f93

andriisoldatenko (approved):

PR #2720, opened 2024-02-09
Adapt the local linting command to new linters

0sewa0:
## Description

Several linters were introduced recently, one in particular `wsl` has its own fixer command, that we should use in the `make go/lint` command.

Also added the linting tools install/update make target as a prerequisite to the `make go/lint` command, so we don't have to think about locally updating the linters.  (it slows the command down a bit, but the golangci-lint is already dog slow, so it doesn't really matter)

## How can this be tested?

run `make go/lint`

## Checklist

### DIFF
--- a/hack/make/go.mk
+++ b/hack/make/go.mk
@@ -29,7 +29,7 @@ go/golangci:
 	golangci-lint run --build-tags "$(shell ./hack/build/create_go_build_tags.sh true)" --timeout 300s
 
 ## Runs all the linting tools
-go/lint: prerequisites/go-linting go/format go/vet go/golangci
+go/lint: prerequisites/go-linting go/format go/vet go/wsl go/golangci
 
 ## Runs all go unit tests and writes the coverprofile to coverage.txt
 go/test:
--- a/hack/make/go.mk
+++ b/hack/make/go.mk
@@ -21,12 +21,15 @@ go/format: go/fmt go/gci
 go/vet:
 	go vet -copylocks=false $(LINT_TARGET)
 
+go/wsl:
+	wsl -fix ./pkg/...
+
 ## Runs golangci-lint
 go/golangci:
 	golangci-lint run --build-tags "$(shell ./hack/build/create_go_build_tags.sh true)" --timeout 300s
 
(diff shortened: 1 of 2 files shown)


andriisoldatenko (dismissed):

hack/make/go.mk
```diff
@@ -21,12 +21,15 @@ go/format: go/fmt go/gci
...
+go/wsl:
+	wsl -fix ./pkg/...
+
 ## Runs golangci-lint
 go/golangci:
 	golangci-lint run --build-tags "$(shell ./hack/build/create_go_build_tags.sh true)" --timeout 300s
 
 ## Runs all the linting tools
-go/lint: go/format go/vet go/golangci
+go/lint: prerequisites/go-linting go/format go/vet go/golangci
```
StefanHauth, line 32: Shouldn't `go/wsl` be added here, not the tool install step?

StefanHauth (changes requested):

hack/make/go.mk
albertogdd, line 32:
```suggestion
go/lint: go/wsl go/format go/vet go/golangci
```
no bueno?

I would only add `prerequisites/` commands in the `make prerequisites` global command

0sewa0: well both actually, nice catch :D

0sewa0: > Also added the linting tools install/update make target as a prerequisite to the make go/lint command, so we don't have to think about locally updating the linters. (it slows the command down a bit, but the golangci-lint is already dog slow, so it doesn't really matter)

albertogdd (approved):

hack/make/go.mk
0sewa0: when I run `make go/lint` I want to make sure that the output shows what the CI would show.

It happened to me several times that the local linter was outdated, everything passed locally, but the CI exploded

but if for popular demand we don't want it, then I can live without it 😅

albertogdd: Yep, now I know the pain. Sounds good to me too :+1:

StefanHauth (approved):

PR #2721, opened 2024-02-09
Revert "Adapt the local linting command to new linters"

StefanHauth:
The original PR breaks `make go/lint` as `wsl -fix` doesn't respect `//nollnt` directives corretly and thus reports errors.


andriisoldatenko (approved):

StefanHauth:
PR is obsolete.

PR #2725, opened 2024-02-09
Implement suggestions made by GoLand

0sewa0:
## Description

I noticed that we are not the greatest at removing code when its no longer used. So I used GoLand to find all the things we don't use.

Then I found more things that don't really make sense to do, and removed those too.
Then I used GoLand further to clean up the warnings and suggestions where it made sense. (there were cases where it didn't, example: 1 out 3 values of an enum was not used, IMO this is fine)

Use/Read the Commits if you want explanation on why certain things happened.

## How can this be tested?

No change in behaviour.

## Checklist


andriisoldatenko (approved):
always wanted to do it! nice!

aorcholski (approved):

PR #2731, opened 2024-02-12
bug: fix sha256 prefix handling in csi driver

andriisoldatenko:
## Description

This PR fixes issue with the problematic code: https://github.com/Dynatrace/dynatrace-operator/blob/9a3bcd1bc6b3eb9f946d056c83474efb23dde74b/pkg/controllers/csi/provisioner/install.go#L71-L97

## How can this be tested?

`make test/e2e/cloudnative/codemodules`

## Checklist


pkg/controllers/csi/provisioner/install.go
```diff
@@ -93,7 +94,9 @@ func (provisioner *OneAgentProvisioner) getDigest(dynakube dynatracev1beta1.Dyna
 		return "", err
 	}
 
-	return string(imageVersion.Digest), nil
+	digest, _ := strings.CutPrefix(string(imageVersion.Digest), "sha256:")
```
waodim, line 97: Would it be good to handle the case if CutPrefix returns false? Or do you think it is not necessary?

andriisoldatenko: it;s not necessary., 
https://go.dev/play/p/71it3wq7o_W

wepudt (commented):
tests work and functionality does what is advertised.

If we would like to add more than a minimal change we could:
* separate the concerns of creating a Registry and getting the Digest in the `getDigets` function
* pass the `context.Context `from `installAgentImage`

andriisoldatenko:
> * separate the concerns of creating a Registry and getting the Digest in the `getDigets` function

I can add context, regarding splitting code, I don't see any benefits. We use this function only once.

wepudt (approved):
lgtm
ran tests, looked at unit test

PR #2736, opened 2024-02-13
Create separate injection reconciler

albertogdd:
## Description

Decouple injection from dynakube reconciler -> `dynakube/injection.go` is removed, new package `dynakube/injection` is created

Now injection has its own reconciler which can be mocked via Controller builder.

## How can this be tested?

Unit tests
`make test/e2e/activegate`
`make test/e2e/cloudnative/*`

## Checklist


pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -0,0 +1,141 @@
...
+
+type ReconcilerBuilder func(client client.Client, apiReader client.Reader, dynakube *dynatracev1beta1.DynaKube, istioReconciler istio.Reconciler, versionReconciler version.Reconciler) Reconciler
+
+func NewReconciler(client client.Client, apiReader client.Reader, dynakube *dynatracev1beta1.DynaKube, istioReconciler istio.Reconciler, versionReconciler version.Reconciler) Reconciler {
+	return &reconciler{
+		client:            client,
+		apiReader:         apiReader,
+		dynakube:          dynakube,
+		istioReconciler:   istioReconciler,
+		versionReconciler: versionReconciler,
```
0sewa0, line 37: On the one hand I would prefer that these 2 are not passed in, but created here
- so the `dyankube-controller` does not have to think about those
   -  the `istioReconciler` is a bit of an outlier, because the `dynakube-controller` will also use it, but maybe we can only reuse the `istioClient`?
   -  The current `istioReconciler` does comply with the `Reconcile` interface, because it was "quickly" converted into something that can support the reconcile of different sections, but that could possible be done better.

On the other hand, by not passing them in you would probably need to increase the params list

0sewa0 (commented):
in general looks good, left a few comments to think about
others should also give their opinions

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/version"
+	"github.com/Dynatrace/dynatrace-operator/pkg/injection/namespace/ingestendpoint"
+	"github.com/Dynatrace/dynatrace-operator/pkg/injection/namespace/initgeneration"
+	"github.com/Dynatrace/dynatrace-operator/pkg/injection/namespace/mapper"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Reconciler interface {
+	Reconcile(ctx context.Context) error
+}
```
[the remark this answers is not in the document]
wepudt: it's in the controllers package: `controllers.Reconciler`

pkg/controllers/dynakube/controller_test.go
```diff
@@ -356,13 +360,14 @@ func TestReconcileComponents(t *testing.T) {
...
 			activeGateReconcilerBuilder:     createActivegateReconcilerBuilder(mockActiveGateReconciler),
+			injectionReconcilerBuilder:      createInjectionReconcilerBuilder(mockInjectionReconciler),
 		}
 		mockedDtc := mockedclient.NewClient(t)
 		err := controller.reconcileComponents(ctx, mockedDtc, nil, dynakube)
 
 		require.Error(t, err)
 		// goerrors.Join concats errors with \n
-		assert.Len(t, strings.Split(err.Error(), "\n"), 4) // ActiveGate, OneAgentInjection, EnrichmentInjection, OneAgent
+		assert.Len(t, strings.Split(err.Error(), "\n"), 3) // ActiveGate, OneAgentInjection, EnrichmentInjection, OneAgent
```
aorcholski, line 370: could you please update the comment? This assert is confusing now

pkg/controllers/dynakube/injection/reconciler.go
albertogdd: Very good point, I adressed it in [this commit](https://github.com/Dynatrace/dynatrace-operator/pull/2736/commits/b721d3d23dd5429ee8dca1c827f673b3cef0fea0). It does increase the number of parameters (which made me even disable revive linter) but now the istio client is shared between active gate and injection, and in injection we create the reconcilers in the constructor.
Let me know your thoughts :+1:

pkg/controllers/dynakube/controller.go
```diff
@@ -375,3 +380,8 @@ func (controller *Controller) reconcileComponents(ctx context.Context, dynatrace
 
 	return goerrors.Join(componentErrors...)
 }
+
+func (controller *Controller) createDynakubeMapper(ctx context.Context, dynakube *dynatracev1beta1.DynaKube) *mapper.DynakubeMapper {
+	dkMapper := mapper.NewDynakubeMapper(ctx, controller.client, controller.apiReader, controller.operatorNamespace, dynakube)
+	return &dkMapper
+}
```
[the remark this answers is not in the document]
albertogdd: I can create a story about this. I would prefer to skip it from this one to not change more code.

pkg/controllers/dynakube/injection/reconciler.go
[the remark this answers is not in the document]
albertogdd: [Fixed here](https://github.com/Dynatrace/dynatrace-operator/pull/2736/commits/ebb2f1966ff99ea79c6424aeedb2c677b4b21fee), thanks guys.

pkg/controllers/dynakube/controller_test.go
albertogdd: You are totally right! [Fixed here](https://github.com/Dynatrace/dynatrace-operator/pull/2736/commits/2541a78f1e4b7aa0fa0625e5190ba0a5c5824aba)

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -0,0 +1,170 @@
...
+) controllers.Reconciler
+
+//nolint:revive
+func NewReconciler(
+	client client.Client,
+	dynatraceClient dynatrace.Client,
+	istioClient *istio.Client,
+	apiReader client.Reader,
+	fs afero.Afero,
+	dynakube *dynatracev1beta1.DynaKube,
```
0sewa0, line 46: small suggestion, so the order makes more sense (I like to have the `client` and `apiReader` to be passed in as a pair)
```suggestion
func NewReconciler(
	client client.Client,
	apiReader client.Reader,
	dynatraceClient dynatrace.Client,
	istioClient *istio.Client,
	fs afero.Afero,
	dynakube *dynatracev1beta1.DynaKube,
```

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -0,0 +1,170 @@
...
+		apiReader:         apiReader,
+		dynakube:          dynakube,
+		istioReconciler:   istioReconciler,
+		versionReconciler: version.NewReconciler(apiReader, dynatraceClient, fs, timeprovider.New().Freeze()),
+	}
+}
+
+func (r *reconciler) Reconcile(ctx context.Context) error {
+	if r.istioReconciler != nil {
+		err := r.istioReconciler.ReconcileAPIUrl(ctx, r.dynakube)
```
0sewa0, line 65:
```suggestion
		err := r.istioReconciler.ReconcileCodeModuleCommunicationHosts(ctx, r.dynakube)
```

you reconcile the `APIUrl` in the dynakube-controller as its a general property, and anything that will want to use the dtclient will have that as a prerequisite

pkg/controllers/dynakube/controller.go
```diff
@@ -318,11 +317,23 @@ func (controller *Controller) setupTokensAndClient(ctx context.Context, dynakube
...
+	var istioReconciler istio.Reconciler
+
+	if istioClient != nil {
+		istioReconciler = controller.istioReconcilerBuilder(istioClient)
+
+		err := istioReconciler.ReconcileAPIUrl(ctx, dynakube)
+		if err != nil {
+			return errors.WithMessage(err, "failed to reconcile istio objects for API url")
+		}
+	}
```
0sewa0, line 336: this should be done earlier

the purpose of `ReconcileAPIUrl` is that the operator can talk to the Dynatrace API even if the operator's namespace is restricted with istio.

So we have to do this BEFORE ever using the dtclient.
- and even the creation of the dtclient requires the use of the dtclient, because we check the scopes at the same time [(kinda dumb, but thats how it works 😅)](https://github.com/Dynatrace/dynatrace-operator/blob/137c5b0b89ed633b11525a537a1bea723f9dddfa/pkg/controllers/dynakube/dynatraceclient/builder.go#L98)

this is why `setupIstio` is at the very beginning of the reconcile

0sewa0 (changes requested):

pkg/controllers/dynakube/injection/reconciler.go
0sewa0: but actually you already do that at: https://github.com/Dynatrace/dynatrace-operator/pull/2736/files#diff-1f094421eabf0cb587da250be2bbd4bba051c5e50cd36d79d48358ff685171afR125

so it actually not necessary

albertogdd: :+1: I will remove it then.

I'm not 100% familiar with how it works. In my mind, as the end goal is to make the reconcilers more independent, I wanted to execute in injection reconciler the same functions that are triggered on active gate reconciler.

e.g. In case injection reconciler is executed 'alone', `func ReconcileAPIUrl` won't be executed because it is done in the dynakube controller, but I guess there is no use case where dynakube reconciler isn't executed alongside injection reconciler, so it makes sense to not repeat `func ReconcileAPIUrl`.

albertogdd: [Fixed here](https://github.com/Dynatrace/dynatrace-operator/pull/2736/commits/5b89aab62c04646dadaffe0d57f2d85009c1b74c)

pkg/controllers/dynakube/controller.go
albertogdd: [Fixed here](https://github.com/Dynatrace/dynatrace-operator/pull/2736/commits/c0e8fd6e775fbe176dd1499d60be3f5f61b33e2f)

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		log.Info("could not remove data-ingest secret")
+		return err
+	}
+	// TODO: remove initgeneration secret as well + handle errors jointly
+
+	return nil
+}
+
+func (r *reconciler) setupOneAgentInjection(ctx context.Context, dynakube *dynatracev1beta1.DynaKube, istioReconciler istio.Reconciler, versionReconciler version.Reconciler) error {
```
aorcholski, line 111: Do we need to pass `istioReconciler` and `versionReconciler` as arguments ? They are members of the `reconciler` object.

albertogdd: Thank you for noticing about this :heart: 
[Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/2736/commits/ccce366127b10ce397dda628f264b7ad34cf7aae)

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+	istioReconciler   istio.Reconciler
+	versionReconciler version.Reconciler
+}
+
+type ReconcilerBuilder func(
+	client client.Client,
+	apiReader client.Reader,
+	dynatraceClient dynatrace.Client,
+	istioClient *istio.Client,
+	fs afero.Afero,
```
aorcholski, line 34: IMHO the `fs` field is not needed in the builder

aorcholski (approved):
LGTM

0sewa0 (approved):

PR #2739, opened 2024-02-13
move oci/registry/mocks to proper location to make it in same way as other mocks.

andriisoldatenko:
## Description

This PR fixes "chicken and egg" problem. From one side we have `make go/gen_mocks` from another side we do some linting and formating. Usually we prefer to avoid lint/format mocks or any automatically generated code.

So I moved `oci/registry/mocks` to proper location to make it in same way as other mocks.

## How can this be tested?

unit-tests

## Checklist


aorcholski (approved):
LGTM I launched 
```
make go/lint
make go/gen_mocks
make go/test
```
in a loop and in a random order. No errors.

PR #2744, opened 2024-02-13
Only support autoUpdate for default registry in EdgeConnect

0sewa0:
Please include the following:
To remain consistent with the DynaKube and avoid confusion, we will not support `autoUpdate` for custom image repositories.

This is so we can avoid requiring the `customPullSecret` for custom image registries.

We will still use the registry to get the version(digest) that allows `autoUpdate` to function.
- This is because we don't have an endpoint in the Dyntrace API to check which is the latest EdgeConnect version.
- The default registry is `docker.io`, so no `customPullSecret` is needed

I left in that if only the `imageRef.Tag` is not default we still get the digest, this is mainly an implementation detail, also its security best practice to use to digest, which is a plus.

## How can this be tested?

The unit-tests are pretty straight forward.

But you can get an edgeconnect image from a non-default registry and then you should still be able to use it even if the operator it self wouldn't be able to pull. (also no digest will be in the imageID of the EdgeConnect CR Status)

## Checklist

### DIFF
--- a/pkg/controllers/edgeconnect/version/updater.go
+++ b/pkg/controllers/edgeconnect/version/updater.go
@@ -2,7 +2,6 @@ package version
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/status"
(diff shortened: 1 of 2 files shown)


aorcholski (approved):
I used `docker.io` as custom registry
'''
  imageRef:                                                                                                                                                                                                          
    repository: docker.io/dynatrace/edgeconnect                                                                                                                                                                      
    tag: latest 
'''    
LGTM

PR #2747, opened 2024-02-14
Push Helm Chart to ECR on Release

wepudt:
## Description

Reworks how we do helm pushes to include Amazon ECR. Moves login logic from action to workflow as it is only called there, adds login to ECR steps to workflow and reuses helm upload action for ECR.

## How can this be tested?

Not as convenient as I would like to: when running next release 😞 

## Checklist


wepudt:
cherrypick from release

PR #2755, opened 2024-02-15
Pass in the platform during CodeModulesImage pull

0sewa0:
## Description

Passes in the `Arch` of the operator during codemodules image pull, so we can pull arm images if needed.

the `defaultOS` I got from the actual source for the image library.

~TODO: We still need to verify if our `arc.Arch` is good enough that the image-library/registry will understand.~
- The `defaultArchitecture` in the image library is `amd64`, however our `arch.Arch` for that is `x86`, will it work? - NO
  - This is why new consts were introduced 

## How can this be tested?

Can an ARM `CodeModulesImage` be pulled on an ARM nodes? (if an image manifest is given)
- Obviously, we should still work on AMD nodes with AMD images :D 

(I tested with images from: https://hub.docker.com/r/dynatrace/dynatrace-codemodules/tags)

## Checklist


pkg/injection/codemodule/installer/image/unpack.go
```diff
@@ -44,7 +49,14 @@ func (installer Installer) pullImageInfo(imageName string) (*containerv1.Image,
 		return nil, errors.WithMessagef(err, "parsing reference %q:", imageName)
 	}
 
-	image, err := remote.Image(ref, remote.WithContext(context.TODO()), remote.WithAuthFromKeychain(installer.keychain), remote.WithTransport(installer.transport))
+	image, err := remote.Image(ref, remote.WithContext(context.TODO()),
+		remote.WithAuthFromKeychain(installer.keychain),
+		remote.WithTransport(installer.transport),
+		remote.WithPlatform(containerv1.Platform{
+			OS:           defaultOS,
```
luhi-DT, line 56: do you have to pass OS?

0sewa0: pretty sure yes
the "true" default value would be an empty string, which is probably problematic if you look at an image manifest, because you tend to see stuff like:
image

andriisoldatenko: yes, because in docker it's called platform=${platform-os}/${platform-arch},
in podman cli client it's has separate options to combine.

andriisoldatenko: @0sewa0 is any chance to test it?

andriisoldatenko (approved):

pkg/injection/codemodule/installer/image/unpack.go
0sewa0: test to not specify?

andriisoldatenko: yeah, but it doesn't make any sense. Ignore my comment

0sewa0: 🫡

PR #2759, opened 2024-02-15
Use the platform for all image related operations

0sewa0:
## Description

In https://github.com/Dynatrace/dynatrace-operator/commit/eab775d9c20655fa23194cf47e32c01a49bfb578, I started to use the platform for the image pull, but then I realised that we have other calls where we can pass the platform.

## How can this be tested?

Same as https://github.com/Dynatrace/dynatrace-operator/pull/2755

## Checklist


luhi-DT (approved):
tested, works

PR #2773, opened 2024-02-19
Enable linters (makezero, nilnil, nlreturn, fieldalignment)

0sewa0:
## Description

My quick thoughts about the linters:
- `makezero`: Makes sense, small easy fix
- `nilnil`: Most of the time I just `nolint`-ed it, because we used the double `nil` return as an actual logical step, and I didn't see a better way to do it. (so maybe not needed?)
- `nlreturn`: Feels arbitrary to me, most of the changes are because of it. (IMO, wsl is enough)
   - Sidenote: AFAIK there was/is no autofix for this
- `fieldalignment`: This is hell without [betteralign](https://github.com/dkorunic/betteralign), and even that needed to be run multiple times to get everything. (because of composite types)

## How can this be tested?

CI (or run linting locally)

## Checklist


andriisoldatenko (approved):

.golangci.yml
```diff
@@ -248,6 +250,9 @@ linters:
   - whitespace                # Tool for detection of leading and trailing whitespace.
   - wsl                       # Add or remove empty lines.
   - zerologlint               # Detects the wrong usage of zerolog that a user forgets to dispatch with Send or Msg.
+  - makezero                  # Finds slice declarations with non-zero initial length.
```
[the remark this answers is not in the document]
0sewa0: done

.golangci.yml
```diff
@@ -220,13 +222,16 @@ linters:
   - importas                  # Enforces consistent import aliases.
   - ineffassign               # Detects when assignments to existing variables are not used
   - loggercheck               # Checks key value pairs for common logger libraries (kitlog,klog,logr,zap).
+  - makezero                  # Finds slice declarations with non-zero initial length.
   - mirror                    # Reports wrong mirror patterns of bytes/strings usage.
   - misspell                  # Finds commonly misspelled English words in comments.
   - nakedret                  # Checks that functions with naked returns are not longer than a maximum size (can be zero).
   - nilerr                    # Finds the code that returns nil even if it checks that the error is not nil.
+  - nilnil                    # Checks that there is no simultaneous return of nil error and an invalid value.
```
waodim, line 230: IMO we can get rid of this linter as it is no-linted almost everywhere

waodim: But if you think it is not worth removing feel free to leave it and merge

andriisoldatenko: @waodim I would recommend you to check these examples https://github.com/Antonboom/nilnil?tab=readme-ov-file#examples

waodim: very good point thanks :slightly_smiling_face: 

I am fine with enabling it and the rest as well :+1:

waodim (approved):
I am fine with enabling those linters

aorcholski (dismissed):
LGTM - `Status` fields are printed in alphabetical order by `kubectl describe` command so different alignment of DynakubeStatus struct doesn't matter

albertogdd (approved):

andriisoldatenko (approved):

PR #2777, opened 2024-02-20
Refactor ProcessmoduleConfig code and add missing unit test

waodim:
## Description

This is a refactor of some parts of the `ProcessModuleConfig` handling. Namely: the `OneAgentProvisioner` sets some values which are currently called every time per `CSI Driver`. Now that they are moved to the `Dynakube` reconciler, they are called on `Dynakube` reconcile. In order to make sure that the `OneAgentTenantToken` is ready at the time when the `ProcessModuleConfig` is getting reconciled. This is because due to the refactor there `OneAgentConnectionInfo` info is added to the `LatestProcessModuleConfig`.

Also: I did some mini refactor in c1e69c33ab31f2fde456bb1a6a2f1fc9c405d4e7

## How can this be tested?

Unit tests

and/or

Deploy operator, inject into namespace and check dynakube-pmc-secret 

## Checklist


pkg/controllers/dynakube/processmoduleconfigsecret/reconciler.go
```diff
@@ -143,6 +171,22 @@ func (r *Reconciler) prepareSecret(ctx context.Context) (*corev1.Secret, error)
 	return newSecret, err
 }
 
+func (r *Reconciler) getAgentTenantToken(ctx context.Context) (string, error) {
```
aorcholski, line 174: it's duplicated [GetDataFromSecretName](https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/util/kubeobjects/secret/secret.go#L180) function. Error message wrapper could be "secret not found or invalid" and details comes from the err returned by the function

waodim: Done in https://github.com/Dynatrace/dynatrace-operator/pull/2777/commits/3ac9f4a406e5def836a71c7ddb0c336baee14ffc

pkg/controllers/dynakube/controller.go
```diff
@@ -261,14 +261,14 @@ func (controller *Controller) reconcileDynaKube(ctx context.Context, dynakube *d
...
-	err = processmoduleconfigsecret.NewReconciler(
-		controller.client, controller.apiReader, dynatraceClient, dynakube, controller.scheme, timeprovider.New()).
-		Reconcile(ctx)
+	err = controller.reconcileComponents(ctx, dynatraceClient, istioClient, dynakube)
 	if err != nil {
 		return err
 	}
 
-	return controller.reconcileComponents(ctx, dynatraceClient, istioClient, dynakube)
+	return processmoduleconfigsecret.NewReconciler(
```
0sewa0, line 269: why here?

ProcessModuleConfig only matters if there is injection (cloudnative/appmon), so it could be in the [injection reconciler](https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/controllers/dynakube/injection/reconciler.go#L62-L91)

0sewa0 (changes requested):

pkg/controllers/dynakube/controller.go
waodim: It used to be here and for the refactoring to work I needed to move reconcilers because the oneagent tenant token needs to be created in order to set oneagent connection info. But I will have a look if I can move it into the injection reconciler.

pkg/controllers/dynakube/controller.go
```diff
@@ -261,14 +261,14 @@ func (controller *Controller) reconcileDynaKube(ctx context.Context, dynakube *d
...
+		controller.client, controller.apiReader, dynatraceClient, dynakube, controller.scheme, timeprovider.New()).
+		Reconcile(ctx)
```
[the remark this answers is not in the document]
waodim: It used to be here and for the refactoring to work I needed to move reconcilers because the oneagent tenant token needs to be created in order to set oneagent connection info. 

But @0sewa0 also commented and I might move processmoduleconfig reconciler into the injection reconciler. I am currently looking into it

wepudt (dismissed):
lgtm

deployed a DynaKube with cloudNative all worked fine

pkg/controllers/dynakube/controller.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/2777/commits/21e52638bf7c80d7a47af88adb0f0875282826bc

pkg/controllers/dynakube/controller.go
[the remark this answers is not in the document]
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/2777/commits/21e52638bf7c80d7a47af88adb0f0875282826bc

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -79,6 +87,13 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 		setupErrors = append(setupErrors, err)
 	}
 
+	err := processmoduleconfigsecret.NewReconciler(
+		r.client, r.apiReader, r.dynatraceClient, r.dynakube, r.scheme, timeprovider.New()).
+		Reconcile(ctx)
+	if err != nil {
+		setupErrors = append(setupErrors, err)
+	}
+
```
0sewa0, line 96: create the reconciler in the constructor, like the `versionReconciler`, so it can be mocked

also use the same `timeprovider` for the reconcilers (however I am not the biggest fan of using `.Freeze()` for no real reason)

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -43,6 +48,7 @@ func NewReconciler(
 	istioClient *istio.Client,
 	fs afero.Afero,
 	dynakube *dynatracev1beta1.DynaKube,
+	scheme *runtime.Scheme,
```
0sewa0, line 51: have the order be 
```
	client client.Client,
	apiReader client.Reader,
	scheme *runtime.Scheme,
	...
```

(we do this "everywhere")

also, fun fact `fs afero.Afero` is not needed, its a leftover in the `versionReconciler` (I encountered it in [my pr](https://github.com/Dynatrace/dynatrace-operator/pull/2781))

pkg/controllers/dynakube/controller_test.go
```diff
@@ -455,7 +455,7 @@ func createVersionReconcilerBuilder(reconciler *mockversion.Reconciler) version.
 }
 
 func createInjectionReconcilerBuilder(reconciler *mockinjection.Reconciler) injection.ReconcilerBuilder {
-	return func(client client.Client, apiReader client.Reader, dynatraceClient dtclient.Client, istioClient *istio.Client, fs afero.Afero, dynakube *dynatracev1beta1.DynaKube) controllers.Reconciler {
+	return func(client client.Client, apiReader client.Reader, dynatraceClient dtclient.Client, istioClient *istio.Client, fs afero.Afero, dynakube *dynatracev1beta1.DynaKube, scheme *runtime.Scheme) controllers.Reconciler {
```
0sewa0, line 458:
```suggestion
	return func(_ client.Client, _ client.Reader, _*runtime.Scheme, _ dtclient.Client, _ *istio.Client, _ *dynatracev1beta1.DynaKube) controllers.Reconciler {
```

0sewa0 (changes requested):

pkg/controllers/dynakube/controller_test.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/2777/commits/39d3778db07cc9d89ba637c88bb091975411c30b

pkg/controllers/dynakube/injection/reconciler.go
waodim: basically done in https://github.com/Dynatrace/dynatrace-operator/pull/2777/commits/39d3778db07cc9d89ba637c88bb091975411c30b

BUT i was not sure, should `fs` be removed as it is not needed anymore? Because I did not do that yet.

pkg/controllers/dynakube/injection/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/2777/commits/39d3778db07cc9d89ba637c88bb091975411c30b

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -9,26 +9,32 @@ import (
...
-	apiReader         client.Reader
-	dynakube          *dynatracev1beta1.DynaKube
-	istioReconciler   istio.Reconciler
-	versionReconciler version.Reconciler
+	client              client.Client
+	apiReader           client.Reader
+	dynakube            *dynatracev1beta1.DynaKube
+	istioReconciler     istio.Reconciler
+	versionReconciler   version.Reconciler
+	pmcSecretreconciler processmoduleconfigsecret.Reconciler
```
0sewa0, line 29: use the `interface` not the exact `struct`, so it can be mocked fully

- currently you can "mock" it by creating a `processmoduleconfigsecret.Reconciler` with fakeClients and stuff, but that is not necessarily what we want always
  - considering the `injection.Reconciler` as a "unit", it should only care that that given reconciler was called
  - you can have a bigger test where you would care how each reconciler actually interact, but that is a system/e2e test and not a unittest. (which you can still do if you use the `interface` instead of the `struct`)

0sewa0 (changes requested):

pkg/controllers/dynakube/injection/reconciler.go
0sewa0: `controllers.Reconciler` is the `interface` I am talking about 😅

waodim: if i understood it correctly I did it in https://github.com/Dynatrace/dynatrace-operator/pull/2777/commits/8d3c72599d3747501778e8602d73690edcee9bce :sweat_smile:

PR #2781, opened 2024-02-21
Move connection-info and version reconcile to oneagent reconciler

0sewa0:
## Description

A bit of refactor:
- Connection-info reconciler was split, so its each component has its own specific one
  - Duo to the injection-reconciler and oneagent-reconciler having some shared requirements there is some duplicate code
     - No double work should be done, thanks to the use of timestamps / early return in case of special errors
     - We want the components to be handled independently, so some duplication is unavoidable
- Some smaller renames
- Removed unused properties from the version-reconciler

Moved the version and connection-info reconcilers, that were only used by the oneagent-reconciler, inside the oneagent-reconciler.
Also now the oneagent-reconciler is mocked(or at least mockable) in the dynakube reconciler.

Big sidenote:
I removed a comment that stated something that is no longer true: (I am highlighting this, because I think its important)
   - Due to the networkzone support we actually no longer can inject into activegates that is deployed by us super easily
     - This is because we don't want to inject until the connection-info for the oneagents are present
     - But in certain cases the networkzone will be created by the activegate we deploy
       - So naturally, the activegate is the first that should be reconciled
     - But if the activegate is the first, then the injection reconcile must run after, so the activegate can't be easily injected
   
 Could this be fixed: probably, but the solution will not be nice nor clear. (deadlocks are fun)

## How can this be tested?

No major change in behaviour

Small changes:
- The connection-info timestamp is checked first, so if its outdated we don't do an extra request to see if the secret is there or not
- Because of the split of the connection-info reconciler there are 2 different loggers, 1 for oneagent and 1 for activegate connection info reconciler

## Checklist

- [x] Unit tests have been updated/added (TODO: Add more)


wepudt (approved):
lgtm

waodim (approved):
LGTM codewise, unit tests ran through - deployed operator and tested it and it worked just fine

PR #2788, opened 2024-02-22
Change default OneAgent and ActiveGate images to use raw tag instead of latest

waodim:
## Description

As the operator should always use the non-configured images (i.e. images with `raw` tag) for `OneAgent` and `ActiveGate`. With this the default images are now changed to use the `raw` tag instead of `latest`.

This change is done because the operator always configures the images by supplying config parameters, so there is no need to take the pre-configured image with the latest tag.

## How can this be tested?

Unit tests

and/or

Deploy operator and a dynakube with oneagent and activegate, check their respective images and check if functionality is still the same.

## Checklist


wepudt (approved):
lgtm

PR #2796, opened 2024-02-23
WIP Add workload annotations

aorcholski:
## Description

Please include the following:

- A summary of the change

## How can this be tested?

## Checklist


pkg/webhook/config.go
```diff
@@ -58,4 +58,9 @@ const (
```
waodim, line 58: tested it and it works, code also looks good to me, but as it is still WIP according to title maybe something will still change, I guess?

wepudt: same- also deployed on my cluster- looking good to me!

aorcholski: IMHO no more changes needed

gkrenn (approved):

0sewa0 (approved):

PR #2816, opened 2024-02-28
Add condition for CodeModuleVersion and ProcessModuleConfig

0sewa0:
## Description

First `Conditions` related PR, review it in a way that following `Conditions` should follow the same pattern.

Adds 2 `Conditions`:
- `CodeModuleVersion`, which is for the version-reconciler
  - A bit of a hacky solution because of the version reconciler structure, but it works 
  - Only serves as more info, not used directly for anything
- `ProcessModuleConfig` which is for the process-module-config-secret reconciler
  - Secret creation/update related, so I moved these `Conditions` into a utils package. 
  - ⚠️ Replaces the timestamp for the Dynatrace API request limit feature
     - "Basic" rule: A "transition" has (True => False, False=>True) for the timestamp to change, that is why there is an `Empty` Reason, so this transition happens, also its a nice failsafe 

## How can this be tested?

Deploy a dynakube that uses the CodeModules.
- Check the status for the `Conditions`

## Checklist


pkg/controllers/dynakube/processmoduleconfigsecret/conditions.go
```diff
@@ -0,0 +1,5 @@
+package processmoduleconfigsecret
+
+const (
+	conditionType = "ProcessModuleConfig"
```
waodim, line 4:
```suggestion
	pmcConditionType = "ProcessModuleConfig"
```

waodim (commented):
Apart from this minor nitpick lgtm, tested it and it works. I would still prefer someone else to also have a look before we merge.

pkg/controllers/dynakube/version/reconciler_test.go
```diff
@@ -341,3 +341,14 @@ func mockLatestAgentVersion(mockClient *dtclientmock.Client, latestVersion strin
 func mockLatestActiveGateVersion(mockClient *dtclientmock.Client, latestVersion string) {
 	mockClient.On("GetLatestActiveGateVersion", mock.AnythingOfType("context.backgroundCtx"), mock.Anything).Return(latestVersion, nil)
 }
+
+func createBoomDTClient(t *testing.T) dtclient.Client {
```
luhi-DT, line 345: Maybe rename to ErrorClient :D

pkg/controllers/dynakube/version/conditions.go
```diff
@@ -0,0 +1,31 @@
...
+	condition := metav1.Condition{
+		Type:    conditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  verifiedReason,
+		Message: "Version verified for component.",
+	}
+	_ = meta.SetStatusCondition(conditions, condition)
+}
+
+func setVerificationSkippedReasonCondition(conditions *[]metav1.Condition, conditionType string) {
```
luhi-DT, line 23:
```suggestion
func setVersionUpdateSkippedReasonCondition(conditions *[]metav1.Condition, conditionType string) {
```

0sewa0: the version update wasn't necessarily skipped, because if the custom version/image was changed we would still set the status to be `VerificationSkipped`, because we didn't verify anything

the name is not the best but I tried to convey the fact that if custom stuff is used then we will just "go with it", instead of trying to be smart about it

pkg/controllers/dynakube/processmoduleconfigsecret/conditions.go
0sewa0: done

pkg/controllers/dynakube/version/reconciler_test.go
0sewa0: done

luhi-DT (approved):

PR #2837, opened 2024-03-05
Data migration of new scheme

andriisoldatenko:
## Description

This PR migrates data from old schema to new schema:

old scheme:
```
- dynakubes
- osagent_volumes
- volumes
``` 

new:
```
app_mounts
code_modules
os_mounts
tenant_configs
volume_meta
```

Note: `migrations` is table where we should see history of applied migrations:
```
select * from migrations;
```

## How can this be tested?

Deploy to any env, and check that new tables are not empty.

## Checklist


pkg/controllers/csi/metadata/models.go
```diff
@@ -14,20 +14,20 @@ type TimeStampedModel struct {
 
 // TenantConfig holds info about a given configuration for a tenant.
 type TenantConfig struct {
-	CodeModule CodeModule `gorm:"foreignKey:Version;references:DowloadedCodeModuleVersion"`
+	CodeModule CodeModule `gorm:"foreignKey:DowloadedCodeModuleVersion;references:Version"`
```
wepudt, line 17: do we need that? seems like "doppeltgemoppelt" to me as Version is already declared primary key in the CodeModule struct?

pkg/controllers/csi/metadata/models.go
```diff
@@ -37,15 +37,15 @@ type OSMount struct {
...
+	VolumeMetaID  string `gorm:"not null"`
 	Location      string `gorm:"not null"`
 	MountAttempts int64  `gorm:"not null"`
 }
 
 // AppMount keeps track of our mounts to user applications, where we provide the codemodules.
 type AppMount struct {
 	VolumeMeta VolumeMeta
-	CodeModule CodeModule `gorm:"foreignKey:Version;references:CodeModuleVersion"`
+	CodeModule CodeModule `gorm:"foreignKey:CodeModuleVersion;references:Version"`
```
wepudt, line 48: I have added ``gorm:"foreignKey:VolumeMetaID"`` as this should reference a VolumeMeta- correct?

pkg/controllers/csi/metadata/models.go
```diff
@@ -37,15 +37,15 @@ type OSMount struct {
...
 	TimeStampedModel
 	VolumeMetaID      string
```
wepudt, line 50: also we are missing a primary key for AppMount- could we use `VolumeMetaID` for that? Would make OPS on an AppMount much easier imho...

pkg/controllers/csi/metadata/models.go
```diff
@@ -14,20 +14,20 @@ type TimeStampedModel struct {
...
+	CodeModule CodeModule `gorm:"foreignKey:DowloadedCodeModuleVersion;references:Version"`
 	TimeStampedModel
-	UID                        string `gorm:"primaryKey" json:"UID,omitempty"`
-	Name                       string `gorm:"not null"`
-	DowloadedCodeModuleVersion string
-	ConfigDirPath              string `gorm:"not null"`
-	TenantUUID                 string
+	UID                        string  `gorm:"primaryKey" json:"UID,omitempty"`
+	Name                       string  `gorm:"not null"`
+	DowloadedCodeModuleVersion string  `gorm:"not null"`
```
wepudt, line 21: we have a typo here: DowNloaded... fix also needs to be in CodeModule association comment further up.

pkg/controllers/csi/metadata/models.go
```diff
@@ -37,15 +37,15 @@ type OSMount struct {
...
 	TenantUUID    string `gorm:"primaryKey" json:"tenantUUID,omitempty"`
-	VolumeMetaID  string
+	VolumeMetaID  string `gorm:"not null"`
 	Location      string `gorm:"not null"`
 	MountAttempts int64  `gorm:"not null"`
 }
 
 // AppMount keeps track of our mounts to user applications, where we provide the codemodules.
 type AppMount struct {
 	VolumeMeta VolumeMeta
```
wepudt, line 47: I have added gorm:"foreignKey:VolumeMetaID" as this should reference a VolumeMeta- correct?


```go
VolumeMeta VolumeMeta `gorm:"foreignKey:VolumeMetaID"`
```

andriisoldatenko: This is redundant, it will work without this line.

andriisoldatenko: Because if you introspect table you can see it works without this extra annotation:
image

pkg/controllers/csi/metadata/models.go
andriisoldatenko: fixed in 43eb3821

pkg/controllers/csi/metadata/models.go
wepudt: true... I just like the explicit- in case of renaming it could save us a headache

pkg/controllers/csi/metadata/models.go
andriisoldatenko: we need, because otherwise it won't find column, column name doesn't fit into GORM convention name for FK.

andriisoldatenko: double checked we need it,otherwise GORM failing:
```
2024/03/05 10:50:08 /Users/andrii.soldatenko/work/dynatrace-operator/pkg/controllers/csi/metadata/sqlite_test.go:674 FOREIGN KEY constraint failed
[0.772ms] [rows:0] INSERT INTO `tenant_configs` (`created_at`,`updated_at`,`deleted_at`,`uid`,`name`,`downloaded_code_module_version`,`config_dir_path`,`tenant_uuid`,`max_failed_mount_attempts`) VALUES ("2024-03-05 10:50:08.341","2024-03-05 10:50:08.341",NULL,"","dynakube","1.288.0.20240227-104015","/data/mts86258/dynakube/config","mts86258",10)
FOREIGN KEY constraint failed
<nil>
```

pkg/controllers/csi/metadata/models.go
andriisoldatenko: fixed.

andriisoldatenko:
@wepudt PTAL!

pkg/controllers/csi/metadata/migrations.go
```diff
@@ -0,0 +1,93 @@
...
+	var dynakubes []Dynakube
+
+	tx.Table("dynakubes").Find(&dynakubes)
+
+	for _, d := range dynakubes {
+		tc := TenantConfig{
+			Name:                        d.Name,
+			TenantUUID:                  d.TenantUUID,
+			ConfigDirPath:               filepath.Join(filepath.Join(dtcsi.DataPath, d.TenantUUID), d.Name, dtcsi.SharedAgentConfigDir),
+			DownloadedCodeModuleVersion: d.LatestVersion,
```
0sewa0, line 20: this could be `ImageDigest` as well

[Relevant code](https://github.com/Dynatrace/dynatrace-operator/blob/efe236974140ed208f4dfda94ad0351599068770/pkg/controllers/csi/provisioner/controller.go#L225-L228)

Context: Downloaded images and dowloaded zips used to be stored in a different location, and the way we differentiated "where to look" is by checking if the `ImageDigest` or the `LatestVersion` is set.
But as of right now we only store things in the same folder. (also we are possible moving away from digests, so that is that :D) 
- This "where to look" was/is only relevant during download, mount, and cleanup, and not during unmount, so that is why you do not see it in the `volumes` table. (its confusing, I know 😅 )

wepudt (approved):
lgtm

pkg/controllers/csi/metadata/migrations.go
```diff
@@ -0,0 +1,93 @@
...
+			MaxFailedMountAttempts:      int64(d.MaxFailedMountAttempts),
+		}
+
+		result := tx.Create(&tc)
+		if result.Error != nil {
+			log.Error(result.Error, "failed to create TenantConfig")
+		}
+
+		cm := CodeModule{
+			Version:  d.LatestVersion,
```
0sewa0, line 30: same as above, it can be the `ImageDigest`

pkg/controllers/csi/metadata/migrations.go
```diff
@@ -0,0 +1,93 @@
...
+func dataMigration(tx *gorm.DB) error {
+	var dynakubes []Dynakube
+
+	tx.Table("dynakubes").Find(&dynakubes)
+
+	for _, d := range dynakubes {
+		tc := TenantConfig{
+			Name:                        d.Name,
+			TenantUUID:                  d.TenantUUID,
+			ConfigDirPath:               filepath.Join(filepath.Join(dtcsi.DataPath, d.TenantUUID), d.Name, dtcsi.SharedAgentConfigDir),
```
0sewa0, line 19: why not use the `PathResolver`, to remain consistent with the rest of the code? (its even in the same package 😅 )

pkg/controllers/csi/metadata/models.go
```diff
@@ -14,20 +14,20 @@ type TimeStampedModel struct {
...
-	UID                        string `gorm:"primaryKey" json:"UID,omitempty"`
-	Name                       string `gorm:"not null"`
-	DowloadedCodeModuleVersion string
-	ConfigDirPath              string `gorm:"not null"`
-	TenantUUID                 string
-	OSMount                    OSMount `gorm:"foreignKey:TenantUUID;references:TenantUUID"`
-	MaxFailedMountAttempts     int64   `gorm:"default:10"`
+	UID                         string  `gorm:"primaryKey" json:"UID,omitempty"`
+	Name                        string  `gorm:"not null"`
+	DownloadedCodeModuleVersion string  `gorm:"not null"`
```
0sewa0, line 21: would this `not null` constraint work in case of `HostMonitoring`?

Because in case of `HostMonitoring` we only do an `OSMount` , in which case we shouldn't download a CodeModule.

(I should have noticed this in the original schema PR 😬 )

0sewa0 (changes requested):

pkg/controllers/csi/metadata/migrations.go
andriisoldatenko: Because it's unclear how to creat an instance of it... but I can do it.

pkg/controllers/csi/metadata/models.go
andriisoldatenko: @0sewa0 without not null you can't set properly constraint, so we need remove link in this case.

pkg/controllers/csi/metadata/migrations.go
0sewa0: you can create a constructor for it, or just `PathResolver{RootDir: dtcsi.DataPath}`

pkg/controllers/csi/metadata/models.go
0sewa0: [but that is how it works](https://github.com/Dynatrace/dynatrace-operator/blob/efe236974140ed208f4dfda94ad0351599068770/pkg/controllers/csi/provisioner/controller.go#L125-L129)😞

pkg/controllers/csi/metadata/migrations.go
andriisoldatenko: fixed in 2639aa76

pkg/controllers/csi/metadata/migrations.go
andriisoldatenko: fixed in 2639aa76

pkg/controllers/csi/metadata/models.go
andriisoldatenko: it's okey we can remove constraint, and just use as string, and for HostMonitoring it will be null.

pkg/controllers/csi/metadata/migrations.go
andriisoldatenko: fixed 23abf8ea

pkg/controllers/csi/metadata/migrations.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		result := tx.Create(&tc)
+		if result.Error != nil {
+			log.Error(result.Error, "failed to create TenantConfig")
+		}
+
+		cm := CodeModule{
+			Version:  version,
+			Location: pr.AgentSharedBinaryDirBase(),
```
0sewa0, line 38: It should be [AgentSharedBinaryDirForAgent](https://github.com/Dynatrace/dynatrace-operator/blob/784e35fc06bca116c3cbdd2d6f775703729f5972/pkg/controllers/csi/metadata/path_resolver.go#L43-L43) as you want to exact location

andriisoldatenko: fixed cbdf7da2

pkg/controllers/csi/metadata/models.go
andriisoldatenko: fixed in cbdf7da2

pkg/controllers/csi/metadata/migrations.go
```diff
@@ -0,0 +1,112 @@
...
+
+func volumeIn(volumeID string, volumes []Volume) bool {
+	for _, v := range volumes {
+		if volumeID == v.VolumeID {
+			return true
+		}
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
0sewa0: You should not add the `OSMount` if its `!Mounted`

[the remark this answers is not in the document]
andriisoldatenko: fixed in bb7051cc

andriisoldatenko:
@0sewa0 PTAL

pkg/controllers/csi/metadata/migrations.go
```diff
@@ -0,0 +1,146 @@
...
+			mountAttempts = 1
+		}
+
+		om := OSMount{
+			TenantUUID:    ov.TenantUUID,
+			VolumeMetaID:  ov.VolumeID,
+			Location:      pr.AgentRunDirForVolume(ov.TenantUUID, ov.VolumeID),
+			MountAttempts: mountAttempts,
+		}
+		tx.Create(&om)
```
0sewa0, line 130:
```suggestion
		result = tx.Create(&om)
		if result.Error != nil {
			log.Error(result.Error, "failed to create OSMount")
		}
```

0sewa0: I think there is a problem here, because when I try testing an upgrade, the `VolumeMeta` gets created for the existing `OsAgentVolume`, but the `OsMount` doesn't

And if we don't log the error, then we will never know

0sewa0 (changes requested):

andriisoldatenko:
@0sewa0  adjusted based on our last conversation
PTAL ee2ef9e8

pkg/controllers/csi/metadata/migrations.go
```diff
@@ -0,0 +1,155 @@
...
+			log.Error(result.Error, "failed to create VolumeMeta")
+		}
+
+		var mountAttempts int64
+		if ov.Mounted {
+			mountAttempts = 1
+		}
+
+		tc := TenantConfig{TenantUUID: ov.TenantUUID}
+		tx.First(&tc)
```
0sewa0, line 126:
```suggestion

		// This is a workaround for not having enough information in the current database tables to migrate 100% correctly.
		// This is fine as we don't currently use this information for anything.
		tc := TenantConfig{TenantUUID: ov.TenantUUID}
		tx.First(&tc)
```

0sewa0 (dismissed):

0sewa0 (approved):

PR #2840, opened 2024-03-05
Run E2E tests in latest FLC version environments

albertogdd:
## Description

Update FLC environments where we run our E2E tests. Now, we use the latest version ones.

## How can this be tested?

https://github.com/Dynatrace/dynatrace-operator/actions/runs/8170692777

## Checklist

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -2,7 +2,7 @@ name: E2E tests
 
 on:
   schedule:
-    # every work day at 5:00 AM
+    # every work day at 00:00 UTC
     - cron: 0 0 * * 1-5
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -2,7 +2,7 @@ name: E2E tests
 
 on:
   schedule:
-    # every work day at 5:00 AM
+    # every work day at 00:00 AM
```
chrismuellner, line 5: should we just say midnight? what time zone?

.github/workflows/e2e-tests.yaml
```diff
@@ -13,11 +13,11 @@ on:
...
 jobs:
   run-in-k8s:
-    name: Run in Kubernetes 1.26 (${{ github.event.inputs.target }})
+    name: Run in Kubernetes latest (${{ github.event.inputs.target }})
     environment: E2E
     env:
-      FLC_NAMESPACE: dto-k8s-ondemand
-      FLC_ENVIRONMENT: dto-k8s-1-26
+      FLC_NAMESPACE: dto-daily
+      FLC_ENVIRONMENT: dto-k8s-latest-flc
```
chrismuellner, line 20: somewhat unrelated to this specific PR, but is there a reason we are not using a [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-using-a-single-dimension-matrix) to reuse the steps? It seems only the `FLC_ENVIRONMENT` is different between `run-in-k8s` and `run-in-ocp` jobs

chrismuellner (changes requested):

.github/workflows/e2e-tests.yaml
albertogdd: I just wrote a user story today about this, you can check it in the epic (I have no access to Jira now)

.github/workflows/e2e-tests.yaml
albertogdd: 00:00 is more clear to me than midnight, just IMO
I will add the timezone 👍

albertogdd: Updated

chrismuellner (approved):

PR #2849, opened 2024-03-06
Create a GORM Client and Implementation of the CSI DB API

wepudt:
## Description

Creates an interface that uses GORM to access the sqlite DB.

## How can this be tested?

Run the created unit-tests.

## Checklist


pkg/controllers/csi/metadata/sqlite.go
```diff
@@ -304,7 +277,7 @@ func (access *SqliteAccess) executeAlterStatement(ctx context.Context, statement
 	return nil
 }
 
-// Setup connects to the database and creates the necessary tables if they don't exist
+// setup connects to the database and creates the necessary tables if they don't exist
```
luhi-DT, line 280: why?

wepudt: that's strange- didn't do it on purpose- I suppose it's a refactoring gone wrong by goland

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,187 @@
...
+	// we should explicitly enable foreign_keys for sqlite
+	if strings.Contains(path, "?") {
+		path += "&_foreign_keys=on"
+	} else {
+		path += "?_foreign_keys=on"
+	}
+
+	db, err := gorm.Open(sqlite.Open(path), &gorm.Config{Logger: logger.Default})
+
+	// db = db.Session(&gorm.Session{FullSaveAssociations: true})
```
waodim, line 53: Should we get rid of this commented out line?

andriisoldatenko:
@wepudt could you please remove unssesary files like:
[pkg/controllers/csi/metadata/publish.db](https://github.com/Dynatrace/dynatrace-operator/pull/2849/files#diff-bad4d805b6ed83f9d90aa22e019bd0f1431bfcbdfafbb163c12bf8a24138590f)

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,186 @@
...
+	"context"
+	"strings"
+
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
+)
+
+type CSIDriverDBAccess interface {
```
andriisoldatenko, line 13: I think since we are already inside `csi/` subfolder, CSI prefix is redundant.

aorcholski:
`publish.db` and `reconcile.db` are not source files. They are unittests related files. Maybe it would be better to move them to the `test/unittestdata/...` directory.

wepudt:
> @wepudt could you please remove unssesary files like: [pkg/controllers/csi/metadata/publish.db](https://github.com/Dynatrace/dynatrace-operator/pull/2849/files#diff-bad4d805b6ed83f9d90aa22e019bd0f1431bfcbdfafbb163c12bf8a24138590f)

I use this publish.db for 11 tests, the reconcile.db for 3.

Idea was to mimic db state from our Reconciler / CSI Driver

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (conn *DBConn) DeleteCodeModule(ctx context.Context, codeModule *CodeModule) error {
+	return conn.db.WithContext(ctx).Delete(codeModule).Error
+}
+
+func (conn *DBConn) ReadTenantConfigByTenantUUID(ctx context.Context, uid string) (*TenantConfig, error) {
+	var tenantConfig TenantConfig
+
+	result := conn.db.WithContext(ctx).First(&tenantConfig, "tenant_uuid = ?", uid)
```
andriisoldatenko, line 126: you can do it in more elegant way IMO:
```
tenantConfig := TenantConfig{TenantUUID: uid}
```
and then:
```
result := conn.db.WithContext(ctx).First(&tenantConfig)
```

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (conn *DBConn) UpdateOSMount(ctx context.Context, osMount *OSMount) error {
+	return conn.db.WithContext(ctx).Updates(osMount).Error
+}
+
+func (conn *DBConn) ReadOSMountByTenantUUID(ctx context.Context, tenantUUID string) (*OSMount, error) {
+	var osMount OSMount
+
+	result := conn.db.WithContext(ctx).Preload("VolumeMeta").First(&osMount, "tenant_uuid = ?", tenantUUID)
```
andriisoldatenko, line 145: same here

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,197 @@
...
+	if result.Error != nil {
+		return nil, result.Error
+	}
+
+	return &osMount, nil
+}
+
+func (conn *DBConn) DeleteOSMount(ctx context.Context, osMount *OSMount) error {
+	if osMount.VolumeMeta != (VolumeMeta{}) {
```
andriisoldatenko, line 154: I think better ti check if `VolumeMeta.ID != nil` and delete by ID.

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (conn *DBConn) UpdateAppMount(ctx context.Context, appMount *AppMount) error {
+	return conn.db.WithContext(ctx).Updates(appMount).Error
+}
+
+func (conn *DBConn) ReadAppMountByVolumeMetaID(ctx context.Context, volumeMetaID string) (*AppMount, error) {
+	var appMount AppMount
+
+	result := conn.db.WithContext(ctx).Preload("VolumeMeta").First(&appMount, "volume_meta_id = ?", volumeMetaID)
```
andriisoldatenko, line 172: same here

pkg/controllers/csi/metadata/sqlite_gorm_client_test.go
```diff
@@ -0,0 +1,392 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	db, err := NewDBAccess(TMP)
+
+	return &db, err
+}
+
+func copyFile(source, dest string) error {
```
andriisoldatenko, line 369: please remove this, because we can jsut do everything in memory.

pkg/controllers/csi/metadata/sqlite_gorm_client_test.go
```diff
@@ -0,0 +1,392 @@
...
+
+	assert.NotNil(t, appMount)
+	assert.Equal(t, "appmount1", appMount.VolumeMeta.ID)
+
+	err = db.DeleteAppMount(context.Background(), appMount)
+	require.NoError(t, err)
+
+	var result []AppMount
+
+	db.db.Raw(`SELECT * FROM app_mounts WHERE volume_meta_id = "appmount1"`).Scan(&result)
```
andriisoldatenko, line 337: why not using GORM? The purpose to migrate to GORM to avoid RAW sqls :)

pkg/controllers/csi/metadata/sqlite_gorm_client_test.go
```diff
@@ -0,0 +1,392 @@
...
+	require.NoError(t, err)
+
+	appMount.MountAttempts = 5
+
+	err = db.UpdateAppMount(context.Background(), appMount)
+	require.NoError(t, err)
+
+	var result []OSMount
+
+	db.db.Raw(`SELECT * FROM app_mounts WHERE volume_meta_id = "appmount1"`).Scan(&result)
```
andriisoldatenko, line 317: same here

pkg/controllers/csi/metadata/sqlite_gorm_client_test.go
```diff
@@ -0,0 +1,391 @@
...
+	ctxDB.Exec("INSERT INTO code_modules (version, location, created_at, updated_at, deleted_at) VALUES ('1.2.3', 'someplace', '2024-03-11 17:07:43.038661+01:00', '2024-03-11 17:07:43.038661+01:00', null);")
+	ctxDB.Exec("INSERT INTO tenant_configs (created_at, updated_at, deleted_at, uid, name, downloaded_code_module_version, config_dir_path, tenant_uuid, max_failed_mount_attempts) VALUES ('2024-03-11 16:42:39.323198+01:00', '2024-03-11 16:42:39.323198+01:00', null, '033dcff9-5c76-4b3a-9e3a-ea0a78bf0b3f', 'somename', '1.2.3', 'somewhere', 'abc123', 10);")
+}
+
+func setupPostPublishData(ctx context.Context, conn *DBConn) {
+	setupPostReconcileData(ctx, conn)
+
+	ctxDB := conn.db.WithContext(ctx)
+
+	ctxDB.Exec("INSERT INTO volume_meta (id, pod_uid, pod_name, pod_namespace, pod_service_account, created_at, updated_at, deleted_at) VALUES ('osmount1', 'pod1', 'podi', 'testnamespace', 'podsa', '2024-03-12 07:49:37.943527+01:00', '2024-03-12 07:49:37.943527+01:00', null);")
```
andriisoldatenko, line 381: why not using ORM and structs?

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,197 @@
...
+			&VolumeMeta{},
+		)
+		if err != nil {
+			return err
+		}
+		// all other constraints, indexes, etc...
+		return nil
+	})
+
+	_ = m.Migrate()
```
gkrenn, line 79: Is it a clever idea to add an auto migration before the actual migration, why not having all the migrationsteps in one array?

The problem I see here is that somewhen in the future as soon as we try a migration, the automigrate step could interfere with the predefined migration steps

@andriisoldatenko

andriisoldatenko: @gkrenn I didn't get what you mean? Could you please provide more details?

**Context why it's written this way:**
go-migrate strangely behaves if we combine Init migrations and just regular ones. So we decided with @0sewa0 to explicitly migrate the Initial schema, and then you put your migrations after line 81.

So, the new migration will look like it; you should copy the following:
```
gormigrate.New(conn.db, gormigrate.DefaultOptions, []*gormigrate.Migration{
		{
			ID:      "X",
			Migrate: <new migrations here>,
			Rollback: func(tx *gorm.DB) error {
				return nil
			},
		},
	}).Migrate()
```
Below line 81, and you're done. You never touch initial migration.

gkrenn: Ok thx, that's what I meant. - Interesting, if there are sideeffects if we combine init migrations and regular ones 🤔

andriisoldatenko: If you check my original PR very first few commits I did this way. It creates two records in migration table but never apply second migration lol. Then we spent some time to debug library to understand how it's working under the hood.

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,197 @@
...
+	"context"
+	"strings"
+
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
+)
+
+type DBAccess interface {
```
0sewa0, line 13: How are we handling the reading of soft-deleted entries?
Do our `Read...` functions consider this?

So, as of right now, the database is not used that much for the GC logic.
- Currently, we check what is on the filesystem, try to match a DB entry, and remove stuff the database doesn't know about.
   - This is bad because if we have to change the location of something, the logic for the GC has to know the old and new locations, which is rather annoying.
- In the future, we would like to do the other way around, so look at the database and decide what to remove from the filesystem.
  - To do this, we will need separate calls to look at what was soft-deleted. 
  - We can add those functions when we look more deeply at how the garbage collection should work.

0sewa0 (changes requested):

pkg/controllers/csi/metadata/sqlite_gorm_client.go
wepudt: Interestingly my test failed when called with "" as uid:
```go
_, err = db.ReadTenantConfigByTenantUUID(context.Background(), "")
require.Error(t, err)
``` 
I expect it to fail with an error, but it simply returns the first (and only entry) in the table.

pkg/controllers/csi/metadata/sqlite_gorm_client_test.go
andriisoldatenko: No need to test ORM, orm works it has it's own unit tests. Focus on testing our stuff. The whole purpose of rewriting is to remove raw SQL everywhere.

pkg/controllers/csi/metadata/sqlite_gorm_client.go
wepudt: true. I tried to make use of our datamodel/framework as much as possible and not duplicate checks, but for some params I was missing.

in this example I would not change the code bc the next check also fails with this code unexpectedly:

```go
_, err = db.ReadTenantConfigByTenantUUID(context.Background(), "unknown")
require.Error(t, err)
```

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,197 @@
...
+	DeleteAppMount(ctx context.Context, appMount *AppMount) error
+	ReadAppMountByVolumeMetaID(ctx context.Context, volumeMetaID string) (*AppMount, error)
+	ReadAppMounts(ctx context.Context) ([]AppMount, error)
+}
+
+type DBConn struct {
+	db *gorm.DB
+}
+
+var _ DBAccess = (*DBConn)(nil)
```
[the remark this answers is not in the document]
wepudt: I will go with the first, even though the second approach seems more clean.

But then I have issues with testability (setting up DBs and stuff)

pkg/controllers/csi/metadata/sqlite_gorm_client.go
wepudt: No our Read funcs only work with live data.
imho better to have a kind of "Cleanup" function that removes all entries older than X or something like this.

As I do not understand the requirements of GC clearly I wanted to get the API in and refactor when reworking the GC of the CSI driver.

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,219 @@
...
+
+type DBConn struct {
+	db *gorm.DB
+}
+
+var _ DBAccess = &DBConn{}
+
+// NewDBAccess creates a new gorm db connection to the database.
+func NewDBAccess(path string) (*DBConn, error) {
+	// we should explicitly enable foreign_keys for sqlite
```
StefanHauth, line 46:
```suggestion
	// we need to enable foreign_keys for sqlite because <add reason here>
```

pkg/controllers/csi/metadata/sqlite_gorm_client.go
wepudt: That won't work as stated above: in case of invalid tenantUUID it will create an invalid object instead of returning an error- at least that's what my tests showed.

pkg/controllers/csi/metadata/sqlite_gorm_client.go
wepudt: That won't work as stated above: in case of invalid tenantUUID it will create an invalid object instead of returning an error- at least that's what my tests showed.

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (conn *DBConn) CreateCodeModule(ctx context.Context, codeModule *CodeModule) error {
+	return conn.db.WithContext(ctx).Create(codeModule).Error
+}
+
+func (conn *DBConn) ReadCodeModuleByVersion(ctx context.Context, version string) (*CodeModule, error) {
+	var codeModule CodeModule
+
+	result := conn.db.WithContext(ctx).First(&codeModule, "version = ?", version)
```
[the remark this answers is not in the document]
wepudt: That won't work as stated above: in case of invalid tenantUUID it will create an invalid object instead of returning an error- at least that's what my tests showed.

pkg/controllers/csi/metadata/sqlite_gorm_client_test.go
```diff
@@ -0,0 +1,615 @@
...
+	require.Error(t, err)
+}
+
+func TestReadOSMount(t *testing.T) {
+	db, err := setupDB()
+	require.NoError(t, err)
+
+	setupPostPublishData(context.Background(), db)
+
+	appMount, err := db.ReadOSMountByTenantUUID(context.Background(), "abc123")
```
0sewa0, line 215: should be `osMount`

0sewa0 (dismissed):

pkg/controllers/csi/metadata/sqlite_gorm_client.go
wepudt: normally I would also separate APIs/interfaces by use-case, so I like your idea to have a different interface for GC.

pkg/controllers/csi/metadata/sqlite_gorm_client.go
andriisoldatenko: @wepudt what do you mean? for me works like a charm:

In case **Invalid** tenantUUID you get an error:
```
    sqlite_test.go:679: 
        	Error Trace:	/Users/andrii.soldatenko/work/dynatrace-operator/pkg/controllers/csi/metadata/sqlite_test.go:679
        	Error:      	Received unexpected error:
        	            	record not found
        	Test:       	TestSchemaMigration
--- FAIL: TestSchemaMigration (0.00s)

FAIL
```

My test:
```
func TestSchemaMigration(t *testing.T) {
	// new schema
	conn, err := NewDBAccess("file:csi_testdb?mode=memory")
	require.NoError(t, err)
	// new migrations
	err = conn.SchemaMigration(context.Background())
	require.NoError(t, err)

	tc := TenantConfig{
		Name:          "test",
		TenantUUID:    "test",
		ConfigDirPath: "/tmp",
	}
	require.NoError(t, conn.db.Create(&tc).Error)

	vm := VolumeMeta{
		ID:           "volume_id",
		PodUid:       "",
		PodName:      "php-pod",
		PodNamespace: "",
	}
	require.NoError(t, conn.db.Create(&vm).Error)

	om := OSMount{
		TenantUUID:      tc.TenantUUID,
		TenantConfigUID: tc.UID,
		VolumeMetaID:    vm.ID,
		MountAttempts:   10,
	}
	require.NoError(t, conn.db.Create(&om).Error)

	osMount := OSMount{TenantUUID: "test111"}
	require.NoError(t, conn.db.Preload("VolumeMeta").First(&osMount).Error)
}
```

wepudt:
> `publish.db` and `reconcile.db` are not source files. They are unittests related files. Maybe it would be better to move them to the `test/unittestdata/...` directory.

done

pkg/controllers/csi/metadata/sqlite_gorm_client_test.go
wepudt: done

pkg/controllers/csi/metadata/sqlite_gorm_client.go
```diff
@@ -0,0 +1,237 @@
...
+	if osMount == nil {
+		return errors.New("Cannot restore nil OSMount")
+	}
+
+	if !osMount.DeletedAt.Valid {
+		return nil
+	}
+
+	return conn.db.WithContext(ctx).Unscoped().Model(osMount).Where("tenant_uuid = ?", osMount.TenantUUID).Update("deleted_at", nil).Error
+}
```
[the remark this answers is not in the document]
wepudt: done

pkg/controllers/csi/metadata/sqlite_gorm_client.go
wepudt: did it for all cases where it works (for the 
`TenantConfig` it does not work for `TenantUUID`) and adapted param checks to pass all tests (until refactoring to `sql.NullString`)

pkg/controllers/csi/metadata/models.go
```diff
@@ -24,8 +24,8 @@ type TenantConfig struct {
 	MaxFailedMountAttempts      int64  `gorm:"default:10"`
 }
 
-func (tc *TenantConfig) BeforeCreate(tx *gorm.DB) error {
-	tc.UID = uuid.NewString()
+func (tenantConfig *TenantConfig) BeforeCreate(_ *gorm.DB) error {
```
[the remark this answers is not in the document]
andriisoldatenko: @wepudt 
>The name of a method's receiver should be a reflection of its identity; often a one or two letter abbreviation of its type suffices (such as "c" or "cl" for "Client"). Don't use generic names such as "me", "this" or "self", identifiers typical of object-oriented languages that gives the method a special meaning. In Go, the receiver of a method is just another parameter and therefore, should be named accordingly. The name need not be as descriptive as that of a method argument, as its role is obvious and serves no documentary purpose. It can be very short as it will appear on almost every line of every method of the type; familiarity admits brevity. Be consistent, too: if you call the receiver "c" in one method, don't call it "cl" in another.

from https://go.dev/wiki/CodeReviewComments#:~:text=The%20name%20of%20a%20method's,the%20method%20a%20special%20meaning.

pkg/controllers/csi/metadata/sqlite_gorm_client.go
andriisoldatenko: For me, one or another is redundant; it's unclear why do we pass `volumeMetaID` 2 times.

wepudt: fixed in #1fd425db51879340cc674fbeb913a9eb5619763b

pkg/controllers/csi/metadata/models.go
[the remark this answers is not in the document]
wepudt: fixed in 34153f4cfc53416af722d0fbf4d55583aa7c24bb

pkg/controllers/csi/metadata/sqlite_gorm_client.go
wepudt: wont change that as it does not work

andriisoldatenko: it works if you refactor using this snippet:
```
func (conn *DBConn) ReadTenantConfigByTenantUUID(ctx context.Context, uid string) (*TenantConfig, error) {
	if uid == "" {
		return nil, errors.New("error")
	} // TODO: check will be obsolete with move to sql.NullString

	tenantConfig := &TenantConfig{TenantUUID: uid}

	result := conn.db.WithContext(ctx).Where(tenantConfig).First(tenantConfig)
	if result.Error != nil {
		return nil, result.Error
	}

	return tenantConfig, nil
}
```

wepudt: fixed in 88cad875c715d861a4a61859035b75333f199764

andriisoldatenko (approved):

PR #2861, opened 2024-03-11
Publish e2e test results on github

aorcholski:
## Description

e2e test results are published on the `E2E tests` workflow page.

image

The main e2e tests can be executed in two ways:
1. to see output on the console:
- test/e2e/standard
- test/e2e/istio
- test/e2e/release
2. to save test results to the `results/` directory:
- test/e2e/standard/publish
- test/e2e/istio/publish
- test/e2e/release/publish

Each `target` creates one file so the following files are created when tests are run by the `e2e-tests` workflow:
- standard.xml
- istio.xml
- release.xml

The workflow then collect data from all files and adds `Test Results` part to the `Summary` page.

## How can this be tested?
Execute e2e tests from github actions.

## Checklist

### DIFF
--- a/.github/scripts/run-e2e-tests.sh
+++ b/.github/scripts/run-e2e-tests.sh
@@ -51,6 +51,6 @@ EOF
 popd
 
 echo "Running tests for environment '$FLC_ENVIRONMENT'..."
-make BRANCH="$TARGET_BRANCH" test/e2e
+make BRANCH="$TARGET_BRANCH" test/e2e-publish
 
 echo "Success!"
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -53,8 +53,7 @@ jobs:
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
     - name: Install gotestsum
-      run: |
-        go install gotest.tools/gotestsum@latest
+      run: go install gotest.tools/gotestsum@latest
     - name: Create cluster
       run: ref/.github/scripts/create-cluster.sh
     - name: Run tests
@@ -65,8 +64,7 @@ jobs:
     - name: Publish test results
       uses: EnricoMi/publish-unit-test-result-action@v2
       with:
-        files: |
-          target/results/*.xml
+        files: target/results/*.xml
       if: always()
   run-in-ocp:
     name: Run in OpenShift latest (${{ github.event.inputs.target || 'main' }})
@@ -106,8 +104,7 @@ jobs:
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
     - name: Install gotestsum
-      run: |
-        go install gotest.tools/gotestsum@latest
+      run: go install gotest.tools/gotestsum@latest
     - name: Create cluster
       run: ref/.github/scripts/create-cluster.sh
     - name: Run tests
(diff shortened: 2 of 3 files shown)


albertogdd:
What happens when a test fails? Does it explicitly say which test failed (printing its name in the summary, etc.), or does it just show the :x: number increased?
I'm asking because I think this is one of the main values of 'publishing' the test results: having a quick overview of what failed.

.github/scripts/run-e2e-tests.sh
```diff
@@ -51,6 +51,6 @@ EOF
 popd
 
 echo "Running tests for environment '$FLC_ENVIRONMENT'..."
-make BRANCH="$TARGET_BRANCH" test/e2e
+make BRANCH="$TARGET_BRANCH" test/e2e-publish
```
albertogdd, line 54: What happens when a test fails? Does it explicitly say which test failed (printing its name in the summary, etc.), or does it just show the ❌ number increased?

I'm asking because I think this is one of the main values of 'publishing' the test results: having a quick overview of what failed.

andriisoldatenko: I think it's okey if we fail a build and publish test results. Not everyone has access to see logs.

aorcholski: I noticed that in case of failing installation (operator pods - ErrPullImage) there is no result file.

.github/workflows/e2e-tests.yaml
```diff
@@ -49,13 +52,22 @@ jobs:
       uses: azure/setup-helm@5119fcb9089d432beecbf79bb2c7915207344b78 # v3.5
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Install gotestsum
+      run: |
+        go install gotest.tools/gotestsum@latest
```
albertogdd, line 57:
```suggestion
      run: go install gotest.tools/gotestsum@latest
```

albertogdd: Same for the rest one line expressions

.github/scripts/run-e2e-tests.sh
albertogdd: Interesting... IMO let's merge it and get experience on how the published result behaves. There will be room for improvement for sure.

.github/workflows/e2e-tests.yaml
```diff
@@ -49,13 +52,22 @@ jobs:
...
     - name: Run tests
       run: ref/.github/scripts/run-e2e-tests.sh
     - name: Destroy cluster
       run: ref/.github/scripts/destroy-cluster.sh
       if: always()
+    - name: Publish test results
+      uses: EnricoMi/publish-unit-test-result-action@v2
+      with:
+        files: |
+          target/results/*.xml
```
albertogdd, line 69:
```suggestion
        files: target/results/*.xml
```

.github/workflows/e2e-tests.yaml
```diff
@@ -93,13 +105,22 @@ jobs:
       uses: azure/setup-helm@5119fcb9089d432beecbf79bb2c7915207344b78 # v3.5
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Install gotestsum
+      run: |
+        go install gotest.tools/gotestsum@latest
```
albertogdd, line 110:
```suggestion
      run: go install gotest.tools/gotestsum@latest
```

albertogdd:
LGTM besides those small details :+1:

albertogdd (dismissed):

PR #2864, opened 2024-03-11
Stop using latest as the default image tag for OA and AG

0sewa0:
## Description

We will always set the `Status.<Component>.ImageID` to a specific version, even in the default case.
If a custom version is provided, we will append the `raw` postfix if it is not present.

Minor change in the behavior of the troubleshooting script: 
- Now we have to rely on the Status, otherwise, we would need to get the latest versions there as well.
- I also removed some tests, as they were totally unnecessary and sometimes even incorrect

## How can this be tested?

When deploying OA or AG components, the version is "always" (except with custom images) in the image tag.

## Checklist

### DIFF
--- a/cmd/support_archive/builder.go
+++ b/cmd/support_archive/builder.go
@@ -16,6 +16,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	clientgocorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
@@ -159,6 +160,11 @@ func (builder CommandBuilder) runCollectors(log logr.Logger, supportArchive arch
 		return err
 	}
 
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(kubeConfig)
+	if err != nil {
+		return err
+	}
+
 	pods := clientSet.CoreV1().Pods(namespaceFlagValue)
 	appName := getAppNameLabel(ctx, pods)
 
@@ -168,7 +174,7 @@ func (builder CommandBuilder) runCollectors(log logr.Logger, supportArchive arch
 	collectors := []collector{
 		newOperatorVersionCollector(log, supportArchive),
 		newLogCollector(ctx, log, supportArchive, pods, appName, collectManagedLogsFlagValue),
-		newK8sObjectCollector(ctx, log, supportArchive, namespaceFlagValue, appName, apiReader, *kubeConfig),
+		newK8sObjectCollector(ctx, log, supportArchive, namespaceFlagValue, appName, apiReader, discoveryClient),
 		newTroubleshootCollector(ctx, log, supportArchive, namespaceFlagValue, apiReader, *kubeConfig),
 		newLoadSimCollector(ctx, log, supportArchive, fileSize, loadsimFilesFlagValue, clientSet.CoreV1().Pods(namespaceFlagValue)),
 	}
--- a/cmd/support_archive/resources.go
+++ b/cmd/support_archive/resources.go
@@ -13,7 +13,6 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
-	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 )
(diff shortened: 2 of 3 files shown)


pkg/util/dtversion/dtversion.go
```diff
@@ -48,3 +50,14 @@ func ToSemver(version string) (string, error) {
 
 	return result, nil
 }
+
+func ToImageTag(version string) string {
```
andriisoldatenko, line 54: @0sewa0 could you please add unit-test to this function?

0sewa0: Question: This function is rather dumb, as it doesn't do any verification, should I add a verification?

The problem with that is that I would need to modify the signature of the function, and the signature of most functions that use it. (shouldn't be that problematic)

andriisoldatenko: you can remove/refactor or change it :) just noticed it doesn't have tests...

cmd/troubleshoot/dynakube_test.go
```diff
@@ -349,10 +358,12 @@ func (builder *testDynaKubeBuilder) withCloudNativeFullStackCustomImage(image st
 func (builder *testDynaKubeBuilder) withHostMonitoringCustomImage(image string) *testDynaKubeBuilder {
 	if builder.dynakube.Spec.OneAgent.HostMonitoring != nil {
 		builder.dynakube.Spec.OneAgent.HostMonitoring.Image = image
+		builder.dynakube.Status.OneAgent.ImageID = image
```
andriisoldatenko, line 361: from what I can see you can do this line out of if/else. cause we do same in both branches. 🤔

cmd/troubleshoot/dynakube_test.go
```diff
@@ -361,13 +372,15 @@ func (builder *testDynaKubeBuilder) withHostMonitoringCustomImage(image string)
...
 		builder.dynakube.Spec.OneAgent.CloudNativeFullStack.CodeModulesImage = image
+		builder.dynakube.Status.CodeModules.ImageID = image
 	} else {
 		builder.dynakube.Spec.OneAgent.CloudNativeFullStack = &dynatracev1beta1.CloudNativeFullStackSpec{
 			AppInjectionSpec: dynatracev1beta1.AppInjectionSpec{
 				InitResources:    &corev1.ResourceRequirements{},
 				CodeModulesImage: image,
 			},
 		}
+		builder.dynakube.Status.CodeModules.ImageID = image
```
andriisoldatenko, line 383: same here.

cmd/troubleshoot/dynakube_test.go
```diff
@@ -335,12 +342,14 @@ func (builder *testDynaKubeBuilder) withClassicFullStackCustomImage(image string
 func (builder *testDynaKubeBuilder) withCloudNativeFullStackCustomImage(image string) *testDynaKubeBuilder {
 	if builder.dynakube.Spec.OneAgent.CloudNativeFullStack != nil {
 		builder.dynakube.Spec.OneAgent.CloudNativeFullStack.Image = image
+		builder.dynakube.Status.OneAgent.ImageID = image
```
andriisoldatenko, line 345: same here

cmd/troubleshoot/dynakube_test.go
```diff
@@ -304,29 +306,34 @@ func (builder *testDynaKubeBuilder) withCloudNativeFullStack() *testDynaKubeBuil
...
 	builder.dynakube.Spec.OneAgent.HostMonitoring = &dynatracev1beta1.HostInjectSpec{}
+	builder.dynakube.Status.OneAgent.ImageID = builder.dynakube.DefaultOneAgentImage(testVersion)
 
 	return builder
 }
 
 func (builder *testDynaKubeBuilder) withClassicFullStackCustomImage(image string) *testDynaKubeBuilder {
 	if builder.dynakube.Spec.OneAgent.ClassicFullStack != nil {
 		builder.dynakube.Spec.OneAgent.ClassicFullStack.Image = image
+		builder.dynakube.Status.OneAgent.ImageID = image
```
andriisoldatenko, line 331: and here :)

config/crd/bases/dynatrace.com_dynakubes.yaml
```diff
@@ -3667,10 +3667,6 @@ spec:
                       performed
                     format: date-time
                     type: string
-                  lastProcessModuleConfigUpdate:
```
andriisoldatenko, line 3670: you need rebase :)

0sewa0: github seems to think its fine :D

andriisoldatenko: image

cmd/troubleshoot/dynakube_test.go
0sewa0: done

cmd/troubleshoot/dynakube_test.go
0sewa0: doen

pkg/util/dtversion/dtversion.go
0sewa0: done

PR #2868, opened 2024-03-12
Make local build script more robust

0sewa0:
## Description

I had problems with the new `build_image.sh`, so I tried to make it nicer according to the following:
- Why isn’t the ${PLATFORM} double quoted? The simplest shellcheck complains about that. 
- PLATFORM  as an env is a bit sketchy, as it’s not very specific, and it may be set for other reasons in certain envs. (like mine 🤦 )
- build_image.sh is only used for development. (if you search for the file name in the codebase, nothing really comes up)
[the CI uses a different logic to build](https://github.com/Dynatrace/dynatrace-operator/blob/34a8044dd7bcf773544d02a69ead2f4883ca1584/.github/actions/build-image/action.yaml#L1-L1)

## How can this be tested?

PLS test it in your environment.

## Checklist


albertogdd (approved):
Works on my machine

aorcholski (approved):
podman on linux OK

andriisoldatenko (approved):
tested on my M1, and it works with/ and without specifying: `OPERATOR_DEV_BUILD_PLATFORM=linux/amd64`

PR #2871, opened 2024-03-12
Add GitHub action to automatically update renovate config

gkrenn:
## Description

Currently we have to manually edit the renovate config file to always point to the last 3 releases.
This PR automates the process and creates a github action that creates a PR that updates the file.

The implementation uses node as the official reference library for JSON5 parsing was implemented in node. 

## How can this be tested?

Create a release branch and check if the PR for `.github/renovate.json5` was created.

## Checklist


hack/build/ci/update-renovate-json5.js
```diff
@@ -0,0 +1,20 @@
...
+// Create version array
+let baseVersions = ["\$default"]
+baseVersions = baseVersions.concat(releaseVersions.map(v => v.replace('origin/', '')));
+
+// Update baseBranches with versions from the version file
+renovateConfig.baseBranches = baseVersions
+renovateConfig.packageRules[0].matchBaseBranches = baseVersions
+
+// Print updated renovate config
+console.log(JSON5.stringify(renovateConfig, null, 2));
```
andriisoldatenko, line 20: instead of printing full config, I would prefer to print:
replacing => old versions into new versions kind of thing.

.github/workflows/renovate.yaml
```diff
@@ -0,0 +1,52 @@
...
+      contents: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Install node
+        uses: actions/setup-node@v4
+      - name: Install json5
+        run: npm install json5
```
0sewa0, line 20: Why not do it in Python? (we already have Python scripts)

https://pypi.org/project/json5/

0sewa0 (changes requested):

.github/workflows/renovate.yaml
albertogdd: +1, not a fan of Javascript

gkrenn: It's done in python now 🥳

hack/build/ci/update-renovate-json5.js
andriisoldatenko: nvm, originally I tought you are printing it like debug info, but then I realized you using pipe and bash to update it.

hack/build/ci/update-renovate-json5.py
```diff
@@ -0,0 +1,24 @@
+import json5
```
andriisoldatenko, line 1: please add json5 to requirements.txt somewhere as dependency.

andriisoldatenko: it;s in readme https://github.com/dynatrace/dynatrace-operator/blob/04b27c3b322371b5ddd7b7ac17449f3ab80514bb/hack/doc/README.md#L16-L23

hack/build/ci/update-renovate-json5.py
```diff
@@ -0,0 +1,24 @@
...
+import sys
+
+# version file contains a list of strings
+versionFile = sys.argv[1]
+renovateFile = sys.argv[2]
+
+# read versions to list
+versions = ['$default']
```
andriisoldatenko, line 9: it's not a blocker, but if you can install `pip install black` and format file, it would be super nice :)

hack/build/ci/update-renovate-json5.py
gkrenn: Does it make sense to add it there as it is only a description for the python scripts in the doc folder.

Maybe we should create a real `requirements.txt` in the hack folder that includes our 3 requirements: 
```
argparse==1.4.0
pyyaml==6.0.1
json5==0.9.24
```

gkrenn: Done 👍

andriisoldatenko (approved):

0sewa0 (approved):

PR #2877, opened 2024-03-13
Ignore terminating namespace during mapping

0sewa0:
## Description

If a Namespace is stuck in Terminating (not because of us, but in general), it will cause errors when creating Secrets for that Namespace. 

These errors are primarily cosmetic because we handle each section separately.
However, they are also easy to fix, and it makes sense just to ignore Terminating namespaces.

## How can this be tested?

- Get a Namespace stuck in Terminating (with a finalizer or something)
- Deploy a DynaKube that should technically monitor the terminating Namespace.
- Check the operator's logs for "forbidden" errors.

## Checklist


andriisoldatenko (approved):

albertogdd (approved):
Oida

PR #2878, opened 2024-03-13
Lower case kind annotation on webhook pod mutation

albertogdd:
## Description

Set the Webhook Pod mutation kind annotation to lower case, according to DT semantic dictionary (https://bitbucket.lab.dynatrace.org/projects/DEUS/repos/semantic-dictionary/browse/source/fields/k8s.yaml)

## How can this be tested?

- Unit tests
- Deploy an app workload and check its pods annotations, should contain the workload kind in lower case

## Checklist

### DIFF
--- a/pkg/webhook/mutation/pod_mutator/dataingest_mutation/mutator.go
+++ b/pkg/webhook/mutation/pod_mutator/dataingest_mutation/mutator.go
@@ -2,6 +2,7 @@ package dataingest_mutation
 
 import (
 	"context"
+	"strings"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/consts"
 	dtingestendpoint "github.com/Dynatrace/dynatrace-operator/pkg/injection/namespace/ingestendpoint"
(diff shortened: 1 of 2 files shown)


pkg/webhook/mutation/pod_mutator/dataingest_mutation/mutator.go
```diff
@@ -124,7 +125,9 @@ func setWorkloadAnnotations(pod *corev1.Pod, workload *workloadInfo) {
 		pod.Annotations = make(map[string]string)
 	}
 
-	pod.Annotations[dtwebhook.AnnotationWorkloadKind] = workload.kind
+	// workload kind annotation in lower case according to dt semantic-dictionary
+	// https://bitbucket.lab.dynatrace.org/projects/DEUS/repos/semantic-dictionary/browse/source/fields/k8s.yaml
```
andriisoldatenko, line 129: it's a private link, better to avoid it to adding here, IMO

andriisoldatenko (approved):

PR #2893, opened 2024-03-18
Enhance EdgeConnect CRD with additional connection properties

andriisoldatenko:
## Description

This PR introduces support for proxy and custom root certificates inside EdgeConnect CRD.

## How can this be tested?

Deploy the operator and apply EdgeConnect CR with new fields. Make sure new fields are projected inside EdgeConnect CR.

## Checklist


pkg/api/v1alpha1/edgeconnect/edgeconnect_types.go
```diff
@@ -28,6 +28,10 @@ type EdgeConnectSpec struct { //nolint:revive
 	// Node selector to control the selection of nodes for the EdgeConnect pods
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 
+	// Set custom proxy settings.
+	// +kubebuilder:validation:Optional
+	Proxy ProxySpec `json:"proxy"`
```
luhi-DT, line 33: omitempty?

andriisoldatenko: fixed in 1492da5b

pkg/api/v1alpha1/edgeconnect/edgeconnect_types.go
```diff
@@ -28,6 +28,10 @@ type EdgeConnectSpec struct { //nolint:revive
 	// Node selector to control the selection of nodes for the EdgeConnect pods
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 
+	// Set custom proxy settings.
```
0sewa0, line 31:
```suggestion
	// General configurations for proxy settings.
```

Just because its a "bigger" section

pkg/api/v1alpha1/edgeconnect/edgeconnect_types.go
```diff
@@ -42,6 +46,10 @@ type EdgeConnectSpec struct { //nolint:revive
 	// Pull secret for your private registry
 	CustomPullSecret string `json:"customPullSecret,omitempty"`
 
+	// Adds custom root certificate from a configmap.
```
0sewa0, line 49:
```suggestion
	// Adds custom root certificate from a configmap. Put the certificate under certs within your configmap.
```

This is probably not decided yet, but it will probably work like that.

pkg/api/v1alpha1/edgeconnect/edgeconnect_types.go
```diff
@@ -91,6 +99,26 @@ type ImageRefSpec struct {
 	Tag string `json:"tag,omitempty"`
 }
 
+type ProxySpec struct {
+	// Proxy scheme.
```
0sewa0, line 103:
```suggestion
	// Proxy scheme (`http` or `https`). 
```

The actual format has probably not been decided yet, but it would be nice to document what is exactly expected.
Would it be possible to have a "special" type alias, that only allows `http` or `https` (does it already exist in some standard lib)? 🤔

pkg/api/v1alpha1/edgeconnect/edgeconnect_types.go
```diff
@@ -91,6 +99,26 @@ type ImageRefSpec struct {
...
+	// Server address (hostname or IP address) of the proxy.
+	Host string `json:"host,omitempty"`
+
+	// NoProxy represents the NO_PROXY or no_proxy environment
+	// variable. It specifies a string that contains comma-separated values
+	// specifying hosts that should be excluded from proxying.
+	NoProxy string `json:"noProxy,omitempty"`
+
+	// Secret name which represents username and password for authentication with the proxy, using the
+	// "Basic" HTTP authentication scheme.
```
0sewa0, line 115:
```suggestion
	// Secret name which contains the username and password used for authentication with the proxy, using the
	// "Basic" HTTP authentication scheme.
```

We could also add which "keys" should be used in the secret. (if we already decided)

0sewa0 (approved):
Added a few suggestions about the docstrings.
- I am approving either way, because most of them depend on implementation that yet to be done

Otherwise LGTM.

andriisoldatenko:
> Added a few suggestions about the docstrings.
> 
> * I am approving either way, because most of them depend on implementation that yet to be done
> 
> Otherwise LGTM.

@0sewa0 PTAL, I added your suggestions :)

0sewa0 (approved):

PR #2907, opened 2024-03-20
Mount trustedCAs configmap to the ActiveGate pod

aorcholski:
## Description

We want ActiveGate to be able to communicate with other HTTPS servers (proxy, dynatrace server) using custom TLS certificates.

`dynakube.spec.trustedCAs` configmap is automatically mounted to the ActiveGate pod. `entrypoint.sh` script of the AG image imports all certificates and configures AG accordingly.

## How can this be tested?

Unittests:
```
make go/test
```

Using a HTTPS proxy with custom certificate:
```
  activeGate:                                                                                                                                                                                                                                                                                                                                                                                 
    capabilities:                                                                                                                                                                                                                                                                                                                                                                             
      - routing                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
  proxy:                                                                                                                                                                                                                                                                                                                                                                                      
      value: https://squid.proxy:3128                                                                                                                                                                                                                                                                                                                                                         
  trustedCAs: proxycert  
```

## Checklist

### DIFF
--- a/pkg/api/v1beta1/dynakube/dynakube_types.go
+++ b/pkg/api/v1beta1/dynakube/dynakube_types.go
@@ -106,7 +106,7 @@ type DynaKubeSpec struct { //nolint:revive
 	Tokens string `json:"tokens,omitempty"`
 
 	// Adds custom RootCAs from a configmap. Put the certificate under certs within your configmap.
-	// Note: Applies only to Dynatrace Operator and OneAgent, not to ActiveGate.
+	// Note: Applies to Dynatrace Operator, OneAgent and ActiveGate.
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Trusted CAs",order=6,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:io.kubernetes:ConfigMap"}
 	TrustedCAs string `json:"trustedCAs,omitempty"`
--- a/pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/ssl_volume.go
+++ b/pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/ssl_volume.go
@@ -1,8 +1,6 @@
 package modifiers
 
 import (
-	"path/filepath"
-
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta1/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/activegate/consts"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/activegate/internal/statefulset/builder"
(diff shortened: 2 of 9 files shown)


pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/sslvolume.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta1/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/activegate/consts"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/activegate/internal/statefulset/builder"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/container"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var _ volumeModifier = SslVolumeModifier{}
```
andriisoldatenko, line 12:
```suggestion
var _ volumeModifier = SSLVolumeModifier{}
```

andriisoldatenko: you should not make camelCase from abbr

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/trustedcasvolume.go
```diff
@@ -0,0 +1,72 @@
+package modifiers
```
andriisoldatenko, line 1: same here

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/ssl_volume.go
```diff
@@ -0,0 +1,57 @@
...
+	return SSLVolumeModifier{
+		dynakube: dynakube,
+	}
+}
+
+type SSLVolumeModifier struct {
+	dynakube dynatracev1beta1.DynaKube
+}
+
+func (mod SSLVolumeModifier) Enabled() bool {
```
andriisoldatenko, line 26: you can use `m` here, it's a convention to use short names here.

> The name of a method's receiver should be a reflection of its identity; often a one or two letter abbreviation of its type suffices (such as "c" or "cl" for "Client"). Don't use generic names such as "me", "this" or "self", identifiers typical of object-oriented languages that gives the method a special meaning. In Go, the receiver of a method is just another parameter and therefore, should be named accordingly. The name need not be as descriptive as that of a method argument, as its role is obvious and serves no documentary purpose. It can be very short as it will appear on almost every line of every method of the type; familiarity admits brevity. Be consistent, too: if you call the receiver "c" in one method, don't call it "cl" in another.

from https://go.dev/wiki/CodeReviewComments#:~:text=The%20name%20of%20a%20method's,the%20method%20a%20special%20meaning.

0sewa0: https://github.com/Dynatrace/dynatrace-operator/blob/623b1a11f86b73bf6cad38221c431ae5c3450ed8/doc/coding-style-guide.md#general

chrismuellner (changes requested):
We should update the comment about on the `trustedCAs` field as well to indicate that it now works with the Activegate: https://github.com/Dynatrace/dynatrace-operator/blob/4d950b09560e49fde84a697b0f026d676bbdbc97/pkg/api/v1beta1/dynakube/dynakube_types.go#L109

0sewa0 (approved):

andriisoldatenko (approved):

pkg/api/v1beta1/dynakube/dynakube_types.go
```diff
@@ -106,7 +106,7 @@ type DynaKubeSpec struct { //nolint:revive
 	Tokens string `json:"tokens,omitempty"`
 
 	// Adds custom RootCAs from a configmap. Put the certificate under certs within your configmap.
-	// Note: Applies only to Dynatrace Operator and OneAgent, not to ActiveGate.
+	// Note: Applies to Dynatrace Operator, OneAgent and ActiveGate.
```
chrismuellner, line 109: we could just remove this whole line instead

chrismuellner (approved):

PR #2915, opened 2024-03-22
Fix missing ConnectionInfo secret in case of no app monitoring

0sewa0:
## Description

It's just a fix for a little logic mess up.

In case of `ClassicFullStack` or `HostMonitroing` the oneagent-connection-info creation was skipped, because the initial "is-needed" check was incorrect. (used `||` instead of `&&`). Added an extra test to make sure this doesn't happen again. 

## How can this be tested?

Deploy a dynakube with `ClassicFullStack` or `HostMonitroing`. 

## Checklist

### DIFF
--- a/pkg/controllers/dynakube/connectioninfo/oneagent/reconciler.go
+++ b/pkg/controllers/dynakube/connectioninfo/oneagent/reconciler.go
@@ -45,7 +45,7 @@ func NewReconciler(clt client.Client, apiReader client.Reader, scheme *runtime.S
 var NoOneAgentCommunicationHostsError = errors.New("no communication hosts for OneAgent are available")
 
 func (r *reconciler) Reconcile(ctx context.Context) error {
-	if !r.dynakube.NeedAppInjection() || !r.dynakube.NeedsOneAgent() {
+	if !r.dynakube.NeedAppInjection() && !r.dynakube.NeedsOneAgent() {
 		meta.RemoveStatusCondition(r.dynakube.Conditions(), oaConnectionInfoConditionType)
 		r.dynakube.Status.OneAgent.ConnectionInfoStatus = dynatracev1beta1.OneAgentConnectionInfoStatus{}
 		// TODO: Delete secret if there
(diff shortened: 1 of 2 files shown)


andriisoldatenko (approved):

albertogdd (approved):

pkg/controllers/dynakube/connectioninfo/oneagent/reconciler.go
```diff
@@ -45,7 +45,7 @@ func NewReconciler(clt client.Client, apiReader client.Reader, scheme *runtime.S
 var NoOneAgentCommunicationHostsError = errors.New("no communication hosts for OneAgent are available")
 
 func (r *reconciler) Reconcile(ctx context.Context) error {
-	if !r.dynakube.NeedAppInjection() || !r.dynakube.NeedsOneAgent() {
+	if !r.dynakube.NeedAppInjection() && !r.dynakube.NeedsOneAgent() {
 		meta.RemoveStatusCondition(r.dynakube.Conditions(), oaConnectionInfoConditionType)
 		r.dynakube.Status.OneAgent.ConnectionInfoStatus = dynatracev1beta1.OneAgentConnectionInfoStatus{}
 		// TODO: Delete secret if there
```
StefanHauth, line 51: Is this a left over or something were we need to do more work?

StefanHauth (changes requested):

pkg/controllers/dynakube/connectioninfo/oneagent/reconciler.go
0sewa0: This TODO was added to be done later, as this is a quick bug fix; I didn't want to change the expected behavior more than necessary.

Context:
There are multiple cases where we do not clean up the secret we create in a reconciler, and I would like to tackle that in 1 go. (firstly, decide if we even care)
On a related note, I already created a follow-up to "standardize" the `kubeobjects` utils. We don't have a utility for cleaning up secrets.(So to do the clean-up here properly, I would need to refactor the Secrets util in kubeobjects, which is not in scope of this change at all.)

aorcholski (approved):

waodim (approved):

StefanHauth (approved):

PR #2917, opened 2024-03-22
Version reconcile only fails if there is no version info in the status

0sewa0:
## Description

We don't want to block a reconciliation if we fail to refresh the version information for a given component.
So we only fail the version reconcile for a component if there is nothing to fall back to; otherwise, we log the error (for specific components that even have a condition `VerificationFailed`) and move on.

## How can this be tested?

The unit test is pretty straightforward, otherwise, you can:

Deploy a Dyankube that will do some "version-reconciling","" check that I didn't break the default behavior.
Then, somehow try to break the few requests we do for version reconciling. (one way would be to lower the latest oneagent in the tenant, so the downgrade check will fail)
And notice that everything is fine because we just fallback to the previous version.

## Checklist

### DIFF
--- a/pkg/controllers/dynakube/version/reconciler.go
+++ b/pkg/controllers/dynakube/version/reconciler.go
@@ -70,7 +70,13 @@ func (r *reconciler) updateVersionStatuses(ctx context.Context, updater StatusUp
 
 	err := r.run(ctx, updater)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


andriisoldatenko (approved):

waodim (approved):
lgtm

PR #2923, opened 2024-03-25
Add conditions for ActiveGate related components to Dynakube status

gkrenn:
## Description

In order to be more expressive what's going on inside the operator, conditions for the ActiveGate are added to the dynakube status.

## How can this be tested?

- Run a dynakube with ActiveGate and check if the conditions are there 

## Checklist


pkg/api/v1beta1/dynakube/dynakube_conditions.go
```diff
@@ -0,0 +1,99 @@
...
+	// ReasonTokenReady is set when a token has passed verifications.
+	ReasonTokenReady string = "TokenReady"
+
+	// ReasonTokenError is set when an unknown error has been found when verifying the token.
+	ReasonTokenError string = "TokenError"
+
+	ReasonCreated         string = "Created"
+	ReasonError           string = "Error"
+	ReasonUnexpectedError string = "UnexpectedError"
+	ReasonUpToDate        string = "UpToDate"
```
0sewa0, line 33: You kinda went against the current pattern 😅 

I would like to avoid bloating the Dynakube package further.

There was this TODO here even related to the TokenScope conditions.
```
// TODO: Move these conditions related consts to a place where they are used, so we don't bloat this package further.
```

Look at https://github.com/Dynatrace/dynatrace-operator/blob/37afc6130005c92ba0ff4fddb503292be48269aa/pkg/util/conditions/error.go#L1-L1 for the common conditions.

pkg/api/v1beta1/dynakube/dynakube_conditions.go
```diff
@@ -0,0 +1,99 @@
...
+
+		return err
+	}
+
+	meta.SetStatusCondition(&dynakube.Status.Conditions, metav1.Condition{
+		Type:    ActiveGateVersionConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  ReasonUpToDate,
+		Message: dynakube.Status.ActiveGate.VersionStatus.ImageID,
+	})
```
0sewa0, line 96: Pretty much a continuation of my previous comment.

There is no reason for bloating the dynakube package with logic that is only relevant in a sub package.

A package-specific condition logic should be defined in that package, [example](https://github.com/Dynatrace/dynatrace-operator/blob/37afc6130005c92ba0ff4fddb503292be48269aa/pkg/controllers/dynakube/connectioninfo/oneagent/conditions.go#L1-L23)

Reasoning: Why would you want to modify an ActiveGateStatefulSet condition in, for example, a OneAgent related package? There is no good reason for that (just bad design maybe 😅 ).
Each package/reconciler/controller should care about only their conditions, not anyone else's.

It only makes sense to store more general/common conditions and condition-related helpers in 1 package, and we already have that. (see my previous comment)

pkg/api/v1beta1/dynakube/dynakube_conditions.go
0sewa0: Also, I would avoid touching the `Token` related conditions as of right now, it is out of scope of your change.

0sewa0 (changes requested):

pkg/api/v1beta1/dynakube/dynakube_conditions.go
gkrenn: Ups, read that we don't want to bloat the file instead of the package. - Will move it to a package more related to activegate, as you described below.

gkrenn: Moved it 👍

pkg/controllers/dynakube/activegate/conditions/conditions.go
```diff
@@ -0,0 +1,70 @@
...
+
+		return
+	}
+
+	meta.SetStatusCondition(conditions, metav1.Condition{
+		Type:   ActiveGateConnectionInfoConditionType,
+		Status: metav1.ConditionTrue,
+		Reason: ReasonCreated,
+	})
+}
```
0sewa0, line 39: could be in the activegate connection info package, and it could be private

pkg/controllers/dynakube/activegate/conditions/conditions.go
```diff
@@ -0,0 +1,70 @@
...
+func SetActiveGateStatefulSetErrorCondition(conditions *[]metav1.Condition, err error) {
+	if err != nil {
+		meta.SetStatusCondition(conditions, metav1.Condition{
+			Type:    ActiveGateStatefulSetConditionType,
+			Status:  metav1.ConditionFalse,
+			Reason:  ReasonError,
+			Message: err.Error(),
+		})
+	}
+}
```
0sewa0, line 50: could be in the activegate statefulset  package, and it could be private

pkg/controllers/dynakube/activegate/conditions/conditions.go
```diff
@@ -0,0 +1,70 @@
...
+		return
+	}
+
+	meta.SetStatusCondition(conditions, metav1.Condition{
+		Type:    ActiveGateVersionConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  ReasonUpToDate,
+		Message: version,
+	})
+}
```
0sewa0, line 70: could be in the version package, and it could be private ([we already have similar ones for OneAgent, so try to reuse what you can](https://github.com/Dynatrace/dynatrace-operator/blob/37afc6130005c92ba0ff4fddb503292be48269aa/pkg/controllers/dynakube/version/conditions.go#L27-L27))

pkg/controllers/dynakube/activegate/conditions/conditions.go
```diff
@@ -0,0 +1,70 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	ReasonCreated         string = "Created"
+	ReasonError           string = "Error"
+	ReasonUnexpectedError string = "UnexpectedError"
+	ReasonUpToDate        string = "UpToDate"
+)
```
0sewa0, line 13: Try to use actual error conditions, stuff like whats in the common conditions package: https://github.com/Dynatrace/dynatrace-operator/blob/37afc6130005c92ba0ff4fddb503292be48269aa/pkg/util/conditions/error.go#L13-L32

pkg/controllers/dynakube/activegate/internal/statefulset/reconciler.go
```diff
@@ -59,10 +63,18 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
...
 		return errors.WithStack(err)
 	}
 
+	meta.SetStatusCondition(&r.dynakube.Status.Conditions, metav1.Condition{
+		Type:    conditions.ActiveGateStatefulSetConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  conditions.ReasonCreated,
+		Message: fmt.Sprintf("StatefulSet for %s has been created", r.capability.DisplayName()),
+	})
+
```
0sewa0, line 77: why not have a specific `set...` function for it, while you have for others 🤔

pkg/controllers/dynakube/activegate/reconciler.go
```diff
@@ -97,24 +99,29 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
...
-		err := r.authTokenReconciler.Reconcile(ctx)
+		err = r.authTokenReconciler.Reconcile(ctx)
 		if err != nil {
 			return errors.WithMessage(err, "could not reconcile Dynatrace ActiveGateAuthToken secrets")
 		}
+	} else {
+		statusconditions := &r.dynakube.Status.Conditions
+		meta.RemoveStatusCondition(statusconditions, conditions.ActiveGateConnectionInfoConditionType)
+		meta.RemoveStatusCondition(statusconditions, conditions.ActiveGateStatefulSetConditionType)
+		meta.RemoveStatusCondition(statusconditions, conditions.ActiveGateVersionConditionType)
```
0sewa0, line 110: each sub reconciler should handle its own conditions, so we can avoid this

gkrenn: Problem is that the sub reconcilers are not called if activegate is removed from the dynakube. Therefore they would not run and cleanup the conditions

pkg/controllers/dynakube/activegate/internal/statefulset/reconciler.go
```diff
@@ -77,18 +78,34 @@ func (r *Reconciler) manageStatefulSet(ctx context.Context) error {
...
-	if deleted || err != nil {
+	if err != nil {
+		conditions.SetKubeApiError(r.dynakube.Conditions(), ActiveGateStatefulSetConditionType, err)
+
 		return errors.WithStack(err)
+	} else if deleted {
+		conditions.SetStatefulSetDeleted(r.dynakube.Conditions(), ActiveGateStatefulSetConditionType, desiredSts.Name)
+
+		return nil
 	}
```
0sewa0, line 100: is this condition needed?
also the `return nil` is probably incorrect

0sewa0 (changes requested):

pkg/util/conditions/suffix.go
```diff
@@ -0,0 +1,26 @@
...
+	return fmt.Sprintf("%s %s", conditionType, updatedSuffix)
+}
+
+func appendDeletedSuffix(conditionType string) string {
+	return fmt.Sprintf("%s %s", conditionType, deletedSuffix)
+}
+
+func appendOutdatedSuffix(conditionType string) string {
+	return fmt.Sprintf("%s %s", conditionType, outdatedSuffix)
+}
```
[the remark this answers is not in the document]
gkrenn: Let's go for `name` then

pkg/controllers/dynakube/activegate/internal/statefulset/reconciler.go
gkrenn: That's at least the logic we [already have](https://github.com/Dynatrace/dynatrace-operator/blob/f3eba0fea8dfa9b10e9059b17e8d828ee55c47c2/pkg/controllers/dynakube/activegate/internal/statefulset/reconciler.go#L85).

gkrenn: Nevertheless, removed it now

gkrenn: Updated to call create afterwards instead of returning

waodim (approved):
I am approving as I think the code LGTM and you have adressed all of @0sewa0 points.

pkg/controllers/dynakube/version/reconciler.go
```diff
@@ -53,13 +54,17 @@ func (r *reconciler) ReconcileOneAgent(ctx context.Context, dynakube *dynatracev
 }
 
 func (r *reconciler) ReconcileActiveGate(ctx context.Context, dynakube *dynatracev1beta1.DynaKube) error {
-	updaters := []StatusUpdater{
-		newActiveGateUpdater(dynakube, r.apiReader, r.dtClient),
+	if !dynakube.NeedsActiveGate() {
+		meta.RemoveStatusCondition(dynakube.Conditions(), activeGateVersionConditionType)
+
+		return nil
 	}
```
0sewa0, line 61: in all other cases we have the this logic in the `updater.Enabled` function

gkrenn: Already doing it there, thanks for pointing out 👍

waodim (approved):

0sewa0 (approved):

PR #2925, opened 2024-03-26
Remove scheme field from EdgeConnect CRD

waodim:
## Description

Since EdgeConnect does not support the scheme field for configuring the proxy we do not need it, so I removed it.

## How can this be tested?

Look at the manifests

## Checklist


andriisoldatenko (changes requested):
Could you please  run `make doc/api-ref` to update docs?

andriisoldatenko (approved):

aorcholski (approved):

PR #2935, opened 2024-03-27
Create edgeconnect.yaml config file and use it inside edgeConnect deployment

andriisoldatenko:
## Description

This PR creates mechanism to configure EdgeConnect via `edgeConnect.yaml` as a secret. Also `edgeConnect.yaml` is mounted to the actual container and is used.

No more Env Vars are used to configure EdgeConnect.

Secrets OAuthClientID & OAuthClient Secret are also passed via `edgeConnect.yaml`.

`caCertsRef` is mounted to a specific path of the container and this path is correctly specified in edgeConnect.yaml

## How can this be tested?

- Create regular EdgeCOnnect custom resource, make sure it's up and running.
- Create provised=true EdgeConnect, make sure it's up and running.

## Checklist


pkg/controllers/edgeconnect/config/config.go
```diff
@@ -0,0 +1,46 @@
+package config
+
+type EdgeConnect struct {
```
0sewa0, line 3: I was kinda confused reading this, then I realised that it should be read like `config.EdgeConnect`, which still reads kinda weird for me, especially if it would be next to a `edgeconnect.EdgeConnect`.

I have no solution to this (other than getting used to it), but thought it is worth mentioning

pkg/controllers/edgeconnect/controller.go
```diff
@@ -331,6 +332,10 @@ func (controller *Controller) reconcileEdgeConnectRegular(edgeConnect *edgeconne
 
 	desiredDeployment.Annotations[hasher.AnnotationHash] = ddHash
 
+	if err := controller.createOrUpdateEdgeConnectConfigSecret(context.Background(), edgeConnect); err != nil {
```
0sewa0, line 335: pass in the `ctx` and use that

pkg/controllers/edgeconnect/deployment/deployment.go
```diff
@@ -156,21 +116,55 @@ func edgeConnectContainer(instance *edgeconnectv1alpha1.EdgeConnect, resource st
...
+		volumes = append(volumes, corev1.Volume{
+			Name: consts.EdgeConnectCustomCAVolumeName,
+			VolumeSource: corev1.VolumeSource{
+				ConfigMap: &corev1.ConfigMapVolumeSource{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: instance.Spec.CaCertsRef,
+					},
+					Items: []corev1.KeyToPath{
+						{
+							Key:  "certs",
```
0sewa0, line 136: "magic" string?
maybe a `const` would be better? 🤔

pkg/controllers/edgeconnect/secret/secret.go
```diff
@@ -0,0 +1,70 @@
...
+	"context"
+
+	edgeconnectv1alpha1 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha1/edgeconnect"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/edgeconnect/config"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/edgeconnect/consts"
+	"gopkg.in/yaml.v3"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func PrepareConfigFile(instance *edgeconnectv1alpha1.EdgeConnect, apiReader client.Reader) ([]byte, error) {
```
0sewa0, line 13: pass in the `ctx` and use it

0sewa0 (changes requested):

pkg/controllers/edgeconnect/secret/secret.go
andriisoldatenko: fixed in bee85f09

pkg/controllers/edgeconnect/controller.go
andriisoldatenko: fixed in bee85f09

pkg/controllers/edgeconnect/deployment/deployment.go
andriisoldatenko: hmmm i think i fixed it

pkg/controllers/edgeconnect/controller.go
```diff
@@ -667,6 +661,32 @@ func (controller *Controller) createOrUpdateEdgeConnectDeployment(ctx context.Co
...
+	configFile, err := secret.PrepareConfigFile(ctx, edgeConnect, controller.apiReader)
+	if err != nil {
+		return err
+	}
+
+	secretData := make(map[string][]byte)
+	secretData[consts.EdgeConnectConfigName] = configFile
+
+	secretConfig, err := k8ssecret.Create(controller.scheme, edgeConnect,
+		k8ssecret.NewNameModifier(consts.EdgeConnectConfigVolumeMountName),
```
0sewa0, line 674: the name of the secret shouldn't be a const
You can have multiple EdgeConnects, so each configuration should have its unique name derived from the name of the EdgeConnect that owns it.

Things to look out for:
- Be careful with the char limit of resource names
- Just "trimming" the name to fit in the limit might lead to conflicts
- AFAIK we have a validation warning for the DynaKube's name's length for the same reason.

pkg/controllers/edgeconnect/deployment/deployment.go
```diff
@@ -156,21 +116,55 @@ func edgeConnectContainer(instance *edgeconnectv1alpha1.EdgeConnect, resource st
...
+			},
+		})
+	}
+
+	return volumes
+}
+
+func prepareVolumeMounts(instance *edgeconnectv1alpha1.EdgeConnect) []corev1.VolumeMount {
+	volumeMounts := []corev1.VolumeMount{
+		{MountPath: consts.EdgeConnectConfigPath, Name: consts.EdgeConnectConfigVolumeMountName},
```
0sewa0, line 150: follow up from my previous review:
this is wrong, this creates a folder called `edgeConnect.yaml`

I used `kubectl debug` to check how it shows up
```
➜  dynatrace-operator git:(configure-edgeconnect-via-edgeconnect-yaml) k debug -n dynatrace provision-test-57585f867f-tx8x9 -it --copy-to=my-debugger --set-image=edge-connect=busybox --container=edge-connect -- sh
If you don't see a command prompt, try pressing enter
~ $ cat edgeConnect.yaml/
cat: read error: Is a directory
~ $ ls edgeConnect.yaml/
edgeConnect.yaml
~ $ cat edgeConnect.yaml/edgeConnect.yaml
name: provision-test
api_endpoint_host: ...
oauth:
    endpoint: ...
    client_id: ...
    client_secret: ...
    resource: ...
```

0sewa0 (changes requested):

pkg/controllers/edgeconnect/deployment/deployment.go
andriisoldatenko: @0sewa0 thanks for troubleshooting, ill fix it.

andriisoldatenko: fixed in d5bba271

pkg/controllers/edgeconnect/controller_test.go
```diff
@@ -55,9 +56,15 @@ func TestReconcile(t *testing.T) {
 			},
 			Spec: edgeconnectv1alpha1.EdgeConnectSpec{
 				ApiServer: "abc12345.dynatrace.com",
+				OAuth: edgeconnectv1alpha1.OAuthSpec{
+					Endpoint:     "https://sso-dev.dynatracelabs.com/sso/oauth2/token",
```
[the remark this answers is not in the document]
andriisoldatenko: fixed 2c9aabbe

pkg/controllers/edgeconnect/controller.go
0sewa0: yes that works but don't forget

> Be careful with the char limit of resource names
Just "trimming" the name to fit in the limit might lead to conflicts
AFAIK we have a validation warning for the DynaKube's name's length for the same reason.

andriisoldatenko: okey let me try out it.

andriisoldatenko: fixed in 095bc819

luhi-DT (changes requested):
There is currently no logic to restart the edgeConnect if there is a change in the CR in order to re-read the edgeConnect.yaml, I would add similar logic like we have with oneAgent or ActiveGate -> hash the value of edgeConnect.yaml and add it as annotation to the deployment

andriisoldatenko:
> There is currently no logic to restart the edgeConnect if there is a change in the CR in order to re-read the edgeConnect.yaml, I would add similar logic like we have with oneAgent or ActiveGate -> hash the value of edgeConnect.yaml and add it as annotation to the deployment

@luhi-DT I"ve added secret hash as annotation of pod.template.spec to force EdgeConnect to restart on secret change.
you can see it cc43e358

pkg/controllers/edgeconnect/consts/consts.go
```diff
@@ -1,24 +1,23 @@
 package consts
 
 const (
-	EdgeConnectUserProvisioned    = "user-provisioned"
-	EdgeConnectContainerName      = "edge-connect"
-	EdgeConnectServiceAccountName = "dynatrace-edgeconnect"
-	EdgeConnectMountPath          = "/etc/edge_connect"
-	EdgeConnectVolumeMountName    = "oauth-secret"
+	EdgeConnectAnnotationSecretHash  = "my-secret-hash"
```
andriisoldatenko, line 4: I'm open to changing it. Should I add the prefix ` "internal.operator.dynatrace.com/" to it?
I just don't want to import it from dynakube package, better to extract to some common constant.

cc @0sewa0

pkg/controllers/edgeconnect/controller.go
```diff
@@ -667,6 +666,32 @@ func (controller *Controller) createOrUpdateEdgeConnectDeployment(ctx context.Co
...
+	query := k8ssecret.NewQuery(ctx, controller.client, controller.apiReader, log)
+
+	err = query.CreateOrUpdate(*secretConfig)
+	if err != nil {
+		log.Info("could not create or update secret for edgeConnect.yaml", "name", secretConfig.Name)
+
+		return "", err
+	}
+
+	return hasher.GenerateHash(secretConfig)
```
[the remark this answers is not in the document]
andriisoldatenko: fixed in b6f2ac5e

andriisoldatenko:
> I tested it an my EdgeConnect pod is failing with:
> 
> ```
> 2024-04-04T08:38:13.497970378+0000 info [edge_connect::edge_connect::runner] Initializing EdgeConnect v1.115.0
> 2024-04-04T08:38:13.498261341+0000 info [initialization] Initializing configuration: Config .yaml file edgeConnect.yaml will be used.
> 2024-04-04T08:38:13.498592433+0000 error [initialization] Error during startup: Error while loading configuration: missing field `name`
> Stream closed EOF for dynatrace/provision-test-57585f867f-tx8x9 (edge-connect)
> ```
> 
> The content of the secret that is mounted: (the `...` parts looked good, just didn't want to leak stuff 😅 )
> 
> ```
> edgeConnect.yaml: |
>   name: provision-test
>   api_endpoint_host: ...
>   oauth:
>       endpoint: ...
>       client_id: ...
>       client_secret: ...
>       resource: ...
> ```

fixed, PTAL

pkg/controllers/edgeconnect/controller.go
```diff
@@ -667,6 +666,32 @@ func (controller *Controller) createOrUpdateEdgeConnectDeployment(ctx context.Co
...
+	query := k8ssecret.NewQuery(ctx, controller.client, controller.apiReader, log)
+
+	err = query.CreateOrUpdate(*secretConfig)
+	if err != nil {
+		log.Info("could not create or update secret for edgeConnect.yaml", "name", secretConfig.Name)
+
+		return "", err
+	}
+
+	return hasher.GenerateHash(secretConfig.Data)
```
luhi-DT, line 696: Why not using the AddAnnotation() from hasher.go instead? This way you can generate the hash and add the annotation at the same time?

pkg/controllers/edgeconnect/consts/consts.go
luhi-DT: If you use AddAnnotation() from hasher.go the annotation name is in line with all the other occurrences, where we do the same pattern

pkg/controllers/edgeconnect/controller.go
andriisoldatenko: because we want to add hash to `desiredDeployment.Spec.Template.Annotations`, but AddAnnotation works only with object itself.

pkg/controllers/edgeconnect/controller.go
```diff
@@ -617,27 +625,18 @@ func (controller *Controller) updateEdgeConnect(ctx context.Context, edgeConnect
...
-	resource, err := k8ssecret.ExtractToken(&secret, consts.KeyEdgeConnectOauthResource)
-	if err != nil {
-		_log.Debug("Unable to extract token from EdgeConnect client secret")
-
-		return err
-	}
+	desiredDeployment := deployment.New(edgeConnect)
 
-	desiredDeployment := deployment.NewProvisioner(edgeConnect, clientSecretName, resource)
+	desiredDeployment.Spec.Template.Annotations = map[string]string{consts.EdgeConnectAnnotationSecretHash: secretHash}
```
0sewa0, line 639: I couldn't find where you calculate the hash for the desired deployment, so maybe I am going to say something dumb. 😅
Shouldn't we have this annotation part of the desired deployment when the hash is generated for it? 🤔

We want to have a hash on the deployment so we don't redeploy the resource in case something (a webhook) dynamically changed it.
We want to have a hash for the secret so if the secret changes, so we redeploy the deployment if the secret changes.
- If we have the secret hash part of the desired deployment at the time of the hash calculation of the deployment, then you only have to check the hash of the deployment to know if a redeployment is needed.

andriisoldatenko: we do it here https://github.com/dynatrace/dynatrace-operator/blob/b6f2ac5ef9dc4214034b12415d9475d36e94e9ef/pkg/controllers/edgeconnect/controller.go#L332-L340, as I already replied to @luhi-DT  in slack, it does't work if I add annotation of secret data to deployment.

I see new secret hash updated, but it doesn't restart pod of deployment.
```
[50](https://dynatrace.slack.com/archives/D05FG9U937D/p1712569856583619)
│                         internal.operator.dynatrace.com/template-hash: 2620308800                                                            │
│                         my-secret-hash: 855148224
```
apply few changes to secret in between:
```
│                         [internal.operator.dynatrace.com/template-hash](http://internal.operator.dynatrace.com/template-hash): 1846043313                                                            │
│                         my-secret-hash: 2079240476
```

pkg/controllers/edgeconnect/consts/consts.go
andriisoldatenko: already replied in another thread that we can't use AddAnnotation() here. I've added similar prefix:
https://github.com/dynatrace/dynatrace-operator/blob/4d096200bc7e61cb5c834a6ea1a9328d2affbbbe/pkg/controllers/edgeconnect/consts/consts.go#L4

and 

https://github.com/dynatrace/dynatrace-operator/blob/4d096200bc7e61cb5c834a6ea1a9328d2affbbbe/pkg/controllers/edgeconnect/consts/consts.go#L5

luhi-DT (approved):

aorcholski (approved):
LGTM

waodim (approved):

0sewa0 (approved):

PR #2938, opened 2024-03-27
Combined certificates from the init secret propagated to the injected agent

aorcholski:
## Description

ActiveGate certificates and trustedCAs certificates should be used by the injected agent on application pods. There is a single file only read by the agent (`custom.pem`) which can be used for that purpose so combined certificates from the init secret are stored the file. 

The following properties have been moved to separate fields:
- data.config.trustedCAs -> data.trustedcas
- data.config.tlsCert -> data.agcerts

These fields are mounted directly to the initContainer of the app pod. 

Based on their content, the `Runner` creates `/mnt/share/custom.pem` and `/mnt/share/custom_proxy.pem` files mounted to the application container. The `/mnt/share/custom_proxy.pem` contains `trustedCAs` certificates only because AG is never an HTTP(s) proxy,

AG certs and TrustedCAs can't be combined into single set of certificates by the Operator because of the Dynatrace Client which is used by the `Runner`. DC would use AG certificate to verify Dynatrace Server if AG certificate was provided without any trustedCAs.

## How can this be tested?

Unittests.

## Checklist

### DIFF
--- a/pkg/injection/startup/files.go
+++ b/pkg/injection/startup/files.go
@@ -96,7 +96,7 @@ func (runner *Runner) createJsonEnrichmentFile() error {
 	)
 	jsonPath := filepath.Join(consts.EnrichmentMountPath, fmt.Sprintf(consts.EnrichmentFilenameTemplate, "json"))
 
-	return runner.createConfFile(jsonPath, jsonContent)
+	return runner.createConfigFile(jsonPath, jsonContent, true)
 }
 
 func (runner *Runner) createPropsEnrichmentFile() error {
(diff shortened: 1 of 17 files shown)


albertogdd (approved):
LGTM

pkg/injection/startup/files.go
```diff
@@ -120,22 +121,42 @@ func (runner *Runner) createCurlOptionsFile() error {
...
 		return errors.WithStack(err)
 	}
 
-	_, err = file.Write([]byte(content))
+	log.Info("created file", "filePath", path)
+
+	return nil
+}
+
+func createConfigFile(fs afero.Fs, path string, content string) error {
```
0sewa0, line 145: Having 3 `createConfigFile` functions that, 2 of them are just wrappers for nicer logs, is a bit hard to read (or it was for me at least 😅 )

could we make it nicer 🤔 
something like:
```go
func (runner *Runner) createConfigFile(path string, content string, verbose bool) error {
	err := createFile(runner.fs, path, content)
	if err != nil {
		return errors.WithStack(err)
	}
	if verbose {
		log.Info("created file", "filePath", path, "content", content)
	} else {
		log.Info("created file", "filePath", path)	
	}

	return nil
}
```
(renamed `createConfigFile` to `createFile`; combined the two `runner.createConfigFile...` to 1)

pkg/webhook/mutation/pod_mutator/oneagent_mutation/config.go
```diff
@@ -22,8 +22,9 @@ const (
 	oneAgentShareVolumeName   = "oneagent-share"
 	injectionConfigVolumeName = "injection-config"
 
-	oneAgentCustomKeysPath = "/var/lib/dynatrace/oneagent/agent/customkeys"
-	customCertFileName     = "custom.pem"
+	oneAgentCustomKeysPath  = "/var/lib/dynatrace/oneagent/agent/customkeys"
+	customCertFileName      = "custom.pem"
+	customProxyCertFileName = "custom_proxy.pem"
```
0sewa0, line 27: can't you use the const you created in [pkg/consts/agent_injection.go](https://github.com/Dynatrace/dynatrace-operator/pull/2938/files#diff-eda55294e0cf89ab2b60121173dcbb8384342001ed988214c5b3d4590f73012eR35-R36) ?

0sewa0 (commented):
LGTM, only have some minor stylistic complaints 😉

pkg/injection/startup/files.go
aorcholski: `verbose` parameter added instead of many functions

0sewa0 (approved):

albertogdd (approved):

waodim (approved):

PR #2942, opened 2024-03-28
Add clean up logic for version and connection-info reconcilers

0sewa0:
## Description

In previous refactors we added a few TODOs about cleaning up after ourselves. 

Here, I added/improved the cleanup for 2 reconcilers + 1:
- version
- connection-info
- +1, kubesystemUUID now is only set once per DynaKube as it should never change, this way we save a request per reconcile

The improvement is not final as:
- ActiveGate conditions are missing, as they are added in https://github.com/Dynatrace/dynatrace-operator/pull/2923
   - So the AG connection info clean-up is not as "clean" as the OA connection-info (only ping the API if there is actually something to clean up)
- Istio reconciler still needs to be updated
- authToken reconciler still needs to be updated (it also will have its own condition, and might make sense to fix the cleanup in that PR)

I think this is a big enough change for now, further improvements will have separate PRs. 
- (easier to review, and if we want a different "style" of doing this, I would prefer to not have to rework everything 😅 ) 

## How can this be tested?

Deploy dynakubes, and turn on/off OA and AG. See that the related conditions and secrets are removed/created.
- only version, and connection-info related stuff should be affected

## Checklist


waodim (approved):
code lgtm and it also works properly for me.

pkg/util/kubeobjects/secret/secret.go
```diff
@@ -48,6 +48,18 @@ func (query Query) Update(secret corev1.Secret) error {
...
 }
 
+func (query Query) Delete(name, namespace string) error {
+	query.Log.Info("updating secret", "name", name, "namespace", namespace)
+	tmp := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace}}
+
+	err := query.KubeClient.Delete(query.Ctx, tmp)
+	if k8serrors.IsNotFound(err) {
+		return nil
+	}
```
albertogdd, line 58: Maybe add a log here?

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -72,19 +72,19 @@ func NewReconciler(
 }
 
 func (r *reconciler) Reconcile(ctx context.Context) error {
-	err := r.versionReconciler.ReconcileCodeModules(ctx, r.dynakube) // Handles their cleanup on their own, like all things should
```
albertogdd, line 75: The comment :rofl:

albertogdd (approved):
Overall LGTM

pkg/util/kubeobjects/secret/secret.go
albertogdd: What about this previous log then
```
query.Log.Info("updating secret", "name", name, "namespace", namespace)
```

0sewa0: well that is nice catch 😅
on it 😉

waodim (approved):

albertogdd (approved):

PR #2947, opened 2024-04-02
Fix release notes generation script

0sewa0:
## Description

Escape the `\` in the helm command.
Also remove some old leftovers.

## How can this be tested?
Running ` ./hack/build/ci/generate-release-notes.sh` locally should give you an idea how it looks.

## Checklist

### DIFF
--- a/hack/build/ci/generate-release-notes.sh
+++ b/hack/build/ci/generate-release-notes.sh
@@ -27,11 +27,6 @@ kubernetes_manifests="kubectl apply -f https://github.com/Dynatrace/dynatrace-op
 openshift_manifests="oc apply -f https://github.com/Dynatrace/dynatrace-operator/releases/download/${tag}/openshift.yaml"
 
 if [ "${pre_release}" = false ] ; then
-  kubernetes_manifests="${kubernetes_manifests}
-kubectl apply -f https://github.com/Dynatrace/dynatrace-operator/releases/download/${tag}/kubernetes-csi.yaml"
-  openshift_manifests="${openshift_manifests}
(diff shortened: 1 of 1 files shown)


hack/build/ci/generate-release-notes.sh
```diff
@@ -44,10 +39,10 @@ For information on how to install the [latest dynatrace-operator](https://github
 
 #### Helm (recommended)
 \`\`\`sh
-helm upgrade dynatrace-operator oci://public.ecr.aws/dynatrace/dynatrace-operator \
-  --version ${tag_without_leading_v} \
-  --create-namespace --namespace dynatrace \
-  --install \
+helm upgrade dynatrace-operator oci://public.ecr.aws/dynatrace/dynatrace-operator \\
+  --version ${tag_without_leading_v} \\
```
waodim, line 43: when i run the script at this place I get this output:


```
helm upgrade dynatrace-operator oci://public.ecr.aws/dynatrace/dynatrace-operator \
  --version tag> \
  --create-namespace --namespace dynatrace \
  --install \
  --atomic
```

```
  --version tag> \
```
this line looks weird. Can you confirm or does this only happen on my machine?

0sewa0: yes, that is kinda ugly, but happens for me aswell, but it happens irrelevant of my change

its because of `tag=${GITHUB_REF_NAME:-<tag>}`, where `<tag>` is just a place holder, it will always be `GITHUB_REF_NAME` when its actually run

(so I wouldn't worry about it 😅 )

waodim (approved):

andriisoldatenko (approved):

PR #2951, opened 2024-04-03
Add ipv6 support to ActiveGate deployment

0sewa0:
## Description

We filled out the `DT_DNS_ENTRY_POINT` envvar based on the IP of the created `Service` to the ActiveGate `StatefulSet` using the envvars that are automatically added by kubernetes. (see in docs https://kubernetes.io/docs/concepts/services-networking/service/)

However, this is problematic when we consider that:
- the IPs of the `Service` can be IPv4 or IPv6, and the DNS entry points are actual URLs (example: `https://10.123.123.12:443/communication`).
- the contents "auto-added" Service envs are not formatted according to _RFC2732_ in the case of IPv6 (makes sense).

Due to this "limitation", we can't rely on the "auto-added" envs, instead, we look at the `Service` we create and get the IPs from it, then add each to the `DT_DNS_ENTRY_POINT` directly, formatted correctly.
- This way we can support even dual-stack setups.

Sidenote: Ignore the first commit, that was my easy/quick POC without magic.

## How can this be tested?

This is semi-difficult, because you need a dual-stack cluster (can be created on GKE) + do changes to the code, that enforces the usage of `DualStack` IPFamily, because otherwise it will always use IPv4. (if you can configure your cluster to not have to do this, then you deserve all the 🌮 )

I did this, so you don't necessarily have to:
- The service
image
- The resulting entry points
image

Otherwise, the unittests are straight forward.

## Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -302,6 +302,8 @@ service:
   golangci-lint-version: 1.55.x # use the fixed version to not introduce new linters unexpectedly
 
 issues:
+  exclude-dirs:
+    - pkg/api/v1alpha1/dynakube # legacy version, should not be changed
   exclude-rules:
     # Exclude duplicate code and function length and complexity checking in test
     # files (due to common repeats and long functions in test code)
@@ -337,7 +339,3 @@ issues:
     - path-except: 'pkg/api/(.+)\.go'
       linters:
         - godot
-
-run:
-  skip-dirs:
-    - pkg/api/v1alpha1/dynakube # legacy version, should not be changed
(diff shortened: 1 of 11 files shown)


andriisoldatenko:
@0sewa0 could you please run `make doc/api-ref`?

0sewa0:
> @0sewa0 could you please run `make doc/api-ref`?

did it, no changes (the first commit would need it, but that is no longer relevant)

andriisoldatenko (approved):

waodim (approved):

PR #2967, opened 2024-04-05
Rename "data-ingest" to "metadata-enrichment"

0sewa0:
## Description

Renames (almost) every occurrence of `data-ingest` to `metadata-enrichment` or something that is more accurate.
- The only occurrences I left alone, for now, are the `DataIngestToken` and related code, as that would be a breaking change, and we can handle that separately.

Also tried to rename related packages to remain consistent. (webhook packages)
 
## How can this be tested?

No changes in logic, just look at the renames and see if they make sense.

## Checklist


waodim (approved):

aorcholski (approved):
LGTM

waodim (approved):

PR #2968, opened 2024-04-05
Allow duplicate oneagent arguments

StefanHauth:
## Description

With a previous overhaul of how OneAgent arguments are handled it was made impossible to use an argument multiple times. For some arguments that's reasonable, for others not so much.

This PR enables using the same argument multiple times for the OneAgent, while keeping some more sensible arguments to being used only once. Only those where the operator potentially uses default values are considered. Namely those which can't be used multiple times are:
* `--set-server`
* `--set-tenant`
* `--set-host-id-source`
* `--set-host-group`
* `--set-network-zone`
* `--set-proxy`

## How can this be tested?

Deploy the operator.
Deploy a DynaKube where you set custom one agent arguments. Use especially multiple `--set-host-property` and `--set-host-tag` entries. They should be usable multiple times.

## Checklist


chrismuellner (changes requested):
Please change the logic as discussed so all arguments are allowed.

We should add a warning when an argument is used that is already configured by the Operator, not block the usage.

StefanHauth:
> Please change the logic as discussed so all arguments are allowed.
> 
> We should add a warning when an argument is used that is already configured by the Operator, not block the usage.

All arguments are allowed, It's just made sure that those, where the operator sets default values, are not duplicated if they are also provided in the Dynakube. In such cases, the operator's default values are overruled by the Dynakube (which always shall take precedence).

chrismuellner:
@StefanHauth I don't see the point of doing that. Let's just leave that logic to the Oneagent.

We allow all values. User provided args are added after operator ones and therefore take priority. If a user provided value is also set by the Operator we only provide a warning and don't remove any existing arg from the list.

aorcholski (approved):
LGTM

pkg/controllers/dynakube/oneagent/daemonset/arguments.go
```diff
@@ -13,7 +13,11 @@ const customArgumentPriority = 2
 const defaultArgumentPriority = 1
 
 func (dsInfo *builderInfo) arguments() ([]string, error) {
-	argMap := prioritymap.New(prioritymap.WithSeparator(prioritymap.DefaultSeparator), prioritymap.WithPriority(defaultArgumentPriority))
+	argMap := prioritymap.New(
```
luhi-DT, line 16: Why do we need a priorityMap at all? Wouldn't it be simpler to just append the args as they are? And not store them in proprietary struct

luhi-DT (changes requested):

pkg/controllers/dynakube/oneagent/daemonset/arguments.go
StefanHauth: Maybe, but not in this PR.

StefanHauth: The priority map ensures proper ordering of the arguments, no matter in which order they are added. IMHO this also provides more resilience against inconsiderate changes in the future.

luhi-DT: I am fine with not having it in this bugfix PR, but we should definitely get rid of this logic, its to complicated for what it brings us in terms of benefits

luhi-DT (approved):

chrismuellner (approved):
PR description is outdated, but changes were done as expected

PR #2977, opened 2024-04-08
Add ServiceAcountName and kubernetesApiAutomation fields to EdgeConnect CRD

aorcholski:
## Description

Adds `serviceAccountName` field and `kubernetesApiAutomation` section to the EdgeConnect CRD. 

if no service account is configured, `dynatrace-edgeconnect` is used as the default value in CRD. Validation webhook checks if provided service name is not empty.

`assertAllowedResponse` function is used by all EC unittest  modules so `assert.True` call simply prints error(s) returned by all `validator` functions in case of assertion failure.  An error may be unrelated to `ApiServer` value.

## How can this be tested?

Unittests.

## Checklist


pkg/api/v1alpha1/edgeconnect/edgeconnect_types.go
```diff
@@ -116,6 +123,11 @@ type ProxySpec struct {
 	Port uint32 `json:"port,omitempty"`
 }
 
+type KubernetesApiAutomationSpec struct {
+	// Enabled enables kubernetes automation
```
0sewa0, line 127:
```suggestion
	// Enables kubernetes automation
```

do we have to say the name of the field 😅

aorcholski: fixed

andriisoldatenko:
could you please run:
```
make doc/api-ref
```

and commit files.

andriisoldatenko (changes requested):
could you please run `make doc/api-ref` ?

waodim (approved):
lgtm

pkg/webhook/validation/edgeconnect/service_account.go
```diff
@@ -0,0 +1,19 @@
...
+	errorInvalidServiceName = `The EdgeConnect's specification has an invalid serviceAccountName.
+`
+)
+
+func isInvalidServiceName(_ context.Context, _ *edgeconnectValidator, edgeConnectCR *edgeconnect.EdgeConnect) string {
+	if edgeConnectCR.Spec.ServiceAccountName == "" {
+		return errorInvalidServiceName
+	}
+
+	return ""
```
0sewa0, line 18: maybe dumb questions:
- Do you always need a `ServiceAccount` or only when `KubernetesApiAutomation` is used?

If so:
- We shouldn't always error for an empty `ServiceAccount`
- Why isn't the `ServiceAccountName` under the `KubernetesApiAutomation` section, will it be used for something else in the future?

aorcholski: fixed [commit](https://github.com/Dynatrace/dynatrace-operator/pull/2977/commits/eb8a00a44ca0b1f7830ee270730082997d74f2aa)

0sewa0: but then you have to have a `*string` for it in the CR, no?

because you want to differentiate between:
1. "not  set" == `nil`
2. "set to nothing" == `""`
3. "set to something" == `myServiceAccount`

A normal `string` can only have 2 states, so it wont work for:
1. "not  set" == `""` ⚠️ 
2. "set to nothing" == `""` ⚠️ 
3. "set to something" == `myServiceAccount`

aorcholski: `serviceAccountName` is NOT optional. We don't need a pointer and `nil` value. It always has a value - either the default one or provided by an user. The problem is one can use empty string and apply such a CR because it is correct manifest. However from our/k8s point of view `""` value has no meaning so has to be immediately rejected.

0sewa0 (approved):
LGTM (only merge after 1.1.0 release branch was created)

waodim (approved):

config/crd/bases/dynatrace.com_edgeconnects.yaml
```diff
@@ -207,6 +207,13 @@ spec:
                     description: Indicates version of the EdgeConnect image to use
                     type: string
                 type: object
+              kubernetesApiAutomation:
+                description: KubernetesApiAutomation enables Automation workflows
+                properties:
+                  enabled:
+                    description: Enables kubernetes automation
```
StefanHauth, line 214:
```suggestion
                    description: Enables Kubernetes automation
```

config/crd/bases/dynatrace.com_edgeconnects.yaml
```diff
@@ -325,6 +332,11 @@ spec:
                       More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
                     type: object
                 type: object
+              serviceAccountName:
+                default: dynatrace-edgeconnect
+                description: ServiceAccountName allows EdgeConnect to perform required
```
StefanHauth, line 337:
```suggestion
                description: ServiceAccountName that allows EdgeConnect to access the Kubernetes API
```

config/crd/bases/dynatrace.com_edgeconnects.yaml
```diff
@@ -325,6 +332,11 @@ spec:
...
+                  operations inside the Kubernetes cluster
```
StefanHauth, line 338:
```suggestion

```

pkg/api/v1alpha1/edgeconnect/edgeconnect_types.go
```diff
@@ -50,6 +53,10 @@ type EdgeConnectSpec struct { //nolint:revive
 	// +kubebuilder:validation:Optional
 	CaCertsRef string `json:"caCertsRef,omitempty"`
 
+	// ServiceAccountName allows EdgeConnect to perform required operations inside the Kubernetes cluster
```
StefanHauth, line 56:
```suggestion
	// ServiceAccountName allows EdgeConnect to access the Kubernetes API
```

StefanHauth (changes requested):

agrimmer:
@amohrd, Can you please confirm that we want to use `kubernetesApiAutomation` in the CRD?
The corresponding feature in Dynatrace is called "Kubernetes Automation for Workflows" without the "API".

config/crd/bases/dynatrace.com_edgeconnects.yaml
```diff
@@ -207,6 +207,13 @@ spec:
                     description: Indicates version of the EdgeConnect image to use
                     type: string
                 type: object
+              kubernetesApiAutomation:
```
StefanHauth, line 210: There was a small change requested, please change it to
```suggestion
              kubernetesAutomation:
```

StefanHauth (changes requested):

amohrd:
> @amohrd, Can you please confirm that we want to use `kubernetesApiAutomation` in the CRD? The corresponding feature in Dynatrace is called "Kubernetes Automation for Workflows" without the "API".

@agrimmer 

kubernetesAPIAutomation: I see it as a quite technical description, as the workflow actions use the API for command execution. kubernetesAutomation: I understand it as a more generic term, and better fits to "Kubernetes Automation for Workflows". Also the whole CRD already include various "API",  therefore I prefer **kubernetesAutomation**

config/crd/bases/dynatrace.com_edgeconnects.yaml
aorcholski: [Renamed](https://github.com/Dynatrace/dynatrace-operator/pull/2977/commits/e2669335c066680be16daa0681c9f34f6563395b)

StefanHauth (approved):

andriisoldatenko (approved):

PR #2991, opened 2024-04-09
Add short names for dynakube

waodim:
## Description
With this the short names `dk` and `dks` are added to the dynakube CRD. They can then be used with the `kubectl get` command like for example: `kubectl get ds` for daemonset.

## How can this be tested?
Deploy operator with this CRD and try out these commands:
- `kubectl get dk`
- `kubectl get dks`

## Checklist

### DIFF
--- a/config/crd/bases/dynatrace.com_dynakubes.yaml
+++ b/config/crd/bases/dynatrace.com_dynakubes.yaml
@@ -13,6 +13,9 @@ spec:
     kind: DynaKube
     listKind: DynaKubeList
     plural: dynakubes
+    shortNames:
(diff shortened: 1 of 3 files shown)


luhi-DT (approved):

0sewa0 (approved):
LGTM, should we also add these for EdgeConnect, to be consistent? 🤔 (`ec/ecs`)

luhi-DT:
> LGTM, should we also add these for EdgeConnect, to be consistent? 🤔 (`ec/ecs`)

@waodim can you please create a ticket for that? :)

PR #3004, opened 2024-04-15
Add prerequisites installation to all python commands

gkrenn:
## Description

[Issue Link](https://dt-rnd.atlassian.net/browse/K8S-9236)

Prerequisites should be automatically installed when a command that requires python is used

## How can this be tested?

Run `make doc/api-ref` and check if it is working


waodim (approved):
lgtm

luhi-DT (approved):

PR #3007, opened 2024-04-15
Fix container-name and container-image env var counter if container is excluded via annotation

luhi-DT:
## Description

[Issue](https://dt-rnd.atlassian.net/browse/K8S-9700)

Currently the initContainer of an injected Pod fails, if there is a second container in the list of containers, which is excluded and in-front of the monitored container, in the list of containers.

## How can this be tested?

Deploy Operator -> cloudNativeFullStack -> create a sample pod with 2 containers -> exclude the first container in the list of containers from monitoring via annotation on Dynakube `container.inject.dynatrace.com/<container name to exclude>: "false"` (important, it has to be the first one)


waodim (approved):
tested it and it works :+1:

gkrenn (approved):

PR #3059, opened 2024-04-24
EdgeConnect pod spec enhancements for Kubernetes Automation (ServiceAccountName, HostAlias)

aorcholski:
## Description

Sets the following fields of EdgeConnect pod spec :
- `serviceAccount`, `serviceAccountName` to _CR.Spec.ServiceAccount_
- `hostAliases[0].IP` to _ClusterIP_ of the _kubernetes.default_ service
- `hostAliases[0].Hostnames[0]` to _CR-name.CR-namespace.UID-of-the-kube-system-namespace_

Both fields are always set  (also if K8S Automation is disabled).

Operator needs new clusterrole to get `kubernetes.default` service.

## How can this be tested?

- Unittests
- Deploy EdgeConnect in normal mode and provisioner mode and verify spec of the EdgeConnect pod
- Deploy EdgeConnect with different ServiceAccount
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: test-normal-edgeconnect
  namespace: dynatrace
  labels:
    app.kubernetes.io/name: edgeconnect
    app.kubernetes.io/version: "0.0.0-snapshot"
    app.kubernetes.io/component: user-provisioned
```


pkg/controllers/edgeconnect/controller.go
```diff
@@ -695,3 +710,30 @@ func (controller *Controller) createOrUpdateEdgeConnectConfigSecret(ctx context.
...
+	var kubernetesService corev1.Service
+
+	err := controller.apiReader.Get(ctx, client.ObjectKey{Namespace: defaultNamespaceName, Name: kubernetesServiceName}, &kubernetesService)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	var kubeSystemNamespace corev1.Namespace
+
+	err = controller.apiReader.Get(ctx, client.ObjectKey{Namespace: "", Name: kubeSystemNamespaceName}, &kubeSystemNamespace)
```
andriisoldatenko, line 724:
```suggestion
	err = controller.apiReader.Get(ctx, client.ObjectKey{Name: kubeSystemNamespaceName}, &kubeSystemNamespace)
```

pkg/controllers/edgeconnect/controller.go
```diff
@@ -319,6 +324,11 @@ func (controller *Controller) reconcileEdgeConnectRegular(ctx context.Context, e
...
 	_log := log.WithValues("namespace", edgeConnect.Namespace, "name", edgeConnect.Name, "deploymentName", desiredDeployment.Name)
 
+	err := controller.hostAlias(ctx, desiredDeployment, edgeConnect.Name, edgeConnect.Namespace)
+	if err != nil {
+		return err
+	}
+
 	if err := controllerutil.SetControllerReference(edgeConnect, desiredDeployment, controller.scheme); err != nil {
 		return errors.WithStack(err)
 	}
```
0sewa0, line 334: Couldn't the `createOrUpdateEdgeConnectDeployment` function also be used here?

`createOrUpdateEdgeConnectDeployment` is only used by `reconcileEdgeConnectProvisioner` currently, but the naming of the function and the contents of it seems generic enough so we might be able to use it here aswell.

andriisoldatenko: good point!

andriisoldatenko: @0sewa0 wdyt if we merge as is, and I'll create a new ticket to refactor it separately?

andriisoldatenko (approved):
practically LGTM, tested in my cluster.

0sewa0 (approved):

PR #3120, opened 2024-05-06
Fix ignore-proxy feature flag for oneAgent daemonset

luhi-DT:
## Description

[Ticket](https://dt-rnd.atlassian.net/browse/K8S-9893)

Currently the oneagent-ignore-proxy feature flag is not considered, when the new way to mount the proxy to the agent via volume is used.

## How can this be tested?

Create a dynakube with a proxy and the annotation oneagent-ignore-proxy, check if the daemonset of the agent does not have a proxy volume and volume mount

### DIFF
--- a/pkg/controllers/dynakube/oneagent/daemonset/volumes_test.go
+++ b/pkg/controllers/dynakube/oneagent/daemonset/volumes_test.go
@@ -227,7 +227,7 @@ func TestPrepareVolumeMounts(t *testing.T) {
 		assert.Contains(t, volumeMounts, getClusterCaCertVolumeMount())
 		assert.Contains(t, volumeMounts, getCSIStorageMount())
 	})
-	t.Run(`has no volume if no proxy is set`, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/controllers/dynakube/oneagent/daemonset/volumes_test.go
```diff
@@ -227,4 +227,28 @@ func TestPrepareVolumeMounts(t *testing.T) {
 		assert.Contains(t, volumeMounts, getClusterCaCertVolumeMount())
 		assert.Contains(t, volumeMounts, getCSIStorageMount())
 	})
+	t.Run(`has no volume if no proxy is set`, func(t *testing.T) {
```
0sewa0, line 230:
```suggestion
	t.Run(`has no volume if proxy is set and proxy ignore feature-flags is used`, func(t *testing.T) {
```

0sewa0 (commented):
LGTM, just minor nitpick for test naming

0sewa0 (approved):

PR #3124, opened 2024-05-06
Use explicit tenant UUID returned by tenant

gkrenn:
## Description

Currently the tenant UUID is derived from the URL provided in the dynakube.
It is never guaranteed that the url contains a valid UUID and can even be an alias. 

This PR updates the corresponding code to use the UUID returned by the tenant itself.

[Jira Ticket](https://dt-rnd.atlassian.net/browse/K8S-9537)

## How can this be tested?

Run cloudNative e2e test and check if everything works as expected. 
Run resilience test to verify that even with limited connectivity the CSI driver works as expected.


pkg/api/v1beta1/dynakube/properties.go
```diff
@@ -414,9 +414,14 @@ func (dk *DynaKube) Tokens() string {
 	return dk.Name
 }
 
-// TenantUUIDFromApiUrl gets the tenantUUID from the ApiUrl present in the struct, if the tenant is aliased then the alias will be returned.
-func (dk *DynaKube) TenantUUIDFromApiUrl() (string, error) {
-	return tenantUUID(dk.Spec.APIURL)
+func (dk *DynaKube) TenantUUIDFromConnectionInfo() (string, error) {
```
0sewa0, line 417:
```suggestion
func (dk *DynaKube) TenantUUID() (string, error) {
```

as there is no other `TennantUUIDFromXYZ` function, just having it with a shorter name would be better

pkg/controllers/dynakube/controller.go
```diff
@@ -297,23 +301,15 @@ func (controller *Controller) setupTokensAndClient(ctx context.Context, dynakube
...
-
 	return dynatraceClient, nil
 }
 
 func (controller *Controller) reconcileComponents(ctx context.Context, dynatraceClient dtclient.Client, istioClient *istio.Client, dynakube *dynatracev1beta1.DynaKube) error {
 	var componentErrors []error
 
+	pullSecretReconciler := dtpullsecret.
+		NewReconciler(controller.client, controller.apiReader, dynakube, controller.tokens)
+
```
0sewa0, line 312: why is this change necessary for the `using the tennantUUID from the connection-info` logic?

Also, it looks kinda wrong (sorry if its a bit rambly 😅 )
- when do you **not** need a pull-secret? => when its application-monitoring
   - for both oneagent and activegate it is needed
   - so why move it into the `injection` reconciler?
   - why move it at all in the context of the intended change?
 - why store a "status" in memory that will be released right after each reconcile?
   - The current way we do it for the`connectionInfo` is that the `injection` and `oneagent` Reconcilers don't each get a `connection-info-reconciler` but create one on there own, without coupling everything together.
      - The "is already reconciled" logic is done using the status and the timestamp within it, so its multipurpose
         - Only gets updated every X minutes
         - Only get updated maximum once per reconcile as the timestamp already handles that

0sewa0 (changes requested):

pkg/controllers/dynakube/controller.go
gkrenn: You are right, the creation does not need to be in the injection reconciler - Added it in the injection reconciler, as I was not sure if the oneagent reconciler will always run before the injection one. But guess this won't change that likely in the future. Will remove it from there. 

The Status is way more elegant did not think about it, thx for pointing out 👍

pkg/controllers/dynakube/oneagent/oneagent_reconciler.go
```diff
@@ -107,6 +115,15 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
...
 
+	log.Info("creating Pull Secret")
+
+	err = dtpullsecret.NewReconciler(r.client, r.apiReader, r.dynakube, r.tokens).Reconcile(ctx)
+	if err != nil {
+		log.Error(err, "could not reconcile pull secret")
+
+		return err
+	}
+
```
0sewa0, line 126: what if you don't want to deploy OneAgents but only an ActiveGate ? 🙃

gkrenn: That's indeed a problem, still have to add the status condition to the pull secret reconciler and write unit tests for it - And then call it a second time. 

Will do that after my vacation, as I don't want to code in another bug in a hurry now.

gkrenn:
Updated the pull secret reconciler and added conditions + cleanup logic

Pull Secret Reconciler is not called from within oneagent reconciler and active gate reconciler

pkg/controllers/csi/provisioner/controller.go
```diff
@@ -161,7 +161,7 @@ func (provisioner *OneAgentProvisioner) Reconcile(ctx context.Context, request r
 }
 
 func (provisioner *OneAgentProvisioner) setupFileSystem(dk *dynatracev1beta2.DynaKube) error {
-	tenantUUID, err := dk.TenantUUIDFromApiUrl()
+	tenantUUID, err := dk.TenantUUID()
```
0sewa0, line 164: this will break the "resilience" feature, because
```
	err = provisioner.setupFileSystem(dk) // <= BOOM, because no tenant-uuidm because network is dead
	if err != nil {
	}
	tenantConfig, err := provisioner.setupTenantConfig(dk) // needed for the CSI-resilience feature
	if err != nil {
	}
```

0sewa0 (changes requested):

pkg/controllers/csi/provisioner/controller.go
gkrenn: Set it back to the old method + follow up ticket created

pkg/controllers/dynakube/dtpullsecret/reconciler.go
```diff
@@ -18,27 +21,49 @@ const (
...
+	}
+
+	if !(r.dynakube.NeedsOneAgent() || r.dynakube.NeedsActiveGate()) {
+		if meta.FindStatusCondition(*r.dynakube.Conditions(), PullSecretConditionType) == nil {
+			return nil // no condition == nothing is there to clean up
+		}
+
+		err := r.deletePullSecretIfExists(ctx)
+		if err != nil {
+			log.Error(err, "failed to clean-up OneAgent tenant-secret")
```
0sewa0, line 53: i think this log line is not correct 😅 

also if you bubble up the error then don't log it, as it will be logged higher up, and we don't want duplicate error logs

pkg/controllers/dynakube/oneagent/oneagent_reconciler.go
```diff
@@ -107,6 +115,15 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 		return r.cleanUp(ctx)
 	}
 
+	log.Info("creating Pull Secret")
```
0sewa0, line 118: do we need this log line here?
(you don't have it in the AG reconciler for the same logic)

pkg/controllers/dynakube/oneagent/oneagent_reconciler.go
```diff
@@ -107,6 +115,15 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
...
+
+	err = dtpullsecret.NewReconciler(r.client, r.apiReader, r.dynakube, r.tokens).Reconcile(ctx)
+	if err != nil {
+		log.Error(err, "could not reconcile pull secret")
```
0sewa0, line 122: if you bubble up the error then don't log it, as it will be logged higher up, and we don't want duplicate error logs

0sewa0 (changes requested):
Works as expected, but there are some inconsistencies in logging that we should fix

pkg/controllers/csi/provisioner/controller.go
```diff
@@ -287,7 +287,7 @@ func (provisioner *OneAgentProvisioner) updateAgentInstallation(
 }
 
 func (provisioner *OneAgentProvisioner) handleMetadata(dk *dynatracev1beta2.DynaKube) (*metadata.TenantConfig, error) {
-	tenantUUID, err := dk.TenantUUIDFromApiUrl()
+	tenantUUID, err := dk.TenantUUIDFromApiUrl() // TODO update to use the tenant uuid from the DynaKube status
```
waodim, line 290: why isnt this updated in this PR?

pkg/controllers/dynakube/dtpullsecret/reconciler.go
gkrenn: We should just log the error and ignore it in the cleanup, as you described in the active gate connection reconciler. 

Will update wrong log line and remove the return so that the conditions are cleanup up as well like in the active gate connection info reconciler.

0sewa0 (approved):

luhi-DT (approved):

waodim (approved):
LGTM

PR #3136, opened 2024-05-07
Fix GORM and remove usage of digest

0sewa0:
## Description

Initial change:
We no longer need to use the digest to pull the image in case of `CodeModulesImage`, so now we just use the ref we get from the user, and use that string (base64 encoded) as the folder name.

**To make the change actually work:**
I had to update the way we do things all around the new GORM implementation, so that we are utilising the data in the new database schema in a way that "make more sense".
- This includes changes to the GC, now it actually runs when there is more likely some stuff to collect (example: when a dynakube is deleted)
- It actually cleans things up (⚠️ this is very important)
  - The `OSMounts` have to be cleaned up, otherwise the `TenantConfigs` can't be cleaned up
     - So we don't fill the database with bunch of deleted `TenantConfigs` (this is more of a cosmetic problem) 
  - The `CodeModule` foreign-key in an `AppMount` has to be set to Null in case the `AppMount` is deleted, so the `CodeModule` can be deleted if its not used. (Deleted `AppMounts` no longer use the related `CodeModule`) 
    - If a soft deleted `CodeModule` is not cleaned-up soon ™️  then you will have a problem with the `UNIQUE` constraint if the same version would be downloaded.

⚠️ ⚠️ ⚠️ ⚠️ ⚠️ REALLY IMPORTANT NOTES ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ 
- If you have previously installed the Operator that was already using the GORM implementation, you WILL face errors
   - The reason for this is that the current way we use the database is very messy, and things don't get cleaned up (just soft deleted)
   - If the soft-deleted things are not removed, there may be problems due to the `UNIQUE` constraint. (mainly for the `CodeModules`)
   - OSMounts were not updated properly, so they will point to a `VolumeMeta` that is actually not correct anymore

⚠️ ⚠️ ⚠️ ⚠️ ⚠️

## How can this be tested?

Play around with the csi-driver, check that the `CodeModulesImage` works, check that there are no database errors in the logs. 
Check if the garbage-collection works, make sure its not too aggressive.

Sidenote: I still want to add more unit-tests, especially to the "Cleaner" parts, but its ready to be reviewed


pkg/controllers/csi/driver/server.go
```diff
@@ -197,7 +197,7 @@ func (svr *Server) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpubli
 }
 
 func (svr *Server) unmountUnknownVolume(volumeInfo csivolumes.VolumeInfo) {
-	log.Info("VolumeID not present in the database", "volumeID", volumeInfo.VolumeID, "targetPath", volumeInfo.TargetPath)
+	log.Info("Couldn't handle volume with any publisher, trying to unmount", "volumeID", volumeInfo.VolumeID, "targetPath", volumeInfo.TargetPath)
```
andriisoldatenko, line 200: this error message is confusing.

pkg/controllers/csi/driver/volumes/app/publisher.go
```diff
@@ -265,16 +267,20 @@ func (publisher *AppVolumePublisher) ensureMountSteps(bindCfg *csivolumes.BindCo
...
+			return false, err
+		}
+
 		publisher.db.CreateAppMount(appMount)
 	} else if err != nil {
 		return false, err
 	}
 
-	if int(appMount.MountAttempts) > bindCfg.MaxMountAttempts {
+	if int(appMount.MountAttempts) > int(tenantConfig.MaxFailedMountAttempts) {
```
andriisoldatenko, line 283:
```suggestion
	if appMount.MountAttempts > tenantConfig.MaxFailedMountAttempts {
```
they both int64

pkg/controllers/csi/metadata/models.go
```diff
@@ -52,7 +52,7 @@ type OSMount struct {
 // AppMount keeps track of our mounts to user applications, where we provide the codemodules.
 type AppMount struct {
 	VolumeMeta VolumeMeta
-	CodeModule CodeModule `gorm:"foreignKey:CodeModuleVersion"`
+	CodeModule CodeModule `gorm:"foreignKey:CodeModuleVersion;constraint:OnUpdate:CASCADE,OnDelete:SET NULL;"`
```
andriisoldatenko, line 55: okey this code works, after upgrading:
```
sqlite> .schema app_mounts
CREATE TABLE `app_mounts` (`created_at` datetime,`updated_at` datetime,`deleted_at` datetime,`volume_meta_id` text,`code_module_version` text,`location` text NOT NULL,`mount_attempts` integer NOT NULL,PRIMARY KEY (`volume_meta_id`),CONSTRAINT `fk_app_mounts_volume_meta` FOREIGN KEY (`volume_meta_id`) REFERENCES `volume_meta`(`id`),CONSTRAINT `fk_app_mounts_code_module` FOREIGN KEY (`code_module_version`) REFERENCES `code_modules`(`version`) ON DELETE SET NULL ON UPDATE CASCADE);
CREATE INDEX `idx_app_mounts_deleted_at` ON `app_mounts`(`deleted_at`);
sqlite>
```

andriisoldatenko (dismissed):
practically LGTM.

waodim (approved):
It looks good to me, probably a third reviewer could not hurt as it is a big PR

albertogdd (approved):
LGTM. Will make more tests tomorrow. I wanted to try the garbage collector but being the latest code modules image then it doesn't get deleted at all, even if I delete the only pod using the version (which is correct).
image

andriisoldatenko (approved):

PR #3169, opened 2024-05-15
Trying things to fix conversion

0sewa0:
## Description
This is all the changes I did in order to find the source of the super annoying error:
```
admission webhook "webhook.dynatrace.com" denied the request: unable to decode dynatrace.com/v1beta1, Kind=DynaKube into *dynakube.DynaKube
```
But in my search of the cause I also did some changes that I think are an improvement OR absolutely necessary.

⚠️This is not a complete, merge and you are done change, the linting and tests have been messed up.

Detailed changes:
- The fix for the admission error:
   - It had nothing really to do with the conversion webhook, the problem was that the `ValidatingWebhookConfiguration` still was asking for `v1beta1` even though the code wanted to work with `v1beta2`, and (I guess) the validation webhook wanted validate the `v1beta1` dynakube  BEFORE the conversion even happened, but failed, as it couldn't convert it on its own.
- Moved the `Hub` to the `v1beta2` version
    - If our code is working only with `v1beta2` it makes no sense to store it in `v1beta1` as that would lead to constant conversions.
       - I know we did it like that for `v1alpha1` but its incorrect, so lets not repeat our mistake.
- Removed all the `operator-sdk` labels from the CR and introduced `kubebuilder` ones
   - The `operator-sdk` labels were just noise, and provided nothing
   - Now we are conforming to how we did it in `EdgeConnect`, yay
   - No more pointer/nil values for simple types, because we can set default values via `kubebuilder`, yaaaay
      - ⚠️ This lead to some code changes
- Removed code from the `v1beta1` package, as they were not used, and shouldn't be used
  - Feautre-flags are an exceptions, as those are needed for conversion (⚠️ the ones that are not converted still need to be removed as they will present in `v1beta2`)
  - Removed the `properties.go`, maybe that was a mistake, could be useful for conversion, but IMO its cleaner if we do the conversion "raw", and nothing is hidden behind magic functions.
- Updated the `ConvertTo` function (⚠️ very important and its not finished, and `ConvertFrom` needs similar type of love)
   -  Before my change, the `ConvertTo/From` function only cared about the conversion that I would consider "more complicated", but it didn't much care for the very annoying an monotonous ones. (like copying all the `v1beta1.CloudNativeSpec` to the 99% same `v1beta2.CloudNativeSpec`) 
   - These super monotonous conversions need to be done, because when you are using the `webhook` conversion method, there is no help from kubernetes, its all in our responsibility to do.
      - There are probably nicer ways to do it, because of how similar the structures are, but I went the safe/least-magic route.

## How can this be tested?

Apply 1 `v1beta1` Dynakube, should work and move most fields (didn't fix the moving of the whole status)
Apply 1 `v1beta2` Dynakube, should just work, didn't touch the conversion logic of that, but didn't have a problem with it as of now


cmd/webhook/builder.go
```diff
@@ -2,12 +2,12 @@ package webhook
 
 import (
 	"context"
+	dynatracev1beta2 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube"
```
chrismuellner, line 5: linter?

0sewa0: pls read the description first 😬 

> ⚠️This is not a complete, merge and you are done change, the linting and tests have been messed up, and there are still changes that need to be finished.

as this is not my ticket I didn't want to do everything, this is not aimed at `main` but another PR.

pkg/api/v1beta2/dynakube/activegate_types.go
```diff
@@ -54,32 +54,27 @@ var ActiveGateDisplayNames = map[CapabilityDisplayName]struct{}{
 type ActiveGateSpec struct {
 
 	// Adds additional annotations to the ActiveGate pods
-	// +optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Annotations",order=27,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:text"}
```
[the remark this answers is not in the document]
0sewa0: part of the description: 😬 
> Removed all the operator-sdk labels from the CR and introduced kubebuilder ones
> - The operator-sdk labels were just noise, and provided nothing
> - Now we are conforming to how we did it in EdgeConnect, yay
> - No more pointer/nil values for simple types, because we can set default values via kubebuilder, yaaaay
>   - ⚠️ This lead to some code changes

we can not remove them, and just add the kubebuilder ones
we can look into if the `csv` related labels are needed in the future

pkg/api/v1beta1/dynakube/convert_to.go
```diff
@@ -0,0 +1,226 @@
...
+func toAppInjectSpec(src AppInjectionSpec) *v1beta2.AppInjectionSpec {
+	dst := &v1beta2.AppInjectionSpec{}
+
+	dst.CodeModulesImage = src.CodeModulesImage
+	dst.InitResources = src.InitResources
+
+	return dst
+}
+
+func isLabelSelectorEmpty(selector metav1.LabelSelector) bool {
```
chrismuellner, line 224: nit: should this be a common util?

pkg/api/v1beta2/dynakube/dynakube_types.go
```diff
@@ -59,9 +57,11 @@ type DynaKubeValueSource struct { //nolint:revive
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:storageversion
```
chrismuellner, line 60: is `+kubebuilder:storageversion` necessary twice?

pkg/api/v1beta2/dynakube/dynakube_types.go
```diff
@@ -134,20 +129,26 @@ type DynaKubeSpec struct { //nolint:revive
 	ActiveGate ActiveGateSpec `json:"activeGate,omitempty"`
 
 	// Configuration for Metadata Enrichment.
-	// +optional
+	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="MetaData Enrichment",order=9,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced"}
 	MetaDataEnrichment MetaDataEnrichment `json:"metaDataEnrichment,omitempty"`
```
chrismuellner, line 134:
```suggestion
	MetadataEnrichment MetaDataEnrichment `json:"metadataEnrichment,omitempty"`
```

pkg/api/v1beta2/dynakube/metada_enrichment.go
```diff
@@ -3,13 +3,13 @@ package dynakube
...
-	// Enabled by default.
-	// +kubebuilder:default:=true
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="MetaDataEnrichment",xDescriptors="urn:alm:descriptor:com.tectonic.ui:selector:booleanSwitch"
-	Enabled *bool `json:"enabled"`
 
 	// The namespaces where you want Dynatrace Operator to inject enrichment.
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Namespace Selector",xDescriptors="urn:alm:descriptor:com.tectonic.ui:selector:core:v1:Namespace"
 	NamespaceSelector metav1.LabelSelector `json:"namespaceSelector,omitempty"`
+	// Enables MetaDataEnrichment.
+	// Enabled by default.
```
chrismuellner, line 11: > Enabled by default.

Feels redundant to me.

```
k explain dynakubes.dynatrace.com.spec.metaDataEnrichment
...

FIELDS:
  enabled	<boolean> -required-
    Enables MetaDataEnrichment.
    Enabled by default.
...
```

pkg/api/v1beta1/dynakube/convert_to.go
0sewa0: its actually not necessary because `src.Spec.NamespaceSelector.Size() != 0` does the same

pkg/api/v1beta2/dynakube/dynakube_types.go
0sewa0: missed it, nice catch 🍪

chrismuellner (approved):

chrismuellner (approved):

PR #3170, opened 2024-05-16
Change logic in WorkLoad retrieval when OwnerReference of Pod is no Controller

waodim:
## Description

Respective ticket: https://dt-rnd.atlassian.net/browse/K8S-9998

Because we normally assign the podName to the workload info in a certain corner case (when we have a ownerreference which is no controller) it can occur that the DT_WORKLOAD_KIND env var is set to empty.

This is fixed by simply using GenerateName instead of Name.

## How can this be tested?

- unit tests
- I am not really sure how to properly test it otherwise at it seems to be a corner case which needs special set up


pkg/webhook/mutation/pod/metadata/workload.go
```diff
@@ -102,6 +102,10 @@ func findRootOwner(ctx context.Context, clt client.Client, partialObjectMetadata
 		}
 	}
 
+	if partialObjectMetadata.ObjectMeta.Name == "" {
```
StefanHauth, line 105: I think this would be done better after the whole recursion is done. Move this to ~ line 60.

StefanHauth (changes requested):

pkg/webhook/mutation/pod/metadata/workload.go
```diff
@@ -46,7 +46,7 @@ func findRootOwnerOfPod(ctx context.Context, clt client.Client, pod *corev1.Pod,
 			Kind:       pod.Kind,
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: pod.ObjectMeta.Name,
+			Name: pod.ObjectMeta.GenerateName,
```
0sewa0, line 49: this is wrong, this is not your problem, this will break more things, the logic does not work like this

see:
```
func findRootOwner(ctx context.Context, clt client.Client, partialObjectMetadata *metav1.PartialObjectMetadata) (workloadInfo, error) {
	if len(partialObjectMetadata.ObjectMeta.OwnerReferences) == 0 {
		if partialObjectMetadata.ObjectMeta.Name == "" {
			// pod is not created directly and does not have an owner reference set
			return newUnknownWorkloadInfo(), nil
		}
		return newWorkloadInfo(partialObjectMetadata), nil
	}
....
```
We handle `Pod`s that are directly created.
Setting the generateName here is, firstly just incorrect, and second it wont result in what you want, which is `UNKNOWN` workload in the env-var (according to the linked ticket)

A problematic workload looks like:
```
apiVersion: v1
kind: Pod
metadata:
  name: 
  namespace: feh0-private-dev
  ownerReferences:
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false                                 <- Pay attention to this
    kind: CatalogSource
    name: ibm-operator-catalog
    uid: 2902465e-7214-46ba-8e7c-48fa2067587b
```

A non-problematic workload  looks like:
```
apiVersion: v1
kind: Pod
metadata:
  generateName: collector-
  name: collector-dm82x
  namespace: gmp-system
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true                             <- Pay attention to this
    kind: DaemonSet
    name: collector
    uid: db5ee989-b467-4e45-8499-5bd47a3bf85c
  resourceVersion: "2317"
```

Your problem is `if owner.Controller != nil && *owner.Controller ` (line 76)
The problem is that we don't handle the owner-references that are not pointing to a `controller`.

I would just do:

```
if owner.Controller != nil && *owner.Controller {
...
} else {
	// pod is created by non-controller workload
	return newUnknownWorkloadInfo(), nil
}
```

or just remove the `is a controller` check, dunno why we have it, but that require asking around

pkg/webhook/mutation/pod/metadata/workload.go
waodim: thanks for pointing out, you are right. done in https://github.com/Dynatrace/dynatrace-operator/pull/3170/commits/8fe5b61c123083cd2161db67f9379d4c06a58b68

pkg/webhook/mutation/pod/metadata/workload.go
```diff
@@ -102,6 +102,10 @@ func findRootOwner(ctx context.Context, clt client.Client, partialObjectMetadata
 		}
 	}
 
+	if partialObjectMetadata.ObjectMeta.Name == "" {
+		log.Info("podName is empty", "objectMeta.name", partialObjectMetadata.ObjectMeta.Name)
```
0sewa0: you are saying something is empty, then printing it, why? 😅

0sewa0 (changes requested):

pkg/webhook/mutation/pod/metadata/workload.go
waodim: It was a long day... 🥲  thank you for pointing out :+1:

pkg/webhook/mutation/pod/metadata/workload.go
waodim: I had a discussion with @luhi-DT and @StefanHauth where we discussed handling the Controller of the Ownerreference and came to the conclusion that we should simply change the name to the general name as the case in which the failure was was a corner case in our opinion.

@StefanHauth even had a sample solution that is similar to yours.

Your explanations lead to me thinking that we should go with your proposed solution. What do you guys think? @luhi-DT @StefanHauth 

Thank you for the input

0sewa0: but then your discussion contradict the ticket that this supposed to solve:

> The attached pod manifest shows the pod of the linked RFA after it has been created (and failed).
> If this happens, resolving the workload info needed should result in an UNKNOWN workload.

also setting the `GenerateName` there will cause that according to the metadata enrichment the workload is a `Pod` but actually is not at all.
 - This is false data, `UNKNOWN` is at least not false, but incomplete
 - When the workload is just a `Pod` it HAS a `Name`, so using the `GenerateName` there is just wrong
   - Setting the `GenerateName` there 	in case the workload is just a `Pod` will cause problems again, because why would a "lonley" `Pod` have a `GenerateName`
   
You should also not forget that this is a recursive function, so the `partialObjectMetadata` can be any of the following:
- `Deployment`
- `DaemonSet`
- `StatefulSet`
-  etc.
And if you check (for example) a `Deployment` in your cluster, it does NOT have a `GenerateName`:
- Example:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    components.gke.io/layer: addon
    deployment.kubernetes.io/revision: "1"
  creationTimestamp: "2024-05-03T08:33:21Z"
  generation: 1
  labels:
    addonmanager.kubernetes.io/mode: Reconcile
    app: managed-prometheus-operator
    app.kubernetes.io/component: operator
    app.kubernetes.io/name: gmp-operator
    app.kubernetes.io/part-of: gmp
  name: gmp-operator
  namespace: gmp-system
  resourceVersion: "2001"
  uid: 54440195-91c9-4cdd-8509-b03fb930e834
spec:
```

pkg/webhook/mutation/pod/metadata/workload_test.go
```diff
@@ -44,8 +45,9 @@ func TestFindRootOwnerOfPod(t *testing.T) {
 						Controller: address.Of(true),
 					},
 				},
-				Name:      resourceName,
-				Namespace: namespaceName,
+				Name:         resourceName,
+				Namespace:    namespaceName,
+				GenerateName: resourceName,
```
0sewa0, line 50: Related to my comment above, a `Deployment` will not have a `GenerateName`, so this test is incorrect:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    components.gke.io/layer: addon
    deployment.kubernetes.io/revision: "1"
  creationTimestamp: "2024-05-03T08:33:21Z"
  generation: 1
  labels:
    addonmanager.kubernetes.io/mode: Reconcile
    app: managed-prometheus-operator
    app.kubernetes.io/component: operator
    app.kubernetes.io/name: gmp-operator
    app.kubernetes.io/part-of: gmp
  name: gmp-operator
  namespace: gmp-system
  resourceVersion: "2001"
  uid: 54440195-91c9-4cdd-8509-b03fb930e834
spec:
```

Disclaimer: I do not know if there is a scenario that a `Deployment` will have a `GenerateName`, but all the `Deployments` I have in my cluster are lacking a `GenerateName`

pkg/webhook/mutation/pod/metadata/workload.go
0sewa0: How `GeneratedName` works: https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values

TLDR: Anyone CAN set a `GeneratedName` instead of a `Name`, so nothing is super certain, like my "Deployments don't have GenerateName set" is not correct, its more accurate that "Deployments don't tend to have GenerateName set"

So this tells me that this whole recursive logic is flawed. (flawed == there are edgecases that we don't cover)

pkg/webhook/mutation/pod/metadata/workload_test.go
waodim: thank you for clearing things up should be done in https://github.com/Dynatrace/dynatrace-operator/pull/3170/commits/92c10486b2ce3fa35e54b105a364f1dff9bd273c

pkg/webhook/mutation/pod/metadata/workload.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3170/commits/92c10486b2ce3fa35e54b105a364f1dff9bd273c

pkg/webhook/mutation/pod/metadata/workload.go
waodim: thank you for clearing everything up, done in https://github.com/Dynatrace/dynatrace-operator/pull/3170/commits/92c10486b2ce3fa35e54b105a364f1dff9bd273c

pkg/webhook/mutation/pod/metadata/workload.go
```diff
@@ -58,6 +58,10 @@ func findRootOwnerOfPod(ctx context.Context, clt client.Client, pod *corev1.Pod,
 		return nil, errors.WithStack(err)
 	}
 
+	if podPartialMetadata.ObjectMeta.Name == "" {
+		log.Info("podName is empty")
+	}
```
0sewa0, line 63: this will just spam the logs.

try it out with a `Deployment` that has 100 replicas
this line will appear 100 times, not the most helpful 😅

0sewa0: also `findRootOwner` does not change `podPartialMetadata`, it just used as a starting point for the recursion

so printing it here is even more questionable IMO, because it would MAYBE be interesting before trying to find the owner.
- The "MAYBE" comes from the fact that it can happen that you have `Name` and `GeneratedName` set, because the pod was not `Created` but  `Updated`, so only looking at the `Name`/`GeneratedName` will give you an incomplete picture.

0sewa0 (changes requested):

pkg/webhook/mutation/pod/metadata/workload.go
waodim: Okay that makes sense of course, I think we could get rid off it anyways, now that we went for dealing with owners when they are no controller

pkg/webhook/mutation/pod/metadata/workload.go
```diff
@@ -99,6 +99,9 @@ func findRootOwner(ctx context.Context, clt client.Client, partialObjectMetadata
 			}
 
 			return findRootOwner(ctx, clt, ownerObjectMetadata)
+		} else {
+			// pod is created by workload of kind that is not well known
+			return newUnknownWorkloadInfo(), nil
```
aorcholski, line 104: there might be more owner references
so `return` statement here seems to be premature

waodim: https://github.com/Dynatrace/dynatrace-operator/pull/3170/commits/d0f8b4e6a080075fff2c926bd4e2321004851a36 Is a proposed solution. I am not sure how stable it is so please if anyone has feedback please let me know.

StefanHauth (approved):
LGTM

pkg/webhook/mutation/pod/metadata/workload_test.go
```diff
@@ -114,11 +114,23 @@ func TestFindRootOwnerOfPod(t *testing.T) {
 		assert.Equal(t, "UNKNOWN", workloadInfo.kind)
 	})
 
-	t.Run("should be empty if owner is not set, but name is empty", func(t *testing.T) {
+	t.Run("should be empty if owner is no controller", func(t *testing.T) {
```
aorcholski, line 117:
```suggestion
	t.Run("should be unknown if no controller is the owner", func(t *testing.T) {
```

aorcholski (commented):
line 93

```
t.Run("should be empty if owner is not well known", func(t *testing.T) {
                 vvvv
t.Run("should be unknown if owner is not well known", func(t *testing.T) {
```

pkg/webhook/mutation/pod/metadata/workload.go
```diff
@@ -102,6 +106,11 @@ func findRootOwner(ctx context.Context, clt client.Client, partialObjectMetadata
 		}
 	}
 
+	if !foundController {
```
aorcholski, line 109: In such a case :
```
pod < controller A < controller B < owner
```
the current way - function returns B
fixed way - function returns UNKNOWN

0sewa0 (approved):

luhi-DT (approved):

0sewa0 (approved):

aorcholski (approved):

PR #3183, opened 2024-05-21
Add readOnlyRootFs check for OneAgent DaemonSet

0sewa0:
## Description

Since v1.291 of the OneAgent the `securityContext. ReadOnlyRootFilesystem = true` is supported in the CloudNative and HostMonitoring cases.

This introduces a check that:
- if the version is older than  v1.291, we **don't** set `ReadOnlyRootFilesystem` to `true`
- if the the version is v1.291 or newer or unknown, we set `ReadOnlyRootFilesystem` to `true`

Related ticket: [K8S-5812](https://dt-rnd.atlassian.net/browse/)

## How can this be tested?

Deploy a older OneAgent version, see that the `ReadOnlyRootFilesystem` is `false`
Deploy a current OneAgent version, see that the `ReadOnlyRootFilesystem` is `true`


pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -42,6 +44,8 @@ const (
 
 	probeMaxInitialDelay         = int32(90)
 	probeDefaultSuccessThreshold = int32(1)
+
+	readOnlyRootFsVersionThreshold = "v1.291"
```
andriisoldatenko, line 48: it's not a threshold, it's aboslutly valid sem version.
wdyt to rename it to: `readOnlyRootFsMinVersion`?

andriisoldatenko: >Range constraint
> Constraint will be satisfied by versions matching >=1.291.0 <1.292.0-0.
> 1.291 is a range constraint. It means that it will match several versions.

https://jubianchi.github.io/semver-check/#/constraint/1.291

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -380,3 +385,22 @@ func (dsInfo *builderInfo) getReadinessProbe() *corev1.Probe {
...
+
+// isRootFsReadonly checks if the given version of the OneAgent supports the `ReadOnlyRootFilesystem` securityContext setting.
+// if the version is not set, ie.: unknown, we  consider the OneAgent to support `ReadOnlyRootFilesystem`.
+func (dsInfo *builderInfo) isRootFsReadonly() bool {
+	if dsInfo.dynakube != nil &&
+		dsInfo.dynakube.NeedsReadOnlyOneAgents() &&
+		dsInfo.dynakube.OneAgentVersion() != "" {
+		agentSemver, err := dtversion.ToSemver(dsInfo.dynakube.OneAgentVersion())
+		if err != nil {
+			log.Error(err, "Unable to determine OneAgent version to enable readonly pod filesystem, skipping")
```
andriisoldatenko, line 397: I would prefer to use Debug instead of error and add more context to it, because we do nothing with it :)

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -380,3 +385,22 @@ func (dsInfo *builderInfo) getReadinessProbe() *corev1.Probe {
 
 	return defaultProbe
 }
+
+// isRootFsReadonly checks if the given version of the OneAgent supports the `ReadOnlyRootFilesystem` securityContext setting.
+// if the version is not set, ie.: unknown, we  consider the OneAgent to support `ReadOnlyRootFilesystem`.
+func (dsInfo *builderInfo) isRootFsReadonly() bool {
```
andriisoldatenko, line 391: I know it's not you, but based on convention it should be `b` or `bi`, dsInfo is a bit confusing. =)

aorcholski (commented):
the default image is :
```
    image: <tenant>.dev.dynatracelabs.com/linux/oneagent:<tag>
```    
no problem in log.

I set exactly the same value in dynakube:
```
cloudNativeFullStack:
    image: <tenant>.dev.dynatracelabs.com/linux/oneagent:<tag>
```    
oa pods are running but operator log says :
```
"error":"custom-image is not possible to convert to a semver format.","level":"error","logger":"oneagent-daemonset","msg":"Unable to determine OneAgent version to enable readonly pod filesystem, skipping","stacktrace":"custom-image is not possible to convert to a semver format.
github.com/Dynatrace/dynatrace-operator/pkg/util/dtversion.ToSemver
...
<huge stack>
```

status.oneagent.version set to `custom-image`  is sth normal. There should be no stack trace because it's confusing and actually not an error - we support custom `image:`

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
waodim: I agree with the name `readOnlyRootFsConstraint` nice catch :+1:

waodim (commented):
apart from the other things I do not have antyhing extra to add here. it LGTM, let us wait until you adress all points

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
0sewa0: [this](https://github.com/Dynatrace/dynatrace-operator/pull/3183/commits/6b2b72a9db41dfcde17aeb5d39da74b78b8c1a49) is your fault now 😛

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
0sewa0: done

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
0sewa0: done

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
andriisoldatenko: love it!

andriisoldatenko (approved):

aorcholski (approved):
LGTM

waodim (approved):

PR #3193, opened 2024-05-22
Add Dynakube status condition for the AG dynakube-activegate-authtoken-secret secret

aorcholski:
## Description

Adds new `ActiveGateAuthTokenSecret` condition :
```
  - lastTransitionTime: "2024-05-24T10:57:12Z"
    message: secret created 0 day(s) ago, token:dt0g02.ABCDEFGH
    reason: SecretCreated
    status: "True"
    type: ActiveGateAuthTokenSecret
```    

`lastTransitionTime` field is updated if:
- token expired and new instance of the secret is created
- k8s client error occurred
- Dynatrace server error occurred

In case of an error the `message` field contains description of the error. Otherwise the message informs about token's lifetime and public part of the token. The field is updated once a day counting down the token's lifetime (it doesn't affect `lastTransitionTime` field).

For example the secret was created on Friday:
```
  - lastTransitionTime: "2024-05-24T10:57:12Z"                                                                                                                                                                                                                                                                                                                                                
    message: secret created 0 day(s) ago, token:dt0g02.LVBZLERV                                                                                                                                                                                                                                                                                                                               
    reason: SecretCreated                                                                                                                                                                                                                                                                                                                                                                     
    status: "True"                                                                                                                                                                                                                                                                                                                                                                            
    type: ActiveGateAuthTokenSecret
```
after weekend:
```
  - lastTransitionTime: "2024-05-24T10:57:12Z"
    message: secret created 2 day(s) ago, token:dt0g02.LVBZLERV
    reason: SecretCreated
    status: "True"
    type: ActiveGateAuthTokenSecret
```


## How can this be tested?

Unittests.

Deploy CN with ActiveGate. 

Check AG status after 1 day :
- `lastTransitionTime` - shoud be the same
- `message` - token should be the same, lifetime should be set to `1 day(s) ago`

Remove AG from the dynakube :
- the condition shoud be removed from the dynakube status

### DIFF
--- a/pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
+++ b/pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
@@ -2,11 +2,14 @@ package authtoken
 
 import (
 	"context"
+	"strconv"
+	"strings"
 	"time"
 
 	dynatracev1beta2 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube"
 	dtclient "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/conditions"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/secret"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -41,19 +44,6 @@ func NewReconciler(clt client.Client, apiReader client.Reader, dynakube *dynatra
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context) error {
-	if !r.dynakube.NeedsActiveGate() {
-		return nil
-	}
-
-	err := r.reconcileAuthTokenSecret(ctx)
-	if err != nil {
-		return errors.WithMessage(err, "failed to create activeGateAuthToken secret")
-	}
-
-	return nil
-}
-
-func (r *Reconciler) reconcileAuthTokenSecret(ctx context.Context) error {
 	var secret corev1.Secret
 
 	err := r.apiReader.Get(ctx,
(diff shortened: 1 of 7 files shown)


pkg/util/conditions/secret.go
```diff
@@ -40,3 +40,13 @@ func SetSecretOutdated(conditions *[]metav1.Condition, conditionType, message st
 	}
 	_ = meta.SetStatusCondition(conditions, condition)
 }
+
+func SetAuthSecretCreated(conditions *[]metav1.Condition, conditionType string, msg string) {
```
0sewa0, line 44: how is this different from `SetSecretCreated`? 😅

pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
```diff
@@ -41,19 +44,6 @@ func NewReconciler(clt client.Client, apiReader client.Reader, dynakube *dynatra
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context) error {
-	if !r.dynakube.NeedsActiveGate() {
```
0sewa0, line 44: why was this removed?

I know we check in the AG reconciler already, but that is not how it should be done:
- Each `Reconciler` should "own" the things they create, so the AG reconciler should just call the `AuthToken` reconciler, that will clean up OR create what is necessary.
  - Otherwise, to do cleanup, each `Reconciler` would need a `CleanUp` function that is called by the original "parent" calling the `Reconciler`.
    - This unnecessarily spreads the logic between packages (makes testing a bit more annoying)

    
The line where you added the cleanup of the new condition has a comment: `// TODO: move cleanup to ActiveGate reconciler`, that now would need another: `// TODO: move cleanup to AuthToken reconciler` 😅 


IMO how a `Reconile` func should look like: (not everything needs to be sooo abstracted, it was just easier to write here 😅 )

```go
	if !r.needed() { 
		return r.cleanUp()
	}
	return r.doStuff()
}
```

0sewa0 (changes requested):

pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
aorcholski: I wasn't sure the meaning of `// TODO: move cleanup to ActiveGate reconciler` so I moved cleanup to AG reconciler :smile: .

Fixed. The auth-token condition is removed by authtoken reconciler.

pkg/util/conditions/secret.go
aorcholski: Moved to authtoken package because it is none-generic function to set specific condition message.

0sewa0 (dismissed):
Tested it, works as expected

only added small comments/compliments 😉 , fix at your discretion (just ping me if you do so I can reapprove)

0sewa0 (approved):

wepudt (approved):
code look good to me

PR #3197, opened 2024-05-23
Fix workload owner detection

luhi-DT:
## Description

Ticket: https://dt-rnd.atlassian.net/browse/K8S-10007

Fixes the current state of the workload detection, the pod itself is always the default owner of itself, if there is a well known owner in the hierarchy, then we take the highest in line as the owner. There is no more unknown, as the default should be the pod itself

## How can this be tested?

Deploy operator -> create pods with owner references and check the env vars

### DIFF
--- a/pkg/webhook/mutation/pod/metadata/workload.go
+++ b/pkg/webhook/mutation/pod/metadata/workload.go
@@ -81,9 +81,11 @@ func findRootOwner(ctx context.Context, clt client.Client, childObjectMetadata *
 			if err != nil {
 				return childObjectMetadata, err
 			}
+
 			if isWellKnownWorkload(parentObjectMetadata) {
 				return parentObjectMetadata, nil
 			}
+
 			return childObjectMetadata, nil
 		}
 	}
--- a/pkg/webhook/mutation/pod/metadata/workload.go
+++ b/pkg/webhook/mutation/pod/metadata/workload.go
@@ -3,7 +3,7 @@ package metadata
 import (
 	"context"
 
-	"github.com/Dynatrace/dynatrace-operator/pkg/consts"
+	kubeobjects "github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/pod"
 	dtwebhook "github.com/Dynatrace/dynatrace-operator/pkg/webhook"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -23,13 +23,6 @@ func newWorkloadInfo(partialObjectMetadata *metav1.PartialObjectMetadata) worklo
 	}
 }
 
-func newUnknownWorkloadInfo() workloadInfo {
-	return workloadInfo{
-		name: consts.EnrichmentUnknownWorkload,
-		kind: consts.EnrichmentUnknownWorkload,
-	}
-}
-
 func (mut *Mutator) retrieveWorkload(request *dtwebhook.MutationRequest) (*workloadInfo, error) {
 	workload, err := findRootOwnerOfPod(request.Context, mut.metaClient, request.Pod, request.Namespace.Name)
 	if err != nil {
@@ -46,14 +39,14 @@ func findRootOwnerOfPod(ctx context.Context, clt client.Client, pod *corev1.Pod,
 			Kind:       pod.Kind,
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: pod.ObjectMeta.Name,
+			Name: kubeobjects.GetName(*pod),
 			// pod.ObjectMeta.Namespace is empty yet
 			Namespace:       namespace,
 			OwnerReferences: pod.ObjectMeta.OwnerReferences,
 		},
 	}
 
-	workloadInfo, err := findRootOwner(ctx, clt, podPartialMetadata)
+	workloadInfo, err := findRootOwner(ctx, clt, podPartialMetadata, podPartialMetadata) // default owner of the pod is the pod itself
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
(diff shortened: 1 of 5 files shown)


pkg/webhook/mutation/pod/metadata/workload_test.go
```diff
@@ -90,31 +90,37 @@ func TestFindRootOwnerOfPod(t *testing.T) {
...
 			TypeMeta: metav1.TypeMeta{
 				Kind: "Pod",
 			},
 			ObjectMeta: metav1.ObjectMeta{
 				OwnerReferences: []metav1.OwnerReference{
 					{
-						APIVersion: "unknown",
-						Kind:       "unknown",
+						APIVersion: "v1",
+						Kind:       "Secret",
```
chrismuellner, line 102: wouldn't `unknown` be fine here? It should just be a resource that is not well known

pkg/webhook/mutation/pod/metadata/workload_test.go
```diff
@@ -90,31 +90,37 @@ func TestFindRootOwnerOfPod(t *testing.T) {
...
 						Name:       "test",
 						Controller: address.Of(true),
 					},
 				},
 				Name: resourceName,
 			},
 		}
-		client := fake.NewClient(&pod)
+		namespace := corev1.Secret{
```
chrismuellner, line 110:
```suggestion
		secret := corev1.Secret{
```

pkg/webhook/mutation/pod/metadata/workload.go
```diff
@@ -46,43 +39,29 @@ func findRootOwnerOfPod(ctx context.Context, clt client.Client, pod *corev1.Pod,
...
 
-func findRootOwner(ctx context.Context, clt client.Client, partialObjectMetadata *metav1.PartialObjectMetadata) (workloadInfo, error) {
-	if len(partialObjectMetadata.ObjectMeta.OwnerReferences) == 0 {
-		if partialObjectMetadata.ObjectMeta.Name == "" {
-			// pod is not created directly and does not have an owner reference set
-			return newUnknownWorkloadInfo(), nil
-		}
-
-		return newWorkloadInfo(partialObjectMetadata), nil
+func findRootOwner(ctx context.Context, clt client.Client, rootObject *metav1.PartialObjectMetadata, wellKnownOwner *metav1.PartialObjectMetadata) (workloadInfo, error) {
```
0sewa0, line 57: IMO this is an over-complicated way of solving the problem. (you are trying to reuse the old implementation to solve a new problem, instead of writing a solution to our current problem)

**As far as I understand the problem:**
The `OwnerReferences` is a tree, where the root is the `owner` that started everything.

What we want is to return the `owner` in the tree that is:
1. a "controller"
2. "well-known"
3. closest to the root 

**A possible way to go about finding the desired owner:**
W need to do is just reach the root of the tree (where we have no more `OwnerReferences`), with the following conditions:
- only go "down" a level if we find a "controller" (1. condition satisfied)

When we're coming back from the root we have to make sure that we only return an `owner` if its "well-known" (2. condition satisfied)
- This will cause the `owner` closest to the root to be returned at the end. (3. condition satisfied)

pkg/webhook/mutation/pod/metadata/workload.go
```diff
@@ -46,43 +39,29 @@ func findRootOwnerOfPod(ctx context.Context, clt client.Client, pod *corev1.Pod,
...
+	if len(rootObject.ObjectMeta.OwnerReferences) == 0 {
+		return newWorkloadInfo(wellKnownOwner), nil
```
0sewa0, line 59: IMO we should only create the `workloadInfo` outside of `findRootOwner`, so we work with only `*metav1.PartialObjectMetadata` making things more straight forward

pkg/webhook/mutation/pod/metadata/workload.go
chrismuellner, line 57:
```suggestion
func findRootOwner(ctx context.Context, clt client.Client, rootObject *metav1.PartialObjectMetadata, prevWellKnownOwner *metav1.PartialObjectMetadata) (workloadInfo, error) {
```

pkg/webhook/mutation/pod/metadata/workload.go
```diff
@@ -99,19 +78,18 @@ func findRootOwner(ctx context.Context, clt client.Client, partialObjectMetadata
...
 
-	if !foundController {
-		// Out of all owner references no one is a Controller
-		return newUnknownWorkloadInfo(), nil
+			return findRootOwner(ctx, clt, ownerObjectMetadata, wellKnownOwner)
+		}
 	}
 
-	return newWorkloadInfo(partialObjectMetadata), nil
+	return newWorkloadInfo(wellKnownOwner), nil
```
0sewa0, line 92: Here is a draft of the solution I was talking about:
```go
package metadata

import (
	"context"

	dtwebhook "github.com/Dynatrace/dynatrace-operator/pkg/webhook"
	"github.com/pkg/errors"
	corev1 "k8s.io/api/core/v1"
)

type workloadInfo struct {
	name string
	kind string
}

func newWorkloadInfo(partialObjectMetadata *metav1.PartialObjectMetadata) *workloadInfo {
	name := partialObjectMetadata.ObjectMeta.Name
	if name == "" {
		name = partialObjectMetadata.ObjectMeta.GenerateName
	}
	return &workloadInfo{
		name: name,
		kind: partialObjectMetadata.Kind,
	}
}

func (mut *Mutator) retrieveWorkload(request *dtwebhook.MutationRequest) (*workloadInfo, error) {
	workload, err := findRootOwnerOfPod(request.Context, mut.metaClient, request.Pod, request.Namespace.Name)
	if err != nil {
		return nil, err
	}

	return workload, nil
}

func findRootOwnerOfPod(ctx context.Context, clt client.Client, pod *corev1.Pod, namespace string) (*workloadInfo, error) {
	podPartialMetadata := &metav1.PartialObjectMetadata{
		TypeMeta: metav1.TypeMeta{
			APIVersion: pod.APIVersion,
			Kind:       pod.Kind,
		},
			Name: pod.ObjectMeta.Name,
			// pod.ObjectMeta.Namespace is empty yet
			Namespace:       namespace,
			OwnerReferences: pod.ObjectMeta.OwnerReferences,
		},
	}

	rootOwner, err := findRootOwner(ctx, clt, podPartialMetadata)
	if err != nil {
	}

	return newWorkloadInfo(rootOwner), nil
}

func findRootOwner(ctx context.Context, clt client.Client, childObjectMetadata *metav1.PartialObjectMetadata) (parentObjectMetadata *metav1.PartialObjectMetadata, err error) {
	objectMetadata := childObjectMetadata.ObjectMeta
	for _, owner := range objectMetadata.OwnerReferences {
		if owner.Controller != nil && *owner.Controller {
			parentObjectMetadata = &metav1.PartialObjectMetadata{
				TypeMeta: metav1.TypeMeta{
					APIVersion: owner.APIVersion,
					Kind:       owner.Kind,
				},
			}

			err = clt.Get(ctx, client.ObjectKey{Name: owner.Name, Namespace: objectMetadata.Namespace}, parentObjectMetadata)
			if err != nil {
				log.Error(err, "failed to query the object",
					"apiVersion", owner.APIVersion,
					"kind", owner.Kind,
					"name", owner.Name,
					"namespace", objectMetadata.Namespace,
				)

				return childObjectMetadata, err
			}

			parentObjectMetadata, err = findRootOwner(ctx, clt, parentObjectMetadata)
			if err != nil {
				return childObjectMetadata, err
			}
			if isWellKnownWorkload(parentObjectMetadata){
				return parentObjectMetadata, nil
			}
			return childObjectMetadata, nil
		}
	}

	return childObjectMetadata, nil
}

func isWellKnownWorkload(owner *metav1.PartialObjectMetadata) bool {
	knownWorkloads := []metav1.TypeMeta{
		{Kind: "ReplicaSet", APIVersion: "apps/v1"},
		{Kind: "Deployment", APIVersion: "apps/v1"},
		{Kind: "ReplicationController", APIVersion: "v1"},
		{Kind: "StatefulSet", APIVersion: "apps/v1"},
		{Kind: "DaemonSet", APIVersion: "apps/v1"},
		{Kind: "Job", APIVersion: "batch/v1"},
		{Kind: "CronJob", APIVersion: "batch/v1"},
		{Kind: "DeploymentConfig", APIVersion: "apps.openshift.io/v1"},
	}

	for _, knownController := range knownWorkloads {
		if owner.Kind == knownController.Kind &&
			owner.APIVersion == knownController.APIVersion {
			return true
		}
	}

	return false
}
```

sidenote: we talked about it yesterday, I felt there was a better way to do this, so I quickly came up with this, (didn't test it)

0sewa0 (commented):
IMO there is better solution 😅 😬

chrismuellner (changes requested):

pkg/webhook/mutation/pod/metadata/workload.go
luhi-DT: applied

0sewa0 (approved):
tested it => works

LGTM

chrismuellner (approved):

PR #3198, opened 2024-05-23
Add Conditions for ActiveGate and CodeModules to the Istio Reconciler

wepudt:
## Description

[K8S-9578](https://dt-rnd.atlassian.net/browse/K8S-9578?atlOrigin=eyJpIjoiMzk3NmZmOGU3MDQ1NGIwYjg3MjA3MzlhZGE0MTUyMzUiLCJwIjoiaiJ9)

Have conditions for the istio reconciler, for the ActiveGate and CodeModule connection-info ServiceEntry creation logic.
Have “Reasons“ for the following scenarios:

* ServiceEntries and VirualServices created/updated

* Error due to kubernetes API problems

## How can this be tested?
 
* Run the unit-tests.
* Deploy Istio and a Dynakube with Istio enabled. Describe the DynaKube and look for the newly added conditions starting with `IstioServiceConfigurationFor*`.

### DIFF
--- a/pkg/controllers/dynakube/activegate/reconciler.go
+++ b/pkg/controllers/dynakube/activegate/reconciler.go
@@ -107,12 +107,10 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 		return err
 	}
 
-	if r.dynakube.NeedsActiveGate() {
-		if r.istioReconciler != nil {
-			err = r.istioReconciler.ReconcileActiveGateCommunicationHosts(ctx, r.dynakube)
-			if err != nil {
-				return err
-			}
+	if r.istioReconciler != nil {
+		err = r.istioReconciler.ReconcileActiveGateCommunicationHosts(ctx, r.dynakube)
+		if err != nil {
+			return err
 		}
 	}
 
--- a/pkg/controllers/dynakube/injection/reconciler.go
+++ b/pkg/controllers/dynakube/injection/reconciler.go
@@ -78,6 +78,14 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 	}
 
 	if !r.dynakube.NeedAppInjection() {
+		if r.istioReconciler != nil {
+			err = r.istioReconciler.ReconcileCodeModuleCommunicationHosts(ctx, r.dynakube)
+
+			if err != nil {
+				log.Info("Error reconciling Istio codemodules", "error", err)
+			}
+		}
+
 		return r.removeAppInjection(ctx)
 	}
 
--- a/pkg/controllers/dynakube/injection/reconciler.go
+++ b/pkg/controllers/dynakube/injection/reconciler.go
@@ -77,15 +77,16 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 		return err
 	}
 
-	if !r.dynakube.NeedAppInjection() {
-		if r.istioReconciler != nil {
-			err = r.istioReconciler.ReconcileCodeModuleCommunicationHosts(ctx, r.dynakube)
+	// do istio reconciliation for CodeModules here to enable cleanup of conditions
+	if r.istioReconciler != nil {
+		err = r.istioReconciler.ReconcileCodeModuleCommunicationHosts(ctx, r.dynakube)
 
-			if err != nil {
-				log.Info("Error reconciling Istio codemodules", "error", err)
-			}
+		if err != nil {
+			log.Info("Error reconciling Istio configuration for CodeModules", "error", err)
 		}
+	}
 
+	if !r.dynakube.NeedAppInjection() {
 		return r.removeAppInjection(ctx)
 	}
 
@@ -149,13 +150,6 @@ func (r *reconciler) removeAppInjection(ctx context.Context) (err error) {
 func (r *reconciler) setupOneAgentInjection(ctx context.Context) error {
 	if !r.dynakube.NeedAppInjection() {
 		return nil
-	}
-
-	if r.istioReconciler != nil {
-		err := r.istioReconciler.ReconcileCodeModuleCommunicationHosts(ctx, r.dynakube)
-		if err != nil {
-			return err
-		}
 	}
 
 	err := initgeneration.NewInitGenerator(r.client, r.apiReader, r.dynakube.Namespace).GenerateForDynakube(ctx, r.dynakube)
--- a/pkg/controllers/dynakube/injection/reconciler.go
+++ b/pkg/controllers/dynakube/injection/reconciler.go
@@ -82,7 +82,7 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 		err = r.istioReconciler.ReconcileCodeModuleCommunicationHosts(ctx, r.dynakube)
 
 		if err != nil {
-			log.Info("Error reconciling Istio configuration for CodeModules", "error", err)
+			log.Info("error reconciling istio configuration for codemodules", "error", err)
 		}
 	}
 
--- a/pkg/controllers/dynakube/injection/reconciler.go
+++ b/pkg/controllers/dynakube/injection/reconciler.go
@@ -82,7 +82,7 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 		err = r.istioReconciler.ReconcileCodeModuleCommunicationHosts(ctx, r.dynakube)
 
 		if err != nil {
-			log.Info("error reconciling istio configuration for codemodules", "error", err)
+			log.Error(err, "error reconciling istio configuration for codemodules")
 		}
 	}
 
--- a/pkg/controllers/dynakube/istio/conditions.go
+++ b/pkg/controllers/dynakube/istio/conditions.go
@@ -6,14 +6,14 @@ import (
 )
 
 func getConditionTypeName(component string) string {
-	return "IstioServiceConfigurationFor" + component
+	return "IstioFor" + component
 }
 
 func setServiceEntryUpdatedConditionForComponent(conditions *[]metav1.Condition, component string) {
 	condition := metav1.Condition{
 		Type:    getConditionTypeName(component),
 		Status:  metav1.ConditionTrue,
-		Reason:  "IstioServiceConfigurationFor" + component + "Changed",
+		Reason:  "IstioFor" + component + "Changed",
 		Message: "ServiceEntries and VirtualServices for " + component + " have been configured.",
 	}
 	_ = meta.SetStatusCondition(conditions, condition)
@@ -21,9 +21,9 @@ func setServiceEntryUpdatedConditionForComponent(conditions *[]metav1.Condition,
 
 func setServiceEntryFailedConditionForComponent(conditions *[]metav1.Condition, component string, err error) {
 	condition := metav1.Condition{
-		Type:    "IstioServiceConfigurationFor" + component,
+		Type:    getConditionTypeName(component),
 		Status:  metav1.ConditionFalse,
-		Reason:  "IstioServiceConfigurationFor" + component + "Failed",
+		Reason:  "IstioFor" + component + "Failed",
 		Message: "Failed to configure Istio ServiceEntries and VirtualServices for " + component + " with error: " + err.Error(),
 	}
 	_ = meta.SetStatusCondition(conditions, condition)
--- a/pkg/controllers/dynakube/istio/conditions.go
+++ b/pkg/controllers/dynakube/istio/conditions.go
@@ -2,32 +2,31 @@ package istio
 
 import (
 	"fmt"
-	"strings"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func createConditionTypeForComponent(component string) string {
-	return fmt.Sprintf("IstioServiceConfigurationFor%s", strings.ToTitle(component))
+func getConditionTypeName(component string) string {
+	return "IstioServiceConfigurationFor" + component
 }
 
 func setServiceEntryUpdatedConditionForComponent(conditions *[]metav1.Condition, component string) {
 	condition := metav1.Condition{
-		Type:    createConditionTypeForComponent(component),
+		Type:    getConditionTypeName(component),
 		Status:  metav1.ConditionTrue,
-		Reason:  fmt.Sprintf("IstioServiceConfigurationFor%sChanged", strings.ToTitle(component)),
-		Message: fmt.Sprintf("ServiceEntries and VirtualServices for %s have been configured.", strings.ToTitle(component)),
+		Reason:  "IstioServiceConfigurationFor" + component + "Changed",
+		Message: "ServiceEntries and VirtualServices for " + component + " have been configured.",
 	}
 	_ = meta.SetStatusCondition(conditions, condition)
 }
 
 func setServiceEntryFailedConditionForComponent(conditions *[]metav1.Condition, component string, err error) {
 	condition := metav1.Condition{
-		Type:    createConditionTypeForComponent(component),
+		Type:    "IstioServiceConfigurationFor" + component,
 		Status:  metav1.ConditionFalse,
-		Reason:  fmt.Sprintf("ServiceEntryFor%sFailed", strings.ToTitle(component)),
-		Message: fmt.Sprintf("Failed to configure Istio ServiceEntries and VirtualServices for %s with error: %s", strings.ToTitle(component), err.Error()),
+		Reason:  "IstioServiceConfigurationFor" + component + "Failed",
+		Message: fmt.Sprintf("Failed to configure Istio ServiceEntries and VirtualServices for %s with error: %s", component, err.Error()),
 	}
 	_ = meta.SetStatusCondition(conditions, condition)
 }
--- a/pkg/controllers/dynakube/istio/conditions.go
+++ b/pkg/controllers/dynakube/istio/conditions.go
@@ -1,8 +1,6 @@
 package istio
 
 import (
-	"fmt"
-
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -26,7 +24,7 @@ func setServiceEntryFailedConditionForComponent(conditions *[]metav1.Condition, 
 		Type:    "IstioServiceConfigurationFor" + component,
 		Status:  metav1.ConditionFalse,
 		Reason:  "IstioServiceConfigurationFor" + component + "Failed",
-		Message: fmt.Sprintf("Failed to configure Istio ServiceEntries and VirtualServices for %s with error: %s", component, err.Error()),
+		Message: "Failed to configure Istio ServiceEntries and VirtualServices for " + component + " with error: " + err.Error(),
 	}
 	_ = meta.SetStatusCondition(conditions, condition)
 }
--- a/pkg/controllers/dynakube/istio/conditions.go
+++ b/pkg/controllers/dynakube/istio/conditions.go
@@ -1,0 +1,44 @@
+package istio
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func createConditionTypeForComponent(component string) string {
+	return fmt.Sprintf("IstioServiceConfigurationFor%s", strings.ToTitle(component))
+}
+
+func setServiceEntryUpdatedConditionForComponent(conditions *[]metav1.Condition, component string) {
+	condition := metav1.Condition{
+		Type:    createConditionTypeForComponent(component),
+		Status:  metav1.ConditionTrue,
+		Reason:  fmt.Sprintf("IstioServiceConfigurationFor%sChanged", strings.ToTitle(component)),
+		Message: fmt.Sprintf("ServiceEntries and VirtualServices for %s have been configured.", strings.ToTitle(component)),
+	}
+	_ = meta.SetStatusCondition(conditions, condition)
+}
+
+func setServiceEntryFailedConditionForComponent(conditions *[]metav1.Condition, component string, err error) {
+	condition := metav1.Condition{
+		Type:    createConditionTypeForComponent(component),
+		Status:  metav1.ConditionFalse,
+		Reason:  fmt.Sprintf("ServiceEntryFor%sFailed", strings.ToTitle(component)),
+		Message: fmt.Sprintf("Failed to configure Istio ServiceEntries and VirtualServices for %s with error: %s", strings.ToTitle(component), err.Error()),
+	}
+	_ = meta.SetStatusCondition(conditions, condition)
+}
+
+//
+// func setIPServiceEntryForActiveGateFailedCondition(conditions *[]metav1.Condition, err error) {
+// 	condition := metav1.Condition{
+// 		Type:    serviceEntryForActiveGateConditionType,
+// 		Status:  metav1.ConditionFalse,
+// 		Reason:  ipServiceEntryForActiveGateFailedReason,
+// 		Message: "Failed to create an IP ServiceEntry for ActiveGate, error: " + err.Error(),
+// 	}
+// 	_ = meta.SetStatusCondition(conditions, condition)
+// }
--- a/pkg/controllers/dynakube/istio/conditions.go
+++ b/pkg/controllers/dynakube/istio/conditions.go
@@ -28,14 +28,3 @@ func setServiceEntryFailedConditionForComponent(conditions *[]metav1.Condition, 
 	}
 	_ = meta.SetStatusCondition(conditions, condition)
 }
-
-//
-// func setIPServiceEntryForActiveGateFailedCondition(conditions *[]metav1.Condition, err error) {
-// 	condition := metav1.Condition{
-// 		Type:    serviceEntryForActiveGateConditionType,
-// 		Status:  metav1.ConditionFalse,
-// 		Reason:  ipServiceEntryForActiveGateFailedReason,
-// 		Message: "Failed to create an IP ServiceEntry for ActiveGate, error: " + err.Error(),
-// 	}
-// 	_ = meta.SetStatusCondition(conditions, condition)
-// }
--- a/pkg/controllers/dynakube/istio/reconciler.go
+++ b/pkg/controllers/dynakube/istio/reconciler.go
@@ -269,17 +269,10 @@ func (r *reconciler) reconcileFQDNServiceEntry(ctx context.Context, fqdnHosts []
 func (r *reconciler) cleanupFQDNServiceEntry(ctx context.Context, component string) error {
 	entryName := BuildNameForFQDNServiceEntry(r.client.Owner.GetName(), component)
 
-	err := r.client.DeleteServiceEntry(ctx, entryName)
-	if err != nil {
-		return err
-	}
+	errServiceEntry := r.client.DeleteServiceEntry(ctx, entryName)
+	errVirtualService := r.client.DeleteVirtualService(ctx, entryName)
 
-	err = r.client.DeleteVirtualService(ctx, entryName)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return goerrors.Join(errServiceEntry, errVirtualService)
 }
 
 func buildObjectMeta(name, namespace string, labels map[string]string) metav1.ObjectMeta {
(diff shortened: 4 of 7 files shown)


pkg/controllers/dynakube/istio/conditions.go
```diff
@@ -0,0 +1,30 @@
+package istio
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func getConditionTypeName(component string) string {
+	return "IstioServiceConfigurationFor" + component
```
0sewa0, line 9: this feels a bit too verbose, my suggestions for alternatives:
- `<component>IstioConfig`
- `IstioFor<component>`

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -62,11 +63,21 @@ func (r *reconciler) ReconcileCodeModuleCommunicationHosts(ctx context.Context,
 
 	oneAgentCommunicationHosts := oaconnectioninfo.GetCommunicationHosts(dynakube)
 
+	const conditionComponent = "CodeModule"
```
0sewa0, line 66: couldn't we reuse the `OneAgentComponent` const here? 
I know that technically `CodeModule` is more accurate

but maybe we could change the `OneAgentComponent` to be more accurate
- Only do this if it does not break any naming

otherwise we could still just reuse `OneAgentComponent` and give it the old `strings.ToTitle()` to make it more "presentable" 😅

wepudt: I actually did that. Turns out `toTitle(..)` uppercases all letters, not just the initial one. :(

Reusing the existing constants would either mean to 
1. have the component part all lowercase, 
2. or all uppercase (using `toTitle(...)`)
3. or do some string-voodoo to just uppercase the initial rune.
4. 
possibilities 1 and 2 seemed bad in respect to readability, 3rd one seemed a bit over-engineered so the simplest solution I could think of was to use a constant... (changing the existing one seemed the worse idea as it would break exisiting behaviour)

pkg/controllers/dynakube/istio/conditions.go
wepudt: fixed in #e131fa0d

pkg/controllers/dynakube/istio/reconciler.go
0sewa0: Did a bit of research, and yes the `ToTitle()` func is confusing AF.
There is a `.Title()` func that does kinda what I want. (makes only the first char in a word upper case) 
But even then, we would end up with the `Codemodule` or `Oneagent`

 so lets leave it as is, and do some refactoring incase in the future it will confuse us

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -83,11 +94,21 @@ func (r *reconciler) ReconcileActiveGateCommunicationHosts(ctx context.Context,
...
 	if err != nil {
+		setServiceEntryFailedConditionForComponent(dynakube.Conditions(), conditionComponent, err)
+
 		return err
 	}
 
+	if len(activeGateEndpoints) > 0 {
+		setServiceEntryUpdatedConditionForComponent(dynakube.Conditions(), conditionComponent)
+	} else {
+		meta.RemoveStatusCondition(dynakube.Conditions(), getConditionTypeName(conditionComponent))
```
0sewa0, line 109: one of the Acceptance Criteria in the linked ticket:
- `condition and istio resources are cleaned up, when not necessary`

you do a clean up here for the condition (and in line 107, the istio stuff is also cleaned up)

but in line 91, we do not clean up, if the activeGate was "turned off", and we should

(also the log line in line 88 is incorrect, but that is a mistake done by someone else 😅 )

the clean-up should also be double-checked in the `ReconcileCodeModuleCommunicationHosts` func, but maybe it just works 🤔

wepudt: fixed in 2d824c6d

I think that this is no issue for the CodeModules as the conditions are 
* set to failed if there were errors
* set to updated if there are CommunicationHosts where the ServiceEntries were updated
* removed if there are none

0sewa0: the `if !dynakube.NeedsActiveGate` check ([line 93-97](https://github.com/Dynatrace/dynatrace-operator/pull/3198/files#diff-6dcf84623910535feca0c7dc62e7a35bd7706001d348bc3c88ba5f69893b100cL80-L82)) will stop the cleanup of the istio resources we created for the ActiveGate, as the clean up happens deeper in the `r.reconcileCommunicationHostsForComponent` :
- [here](https://github.com/Dynatrace/dynatrace-operator/blob/e93ae0ad8f1a0552cfce5ff7ee5818809fe3e6d8/pkg/controllers/dynakube/istio/reconciler.go#L150-L157) and [here](https://github.com/Dynatrace/dynatrace-operator/blob/e93ae0ad8f1a0552cfce5ff7ee5818809fe3e6d8/pkg/controllers/dynakube/istio/reconciler.go#L184-L194)

we don't have a similar check for the CodeModules, so those get cleaned up (probably, only looked at code, didn't run tests)

the most "efficient" way to handle the creation/update/delete with the help of the condition: (we already do this for 1 or 2 secrets)
1. `if needed && conditionNotSet` => Create resources and set condition
2. `if needed && conditionPresent && conditionUpToDate` => do nothing
3. `if needed && conditionPresent && conditionNOTUpToDate` => update resources and condtion
4. `if notNeeded && conditionPresent` => cleanup resources and condition
5. `if notNeeded && conditionNotPresent` => do nothing

(it looks something like [this](https://github.com/Dynatrace/dynatrace-operator/blob/290c4dac9064a4c97be150f373c8e39c7dcd6aed/pkg/controllers/dynakube/connectioninfo/activegate/reconciler.go#L43-L69), its not just a bunch of `ifs`, more of a flow)

wepudt: fixed in ecac0abd

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -62,35 +67,89 @@ func (r *reconciler) ReconcileCodeModuleCommunicationHosts(ctx context.Context,
...
+	// try to clean up all entries even if one fails
+	if err1 != nil {
+		return err1
+	}
+
+	if err2 != nil {
+		return err2
+	}
+
 	return nil
```
0sewa0, line 144:
```suggestion
	return errors.Join(err1, err2)
```

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -62,35 +67,89 @@ func (r *reconciler) ReconcileCodeModuleCommunicationHosts(ctx context.Context,
...
+		if isIstioConfigured(dynakube, conditionComponent) {
+			log.Info("ActiveGate disabled, cleaning up")
+
+			return r.CleanupIstio(ctx, dynakube, conditionComponent)
+		} else {
+			return nil
+		}
+	}
+
+	if !conditions.IsOutdated(r.timeProvider, dynakube, getConditionTypeName(conditionComponent)) {
```
0sewa0, line 107: I like this, could we do this for the ` ReconcileCodeModuleCommunicationHosts`?

you should just need to change the `if` to `!dk.NeedAppInjection()`

pkg/controllers/dynakube/istio/reconciler.go
wepudt: fixed in f30e0998

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -2,14 +2,18 @@ package istio
...
 
 	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta1/dynakube"
 	dtclient "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/connectioninfo/activegate"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/connectioninfo/oneagent"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/conditions"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/labels"
-	"github.com/pkg/errors"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/timeprovider"
+	errorshelper "github.com/pkg/errors"
```
0sewa0, line 15: pretty sure you can just remove this and use `"errors"` instead

0sewa0: the `WithMessage` func is missing in `"errors"` 😭

wepudt: that's why I just left it in... we decided to postpone the change to a different story, but after doing some fast research I found, that there are a bunch of libs, each with advantages and disadvantages. The one we use is pretty simple so I would just keep it for now...

0sewa0, line 15: just to be consistent:

we tend to alias the `errors` to `goerrors`
and leave `github.com/pkg/errors` as `errors`

https://github.com/Dynatrace/dynatrace-operator/blob/7d61cb9fed001f98c64157b193f4bc25eca742cf/pkg/controllers/dynakube/controller.go#L5

wepudt: fixed in cad3eb41

wepudt:
> pls rebase changes to main to resolve the conflict
> 
> tried it out, and the cleanup does not work because `ReconcileCodeModuleCommunicationHosts` and `ReconcileActiveGateCommunicationHosts` are called in block that checks `if featuresIsNeeded`, which is not necessary.
> 
> * https://github.com/Dynatrace/dynatrace-operator/blob/7d61cb9fed001f98c64157b193f4bc25eca742cf/pkg/controllers/dynakube/activegate/reconciler.go
> * https://github.com/Dynatrace/dynatrace-operator/blob/7d61cb9fed001f98c64157b193f4bc25eca742cf/pkg/controllers/dynakube/injection/reconciler.go

fixed in cad3eb41

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -78,6 +78,14 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
 	if !r.dynakube.NeedAppInjection() {
+		if r.istioReconciler != nil {
+			err = r.istioReconciler.ReconcileCodeModuleCommunicationHosts(ctx, r.dynakube)
+
+			if err != nil {
+				log.Info("Error reconciling Istio codemodules", "error", err)
+			}
+		}
+
 		return r.removeAppInjection(ctx)
```
0sewa0, line 89: why are you only reconciling istio when no app injection is needed? -> this will cause to only clean up things, but never create them

you should try to reconcile in every case so we create/update when we need to or cleanup

wepudt: actually this is due to the way we do reconciliation.

I wanted to have minimal invasive changes. Before we had a short to just check if AppInjection is disabled here.

The actual reconciliation is done way down in `setupOneAgentInjection` but that never got called due to the if further up...

0sewa0: moving it from `setupOneAgentInjection` to here without the is IMO less invasive than having it 2 times in different places.

luhi-DT: Also its checked twice now, here and in the function you check for `!r.dynakube.NeedAppInjection()`

wepudt: fixed in 94a715a4

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -157,40 +221,62 @@ func (r *reconciler) reconcileIPServiceEntry(ctx context.Context, ipHosts []dtcl
...
+	return nil
+}
+
 func (r *reconciler) reconcileFQDNServiceEntry(ctx context.Context, fqdnHosts []dtclient.CommunicationHost, component string) error {
 	owner := r.client.Owner
 	entryName := BuildNameForFQDNServiceEntry(owner.GetName(), component)
 
 	if len(fqdnHosts) != 0 {
-		meta := buildObjectMeta(
+		objectMeta := buildObjectMeta(
```
andriisoldatenko, line 240: out of curiosity, why do we need this change here? it seems unrelated to this PR.

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -77,6 +77,15 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 		return err
 	}
 
+	// do istio reconciliation for CodeModules here to enable cleanup of conditions
+	if r.istioReconciler != nil {
+		err = r.istioReconciler.ReconcileCodeModuleCommunicationHosts(ctx, r.dynakube)
+
+		if err != nil {
+			log.Info("Error reconciling Istio configuration for CodeModules", "error", err)
```
andriisoldatenko, line 85:
```suggestion
			log.Info("error reconciling istio configuration for codeModules", "error", err)
```

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -54,43 +60,101 @@ func (r *reconciler) ReconcileAPIUrl(ctx context.Context, dynakube *dynatracev1b
...
+			log.Info("ActiveGate disabled, cleaning up")
+
+			return r.CleanupIstio(ctx, dynakube, conditionComponent, ActiveGateComponent)
+		} else {
+			return nil
+		}
+	}
+
+	if !conditions.IsOutdated(r.timeProvider, dynakube, getConditionTypeName(conditionComponent)) {
+		log.Info("Condition still within time threshold...skipping further reconciliation")
```
andriisoldatenko, line 119:
```suggestion
		log.Info("condition still within time threshold...skipping further reconciliation")
```

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -135,20 +199,20 @@ func (r *reconciler) reconcileIPServiceEntry(ctx context.Context, ipHosts []dtcl
 	entryName := BuildNameForIPServiceEntry(owner.GetName(), component)
 
 	if len(ipHosts) != 0 {
-		meta := buildObjectMeta(
+		objectMeta := buildObjectMeta(
```
andriisoldatenko, line 202: same as above.

pkg/controllers/dynakube/istio/reconciler.go
wepudt: true, but I got the hint `Variable 'meta' collides with imported package name` so I thought it might improve readability...

andriisoldatenko: okey, it should work.

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -157,40 +221,62 @@ func (r *reconciler) reconcileIPServiceEntry(ctx context.Context, ipHosts []dtcl
...
+
+	err := r.client.DeleteServiceEntry(ctx, entryName)
+	if err != nil {
+		return err
+	}
+
+	err = r.client.DeleteVirtualService(ctx, entryName)
+	if err != nil {
+		return err
 	}
```
andriisoldatenko, line 280: here if you was not able to delete SE, we never try to delete VS. May be you should use same trick you did with errors.Join?
```
package main

import (
	"errors"
	"fmt"
)

func main() {
	err1 := errors.New("err1")
	var err2 error
	err := errors.Join(err1, err2)
	fmt.Println(err)
	if errors.Is(err, err1) {
		fmt.Println("err is err1")
	}
}
```

pkg/controllers/dynakube/istio/reconciler.go
wepudt: fixed in dc69fe66

pkg/controllers/dynakube/injection/reconciler.go
wepudt: fixed in dc69fe66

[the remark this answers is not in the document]
wepudt: fixed in dc69fe66

pkg/controllers/dynakube/istio/reconciler.go
wepudt: I like it, should improve cleanup.

wepudt: fixed in c09681fc

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -157,37 +221,43 @@ func (r *reconciler) reconcileIPServiceEntry(ctx context.Context, ipHosts []dtcl
...
 
+func (r *reconciler) cleanupIPServiceEntry(ctx context.Context, component string) error {
+	entryName := BuildNameForIPServiceEntry(r.client.Owner.GetName(), component)
+
+	err := r.client.DeleteServiceEntry(ctx, entryName)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
andriisoldatenko, line 232: this code can be simplified, you are doing nothing with err. so
```
return r.client.DeleteServiceEntry(ctx, entryName)
```

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -54,43 +60,101 @@ func (r *reconciler) ReconcileAPIUrl(ctx context.Context, dynakube *dynatracev1b
...
 	}
 
 	if !dynakube.NeedsActiveGate() {
+		if isIstioConfigured(dynakube, conditionComponent) {
+			log.Info("activegate disabled, cleaning up")
+
+			return r.CleanupIstio(ctx, dynakube, conditionComponent, ActiveGateComponent)
+		} else {
+			return nil
+		}
```
andriisoldatenko, line 115:
```suggestion
		if isIstioConfigured(dynakube, conditionComponent) {
			log.Info("activegate disabled, cleaning up")

			return r.CleanupIstio(ctx, dynakube, conditionComponent, ActiveGateComponent)
		}
		return nil
```
else is redundant.

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -54,43 +60,101 @@ func (r *reconciler) ReconcileAPIUrl(ctx context.Context, dynakube *dynatracev1b
...
 	err := r.reconcileCommunicationHostsForComponent(ctx, activeGateEndpoints, ActiveGateComponent)
 	if err != nil {
+		setServiceEntryFailedConditionForComponent(dynakube.Conditions(), conditionComponent, err)
+
 		return err
 	}
 
+	if len(activeGateEndpoints) > 0 {
+		setServiceEntryUpdatedConditionForComponent(dynakube.Conditions(), conditionComponent)
+	} else {
```
andriisoldatenko, line 135: else is redundunt here.

pkg/controllers/dynakube/activegate/reconciler.go
```diff
@@ -107,12 +107,10 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
...
 
-	if r.dynakube.NeedsActiveGate() {
-		if r.istioReconciler != nil {
-			err = r.istioReconciler.ReconcileActiveGateCommunicationHosts(ctx, r.dynakube)
-			if err != nil {
-				return err
-			}
+	if r.istioReconciler != nil {
+		err = r.istioReconciler.ReconcileActiveGateCommunicationHosts(ctx, r.dynakube)
+		if err != nil {
```
andriisoldatenko, line 112: useless error checking, just return `r.istioReconciler.ReconcileActiveGateCommunicationHosts(ctx, r.dynakube)`

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -77,6 +77,15 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 		return err
 	}
 
+	// do istio reconciliation for CodeModules here to enable cleanup of conditions
+	if r.istioReconciler != nil {
+		err = r.istioReconciler.ReconcileCodeModuleCommunicationHosts(ctx, r.dynakube)
+
+		if err != nil {
+			log.Info("error reconciling istio configuration for codemodules", "error", err)
```
andriisoldatenko, line 85:
```suggestion
			log.Error("error reconciling istio configuration for codemodules", "error", err)
```
may be instead?

pkg/controllers/dynakube/istio/conditions.go
```diff
@@ -0,0 +1,30 @@
...
+	}
+	_ = meta.SetStatusCondition(conditions, condition)
+}
+
+func setServiceEntryFailedConditionForComponent(conditions *[]metav1.Condition, component string, err error) {
+	condition := metav1.Condition{
+		Type:    getConditionTypeName(component),
+		Status:  metav1.ConditionFalse,
+		Reason:  "IstioFor" + component + "Failed",
+		Message: "Failed to configure Istio ServiceEntries and VirtualServices for " + component + " with error: " + err.Error(),
```
andriisoldatenko, line 27: @0sewa0 do we have limit lenght for message? if err is long ? nvm already found:
```
	// message is a human readable message indicating details about the transition.
	// This may be an empty string.
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MaxLength=32768
	Message string `json:"message" protobuf:"bytes,6,opt,name=message"`
```

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -54,43 +60,101 @@ func (r *reconciler) ReconcileAPIUrl(ctx context.Context, dynakube *dynatracev1b
...
 	err := r.reconcileCommunicationHostsForComponent(ctx, oneAgentCommunicationHosts, OneAgentComponent)
 	if err != nil {
+		setServiceEntryFailedConditionForComponent(dynakube.Conditions(), conditionComponent, err)
+
 		return err
 	}
 
+	if len(oneAgentCommunicationHosts) > 0 {
+		setServiceEntryUpdatedConditionForComponent(dynakube.Conditions(), conditionComponent)
+	} else {
```
andriisoldatenko, line 92: else is redundunt.

> The Go convention is that your "happy path" should be left-aligned and "exceptions" should be indented.

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -54,43 +60,101 @@ func (r *reconciler) ReconcileAPIUrl(ctx context.Context, dynakube *dynatracev1b
...
+		meta.RemoveStatusCondition(dynakube.Conditions(), getConditionTypeName(conditionComponent))
+	}
+
 	return nil
 }
 
 func (r *reconciler) ReconcileActiveGateCommunicationHosts(ctx context.Context, dynakube *dynatracev1beta2.DynaKube) error {
+	const conditionComponent = "ActiveGate"
```
andriisoldatenko, line 100: I don't get the point why we need this one, if we have `ActiveGateComponent` from pkg/controllers/dynakube/istio/config.go:16

pkg/controllers/dynakube/istio/reconciler.go
wepudt: fixed in a6858713

pkg/controllers/dynakube/istio/reconciler.go
wepudt: If I use `ActiveGateComponent` the condition would then be called `IstioForactivegate` which I don't like. 
Uppercasing the first rune is also no good solution imho- adds to code complexity for only mediocre improvement. Simplest solution imho is to use this const.

andriisoldatenko: but why not to more to all constants? so later we can refactor it. BTW we use ActiveGateComponent only in 1 place. So it makes sense to tweak it and then just apply strings.Lower(ActiveGateComponent) in 2 places.

pkg/controllers/dynakube/istio/reconciler.go
wepudt: fixed in c4c2371a

pkg/controllers/dynakube/injection/reconciler.go
wepudt: fixed in c4c2371a

pkg/controllers/dynakube/activegate/reconciler.go
wepudt: fixed in c4c2371a

pkg/controllers/dynakube/istio/reconciler.go
wepudt: fixed in c4c2371a

pkg/controllers/dynakube/injection/reconciler.go
andriisoldatenko: this commit hash about different change.

pkg/controllers/dynakube/istio/reconciler.go
andriisoldatenko: wrong commit hash.

pkg/controllers/dynakube/istio/reconciler.go
andriisoldatenko: so something like it:
```diff
--- pkg/controllers/dynakube/istio/config.go
+++ pkg/controllers/dynakube/istio/config.go
@@ -13,7 +13,7 @@ var (
 const (
        OperatorComponent   = "operator"
        OneAgentComponent   = "oneagent"
-       ActiveGateComponent = "activegate"
+       ActiveGateComponent = "ActiveGate"
        IstioGVRName        = "networking.istio.io"
        IstioGVRVersion     = "v1beta1"
 )
```


```diff
 func (r *reconciler) ReconcileActiveGateCommunicationHosts(ctx context.Context, dynakube *dynatracev1beta2.DynaKube) error {
-       const conditionComponent = "ActiveGate"
-
        log.Info("reconciling istio components for activegate communication hosts")

        if dynakube == nil {
@@ -106,16 +105,16 @@ func (r *reconciler) ReconcileActiveGateCommunicationHosts(ctx context.Context,
        }

        if !dynakube.NeedsActiveGate() {
-               if isIstioConfigured(dynakube, conditionComponent) {
+               if isIstioConfigured(dynakube, ActiveGateComponent) {
                        log.Info("ActiveGate disabled, cleaning up")

-                       return r.CleanupIstio(ctx, dynakube, conditionComponent, ActiveGateComponent)
+                       return r.CleanupIstio(ctx, dynakube, ActiveGateComponent, strings.ToLower(ActiveGateComponent))
                } else {
                        return nil
                }
        }

-       if !conditions.IsOutdated(r.timeProvider, dynakube, getConditionTypeName(conditionComponent)) {
+       if !conditions.IsOutdated(r.timeProvider, dynakube, getConditionTypeName(ActiveGateComponent)) {
                log.Info("Condition still within time threshold...skipping further reconciliation")

                return nil
@@ -123,17 +122,17 @@ func (r *reconciler) ReconcileActiveGateCommunicationHosts(ctx context.Context,

        activeGateEndpoints := activegate.GetEndpointsAsCommunicationHosts(dynakube)

-       err := r.reconcileCommunicationHostsForComponent(ctx, activeGateEndpoints, ActiveGateComponent)
+       err := r.reconcileCommunicationHostsForComponent(ctx, activeGateEndpoints, strings.ToLower(ActiveGateComponent))
        if err != nil {
-               setServiceEntryFailedConditionForComponent(dynakube.Conditions(), conditionComponent, err)
+               setServiceEntryFailedConditionForComponent(dynakube.Conditions(), ActiveGateComponent, err)

                return err
```

pkg/controllers/dynakube/injection/reconciler.go
wepudt: sorry my bad 76f4aaa0

pkg/controllers/dynakube/istio/reconciler.go
wepudt: copy paste error 😞 76f4aaa0

pkg/controllers/dynakube/istio/reconciler.go
wepudt: done in d56e6ba3

wepudt: done in a9c0115d1d132628574371076b3627808106ae13

pkg/controllers/dynakube/activegate/reconciler.go
albertogdd: Yep, Werner is right. The function continues after this.

albertogdd (commented):
Code LGTM, but I deployed it in my cluster with CloudNativeFullStack and Istio and I didn't find the conditions in the Dynakube :thinking: But I think is an issue from my Istio setup

aorcholski (approved):

aorcholski (approved):
LGTM. I see new conditions, appropriete condition disappeared after update of dynakube (AG removed)

0sewa0 (approved):

albertogdd:
Checked it again this morning and it works :+1:

PR #3199, opened 2024-05-24
Consider all namespaceselectors in the Pod webhook

waodim:
## Description

Corresponding ticket: https://dt-rnd.atlassian.net/browse/K8S-9333

With this the newly added (or more presicely: distributed) namespaceselectors are considered when in the Pod webhook. the namespaces that are stored in the request are checked against the namespaceselector that can be set in the DynaKube. 

## How can this be tested?

Deploy the operator with e DynaKube with a namespaceselector. Create namespaces that match the namespaceselector but also some that do not match it. Inject into those namespaces and check if everything works accordingly


pkg/webhook/mutation/pod/metadata/mutator.go
```diff
@@ -36,7 +38,19 @@ func (mut *Mutator) Enabled(request *dtwebhook.BaseRequest) bool {
 		request.DynaKube.FeatureAutomaticInjection())
 	enabledOnDynakube := request.DynaKube.MetaDataEnrichmentEnabled()
 
-	return enabledOnPod && enabledOnDynakube
+	matchesNamespace := true // if no namespace selector is configured, we just pass set this to true
+
+	if enabledOnDynakube && request.DynaKube.MetadataEnrichmentNamespaceSelector() != nil && request.DynaKube.MetadataEnrichmentNamespaceSelector().Size() > 0 {
```
0sewa0, line 43:
```suggestion
	if request.DynaKube.MetadataEnrichmentNamespaceSelector().Size() > 0 {
```
wouldn't this be enough?
the first line in the `.Size()` implementation has a nil check in it.

you can write a test for that

checking the `enabledOnDynaKube` feels like a bit of optimisation that probably is not worth it, because to me its much cleaner to define each condition `matchesNamespace, enabledOnPod, enabledOnDynakube` one by one, and at the end "combine them" accordingly
- you already check the `enabledOnDynakube` at the end, so doing it here is not fully necessary

pkg/api/v1beta2/dynakube/properties.go
```diff
@@ -432,6 +432,10 @@ func (dk *DynaKube) MetaDataEnrichmentEnabled() bool {
 	return dk.Spec.MetadataEnrichment.Enabled
 }
 
+func (dk *DynaKube) MetadataEnrichmentNamespaceSelector() *metav1.LabelSelector {
+	return &dk.Spec.MetadataEnrichment.NamespaceSelector
+}
```
0sewa0, line 437: do we have to return a pointer? what do we gain by that?

we can easily check if its unset/empty with `.Size()`

0sewa0: (if we use a pointer for the `OneAgentNamespaceSelector` then we are dumb, and should fix that at some point 😛 )

waodim: Indeed we return a pointer. for OneAgentNamespaceSelector. Should I update this within this PR? Wdyt?

0sewa0: you can try, 
BUT if it introduces way too many changes then just leave it, lets not blow up the PR
- We can create a ticket about it, that can be more general, and I think we might already have one, related to using the kubebuilder "marker-comments" more, we could do it there

waodim: It seems to me as though we can not call `Size()` on `MetadataEnrichmentSelector` if we do not return a pointer. Can you confirm?

pkg/webhook/mutation/pod/metadata/mutator.go
```diff
@@ -36,7 +38,19 @@ func (mut *Mutator) Enabled(request *dtwebhook.BaseRequest) bool {
...
 	enabledOnDynakube := request.DynaKube.MetaDataEnrichmentEnabled()
 
-	return enabledOnPod && enabledOnDynakube
+	matchesNamespace := true // if no namespace selector is configured, we just pass set this to true
+
+	if request.DynaKube.MetadataEnrichmentNamespaceSelector().Size() > 0 {
+		selector, err := metav1.LabelSelectorAsSelector(request.DynaKube.MetadataEnrichmentNamespaceSelector())
+
+		if err != nil {
+			matchesNamespace = false //nolint:ineffassign,wastedassign
```
chrismuellner, line 47: since nothing else is done in the error case, the assignment is "wasted" as `matchesNamespace` gets a value after the `if` anyway

waodim: adressed in https://github.com/Dynatrace/dynatrace-operator/pull/3199/commits/a4cfb344db2357b76fd158e5284e7454fef9ce10

pkg/api/v1beta2/dynakube/properties.go
0sewa0: you can't use it directly, because `.Size()` was written for a pointer receiver.

usually Golang hides this fact, and auto converts from "value" to "pointer", but it can't do it with the return value of a function.

so incase of:
```go

func (dk *DynaKube) OneAgentNamespaceSelector() metav1.LabelSelector { ... }
```
```go
dk.OneAgentNamespaceSelector().Size() // not allowed

nl := dk.OneAgentNamespaceSelector()
nl.Size() // allowed
```

0sewa0: but don't worry about it, we should fix this an other time, because I am flip-flopping about which should be the better solution

0sewa0: (reason for flip-flopping:
We store the `LabelSelector` in the CR as a "value" so `MetadataEnrichmentNamespaceSelector()` will **never** return `nil`, as an empty `LabelSelector`'s pointer is not nil.
And if its nicer to use, then maybe its better this way 😅 
)

waodim: Thanks for elaboration. Then I will let it as it is here and maybe we will create stories to fix this in the future. cc @chrismuellner

pkg/webhook/mutation/pod/oneagent/mutator.go
```diff
@@ -36,7 +38,15 @@ func NewMutator(image, clusterID, webhookNamespace string, client client.Client,
...
-	return maputils.GetFieldBool(request.Pod.Annotations, dtwebhook.AnnotationOneAgentInject, request.DynaKube.FeatureAutomaticInjection())
+	matchesNamespaceSelector := true // if no namespace selector is configured, we just pass set this to true
+
+	if request.DynaKube.OneAgentNamespaceSelector().Size() > 0 {
+		selector, _ := metav1.LabelSelectorAsSelector(request.DynaKube.OneAgentNamespaceSelector())
+
+		matchesNamespaceSelector = selector.Matches(labels.Set(request.Namespace.Labels))
+	}
+
+	return matchesNamespaceSelector && maputils.GetFieldBool(request.Pod.Annotations, dtwebhook.AnnotationOneAgentInject, request.DynaKube.FeatureAutomaticInjection())
```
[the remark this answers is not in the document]
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3199/commits/9ec656f97096459ed847c90bcf82b2cda7dbdd20

0sewa0 (dismissed):

andriisoldatenko (approved):
code LGTM

andriisoldatenko (approved):

0sewa0 (approved):

PR #3205, opened 2024-05-28
Fix failing e2e release test

waodim:
## Description

With the introduction of `v1beta2` this e2e test failed. This was because the operator was installed with using the `0.15.0` tag. This version of the operator does not know `v1beta2`, thus it failed. I fixed this by simply checking if the used version for the helm command does know `v1beta2` or not.

related ticket: https://dt-rnd.atlassian.net/browse/K8S-10110

## How can this be tested?

run `make test/e2e/release`

### DIFF
--- a/test/helpers/components/dynakube/dynakube.go
+++ b/test/helpers/components/dynakube/dynakube.go
@@ -9,6 +9,8 @@ import (
 	"time"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/status"
+	dynatracev1beta1 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta1"
+	dynakubev1beta1 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta1/dynakube"
 	dynatracev1beta2 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2"
 	dynakubev1beta2 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/test/helpers/components/oneagent"
@@ -27,9 +29,16 @@ const (
 	defaultName = "dynakube"
 )
 
-func Install(builder *features.FeatureBuilder, level features.Level, secretConfig *tenant.Secret, testDynakube dynakubev1beta2.DynaKube) {
-	Create(builder, level, secretConfig, testDynakube)
-	VerifyStartup(builder, level, testDynakube)
+func Install(builder *features.FeatureBuilder, level features.Level, secretConfig *tenant.Secret, testDynakube dynakubev1beta2.DynaKube, usesOldVersion bool) {
+	if usesOldVersion {
+		dkv1beta1 := dynakubev1beta1.DynaKube{}
+		dkv1beta1.ConvertFrom(&testDynakube)
+		CreateV1Beta1(builder, level, secretConfig, dkv1beta1)
+		VerifyStartupV1Beta1(builder, level, dkv1beta1)
+	} else {
+		Create(builder, level, secretConfig, testDynakube)
+		VerifyStartup(builder, level, testDynakube)
+	}
 }
 
 func Create(builder *features.FeatureBuilder, level features.Level, secretConfig *tenant.Secret, testDynakube dynakubev1beta2.DynaKube) {
@@ -44,6 +53,26 @@ func Create(builder *features.FeatureBuilder, level features.Level, secretConfig
 
 func Update(builder *features.FeatureBuilder, level features.Level, testDynakube dynakubev1beta2.DynaKube) {
 	builder.WithStep("dynakube updated", level, update(testDynakube))
+}
+
+func CreateV1Beta1(builder *features.FeatureBuilder, level features.Level, secretConfig *tenant.Secret, testDynakube dynakubev1beta1.DynaKube) {
+	if secretConfig != nil {
+		builder.WithStep("created tenant secret", level, tenant.CreateTenantSecret(*secretConfig, testDynakube.Name, testDynakube.Namespace))
+	}
+	builder.WithStep(
+		fmt.Sprintf("'%s' dynakube created", testDynakube.Name),
+		level,
+		createv1beta1(testDynakube))
+}
+
+func VerifyStartupV1Beta1(builder *features.FeatureBuilder, level features.Level, testDynakube dynakubev1beta1.DynaKube) {
+	if testDynakube.NeedsOneAgent() {
+		builder.WithStep("oneagent started", level, oneagent.WaitForDaemonsetV1Beta1(testDynakube))
+	}
+	builder.WithStep(
+		fmt.Sprintf("'%s' dynakube phase changes to 'Running'", testDynakube.Name),
+		level,
+		WaitForPhaseV1Beta1(testDynakube, status.Running))
 }
 
 func Delete(builder *features.FeatureBuilder, level features.Level, testDynakube dynakubev1beta2.DynaKube) {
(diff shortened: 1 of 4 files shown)


test/scenarios/release/release_test.go
```diff
@@ -5,22 +5,27 @@ package release
...
 	"github.com/Dynatrace/dynatrace-operator/test/helpers/tenant"
+	"golang.org/x/mod/semver"
 	"sigs.k8s.io/e2e-framework/pkg/env"
 	"sigs.k8s.io/e2e-framework/pkg/features"
 )
 
 var testEnv env.Environment
+var thresholdVersion, _ = dtversion.ToSemver("1.1.0")
+
+const usedVersion = "0.15.0"
```
andriisoldatenko, line 21: hmm... any reason why do we use old version for running e2e tests?

test/features/cloudnative/codemodules/codemodules.go
```diff
@@ -98,7 +98,7 @@ func InstallFromImage(t *testing.T) features.Feature {
 	builder.Assess("create sample namespace", sampleApp.InstallNamespace())
 
 	// Register dynakube install
-	dynakube.Install(builder, helpers.LevelAssess, &secretConfigs[0], cloudNativeDynakube)
+	dynakube.Install(builder, helpers.LevelAssess, &secretConfigs[0], cloudNativeDynakube, false)
```
andriisoldatenko, line 101: I don't understand this change, if we never use `true`, why do we need it?

waodim: it is used in the release e2e test, where I check if we use a old version which does not know v1beta2

waodim: I am not sure how to do that differently. In the case we use a version that does not know v1beta2 we convert our dynakube to v1beta1 and use v1beta1 then instead of v1beta2

test/scenarios/release/release_test.go
chrismuellner: We should bump this version as the test should just test the upgrade from previous to latest version

andriisoldatenko: ☝️  that was my point as well

waodim: Fair enough. I will bump it to 1.0.0 then I guess?

wepudt (dismissed):
tests pass, but the solution will have to be adapted based on the version we use and whether this version uses CRD v1beta1 or v1beta2, I guess.

test/helpers/components/dynakube/dynakube.go
```diff
@@ -27,9 +29,16 @@ const (
...
+	if usesOldVersion {
+		dkv1beta1 := dynakubev1beta1.DynaKube{}
+		dkv1beta1.ConvertFrom(&testDynakube)
+		CreateV1Beta1(builder, level, secretConfig, dkv1beta1)
+		VerifyStartupV1Beta1(builder, level, dkv1beta1)
+	} else {
+		Create(builder, level, secretConfig, testDynakube)
+		VerifyStartup(builder, level, testDynakube)
+	}
 }
```
0sewa0, line 42: you could just introduce an `InstallAsOld` or something, so you only change a single package and not every

```
func InstallAsOld(builder *features.FeatureBuilder, level features.Level, secretConfig *tenant.Secret, testDynakube dynakubev1beta2.DynaKube) {
	dkv1beta1 := dynakubev1beta1.DynaKube{}
	dkv1beta1.ConvertFrom(&testDynakube)
	CreateV1Beta1(builder, level, secretConfig, dkv1beta1)
	VerifyStartupV1Beta1(builder, level, dkv1beta1)
}
```

The name `InstallAsOld` is not my greatest, so pls try to change it 😅

0sewa0 (changes requested):

test/helpers/components/dynakube/dynakube.go
waodim: I had this idea also yesterday in the evening, on it :+1:

waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3205/commits/0ce7b62c82226527c1db1832f2fb808d8d81983a

Please give me feedback for the naming 😄

waodim:
@0sewa0 @andriisoldatenko @wepudt 

Guys I did some changes:

- i removed the releaseTag out of the helm install command, i did this because then the helm command takes the latest stable released version and installs the operator
- I added logic to get our latest tag from the operator and check if this version knows v1beta2
- I used 1.2.0 as threshold version, as this will be the first version that knows v1beta2.

Please give me feedback what you think about those changes.

@chrismuellner gave me the hint that helm always gets the latest version when no specific version is set.

test/scenarios/release/release_test.go
```diff
@@ -30,8 +34,19 @@ func TestMain(m *testing.M) {
...
+	latestTag, err := cmd.CombinedOutput()
+	if err != nil {
+		t.Fatalf("Failed to run 'git describe': %v", err)
+	}
+
+	usesOldVersion := false
+	usedSemVer, _ := dtversion.ToSemver(string(latestTag))
+	if semver.Compare(thresholdVersion, usedSemVer) >= 1 {
+		usesOldVersion = true
+	}
```
0sewa0, line 47: will this work in the CI?

if we are still running the things on Concourse in the background then we do not have `git` when we run this.
- Concourse just gets the content of the branch/tag, but doesn't do a full `git clone`, so we can't use git
  - If we could use `git` then we would have done this from the beginning

andriisoldatenko: As we run it using the new Github Action Runner maybe you try to read it from envy or so.

@0sewa0 wdyt if we just bump and hardcode version again?

0sewa0: as this supposed to be a bugfix, I would actually prefer it, we should revisit the "get the version dynamically" problem in a separate ticket

andriisoldatenko: @waodim  ☝️

waodim: Got it. I will go back to the previous approach.

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/3205/commits/d771f4da34f9527c3661b0cfb31d886a11db4bbf

test/scenarios/release/release_test.go
```diff
@@ -30,8 +35,13 @@ func TestMain(m *testing.M) {
 }
 
 func TestRelease(t *testing.T) {
+	usesOldVersion := false
+	usedSemVer, _ := dtversion.ToSemver(releaseTag)
+	if semver.Compare(thresholdVersion, usedSemVer) >= 1 {
+		usesOldVersion = true
+	}
```
0sewa0, line 42: we are comparing basically 2 `const` here, where the outcome is always the same, why are we doing this? 😅 

is there a case where the "upgrade" test scenario will become a "downgrade" test scenario? (I hope not 😅 )

can we determine the `usesOldVersion` is any nice way?
- meaning: will there be some check that will magically tell us if version A supports the new CR version that version B uses?
  - I don't think so, or at least not now

waodim: My thought process was: currently we are at `1.1.0`, which does not know `v1beta2`. When `1.2.0` is released we have a version that knows `v1beta2`. From then on (after we bump the `releaseTag` in the test manually, because it is hardcoded) we always use the normal install method. After that the here introduced changes become useless and may be removed. But at the moment I do not see a other way to fix this test. What do you think about this? Do you have some tips what I could change here?

0sewa0: There are multiple perspectives:
1. **Does it hurt to always check if the conversion works properly?** (ie. just always consider `usesOldVersion` as `true`)
  - IMO, it should be fine to check if the conversion is working properly
    - kinda mixes the responsibilities of the tests, which is not nice, but not horrible 
  - this results in a "dumber" logic when it comes to the e2e test code (we don't have extra checks)
2. **Automate it to the fullest:**
  - So we never have to manually update any values, it "just works" and tests the upgrade from `latest released version` to the `current branch`.
  - Your check would kinda make sense in that situation
  - For this to work the following would need to be true:
     1. We (dynamically) know what is the `latest released version`
     2. We (dynamically) know what is the `latest CR version` that the `latest released version` supports
     3. We (dynamically) know what is the `latest CR version` that the `current branch` supports
     4. We (dynamically) know what is the "oldVersion" to use to convert to 🙃 
     5. ...
3. **Partially automate it**
 - So we only have to intervene, when a new version is introduced 
 - This is what you did
   - More accurately, "would have done", if the `git` magic worked 😅 


I do not think option 2. is anyway feasible.

I like option 1, for its simplicity, but can be annoying to keep up to date

Option 3 would make sense and would be the best "middle ground" IF the `git` magic can be used.
- As of right now we do not now
- Introducing code that "might be used in the future" is something we should avoid, as it will only cause confusion.
  - If I asked anyone, 2 weeks from now, "Why do we have that code?", the likely answer would be "dunno, does not make much sense, but don't want to touch it as it works, right?"

waodim: I guess the most practical way currently is to go for option 1, then. I will go for it. Thank you for elaborating.

0sewa0: np, as mentioned before, we should look into **option 3.** separately  😉

waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3205/commits/ba567e6920800e3d0d4ab60466bb2464fe1e8390

I removed the whole logic with the check if we use an old version because if we assume it is always true we do not need that flag.

0sewa0 (approved):

luhi-DT (approved):

PR #3206, opened 2024-05-28
CRUD Dynatrace Connection Settings object on EdgeConnect reconciliation

albertogdd:
## Description

EdgeConnect client now handles **environment-api**, creating K8s Connection Settings objects in Dynatrace.

- **New field in EdgeConnect CR Status** "KubeSystemUID".
- **New endpoints**
- **New CRUD functions**
- _**ONLY ON PROVISIONER MODE && KUBERNETESAUTOMATION TRUE:**_ When applying a EdgeConnect CR, **a Connection Settings object gets created in Dynatrace**. This gets updated with the connection token rotation. E.G:
```yaml
kind: EdgeConnect
metadata:
  name: test-edgeconnect
  namespace: dynatrace
spec:
  apiServer: dib18963.dev.apps.dynatracelabs.com
  kubernetesAutomation:
    enabled: true
  hostPatterns:
    - '*.test.host.org'
  oauth:
    provisioner: true
    clientSecret: edgeconnect
    endpoint: https://sso-dev.dynatracelabs.com/sso/oauth2/token
    resource: urn:dtaccount:xxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
- On deletion, **reconciler deletes the Settings object**

## How can this be tested?

- Have a dev tenant
- Install the Hub app _"Kubernetes Automation for Workflows"_
image
- You should now have access to the _new_ App settings `https://<your-tenant>.dev.apps.dynatracelabs.com/ui/apps/dynatrace.classic.settings/ui/settings/app:dynatrace.kubernetes.connector:connection?gf=all`.
**NOTE:** This URL in dev environment tends to fail. Alternatively you can check the settings objects via [HTTP call directly to the API](https://dib18963.dev.apps.dynatracelabs.com/platform/swagger-ui/index.html?urls.primaryName=Classic%20Environment%20API%20v2%20(for%20SDK%20generation)#/Settings%20-%20Objects/getSettingsObjects) with this schema ID as parameter `app:dynatrace.kubernetes.connector:connection`
- Now, go to your K8s cluster, **deploy operator**
- **Install EdgeConnect CR** on Provisioner mode and K8S Automation true (with DT Account OAuth credentials, see example above)
- After the Pod is deployed, refresh the previous Dynatrace website. You should see the **new settings object.**
- **Delete the EdgeConnect CR**, the settings object should be removed automatically.


albertogdd:
E2E tests execution > https://github.com/Dynatrace/dynatrace-operator/actions/runs/9291083181

pkg/clients/edgeconnect/environment_settings.go
```diff
@@ -0,0 +1,159 @@
...
+
+type EnvironmentSettingValue struct {
+	Name      string `json:"name"`
+	Uid       string `json:"uid"`
+	Namespace string `json:"namespace"`
+	Token     string `json:"token"`
+}
+
+type SettingsObject struct {
+}
```
0sewa0, line 35: does this serve some purpose or was just left in by accident?

pkg/clients/edgeconnect/environment_settings.go
```diff
@@ -0,0 +1,159 @@
...
+	KubernetesConnectionScope    = "environment"
+)
+
+type EnvironmentSetting struct {
+	ObjectId      *string                 `json:"objectId"`
+	SchemaId      string                  `json:"schemaId"`
+	SchemaVersion string                  `json:"schemaVersion"`
+	Scope         string                  `json:"scope"`
+	Value         EnvironmentSettingValue `json:"value"`
+}
```
0sewa0, line 25: more of a note:

looks awfully familiar to https://github.com/Dynatrace/dynatrace-operator/blob/8b5ac38fd58335674e93353747ea430abcba36a5/pkg/clients/dynatrace/kubernetes_settings.go#L33-L39

we should look into working with the SettingsApi in a more general way if possible:
- the `kubernetes_settings.go` is rather old, so I wouldn't base it necessarily on it, but would be nice to have a more general implementation for a more general use API (if I understand the SettingsApi correctly)

pkg/controllers/edgeconnect/controller.go
```diff
@@ -132,38 +133,81 @@ func (controller *Controller) reconcileEdgeConnectDeletion(ctx context.Context,
 
 	edgeConnect.ObjectMeta.Finalizers = nil
 	if err := controller.client.Update(ctx, edgeConnect); err != nil {
-		_log.Debug("updating the EdgeConnect object failed")
+		_log.Debug("reconcile deletion: updating the EdgeConnect object failed")
```
0sewa0, line 136:
```suggestion
		_log.Debug("reconcile deletion: updating the EdgeConnect object failed, couldn't remove the finalizers")
```

pkg/controllers/edgeconnect/controller.go
```diff
@@ -132,38 +133,81 @@ func (controller *Controller) reconcileEdgeConnectDeletion(ctx context.Context,
...
 
 		return errors.WithStack(err)
 	}
 
 	edgeConnectClient, err := controller.buildEdgeConnectClient(ctx, edgeConnect)
 	if err != nil {
-		_log.Debug("building EdgeConnect client failed")
+		_log.Debug("reconcile deletion: building EdgeConnect client failed")
```
0sewa0, line 143: if you are constantly adding the same prefix (`reconcile deletion:`), it might make sense to update the `_log := log.WithValues(...)`, to have this in it instead. Like `"scenario: "deletion"` or something.

(for the first 3 logs u use the prefix, but for the rest you don't, kind inconsistent 😅 )

pkg/controllers/edgeconnect/controller.go
```diff
@@ -132,38 +133,81 @@ func (controller *Controller) reconcileEdgeConnectDeletion(ctx context.Context,
...
+			_log.Info("EdgeConnect client secret is missing")
+		}
 	default:
-		if tenantEdgeConnect.ID != edgeConnectIdFromSecret {
-			_log.Info("EdgeConnect client secret contains invalid Id")
+		{
+			if tenantEdgeConnect.ID != edgeConnectIdFromSecret {
+				_log.Info("EdgeConnect client secret contains invalid Id")
+			}
 		}
```
0sewa0, line 177: just for my understanding:

Why are we doing this?
- Extra cases for log lines, that do not do anything else?
- These log lines read more like debug logs
- At the end of the day, do we care about these in context of a deletion?

Like, the problem `"EdgeConnect client secret contains invalid Id"`, sounds like something that should never happen, and if it does then it should be handled and logged, BUT I (personally) don't care about it during deletion.

pkg/controllers/edgeconnect/controller.go
```diff
@@ -132,38 +133,81 @@ func (controller *Controller) reconcileEdgeConnectDeletion(ctx context.Context,
...
+	return edgeConnectClient.DeleteEdgeConnect(tenantEdgeConnect.ID)
+}
+
+func (controller *Controller) deleteConnectionSetting(ctx context.Context, edgeConnectClient edgeconnect.Client) error {
+	var kubeSystemNamespace corev1.Namespace
+
+	err := controller.apiReader.Get(ctx, client.ObjectKey{Name: kubeSystemNamespaceName}, &kubeSystemNamespace)
+	if err != nil {
+		return err
+	}
```
0sewa0, line 199: note for future improvement:

- you could store the `kubeSystemNamespace.UID` in the `Status` of the EdgeConnect object, so you don't always do a request for it.
  - `kubeSystemNamespace.UID` does not change overtime (AFAIK), so constantly requesting it is unnecessary

pkg/controllers/edgeconnect/controller.go
```diff
@@ -357,6 +401,21 @@ func (controller *Controller) reconcileEdgeConnectRegular(ctx context.Context, e
...
+	// 	return err
+	// }
+
+	// err = controller.createOrUpdateConnectionSetting(ctx, edgeConnectClient, edgeConnect, edgeConnectToken)
+	// if err != nil {
+	// 	_log.Debug("creating EdgeConnect connection setting failed")
+
+	// 	return err
+	// }
+
```
0sewa0, line 418: Committing commented code is "not recommended." 🙃

pkg/controllers/edgeconnect/controller.go
```diff
@@ -676,32 +741,100 @@ func (controller *Controller) createOrUpdateEdgeConnectDeployment(ctx context.Co
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (controller *Controller) createOrUpdateEdgeConnectConfigSecret(ctx context.Context, edgeConnect *edgeconnectv1alpha1.EdgeConnect) (string, string, error) {
```
0sewa0, line 817: if you are going to have multiple return values (ie. not 1 value and 1 error), AND each value does not have a distinct type, then I would recommend naming them, so its easier to understand which value is supposed to be what. (otherwise you have to sniff through the code of the function to figure it out)

```suggestion
func (controller *Controller) createOrUpdateEdgeConnectConfigSecret(ctx context.Context, edgeConnect *edgeconnectv1alpha1.EdgeConnect) (token string, hash string, err error) {
```

pkg/controllers/edgeconnect/controller_test.go
```diff
@@ -718,6 +738,18 @@ func mockNewEdgeConnectClientCreate(edgeConnectClient *edgeconnectmock.Client, h
...
+		edgeConnectClient.On("GetConnectionSetting", mock.Anything).Return(edgeconnect.EnvironmentSetting{
+			ObjectId:      &v,
+			SchemaId:      edgeconnect.KubernetesConnectionSchemaID,
+			SchemaVersion: edgeconnect.KubernetesConnectionVersion,
+			Scope:         edgeconnect.KubernetesConnectionScope,
+			Value:         edgeconnect.EnvironmentSettingValue{},
+		}, nil).Maybe()
+		edgeConnectClient.On("CreateConnectionSetting", mock.Anything).Return(nil).Maybe()
+		edgeConnectClient.On("UpdateConnectionSetting", mock.Anything).Return(nil).Maybe()
+		edgeConnectClient.On("DeleteConnectionSetting", mock.Anything).Return(nil).Maybe()
```
0sewa0, line 751: can we avoid the use of `Maybe()`, and mock the client for the exact scenario we are testing?

pkg/clients/edgeconnect/environment_settings.go
StefanHauth: Let's do this in a follow up.

pkg/clients/edgeconnect/endpoints.go
```diff
@@ -2,10 +2,30 @@ package edgeconnect
...
+func (c *client) getEnvironmentApiUrl() string {
+	return c.baseURL + "/platform/classic/environment-api/v2"
+}
+
+func (c *client) getSettingsObjectsUrl() string {
+	return c.getEnvironmentApiUrl() + "/settings/objects"
+}
+
+func (c *client) getSettingsObjectsIdUrl(objectId string) string {
+	return fmt.Sprintf("%s/%s", c.getSettingsObjectsUrl(), objectId)
```
andriisoldatenko, line 30: Neat: can we use the same style of string concatenation? 2 functions using `+`, and the last one is Sprintf.

pkg/clients/edgeconnect/environment_settings.go
andriisoldatenko: okey so @albertogdd please create a ticket before closing the thread, to not forget.

pkg/clients/edgeconnect/iface.go
```diff
@@ -16,4 +16,12 @@ type Client interface {
 
 	// GetEdgeConnects returns list of edge connects
 	GetEdgeConnects(name string) (ListResponse, error)
+
+	GetConnectionSetting(uid string) (EnvironmentSetting, error)
```
andriisoldatenko, line 20: add comment like other methods. See line 17.

pkg/controllers/edgeconnect/controller_test.go
andriisoldatenko: ☝️  very good point

pkg/clients/edgeconnect/environment_settings.go
albertogdd: Fine for me :+1: Will do

pkg/controllers/edgeconnect/controller.go
albertogdd: This is a debugging session desperation leftover, to make sure that the switch statement wasn't just running the next code line and jumping the rest of the code :laughing: I will remove it :+1:

pkg/controllers/edgeconnect/controller.go
albertogdd: Awesome suggestion! I did it [here](https://github.com/Dynatrace/dynatrace-operator/pull/3206/commits/b16b883aa01c734fe24de6173b3c79978d4035f7)

pkg/clients/edgeconnect/iface.go
albertogdd: Done [here](https://github.com/Dynatrace/dynatrace-operator/pull/3206/commits/401af9d21c6e195940056f10fb01027e4976bd81)

pkg/controllers/edgeconnect/controller_test.go
albertogdd: Done [here](https://github.com/Dynatrace/dynatrace-operator/pull/3206/commits/9ab3b9530311f58228cd4bfe687929fba222a126)

pkg/clients/edgeconnect/environment_settings.go
```diff
@@ -0,0 +1,156 @@
...
+		if item.Value.Uid == uid {
+			return item, nil
+		}
+	}
+
+	return EnvironmentSetting{}, nil
+}
+
+func (c *client) CreateConnectionSetting(es EnvironmentSetting) error {
+	log.Info("Creating connection setting", "EnvironmentSetting", es)
```
aorcholski, line 83: private part of the token is printed to log
```
{"level":"info","ts":"2024-06-04T11:35:52.309Z","logger":"edgeconnect-client","msg":"Creating connection setting","EnvironmentSetting":{"objectId":null,"schemaId":"app:dynatrace.kubernetes.connector:connection","schemaVersion":"0.1.2","scope":"environment","value":{"name":"ao-ec-provisioner","uid":"273ec656-603d-46c8-b5f5-5c47a6903dff","namespace":"dynatrace","token":"dt0e01.2Xxxx.WPyyy"}}}
```

aorcholski (changes requested):

pkg/clients/edgeconnect/environment_settings.go
albertogdd: True! Thank you.
Fixed [here](https://github.com/Dynatrace/dynatrace-operator/pull/3206/commits/6a4b8f2e72b859f69c0e43e5aecf0d9b0c615800)

pkg/controllers/edgeconnect/controller.go
```diff
@@ -676,32 +714,93 @@ func (controller *Controller) createOrUpdateEdgeConnectDeployment(ctx context.Co
 		return err
 	}
 
+	edgeConnectClient, err := controller.buildEdgeConnectClient(ctx, edgeConnect)
```
aorcholski, line 717: `createOrUpdateEdgeConnectDeployment` function is not related to Deployment any more.

maybe `createOrUpdateEdgeConnectDeploymentAndSettings` name would be better or a new wrapper function could be added to group Deployment and Settings changes together and to keep `createOrUpdateEdgeConnectDeployment` function focused on a single task

albertogdd: Agree with you :+1: 
Updated [here](https://github.com/Dynatrace/dynatrace-operator/pull/3206/commits/8628226038be2c03a62e1b3d9ed753143213ddcf)

aorcholski (approved):
LGTM

config/helm/chart/default/templates/Common/crd/dynatrace-operator-crd.yaml
```diff
@@ -5315,6 +5315,9 @@ spec:
                   - type
                   type: object
                 type: array
+              kubeSystemUid:
```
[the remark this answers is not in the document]
andriisoldatenko: do we want to stick with this name?

albertogdd:
Question that I just came up with: What would happen if the tenant doesn't have the App installed? :thinking: I will try it

StefanHauth:
> Question that I just came up with: What would happen if the tenant doesn't have the App installed? 🤔 I will try it

I think in this case, the connection object scheme is missing and the request will fail. But that's nothing that we can do anything about it, is it?

pkg/clients/edgeconnect/environment_settings.go
```diff
@@ -0,0 +1,154 @@
...
+	ObjectId      *string                 `json:"objectId"`
+	SchemaId      string                  `json:"schemaId"`
+	SchemaVersion string                  `json:"schemaVersion"`
+	Scope         string                  `json:"scope"`
+	Value         EnvironmentSettingValue `json:"value"`
+}
+
+type EnvironmentSettingValue struct {
+	Name      string `json:"name"`
+	Uid       string `json:"uid"`
```
andriisoldatenko, line 29:
```suggestion
	UID      string `json:"uid"`
```

aorcholski (changes requested):
we need different types of server error response structs for 
EdgeConnect related endpoints :
```
{
  "error": {
    "code": 0,
    "message": "string",
    "details": ...
  }
}
```
and Settings related endpoints :
```
  "error": {
    "code": 0,
    "constraintViolations": [ ...  ],
    "message": "string"
  }
```

albertogdd:
> we need different types of server error response structs for EdgeConnect related endpoints :
> 
> ```
> {
>   "error": {
>     "code": 0,
>     "message": "string",
>     "details": ...
>   }
> }
> ```
> 
> and Settings related endpoints :
> 
> ```
>   "error": {
>     "code": 0,
>     "constraintViolations": [ ...  ],
>     "message": "string"
>   }
> ```

Fixed [here](https://github.com/Dynatrace/dynatrace-operator/pull/3206/commits/3ef2a272cf32bb31841cd028c1969fead45310fc)

aorcholski (approved):
LGTM , settings object created and removed

aorcholski (approved):
LGTM

StefanHauth (commented):
LGTM

StefanHauth (approved):
LGTM

PR #2659, opened 2024-02-02
Enable the usage of verbosity levels for logging

StefanHauth:
## Description

The main purpose of this PR is to straighten our logging and make the usage of verbosity levels (aka. `log.V(n)` possible. This also includes a bit of refactoring and code cleanup. This PR intentionally does not change existing log lines or introduce lot's of debug logging. That's for later. Also the actual log output should be kept as close as possible to what we have now.

This PR is split in two commits to make reviewing a bit easier. 
* 2952d2ad8ff45cb54c380004e256e27cc4527cd3 get's rid of a useless factory pattern. This one touches lot's of files. 
* The core refactoring is in 21e6c151d569468d089462fa67551a0f3b8c5e14.

## How can this be tested?

Deploy the operator and check if logs are still arriving and are readable (at least as readable as they used to be ;-)

## Checklist

- [x] Unit tests have been updated/added
- [x] PR is labeled accordingly with a single label
- [x] I have read and understood the [contribution guidelines](https://github.com/Dynatrace/dynatrace-operator/blob/main/CONTRIBUTING.md)

PR #3208, opened 2024-05-29
Make log lines more consistent

0sewa0:
## Description

The https://github.com/Dynatrace/dynatrace-operator/pull/2659 PR introduced a bit of a problem in the logs.

Example:
- Before the change: (there is a predefined order in the keys)
```
{"level":"info","ts":"2024-05-28T14:00:03.031Z","logger":"main.controller-runtime.metrics","msg":"Serving metrics server","bindAddress":":8080","secure":false}
{"level":"info","ts":"2024-05-28T14:00:03.031Z","logger":"main","msg":"starting server","kind":"health probe","addr":":10080"}
```
- After the change: (keys are ordered alphanumerically)
```
{"bindAddress":":8080","level":"info","logger":"main.controller-runtime.metrics","msg":"Serving metrics server","secure":false,"ts":"2024-05-28T14:05:53.774Z"}
{"addr":":10080","level":"info","logger":"main","msg":"starting server","name":"health probe","ts":"2024-05-28T14:05:53.774Z"}
``` 

The source of the problem is that we used to send error to `stderr` and everything else to `stdout`, and we had a different `io.Writer` for each.
- We didn't use to touch `stdout`
- We only prettified the `stderr` input, to have nicer stacktraces
   - these logs probably always had this problem, but because of the messiness of the stacktrace it wasn't noticed  

But now we send everything through the "prettifier", so every log line is out of order.

The ordering mess-up happens because we unmarshal the log line, do some changes, then just marshal it again.

## How can this be tested?

Look at the log lines produced by the operator:
- "keys" in info logs should have a predefined order
- stacktraces show up in a "pretty" way still

### DIFF
--- a/pkg/logd/pretty_log_writer.go
+++ b/pkg/logd/pretty_log_writer.go
@@ -38,12 +38,18 @@ func (pretty *prettyLogWriter) Write(payload []byte) (int, error) {
 		return 0, errors.New("no output set on prettyLogWriter")
 	}
 
-	payload, err := removeDuplicatedStacktrace(payload)
+	prettified, err := removeDuplicatedStacktrace(payload)
 	if err != nil {
 		return pretty.out.Write(payload)
 	}
 
-	return pretty.out.Write(correctLineEndings(payload))
+	if prettified != nil {
+		prettified = correctLineEndings(prettified)
+
+		return pretty.out.Write(prettified)
+	}
+
+	return pretty.out.Write(payload)
 }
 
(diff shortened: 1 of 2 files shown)


0sewa0:
@andriisoldatenko We don't want that, the change I mention in the description is the PR that is linked there, but it may be confusing so I will reword it

andriisoldatenko (approved):

wepudt (approved):
lgtm
deployed, logs look good

PR #3209, opened 2024-05-29
Only convert metadata enrichment if oneagent injection was necessary

0sewa0:
## Description

~This small change in conversion logic was discussed last week, but it seems to have been forgotten.~
~- Or we decided against it, but then I wasn't informed, and we can just close it 😅~

As it turns out we need to be even stricter and only convert metadata enrichment if oneagent injection was necessary in the original dynakube

## How can this be tested?

Deploy a classicFullStack/hostmonitoring/activegate-only v1beta1 DynaKube, and see what happens 😉


andriisoldatenko (dismissed):

waodim (dismissed):

pkg/api/v1beta1/dynakube/convert_to.go
```diff
@@ -86,7 +86,7 @@ func (src *DynaKube) toActiveGateSpec(dst *v1beta2.DynaKube) {
 }
 
 func (src *DynaKube) toMovedFields(dst *v1beta2.DynaKube) error {
-	if src.Annotations[AnnotationFeatureMetadataEnrichment] == "false" || src.ClassicFullStackMode() {
```
chrismuellner, line 89: As discussed on Slack, we need to also disable metadata enrichment if there was no injection in `v1beta1`, e.g. only Activegate configured.

Enabling metadata enrichment by default for all namespaces (no `namespaceSelector` configured because it's not used) would conflict with all other Dynakubes that may have a correctly configured `namespaceSelector`

chrismuellner (changes requested):

pkg/api/v1beta1/dynakube/convert_to.go
0sewa0: done

chrismuellner (approved):

waodim (approved):

PR #3214, opened 2024-05-29
Remove resources limit field from csi provisioner

waodim:
## Description

After some discussions the decision was made that the limit field in the resources of the csi provisioner is not needed due to it not being used frequently enough. So it was decided to remove it.

Ticket: https://dt-rnd.atlassian.net/browse/K8S-9743

## How can this be tested?

- run helm unittests
- deploy operator via helm and check resources of the csi provisioner

### DIFF
--- a/config/helm/chart/default/questions.yml
+++ b/config/helm/chart/default/questions.yml
@@ -179,20 +179,6 @@ questions:
     type: string
     group: "CSI Driver Deployment Configuration"
 
-  - variable: csidriver.provisioner.limits.cpu
-    label: "CPU resource limits settings for Dynatrace CSI Driver's provisioner container"
-    description: "The maximum amount of CPU resources that the Dynatrace CSI Driver's provisioner container can use. Default: 300m"
-    default: "300m"
-    type: string
-    group: "CSI Driver Deployment Configuration"
-
-  - variable: csidriver.provisioner.limits.memory
-    label: "Memory resource limits settings for Dynatrace CSI Driver's provisioner container"
-    description: "The maximum amount of memory that the Dynatrace CSI Driver's provisioner container can use. Pod restarted if exceeded. Default: 100Mi"
-    default: "100Mi"
-    type: string
-    group: "CSI Driver Deployment Configuration"
-
   - variable: csidriver.registrar.requests.cpu
     label: "CPU resource requests settings for Dynatrace CSI Driver's registrar container"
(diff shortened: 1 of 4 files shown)


0sewa0 (commented):
didn't test this, but I am curious:

we use the `resources` section like:
```
        resources:
          {{- if .Values.csidriver.provisioner.resources }}
          {{- toYaml .Values.csidriver.provisioner.resources | nindent 10 }}
          {{- end }}
```
so we just put in what the customer provides, I don't think removing `resources.limit` from the values.yaml will disallow them from setting limits.
- Do we want this?
- The current change just changes/removes the defaults

EDIT: Also you might have missed some misc files:
- https://github.com/Dynatrace/dynatrace-operator/blob/ec926761c3009f1972f37d372350dbce6d0ac004/config/helm/schema.yaml#L301-L314
- https://github.com/Dynatrace/dynatrace-operator/blob/ec926761c3009f1972f37d372350dbce6d0ac004/config/helm/chart/default/questions.yml#L168-L180

waodim:
> didn't test this, but I am curious:
> 
> we use the `resources` section like:
> 
> ```
>         resources:
>           {{- if .Values.csidriver.provisioner.resources }}
>           {{- toYaml .Values.csidriver.provisioner.resources | nindent 10 }}
>           {{- end }}
> ```
> 
> so we just put in what the customer provides, I don't think removing `resources.limit` from the values.yaml will disallow them from setting limits.
> 
>     * Do we want this?
> 
>     * The current change just changes/removes the defaults
> 
> 
> EDIT: Also you might have missed some misc files:
> 
>     * https://github.com/Dynatrace/dynatrace-operator/blob/ec926761c3009f1972f37d372350dbce6d0ac004/config/helm/schema.yaml#L301-L314
> 
>     * https://github.com/Dynatrace/dynatrace-operator/blob/ec926761c3009f1972f37d372350dbce6d0ac004/config/helm/chart/default/questions.yml#L168-L180

nice catch thank you.

My personal opinion is: If anyone wants to set limits he is absolutely free to do so and we simply changed the defaults. But then again maybe someone has other opinions to that. I do not have a strong one here.

chrismuellner:
> My personal opinion is: If anyone wants to set limits he is absolutely free to do so and we simply changed the defaults. But then again maybe someone has other opinions to that. I do not have a strong one here.

I agree, it should still be possible to configure `limits` for this container, just the defaults should not have them.

0sewa0 (approved):

luhi-DT (approved):

PR #3220, opened 2024-06-04
Increase matrix: max-parallel and set fail-fast: false

andriisoldatenko:
Currently, if the single job fails, Github will cancel all other jobs, but they are independent, so we want to run all of them.

Also running 2 parallel jobs is a bit slow, so I've increased to 4, since every job takes ~ 2hours or so.


luhi-DT (approved):

albertogdd (approved):
We could ask Asperitas how many of them in parallel is ok to run. They are independent clusters so in theory we could remove the maximum threshold, or?

PR #3234, opened 2024-06-05
Add EC validator to reject CR with Kubernetes Automation feature enabled

aorcholski:
## Description

New validator rejects EdgeConnect CR with KubernetesAutomation enabled or customized ServiceAccountName.

ServiceAccountName is set to the default value on the CRD level so it's never empty. 

## How can this be tested?

1) Unittests.

2) Try to apply EC with KubernetesAutomation enabled and/or customized serviceAccountName :
```
spec:
  kubernetesAutomation:
    enabled: false / true
  serviceAccountName: dynatrace-edgeconnect / test-dynatrace-edgeconnect
```
3) Apply valid EdgeConnect then change value of `enabled` and/or `serviceAccountName` and try to save changes.

### DIFF
--- a/pkg/webhook/validation/edgeconnect/api_server_test.go
+++ b/pkg/webhook/validation/edgeconnect/api_server_test.go
@@ -37,10 +37,10 @@ func TestApiServer(t *testing.T) {
 						Endpoint:     "endpoint",
 						Resource:     "resource",
 					},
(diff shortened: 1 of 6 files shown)


andriisoldatenko (approved):

StefanHauth (approved):
Tested it, LGTM. Just minor nitpicks that are not absolutely necessary.

PR #3235, opened 2024-06-05
Fixes autoUpdate being enabled when applying a DynaKube in v1beta1 with autoUpdate set to false.

wepudt:
[K8S-10181](https://dt-rnd.atlassian.net/browse/K8S-10181)

Problem

Applying a DK v1beta1 with OneAgent autoUpdate set to false gets set to true by the conversion webhook.

The reason for this is how the object gets marshalled to JSON as autoUpdate is configured to be `omitempty`. as `false` is the default `bool` value it get's omitted and set to it's default value `true`.

## How can this be tested?

apply a DynaKube in version v1beta1 with 
```oneAgent:
    cloudNativeFullStack:
      autoUpdate: false
```

check the DK in both v1beta1 and v1beta2 and validate that autoUpdate is still false.


pkg/api/v1beta1/dynakube/oneagent_types.go
```diff
@@ -60,7 +60,7 @@ type HostInjectSpec struct {
 	// Enabled by default.
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Automatically update Agent",order=13,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
-	AutoUpdate *bool `json:"autoUpdate,omitempty"`
```
0sewa0, line 63: here its fine, and probably the way to do it, it doesn't make sense to store "nothing"

0sewa0 (approved):

luhi-DT (commented):
Tested and works

luhi-DT (approved):

PR #3242, opened 2024-06-06
Fix hostGroup conversion logic for v1beta2

luhi-DT:
## Description
Ticket: https://dt-rnd.atlassian.net/browse/K8S-10198

Currently the hostGroup parameter is not converted from v1beta1 to v1beta2.

## How can this be tested?
Install Operator -> Apply a v1beta1 Dynakube with spec.oneAgent.hostGroup -> check if the converted v1beta2 Dynakube also has hostGroup set

PR #3243, opened 2024-06-06
Fix host group conversion logic #3242

luhi-DT:
## Description

Same as in #3242 

## How can this be tested?

Same as in #3242


wepudt (approved):
LGTM

checked im my cluster with v1beta1 and v1beta2 DKs, both successfully converted in both directions.

PR #3254, opened 2024-06-07
Make conversion nil safe in case of empty annotations

0sewa0:
## Description

We didn't handle the `Annotations` map safely during conversion, which can cause a panic.

## How can this be tested?

Deploy a v1beta2 DynaKube, with no-annotation and `metadataEnrichment.Enabled = true`. (or any new field that would be converted from an annotation)


wepudt (commented):
LGTM
aside from the linting errors:
code looks good, deployed a DK with no annotations, all went smoothly

andriisoldatenko (approved):

PR #3266, opened 2024-06-10
Make otel error log less scary

0sewa0:
## Description

Setting up OTel requires a specific secret to be created.
This is totally optional, so most of the time this secret won't be there, so logging the error is a bit excessive.

## How can this be tested?

Deploy the operator without having the Otel secret setup,and check the log of the operator is less scary

### DIFF
--- a/pkg/util/dtotel/otel.go
+++ b/pkg/util/dtotel/otel.go
@@ -22,7 +22,7 @@ type shutdownFn func(ctx context.Context) error
 func Start(ctx context.Context, otelServiceName string, apiReader client.Reader, webhookNamespace string) func() {
 	endpoint, apiToken, err := getOtelConfig(ctx, apiReader, webhookNamespace)
 	if err != nil {
-		log.Info("failed to read OpenTelementry config secret", "err", err.Error())
+		log.Info("Otel was not configured, add the secret if you want to enable this functionality.")
 
 		return setupNoopOTel()
 	}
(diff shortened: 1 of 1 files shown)


andriisoldatenko (approved):
thank you!! was in my list!

pkg/util/dtotel/otel.go
```diff
@@ -22,7 +22,7 @@ type shutdownFn func(ctx context.Context) error
 func Start(ctx context.Context, otelServiceName string, apiReader client.Reader, webhookNamespace string) func() {
 	endpoint, apiToken, err := getOtelConfig(ctx, apiReader, webhookNamespace)
 	if err != nil {
-		log.Info("failed to read OpenTelementry config secret", "err", err.Error())
+		log.Info("Otel was not configured, add the secret if you want to enable this functionality.")
```
chrismuellner, line 25:
```suggestion
		log.Info("Opentelemetry instrumentation was not configured")
```

pkg/util/dtotel/otel.go
```diff
@@ -22,7 +22,7 @@ type shutdownFn func(ctx context.Context) error
...
 
 		return setupNoopOTel()
 	}
```
chrismuellner, line 28: should we have a log as well that indicates that otel was enabled?

chrismuellner (changes requested):

pkg/util/dtotel/otel.go
0sewa0: IMO not necessary

chrismuellner (approved):

andriisoldatenko (approved):

PR #3268, opened 2024-06-10
WIP The operator merges trustedCAs certificates with ones shipped in the image

aorcholski:
## Description

The operator should merge `trustedCAs` with the default certificates. It's more convenient approach for users as they can provide just an extra certificate and don't have to worry about for example DT server certificates.

## How can this be tested?

1. e2e istio tests
2. deploy CN dynakube with any certificate using trustedCAs. The operator should be able to connect to the DT server because knows well-know RootCAs shipped with the OS


aorcholski:
By mistake created on top of the main branch

PR #3274, opened 2024-06-11
Add conditions and proper cleanup to injection reconciler

gkrenn:
## Description

The dynakube conditions should show what happend in our operator.
We want to extend the conditions by the result of the injection reconciler. 

In addition we can cleanup the created secrets when needed.

[Jira Ticket](https://dt-rnd.atlassian.net/browse/K8S-9579)

## How can this be tested?

- Run a dynakube with cloudnative and metrics ingest enabled => check if the 2 created conditions are there
- Remove everathing from the dynakube => check if secrets in monitored namespaces are deleted

### DIFF
--- a/pkg/controllers/dynakube/injection/conditions.go
+++ b/pkg/controllers/dynakube/injection/conditions.go
@@ -9,7 +9,7 @@ const (
 	metaDataEnrichmentConditionType   = "MetadataEnrichment"
 	codeModulesInjectionConditionType = "CodeModulesInjection"
 
-	nsAndSecretsCreatedReason  = "NamespaceAndSecretsCreated"
+	nsAndSecretsCreatedReason  = "SecretsCreated"
 	nsAndSecretsCreatedMessage = "Namespaces mapped and secrets created"
 )
 
--- a/pkg/controllers/dynakube/injection/conditions.go
+++ b/pkg/controllers/dynakube/injection/conditions.go
@@ -1,0 +1,34 @@
+package injection
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	metaDataEnrichmentConditionType   = "MetadataEnrichment"
+	codeModulesInjectionConditionType = "CodeModulesInjection"
+
+	nsAndSecretsCreatedReason  = "NamespaceAndSecretsCreated"
+	nsAndSecretsCreatedMessage = "Namespaces mapped and secrets created"
+)
+
+func setMetadataEnrichmentCreatedCondition(conditions *[]metav1.Condition) {
+	condition := metav1.Condition{
+		Type:    metaDataEnrichmentConditionType,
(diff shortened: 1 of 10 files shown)


pkg/controllers/dynakube/injection/conditions.go
```diff
@@ -0,0 +1,34 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	metaDataEnrichmentConditionType   = "MetadataEnrichment"
+	codeModulesInjectionConditionType = "CodeModulesInjection"
+
+	nsAndSecretsCreatedReason  = "NamespaceAndSecretsCreated"
+	nsAndSecretsCreatedMessage = "Namespaces mapped and secrets created"
```
0sewa0, line 13: do we have to be this specific? (mainly in the `Reason`)

waodim: Is it even correct that the namespace is created alongside the secret?

0sewa0: true, it would be more correct to say `NamespacesMappedAndSecretsCreated`, but pls no 😅

waodim: Yeah I agree that it is to specific anyways 😅 

Probably `SecretCreated` is the way to go I suppose

pkg/controllers/dynakube/injection/conditions.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	metaDataEnrichmentConditionType   = "MetadataEnrichment"
+	codeModulesInjectionConditionType = "CodeModulesInjection"
+
+	nsAndSecretsCreatedReason  = "SecretsCreated"
```
0sewa0, line 12: forgot to update the variable name :D

gkrenn: Left on purpose, because I want the Reason and Message to have the same prefix to indicate they belong together. 

`SecretsCreatedMessage` would not fit there - Maybe renaming the prefix to `nsMappedAndSecretsCreated`, but that's rather long.

0sewa0: so my main problem with the naming is that (unfortunately) the `injection` reconciler's outcome is not as straightforward.

the `oneagent` reconciler has the `DaemonSet...` conditions as its outcomes, even though it also creates Secrets.

the `injection` reconciler has the "outcome" that "many things have been set up" 😅 
- as of https://github.com/Dynatrace/dynatrace-operator/pull/3327 you would even have `NamespacesMappedSecretsCreatedStatutsUpdated"

0sewa0 (approved):

albertogdd (approved):
LGTM

waodim (approved):

PR #3329, opened 2024-06-18
Add a wait timeout to E2E Test Setup for OCP Environments and Check Result of Deployments for all Environments

wepudt:
## Description

* Check the deployment status of FLC deployments and only continue if deployment was successful.
* Wait for proper OCP setup so teardown can clean up properly.

## How can this be tested?

Run E2E tests with workflow from this branch while using a release branch to run tests against.

### DIFF
--- a/.github/scripts/create-cluster.sh
+++ b/.github/scripts/create-cluster.sh
@@ -3,6 +3,7 @@
 set -x
 
 DEFAULT_TIMEOUT="60m" # K8s takes <10min, Openshift >40min
+desired_state="environment-deployed"
 
 kubectl version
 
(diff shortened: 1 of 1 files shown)


.github/scripts/create-cluster.sh
```diff
@@ -3,6 +3,7 @@
 set -x
 
 DEFAULT_TIMEOUT="60m" # K8s takes <10min, Openshift >40min
+desired_state="environment-deployed"
```
andriisoldatenko, line 6:
```suggestion
DESIRED_STATE="environment-deployed"
```

andriisoldatenko:
@wepudt For other reviewers, it would be great if you rename PR; it's not refactoring. It's about adding a timeout and checking the status after a timeout has elapsed.

andriisoldatenko (approved):

albertogdd (approved):
Very nice :+1:

PR #3339, opened 2024-06-19
Make csi-gc less aggressive

0sewa0:
## Description
https://dt-rnd.atlassian.net/browse/K8S-10338 ~(part of the fix, not all)~

Removing the whole folder may cause problems in a quick/messy uninstall.

The idea behind the change is that the `BidirectionalMountPropagation` might not be "fast enough" if we remove the whole folder that was mounted very close (ie.: almost at the same time) to the unmount then it might not get propagated properly, causing the pod with the mount to be stuck forever, and requiring manual.

## How can this be tested?

1. Deploy operator + cloud-native
2. remove operator (cloud-native oneagents should get stuck `Terminating`)
3. reinstall operator (cloud-native oneagents should get unstuck)

### DIFF
--- a/pkg/controllers/csi/gc/binaries.go
+++ b/pkg/controllers/csi/gc/binaries.go
@@ -20,6 +20,12 @@ func (gc *CSIGarbageCollector) runBinaryGarbageCollection() {
 	}
 
 	for _, codeModule := range codeModules {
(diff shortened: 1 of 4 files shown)


waodim (approved):
Code LGTM

Tested it: After reinstalling the operator, the oneagents stuck in termination were removed

StefanHauth (approved):
LGTM

PR #3342, opened 2024-06-20
Remove log line from support archive zip.

StefanHauth:
## Description

A rogue log line logging loglevels corrupts the support archive zip if written directly to stdout. Moved logging of loglevel analogous to logging of version string.

## How can this be tested?

Deploy the operator and create a support archive using `--stdout`. Check if resulting zip file can be unpacked. Linux `unzip` was able to handle the corrupted zip file, so try to use other unzippers (e.g. Windows built in zip, 7Zip).


0sewa0 (approved):

aorcholski (approved):
Linux MidnightCommander was not able to properly open corrupted zip file. Now it works. I can read logs, manifests, other files. No need to unzip the package first to see them all.

waodim (approved):

PR #3343, opened 2024-06-20
Create istio exclusion for CSI Driver in case of codeModules or public registry

waodim:
## Description

With this the CSI Driver is excluded from istio blocking traffic if:
- `codeModulesImage` is set
- `public registry` is enabled

`VirtualService` and `ServiceEntry` is created that points to the host where the image is stored. If this is not the case the injection into a pod would not work as everything got stuck because the CSI Driver could not be able to retrieve the image.

~I placed the reconciliation of the `istioReconciler` right before the injection reconciliation, please give me feedback if you are fine with this.~
I placed the reconciliation after the codeModules versionReconciler is finished.
 Currently I am only checking if the `ImageID` is set in the status of the `CodeModule`, this should cover both cases (`codeModulesImage` set or `public registry` enabled).

Respective ticket: https://dt-rnd.atlassian.net/browse/K8S-9622

## How can this be tested?

Deploy the operator with istio, do not forget to label the to be injected namespace accordingly (`istio-injection:enabled`)
Apply a dynakube with istio enabled and
- with a `codeModulesImage` set
- with `public registry `feature flag enabled

Also check for `VirtualServices` and `ServiceEntries` and look if they are created correctly.


pkg/controllers/dynakube/controller.go
```diff
@@ -324,6 +324,15 @@ func (controller *Controller) reconcileComponents(ctx context.Context, dynatrace
...
 
+	if istioClient != nil {
+		istioReconciler := controller.istioReconcilerBuilder(istioClient)
+
+		err := istioReconciler.ReconcileCSIDriver(ctx, dynakube)
+		if err != nil {
+			return errors.WithMessage(err, "failed to reconcile istio objects for CSI Driver")
+		}
+	}
+
```
0sewa0, line 335: this is probably not the correct space for it. you need the version reconciliation to happen before you can do this, 

also I think this is only relevant in case of (oneagent) injection is needed, so probably a nicer place to put is the injection reconciler (near where the relevant version reconcilitation is happening)

https://github.com/Dynatrace/dynatrace-operator/blob/b74cbbdb67c0d2784b42c272c6a027c5adcf0d57/pkg/controllers/dynakube/injection/reconciler.go#L70

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -38,6 +39,39 @@ func NewReconciler(istio *Client) Reconciler {
...
 }
 
+func (r *reconciler) ReconcileCSIDriver(ctx context.Context, dynakube *dynatracev1beta2.DynaKube) error {
+	log.Info("reconciling istio components for the CSI driver")
+
+	if dynakube == nil {
+		return errors.New("can't reconcile csi driver of nil dynakube")
+	}
+
+	codeModulesURL := dynakube.Status.CodeModules.ImageID
```
0sewa0, line 49: `dynakube.CodeModulesImage()` does the same

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -38,6 +39,39 @@ func NewReconciler(istio *Client) Reconciler {
...
+	}
+
+	log.Info("reconciled istio objects for CSI driver")
+
+	return nil
+}
+
+// if the URL does not have the correct format parsing does not work‚
+func hasCorrectFormat(url string) bool {
+	return strings.HasPrefix(url, "https://")
```
0sewa0, line 72: what if I like to live dangerously an I have `http://`

waodim: Good point. I am not sure how I could implement this as usualy the `imageID` that is set comes with no protocol at all. `dtclient.ParseEndpoint` fails when there is no protocol is set, which is why I did it that way. Now I am unsure how I could make this check between `http` and `https`

Do you have any suggestions?

pkg/controllers/dynakube/istio/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3343/commits/b5566bd5bdf7edc5afefd8d6d43208d1790c1cf8

pkg/controllers/dynakube/controller.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3343/commits/b5566bd5bdf7edc5afefd8d6d43208d1790c1cf8

pkg/controllers/dynakube/istio/reconciler.go
0sewa0: you can check if there is a schema or not with the `url` library, and append `https` if nothing is present , but the current logic would just append `https://` to almost anything :D

waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3343/commits/c92b4c3055a108823e37d4945a6230151c798548

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -120,6 +125,17 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+func (r *reconciler) reconcileIstioForCSIDriver(ctx context.Context) error {
+	if r.istioReconciler != nil {
+		err := r.istioReconciler.ReconcileCSIDriver(ctx, r.dynakube)
+		if err != nil {
+			return errors.WithMessage(err, "failed to reconcile istio objects for CSI Driver")
+		}
+	}
+
+	return nil
+}
```
0sewa0, line 137: we already have a `r.istioReconciler != nil` check, why have it separately?

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -38,6 +40,58 @@ func NewReconciler(istio *Client) Reconciler {
...
+		return "", errors.New("can't parse the codeModules image URL")
+	}
+
+	if parsedURL.Scheme == "" {
+		parsedURL.Scheme = "https"
+
+		if !strings.HasPrefix(rawUrl, "//") {
+			// if no scheme at all is set we want to add this prefix
+			rawUrl = "//" + rawUrl
+		}
```
0sewa0, line 84: maybe I am not getting something here, but why is this necessary?

pro-tip: `parseCodeModulesImageURL` is nice and easily testable function, writing a bunch of unittests with all the weird possible `rawUrl` examples might help understanding (for all parties) 😉

waodim: No you are right, this rules out my logic tbh 😅  will remove it and add proper tests 👍

waodim: Note: It is indeed necessary now I remember again why: if I have some `url` that looks like this: `some.test.url `and do not have this if check the function it will concatenate the scheme which will be `https` in this case with `:` and the `rawUrl`, but I will fix it by simply concatenating the scheme with the string `://` and the `rawUrl`, this does the same.

waodim: pls have a look at the tests and especially the case where i test `//some.url.com/test` 

Tell me if this is a actually possible case. Because if not I can remove the logic which you questioned again, because for now I let it in.

https://github.com/Dynatrace/dynatrace-operator/pull/3343/commits/8075a6597276e226add71a643131b960503aa5e8

pkg/controllers/dynakube/injection/reconciler.go
waodim: The only check I see is the one for `ReconcileCodeModuleCommunicationHosts` which comes after the `connectionInfoReconciler`. The CSIDriver reconciler comes before it, how would you like to have this?

0sewa0: why does the CSI reconcile need to before the `connectionInfoReconciler`? nothing depends on the CSI reconcile (other then the csi-driver, but that is in a whole different pod, so I wouldn't worry about it)
- The csi-driver also depends on the `r.pmcSecretreconciler.Reconcile(ctx)` which is at the very end of the injection reconcile, so 😛 

the only "requirement" is that the CSI reconcile happens after the version reconcile

waodim: You are totally right, thanks

https://github.com/Dynatrace/dynatrace-operator/pull/3343/commits/eeb60eb316cf1d7799d179f74d01bbe780ee3cef

pkg/controllers/dynakube/istio/config.go
```diff
@@ -11,6 +11,7 @@ var (
 )
 
 const (
+	CSIDiverComponent   = "csi-driver"
```
gkrenn, line 14:
```suggestion
	CSIDriverComponent   = "csi-driver"
```

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -22,6 +23,7 @@ type Reconciler interface {
 	ReconcileAPIUrl(ctx context.Context, dynakube *dynatracev1beta2.DynaKube) error
 	ReconcileCodeModuleCommunicationHosts(ctx context.Context, dynakube *dynatracev1beta2.DynaKube) error
 	ReconcileActiveGateCommunicationHosts(ctx context.Context, dynakube *dynatracev1beta2.DynaKube) error
+	ReconcileCSIDriver(ctx context.Context, dynakube *dynatracev1beta2.DynaKube) error
```
gkrenn, line 26: `ReconcileCSIDriver` should always be called after `ReconcileCodeModuleCommunicationHosts` as far as I got it.
I would call `ReconcileCSIDriver` then directly within `ReconcileCodeModuleCommunicationHosts` instead of creating an additional interface function.

waodim: I can happily do that, there is only one question: Wouldn't this lead to `ReconcileCodeModuleCommunicationHosts` as a method doing too many things? Or can the `host` of the `codeModulesImage` URL be considered as a communication host?

also maybe interesting for you @0sewa0

0sewa0: > Or can the host of the codeModulesImage URL be considered as a communication host?

that is what you are doing here: `err = r.reconcileCommunicationHosts(ctx, []dtclient.CommunicationHost{codeModulesHost}, CSIDriverComponent)`

the fact that the `istio.Reconciler` works with `dtclient.CommunicationHost` is just for historical reasons, there is no better reason for it, it was just always used for creating `ServiceEntries` for a given list of `dtclient.CommunicationHost`s

I like the idea that we don't bloat the (hacky) interface, and just do both in the `ReconcileCodeModuleCommunicationHosts`, but you would probably want to rename that function to something "more correct". Like `ReconcileInjectionEndpoints` or something smarter

waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3343/commits/fdc1c92c54d8f068b0b02b3b4637cdaf7048034e

pkg/controllers/dynakube/istio/reconciler_test.go
```diff
@@ -450,6 +508,57 @@ func TestReconcileActiveGateCommunicationHosts(t *testing.T) {
...
+			input:           ":example.com/test",
+			output:          "",
+			parsedCorreclty: false,
+		},
+		{
+			input:           "//some.url.com/test",
+			output:          "https://some.url.com/test",
+			parsedCorreclty: true,
+		},
+	}
```
0sewa0, line 542: be aware that the URL you are parsing is an image url (as you pass in the `dk.Status.CodeModules.ImageID` directly), so it will have a `:tag` as a postfix, be sure to test that scenario

pkg/controllers/dynakube/istio/reconciler_test.go
```diff
@@ -450,6 +508,57 @@ func TestReconcileActiveGateCommunicationHosts(t *testing.T) {
...
+		} else {
+			if err != nil {
+				t.Errorf("Did not expect error for input %s, but got one: %v", test.input, err)
+			}
+
+			if output != test.output {
+				t.Errorf("For input %s, expected output %s, but got this %s", test.input, test.output, output)
+			}
+		}
+	}
```
0sewa0, line 559: why not use `require` and `assert`?

```suggestion
		if !test.parsedCorreclty {
			require.Error(t, err, "Expected error for input, but did not get any", "input", test.input)
		} else {
			require.NoError(t, err, "Did not expect error for input, but got one", "input", test.input)
			assert.Equal(t, test.output, output)
		}
	}
```

pkg/controllers/dynakube/istio/reconciler.go
0sewa0: I have no idea when `//some.url.com/test` would happen and what it means

0sewa0 (changes requested):

pkg/controllers/dynakube/istio/reconciler.go
waodim: will remove it

pkg/controllers/dynakube/istio/reconciler_test.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3343/commits/619e50a181dc45bb873290e4cd0f59372eadecdc

pkg/controllers/dynakube/istio/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3343/commits/619e50a181dc45bb873290e4cd0f59372eadecdc

pkg/controllers/dynakube/istio/reconciler_test.go
```diff
@@ -450,6 +508,55 @@ func TestReconcileActiveGateCommunicationHosts(t *testing.T) {
 	})
 }
 
+func TestParseCodeModulesImageURL(t *testing.T) {
```
0sewa0, line 511: there is a special case that was missed:
- when the image used in dockerhub: `codeModulesImage: dynatrace/dynatrace-codemodules:1.283.139.20240209-194956`

the resulting `ServiceEntry`:
```
apiVersion: networking.istio.io/v1
kind: ServiceEntry
metadata:
  annotations: ...
  labels: ...
  name: test-fqdn-csi-driver
  namespace: dynatrace
  ownerReferences: ...
spec:
  hosts:
  - dynatrace
  ports:
  - name: https-443
    number: 443
    protocol: HTTPS
  resolution: DNS
```

which doesn't help the CSI-driver: 
```
{"level":"info","ts":"2024-07-01T07:55:00.825Z","logger":"oneagent-image","msg":"failed to extract agent binaries from image via proxy","image":"dynatrace/dynatrace-codemodules:1.283.139.20240209-194956","imageCacheDir":"/data/cache","err":"getting image \"dynatrace/dynatrace-codemodules:1.283.139.20240209-194956\": Get \"https://index.docker.io/v2/\": read tcp 10.108.3.147:42910->3.219.239.5:443: read: connection reset by peer","errVerbose":"Get \"https://index.docker.io/v2/\": read tcp 10.108.3.147:42910->3.219.239.5:443: read: connection reset by peer\ngetting image \"dynatrace/dynatrace-codemodules:1.283.139.20240209-194956\""}
```

0sewa0: how to figure out if the URL is incomplete because its referencing a dockerhub image?

I have no clue really 😬 

How the image library does it: https://github.com/google/go-containerregistry/blob/main/pkg/name/repository.go#L80-L87

waodim: I did it now that way, this ONLY covers docker so I am not sure if we should go for it, pls let me know.
https://github.com/Dynatrace/dynatrace-operator/pull/3343/commits/ea80b5816bc78d79ae07ff2fea28b051643afd75

0sewa0: still doesn't work: 😢  (probably need `index.docker.com`)

```
{"error":"getting image \"dynatrace/dynatrace-codemodules:1.293.133.20240618-095559\": Get \"https://index.docker.io/v2/\": read tcp 10.12.1.11:42066-\u003e34.226.69.105:443: read: connection reset by peer","level":"info","logger":"oneagent-image","msg":"pullImageInfo","stacktrace":"Get \"https://index.docker.io/v2/\": read tcp 10.12.1.11:42066-\u003e34.226.69.105:443: read: connection reset by peer
provisioner getting image \"dynatrace/dynatrace-codemodules:1.293.133.20240618-095559\"","ts":"2024-07-03T12:20:54.687Z"}
provisioner {"level":"info","ts":"2024-07-03T12:20:54.687Z","logger":"oneagent-image","msg":"failed to extract agent binaries from image via proxy","image":"dynatrace/dynatrace-codemodules:1.293.133.20240618-095559","imageCacheDir":"/data/cache","err":"getting image \"dynatrace/dynatrace-codemodules:1.293.133.20240618-095559\": Get \"https://index.docker.io/v2/\": read tcp 10.12.1.11:42066->34.226.69.105:443: read: connection reset by peer","errVerbose":"Get \"https://index.docker.io/v2/\": read tcp 10.12.1.11:42066->34.226.69.105:443: read: connection reset by peer\ngetting image \"dynatrace/dynatrace-codemodules:1.293.133.20240618-095559\""}
```

waodim: I am on it and I'll try it out

waodim: Tried it out and it seems to work for me https://github.com/Dynatrace/dynatrace-operator/pull/3343/commits/379c22d8aefc684fc645c69fddc125066c3c59de

waodim:
```
provisioner {"level":"info","ts":"2024-07-03T12:46:47.397Z","logger":"oneagent-image","msg":"installing agent from image"}                                                                                         │
│ provisioner {"level":"info","ts":"2024-07-03T12:46:47.397Z","logger":"oneagent-image","msg":"installing agent","target dir":"/data/codemodules/ZHluYXRyYWNlL2R5bmF0cmFjZS1jb2RlbW9kdWxlczoxLjI5My4xMzMuMjAyNDA2MTg │
│ provisioner {"level":"info","ts":"2024-07-03T12:46:47.885Z","logger":"oneagent-image","msg":"pullOciImage","ref_identifier":"1.293.133.20240618-095559","ref.Name":"index.docker.io/dynatrace/dynatrace-codemodule │
│ provisioner {"level":"info","ts":"2024-07-03T12:46:57.345Z","logger":"oneagent-image","msg":"unpackOciImage","sourcePath":"/data/cache/ZHluYXRyYWNlL2R5bmF0cmFjZS1jb2RlbW9kdWxlczoxLjI5My4xMzMuMjAyNDA2MTgtMDk1NTU │
│ provisioner {"level":"info","ts":"2024-07-03T12:46:57.345Z","logger":"oneagent-zip","msg":"extracting tar gzip","source":"/data/cache/ZHluYXRyYWNlL2R5bmF0cmFjZS1jb2RlbW9kdWxlczoxLjI5My4xMzMuMjAyNDA2MTgtMDk1NTU5 │
│ provisioner {"level":"info","ts":"2024-07-03T12:47:07.814Z","logger":"oneagent-zip","msg":"moving unpacked archive to target","targetDir":"/data/codemodules/ZHluYXRyYWNlL2R5bmF0cmFjZS1jb2RlbW9kdWxlczoxLjI5My4xM │
│ provisioner {"level":"info","ts":"2024-07-03T12:47:07.815Z","logger":"oneagent-image","msg":"unpackOciImage","targetDir":"/data/codemodules/ZHluYXRyYWNlL2R5bmF0cmFjZS1jb2RlbW9kdWxlczoxLjI5My4xMzMuMjAyNDA2MTgtMD │
│ provisioner {"level":"info","ts":"2024-07-03T12:47:07.897Z","logger":"oneagent-symlink","msg":"found version","version":"1.293.133.20240618-095559"}                                       
```

0sewa0: 1. Is your csi-driver injected by istio?
2. Is your istio configured to be in restricted mode?

waodim: After configuring my istio as you suggested:

`istioctl install --set meshConfig.outboundTrafficPolicy.mode=REGISTRY_ONLY`

I can confirm that the csi driver is injected with the istio containers. 

`dynatrace-oneagent-csi-driver-5htgs   5/5     Running   0               5m55s`

0sewa0: you are going to love this (IMO, I don't even want to do anything like this, because this shows how unpredictable this whole scenario is)

It still does not work:
```
provisioner {"error":"getting image \"dynatrace/dynatrace-codemodules:1.293.133.20240618-095559\": Get \"https://auth.docker.io/token?scope=repository%3Adynatrace%2Fdynatrace-codemodules%3Apull\u0026service=registry.docker.io\": read tcp 10.12.3.29:43420-\u003e54.196.99.49:443: read: connection reset by peer","level":"info","logger":"oneagent-image","msg":"pullImageInfo","stacktrace":"Get \"https://auth.docker.io/token?scope=repository%3Adynatrace%2Fdynatrace-codemodules%3Apull\u0026service=registry.docker.io\": read tcp 10.12.3.29:43420-\u003e54.196.99.49:443: read: connection reset by peer"}
```

so just add `auth.docker.io`, right?
nope, then you will get:
```
provisioner {"level":"info","ts":"2024-07-04T12:23:46.322Z","logger":"oneagent-image","msg":"saving v1.Image img as an OCI Image Layout at path","/data/cache":"Get \"https://production.cloudflare.docker.com/registry-v2/docker/registry/v2/blobs/sha256/64/644c3f0771f461096ccb2c5ce4132447032b833b466cfd845c1a89fd4e5eb762/data?verify=1720098825-yv65IR1rXZ98%2FyHhfEAM%2FMx6cJY%3D\": read tcp 10.12.3.29:55420->104.16.99.215:443: read: connection reset by peer"}
```

so you have to add `production.cloudflare.docker.com` as well

Sidenote:
- using a wildcard like `*.docker.io` or `*.docker.com` is not that simple (you have to mess with the DNS within your application [if I understand it correctly](https://istio.io/latest/docs/reference/config/networking/service-entry/))

waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3343/commits/d23af95e7ef75fcea97566642a47125edd6b6b9a although i do not like this solution at all tbh. Seems to constructed for this case...

waodim:
closed this PR for now, as the way it is done here is way too hacky. We will revisit this and rework.

PR #3356, opened 2024-06-25
Add EEC ServiceAccount, Role and RoleBinding to Helm chart

albertogdd:
## Description

[K8S-10119](https://dt-rnd.atlassian.net/browse/K8S-10119)

Add EEC `ServiceAccount` and Openshift's `Role` and `RoleBinding` to Helm chart.

## How can this be tested?

- `make test/helm`
- `make deploy/helm`, check that `ServiceAccount` exists. 
- `Role` and `RoleBinding` should only exist when deploying on Openshift.

### DIFF
--- a/config/helm/chart/default/templates/Common/extensions-controller/serviceaccount-extensions-controller.yaml
+++ b/config/helm/chart/default/templates/Common/extensions-controller/serviceaccount-extensions-controller.yaml
@@ -1,0 +1,24 @@
+{{- include "dynatrace-operator.platformRequired" . }}
+{{ if eq (include "dynatrace-operator.partial" .) "false" }}
+# Copyright 2021 Dynatrace LLC
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: dynatrace-extensions-controller
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "dynatrace-operator.extensionsControllerLabels" . | nindent 4 }}
+{{ end }}
+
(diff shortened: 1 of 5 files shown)


config/helm/chart/default/templates/Common/extensions-controller/serviceaccount-extensions-controller.yaml
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: dynatrace-extensions-controller
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "dynatrace-operator.extensionsControllerLabels" . | nindent 4 }}
+{{ end }}
```
andriisoldatenko, line 23: funny thing, I know helm supports both, but in second file we use `{{- end -}}`, so is it possible to sync it?

albertogdd: I see on the templates we mostly use `{{ end }}`, so IMO this fits more the standard.
[I had to google about it](https://stackoverflow.com/questions/67293728/what-is-the-difference-between-helm-syntax-something-and-something), adding the `-` puts an empty space if rendered, so I would say in this case makes sense to have only {{ end }}.

andriisoldatenko: sure, my point is to have the same syntax in these new files. Maybe later, you can create a new PR to remove from all files.

albertogdd: :+1::+1::+1:
As it is now the PR, it follows the standards we already had in this files. We can later create a followup PR and update them if we want to normalize them (but IMO is fine as it is).

config/helm/chart/default/templates/_labels.tpl
```diff
@@ -100,3 +100,11 @@ OneAgent labels
...
 app.kubernetes.io/component: oneagent
 {{- end -}}
+
+{{/*
+Extensions Controller (EEC) labels
+*/}}
+{{- define "dynatrace-operator.extensionsControllerLabels" -}}
+{{ include "dynatrace-operator.commonLabels" . }}
+app.kubernetes.io/component: dynatrace-extensions-controller
+{{- end -}}
```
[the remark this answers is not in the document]
albertogdd: https://github.com/Dynatrace/dynatrace-operator/pull/3356#discussion_r1652742223

andriisoldatenko (approved):

aorcholski (approved):

PR #3358, opened 2024-06-26
Fix wrong hostname used for Kubernetes API in EdgeConnect configuration.

StefanHauth:
## Description

The wrong (incomplete) hostname was used for host restrictions in the EdgeConnect config file. This has been fixed. Additonally, `HostMappings` was moved to keep dependencies cleaner.

## How can this be tested?

Deploy EdgeConnect wtih enabled `kuberneteAutomation` and check the EdgeConnec configuration.

### DIFF
--- a/pkg/api/v1alpha1/edgeconnect/edgeconnect_types.go
+++ b/pkg/api/v1alpha1/edgeconnect/edgeconnect_types.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/status"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha1"
-	"github.com/Dynatrace/dynatrace-operator/pkg/clients/edgeconnect"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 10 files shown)


pkg/api/v1alpha1/edgeconnect/edgeconnect_types.go
```diff
@@ -214,9 +213,14 @@ func (e *EdgeConnect) HostPatterns() []string {
 	return hostPatterns
 }
 
-func (e *EdgeConnect) HostMappings() []edgeconnect.HostMapping {
-	hostMappings := make([]edgeconnect.HostMapping, 0)
-	hostMappings = append(hostMappings, edgeconnect.HostMapping{From: e.K8sAutomationHostPattern(), To: kubernetesDefaultDNS})
+type HostMapping struct {
+	From string `json:"from"`
+	To   string `json:"to"`
+}
```
albertogdd, line 219: Shouldn't we keep this on the client package? Because it's only used through the API and it's not in the CR (for now).
But I don't have a strong opinion about it

StefanHauth: What I kind of disklike is that, we have an dependency from the API pkg downward.

albertogdd (approved):

agrimmer:
Thanks! The fix works as expected!

andriisoldatenko (approved):

PR #3359, opened 2024-06-26
Check `name` and `namespace` when filtering EdgeConnect connection settings object

albertogdd:
## Description

[K8S-10407](https://dt-rnd.atlassian.net/browse/K8S-10407)

Add name and namespace when getting the EdgeConnect settings object

## How can this be tested?

Deploy EdgeConnect on provisioner: true and kubernetesAutomation.enabled: true. Check that object gets created/updated/deleted

### DIFF
--- a/pkg/clients/edgeconnect/environment_settings.go
+++ b/pkg/clients/edgeconnect/environment_settings.go
@@ -35,7 +35,7 @@ type EnvironmentSettingsResponse struct {
 	PageSize   int                  `json:"pageSize"`
 }
 
-func (c *client) GetConnectionSetting(uid string) (EnvironmentSetting, error) {
+func (c *client) GetConnectionSetting(name, namespace, uid string) (EnvironmentSetting, error) {
 	settingsObjectsUrl := c.getSettingsObjectsUrl()
 
 	req, err := http.NewRequestWithContext(c.ctx, http.MethodGet, settingsObjectsUrl, nil)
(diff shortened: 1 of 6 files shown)


StefanHauth (dismissed):
Code LGTM, but I'm not able to test it because of missing setup.

pkg/clients/edgeconnect/environment_settings.go
```diff
@@ -69,7 +69,7 @@ func (c *client) GetConnectionSetting(uid string) (EnvironmentSetting, error) {
 	}
 
 	for _, item := range resDataJson.Items {
-		if item.Value.UID == uid {
+		if item.Value.Name == name && item.Value.Namespace == namespace && item.Value.UID == uid {
```
andriisoldatenko, line 72: could you please add a test, which checks that we mismatch `name` or `namespace` like written in ticket?
As you can see we never return empty:
image

andriisoldatenko (changes requested):

pkg/clients/edgeconnect/environment_settings.go
albertogdd: Done [here](https://github.com/Dynatrace/dynatrace-operator/pull/3359/commits/d3803495de1b8596a2e998b139c62675ef5b4131)
I also refactored a bit the implementation, removing logic from the client into the controller (it was also necessary to test the logic)

andriisoldatenko (approved):

StefanHauth (approved):

PR #3361, opened 2024-06-26
WIP Ensure ActiveGate is Enabled for Extensions

wepudt:
## Description

see [K8S-10118](https://dt-rnd.atlassian.net/browse/K8S-10118)

ActiveGate is also created when Extensions are enabled in the DynaKube. No other AG capabilities are needed.

Dynakube v1beta2 is manually converted to v1beta3 in all cases when `dynakube.NeedsActiveGate()` or `NeedsActiveGateService()` is called.

## How can this be tested?

1) Unittests.
2) Apply dynakubes below. ActiveGate pod should be always created.
Check enabled capabilities on the AG pod using:
```
kubectl -n dynatrace get -o yaml pod/dynakube-activegate-0 -o jsonpath="{.spec.containers[0].env[0]}"
```

a)
dynakube:
```
spec:
  activeGate:
    capabilities:
      - routing
    replicas: 1
  extensions:
    prometheus:
      enabled: true
  oneAgent:
    cloudNativeFullStack: {}
```
capabilities:
```
{"name":"DT_CAPABILITIES","value":"MSGrouter,extension_controller"}
```

b)
dynakube:
```
spec:
  activeGate:
    replicas: 1
  extensions:
    prometheus:
      enabled: true
  oneAgent:
    cloudNativeFullStack: {}
```
capabilities:
```
{"name":"DT_CAPABILITIES","value":"extension_controller"}
```

c)
dynakube:
```
spec:
  extensions:
    prometheus:
      enabled: true
  oneAgent:
    cloudNativeFullStack: {}
```
capabilities:
```
{"name":"DT_CAPABILITIES","value":"extension_controller"}
```


pkg/api/v1beta2/dynakube/dynakube_types.go
```diff
@@ -76,6 +76,9 @@ type DynaKube struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	Spec DynaKubeSpec `json:"spec,omitempty"`
+
+	// TODO: remove when extensions are present
```
luhi-DT, line 80: why do we add this in the dynakube? Cant it be an annotation for now

0sewa0: I agree with ☝️ , but also:

why start implementing a feature that is based on a CRD change that is not done yet?
- the way the `v1beta3` CRD introduction will be handled is still up in the air a bit, but AFAIK parts that will need the new fields in `v1beta3` will do some conversion from `v1beta2`, as the `v1beta3` CR will be stored in the annotation, so back and forth conversion without data loss can be done/
   - but still, this is still uncertain 

TLDR: why implement something that will need to be reimplemented in a few weeks?

and if this is just a POC that should never be merged, why the PR?

wepudt: actually the change is minuscule and this field exists just for easy testing now and will be removed and not be missed at all :)

Instead the method [`HasExtensionsEnabled()`](https://github.com/Dynatrace/dynatrace-operator/blob/9c2f68535bc40c1a1855297817c4b6d9e72a330b/pkg/api/v1beta2/dynakube/properties.go#L110) will be used to check if extensions are enabled. In Tests this then can be easily enabled, almost like now...

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/eec.go
```diff
@@ -0,0 +1,81 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+var _ volumeModifier = EecModifier{}
+var _ volumeMountModifier = EecModifier{}
+var _ builder.Modifier = EecModifier{}
+
+const (
+	eecVolumeName = "eec-token"
+	eecDir        = "/var/lib/dynatrace/secrets/eec"
```
aorcholski, line 18: please set it to "/var/lib/dynatrace/secrets/eec/token"

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/eec.go
```diff
@@ -0,0 +1,81 @@
...
+func (mod EecModifier) Modify(sts *appsv1.StatefulSet) error {
+	baseContainer := container.FindContainerInPodSpec(&sts.Spec.Template.Spec, consts.ActiveGateContainerName)
+	sts.Spec.Template.Spec.Volumes = append(sts.Spec.Template.Spec.Volumes, mod.getVolumes()...)
+	baseContainer.VolumeMounts = append(baseContainer.VolumeMounts, mod.getVolumeMounts()...)
+
+	baseContainer.Command = []string{"/bin/sh"}
+	baseContainer.Args = []string{
+		"-c",
+		"echo abc ; cp -v " + eecDir + "/" + eecFile + " /var/lib/dynatrace/gateway/config/ ; /opt/dynatrace/gateway/entrypoint.sh",
+	}
```
aorcholski, line 45: The `cp` command is not needed any more. AG-ExtensionController can read from '/var/lib/dynatrace/secrets/eec/token` directory (AG v 1.296.0.20240628)

aorcholski (changes requested):

pkg/controllers/dynakube/activegate/capability/capability.go
```diff
@@ -76,18 +76,29 @@ func NewMultiCapability(dk *dynatracev1beta2.DynaKube) Capability {
...
 	mc.properties = &dk.Spec.ActiveGate.CapabilityProperties
-	capabilityNames := []string{}
-	capabilityDisplayNames := make([]string, len(dk.Spec.ActiveGate.Capabilities))
 
-	for i, capName := range dk.Spec.ActiveGate.Capabilities {
+	if len(dk.Spec.ActiveGate.Capabilities) == 0 {
+		mc.properties.Replicas = 1
+	}
+
+	var capabilityNames []string //nolint:prealloc
```
aorcholski, line 84:
```suggestion
	capabilityNames:= make([]string, 0)
```

aorcholski (approved):

cmd/troubleshoot/image.go
```diff
@@ -4,7 +4,8 @@ import (
 	"context"
 	"net/http"
 
-	dynatracev1beta2 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube"
+	dynakubev1beta3 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
```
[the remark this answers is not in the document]
0sewa0: I read through more of the code, and it always comes back up.
And I started to ask my self: "What is the benefit of handling 2 versions in a package?"

It just confusing, and only will cause more work, because when we move away from the older version, all of these packages need to be changed.
There is no information loss when converting between the versions, so what the old-version had the new-version also has. So doing this "use double the version, and sometimes pass in 2 dynakubes" are just asking for mistakes caused by confusion.

IMO The biggest flaw with this whole approach (converting at "random" places) is that we like to pass the DynaKube via reference, and we expect the Status part to be mutated as we pass it around, but the moment you introduce conversions in places, and not handle the mutation carefully, some data may be lost

How I would do it is:
- Only 1 version is allowed in a package (conversion and main dk-controller is exempt, maybe troubleshoot and istio aswell)
- If a package uses the new-version in any way, the input for that package should be the new-version
  - The caller of the package will do the convertTo and From, so it will own the reference that is passed it, so no changes to the Status should be lost.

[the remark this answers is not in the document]
albertogdd: While I agree with you, I think we should focus on unblocking this PR (because Werner is on vacation, many stories depends on this and it's getting stucked) and then start switching to beta3 everywhere. We will have to do it anyways with the oncoming stories so we can do it in small increments, or maybe do it all at once after this is merged?

I'm just trying to find the most painless way of moving forward with this.

pkg/controllers/dynakube/activegate/capability/capability.go
albertogdd: Resolved

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/eec.go
albertogdd: @aorcholski If you already changed this, please close the discussions to make the PR more readable

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/eec.go
albertogdd: @aorcholski If you already changed this, please close the discussions to make the PR more readable

pkg/api/v1beta2/dynakube/dynakube_types.go
albertogdd: This change was removed from the PR, we can close the discussion (feel free to reopen it)

aorcholski:
new PR based on v1beta3 branch will be created

PR #3365, opened 2024-06-27
Add all pull secrets to all components

aorcholski:
## Description

[K8S-10395](https://dt-rnd.atlassian.net/browse/K8S-10395)

All dynakube image pull secrets added to OneAgent, ActiveGate specs.

All dynakube image pull secrets used by csi.ImageInstaller.

EdgeConnect controller uses registry.Client but depends on EdgeConnect.Spec.CustomPullSecret.

dynakube.controller.registryClientBuilder is not used (~registry.Client~).

csiprovisioner.OneAgentProvisioner.registryClientBuilder is not used (~registry.Client~).

How it works:

dockerAuth(s) loaded by `configFile.LoadFromReader(...)` from different secrets are stored in a single instance of `configFile.AuthConfigs` map. `configFile.LoadFromReader(...)` basically does `map[k] = v` so it works for many secrets.

See:
- https://github.com/docker/cli/blob/master/cli/config/config.go#L106
- https://github.com/docker/cli/blob/master/cli/config/configfile/file.go#L83


## How can this be tested?

Copy codemodules image to a private registry:
```
podman pull public.ecr.aws/dynatrace/dynatrace-codemodules:1.293.133.20240618-095559 
podman tag public.ecr.aws/dynatrace/dynatrace-codemodules:1.293.133.20240618-095559 <registry>/codemodules:1.293.133.20240618-095559 
podman push <registry>/codemodules:1.293.133.20240618-095559 
```
Create pull image secret:
```
kubectl -n dynatrace create secret docker-registry <secretname>  --docker-server=<registry> --docker-username=<username>  --docker-password=<password>
```
Enable DEBUG log level:
```
kubectl -n dynatrace set env daemonset.apps/dynatrace-oneagent-csi-driver -c provisioner --env=LOG_LEVEL=debug
```
Apply dynakube:
```
spec:
  customPullSecret: <secretname>                                                                                                                                                                                                                                                                                                                                                                      
  oneAgent:                                                                                                                                                                                                                                                                                                                                                                                   
    cloudNativeFullStack:                                                                                                                                                                                                                                                                                                                                                                     
      codeModulesImage: "<registry>/codemodules:1.293.133.20240618-095559" 
```
Check csi logs:
```
[]$ kubectl -n dynatrace logs -f pod/dynatrace-oneagent-csi-driver-<svbwp>

{"level":"debug","ts":"2024-07-04T11:52:56.889Z","logger":"docker-keychain","msg":"loaded docker configs","registries":["<registry>","<tenant>"]}
{"level":"info","ts":"2024-07-04T11:52:56.889Z","logger":"oneagent-image","msg":"installing agent from image"}
{"level":"info","ts":"2024-07-04T11:52:56.889Z","logger":"oneagent-image","msg":"installing agent","target dir":"/data/codemodules/cXVheS5pby9hZGFtX29yY2hvbHNraS9keW5vcDoxLjI5My4xMzMuMjAyNDA2MTgtMDk1NTU5"}
{"level":"info","ts":"2024-07-04T11:52:57.298Z","logger":"oneagent-image","msg":"pullOciImage","ref_identifier":"1.293.133.20240618-095559","ref.Name":"<registry>/codemodules:1.293.133.20240618-095559","ref.String":"<registry>/codemodules:1.293.133.20240618-095559"}
{"level":"info","ts":"2024-07-04T11:53:00.736Z","logger":"oneagent-image","msg":"unpackOciImage","sourcePath":"/data/cache/cXVheS5pby9hZGFtX29yY2hvbHNraS9keW5vcDoxLjI5My4xMzMuMjAyNDA2MTgtMDk1NTU5/blobs/sha256/9592808cd833be900caf7c72a18a97cc8c6c0f54077ddcca84a1030d8fec18b6"}
```
Check OneAgent image pull secrets:
```
kubectl -n dynatrace get daemonset.apps/dynakube-oneagent -o jsonpath="{.spec.template.spec.imagePullSecrets}"
```
there should be two secrets:
```
[{"name":"dynakube-pull-secret"},{"name":"<secretname>"}]
```

### DIFF
--- a/pkg/api/v1beta2/dynakube/properties.go
+++ b/pkg/api/v1beta2/dynakube/properties.go
@@ -211,14 +211,16 @@ func (dk *DynaKube) PullSecretName() string {
 	return dk.Name + PullSecretSuffix
 }
 
-// PullSecretWithoutData returns a secret which can be used to query the actual secrets data from the cluster.
-func (dk *DynaKube) PullSecretWithoutData() corev1.Secret {
-	return corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      dk.PullSecretName(),
-			Namespace: dk.Namespace,
-		},
+// PullSecretsNames returns the names of the pull secrets to be used for immutable images.
+func (dk *DynaKube) PullSecretsNames() []string {
+	names := []string{
+		dk.Name + PullSecretSuffix,
 	}
+	if dk.Spec.CustomPullSecret != "" {
+		names = append(names, dk.Spec.CustomPullSecret)
+	}
+
+	return names
 }
 
 func (dk *DynaKube) NeedsReadOnlyOneAgents() bool {
--- a/pkg/api/v1beta2/dynakube/properties.go
+++ b/pkg/api/v1beta2/dynakube/properties.go
@@ -223,6 +223,17 @@ func (dk *DynaKube) PullSecretsNames() []string {
 	return names
 }
 
+func (dk *DynaKube) ImagePullSecretReferences() []corev1.LocalObjectReference {
+	imagePullSecrets := make([]corev1.LocalObjectReference, 0)
+	for _, pullSecretName := range dk.PullSecretsNames() {
+		imagePullSecrets = append(imagePullSecrets, corev1.LocalObjectReference{
+			Name: pullSecretName,
+		})
+	}
+
+	return imagePullSecrets
+}
+
 func (dk *DynaKube) NeedsReadOnlyOneAgents() bool {
 	return dk.HostMonitoringMode() || dk.CloudNativeFullstackMode()
 }
--- a/pkg/api/v1beta2/dynakube/properties.go
+++ b/pkg/api/v1beta2/dynakube/properties.go
@@ -211,8 +211,8 @@ func (dk *DynaKube) PullSecretName() string {
 	return dk.Name + PullSecretSuffix
 }
 
-// PullSecretsNames returns the names of the pull secrets to be used for immutable images.
-func (dk *DynaKube) PullSecretsNames() []string {
+// PullSecretNames returns the names of the pull secrets to be used for immutable images.
+func (dk *DynaKube) PullSecretNames() []string {
 	names := []string{
 		dk.Name + PullSecretSuffix,
 	}
@@ -225,7 +225,7 @@ func (dk *DynaKube) PullSecretsNames() []string {
 
 func (dk *DynaKube) ImagePullSecretReferences() []corev1.LocalObjectReference {
 	imagePullSecrets := make([]corev1.LocalObjectReference, 0)
-	for _, pullSecretName := range dk.PullSecretsNames() {
+	for _, pullSecretName := range dk.PullSecretNames() {
 		imagePullSecrets = append(imagePullSecrets, corev1.LocalObjectReference{
 			Name: pullSecretName,
 		})
--- a/pkg/api/v1beta3/dynakube/properties.go
+++ b/pkg/api/v1beta3/dynakube/properties.go
@@ -212,7 +212,7 @@ func (dk *DynaKube) PullSecretName() string {
 }
 
 // PullSecretsNames returns the names of the pull secrets to be used for immutable images.
-func (dk *DynaKube) PullSecretsNames() []string {
+func (dk *DynaKube) PullSecretNames() []string {
 	names := []string{
 		dk.Name + PullSecretSuffix,
 	}
@@ -225,7 +225,7 @@ func (dk *DynaKube) PullSecretsNames() []string {
 
 func (dk *DynaKube) ImagePullSecretReferences() []corev1.LocalObjectReference {
 	imagePullSecrets := make([]corev1.LocalObjectReference, 0)
-	for _, pullSecretName := range dk.PullSecretsNames() {
+	for _, pullSecretName := range dk.PullSecretNames() {
 		imagePullSecrets = append(imagePullSecrets, corev1.LocalObjectReference{
 			Name: pullSecretName,
 		})
--- a/pkg/api/v1beta3/dynakube/properties.go
+++ b/pkg/api/v1beta3/dynakube/properties.go
@@ -211,14 +211,27 @@ func (dk *DynaKube) PullSecretName() string {
 	return dk.Name + PullSecretSuffix
 }
 
-// PullSecretWithoutData returns a secret which can be used to query the actual secrets data from the cluster.
-func (dk *DynaKube) PullSecretWithoutData() corev1.Secret {
-	return corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      dk.PullSecretName(),
-			Namespace: dk.Namespace,
-		},
+// PullSecretsNames returns the names of the pull secrets to be used for immutable images.
+func (dk *DynaKube) PullSecretsNames() []string {
+	names := []string{
+		dk.Name + PullSecretSuffix,
 	}
+	if dk.Spec.CustomPullSecret != "" {
+		names = append(names, dk.Spec.CustomPullSecret)
+	}
+
+	return names
+}
+
+func (dk *DynaKube) ImagePullSecretReferences() []corev1.LocalObjectReference {
+	imagePullSecrets := make([]corev1.LocalObjectReference, 0)
+	for _, pullSecretName := range dk.PullSecretsNames() {
+		imagePullSecrets = append(imagePullSecrets, corev1.LocalObjectReference{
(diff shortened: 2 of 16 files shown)


pkg/oci/dockerkeychain/docker_keychain.go
```diff
@@ -20,6 +21,51 @@ type DockerKeychain struct {
...
+		AuthConfigs: make(map[string]dockertypes.AuthConfig),
+	}
+
+	for _, pullSecretName := range pullSecretsNames {
+		pullSecret := corev1.Secret{}
+
+		if err := apiReader.Get(ctx, client.ObjectKey{Namespace: namespaceName, Name: pullSecretName}, &pullSecret); err != nil {
+			log.Info("failed to load registry pull secret", "name", pullSecretName, "namespace", namespaceName)
+
+			return errors.WithStack(err)
```
luhi-DT, line 46: Please do not rerturn an error here -> https://github.com/Dynatrace/dynatrace-operator/pull/3407

StefanHauth:
I tested this PR by pulling a code module image from the Google Cloud Artifactory Repository. Worked fine!

StefanHauth:
I think the troubleshoot command might also need some adoption in a follow-up taking these changes into account.

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -307,7 +307,14 @@ func (b *builder) imagePullSecrets() []corev1.LocalObjectReference {
 		return []corev1.LocalObjectReference{}
 	}
 
-	return []corev1.LocalObjectReference{{Name: b.dk.PullSecretName()}}
+	imagePullSecrets := make([]corev1.LocalObjectReference, 0)
```
StefanHauth, line 310: Same code snippet is used in StatefulSet creation. Maybe worth putting somewhere in a comon place.

pkg/oci/dockerkeychain/docker_keychain.go
```diff
@@ -20,6 +21,54 @@ type DockerKeychain struct {
...
+		}
+
+		dockerAuths, err := extractDockerAuthsFromSecret(&pullSecret)
+		if err != nil {
+			log.Info("failed to parse pull secret content", "name", pullSecret.Name, "namespace", pullSecret.Namespace)
+
+			return err
+		}
+
+		err = configFile.LoadFromReader(bytes.NewReader(dockerAuths))
```
StefanHauth, line 56: Food for thought: If multiple pull secrets contain credentials for the same registry, the last one wins, right?  How is the order determined and can we run into unexpected behaviour?

StefanHauth (changes requested):

pkg/oci/dockerkeychain/docker_keychain.go
luhi-DT: I would call this user error to be honest, the other solution would be to compare them one by one by unpacking all of the pull secrets

pkg/injection/codemodule/installer/image/installer.go
```diff
@@ -30,15 +30,14 @@ type Properties struct {
...
-	pullSecret := props.Dynakube.PullSecretWithoutData()
 	defaultTransport := http.DefaultTransport.(*http.Transport).Clone()
 
 	transport, err := registry.PrepareTransportForDynaKube(ctx, props.ApiReader, defaultTransport, props.Dynakube)
 	if err != nil {
 		return nil, err
 	}
 
-	keychain, err := dockerkeychain.NewDockerKeychain(ctx, props.ApiReader, pullSecret)
+	keychain, err := dockerkeychain.NewDockerKeychains(ctx, props.ApiReader, props.Dynakube.Namespace, props.Dynakube.PullSecretsNames())
```
luhi-DT, line 40: you could pass the dynakube here instead of 2 parameters, since you can access PullSecrets and Namespace inside the function

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
0sewa0:
```
func (dk *DynaKube) ImagePullSecrets() []corev1.LocalObjectReference {
  ...
}
```

pkg/oci/dockerkeychain/docker_keychain.go
aorcholski: `PullSecretNames` function returns list of secrets in the following order: [dynakube-pull-secret , customPullSecret]. 
dynakube-pull-secret docker config is put into the map and then the one from customPullSecret overwriting the previous value. So customPullSecret has higher priority. 
IMHO customPullSecret for the tenant registry is unexpected scenario. In such a case download will fail or not if docker config is copied from dynakube-pull-secret.

pkg/injection/codemodule/installer/image/installer.go
0sewa0: you shouldn't, that will make it bound to the DynaKube, what if you want to use it for EdgeConnect?
or in a case where you don't have the whole DynaKube?

passing around the DynaKube is nice and easy, but lets not make it become a kind of a "singleton" that we pass around in every helper package and function

0sewa0: you could do some renaming, there could be a `dockerkeychain.New` (so we don't stutter) where you request just what you need, and then introduce a `dockerkeychain.NewForDynaKube()`, where you would do what @luhi-DT was requesting. (but IMO `dockerkeychain.NewForDynaKube()` feels a bit of an overkill, but up to you)

pkg/api/v1beta2/dynakube/properties.go
```diff
@@ -211,14 +211,27 @@ func (dk *DynaKube) PullSecretName() string {
...
 
-// PullSecretWithoutData returns a secret which can be used to query the actual secrets data from the cluster.
-func (dk *DynaKube) PullSecretWithoutData() corev1.Secret {
-	return corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      dk.PullSecretName(),
-			Namespace: dk.Namespace,
-		},
+// PullSecretsNames returns the names of the pull secrets to be used for immutable images.
+func (dk *DynaKube) PullSecretsNames() []string {
```
albertogdd, line 215:
```suggestion
// PullSecretNames returns the names of the pull secrets to be used for immutable images.
func (dk *DynaKube) PullSecretNames() []string {
```
IMO it should only be plural at the end

pkg/api/v1beta3/dynakube/properties.go
```diff
@@ -211,14 +211,27 @@ func (dk *DynaKube) PullSecretName() string {
...
 
-// PullSecretWithoutData returns a secret which can be used to query the actual secrets data from the cluster.
-func (dk *DynaKube) PullSecretWithoutData() corev1.Secret {
-	return corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      dk.PullSecretName(),
-			Namespace: dk.Namespace,
-		},
+// PullSecretsNames returns the names of the pull secrets to be used for immutable images.
+func (dk *DynaKube) PullSecretsNames() []string {
```
albertogdd, line 215:
```suggestion
// PullSecretNames returns the names of the pull secrets to be used for immutable images.
func (dk *DynaKube) PullSecretNames() []string {
```
Same here

pkg/oci/dockerkeychain/docker_keychain.go
```diff
@@ -20,6 +21,54 @@ type DockerKeychain struct {
...
+	err := keychain.loadDockerConfigFromSecrets(ctx, apiReader, namespaceName, pullSecretsNames)
+
+	return keychain, err
+}
+
+func (keychain *DockerKeychain) loadDockerConfigFromSecrets(ctx context.Context, apiReader client.Reader, namespaceName string, pullSecretsNames []string) error {
+	if len(pullSecretsNames) == 0 {
+		return nil
+	}
+
```
albertogdd, line 35: You don't need two functions here; you can put them together in a single one (only `NewDockerKeychains` is referenced across the code)

albertogdd (changes requested):
LGTM, small requested style changes

pkg/oci/dockerkeychain/docker_keychain.go
aorcholski: NewDockerKeychain is used by `troubleshoot` command

albertogdd: That's what I mean! You can remove `loadDockerConfigFromSecrets()` and put its implementation inside `NewDockerKeychains()`

StefanHauth (approved):

PR #3373, opened 2024-06-28
Create new v1beta3 dynakube CRD version

andriisoldatenko:
## Description

Issue [K8S-10079](https://dt-rnd.atlassian.net/browse/K8S-10079)

This PR introduce new dynakube version `v1beta3` but we still gonna store everything as `v1beta2` (as hub):

```mermaid
flowchart LR

A[v1beta1] -->B(v1beta2 HUB)
C[v1beta3]-->B(v1beta2 HUB)
```

During converation between (hub) version `v1beta2` and new version `v1beta3` i'm using annotations to store new attributes from `.spec`.

## How can this be tested?

- Deploy v1beta1 make sure all working:
```

```
- Deploy v1beta2 make sure all working

- Deploy v1beta3 make sure all working (check that you have annotations; and inside logs you see that extensions reconciled)


pkg/api/v1beta3/dynakube/dynakube_types.go
```diff
@@ -0,0 +1,175 @@
...
+	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// Set custom proxy settings either directly or from a secret with the field proxy.
+	// Note: Applies to Dynatrace Operator, ActiveGate, and OneAgents.
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Proxy",order=3,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	Proxy *DynaKubeProxy `json:"proxy,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	ExtensionExecutionController ExtensionExecutionControllerSpec `json:"extensionExecutionController,omitempty"`
```
StefanHauth, line 92: This element should be in a `templates` subsection, not directly on this level.

pkg/api/v1beta3/dynakube/dynakube_types.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Defines a custom pull secret in case you use a private registry when pulling images from the Dynatrace environment.
+	// To define a custom pull secret and learn about the expected behavior, see Configure customPullSecret
+	// (https://www.dynatrace.com/support/help/setup-and-configuration/setup-on-container-platforms/kubernetes/get-started-with-kubernetes-monitoring/dto-config-options-k8s#custompullsecret).
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Custom PullSecret",order=8,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:io.kubernetes:Secret"}
+	CustomPullSecret string `json:"customPullSecret,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	OpenTelemetryCollector OpenTelemetryCollectorSpec `json:"openTelemetryCollector,omitempty"`
```
StefanHauth, line 129: This element should be in a `templates` subsection, not directly on this level.

pkg/api/v1beta3/dynakube/dynakube_types.go
andriisoldatenko: fixed.

pkg/api/v1beta3/dynakube/dynakube_types.go
andriisoldatenko: fixed.

hack/make/manifests/config.mk
```diff
@@ -1,4 +1,4 @@
-CRD_OPTIONS ?= "crd:crdVersions=v1"
```
chrismuellner, line 1: you are removing the description for all CRDs, even edgeconnect, and therefore making the generated API ref incorrect

is this really the best way to handle this problem?

aorcholski: Dynakube descriptions are truncated only

albertogdd (dismissed):
Tested different dynakubes api versions in my cluster and worked fine
PR is huge to review thoroughly... IMO is ok to merge and we will find out if something doesn't work

pkg/api/v1beta3/dynakube/extensions.go
```diff
@@ -0,0 +1,104 @@
...
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// Adds TopologySpreadConstraints for the OtelCollector pods
+	// +kubebuilder:validation:Optional
+	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+
+	// Number of replicas for your OtelCollector
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=1
+	Replicas int32 `json:"replicas,omitempty"`
```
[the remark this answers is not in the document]
0sewa0: general rule of thumb, don't use `omitempty` and `kubebuilder:default` at the same time, because then you won't be able to set a value to its "default-golang-value"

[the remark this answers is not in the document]
aorcholski: `omitempty` removed. The value can be set to 0. The property has the default value if not provided by a user.

StefanHauth (approved):

0sewa0 (approved):
The image wasn't built properly for some reason, as I got the following: 
```
Error from server (InternalError): error when creating "./.vscode/deploy/dynakube_zib50933.yaml": Internal error occurred: failed calling webhook "webhook.dynatrace.com": failed to call webhook: the server could not find the requested resource
```
(the validation move was only partly in)

but I built it locally and it works (deployed v1beta1, and v1beta2)

hack/make/manifests/config.mk
StefanHauth: This will for now only affect dev. Before the next release we will take care that `v1beta3` can't be used if it's not ready and that descriptions of `v1beta2` will be fully available.

albertogdd (approved):

PR #3376, opened 2024-06-28
Move dynakube validation logic into admission.CustomValidator

0sewa0:
## Description

[K8S-10054](https://dt-rnd.atlassian.net/browse/K8S-10054)

"Convert" the dynakube validation logic to follow the `admission.CustomValidator` interface , and moves it to the dynakube's api folder that we actually validate, so in the future we can have validation logic for all supported api versions of a CRD.

90% of the logic is the same, as it just moved, and makes it that it works according to the interface. (+some renames)

In 30baa56807603dcfa9519241fce973fa7cfb1693 I remove a check from the troubleshoot script as (IMO) it is useless, and just makes the code messier. (why have a check in the troubleshoot script that should have been caught by the validation webhook???)

## How can this be tested?

Deploy all kinds of `DynaKubes`, broken and not broken ones. (the unit tests didn't need major changes, which is a good sign)


wepudt (approved):
lgtm

deployed several DKs with different versions and configs, also misconfigured.

code also lgtm

albertogdd (approved):
Alles in Ordnung

PR #3383, opened 2024-07-01
Make csidriver updateStrategy configurable

0sewa0:
## Description
https://dt-rnd.atlassian.net/browse/K8S-9601

Make the `updateStrategy` part of the csi DaemonSet configurable via Helm.

## How can this be tested?

`make deploy/helm` should use the defaults set in the values.yaml
- play around with your own values.yaml

### DIFF
--- a/config/helm/chart/default/templates/Common/csi/daemonset.yaml
+++ b/config/helm/chart/default/templates/Common/csi/daemonset.yaml
@@ -272,7 +272,5 @@ spec:
           operator: Exists
           effect: NoSchedule
   updateStrategy:
-    rollingUpdate:
-      maxUnavailable: 1
-    type: RollingUpdate
+    {{- toYaml .Values.csidriver.updateStrategy | nindent 4 }}
 {{- end -}}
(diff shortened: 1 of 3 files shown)


luhi-DT (approved):
Tested, works for me!

config/helm/chart/default/templates/Common/csi/daemonset.yaml
```diff
@@ -272,7 +272,5 @@ spec:
           operator: Exists
           effect: NoSchedule
   updateStrategy:
```
chrismuellner, line 274: when setting `updateStrategy: ` or `updateStrategy: {}` in the values file I get the following generated yaml for the daemonset:
```yaml
  updateStrategy:
    null
```
and
```yaml
  updateStrategy:
    {}
```

They both seem to lead to the following upgrade strategy in the actual daemonset: 
```yaml
  updateStrategy:
    rollingUpdate:
      maxSurge: 0
      maxUnavailable: 1
    type: RollingUpdate
```

This is functionally equivalent to what we are explicitly setting (default value of `maxSurge` is 0), so my question would be: do we even need the default value in the values file?

0sewa0: I prefer to have the defaults set explicitly

0sewa0: can do

chrismuellner (approved):

PR #3384, opened 2024-07-01
Use -s param instead of "partial: ..." to generate partial manifests using Helm

0sewa0:
## Description

[K8S-9602](https://dt-rnd.atlassian.net/browse/K8S-9602)

We want to generate only given parts of our Helm-chart for the release, and we used to do it by setting a "secret" param called `partial`, which led to a very overcomplicated and confusing logic.

I replaced it with using the `-s` (`--show-only`) in the makefiles instead.
[official docu:](https://helm.sh/docs/helm/helm_template/)
```
 -s, --show-only stringArray                      only show manifests rendered from the given templates
```

Sidenote: I tried to make it "prettier" by trying not to list each CSI-related file 1-by-1, but didn't find a way

## How can this be tested?
Try out all helm/manifests related commands 😅 (no change in behaviour should be expected)
(Important to check not only the directly touched make commands, because all yaml files were touched do to the removal of the `partial` check)


hack/make/manifests/crd.mk
```diff
@@ -18,7 +18,5 @@ manifests/crd/helm: prerequisites/kustomize helm/version manifests/crd/generate
 manifests/crd/release: manifests/crd/helm
 	helm template dynatrace-operator config/helm/chart/default \
 			--namespace dynatrace \
-			--set installCRD=true \
-			--set csidriver.enabled=false \
-			--set partial="crd" > $(RELEASE_CRD_YAML)
+			-s templates/Common/crd/dynatrace-operator-crd.yaml > $(RELEASE_CRD_YAML)
```
chrismuellner, line 21: we shouldn't use the shorthand version here, as `--show-only` makes it clearer what is happening.

albertogdd (changes requested):
Everything else LGTM

hack/make/manifests/openshift.mk
```diff
@@ -2,8 +2,13 @@
 manifests/openshift/csi:
```
chrismuellner, line 2: unrelated, but why does the `Common` folder start with an uppercase letter?

hack/make/manifests/kubernetes.mk
```diff
@@ -2,8 +2,13 @@
 manifests/kubernetes/csi:
 	# Generate kubernetes-csi.yaml
 	helm template dynatrace-operator config/helm/chart/default \
+		-s templates/Common/csi/clusterrole-csi.yaml \
+		-s templates/Common/csi/csidriver.yaml \
+		-s templates/Common/csi/daemonset.yaml \
+		-s templates/Common/csi/priority-class.yaml \
+		-s templates/Common/csi/role-csi.yaml \
+		-s templates/Common/csi/serviceaccount-csi.yaml \
```
chrismuellner, line 10: it seems the `--show-only` parameter supports regex, so we could just simplify this (and make it harder to forget to add new files) by adding all files in the `templates/common/csi` folder.

```suggestion
		-s templates/Common/csi/*.yaml \
```

chrismuellner (changes requested):

hack/make/manifests/openshift.mk
0sewa0: no clue; too afraid to change it now 😅 (maybe related to GKE marketplace 🤷 )

config/helm/chart/default/templates/Common/crd/dynatrace-operator-crd.yaml
```diff
@@ -1,5 +1,5 @@
 {{- include "dynatrace-operator.platformRequired" . }}
-{{ if and .Values.installCRD (or (eq (include "dynatrace-operator.partial" .) "false") (eq (include "dynatrace-operator.partial" .) "crd")) }}
+{{ if .Values.installCRD }}
```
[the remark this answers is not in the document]
0sewa0: done

hack/make/manifests/crd.mk
0sewa0: done

hack/make/manifests/kubernetes.mk
0sewa0: done

chrismuellner: I don't see this change in the latest commit, you only changed `-s` to `--show-only`

0sewa0: sorry, can't read 🙈 

also you fixed my sidenote: 😘  ❤️  
> Sidenote: I tried to make it "prettier" by trying not to list each CSI-related file 1-by-1, but didn't find a way (it's not pretty, but still much prettier than partial)

hack/make/manifests/openshift.mk
```diff
@@ -2,8 +2,13 @@
...
 	# Generate openshift-csi.yaml
 	helm template dynatrace-operator config/helm/chart/default \
+		--show-only templates/Common/csi/clusterrole-csi.yaml \
+		--show-only templates/Common/csi/csidriver.yaml \
+		--show-only templates/Common/csi/daemonset.yaml \
+		--show-only templates/Common/csi/priority-class.yaml \
+		--show-only templates/Common/csi/role-csi.yaml \
+		--show-only templates/Common/csi/serviceaccount-csi.yaml \
```
chrismuellner, line 10:
```suggestion
		--show-only templates/Common/csi/*.yaml \
```

chrismuellner (approved):

albertogdd (approved):

PR #3327, opened 2024-06-17
Add enrichment-rules to client and dynakube

0sewa0:
## Description

[K8S-9940](https://dt-rnd.atlassian.net/browse/K8S-9940)

Using the settings API it stores the metadata-enrichment rules in the `DynaKube` when `metadata-enrichment` is enabled.

- It only checks for rules only every `ApiRequestThreshold`
- It uses conditions to store when the last request was done
- If the `metadata-enrichment` feature gets turned off, the status is cleared

Sidenote:
- Tried to restructure the Settings API related logic in a less messy way, for (hopefully) better readability.

## How can this be tested?
0. On your tenant create rules, you need a feature-flag for it to be available, see details in: [K8S-9128 ](https://dt-rnd.atlassian.net/browse/K8S-9128?focusedCommentId=3336492)
   - You need to create the rules for a Kubernetes-cluster monitored-entity, rules defined at the Environment level will no show up (as of right now)
   - You should have a page like this: (follow the "breadcrumbs" in the top left)
image

2. Deploy a `DynaKube` with `metadataEnrichment.enabled: true`
3. Check the status for the rules:
```
apiVersion: dynatrace.com/v1beta2
kind: DynaKube
metadata: ...
spec:
  metadataEnrichment:
    enabled: true
status:
...
  conditions:
  - lastTransitionTime: "2024-06-17T13:38:41Z"
    message: Metadata-enrichment rules are up-to-date in the status
    reason: StatusUpdated
    status: "True"
    type: MetadataEnrichmentRules
  metadataEnrichment:
    rules:
    - enabled: true
      key: cost-center
      mapping: dt.cost.costcenter
      type: LABEL
    - enabled: true
      key: security-context
      mapping: dt.security_context
      type: ANNOTATION
```

PR #3386, opened 2024-07-01
Propagate annotations from namespace to pods for metadata enrichment

gkrenn:
## Description

In a previous PR we already fetch custom rules that define which namespace annotations and labels should be copied to the pod. This PR actually copies the metadata.

[Jira](https://dt-rnd.atlassian.net/browse/K8S-9942)

## How can this be tested?

- Setup in the same way as in #3327. 
- Add labels to the namespace
- Create dynakube with enabled metadata enrichment
- Create test pod and check if metadata enrichment annotations exist


pkg/webhook/mutation/pod/metadata/mutator.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"strings"
 
+	dynatracev1beta2 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube"
```
chrismuellner, line 7:
```suggestion
	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube"
```

pkg/webhook/mutation/pod/metadata/mutator.go
```diff
@@ -141,6 +143,48 @@ func setWorkloadAnnotations(pod *corev1.Pod, workload *workloadInfo) {
...
+			setPodAnnotationIfNotExists(pod, key, value)
+		}
+	}
+}
+
+func copyMetadataAccordingToCustomRules(pod *corev1.Pod, namespace corev1.Namespace, dynakube dynatracev1beta2.DynaKube) {
+	for _, rule := range dynakube.Status.MetadataEnrichment.Rules {
+		var valueFromNamespace string
+
+		var exists bool
```
chrismuellner, line 163: do you need this bool or is it enough and clearer to check if `valueFromNamespace` is not empty?

gkrenn: I would keep it yes, otherwise you would swallow explicitly set empty values

pkg/webhook/mutation/pod/metadata/mutator.go
gkrenn: Done

luhi-DT (approved):
lgtm

PR #3387, opened 2024-07-02
Fix failing initContainer incase of classicFullStack and metadataEnrichment combination

luhi-DT:
Ticket: https://dt-rnd.atlassian.net/browse/K8S-10259

## Description

Since we want to combine classicFullStack and metaDataEnrichment, combining those two will cause an error -> the agent mounts over /var/lib/dynatrace/enrichment, causing the initContainer to fail with createFile errors (lack of permission) on the hostFileSystem.
We should instead of directly creating the files there, create them in a temp location and mount the with subPath (only files not the whole folder) 
Afterwards the agent will make modifications to not overwrite our files

## How can this be tested?

Apply a Dynakube with metadata enrichment only and check if all files (.json and .properties are created)
Apply another Dynakube afterwards with classicFullStack and metadataEnrichment to check if the containers start successfully (note: the files will still be gone, but the agent has a ticket to resolve this)


pkg/consts/enrichment_injection.go
```diff
@@ -2,12 +2,14 @@ package consts
...
 	EnrichmentInjectedEnv        = "METADATA_ENRICHMENT_INJECTED"
 	EnrichmentWorkloadKindEnv    = "DT_WORKLOAD_KIND"
 	EnrichmentWorkloadNameEnv    = "DT_WORKLOAD_NAME"
+	EnrichmentPropertiesFilename = "dt_metadata.properties"
+	EnrichmentJsonFilename       = "dt_metadata.json"
 )
 
 var (
 	EnrichmentMountPath = "/var/lib/dynatrace/enrichment"
+	EnrichmentInitPath  = "/tmp/enrichment"
```
0sewa0, line 14: why aren't these `const`?

luhi-DT: Not sure, did not want to change them, but IMHO they should be consts, good point

luhi-DT: done

0sewa0 (approved):
tested it, deployed sample apps didn't have problems starting

waodim (approved):
tested it, got no problems applying sample apps when reapplied a classicfullstack dk

PR #3388, opened 2024-07-02
Block osmount incase unmount is still in progress

0sewa0:
## Description

[K8S-10338](https://dt-rnd.atlassian.net/browse/K8S-10338)

[Adds a bunch of safeguards](https://github.com/Dynatrace/dynatrace-operator/pull/3388/commits/5aa5128aa7fc057747cbae8edbdf9b5404f8e942) to the csi-gc to never-ever remove things that are mounted, however, this was not the source of the problem, but they shouldn't hurt.

[The real source of the issue](https://github.com/Dynatrace/dynatrace-operator/pull/3388/commits/bb1a4fe96e0e186beb10b4fc51894237b3cb177e) was that in the scenario(can happen during agent upgrade) when the Terminating OsAgent has not been unmounted, but the OsAgent replacing it already needs the csi-volume, we basically "did nothing", which is basically telling `kubelet` that we did our job, but in reality we didn't mount anything.
- This caused the problem that the "new" OsAgent couldn't Terminate when deleted. (also it didn't have the volume it needed, but that didn't cause an fail of the pod for some reason)
	- Sidenote: with "normal" workloads there was no problem with Terminating (tested it with a few sample apps)

## How can this be tested?

Run a command like: (its timing based, so it might not happen on the first try)
```
<deploy-dynakube-with-cloudnative> && sleep 5 && <delete-dynakube> && sleep 5 && <deploy-dynakube-with-cloudnative>
```
This should reproduce the original issue, and you should see logs in the csi's `server`:
```
previous OSMount is yet to be unmounted, there can be only 1 OSMount per tenant per node, blocking until unmount
```
No OneAgents should be stuck at the end.


albertogdd (approved):
Works. Still the pods get stuck sometimes when running `make undeploy`, but at least when deleting applying dynakubes seems to be more stable now.

luhi-DT:
> Works. Still the pods get stuck sometimes when running `make undeploy`, but at least when deleting applying dynakubes seems to be more stable now.

Make undeploy does not do it correctly i guess, since it deletes CSI Driver immediatly and does not wait for dynakube and oneagent pods to be terminated

PR #3405, opened 2024-07-04
Add Verification of OneAgent Version

wepudt:
see [K8S-10190](https://dt-rnd.atlassian.net/browse/K8S-10190)

## Description

The operator should reject OneAgent versions that are not conformant to major.minor.patch pattern subset of SemVer. e.g. 1.0.0

## How can this be tested?

Apply a DK with OneAgent version set to an invalid version such as `latest`, `raw` or `1.296.0-raw` and to a valid one such as 1.296.0

### DIFF
--- a/pkg/api/v1beta2/dynakube/validation/oneagent.go
+++ b/pkg/api/v1beta2/dynakube/validation/oneagent.go
@@ -3,6 +3,7 @@ package validation
 import (
 	"context"
 	"fmt"
+	"regexp"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/env"
@@ -135,6 +136,21 @@ func conflictingOneAgentVolumeStorageSettings(_ context.Context, _ *Validator, d
 func conflictingHostGroupSettings(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
 	if dk.HostGroupAsParam() != "" {
 		return warningHostGroupConflict
+	}
+
+	return ""
+}
+
+var oneAgentSemVerVersionSpec = regexp.MustCompile(`^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)$`)
+
+func validateOneAgentVersionIsSemVerCompliant(_ context.Context, _ *dynakubeValidator, dk *dynakube.DynaKube) string {
+	agentVersion := dk.CustomOneAgentVersion()
+	if agentVersion == "" {
+		return ""
+	}
+
+	if !oneAgentSemVerVersionSpec.MatchString(agentVersion) {
+		return "Only semantic versions are allowed!"
 	}
 
 	return ""
--- a/pkg/api/v1beta2/dynakube/validation/oneagent.go
+++ b/pkg/api/v1beta2/dynakube/validation/oneagent.go
@@ -3,10 +3,11 @@ package validation
 import (
 	"context"
 	"fmt"
-	"regexp"
+	"strings"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/env"
+	"golang.org/x/mod/semver"
 	"k8s.io/apimachinery/pkg/labels"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/webhook/validation/dynakube/oneagent.go
```diff
@@ -106,36 +107,51 @@ func hasConflictingMatchLabels(labelMap, otherLabelMap map[string]string) bool {
...
-	if dynakube.HostGroupAsParam() != "" {
+func conflictingHostGroupSettings(_ context.Context, _ *dynakubeValidator, dk *dynakube.DynaKube) string {
+	if dk.HostGroupAsParam() != "" {
 		return warningHostGroupConflict
 	}
 
 	return ""
 }
+
+var oneAgentSemVerVersionSpec = regexp.MustCompile(`^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)$`)
```
0sewa0, line 144: do we need this?

the `semver` package has a `IsValid` func.
the only downside of it is that `semver` always has a `v` prefix.

we already have some implementation that tries to handle this here: https://github.com/Dynatrace/dynatrace-operator/blob/5aa5128aa7fc057747cbae8edbdf9b5404f8e942/pkg/util/dtversion/dtversion.go#L26

wepudt: unfortunately we do in my opinion.

reason: we want to allow only a subset of SemVer: major.minor.patch without prerelease or build parts that would be allowed due to SemVer spec.

To technically speaking it is not SemVer at all 😄

pkg/webhook/validation/dynakube/oneagent.go
```diff
@@ -106,36 +107,51 @@ func hasConflictingMatchLabels(labelMap, otherLabelMap map[string]string) bool {
...
+
+func validateOneAgentVersionIsSemVerCompliant(_ context.Context, _ *dynakubeValidator, dk *dynakube.DynaKube) string {
+	agentVersion := dk.CustomOneAgentVersion()
+	if agentVersion == "" {
+		return ""
+	}
+
+	if !oneAgentSemVerVersionSpec.MatchString(agentVersion) {
```
aorcholski, line 152:
```suggestion
        version := "v"+agentVersion
	if !(semver.IsValid(version) && semver.Prerelease(version) == "" && semver.Build(version) == "" && len(strings.Split(version, ".")) == 3) {
```

pkg/webhook/validation/dynakube/oneagent_test.go
```diff
@@ -424,3 +424,41 @@ func createDynakubeWithHostGroup(args []string, hostGroup string) *dynatracev1be
 		},
 	}
 }
+
+func TestValidateOneAgentVersionIsSemVer(t *testing.T) {
+	testCasesAcceptedVersions := []string{"", "1.0.0", "1.200.1"}
+
+	testCasesNotAcceptedVersions := []string{"latest", "raw", "1.200.1-raw"}
```
aorcholski, line 431:
```suggestion
	testCasesNotAcceptedVersions := []string{"latest", "raw", "1.200.1-raw", "1.200.1+build", "1.200", "1", "1.0", "v1.200.0"}
```

pkg/webhook/validation/dynakube/oneagent.go
wepudt: That would definitely work.
I was disliking all the operations when matching a regex could get the job done, but it boils down preference- it's up to guys how you want to do it...

pkg/webhook/validation/dynakube/oneagent.go
wepudt: changed to Adams suggestion in 3d3c8add5a6a1bb9ef396db5a088d07373068bbe

pkg/webhook/validation/dynakube/oneagent.go
wepudt: fixed in 3d3c8add5a6a1bb9ef396db5a088d07373068bbe

pkg/webhook/validation/dynakube/oneagent_test.go
wepudt: fixed in 8784b8194be7c36fec3bafd5de700098203ad796

aorcholski (dismissed):
LGTM

pkg/webhook/validation/dynakube/oneagent_test.go
```diff
@@ -424,3 +424,41 @@ func createDynakubeWithHostGroup(args []string, hostGroup string) *dynatracev1be
...
+			})
+		})
+	}
+
+	for _, tc := range testCasesNotAcceptedVersions {
+		t.Run("should accept version "+tc, func(t *testing.T) {
+			assertDeniedResponse(t, []string{"Only semantic versions in the form of major.minor.patch (e.g. 1.0.0) are allowed!"}, &dynatracev1beta2.DynaKube{
+				ObjectMeta: defaultDynakubeObjectMeta,
+				Spec: dynatracev1beta2.DynaKubeSpec{
+					APIURL: testApiUrl,
```
luhi-DT, line 454: since you touched parts of the file, would it make senes to change the imports in this file aswell? so its dynakube instead of dynatracev1beta2?

luhi-DT (approved):

albertogdd (approved):
LGTM

PR #3429, opened 2024-07-09
Fix reconcile interval incase of an error when updating the agent in the provisioner

luhi-DT:
## Description

Ticket: https://dt-rnd.atlassian.net/browse/K8S-10521

If the download of the codemodule fails, the csi-driver will not do a faster requeue (ie.: retry) to try to download the codemodule again. We should do a faster reconcile in such scenarios

## How can this be tested?

Deploy a Dynakube -> set a wrong image in the codeModulesImage field

### DIFF
--- a/pkg/controllers/csi/provisioner/controller.go
+++ b/pkg/controllers/csi/provisioner/controller.go
@@ -233,14 +233,14 @@ func (provisioner *OneAgentProvisioner) collectGarbage(ctx context.Context, requ
 	return result, nil
 }
 
-func (provisioner *OneAgentProvisioner) provisionCodeModules(ctx context.Context, dk *dynatracev1beta2.DynaKube, tenantConfig *metadata.TenantConfig) (bool, error) {
+func (provisioner *OneAgentProvisioner) provisionCodeModules(ctx context.Context, dk *dynatracev1beta2.DynaKube, tenantConfig *metadata.TenantConfig) (requeue bool, err error) {
 	// creates a dt client and checks tokens exist for the given dynakube
 	dtc, err := buildDtc(provisioner, ctx, dk)
 	if err != nil {
 		return true, err
 	}
 
-	requeue, err := provisioner.updateAgentInstallation(ctx, dtc, tenantConfig, dk)
+	requeue, err = provisioner.updateAgentInstallation(ctx, dtc, tenantConfig, dk)
 	if err != nil {
 		return requeue, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/controllers/csi/provisioner/controller.go
```diff
@@ -229,25 +233,25 @@ func (provisioner *OneAgentProvisioner) collectGarbage(ctx context.Context, requ
 	return result, nil
 }
 
-func (provisioner *OneAgentProvisioner) provisionCodeModules(ctx context.Context, dk *dynatracev1beta2.DynaKube, tenantConfig *metadata.TenantConfig) error {
+func (provisioner *OneAgentProvisioner) provisionCodeModules(ctx context.Context, dk *dynatracev1beta2.DynaKube, tenantConfig *metadata.TenantConfig) (bool, error) {
```
andriisoldatenko, line 236: I think since you changed the method's signature, and now it returns a bool, we should consider renaming it to something like `isProvisionCodeModules`. Then you magic constants like true / false will be more clear.

pkg/controllers/csi/provisioner/controller.go
```diff
@@ -129,8 +129,12 @@ func (provisioner *OneAgentProvisioner) Reconcile(ctx context.Context, request r
 		return reconcile.Result{RequeueAfter: dtcsi.ShortRequeueDuration}, err
 	}
 
-	err = provisioner.provisionCodeModules(ctx, dk, tenantConfig)
+	requeue, err := provisioner.provisionCodeModules(ctx, dk, tenantConfig)
```
andriisoldatenko, line 132:
```suggestion
	ok, err := provisioner.isProvisionCodeModules(ctx, dk, tenantConfig)
```

pkg/controllers/csi/provisioner/controller.go
```diff
@@ -129,8 +129,12 @@ func (provisioner *OneAgentProvisioner) Reconcile(ctx context.Context, request r
...
 	if err != nil {
+		if requeue {
```
andriisoldatenko, line 134:
```suggestion
		if !ok {
		log.Warnig("requiring ... something"
```

waodim: shouldn't it be `ok` and not `!ok`? or am i missing something here?

andriisoldatenko: may be :) it depends on func logic, it's just an idea.

pkg/controllers/csi/provisioner/controller.go
luhi-DT: well the bool indicates if we need a requeue or not in case of an error, which is something different imho than if the codeModules got provisioned successfully

luhi-DT: @andriisoldatenko Fixed with naming the return types :)

PR #3447, opened 2024-07-11
Add monitored entity reconciler

waodim:
## Description

With this a new reconciler is added that deals with the monitored entity from the cluster. It will be stored in the dynakube status from now on, such that if we need to store the value we can guarantee that it is stored somewhere.

Because of this I removed the API call which was inside the settings enrichment reconciler, as it is unneeded as the main purpose was to get the entityID. From now on the ID is simply taken from the dynakube status.

JIRA-ticket: https://dt-rnd.atlassian.net/browse/K8S-9857

## How can this be tested?

Deploy the operator and a dynakube with metadata enrichment enabled. Check the logs, also check the dynakube status.


pkg/controllers/dynakube/controller.go
```diff
@@ -331,6 +332,15 @@ func (controller *Controller) reconcileComponents(ctx context.Context, dynatrace
 		return err
 	}
 
+	log.Info("start reconciling monitored entities")
```
chrismuellner, line 335: what's our plan for these type of logs? Do they provide much benefit?

I'd expect a similar closing log as well that the reconcile finished successfully, if we have them, or only logs for interesting events during the reconcile, potentially with a logger name that provides more information.

pkg/controllers/dynakube/monitored_entities/reconciler.go
```diff
@@ -0,0 +1,71 @@
...
+
+			return nil
+		}
+
+		r.dk.Status.MonitoredEntityID = findLatestEntity(monitoredEntities).EntityId
+
+		log.Info("set the monitoredEntityID to the dynakube status")
+	}
+
+	log.Info("monitoredEntityID already set, done reconciling")
```
chrismuellner, line 57: could it help us to log the actual id instead of logging a bunch of similar statements that don't provide much debug value?

config/helm/chart/default/templates/Common/crd/dynatrace-operator-crd.yaml
```diff
@@ -4442,6 +4442,10 @@ spec:
                       type: object
                     type: array
                 type: object
+              monitoredEntityID:
```
luhi-DT, line 4445: Should we rename that to KubernetesClusterMEID?

pkg/controllers/dynakube/monitored_entities/reconciler.go
```diff
@@ -0,0 +1,71 @@
...
+
+		if len(monitoredEntities) == 0 {
+			log.Info("no MEs found, no monitoredentityID will be set in the dynakube status")
+
+			return nil
+		}
+
+		r.dk.Status.MonitoredEntityID = findLatestEntity(monitoredEntities).EntityId
+
+		log.Info("set the monitoredEntityID to the dynakube status")
```
luhi-DT, line 54: should you return here to not log twice?

waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/3d3d6a44b98a09ef8d9842a3ea3e263480209cbe

config/helm/chart/default/templates/Common/crd/dynatrace-operator-crd.yaml
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/3d3d6a44b98a09ef8d9842a3ea3e263480209cbe

pkg/controllers/dynakube/monitored_entities/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/3d3d6a44b98a09ef8d9842a3ea3e263480209cbe

pkg/controllers/dynakube/controller.go
waodim: I do not have a strong opinion here, I can happily add another log line on this level that tells us that the reconciliation is done. Currently I am doing that on the level of the monitored entities reconciler, so I could do that on the higher level as well. For me it is just important to have some kind of entry point that indicates that reconciliation started, in order to make debugging somewhat easier.

maybe the others could share their opinion here as well.

pkg/clients/dynatrace/settings_enrichment.go
```diff
@@ -29,20 +29,14 @@ type RulesResponseValue struct {
...
-		log.Info("failed to retrieve MEs")
-
-		return GetRulesSettingsResponse{}, err
-	}
-
-	if len(monitoredEntities) == 0 {
-		log.Info("no MEs found, skipping getting enrichment rules")
+	if entityID == "" {
+		// if monitored entities were empty this field also stays empty, we return with no error
+		log.Info("No Monitored Entity ID, skip loading/getting/querying enrichment rules")
```
chrismuellner, line 39: that was meant as suggestion for you to choose one verb, not put all of them 😄

waodim: fixed in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/7806dfbbe1fd6bfb933a74ecc937edf987f1c81b

pkg/controllers/dynakube/monitored_entities/reconciler.go
```diff
@@ -0,0 +1,69 @@
...
+
+func NewReconciler( //nolint
+	dtClient dynatrace.Client,
+	dk *dynakube.DynaKube,
+	kubeSystemUUID string,
+) controllers.Reconciler {
+	return &Reconciler{
+		dk:             dk,
+		dtClient:       dtClient,
+		kubeSystemUUID: kubeSystemUUID,
```
0sewa0, line 25: the dynakube already has the `kubeSystemUUID` in its status, not necessary to pass it in separately

pkg/controllers/dynakube/monitored_entities/reconciler.go
```diff
@@ -0,0 +1,69 @@
...
+			return err
+		}
+
+		log.Info("retrieved MEs")
+
+		if len(monitoredEntities) == 0 {
+			log.Info("no MEs found, no monitoredentityID will be set in the dynakube status")
+
+			return nil
+		}
```
0sewa0, line 50: should we care about the scenario when there is no monitored entity and shouldn't be one? is that a valid scenario ?

because if it is, then we will now "constantly"(ie.: every reconciliation) ping the tenant about it

pkg/controllers/dynakube/apimonitoring/reconciler.go
```diff
@@ -75,6 +75,8 @@ func (r *Reconciler) createObjectIdIfNotExists(ctx context.Context) (string, err
 		return "", errors.WithMessage(err, "error creating dynatrace settings object")
 	}
 
+	r.dk.Status.KubernetesClusterMEID = meID
```
0sewa0, line 78: you could simplify the logic above a bit, you are basically doing a monitored-entity reconcile, that is unnecessary if we already know the MEID.

you can just have a monitored-entity reconciler at the beginning as well, it will not run if we already know the MEID, so it will not conflict with any other run of the monitored-entity reconciler, and you can simplify the code here.

(also you can get the kubeSystemUUID from the dynakube)

pkg/controllers/dynakube/monitored_entities/config.go
```diff
@@ -0,0 +1,9 @@
+package monitoredentities
```
0sewa0, line 1: we don't have folders with snake-case, so we shouldn't start now, to remain consistent

0sewa0 (changes requested):

pkg/controllers/dynakube/monitored_entities/reconciler.go
waodim: no, you are right, thinking about it this should not be a valid scenario. Will remove this part.

pkg/controllers/dynakube/controller.go
0sewa0: I would do it inside the monitored-entity reconciler, before it does the actual work, so it should only show up once for a given dynakube

pkg/controllers/dynakube/monitored_entities/reconciler.go
0sewa0: if it's not valid, then we should probably error 😅

pkg/controllers/dynakube/monitored_entities/config.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/7e1e568e4e9045f69237947d9217d7ec22f8d380

pkg/controllers/dynakube/monitored_entities/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/7e1e568e4e9045f69237947d9217d7ec22f8d380

pkg/controllers/dynakube/monitored_entities/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/7e1e568e4e9045f69237947d9217d7ec22f8d380

pkg/controllers/dynakube/apimonitoring/reconciler.go
waodim: (hopefully) done in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/3459e905e7dbe1da30717fe4156a47e3064ace70

pkg/controllers/dynakube/monitoredentities/reconciler.go
```diff
@@ -0,0 +1,67 @@
...
+
+type ReconcilerBuilder func(
+	dtClient dynatrace.Client,
+	dk *dynakube.DynaKube,
+) controllers.Reconciler
+
+func NewReconciler( //nolint
+	dtClient dynatrace.Client,
+	dk *dynakube.DynaKube,
+	kubeSystemUUID string,
```
0sewa0, line 20: not needed

pkg/controllers/dynakube/monitoredentities/reconciler.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+type Reconciler struct {
+	dk       *dynakube.DynaKube
+	dtClient dynatrace.Client
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context) error {
+	log.Info("start reconciling monitored entities")
```
0sewa0, line 34: move it into the if so we don't spam the logs (would  be present during every reconcile)

pkg/controllers/dynakube/monitoredentities/reconciler.go
```diff
@@ -0,0 +1,67 @@
...
+		log.Info("retrieved MEs")
+
+		if len(monitoredEntities) == 0 {
+			return errors.New("MEs are empty, at this point this should not be the case")
+		}
+
+		r.dk.Status.KubernetesClusterMEID = findLatestEntity(monitoredEntities).EntityId
+	}
+
+	log.Info("kubernetesClusterMEID set in dynakube status, done reconciling", "kubernetesClusterMEID", r.dk.Status.KubernetesClusterMEID)
```
0sewa0, line 53: move it into the if or remove so we don't spam the logs (would  be present during every reconcile)

0sewa0 (changes requested):

pkg/controllers/dynakube/monitoredentities/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/8fcfcc3ee0a69d518855231cd6151675b1e180bb

pkg/controllers/dynakube/monitoredentities/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/8fcfcc3ee0a69d518855231cd6151675b1e180bb

pkg/controllers/dynakube/monitoredentities/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/8fcfcc3ee0a69d518855231cd6151675b1e180bb

0sewa0 (commented):
1 final nitpick, otherwise looks good, tested it and works as expected (noticed the nitpick during the tests)

0sewa0 (changes requested):
1 final nitpick, otherwise looks good, tested it and works as expected (noticed the nitpick during the tests)

pkg/controllers/dynakube/monitoredentities/reconciler.go
```diff
@@ -0,0 +1,66 @@
...
+		log.Info("start reconciling monitored entities")
+
+		monitoredEntities, err := r.dtClient.GetMonitoredEntitiesForKubeSystemUUID(ctx, r.dk.Status.KubeSystemUUID)
+		if err != nil {
+			log.Info("failed to retrieve MEs")
+
+			return err
+		}
+
+		log.Info("retrieved MEs")
```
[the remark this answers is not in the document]
waodim: nice catch. done in https://github.com/Dynatrace/dynatrace-operator/pull/3447/commits/5f90422b9eebea1f3d74ad4aaa1bc19e30f044e6

0sewa0 (approved):

0sewa0:
CRD needs to be regenerated:
```
> git status        
On branch feature/add-monitored-entity-reconciler
Your branch is up to date with 'origin/feature/add-monitored-entity-reconciler'.

  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   config/crd/bases/dynatrace.com_dynakubes.yaml

no changes added to commit (use "git add" and/or "git commit -a")
```

0sewa0 (approved):

luhi-DT (approved):
LGTM

PR #3453, opened 2024-07-11
Handle the force-delete scenario in the OS volume publisher

0sewa0:
## Description

If the OSAgents were removed forcefully, we might not get the unmount request, so we can't fully relay on the database, and have to directly check if its mounted or not.

## How can this be tested?

Deploy OSAgents, **force** delete 1, it should be able to come back up.


luhi-DT (dismissed):

luhi-DT (approved):

luhi-DT:
Please dont forget to backport to main and release-1.3 :) ty

waodim (approved):

PR #3469, opened 2024-07-16
Specify ns when listing dynakubes during validation

0sewa0:
## Description

During the move of the validation logic, I also made it that the validation webhook only uses a `client.Reader`, as it doesn't/shouldn't create anything.
A side effect of this was that: A `client.Client` will respect the namespace of the manager  you get it from when doing a `.List(...)` a`client.Reader`  does not

Also adding a `matchPolicy: Exact` to the validation configuration, this is needed because the default is `Equivalent`, which will cause that all validation webhooks will be called for a given CR regardless of the version, which we do not want. ([official docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchpolicy))
- Also removed the `v1alpha1/v1beta1` of the `admissionReviewVersions`, as they do nothing, and `v1` was thing since 2019, so we shouldn't be supporting versions not having the `v1` APIs 

## How can this be tested?
No errors in the webhook when applying dynakubes.
- The "nodeSelector" check was the problematic one, so the dynakube needs to be in a mode that can be configured with a `nodeSelector`


aorcholski (approved):
Applied v1beta2 and v1beta3 dynakubes. No errors

andriisoldatenko (approved):

gkrenn (approved):
Nice improvements 👍 

Checked with v1beta2 => correct validation is called exactly once

PR #3495, opened 2024-07-18
Mark v1beta1 and v1beta2 package as deprecated

gkrenn:
## Description

In order to prevent usage of old dynakube versions in code, this PR deprecates the whole package.

In addition the deprecation linter is enabled.

[Jira](https://dt-rnd.atlassian.net/browse/K8S-10057)

## How can this be tested?

- Check if linting fails if old dynakube is imported in your code.


0sewa0 (dismissed):

pkg/api/v1beta1/dynakube/doc.go
```diff
@@ -0,0 +1,3 @@
+// Deprecated: This package is deprecated and will be removed in future versions.
```
chrismuellner, line 1: What about `v1beta2`? Should that also be deprecated and no longer used in the code?

It would not impact anyone that still uses those versions, since we have conversion, but the code should be all `v1beta3`

gkrenn: done

0sewa0 (approved):

waodim (dismissed):
lgtm

0sewa0:
could we add the `kubebuilder:deprecatedversion` marker comment on the deprecated versions? (you would have to put it somewhere around [here](https://github.com/Dynatrace/dynatrace-operator/blob/e68f16919f1b14938b246621552758697cefb0df/pkg/api/v1beta2/dynakube/dynakube_types.go#L61-L70))

gkrenn:
Added the `kubebuilder` annotation and validated that it works as expected.

As described [here](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-deprecation), the deprecation field informs the kubeapi, that it should set a header on every request to inform the user about the deprecation. It is a non-blocking operation. 

The output of kubectl includes a warning that states the following: `dynatrace.com/v1beta2 DynaKube is deprecated; use dynatrace.com/v1beta3 DynaKube`.

gkrenn:
Removed deprecation again, as we first have to establish a timeline for deprecating/dropping old apis

0sewa0 (approved):

chrismuellner (approved):

PR #3518, opened 2024-07-23
Add k8s.cluster.name to enrichment + renames

0sewa0:
## Description
[K8S-9858](https://dt-rnd.atlassian.net/browse/K8S-9858)

Adds the `k8s.workload.*` renames:
- [Update init container to add new props alongside old ones](https://github.com/Dynatrace/dynatrace-operator/commit/974d02539901cf1bb000d2198f3cbf92b11f32da) 

Adds the `k8s.cluster.name` to the metadata enrichment files:
- [Updated the webhook to add a new ENV based on the dk.Status](https://github.com/Dynatrace/dynatrace-operator/commit/470a273356e9d4c23bfa3d1bf4974b293fa48c51)
  - Wanted to add it to the init-secret (why [this ](https://github.com/Dynatrace/dynatrace-operator/commit/8fbce49bd4cc1a87b62a0052a150fa9da90e29f7)was done), but that is mainly for the OA injection, so decided against it. 	
- [Updated the init container to use the new ENV ](https://github.com/Dynatrace/dynatrace-operator/commit/680f88386b397a62b8a562175b31c51827001325)

## How can this be tested?

1. Deploy DK with metadataEnrichment enabled
2. Deploy sample that will get injected
3. Look for the new envvars on the Pod
4. Check Pod's fs to see if the properties files are correct
   - `/var/lib/dynatrace/enrichment/dt_metadata.properties`
   - `/var/lib/dynatrace/enrichment/dt_metadata.json`


pkg/consts/enrichment_injection.go
```diff
@@ -5,6 +5,7 @@ const (
 	EnrichmentInjectedEnv        = "METADATA_ENRICHMENT_INJECTED"
 	EnrichmentWorkloadKindEnv    = "DT_WORKLOAD_KIND"
 	EnrichmentWorkloadNameEnv    = "DT_WORKLOAD_NAME"
+	EnrichmentClusterNameEnv     = "DT_CLUSTER_NAME"
```
waodim, line 8: By testing it I saw that the value is essentially the same as the KubernetesClusterMEID in the dynakube status that I introduced with the monitoredEntitiyReconciler. Should we align the names as both field and env var have the same value or should we stick with as it is?

waodim (commented):
Apart from my comment I would approve:

- files are present
- env var are set correctly

pkg/consts/enrichment_injection.go
0sewa0: thought about it and I am still conflicted about it, because I don't like the inconsistency
but I like `KubernetesClusterMEID` name more, as its more exact, but in the context of the enrichment that does not work

waodim: I mean if it does not work I think we should stick with your approach, was just a thought that came across 👍

waodim (approved):

pkg/consts/enrichment_injection.go
0sewa0: "does no work" == the property is just name differently, and we have no control over it, so I can't rename it 😅 

And the name of `...MEID` works better in other parts of our code

pkg/webhook/mutation/pod/metadata/mutator.go
```diff
@@ -76,7 +76,7 @@ func (mut *Mutator) Mutate(ctx context.Context, request *dtwebhook.MutationReque
 
 	setupVolumes(request.Pod)
 	mutateUserContainers(request.BaseRequest)
-	updateInstallContainer(request.InstallContainer, workload)
+	updateInstallContainer(request.InstallContainer, workload, request.DynaKube.Status.KubernetesClusterMEID)
```
gkrenn, line 79: It can happen that this field is empty. Setting an empty meid will lead to an init container crashloopbackoff, blocking the run of injected pods.

gkrenn (changes requested):

pkg/webhook/mutation/pod/metadata/mutator.go
waodim: Which scenario can lead to this to happen? We introduced the monitored entity reconciler exactly for the `KubernetesClusterMEID` field to be set.

0sewa0: should be fixed now

gkrenn:
`k8s.container.name´ is still missing, otherwise looks fine to me

gkrenn (changes requested):

pkg/injection/startup/enrichment.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	props := map[string]string{}
+
+	err = json.Unmarshal(raw, &props)
+	if err != nil {
+		return err
+	}
+
+	propsContent := ""
```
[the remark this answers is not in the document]
0sewa0: done

0sewa0:
Adding `k8s.container.name` is more complicated, so I will create a separate PR for that

aorcholski (approved):
LGTM

gkrenn (approved):

PR #3521, opened 2024-07-23
Automates deployment of EEC StatefulSet

aorcholski:
## Description

[K8S-10080](https://dt-rnd.atlassian.net/browse/K8S-10080)

The Operator creates Extension Controller statefulset if extensions are enabled.

To be done as separate tickets :
- image version
- pvc
- configuration file

## How can this be tested?

Apply dynakube (provide `customPullSecret` and `imageRef`):

```
spec:
  apiUrl: ...
  customPullSecret: ...
  extensions:
    prometheus:
      enabled: true
  templates:
    extensionExecutionController:
      imageRef:
        repository: ...
        tag: ...  
```


pkg/controllers/dynakube/extension/statefulset.go
```diff
@@ -0,0 +1,387 @@
+package extension
```
andriisoldatenko, line 1: I have similar task to create OTEL collector StatefulSet, so IMO it's better if we create subpackages or somehow split this file, otherwise I need to refactor you part as well.

pkg/controllers/dynakube/extension/statefulset.go
```diff
@@ -0,0 +1,387 @@
...
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
```
andriisoldatenko, line 21: we have const file for that needs. It's just new file but same package...

pkg/controllers/dynakube/extension/statefulset.go
```diff
@@ -0,0 +1,387 @@
...
+	eecFile                 = "eec.token"
+	logVolumeName           = "log"
+	logMountPath            = "/var/lib/dynatrace/remotepluginmodule/log"
+	runtimeVolumeName       = "agent-runtime"
+	runtimeMountPath        = "/var/lib/dynatrace/remotepluginmodule/agent/runtime"
+	configurationVolumeName = "runtime-configuration"
+	configurationMountPath  = "/var/lib/dynatrace/remotepluginmodule/agent/conf/runtime"
+)
+
+func (r *reconciler) reconcileStatefulset(ctx context.Context) error {
```
andriisoldatenko, line 50:
```suggestion
func (r *reconciler) reconcileEECStatefulSet(ctx context.Context) error {
```

pkg/controllers/dynakube/extension/statefulset.go
```diff
@@ -0,0 +1,387 @@
...
+
+		return err
+	}
+
+	conditions.SetStatefulSetCreated(r.dk.Conditions(), extensionsControllerStatefulSetConditionType, desiredSts.Name)
+
+	return nil
+}
+
+func setReplicas() func(o *appsv1.StatefulSet) {
```
andriisoldatenko, line 125: must be part of the framework, so we can reuse it. So please move it to builder.go, check how I did service it will help you to inspire.

pkg/controllers/dynakube/extension/consts/consts.go
```diff
@@ -0,0 +1,15 @@
+package consts
```
andriisoldatenko, line 1: need to update in some other places.

pkg/util/kubeobjects/statefulset/builder.go
```diff
@@ -59,11 +60,11 @@ func SetTopologySpreadConstraints(topologySpreadConstraints []corev1.TopologySpr
 	}
 }
 
-func SetAllLabels(appLabels, labels map[string]string) builder.Option[*appsv1.StatefulSet] {
+func SetAllLabels(appLabels *labels.AppLabels, customLabels map[string]string) builder.Option[*appsv1.StatefulSet] {
```
andriisoldatenko, line 63: I don't like idea to have dependency here of labels package, because if I want just map[string]string this code won't work.

andriisoldatenko (dismissed):

andriisoldatenko (changes requested):
Failed to apply default image tag ":": couldn't parse image name ":": invalid │
│  reference format
if we don't provide image:
```spec:
  apiUrl: ...
  customPullSecret: ...
  extensions:
    prometheus:
      enabled: true
```

wepudt (dismissed):
lgtm

code lgtm
also deployed in my cluster

wepudt:
> Failed to apply default image tag ":": couldn't parse image name ":": invalid │ │ reference format if we don't provide image:
> 
> ```
>   apiUrl: ...
>   customPullSecret: ...
>   extensions:
>     prometheus:
>       enabled: true
> ```

did reproduce.

andriisoldatenko (approved):

albertogdd:
I guess these logs are normal until further tickets are done? @aorcholski 
image

pkg/controllers/dynakube/extension/consts/consts.go
```diff
@@ -0,0 +1,19 @@
...
```
albertogdd, line 1: Just want to point out that in other packages inside `pkg/controllers/dynakube` we don't create a subpackage for the consts (e.g. `pkg/controllers/dynakube/proxy/consts.go`).
We should keep the same package structure in the directory IMO

So either rolling back this change, or updating the existing consts.go files to its own subdir.

aorcholski: `consts` sub-directory / package is needed because otherwise can't import them from `extension/eec` , `extension/otel` packages because of circular dependency
```
extension <- extension/eec <- extension
```

aorcholski: billions of messages to be solved by newer EEC image or a configuration file

PR #3528, opened 2024-07-24
Introduce statefulset inside kubeobject

andriisoldatenko:
## Description

This PR introduces a statefulSet builder.

> Context: we are going soon to introduce 2 similar stateful sets for OTEL and EEC, so it makes sense to have common logic inside the builder.

##  how to test
`make go/test`


0sewa0 (approved):
Hard to review without an actual usage of it, but lets see

andriisoldatenko:
> Hard to review without an actual usage of it, but lets see

yeah but we need some groudwork :)

aorcholski (approved):

PR #3534, opened 2024-07-24
Set owner during update

0sewa0:
## Description
Set owner during update otherwise after an update the owner reference disappears

## How can this be tested?

Wait, or make a small change (add an env) to make an update happen, and see that the owner reference is still present and if you delete the dynakube the related resource is deleted

### DIFF
--- a/pkg/util/kubeobjects/internal/query/query.go
+++ b/pkg/util/kubeobjects/internal/query/query.go
@@ -56,6 +56,13 @@ func (c Generic[T, L]) Create(ctx context.Context, object T) error {
 func (c Generic[T, L]) Update(ctx context.Context, object T) error {
 	c.log(object).Info("updating")
 
+	if c.Owner != nil {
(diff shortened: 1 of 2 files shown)


andriisoldatenko (dismissed):

pkg/util/kubeobjects/internal/query/query.go
```diff
@@ -56,6 +56,13 @@ func (c Generic[T, L]) Create(ctx context.Context, object T) error {
 func (c Generic[T, L]) Update(ctx context.Context, object T) error {
 	c.log(object).Info("updating")
 
+	if c.Owner != nil {
```
andriisoldatenko, line 59: just curios: can we write a test for it ?

0sewa0: done

andriisoldatenko (approved):

luhi-DT (approved):

PR #3536, opened 2024-07-25
Update token scope api call version

0sewa0:
## Description

We use the older v1 tenant API for token scope check, this APIs are now deprecated.
However this APIs didn't change in the v2 tenant API, so we only need to change the URL we call, and nothing more.

## How can this be tested?

no change in behaviour (token scope check)


waodim (approved):

albertogdd (approved):
Deployed branch in my cluster, LGTM

chrismuellner (changes requested):
doesn't work

pkg/clients/dynatrace/dynatrace_client_test.go
```diff
@@ -174,9 +174,9 @@ func handleRequest(request *http.Request, writer http.ResponseWriter) {
 		(&ipHandler{}).ServeHTTP(writer, request)
 	case "/v1/deployment/installer/agent/connectioninfo":
 		handleCommunicationHosts(request, writer)
-	case "/v1/events":
+	case "/v2/events":
```
chrismuellner, line 177: this endpoint does not exist in v2 as post:
- `/v1/events` to list/post events
- `/v2/events` to list all events
- `/v2/events/ingest` to ingest a custom event

chrismuellner (changes requested):

pkg/clients/dynatrace/dynatrace_client_test.go
0sewa0: thankfully we only POST, so we don't have to care about `/v2/events` 😌 

(quickly added it, still need to test that it actually works)

0sewa0: removed it as it requires a different API scope

chrismuellner (approved):

aorcholski (approved):
```
{"level":"info","ts":"2024-07-25T14:49:22.240Z","logger":"dynatrace-client","msg":"token verified"}
```

PR #3541, opened 2024-07-25
Improve reconcile flow for apimonitoring

0sewa0:
## Description

The reconcile flow in the apimonitoring reconciler is/was flawed.
- It required multiple runs to set up everything

Problem:
- `CreateOrUpdateKubernetesSetting` will:
  - Create a `Monitored Entity` if no scope was given (scope == monitored entity ID), this new `Monitored Entity` will be used as the scope for the created `Settings` object
  - Create/Update a `Settings` object with scope

So to not have to run the whole api-monitoring reconciler to set the monitored entity ID in the dynakube status, we should run the monitored-enitity-reconciler before and after `CreateOrUpdateKubernetesSetting`
- "before" is needed so we use the correct scope for creation/update if it exists
- "after" is needed so we actually set/know the monitored entity ID in the dynakube, so other features can use it.

Also moved the Kubernetes-app related logic to be the last thing, as that also relies on the fact that the `Monitored Entity` AND`Settings` object exists 

## How can this be tested?
When your tenant doesn't have a `Monitored Entity` for a given cluster, apply a dynakube with an `kubernetes-monitoring` ActiveGate, and see that `dk.Status.KubernetesClusterMEID` is set without having to wait for a while.


waodim:
Switching to ready for review, as I think it should work now as intended. Please test it as well properly maybe I missed something.

pkg/controllers/dynakube/apimonitoring/reconciler.go
```diff
@@ -45,17 +47,33 @@ func (r *Reconciler) createObjectIdIfNotExists(ctx context.Context) (string, err
...
+
+	if r.dk.Status.KubernetesClusterMEID != "" {
+		monitoredEntity = []dtclient.MonitoredEntity{
+			{
+				EntityId: r.dk.Status.KubernetesClusterMEID,
+			},
+		}
+	}
+
+	objectID, err := r.dtc.CreateOrUpdateKubernetesSetting(ctx, r.clusterLabel, r.dk.Status.KubeSystemUUID, r.dk.Status.KubernetesClusterMEID)
```
luhi-DT, line 65: can you move this to the else branch of the above if? because if you have MEID you already have KubernetesSetting Object correct?

pkg/controllers/dynakube/apimonitoring/reconciler.go
```diff
@@ -45,17 +47,33 @@ func (r *Reconciler) createObjectIdIfNotExists(ctx context.Context) (string, err
...
+	if err != nil {
+		return "", errors.WithMessage(err, "error creating dynatrace settings object")
+	}
+
+	if r.dk.Status.KubernetesClusterMEID == "" {
+		// the CreateOrUpdateKubernetesSetting call will create the ME(monitored-entity) if no scope was given (scope == entity-id), this happens on the "first run"
+		// so we have to run the entity reconciler AGAIN to set it in the status.
+		err := r.monitoredEntitiesReconciler(r.dtc, r.dk).Reconcile(ctx)
```
luhi-DT, line 73: Same with this, i think this could be moved to the else branch of the first if

pkg/controllers/dynakube/apimonitoring/reconciler.go
waodim: If i get you correctly: The function would only be called if MEID is not set -> create part. If it is set we should not call the function: What about the update part then? Am i missing something here?

luhi-DT (changes requested):
We have to decouple the MEID reconciler from the activegate reconciler, since a K8sMEID can be created and exist without an activeGate. Also we have to update each reconcile the MEID since it can change (e.g if the k8s cluster is deleted in the UI and a new one will be created)

pkg/controllers/dynakube/monitoredentities/reconciler.go
```diff
@@ -18,38 +20,48 @@ func NewReconciler( //nolint
...
+		log.Info("failed to retrieve MEs")
 
-		r.dk.Status.KubernetesClusterMEID = findLatestEntity(monitoredEntities).EntityId
+		return err
+	}
 
-		log.Info("kubernetesClusterMEID set in dynakube status, done reconciling", "kubernetesClusterMEID", r.dk.Status.KubernetesClusterMEID)
+	if len(monitoredEntities) == 0 {
+		log.Info("no MEs found, no kubernetesClusterMEID will be set in the dynakube status")
+		setEmptyMEIDCondition(r.dk.Conditions())
```
gkrenn, line 55: I would not create an empty MEID condition, just create no condition if there are no monitored entities

pkg/controllers/dynakube/monitoredentities/reconciler.go
```diff
@@ -18,38 +20,48 @@ func NewReconciler( //nolint
...
 
 func (r *Reconciler) Reconcile(ctx context.Context) error {
-	if r.dk.Status.KubernetesClusterMEID == "" {
-		log.Info("start reconciling monitored entities")
+	log.Info("start reconciling monitored entities")
 
-		monitoredEntities, err := r.dtClient.GetMonitoredEntitiesForKubeSystemUUID(ctx, r.dk.Status.KubeSystemUUID)
-		if err != nil {
-			log.Info("failed to retrieve MEs")
+	if !conditions.IsOutdated(r.timeProvider, r.dk, MEIDConditionType) {
```
gkrenn, line 38: You either have to let a nil meid pass or just not store an empty meid condition. Otherwise your apimonitoring rereconciler implementation won't work, as in the second call the status would still not be set

pkg/controllers/dynakube/monitoredentities/reconciler.go
```diff
@@ -18,38 +20,48 @@ func NewReconciler( //nolint
...
-		log.Info("kubernetesClusterMEID set in dynakube status, done reconciling", "kubernetesClusterMEID", r.dk.Status.KubernetesClusterMEID)
+	if len(monitoredEntities) == 0 {
+		log.Info("no MEs found, no kubernetesClusterMEID will be set in the dynakube status")
+		setEmptyMEIDCondition(r.dk.Conditions())
+
+		return nil
 	}
 
+	r.dk.Status.KubernetesClusterMEID = findLatestEntity(monitoredEntities).EntityId
+	conditions.SetStatusUpdated(r.dk.Conditions(), MEIDConditionType, "Kubernestes Cluster MEID is up to date in the status: "+r.dk.Status.KubernetesClusterMEID)
```
gkrenn, line 61: Would create a shorter message here, like "Kubernestes Cluster MEID is up to date" only

gkrenn (changes requested):

pkg/controllers/dynakube/monitoredentities/reconciler.go
waodim: If I understood you correctly, removing the empty me condition will fix this?

pkg/controllers/dynakube/monitoredentities/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3541/commits/f02da5dac342c1f5adb3818ea08fc186560da86e

pkg/controllers/dynakube/monitoredentities/reconciler.go
waodim: possibly (?) done in https://github.com/Dynatrace/dynatrace-operator/pull/3541/commits/f02da5dac342c1f5adb3818ea08fc186560da86e

pkg/controllers/dynakube/monitoredentities/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3541/commits/f02da5dac342c1f5adb3818ea08fc186560da86e

pkg/controllers/dynakube/monitoredentities/reconciler.go
gkrenn: Turns out you set the condition status to `false` so would have worked as expected, so my bad 😅 
But still like it better as it is now, otherwise the status condition would have never become true, if there is no monitored entity

gkrenn (approved):
Code looks fine 

Checked with apimonitoring active gate => worked

luhi-DT (approved):

PR #3554, opened 2024-08-01
Revert GORM changes

0sewa0:
## Description

[K8S-10709](https://dt-rnd.atlassian.net/browse/K8S-10709)

Reverts/removes all the changes that introduced GORM and the new database schema, as it causes more trouble then its worth.

The state should be similar to what we had in 1.1, with a few additions:
- Digest is no longer used for the image-installer (so when CodeModulesImage is set)
  - Using the "old" database schema, we store the `ImageUri` (base64 encoded, see comment in code) in the `ImageDigest` column
    - Smallest and least intrusive change possible, as it's a bug fix
- Added extra safety checks in the GC, when handling the codemodule binary folders.
  - This is necessary so if someone moves from the "new"(GORM) database schema to the "old" one, will not have problems.
    - Without the checks, we might try to remove the binary folders that are still mounted as we don't have any information about them.

## How can this be tested?
Try scenarios where you upgrade from a released 1.2 version, while codemodules are mounted, and oneagents are rolled out.

Example: (what I tried)
1. Deploy Operator 1.2.1
2. Deploy cloudnative DynaKube (with a given codeModulesImage set)
3. Deploy Sample apps
4. Update cloudnative DynaKube to not have codeModulesImage
5. Upgrade the Operator to branch version.
6. Check that the GC doesn't delete codeModulesImage binary as the Sample apps use it
7. Delete Sample apps (they don't get stuck)
8. Delete DynaKube (oneagents don't get stuck)


.golangci.yml
```diff
@@ -137,7 +137,6 @@ linters-settings:
           - "github.com/docker/cli"
           - "github.com/go-gormigrate/gormigrate"
           - "github.com/google/uuid"
-          - "github.com/google/go-cmp/cmp"
```
andriisoldatenko, line 140: we need it.

andriisoldatenko: testutil.PartialEqual

albertogdd:
Is there a ticket for this? It's missing in the description

gkrenn (commented):
Pretty sure that monitoring works for all pods, not sure if the GC works as expected, would need to dig deeper there. - Was deploying a cloudnativeFS deployment, but the run directory was not cleaned up completely

pkg/controllers/csi/gc/reconciler.go
```diff
@@ -38,6 +40,8 @@ func NewCSIGarbageCollector(apiReader client.Reader, opts dtcsi.CSIOptions, db m
 		fs:                    afero.NewOsFs(),
 		db:                    db,
 		path:                  metadata.PathResolver{RootDir: opts.RootDir},
+		mounter:               mount.New(""),
+		isNotMounted:          mount.IsNotMountPoint,
```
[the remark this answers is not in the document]
luhi-DT: I will create a follow up, however i dont want to make this change in the bugfix release

pkg/controllers/csi/metadata/sqlite.go
```diff
@@ -723,6 +730,47 @@ func (access *SqliteAccess) GetTenantsToDynakubes(ctx context.Context) (map[stri
...
+	defer func() { _ = rows.Close() }()
+
+	var volumes = make([]*Volume, 0)
+
+	for rows.Next() {
+		var code_module_version string
+
+		var volume_meta_id string
+
+		var location string
```
andriisoldatenko, line 750: `var code_module_version, volume_meta_id, location string`
you can combine these 3 lines into 1

luhi-DT: good one

luhi-DT: fixed in newest commit

andriisoldatenko (approved):

aorcholski (commented):
Test scenario works. 
/data/codemodules/cHVibGljLmVjci5hd3MvZHluYXRyYWNlL2R5bmF0cmFjZS1jb2RlbW9kdWxlczoxLjI5My4xNDEuMjAyNDA2MjEtMTIyMzI4 directory was removed after sample app and dynakube were deleted

wepudt (commented):
also tested the scenario and inspected the sqlite db.

CM image was on disk until change, then got deleted after restarting sample app pod.

not looked into all details of the 44changed files!

PR #3555, opened 2024-08-01
Add missing install deadcode

andriisoldatenko:
This PR fixes: deadcode command not found and as side effect fixing all linter issues.


pkg/util/conditions/service.go
```diff
@@ -21,16 +21,6 @@ func SetServiceCreated(conditions *[]metav1.Condition, conditionType, name strin
 	_ = meta.SetStatusCondition(conditions, condition)
 }
 
-func SetServiceUpdated(conditions *[]metav1.Condition, conditionType, name string) {
```
gkrenn, line 24: `ServiceUpdatedReason` const is unused now

pkg/util/conditions/statefulset.go
```diff
@@ -21,26 +21,6 @@ func SetStatefulSetCreated(conditions *[]metav1.Condition, conditionType, name s
 	_ = meta.SetStatusCondition(conditions, condition)
 }
 
-func SetStatefulSetUpdated(conditions *[]metav1.Condition, conditionType, name string) {
```
gkrenn, line 24: Updated/Deleted consts are unused now

pkg/util/conditions/suffix.go
```diff
@@ -17,10 +17,6 @@ func appendUpdatedSuffix(name string) string {
 	return fmt.Sprintf("%s %s", name, updatedSuffix)
 }
 
-func appendDeletedSuffix(name string) string {
```
gkrenn, line 20: delete suffix unused

andriisoldatenko: it's unclear what do you mean :)

aorcholski (dismissed):

pkg/util/conditions/suffix.go
andriisoldatenko: fixed in cd8132dc

pkg/util/conditions/statefulset.go
andriisoldatenko: fixed in cd8132dc

pkg/util/conditions/service.go
andriisoldatenko: fixed in cd8132dc

hack/make/go.mk
```diff
@@ -52,6 +52,6 @@ go/gen_mocks: prerequisites/mockery
 	mockery
 
 ## Runs deadcode https://go.dev/blog/deadcode
-go/deadcode:
```
gkrenn, line 55: Any reason why this is not called within `go/lint`?

andriisoldatenko: we are not using go/lint in CI, we use github action ... it was like this before I think.

gkrenn: That's ok, would still be nice if you call `go/deadcode` within the `go/lint` command so it automatically runs as soon as we lint

andriisoldatenko: do you mean this b0d4b8bf?

gkrenn (changes requested):
Would still like to see the deadcode command to run with lint command, otherwise looks good to me. 

Maybe I wouldn't remove the deadcode from the utils, as this will be helpful in the future, but that's more personal opinion.

hack/make/go.mk
gkrenn: Ups, didn't see that 👍

gkrenn: Well, that commit was on the wrong branch 😉

andriisoldatenko: fixed. :)

gkrenn (approved):

aorcholski (approved):

PR #3561, opened 2024-08-02
Add k8s.container.name to the metadata enrichment files

0sewa0:
## Description
A follow-up of https://github.com/Dynatrace/dynatrace-operator/pull/3518, so related to the same ticket ([K8S-9858](https://dt-rnd.atlassian.net/browse/K8S-9858))

The complicated part of the change is:
- [Extract container info handling to general pod-webhook from oa-mutator](https://github.com/Dynatrace/dynatrace-operator/commit/884d5dec73936451533caa052dc24d2d96427139)
   - Needed so the metadata-enrichment mutator can also use it. (in conjunction with the init-container)
   - Also includes a slight rework, because the "container counting and several envvars" way of passing the container info to the init-container was waaay too overcomplicated. 
     - So no more `CONTAINER_COUNT` and `CONTAINER_1_NAME` envs littering the init-container's env just a single `CONTAINER_INFO` with some nice json in it.
 
After that the [Add k8s.container.name to the metadata enrichment files](https://github.com/Dynatrace/dynatrace-operator/commit/e7be8654d816331592fc623d40bd9eb191317132) part is trivial.
- Now you should always see the `k8s.container.name` fields in the metadata enrichment files in the injected user containers.

## How can this be tested?
When injecting user pods:
- You should no longer see `CONTAINER_COUNT` , `CONTAINER_1_NAME` and `CONTAINER_1_IMAGE` envs only `CONTAINER_INFO` on the init-container.
   - This new env should always be present, no matter what injection is enabled
   - If no injection is enabled for a given container then that container's info shouldn't be part of the `CONTAINER_INFO` env. (it wouldn't hurt anyone, but still)
- The OA injection shouldn't experience any change in behaviour. (different envs on init, but still works as expected)
- The metadata injection now should produce files that have the `k8s.container.name` key and value in them. (see #3518 for where the files can be found)


pkg/webhook/mutation/pod/metadata/containers.go
```diff
@@ -2,41 +2,28 @@ package metadata
...
-		container := &request.Pod.Spec.Containers[i]
-
-		if dtwebhookutil.IsContainerExcludedFromInjection(request, container.Name) {
-			log.Info("Container excluded from metadata-enrichment injection", "container", container.Name)
-
-			continue
-		}
-
+	newContainers := request.NewContainers(ContainerIsInjected)
+	for i := range newContainers {
```
wepudt, line 10: is there a reason you use this instead of
```
for _, container := newContainers {
...
```
?

wepudt (commented):
tested on my cluster with single container nginx, env vars present, metadata files correct

aorcholski (commented):
I did some tests on my cluster using java app with app container and istio container.
```
install-oneagent:
  env:
    CONTAINER_INFO:  [{"name":"app","image":"..."},{"name":"istio-proxy","image":"..."}]
```
`app` container:
```
$ cat /var/lib/dynatrace/enrichment/dt_metadata.properties
k8s.container.name=app
k8s.node.name=gke...
...

$  cat /var/lib/dynatrace/oneagent/agent/config/container.conf
[container]
containerName app
k8s_containername app
k8s_node_name gke...
...
```
`istio-proxy` container:
```
$ cat /var/lib/dynatrace/enrichment/dt_metadata.properties
k8s.container.name=istio-proxy
k8s.node.name=gke...
...

$ cat /var/lib/dynatrace/oneagent/agent/config/container.conf
[container]
containerName istio-proxy
k8s_containername istio-proxy
k8s_node_name gke...
...
```

POD `inject` annotation works:
```
annotations:
  container.inject.dynatrace.com/istio-proxy: false

install-oneagent:
  env:  
    CONTAINER_INFO:    [{"name":"app","image":"..."}]
```

Looks good.

waodim (approved):
tested with a php app, all files created correctly, values set correctly, code also lgtm

aorcholski (approved):
LGTM

PR #3568, opened 2024-08-05
Fix standalone metadata enrichment if no oneAgent section is configured

luhi-DT:
## Description
Ticket: https://dt-rnd.atlassian.net/browse/K8S-10845

Currently the operator does not add metadata enrichment if no oneAgent section is configured, as the reconciler unmaps all the namespaces

## How can this be tested?

Deploy a Dynakube with only metadataEnrichment and no oneAgent or activeGate section -> check if sample apps get the initContainer and have the /var/lib/dynatrace/enrichment/* files (metadata.json/metadata.properties)


wepudt (approved):
lgtm

tested on my cluster with nginx, all files present

do we need #3561 for that too?

PR #3595, opened 2024-08-08
Add logAgent configuration options for Operator

waodim:
## Description

Jira: https://dt-rnd.atlassian.net/browse/K8S-10533

With this PR the LogAgent is introduced in the CR. It can be configured and applied. It is only added to v1beta3 and no conversion logic is needed. Further logic will be added in follow ups.

## How can this be tested?

Apply dynakubes that use the newly introduced field, check if the fields are correctly set, see if everything works as before. Also read through my comments and correct if necessary.

### DIFF
--- a/config/crd/bases/dynatrace.com_dynakubes.yaml
+++ b/config/crd/bases/dynatrace.com_dynakubes.yaml
@@ -5013,6 +5013,14 @@ spec:
                     - enabled
                     type: object
                 type: object
+              logMonitoring:
+                description: General configuration about LogMonitoring instances.
+                properties:
+                  enabled:
+                    type: boolean
+                required:
+                - enabled
+                type: object
               metadataEnrichment:
                 description: Configuration for Metadata Enrichment.
                 properties:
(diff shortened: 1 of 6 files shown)


config/crd/bases/dynatrace.com_dynakubes.yaml
```diff
@@ -6616,6 +6624,147 @@ spec:
                           type: object
                         type: array
                     type: object
+                  logAgentDaemonSet:
```
andriisoldatenko, line 6627: IN the ticket i see `LogAgentSpec` not logAgentDaemonSet. May be @chrismuellner  can say final word?

For example for EEC and OtelC we do use without Daemonset suffix, but it's Daemonset.

config/crd/bases/dynatrace.com_dynakubes.yaml
```diff
@@ -5013,6 +5013,14 @@ spec:
                     - enabled
                     type: object
                 type: object
+              logMonitoring:
```
andriisoldatenko, line 5016: you have to run `make doc/api-ref` to update docs.

waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3595/commits/d28e3b77f6b09a62c54494ddc435a1e7538c5ee5

pkg/api/v1beta3/dynakube/extensions.go
```diff
@@ -102,3 +104,45 @@ type ImageRefSpec struct {
 	// Indicates a tag of the image to use
 	Tag string `json:"tag,omitempty"`
 }
+
+type LogAgentDaemonSetSpec struct {
```
aorcholski, line 108:
```suggestion
type LogAgentSpec struct {
```

config/crd/bases/dynatrace.com_dynakubes.yaml
waodim: I agree and removed the suffix. https://github.com/Dynatrace/dynatrace-operator/pull/3595/commits/d28e3b77f6b09a62c54494ddc435a1e7538c5ee5

If it is needed, I can readd it.

pkg/api/v1beta3/dynakube/extensions.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3595/commits/a8bbce9077f2aa4e354807686d30ae72e8518cd4

pkg/api/v1beta3/dynakube/extensions.go
```diff
@@ -102,3 +104,45 @@ type ImageRefSpec struct {
 	// Indicates a tag of the image to use
 	Tag string `json:"tag,omitempty"`
 }
+
+type LogAgentSpec struct {
```
aorcholski, line 108: `Templates` section is a container for different features now.

I think `LogAgentSpec` struct should be part of log_monitoring_types.go . 
`TemplatesSpec` struct could be moved to dynakube_types.go .

waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3595/commits/3072f91d4046c25f070d99d2d5415ec4b331a85b

aorcholski (approved):
LGTM

pkg/api/v1beta3/dynakube/extensions.go
```diff
@@ -10,14 +10,6 @@ type ExtensionsSpec struct {
 	Prometheus PrometheusSpec `json:"prometheus,omitempty"`
 }
 
-type TemplatesSpec struct {
```
luhi-DT, line 13: Why is this removed?

waodim: see the comment from @aorcholski 

https://github.com/Dynatrace/dynatrace-operator/pull/3595#discussion_r1709124154

it was moved to dynakube_types.go

PR #3607, opened 2024-08-12
Undeploy operator gracefully

albertogdd:
## Description

<no-ticket>

WDYT about this idea, doing dynakube CRs cleanup before deleting all operator components on `make undeploy`?

This would fix pods getting hanged-up on Terminating state forever because its mounted volume got deleted.

## How can this be tested?

Apply a dynakube and run `make undeploy`. You should see how the make command waits for the 'operator-managed' pods to get completely deleted before uninstalling the operator - no hanged-up/orphan pods


hack/make/deploy/undeploy.mk
```diff
@@ -8,6 +8,8 @@ undeploy/%:
 
 ## Remove the operator with csi-driver, on kubernetes
 undeploy: manifests/crd/helm
+	kubectl delete dynakube --all -n dynatrace
+	kubectl get pods -n dynatrace -o=name | grep dynakube | while read -r POD ; kubectl wait --for=delete $$POD --timeout=60s; done
```
chrismuellner, line 12: we have instructions in our official documentation that should work: https://docs.dynatrace.com/docs/shortlink/guides-k8s-update-uninstall-operator#manifest-uninstall (under "With CSI driver")

if the command there has issues, we need to rethink those instructions as well

albertogdd: Thank you, with that command it works perfectly :+1: 
[Fixed here](https://github.com/Dynatrace/dynatrace-operator/pull/3607/commits/d7981fe8fe7c47247b6da587eadf52341036bb7e)

luhi-DT (approved):

chrismuellner (approved):

andriisoldatenko (approved):

PR #3613, opened 2024-08-13
Change slack notification target for e2e tests

luhi-DT:
## Description

In order to have a proper workflow for the e2e tests, we need to change the target for the notification if the tests are failing. This will allow us to run custom things, like creating jira issues etc.

## How can this be tested?

Run e2e test and let the fail


andriisoldatenko (approved):

albertogdd (approved):
I trust your repo owner secret handling powers

PR #3623, opened 2024-08-14
Rename LogAgent to LogModule

0sewa0:
## Description

Just to conform with the naming.
Naming changed to `LogModule` after initial ticket creation.

re-Generated:
- deepcopy
- api-ref
- crd

## How can this be tested?
Not much. Check the naming and doc fields.
Can check if I committed everything:
- `make manifests`
- `make doc/api-ref`
- `make manifests/deepcopy`


pkg/api/v1beta3/dynakube/log_module_types.go
```diff
@@ -2,48 +2,48 @@ package dynakube
...
 
-type LogMonitoringSpec struct {
+type LogModuleSpec struct {
 	Enabled bool `json:"enabled"`
 }
 
-type LogAgentSpec struct {
-	// Add custom LogAgent annotations
+type LogModuleDaemonSetSpec struct {
+	// Add custom annotations to the LogModule DaemonSet
```
chrismuellner, line 10: don't we only set it to pods? I'd just replace `LogModule Daemonset` with `Log Module` in this file

chrismuellner (changes requested):

pkg/api/v1beta3/dynakube/log_module_types.go
0sewa0: should we still have `LogModule pods` ?

example: `// Set tolerations for the LogModule pods`

I would replace everything to `... LogModule pods`, WDYT?

0sewa0: Updated, pls check 🙈

luhi-DT (approved):

PR #3626, opened 2024-08-14
Ignore enrichment rules without mapping set

0sewa0:
## Description

[K8S-10741](https://dt-rnd.atlassian.net/browse/K8S-10741)

Handles the case when the `Mapping` of an `EnrichmentRule` is empty, we just ignore it and log it.
We still store it in the dk.Status so we have the actual info we get from the API

## How can this be tested?

The unittest is pretty obvious.


pkg/api/v1beta3/dynakube/dynakube_status.go
```diff
@@ -162,6 +162,10 @@ type EnrichmentRule struct {
 }
 
 func (rule EnrichmentRule) ToAnnotationKey() string {
+	if rule.Mapping == "" {
```
luhi-DT, line 165: Why do you have to check here, if you already skipped the item in the for loop?

luhi-DT (changes requested):

pkg/api/v1beta3/dynakube/dynakube_status.go
0sewa0: Mainly for consistency, has no use for now. Also kinda make sense to not append "nothing" to a prefix 😅 

Was thinking of doing a check like `if rule.ToAnnotationKey() == "" { continue }`, but I found the  `if rule.Mapping == "" { continue }` more descriptive.

luhi-DT (approved):

wepudt (approved):
lgtm

PR #3627, opened 2024-08-14
Add labels for rhcc manifest support to CSVs for OLM marketplace releases

luhi-DT:
## Description

Related Ticket: https://dt-rnd.atlassian.net/browse/K8S-10607

After we have introduce an image for all architectures on rhcc (#1225), we need to indicate that our operator runs on all architectures. In order to let OLM know we need to put labels on all 3 OLM release (certified operators, rhmp, community-operators-prod)

## How can this be tested?

make bundle should generate CSVs with this annotations


config/helm/chart/default/templates/_helpers.tpl
```diff
@@ -29,6 +29,8 @@ Check if default image or imageref is used
 {{- else -}}
     {{- if (.Values.imageRef).repository -}}
         {{- .Values.imageRef.tag | default (printf "v%s" .Chart.AppVersion) | printf "%s:%s" .Values.imageRef.repository -}}
+    {{- else if eq (include "dynatrace-operator.platform" .) "openshift" -}}
```
0sewa0, line 32: how does it work with https://github.com/Dynatrace/dynatrace-operator/pull/3628?

0sewa0 (changes requested):

config/helm/chart/default/templates/_helpers.tpl
luhi-DT: platform still exists -> https://github.com/Dynatrace/dynatrace-operator/pull/3628/files#diff-0425575c0ee7e3b71e1b1457ed08cb886f1d5d7526a474dc9388ff1b9de67fc7L16

luhi-DT: and here https://github.com/Dynatrace/dynatrace-operator/pull/3628/files#diff-096ff34b076d51e9cedae8837f8d85cf95173899d1f3e269c2d497e72b69cd03L18

0sewa0 (approved):

andriisoldatenko (approved):

PR #3628, opened 2024-08-14
Add `install-source` label to Operator deployment

chrismuellner:
## Description
Any value for `platform` is now allowed, with the default assumption that manifests for `kubernetes` are generated.
- Manifest and Olm are set as install-source if helm is run with those values
- Any other platform is assumed as special marketplace and passed to the label without modification
- Helm is the default value, if nothing else matches

> [K8S-10736](https://dt-rnd.atlassian.net/browse/K8S-10736)

## How can this be tested?

- Run `make test/helm`
- Try out different values for `platform` and check the resulting manifests are for the right platform and the label is set correctly


luhi-DT (approved):

config/helm/chart/default/templates/Common/activegate/clusterrole-activegate.yaml
```diff
@@ -1,4 +1,3 @@
-{{- include "dynatrace-operator.platformRequired" . }}
```
0sewa0, line 1: ❤️

0sewa0 (approved):

PR #3644, opened 2024-08-20
Use csi liveness-probe sidecar

0sewa0:
## Description

[K8S-10646](https://dt-rnd.atlassian.net/browse/K8S-10646)

1. Use the `livenessprobe` sidecar for the `server` container's `livenessProbe` (which it's only purpose)
   - [`/healthz` must be used, otherwise it errors](https://github.com/kubernetes-csi/livenessprobe/blob/master/cmd/livenessprobe/main.go?rgh-link-date=2024-08-22T11%3A26%3A52Z#L171), ~maybe `--http-endpoint`could be used to allow `/livez`~
   - Calls https://github.com/Dynatrace/dynatrace-operator/blob/55e8026b06bf0b676443bc0bcec0d0fed5f729e9/pkg/controllers/csi/driver/server.go#L137-L139 to determine liveness, maybe we can improve it in the future?
   - ONLY use it for the `server` container, the `provisioner` shouldn't depend on the `server`'s liveness.  
2. Removes all the liveness-probe logic from the CSI `server`'s Manager:
   - They are not needed, the sidecar container is meant to do it
   - The Manager's only purpose now is to provide the `/metrics` server

Previously the `server` created it's own `/livez` endpoint and used it for its `livenessProbe`. Which works fine but not fully correct.

## How can this be tested?

When deploying the CSI-driver the `server` container shouldn't restart and use the sidecar for liveness.
(changing the ports/endpoint in the yaml will cause the container to restart, so I guess it's working 😅 )


andriisoldatenko (approved):
practically LGTM.

albertogdd:
Nitpick, switching from livez to healthz doesnt contradict the official doc guidelines?
https://kubernetes.io/docs/reference/using-api/health-checks/
> The healthz endpoint is deprecated (since Kubernetes v1.16), and you should use the more specific livez and readyz endpoints instead

0sewa0:
@albertogdd 
That page has the following:
> The Kubernetes [API server](https://kubernetes.io/docs/concepts/overview/components/#kube-apiserver) provides API endpoints to indicate the current status of the API server.  This page describes these API endpoints and explains how you can use them.

So the `/healthz` endpoint is deprecated on the API server, but not in the whole of kubernetes.

The `livenessProbe` of a pod could call any random endpoint that will return a response, no matter what the actual endpoint is called.
In our case we can only call the `/healthz` endpoint [because the sidecar container will only listen on that endpoint.](https://github.com/kubernetes-csi/livenessprobe/blob/master/cmd/livenessprobe/main.go#L171)

I agree that `/livez` would be nicer, but that is not up to me. 😅

0sewa0:
added `do-not-merge` label to figure out how this affects the google autopilot whitelisting

aorcholski (dismissed):
LGTM. `Server` container is restarted if `livenessprobe` port is changed (--health-port=9807) so it seems that `livenessprobe` container does it jobs.

andriisoldatenko (approved):

gkrenn (approved):

PR #3648, opened 2024-08-21
Rename fields of EnrichmentRules

0sewa0:
## Description

[K8S-10946](https://dt-rnd.atlassian.net/browse/K8S-10946)

In the `EnrichmentRule` struct `key` was renamed to `source` and `mapping` to `target`. (because the API changed)

## How can this be tested?

same as https://github.com/Dynatrace/dynatrace-operator/pull/3327, but the fields should be different(renamed) in the Status:
```
  metadataEnrichment:
    rules:
    - enabled: true
      source: cost-center
      target: dt.cost.costcenter
      type: LABEL
    - enabled: true
      source: security-context
      target: dt.security_context
      type: ANNOTATION
```

### DIFF
--- a/config/crd/bases/dynatrace.com_dynakubes.yaml
+++ b/config/crd/bases/dynatrace.com_dynakubes.yaml
@@ -4410,9 +4410,9 @@ spec:
                       properties:
                         enabled:
                           type: boolean
-                        key:
+                        source:
                           type: string
-                        mapping:
+                        target:
                           type: string
                         type:
                           type: string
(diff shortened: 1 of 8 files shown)


aorcholski (approved):
LGTM, I see my rule with new naming in the status
```
  Metadata Enrichment:
    Rules:
      Enabled:  true
      Source:   security_context
      Target:   dt.security_context
      Type:     ANNOTATION
```

andriisoldatenko (approved):

Issue #3535, opened 2024-07-24
expose failurePolicy in values .yaml


Aimen618:
**Is your feature request related to a problem? Please describe.**
the following parameter failurePolicy is hardcoded to ignore and it will be useful if we make it configurable:
webhooks:
  - name: webhook.pod.dynatrace.com
    reinvocationPolicy: IfNeeded
    failurePolicy: Ignore

**Describe the solution you'd like**
change it to:
{{.Values.webhook.mutatingWebhook.failurePolicy}}

**Describe alternatives you've considered**
manual patching after the chart deployment

**Additional context**
link to code:
https://github.com/Dynatrace/dynatrace-operator/blob/v1.1.0/config/helm/chart/default/templates/Common/webhook/mutatingwebhookconfiguration.yaml#L25

chrismuellner:
This has the implication that a failure in the webhook or problems with the webhook deployment, in general, would block the rollout of any pods that are expected to be monitored. We have added many features to reduce the impact of failure in any of our monitoring components on the general cluster. Still, we can make this setting configurable.

We'll add this in a future release of the Operator.

PR #3652, opened 2024-08-23
Make mutating webhook failurePolicy configurable via Helm

0sewa0:
## Description

[K8S-10756](https://dt-rnd.atlassian.net/browse/K8S-10756)

Makes the mutating webhook's failurePolicy configurable via Helm

fixes: https://github.com/Dynatrace/dynatrace-operator/issues/3535

## How can this be tested?

Unittest makes it pretty simple.

Otherwise, setting the new `webhook.mutatingWebhook.failurePolicy` to any(but valid) value. (in case of `helm template` the "valid" part doesn't matter 😉 )


andriisoldatenko (approved):

config/helm/chart/default/tests/Common/webhook/mutatingwebhookconfiguration_test.yaml
```diff
@@ -67,3 +67,16 @@ tests:
       - equal:
           path: webhooks[1].timeoutSeconds
           value: 13
+  - it: should change failurePolicy
```
chrismuellner, line 70: we could use a test for the default value

chrismuellner (changes requested):

config/helm/chart/default/tests/Common/webhook/mutatingwebhookconfiguration_test.yaml
0sewa0: done in [Add extra test for defaults](https://github.com/Dynatrace/dynatrace-operator/pull/3652/commits/34940c3875268991308afcdc73e133d4caf454e5)

andriisoldatenko (approved):

chrismuellner (approved):

PR #3663, opened 2024-08-26
Reduce influence of each component reconcile on each other

0sewa0:
## Description
[K8S-11003](https://dt-rnd.atlassian.net/browse/K8S-11003)

Basic idea:
- An error related to the reconcile of 1 section, shouldn't cause the skipping the reconcile of other sections
   -  Example: An error happens in the Extensions-reconcile, OneAgents and ActiveGates still get reconciled

3 fixes: (1 per commit)
- [Make error in extension reconciler not break reconcile loop](https://github.com/Dynatrace/dynatrace-operator/commit/e57d7e23aa9ec9bb88dfb42653b15a8388be368d)
  - Simple, straight forward, titles says it all
- [Elevate proxy reconciler to "prerequisite"](https://github.com/Dynatrace/dynatrace-operator/commit/4c3755de5a02893a39f0dce0b93f1205697145f3)
   - The "proxy secret" we create is used in multiple components, and if there is a problem with creating this, we will have a problem down the line, so it's safe to "elevate" it. (see more in commit message)
- [Move monitored-entities reconciler to section where it's actualy needed](https://github.com/Dynatrace/dynatrace-operator/commit/8cde6c70d1bddf18ff5d1d375a6a8a29a3691689)
  - We should reconcile information closes to where it's going to be used, more details in commit message 
  
## How can this be tested?

Normal scenario => no changes
Error in extensions => everything else is reconciled/deployed
Error in proxy reconcile => nothing is reconciled/deployed
Error in monitored-entity reconcile => only metadata-enrichment is not reconciled


wepudt (approved):
lgtm

gkrenn (approved):

PR #3665, opened 2024-08-26
Move CR name length validation to kubebuilder markers

0sewa0:
## Description
[K8S-10715](https://dt-rnd.atlassian.net/browse/K8S-10715)

**Original idea:** use `+kubebuilder:validation:MaxLength` to restrict the length of certain fields in our CRs.

## Problem:
- `+kubebuilder:validation:MaxLength` can’t be used for `metadata.name` as the marker comment needs to be on the field we want to validate, but the `metadata` struct is an external dependency.

## Solution:
Use `+kubebuilder:validation:XValidation:rule="size(self.metadata.name) < 40"` instead, we can put this marker on the `EdgeConnect` / `DynaKube` type.

### More problems
- `fieldPath=".metadata.name"` can’t be used, because you will get an error like: (this is a cosmetic problem, fieldPath is not mandatory)
   - `* CustomResourceDefinition.apiextensions.k8s.io "edgeconnects.dynatrace.com" is invalid: spec.versions[0].schema.openAPIV3Schema.x-kubernetes-validations[0].fieldPath: Invalid value: ".metadata.name": fieldPath must be a valid path`

   - reason: `metadata.name` is not fully in the schema, so we get this error, only the `metadata` field is present as an `object`
   - leaving `fieldPath` empty will cause:  `The EdgeConnect "test...." is invalid: <nil>: Invalid value: "object": The length limit for the name of...`
    - “Workaround“: set `fieldPath=".metadata"`
- We already have a “CRD too big “ problem, this change will store the validation logic/message in the CRD aswell, making our problems worse 🤷‍♂️
- This XValidation [feature is “stable“ only since 1.29](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules), has been in beta since 1.25

## How can this be tested?

Try deploying a CR with a name that is 40 or more characters


pkg/api/v1beta1/dynakube/dynakube_types.go
```diff
@@ -68,6 +68,7 @@ type DynaKubeValueSource struct { //nolint:revive
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 // +operator-sdk:csv:customresourcedefinitions:displayName="Dynatrace DynaKube"
 // +operator-sdk:csv:customresourcedefinitions:resources={{StatefulSet,v1,},{DaemonSet,v1,},{Pod,v1,}}
+// +kubebuilder:validation:XValidation:rule="size(self.metadata.name) < 40",reason="FieldValueInvalid",fieldPath=".metadata",message="The length limit for the name of a DynaKube is 40, because it is the base for the name of resources related to the DynaKube. (example: dkName-activegate-<some-hash>) The limit is necessary because kubernetes uses the name of some resources (example: StatefulSet) for the label value, which has a limit of 63 characters. (see https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)"
```
chrismuellner, line 71: why would you update this for older crd versions?

pkg/api/v1beta3/dynakube/dynakube_types.go
```diff
@@ -69,6 +69,7 @@ type DynaKubeValueSource struct { //nolint:revive
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 // +operator-sdk:csv:customresourcedefinitions:displayName="Dynatrace DynaKube"
 // +operator-sdk:csv:customresourcedefinitions:resources={{StatefulSet,v1,},{DaemonSet,v1,},{Pod,v1,}}
+// +kubebuilder:validation:XValidation:rule="size(self.metadata.name) < 40",reason="FieldValueInvalid",fieldPath=".metadata",message="The length limit for the name of a DynaKube is 40, because it is the base for the name of resources related to the DynaKube. (example: dkName-activegate-<some-hash>) The limit is necessary because kubernetes uses the name of some resources (example: StatefulSet) for the label value, which has a limit of 63 characters. (see https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)"
 type DynaKube struct {
```
chrismuellner, line 73: is it possible to put the marker on the `metav1.ObjectMeta` field?

pkg/api/v1alpha1/edgeconnect/edgeconnect_types.go
```diff
@@ -166,6 +166,7 @@ func (dk *EdgeConnectStatus) SetPhase(phase status.DeploymentPhase) bool {
 // +kubebuilder:printcolumn:name="ApiServer",type=string,JSONPath=`.spec.apiServer`
 // +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.phase`
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+// +kubebuilder:validation:XValidation:rule="size(self.metadata.name) < 40",reason="FieldValueInvalid",fieldPath=".metadata",message="The length limit for the name of a EdgeConnect is 40, because it is the base for the name of resources related to the EdgeConnect. The limit is necessary because kubernetes uses the name of some resources for the label value, which has a limit of 63 characters."
```
andriisoldatenko, line 169: it's not accurate, previously it was max was 40, so `40 > 40` is false, it means it will show error only on 41, but now we show error on 40. May be it's want we want, but it's new behaviour.

andriisoldatenko: so you should change it to `<=` or `< 41`

pkg/api/v1beta3/dynakube/dynakube_types.go
0sewa0: unfortunately no

pkg/api/v1beta1/dynakube/dynakube_types.go
0sewa0: wanted to be more consistent, but I can undo it for older CRs, and only have it for the latest ones.

should I do that? 🤔

andriisoldatenko (dismissed):

andriisoldatenko:
btw it works :) 
```
>>> "t" * 41
'ttttttttttttttttttttttttttttttttttttttttt'
```

```
The DynaKube "ttttttttttttttttttttttttttttttttttttttttt" is invalid: metadata: Invalid value: "object": The length limit for the name of a DynaKube is 40, because it is the base for the name of resources related to the DynaKube. (example: dkName-activegate-<some-hash>) The limit is necessary because kubernetes uses the name of some resources (example: StatefulSet) for the label value, which has a limit of 63 characters. (see https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)

```

aorcholski (dismissed):
The limit works

andriisoldatenko:
> The limit works but 39 characters are allowed

we even allow you 40 :) but 41 is not

pkg/api/v1beta1/dynakube/dynakube_types.go
chrismuellner: Wasn't the point of keeping older Dynakube versions to keep them unchanged? So I'd prefer to only change this in the latest version

0sewa0: on it 🫡

0sewa0: done

andriisoldatenko (approved):

aorcholski (approved):

PR #3666, opened 2024-08-27
Exclude e2e curl pod from injection

0sewa0:
## Description

The e2e curl helper pod shouldn't be injected. It's only meant to check network traffic, and we shouldn't interfere with it at all.

## How can this be tested?

Run e2e tests (ones that test the network, like the `istio` scenario). Check that the curl pod doesn't get injected.


andriisoldatenko (approved):

wepudt (approved):
lgtm

pods inspected, no OneAgent initcontainer found.

PR #3667, opened 2024-08-27
Propagate no proxy settings to all components

gkrenn:
## Description

This PR ensures that even the no proxy settings are propagated to all components. 
In addition it deprecates the following feature flags
- `oneagent-ignore-proxy`
- `activegate-ignore-proxy`

[Jira](https://dt-rnd.atlassian.net/browse/K8S-10423) 

## How can this be tested?

- deploy cloud native dynakube with proxy settings and no-proxy field set 
- check the following places if it was propagated: 
  - ActiveGate: `cat /var/lib/dynatrace/gateway/config_template/custom.properties`
  - osAgent: `cat /mnt/volume_storage_mount/var/agent/config/ruxitagentproc.conf`
  - codeModule: `cat /opt/dynatrace/oneagent-paas/agent/conf/ruxitagentproc.conf`
- to check the deprecation of the feature flags, just apply the dynakube => a warning should be issued


pkg/api/v1beta2/dynakube/validation/featureflag.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube"
+)
+
+const (
+	warningFeatureFlagDeprecated = `Feature flag %s is deprecated.`
+)
+
+var deprecatedFeatureFlags []string
+
+func init() {
```
0sewa0, line 17: why `init` and not just `var deprecatedFeatureFlags = []string{...}` ?

pkg/api/v1beta3/dynakube/validation/featureflag.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+var deprecatedFeatureFlags []string
+
+func init() {
+	deprecatedFeatureFlags = []string{
+		dynakube.AnnotationFeatureOneAgentIgnoreProxy,   //nolint:staticcheck
+		dynakube.AnnotationFeatureActiveGateIgnoreProxy, //nolint:staticcheck
+	}
+}
```
0sewa0, line 22: why `init` and not just `var deprecatedFeatureFlags = []string{...}` ?

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/customprops.go
```diff
@@ -30,7 +30,7 @@ type CustomPropertiesModifier struct {
 }
 
 func (mod CustomPropertiesModifier) Enabled() bool {
-	return mod.hasCustomProperties()
+	return mod.hasCustomProperties() || mod.dk.NeedsCustomNoProxy()
```
0sewa0, line 33: why not add the ` mod.dk.NeedsCustomNoProxy()` check inside the `hasCustomProperties` instead?

edit: you actually did that, then why do it again 😅

pkg/controllers/dynakube/activegate/internal/customproperties/reconciler.go
```diff
@@ -37,35 +46,107 @@ func NewReconciler(clt client.Client, dk *dynakube.DynaKube, customPropertiesOwn
...
+func (r *Reconciler) buildCustomPropertiesValue(ctx context.Context) ([]byte, error) {
+	value := ""
+
+	if r.customPropertiesSource != nil {
+		if r.customPropertiesSource.Value != "" {
+			value = r.customPropertiesSource.Value
+		} else if r.customPropertiesSource.ValueFrom != "" {
+			customPropertiesSecret := &corev1.Secret{}
+
+			err := r.client.Get(ctx, types.NamespacedName{Name: r.customPropertiesSource.ValueFrom, Namespace: r.dk.Namespace}, customPropertiesSecret)
```
0sewa0, line 107: why not use `query.Get`?

pkg/controllers/dynakube/activegate/internal/customproperties/conditions.go
```diff
@@ -0,0 +1,7 @@
+package customproperties
+
+const customPropertiesConditionType string = "CustomProperties"
+
+func customPropertiesConditionTypeString(customPropertiesOwnerName string) string {
+	return customPropertiesConditionType + "-" + customPropertiesOwnerName
```
0sewa0, line 6: this is not super related to your change, but what is this `customPropertiesOwnerName` and why do we care 😅 ?
I know its based on the service-account to be used by the activegate, but yet again, why do we care ?

1 dynakube will create 1 activegate (from the activegate section) so that postfix means nothing, just looks weird (`Type: CustomProperties-kubernetes-monitoring`), also it wouldn't work at all if we would deploy multiple activegates from 1 dynakube

I wouldn't change the behaviour of the old code related to this (it's not related to your intended change), BUT I would definitely not have it in new code. (ie.: the conditions), just have the `Type` be `CustomPropertiesSecret`

0sewa0 (changes requested):
works, tested it, no-proxy was present in the activegate/oa/codemodules

I left 1 comment that is mainly cosmetic and code "tidiness" related

pkg/controllers/dynakube/activegate/internal/customproperties/conditions.go
gkrenn: You are right, will remove it 👍

0sewa0 (approved):

pkg/api/v1beta3/dynakube/proxy.go
```diff
@@ -48,6 +48,10 @@ func (dk *DynaKube) HasProxy() bool {
 	return dk.Spec.Proxy != nil && (dk.Spec.Proxy.Value != "" || dk.Spec.Proxy.ValueFrom != "")
 }
 
+func (dk *DynaKube) NeedsCustomNoProxy() bool {
```
albertogdd, line 51: IMO I would rename this function
```suggestion
func (dk *DynaKube) NeedsNoProxy() bool {
```
Or even `NeedsProxy()` (switching some logic)

gkrenn: Would leave it with `CustomNoProxy` because of 2 reasons: 

- It is not the only reason why to set the `noProxy` field, we also add some values in case the ActiveGate is there and a proxy was provided, where `NeedsNoProxy` would be confusing. - `NeedsCustomNoProxy` is only true if the customer sets a value in the feature flag
- Removing the `no` would not make sense, the field which we set is called `noproxy` and commonly recognized as such

albertogdd (commented):
I found the dk.proxy value on an injected codemodule file, is this the expected behaviour?
```
/ # cat /opt/dynatrace/oneagent-paas/agent/conf/ruxitagentproc.conf | grep squid.proxy
proxy http://squid.proxy.svc.cluster.local:3128/
```

gkrenn:
@albertogdd Yes, that's correct, although this was already implemented. This PR only changed the noProxy field there, merging the custom no proxy feature flag value (`feature.dynatrace.com/no-proxy`) with the value from the ActiveGate we anyway already added.

PR #3668, opened 2024-08-27
Remove unnecessary roles from CSI serviceAccount

0sewa0:
## Description

[K8S-10892](https://dt-rnd.atlassian.net/browse/K8S-10892)

Remove unnecessary roles from CSI serviceAccount
They were probably copy-pasted from the Operator or some example, and nobody questioned them.

## How can this be tested?

deploy Operator and a DynaKube that uses the CSI-driver, nothing should break. (no actual change in behaviour)


config/helm/chart/default/templates/Common/csi/clusterrole-csi.yaml
```diff
@@ -37,30 +29,6 @@ rules:
       - create
```
chrismuellner, line 29: did you check if these need to be different from other places we use events?

0sewa0: you are referring to the weird `create/patch` in both clusterRole and none-clusterRoles?

it makes little sense: (I tested it now that you asked)
- Operator: Only creates/updates/patch (don't know the exact "verbs") in its own namespace, so the clusterRole is not necessary
    - purpose: needed for leader-election
- Webhook: Only creates/updates/patch (don't know the exact "verbs") in its own namespace so the clusterRole is not necessary
    - purpose: we create events when we inject or fail to inject a pod (only in case of oneagent injection)
    - seems kinda useless 😬 
- CSI-driver: Only creates/updates/patch (don't know the exact "verbs") in its own namespace
    - purpose: we create events when download a given version
    - seems kinda useless 😬

chrismuellner: thanks for testing that. I had an inkling they wouldn't be ncessary 😄 

we can just remove them here as well then!

chrismuellner: If we can remove verbs from the CSI driver roles/clusterroles let's do that and keep the rest as is for now.

0sewa0: related change [Move event permissions to role from clusterRole](https://github.com/Dynatrace/dynatrace-operator/pull/3668/commits/cfdbfad917dc660a7ddf47dc547475d5adeb7391)

gkrenn (approved):

chrismuellner (approved):

PR #3672, opened 2024-08-27
WIP: Create self-signed ActiveGate certificate when not specified

albertogdd:

### DIFF
--- a/pkg/controllers/dynakube/controller.go
+++ b/pkg/controllers/dynakube/controller.go
@@ -328,13 +328,11 @@ func (controller *Controller) setupActiveGateTlsSecret(ctx context.Context, dk *
 			certDomain := fmt.Sprintf(certificates.SelfSignedCertificateActiveGateDomain, dk.Name, dk.Namespace)
 
 			certAltNames := []string{}
-			for _, ip := range dk.Status.ActiveGate.ServiceIPs {
-				certAltNames = append(certAltNames, fmt.Sprintf("IP:%s", ip))
-			}
-			certAltNames = append(certAltNames, fmt.Sprintf("DNS:%s-activegate.%s", dk.Name, dk.Name))
-			certAltNames = append(certAltNames, fmt.Sprintf("DNS:%s-activegate.%s.svc", dk.Name, dk.Name))
 
-			cert, privateKey, err := certificates.CreateSelfSignedCertificate(certDomain, certAltNames)
+			certAltNames = append(certAltNames, fmt.Sprintf("%s-activegate.%s", dk.Name, dk.Name))
+			certAltNames = append(certAltNames, fmt.Sprintf("%s-activegate.%s.svc", dk.Name, dk.Name))
+
+			cert, privateKey, err := certificates.CreateSelfSignedCertificate(certDomain, certAltNames, os.Getenv("AGIP"))
 			if err != nil {
 				return err
 			}
@@ -347,7 +345,7 @@ func (controller *Controller) setupActiveGateTlsSecret(ctx context.Context, dk *
 
 			secret, err := secret.BuildForNamespace(secretName, dk.Namespace, certData, k8ssecret.SetLabels(coreLabels.BuildLabels()))
 
-			err = k8ssecret.Query(controller.client, controller.apiReader, log).Create(ctx, secret)
+			err = k8ssecret.Query(controller.client, controller.apiReader, log).WithOwner(dk).Create(ctx, secret)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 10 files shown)


pkg/util/certificates/certificates.go
```diff
@@ -25,3 +39,71 @@ func ValidateCertificateExpiration(certData []byte, renewalThreshold time.Durati
...
+	if err != nil {
+		return nil, err
+	}
+
+	var p12PemData []byte
+	for _, b := range p12Blocks {
+		p12PemData = append(p12PemData, pem.EncodeToMemory(b)...)
+	}
+
+	data["tls.crt"] = certPem
```
andriisoldatenko, line 102: // TLSKeyDataName is the key used to store a TLS private key in the secret's data field.
	TLSKeyDataName = "tls.key"

	TLSCrtDataName = "tls.crt"
)
we have constants, but may be we need to move them to another package so you can use them

pkg/util/certificates/certificates.go
```diff
@@ -1,13 +1,27 @@
...
+	"math/big"
 	"time"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
+	"golang.org/x/crypto/pkcs12"
 )
 
+const (
+	intSerialNumberLimit = 128
+	certPassword         = "changeit"
```
andriisoldatenko, line 20: I think it's possible to generate without password.

andriisoldatenko: but I need to check it :)

pkg/util/certificates/certificates.go
```diff
@@ -25,3 +39,71 @@ func ValidateCertificateExpiration(certData []byte, renewalThreshold time.Durati
...
+
+	privateKeyBytes, err := x509.MarshalECPrivateKey(privateKey)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return certBytes, privateKeyBytes, nil
+}
+
+func CreateP12CertificateSecretData(cert []byte, privateKey []byte) (map[string][]byte, error) {
```
andriisoldatenko, line 86: IMO it should return a struct with fields, and then we can transform it to data for secret. So this function must be responsible only for creating Certificate struct. Also IMO it's better to create this new type and create new so we can respect go way of doing it instead of function and map.

andriisoldatenko: good starting point is here https://github.com/dynatrace/dynatrace-operator/blob/7d5e26d9af7766f3d410098968d412aaaddb64d7/pkg/util/dttoken/token.go#L13-L37

pkg/util/certificates/certificates.go
```diff
@@ -25,3 +39,71 @@ func ValidateCertificateExpiration(certData []byte, renewalThreshold time.Durati
...
+	var data map[string][]byte
+
+	certPem := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert})
+	privateKeyPem := pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: privateKey})
+
+	p12Blocks, err := pkcs12.ToPEM(cert, certPassword)
```
aorcholski, line 92: AG needs binary format of p12 file. Not sure if "golang.org/x/crypto/pkcs12" can provide it

pkg/util/certificates/certificates.go
```diff
@@ -25,3 +41,71 @@ func ValidateCertificateExpiration(certData []byte, renewalThreshold time.Durati
...
+		return nil, err
+	}
+
+	var p12PemData []byte
+	for _, b := range p12Blocks {
+		p12PemData = append(p12PemData, pem.EncodeToMemory(b)...)
+	}
+
+	data["tls.crt"] = certPem
+	data["tls.key"] = privateKeyPem
```
aorcholski, line 105: please remove it (in a final commit) because AG needs `server.p12` and `password` fields only

aorcholski (changes requested):

pkg/util/certificates/certificates.go
albertogdd: Fixed

pkg/controllers/dynakube/controller.go
```diff
@@ -303,6 +314,44 @@ func (controller *Controller) setupTokensAndClient(ctx context.Context, dk *dyna
...
+			err = k8ssecret.Query(controller.client, controller.apiReader, log).Create(ctx, secret)
+			if err != nil {
+				return err
+			}
+
+		} else if err != nil {
+			return err
+		}
+
+		dk.Spec.ActiveGate.TlsSecretName = fmt.Sprintf(certificates.SelfSignedCertificateSecretName, dk.Name)
```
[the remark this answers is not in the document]
albertogdd: Fixed

pkg/util/certificates/certificates.go
albertogdd: Fixed with new library! :smile:

PR #3679, opened 2024-08-28
Propagate metadata annotations from pod to metadata props files

0sewa0:
## Description
[K8S-10838](https://dt-rnd.atlassian.net/browse/K8S-10838)

Basic idea:
Every annotation that:
- Has the `metadata.dynatrace.com/` prefix in it's key
- Is on the injected **pod**
   - annotations propagated by use via the EnrichmentRules also count (so select annotations from the namespace)
should be written into the `/var/lib/dynatrace/enrichment/dt_metadata.properties` and `/var/lib/dynatrace/enrichment/dt_metadata.json` files on each container of the injected pod.
- in the files, the key should no longer have the metadata prefix.

Workflow:
1. Propagate the annotations from namespace to pod (according to EnrichmentRules)
2. Propagate all annotations with metadata prefix from the pod to init-container envVar (as json)
3. Use the envVar's contents during the creation of the `/var/lib/dynatrace/enrichment/dt_metadata.*` files

## How can this be tested?
Same as https://github.com/Dynatrace/dynatrace-operator/pull/3386, but now the annotation's in question should also appear in the `/var/lib/dynatrace/enrichment/dt_metadata.*` files. (without the prefix)

An "easier" way to test it is to just put annotations with the metadata prefix on the to-be-injected Pod, and see those get propagated to the files inside the containers.


chrismuellner:
Did some testing on this together with @0sewa0 and we figured out that the enrichment rules do not work without an Activegate present. Currently discussing how  we want to proceed before this PR can be merged!

aorcholski (commented):
Works
```
[]$ sctl get -o yaml pod/java-eleven-musl-5fbc6cf5f7-m76kq | grep -A15 annotations:
  annotations:
    metadata.dynatrace.com/dt.security_context: blackhole          <--- from namespace, Enrichment rule
    metadata.dynatrace.com/k8s.workload.kind: deployment
    metadata.dynatrace.com/k8s.workload.name: java-eleven-musl
    metadata.dynatrace.com/security_ctx: reddwarf          <--- from deployment.pod.spec

[]$ sctl exec -it  pod/java-eleven-musl-5fbc6cf5f7-m76kq -- /bin/sh
Defaulted container "app" out of: app, artefact (init), install-oneagent (init)
~ $ cat /var/lib/dynatrace/enrichment/dt_metadata.properties 
dt.security_context=blackhole
k8s.workload.kind=Deployment
k8s.workload.name=java-eleven-musl
security_ctx=reddwarf

k8s.pod.name=java-eleven-musl-5fbc6cf5f7-m76kq
k8s.pod.uid=c34f4145-...
k8s.cluster.name=KUBERNETES_CLUSTER-...
k8s.cluster.uid=ee25d4f3-...
k8s.container.name=app
k8s.namespace.name=stargate
k8s.node.name=gke-...

dt.kubernetes.workload.kind=Deployment
dt.kubernetes.workload.name=java-eleven-musl
dt.entity.kubernetes_cluster=KUBERNETES_CLUSTER-...
dt.kubernetes.cluster.id=ee25d4f3-...
```

PR #3681, opened 2024-08-28
Add fallback for enrichment rules API call

0sewa0:
## Description

In case no monitored entity exists, we use the environment scope to get rules configured for the whole environment.

## How can this be tested?

Deploy a dynakube, without activegate, but metadataEnrichment true.
Have enrichment rules set up on your tenant (globally).
EnrichmentRules should be in the DynaKube's status


chrismuellner (dismissed):

pkg/clients/dynatrace/settings_enrichment.go
```diff
@@ -35,10 +36,10 @@ func (dtc *dynatraceClient) GetRulesSettings(ctx context.Context, kubeSystemUUID
 	}
 
 	if entityID == "" {
-		// if monitored entities were empty this field also stays empty, we return with no error
-		log.Info("No Monitored Entity ID, skip getting enrichment rules")
+		// if monitored entities were empty we then fallback to the enrichment-rules defined globally
+		log.Info("No Monitored Entity ID, getting environment enrichment rules")
 
-		return GetRulesSettingsResponse{}, nil
+		entityID = globalScope
```
chrismuellner, line 42:
```suggestion
		scope = globalScope
```

chrismuellner (approved):

andriisoldatenko (approved):

PR #3691, opened 2024-08-30
WIP Runs e2e tests for each commit on the latest release branch

aorcholski:
[K8S-9688](https://dt-rnd.atlassian.net/browse/K8S-9688)

## Description

Runs e2e tests for each commit on the latest release branch. Tests run on latest supported Kubernetes and OpenShift versions.

## How can this be tested?

Use forked repo. Change `Run in test` step of `run-in-k8s` and `run-in-ocp` 
```

### DIFF
--- a/.github/workflows/e2e-tests-release.yaml
+++ b/.github/workflows/e2e-tests-release.yaml
@@ -1,0 +1,111 @@
+name: E2E tests of latest release
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


luhi-DT (changes requested):
I would not run it on every commit, but once per day as the ticket states :)

PR #3692, opened 2024-08-30
Add proper handling of skipping the oa mutator

0sewa0:
## Description
The e2e curl helper pods were flaky when they were injected by the webhook, because sometimes they got injected 2 times with the init-container, which shouldn't happen. (the pod was rejected by kubernetes due to it)

(probably) How the problem happens:
1. No OneAgent connection info available during injection
2. OA mutator adds `oneagent.dynatrace.com/injected: false` + the reason annotation, to "skip" the OA injection
3. The webhook still adds the init-container (this SHOULDN'T happen, if no other mutator ran)
4. Another webhook (istio in our case) mutates the pod, causing a reinvocation of our webhook
5. Our webhook checks if it already injected, by checking the `oneagent.dynatrace.com/injected` and `metadata-enrichment.dynatrace.com/injected` annotations
   - Problem 1: It doesn't only check the existence of the annotations, but their value, so "not there" and "set to false' is the same, but it's really not, if the annotation is present then the mutator has ran before, no matter the value.
   - Problem 2: We could double check if the init-container is already in the pod spec.
6. According to the annotation, the pod wasn't mutated before, so the "mutation" logic will run instead of the "reinvocation" logic. This will add the init-container again.

Source of the problem:
- The `Mutator` interface in the webhook has been written 2 years ago for different requirements, and has not been updated since. 
- Back then, we didn't have this "skip but annotate" scenario that breaks things
- Will need to rework the interface (ie.: the whole webhook) to meet the current and possible future requirements, as this is just supposed to be a bug fix.

## How can this be tested?

pre-requisite: have a networkzone that has no fallback:
```
curl -X PUT https:/<my-tenant>/api/v2/networkZones/boom \
  -H "Authorization: Api-Token <my-token>" \
  -H "Content-Type: application/json" \
  -d "{ \"fallbackMode\": \"NONE\" }"
```
- this will create a `boom` networkzone that  has no fallback
- DO NOT USE ACTIVEGATE, it will create the networkzone, we do not want that
- make sure your token has the networkzone write permission for this command
- this bug is timing based, so we are creating a broken env, to eliminate the timing variable

(sidenote: you can do the 2 scenarios in 1 go, if you use the pod annotations to turn on/off certain injections instead of using different dynakubes)

Scenario 1.:
1. Configure a `networkZone` in the DynaKube that does not exist. 
2. Configure only the OneAgent the injections to be on
3. Double check the the `.Status.OneAgent.ConnectionInfo.CommunicationHost` are not in the status
4. Create a sample namespace with pods that will be injected by us AND another webhook (istio for example), so our webhook will be called 2 times
5. Check the sample pod is succesfully created, and only annotations are "injected", no init-container.

Scenario 2.:
1. Configure a `networkZone` in the DynaKube that does not exist. 
2. Configure **ALL** the injections to be on
3. Double check the the `.Status.OneAgent.ConnectionInfo.CommunicationHost` are not in the status
4. Create a sample namespace with pods that will be injected by us AND another webhook (istio for example), so our webhook will be called 2 times
5. Check the sample pod is succesfully created, and only metadata-enrichment is configured + oneagent annotations are "injected", init-container should exist.

How to verify that your setup is properly "broken":
- Deploy operator from `main`
- Run `Scenario 1.`
- The following should be in the replicaset events of your sample deployment:
```
Warning  FailedCreate  29s                replicaset-controller  Error creating: Pod "php-glibc-oa-5d4db695f7-dbkmz" is invalid: spec.initContainers[2].name: Duplicate value: "install-oneagent"
```


pkg/webhook/mutation/pod/metadata/containers.go
```diff
@@ -5,18 +5,18 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func mutateUserContainers(request *dtwebhook.BaseRequest) {
-	newContainers := request.NewContainers(ContainerIsInjected)
+func (mut *Mutator) mutateUserContainers(request *dtwebhook.BaseRequest) {
```
andriisoldatenko, line 8:
```suggestion
func (m *Mutator) mutateUserContainers(request *dtwebhook.BaseRequest) {
```

pkg/webhook/mutation/pod/metadata/mutator.go
```diff
@@ -133,7 +133,7 @@ func setWorkloadAnnotations(pod *corev1.Pod, workload *workloadInfo) {
 	pod.Annotations[dtwebhook.AnnotationWorkloadName] = workload.name
 }
 
-func ContainerIsInjected(container corev1.Container) bool {
+func (_ *Mutator) IsContainerInjected(container corev1.Container) bool {
```
andriisoldatenko, line 136: what is the purpose of creating static method ?

0sewa0:
Closing this in favour of https://github.com/Dynatrace/dynatrace-operator/pull/3700

This is just a too complicated change without solving the problem in the long term. We should go with the "simple but works" solution for now, and come up with a new version of the interface, that will fit our purposes.

PR #3698, opened 2024-09-02
Remove extensions logic out of code base everywhere BUT api package

waodim:
## Description

NOTE: The next PR in the series of PRs to remove v1beta3 from operator v1.3

In this PR I removed extensions logic in the codebase everywhere BUT the api package. This will be done in the next PR where I also change storage version of our CRD and so on.

## How can this be tested?

- Unit tests should pass
- Proof read and check if I missed something
- Functionality is the same


andriisoldatenko:
linting errors, @waodim  you have to undo also new code from kubeobjects :(

0sewa0 (approved):

andriisoldatenko (approved):

PR #3715, opened 2024-09-04
Change default value of metadata enrichment

luhi-DT:
https://dt-rnd.atlassian.net/browse/K8S-10945

## Description

Change default value of metadata enrichment to false

## How can this be tested?

Deploy a dynakube with metaDataEnrichment section but without the enabled flag -> metadataEnrichment should be not enabled and not working


0sewa0:
did you regenerate the CRD?

0sewa0 (approved):

PR #3740, opened 2024-09-06
Mount certificate for extension signature verification

andriisoldatenko:
## Description

This PR mount custom certificates for extension signature verification 

[K8S-9688](https://dt-rnd.atlassian.net/browse/K8S-9688)

## How can this be tested?

1. Create a secret based on any kind of certificate:

```
kubectl create secret tls custom-tls-cert --cert=cert.pem --key=key.pem
```

2. Apply dk and ref this secret as custom certificates `customExtensionCertificates`:

```
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
spec:
  customPullSecret: <pull secret>
  extensions:
    prometheus:
      enabled: true
  templates:
    extensionExecutionController:
      customExtensionCertificates: custom-tls-cert
      imageRef:
        repository: extk8sregistry.azurecr.io/eec/dynatrace-eec
        tag: 1.301.0.20240823-092629
```


config/crd/bases/dynatrace.com_dynakubes.yaml
```diff
@@ -6194,6 +6194,10 @@ spec:
                         description: Defines name of ConfigMap containing custom configuration
                           file
                         type: string
+                      customExtensionCertificates:
+                        description: Defines name of Secret containing certificate
```
StefanHauth, line 6198:
```suggestion
                        description: Defines name of Secret containing certificates for custom extensions signature validation
```

config/helm/chart/default/templates/Common/crd/dynatrace-operator-crd.yaml
```diff
@@ -6206,6 +6206,10 @@ spec:
                         description: Defines name of ConfigMap containing custom configuration
                           file
                         type: string
+                      customExtensionCertificates:
+                        description: Defines name of Secret containing certificate
```
StefanHauth, line 6210:
```suggestion
                        description: Defines name of Secret containing certificates for custom extensions signature validation
```

doc/api/dynakube-api-ref.md
```diff
@@ -138,6 +138,7 @@
 |:-|:-|:-|:-|
 |`annotations`|Adds additional annotations to the ExtensionExecutionController pods|-|object|
 |`customConfig`|Defines name of ConfigMap containing custom configuration file|-|string|
+|`customExtensionCertificates`|Defines name of Secret containing certificate files|-|string|
```
StefanHauth, line 141:
```suggestion
|`customExtensionCertificates`|Defines name of Secret containing certificates for custom extensions signature validation|-|string|
```

StefanHauth (changes requested):

config/crd/bases/dynatrace.com_dynakubes.yaml
andriisoldatenko: fixed in fdd4b780

config/helm/chart/default/templates/Common/crd/dynatrace-operator-crd.yaml
andriisoldatenko: fixed in fdd4b780

StefanHauth (dismissed):
Tested it on PoC, LGTM.

pkg/controllers/dynakube/extension/eec/statefulset.go
```diff
@@ -47,6 +48,8 @@ const (
 	envEecHttpsPrivKeyPathPem   = httpsCertMountPath + "/" + consts.TLSKeyDataName
 	// Volume names and paths
 	eecTokenMountPath                  = "/var/lib/dynatrace/remotepluginmodule/secrets/tokens"
+	eecCustomCertificateMountPath      = "/var/lib/dynatrace/remotepluginmodule/secrets/extensions"
+	customCertificateVolumeName        = "extension-custom-certs"
```
0sewa0, line 52: why does some const have an `eec` prefix in their name, while others don't ?
(especially considering that we are in the `eec` package)

pkg/api/v1beta3/dynakube/extensions.go
```diff
@@ -50,6 +50,10 @@ type ExtensionExecutionControllerSpec struct {
 	// +kubebuilder:validation:Optional
 	CustomConfig string `json:"customConfig,omitempty"`
 
+	// Defines name of Secret containing certificates for custom extensions signature validation
+	// +kubebuilder:validation:Optional
+	CustomExtensionCertificates string `json:"customExtensionCertificates,omitempty"`
```
0sewa0, line 55: Couldn't it be just `Certificates`?

1. As of right now, it reads very "stuttery", do we need `Extension` in the name, when it is already part of the `ExtensionExecution...` section?
2. If the user has to specify them they are already `Custom`, no? Example: We don't have a `CustomLabels` field, but a `Labels` field, as "set-by-the-user" already means `Custom` (IMO)
   - Sidenote: We have `CustomPullSecret` which I have the same problems with.

Would the name `Certificates` confuse customers, because the of the higher level `TrustedCAs` field? (it is kinda confusing me 😅 )
- Maybe name it for what is used for: (if it's that specific)
    - According to the doc string, it could be: `SignatureValidationCAs` 
- Or just tell the user to put it into the `TrustedCAs` ConfigMap 🤷 
    - (This is way out of scope for this PR) `TrustedCAs` is a ConfigMap, that could have multiple keys (it only has 1 right now, `certs`), wouldn't it be nicer to have different keys for different certs (an `extensions` key for this cert)?

TLDR:  I am just questioning the whole way we ask the user to specify the certificates 😅  (IMO it's confusing)

andriisoldatenko: @StefanHauth  can you please reply here?

pkg/controllers/dynakube/extension/eec/statefulset.go
andriisoldatenko: on it,.

andriisoldatenko: @0sewa0 fixed in 60ddd312

pkg/api/v1beta3/dynakube/extensions.go
0sewa0: had discussions offline:
- The separate field is necessary, so that is fine, implementation also looks fine
- The **problem** is that the name `customExtensionCertificates` is rather confusing, we should be more specific.
   - This might stem from the fact that our current certificate-related fields(`trustedCAs` and `tlsSecretName`) are also kinda badly named  😬

StefanHauth: `CustomExtension` is a known and used  term by itself in the world of extensions. It stands for extensions, that customers build themselves as opposite to extensions that come ootb in Dynatrace. Those certificates are used to validate the signature of custom extensions. 
These certificates have nothing to do with validating HTTPs connections, hence I don't see them in the trusteCAs, which get propagated to all components. These "CustomExtension" certificates are very, very specific to the EEC.

`CustomExtensionSignatureValidationCertificates` would be the over-correct name for them, but that's a wee bit longish.

andriisoldatenko: that's why we have description:
```
                        description: Defines name of Secret containing certificates
                          for custom extensions signature validation
```
anyway.

are we good to merging it? I need approvals guys!
cc @StefanHauth / @0sewa0

0sewa0: it make sense, thank you for the explanation! 😘

StefanHauth: IMHO the field name very well describes its purpose for someone who is familiar with the extensions topic.

PR #3761, opened 2024-09-10
WIP Do not discard mutating results of all mutators, if one fails.

StefanHauth:
## Description

Make sure one failing pod mutator doesn't block all others.

## How can this be tested?

TBD

### DIFF
--- a/pkg/webhook/mutation/pod/init_container.go
+++ b/pkg/webhook/mutation/pod/init_container.go
@@ -149,9 +149,8 @@ func addSeccompProfile(ctx *corev1.SecurityContext, dk dynakube.DynaKube) {
 }
 
 func updateContainerInfo(request *dtwebhook.ReinvocationRequest, installContainer *corev1.Container) bool {
-	pod := request.Pod
 	if installContainer == nil {
-		installContainer = findInstallContainer(pod.Spec.InitContainers)
+		installContainer = findInstallContainer(request.Pod.Spec.InitContainers)
 	}
 
 	newContainers := request.NewContainers(containerIsInjected)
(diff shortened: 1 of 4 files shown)


pkg/webhook/mutation/pod/init_container.go
```diff
@@ -149,9 +149,8 @@ func addSeccompProfile(ctx *corev1.SecurityContext, dk dynakube.DynaKube) {
 }
 
 func updateContainerInfo(request *dtwebhook.ReinvocationRequest, installContainer *corev1.Container) bool {
-	pod := request.Pod
 	if installContainer == nil {
-		installContainer = findInstallContainer(pod.Spec.InitContainers)
+		installContainer = findInstallContainer(request.Pod.Spec.InitContainers)
```
andriisoldatenko, line 153: why do we need this change?

StefanHauth: Just streamlined the code a bit.

pkg/webhook/mutation/pod/webhook.go
```diff
@@ -166,24 +164,26 @@ func (wh *webhook) handlePodMutation(ctx context.Context, mutationRequest *dtweb
 			continue
 		}
 
-		if err := mutator.Mutate(ctx, mutationRequest); err != nil {
-			return err
+		resultMutationRequest := mutationRequest.DeepCopy()
+		if err := mutator.Mutate(ctx, resultMutationRequest); err != nil {
+			log.Error(err, "pod mutator failed", "pod", mutationRequest.Pod.Name)
```
0sewa0, line 169: you could collect the errors instead of logging each of them, like [here](https://github.com/Dynatrace/dynatrace-operator/blob/1fad0e42ee963470471b8e9bb556473679c5c725/pkg/controllers/dynakube/controller.go#L388)

StefanHauth:
We will go all-or-nothing. If one mutator fails all mutations shall be discarded.

PR #3763, opened 2024-09-10
Replace validation logic with kubebuilder tags

waodim:
## Description

With this validation logic regarding the serviceAccount name of the edgeconnect is replaced by simply adding a kubebuilder tag to make the field required. As there is anyways a default name set with another kubebuilder tag, I am not sure if it is needed at all. Please give me feedback there.

https://dt-rnd.atlassian.net/browse/K8S-10716

I went through all of our validations and did NOT find any other function that can be replaced by a kubebuilder tag. Please also have a look, maybe you find something, I missed.

Here is the list of all kubebuilder tags:

https://book.kubebuilder.io/reference/markers/crd-validation

Give me feedback if you find something, that I can replace.

## How can this be tested?

- Unit tests
- Otherwise it is not really testable, as there is always a default name for the serviceAccount


pkg/api/v1alpha2/edgeconnect/edgeconnect_types.go
```diff
@@ -57,7 +57,8 @@ type EdgeConnectSpec struct { //nolint:revive
 
 	// ServiceAccountName that allows EdgeConnect to access the Kubernetes API
 	// +kubebuilder:default:=dynatrace-edgeconnect
-	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+	// +kubebuilder:validation:Required
+	ServiceAccountName string `json:"serviceAccountName"`
```
0sewa0, line 61: does having both the `default` and `required` make sense?

if not set -> `default` is used
the `required` check will never be hit

all things considered, IMO the only necessary change is that the `isInvalidServiceName` function should be removed as it's never used (due to the `default`)

waodim: Yes I pointed that out in the description and hoped, that anyone kind of confirms my thoughts. Thanks, will remove the required tag again.

waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3763/commits/07540e5000683e05f15abed6ad07d749e9cb0c18


 Please do not forget to still have a look, maybe you find some validation func I can replace with a tag, that I missed.

gkrenn: It would make sense to add a `+kubebuilder:validation:MinLength=1` annotation in my opinion. Otherwise empty values would still be possible as only nil values trigger the default value logic.

To check just set `serviceAccountName: ""`

waodim: I think a regex pattern matching would be better here, as with `+kubebuilder:validation:MinLength=1` `" "` would still be possible.

With `// +kubebuilder:validation:Pattern=\w` I get

`The EdgeConnect "alkr-ec" is invalid: spec.serviceAccountName: Invalid value: "": spec.serviceAccountName in body should match '\w'
`

`The EdgeConnect "alkr-ec" is invalid: spec.serviceAccountName: Invalid value: " ": spec.serviceAccountName in body should match '\w'
`

But all in all I do not know if we should remove the validation function at all and if it would not be better to just revert the changes again... I am not sure if we can cover everything in an "unhacky" way by using kubebuilder tags. What do you guys think?

0sewa0: we can't and shouldn't try to protect from every possible dumb way

if we want to overcomplicate the check then yes, probably better to do it with go code

but the go code you removed would also be fooled by " "

I would go with Gabriel's suggestion of `+kubebuilder:validation:MinLength=1`
And pls check all other `+kubebuilder:default` for `strings` and add `+kubebuilder:validation:MinLength=1` (if it make sense)

waodim: Okay, thanks for the feedback, will do.

waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3763/commits/af7b7b4c03d813ccf51243c23947a13c4035e6c1

did not find other default strings that needed change.


This is the resulting error message:

`The EdgeConnect "alkr-ec" is invalid: spec.serviceAccountName: Invalid value: "": spec.serviceAccountName in body should be at least 1 chars long`

I did not find a way to overwrite this message somehow, so I guess we have to go with it. (Just do not like this message that is why I looked for ways to overwrite)

andriisoldatenko:
tested in my cluster: LGTM
```
secret/edge-connect-client-secret created
The EdgeConnect "andrii-test" is invalid: spec.serviceAccountName: Invalid value: "": spec.serviceAccountName in body should be at least 1 chars long
```

andriisoldatenko (approved):

StefanHauth:
IMHO this is wrong. The ServiceAccount shouldn't be mandatory to be given in the EC CR. It is only mandatory if K8S automation is enabled because then special permissions have to be granted by customers. Otherwise, the SA created on operator deployment shall be used by default.

StefanHauth (changes requested):
IMHO this is wrong. The SA is only mandatory if K8S automation feature is used. Otherwise the default SA will be set by the operator in the EC deployment.

pkg/api/v1alpha2/edgeconnect/edgeconnect_types.go
0sewa0: not fully, if someone set it directly to `serviceAccountName: ""` then it would have caught it

0sewa0 (approved):

PR #3776, opened 2024-09-11
Add check for deepcopy files to CI

gkrenn:
## Description

This PR does 2 things: 
- It speeds up the deepcopy command and adds it to the manifest. This speed improvement is achieved by only searching in the relevant folders
- It adds a check to the CI to verify that the deepcopy files were updated if needed 

[Jira](https://dt-rnd.atlassian.net/browse/K8S-10776)

## How can this be tested?

- change file in the api folder that generates a deepcopy file
- run `make manifests` and check if deepcopy file was updated

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -103,6 +103,14 @@ jobs:
         id: deadcode
         run: |
           make go/deadcode
+      - name: Deepcopy generation
+        id: deepcopy
+        run: |
+          make manifests/deepcopy
+      - name: Check if deepcopy generation changed files
+        id: check-deepcopy
+        run: |
+          git diff --exit-code
 
   security:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -103,6 +103,14 @@ jobs:
         id: deadcode
         run: |
           make go/deadcode
+      - name: Deepcopy generation
```
chrismuellner, line 106: it kind of fits linting, but could we move this to a dedicated job that is separate from linting so it's clear what fails and we can extend it with other checks in the future

gkrenn: Done

0sewa0 (approved):

chrismuellner (approved):

PR #3778, opened 2024-09-11
Reconcile and mount self-signed TLS certificate for EEC and OtelC

albertogdd:
## Description

[K8S-10820](https://dt-rnd.atlassian.net/browse/K8S-10820)

- New util library for TLS certificate generation
- EEC reconciler now also reconciles the self-signed certificate (if needed, only when no `tlsRefName` is specified)
- Always mount TLS certificate both on EEC and OtelC

## How can this be tested?

- Deploy DynaKube with
```
spec:
  extensions:
    prometheus:
      enabled: true
  templates:
    extensionExecutionController:
      imageRef:
        repository: extk8sregistry.azurecr.io/eec/dynatrace-eec
        tag: <latest-tag>
  activeGate:
    tlsSecretName: dynakube-ag-tls ## you have to create this one manually
```
- Check that `dynakube-extensions-controller-tls` secret gets reconciled correctly (when spec.templates.extensionsExecutionController.tlsRefName` is NOT set)
- Check that tls certificate gets added as volume mount in EEC and OtelC StatefulSets


pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +81,57 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+	}
+
+	if r.dk.Spec.Templates.ExtensionExecutionController.TlsRefName != "" {
+		if secret != nil {
+			return query.Delete(ctx, secret)
+		}
+	} else {
+		if k8serrors.IsNotFound(err) {
+			// TODO: domain/altNames/ip ?
+			cert, err := certificates.New("", []string{}, "")
```
andriisoldatenko, line 103: is it something we need to clarify?

cc @StefanHauth

pkg/util/certificates/certificates.go
```diff
@@ -1,13 +1,133 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	netIp := net.ParseIP(ip)
+
+	cert := &x509.Certificate{
+		SerialNumber: serialNumber,
+		Subject: pkix.Name{
+			Country:            []string{"AT"},
```
andriisoldatenko, line 49: can we extract to constants all these strings?

pkg/util/certificates/certificates.go
```diff
@@ -1,13 +1,133 @@
...
+			Locality:           []string{"Linz"},
+			Organization:       []string{"Dynatrace"},
+			OrganizationalUnit: []string{"Operator"},
+			CommonName:         domain,
+		},
+		DNSNames:    altNames,
+		IPAddresses: []net.IP{netIp},
+
+		NotBefore: time.Now(),
+		NotAfter:  time.Now().Add(7 * 24 * time.Hour),
```
andriisoldatenko, line 60: same here, we should keep in mind we don't recreate when it's expired, so at least we can replace it with global var.

pkg/util/certificates/certificates.go
```diff
@@ -1,13 +1,133 @@
...
+		Bytes: pkx509Encoded,
+	})
+
+	c.signedCert = certPem
+	c.signedPk = pkPem
+
+	return nil
+}
+
+func (c *Certificate) CaSign(ca *x509.Certificate, caPk *ecdsa.PrivateKey) error {
```
andriisoldatenko, line 93:
```suggestion
func (c *Certificate) CAsign(ca *x509.Certificate, caPk *ecdsa.PrivateKey) error {
```
CA is abbreviation, so should not be used in this style.

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: More info in this regard here
https://bitbucket.lab.dynatrace.org/users/stefan.hauth/repos/extensions-demo/browse/preview/cert-request.txt
I will try using the same alt names

pkg/util/certificates/certificates.go
albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/8f912b88edf5f695f95af496983cced770f5ef14)

pkg/util/certificates/certificates.go
albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/8f912b88edf5f695f95af496983cced770f5ef14)

pkg/util/certificates/certificates.go
albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/8f912b88edf5f695f95af496983cced770f5ef14)

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: [FIxed](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/8f912b88edf5f695f95af496983cced770f5ef14)

pkg/controllers/dynakube/extension/consts/consts.go
```diff
@@ -20,11 +20,23 @@ const (
...
+	ExtensionsTlsSecretSuffix = "-extensions-controller-tls"
 
 	// TLSKeyDataName is the key used to store a TLS private key in the secret's data field.
 	TLSKeyDataName = "tls.key"
 
 	// TLSCrtDataName is the key used to store a TLS certificate in the secret's data field.
 	TLSCrtDataName = "tls.crt"
 )
+
+func GetCertificateAltNames(dkName string) []string {
```
andriisoldatenko, line 32: please don't put functions in constants files. create new file or package, similar I force to do in @wepudt  PR.

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +83,87 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+		return err
+	}
+
+	return nil
+}
+
+func (r *reconciler) reconcileTlsSecretExpiration(ctx context.Context, secret *corev1.Secret) error {
+	// WIP
+	return nil
+}
```
andriisoldatenko, line 169: you can put it just here `GetTlsSecretName` I mean this one,

pkg/controllers/dynakube/extension/eec/statefulset.go
```diff
@@ -17,7 +17,7 @@ import (
 	"golang.org/x/net/context"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
andriisoldatenko, line 20: linting error, please fix it,

pkg/controllers/dynakube/extension/otel/statefulset.go
```diff
@@ -224,6 +222,12 @@ func setImagePullSecrets(imagePullSecrets []corev1.LocalObjectReference) func(o
 }
 
 func setVolumes(dk *dynakube.DynaKube) func(o *appsv1.StatefulSet) {
+	tlsSecretName := dk.Name + consts.ExtensionsTlsSecretSuffix
+
+	if tlsSecretName != "" {
```
andriisoldatenko, line 227: looks wrong, you just set it to constant on line 225, but now you are trying to compare with empty string 🤔  am I miss something?

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +83,87 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+	secret, err := query.Get(ctx, types.NamespacedName{
+		Name:      consts.GetTlsSecretName(r.dk.Name),
+		Namespace: r.dk.Namespace,
+	})
+
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
+
+	if r.dk.Spec.Templates.ExtensionExecutionController.TlsRefName != "" {
```
waodim, line 99: Saw that this is used several times, should we abstract it for the sake of readability?

pkg/controllers/dynakube/extension/eec/statefulset.go
```diff
@@ -295,7 +294,7 @@ func buildContainerVolumeMounts(dk *dynakube.DynaKube) []corev1.VolumeMount {
 
 func setVolumes(dk *dynakube.DynaKube) func(o *appsv1.StatefulSet) {
 	return func(o *appsv1.StatefulSet) {
-		tlsSecretName := extensionsControllerTlsSecretName
+		tlsSecretName := dk.Name + consts.ExtensionsTlsSecretSuffix
```
andriisoldatenko, line 297: i found we use it quite often, can we create a function or method? to handle this logic easily in future?

pkg/controllers/dynakube/extension/eec/reconciler.go
andriisoldatenko: good catch, I saw we do it via method on dk.

pkg/controllers/dynakube/extension/eec/statefulset.go
albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/c1fb51c95e19d48604f6c04fc85b977cb720dace)

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/c1fb51c95e19d48604f6c04fc85b977cb720dace

pkg/controllers/dynakube/extension/otel/statefulset.go
albertogdd: Yes, missed it :D

Fixed

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: Fixed

pkg/controllers/dynakube/extension/consts/consts.go
albertogdd: Fixed

pkg/controllers/dynakube/extension/eec/reconciler.go
andriisoldatenko: commit not found https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/8f912b88edf5f695f95af496983cced770f5ef14

albertogdd: Links broke because of force pushes in between, but the fix is there

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +84,104 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+	secret, err := query.Get(ctx, types.NamespacedName{
+		Name:      getTLSSecretName(r.dk.Name),
+		Namespace: r.dk.Namespace,
+	})
+
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
+
+	if r.dk.Spec.Templates.ExtensionExecutionController.TlsRefName != "" {
```
andriisoldatenko, line 100: I still don't understand this check, here we introduced new method `GetExtensionsTlsSecretName `, but on line 100 we check only if user set custom TlsRefName, what about default one? and on line 92 we use alternative method?

pkg/controllers/dynakube/extension/eec/reconciler_test.go
```diff
@@ -659,7 +659,7 @@ func TestVolumes(t *testing.T) {
 				Name: httpsCertVolumeName,
 				VolumeSource: corev1.VolumeSource{
 					Secret: &corev1.SecretVolumeSource{
-						SecretName: extensionsControllerTlsSecretName,
+						SecretName: dk.Name + consts.ExtensionsTlsSecretSuffix,
```
andriisoldatenko, line 662: are you intentionally using here string concatination?

pkg/controllers/dynakube/extension/otel/statefulset.go
```diff
@@ -293,13 +289,11 @@ func buildContainerVolumeMounts(dk *dynakube.DynaKube) []corev1.VolumeMount {
...
 
-	if dk.Spec.Templates.ExtensionExecutionController.TlsRefName != "" {
-		vm = append(vm, corev1.VolumeMount{
-			Name:      consts.ExtensionsCustomTlsCertificate,
-			MountPath: customEecTlsCertificatePath,
-			ReadOnly:  true,
-		})
-	}
+	vm = append(vm, corev1.VolumeMount{
+		Name:      dk.Name + consts.ExtensionsTlsSecretSuffix,
```
andriisoldatenko, line 293: what about here? why not use method?

pkg/util/certificates/certificates.go
```diff
@@ -1,13 +1,136 @@
...
+	"errors"
+	"math/big"
 	"time"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
 )
 
+const (
+	intSerialNumberLimit  = 128
+	defaultCertExpiration = 7 * 24 * time.Hour
```
andriisoldatenko, line 19: why 7 days? cc @StefanHauth do we want short live Certificate here?

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: We still support if the user sets `Spec.Templates.ExtensionExecutionController.TlsRefName`. Hence, here I check if the user has set it, and if so, I delete whatever existing self-signed cert was created before.

We only create the self-signed certificate if the user doesn't set it themselves.

pkg/controllers/dynakube/extension/eec/reconciler_test.go
albertogdd: Thanks, I missed updating this to the new function :D

pkg/controllers/dynakube/extension/otel/statefulset.go
albertogdd: Fixed

pkg/util/certificates/certificates.go
albertogdd: I asked Stefan, and he mentioned not having an expiration date. I'll update it.

pkg/controllers/dynakube/extension/eec/reconciler.go
andriisoldatenko: yeah make sense.

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +85,109 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *reconciler) createOrUpdateTLSSecret(ctx context.Context) error {
+	cert, err := certificates.New()
```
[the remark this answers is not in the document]
albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/3660490530f3de7087e93592579179710217f5c7)

albertogdd:
Currently looking into switching to signature algorithm `sha256WithRSAEncryption`...

albertogdd:
> Currently looking into switching to signature algorithm `sha256WithRSAEncryption`...

Done! Generated cert example
```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            a6:9c:dc:b5:5b:99:f9:7d:aa:69:6f:0e:8e:fd:38:13
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: C = AT, ST = UA, L = Linz, O = Dynatrace, OU = Operator
        Validity
            Not Before: Sep 16 14:18:16 2024 GMT
            Not After : Sep 16 14:18:16 2025 GMT
        Subject: C = AT, ST = UA, L = Linz, O = Dynatrace, OU = Operator
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                Public-Key: (2048 bit)
                Modulus:
                    00:e6:01:33:d0:4a:64:01:ff:30:35:f8:32:7e:51:
                    05:96:b3:46:4e:62:6c:e1:13:53:76:9d:9b:57:44:
                    4b:72:79:31:1c:d7:70:f0:d0:db:f4:fa:f5:66:02:
                    21:58:d4:9d:23:0a:0d:09:33:06:07:31:4e:fa:35:
                    8f:88:6d:03:58:7a:fa:38:ae:f7:e5:d4:df:1a:5a:
                    3c:b6:b9:21:1b:58:f0:bd:a8:cf:bc:c9:7b:67:aa:
                    cb:4f:68:c2:5f:ef:ed:aa:8a:87:40:39:a4:e2:08:
                    21:fa:a8:36:61:07:8e:d7:21:17:37:86:2d:25:b4:
                    46:66:63:98:3b:dc:a9:e0:b3:09:d3:b7:33:82:ed:
                    16:22:60:77:f8:48:a7:4e:b9:23:cd:1c:51:fe:ba:
                    28:53:f4:69:15:ca:b2:4f:04:53:52:2d:dd:49:cb:
                    ec:a3:d8:00:10:b1:77:02:4a:7f:77:0d:56:4b:2b:
                    ea:28:16:03:b3:2a:b6:a8:44:cd:45:29:40:71:19:
                    65:87:17:3f:6e:bd:80:99:3e:bb:4f:d8:65:87:14:
                    98:73:13:f9:a5:8d:d4:95:f4:03:53:4b:5f:92:e7:
                    1c:fb:38:62:a7:d8:59:84:5f:32:61:85:8d:ee:db:
                    3b:7b:84:fa:fe:ff:87:25:c0:07:ee:d1:75:43:a8:
                    42:0f
                Exponent: 65537 (0x10001)
        X509v3 extensions:
            X509v3 Key Usage: critical
                Key Encipherment, Data Encipherment
            X509v3 Extended Key Usage: 
                TLS Web Server Authentication
            X509v3 Basic Constraints: critical
                CA:TRUE
            X509v3 Subject Key Identifier: 
                E0:81:62:B9:07:D3:FA:C1:1E:FC:C8:78:84:9D:C8:12:FD:5F:90:5E
            X509v3 Subject Alternative Name: 
                DNS:dynakube-extensions-controller.dynatrace, DNS:dynakube-extensions-controller.dynatrace.svc, DNS:dynakube-extensions-controller.dynatrace.svc.cluster.local
    Signature Algorithm: sha256WithRSAEncryption
    Signature Value:
        4b:18:b5:b3:90:88:fa:58:c3:3e:0d:bf:c3:c7:32:51:0b:18:
        3d:70:07:47:01:78:42:3f:c5:c4:84:b9:d2:32:63:eb:12:61:
        b0:a0:a1:c1:66:36:46:4e:71:b4:0c:55:ed:c6:45:b7:07:f5:
        c5:42:cb:fe:fe:1e:32:f9:f1:e4:58:62:ed:5b:65:d3:14:4d:
        60:1a:65:6e:71:44:d2:e5:88:af:ec:33:1e:94:a2:69:33:20:
        cd:75:3b:59:5f:96:d6:b2:8c:2f:eb:b2:f1:69:7e:af:ad:19:
        7e:bf:60:97:5f:12:e4:ea:d2:ae:20:03:b2:fe:c7:8c:93:16:
        19:b7:f4:bd:4a:a5:30:56:b1:a5:bb:0e:2a:f5:8e:8f:53:fd:
        c8:42:e2:45:0d:0b:db:79:d3:0a:74:cb:eb:dc:01:5e:00:2f:
        2f:c9:9f:fb:24:5f:a7:f0:6b:9f:fc:5a:af:55:c9:ec:46:ef:
        f9:cc:5c:e1:05:55:88:eb:bd:f0:3b:59:7f:2d:a3:3a:2a:a8:
        e2:89:e1:f6:fb:1f:ce:51:4a:f1:c6:3a:87:b2:61:d0:39:62:
        97:ce:1f:59:3e:98:c7:4b:b1:99:1f:74:a1:16:c1:a4:38:c7:
        b1:08:5f:38:74:88:ab:6d:ed:65:e1:c7:da:62:ee:eb:d4:51:
        ed:c9:dd:a6
```

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +84,105 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+		Name:      getSelfSignedTLSSecretName(r.dk.Name),
+		Namespace: r.dk.Namespace,
+	})
+
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
+
+	if r.dk.Spec.Templates.ExtensionExecutionController.TlsRefName != "" {
+		if secret != nil {
```
StefanHauth, line 101: The secret is also not nil if err == NotFound. In that case the delete fails, because the resource name is empty.
```suggestion
		if secret != nil && !k8serrors.IsNotFound(err) {
```

StefanHauth (changes requested):

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +84,106 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+	_, err = query.CreateOrUpdate(ctx, secret)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *reconciler) reconcileTLSSecretExpiration(ctx context.Context, secret *corev1.Secret) error {
+	isValid, err := certificates.ValidateCertificateExpiration(secret.Data[consts.TLSCrtDataName], consts.ExtensionsSelfSignedTLSRenewalThreshold, time.Now(), log)
```
andriisoldatenko, line 169: can you please use timeProvider like other reconcilers, instead of hardcoding time.Now(), it will help to unit-test it in future.

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +84,106 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+		err = r.reconcileTLSSecretExpiration(ctx, secret)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *reconciler) createOrUpdateTLSSecret(ctx context.Context) error {
```
andriisoldatenko, line 127: also what I found, there is no unit-tests for reconcile secret method.

aorcholski (commented):
EEC doesn't like the certificate
```
2024-09-17 09:20:53.992 UTC [0000000e] warning [native] EEC REST server (0.0.0.0:14599) startup failed because: setSslConfig(): unsupported (DECODER routines)
```

pkg/api/v1beta3/dynakube/extensions.go
```diff
@@ -102,3 +103,11 @@ type ImageRefSpec struct {
 	// Indicates a tag of the image to use
 	Tag string `json:"tag,omitempty"`
 }
+
+func (dk *DynaKube) GetExtensionsTLSSecretName() string {
```
0sewa0, line 107: shouldn't this be in the `extensions_props.go` ?

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +84,106 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 
 	return r.createOrUpdateStatefulset(ctx)
 }
+
+func (r *reconciler) reconcileTLSSecret(ctx context.Context) error {
```
0sewa0, line 88: IMO you could simplify the flow:

the delete is safe for not found, see  https://github.com/Dynatrace/dynatrace-operator/blob/4407c4ee51002a764ba82dd00eea46b635d09525/pkg/util/kubeobjects/internal/query/query.go#L74

so what you could do is the following:

```suggestion
func (r *reconciler) reconcileTLSSecret(ctx context.Context) error {
	query := k8ssecret.Query(r.client, r.client, log)

	if r.dk.Spec.Templates.ExtensionExecutionController.TlsRefName != ""  {  // you could also create a property for this
		err := query.Delete()
		if err != nil {
			return err
		}
		return nil
	}
	... // do the rest
	
```

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: Refactored to this (we only get no error when the secret exists)
```
if err == nil
```

pkg/controllers/dynakube/extension/consts/consts.go
```diff
@@ -20,11 +22,13 @@ const (
 	ExtensionsCollectorComPort        = 14599
 	ExtensionsCollectorTargetPortName = "collector-com"
 
-	ExtensionsCustomTlsCertificate = "custom-tls-certificates"
+	ExtensionsSelfSignedTlsSecretSuffix     = "-extensions-controller-tls"
+	ExtensionsSelfSignedTlsCommonNameSuffix = "-extensions-controller.dynatrace"
+	ExtensionsSelfSignedTlsRenewalThreshold = 12 * time.Hour
 
-	// TLSKeyDataName is the key used to store a TLS private key in the secret's data field.
-	TLSKeyDataName = "tls.key"
```
andriisoldatenko, line 26: any reason for this change?

albertogdd: Just to keep all TLS references the same in code, as we have it like this from CRD `TlsRefName` for example
`dk.Spec.Templates.ExtensionExecutionController.TlsRefName`

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: Fixed

pkg/api/v1beta3/dynakube/extensions.go
albertogdd: Fixed!

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: Fixed, you're right thanks!!

pkg/controllers/dynakube/extension/consts/consts.go
andriisoldatenko: TLS is an abbreviation, so the correct camelCase is TLS. https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)?redirectedfrom=MSDN

pkg/api/v1beta3/dynakube/extensions_props.go
```diff
@@ -12,3 +14,15 @@ func (dk *DynaKube) IsExtensionsEnabled() bool {
 func (dk *DynaKube) PrometheusEnabled() bool {
 	return dk.Spec.Extensions.Prometheus.Enabled
 }
+
+func (dk *DynaKube) GetExtensionsTlsRefName() string {
```
chrismuellner, line 18:
```suggestion
func (dk *DynaKube) ExtensionsTlsRefName() string {
```

pkg/api/v1beta3/dynakube/extensions_props.go
```diff
@@ -12,3 +14,15 @@ func (dk *DynaKube) IsExtensionsEnabled() bool {
...
+	return dk.Spec.Templates.ExtensionExecutionController.TlsRefName
+}
+
+func (dk *DynaKube) GetExtensionsTlsSecretName() string {
```
chrismuellner, line 22:
```suggestion
func (dk *DynaKube) ExtensionsTlsSecretName() string {
```

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +86,100 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 
 	return r.createOrUpdateStatefulset(ctx)
 }
+
+func (r *reconciler) reconcileTlsSecret(ctx context.Context) error {
+	query := k8ssecret.Query(r.client, r.client, log)
+
+	if r.dk.GetExtensionsTlsRefName() != "" {
```
chrismuellner, line 93: can this be moved to properties and renamed so it's clearer what this does, like `NeedsSelfSignedTls()`?

pkg/controllers/dynakube/extension/consts/consts.go
andriisoldatenko: it's a good point to make it similar, but if you try to grep current codebase you will find many inconsistency in different places:

```
er.p12`: certificate+key pair in pkcs12 format<br/>`password`: passphrase to read server.p12|-|string|

pkg/webhook/mutation/pod/oneagent/volumes_test.go
76:                                     TlsSecretName: "test",

assets/samples/dynakube/v1beta1/kubernetesObservability.yaml
140:    # Optional: Specifies the name of a secret containing a TLS certificate, a TLS key and the TLS key's password to be used by ActiveGate i
nstances
143:    # server.p12: TLS certificate and TLS key pair in pkcs12 format
144:    # password: passphrase to decrypt the TLS certificate and TLS key pair
146:    # tlsSecretName: "my-tls-secret"
```

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +86,100 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+
+	err = r.reconcileTlsSecretExpiration(ctx, secret)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *reconciler) createOrUpdateTlsSecret(ctx context.Context) error {
```
chrismuellner, line 123: can this whole implementation be made more generic so it can be reused for other components as well?

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -56,6 +67,11 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 		return nil
 	}
 
+	err := r.reconcileTlsSecret(ctx)
```
0sewa0, line 70: Shouldn't this be done after other checks like `if r.dk.Status.KubeSystemUUID == "" `?
Logic: why create a secret if the status is not even ready so we will fail either way

pkg/controllers/dynakube/extension/consts/consts.go
andriisoldatenko:
```
pkg/clients/dynatrace/client.go
5:      "crypto/tls"
163:// SkipCertificateValidation creates an Option that specifies whether validation of the server's TLS
169:                    if t.TLSClientConfig == nil {
170:                            t.TLSClientConfig = &tls.Config{} //nolint:gosec // fix is expected to be delivered soon
173:                    t.TLSClientConfig.InsecureSkipVerify = true
217:            if t.TLSClientConfig == nil {
218:                    t.TLSClientConfig = &tls.Config{} //nolint:gosec // fix is expected to be delivered soon
221:            t.TLSClientConfig.RootCAs = rootCAs

pkg/oci/registry/client_test.go
70:             assert.True(t, transport.TLSClientConfig.InsecureSkipVerify)
77:             assert.False(t, transport.TLSClientConfig.InsecureSkipVerify)

pkg/oci/registry/client.go
5:      "crypto/tls"
211:    if transport.TLSClientConfig == nil {
212:            transport.TLSClientConfig = &tls.Config{} //nolint:gosec
215:    transport.TLSClientConfig.RootCAs = rootCAs
221:    if transport.TLSClientConfig == nil {
222:            transport.TLSClientConfig = &tls.Config{} //nolint:gosec
225:    transport.TLSClientConfig.InsecureSkipVerify = skipCertCheck
```

chrismuellner (changes requested):

pkg/controllers/dynakube/extension/consts/consts.go
albertogdd: Its okay! I'm undoing it and making all TLS :+1:

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: Yes, I thought the same, on it

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -70,3 +86,100 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 
 	return r.createOrUpdateStatefulset(ctx)
 }
+
+func (r *reconciler) reconcileTlsSecret(ctx context.Context) error {
+	query := k8ssecret.Query(r.client, r.client, log)
+
+	if r.dk.GetExtensionsTlsRefName() != "" {
+		return query.DeleteForNamespaces(ctx, getSelfSignedTlsSecretName(r.dk.Name), []string{r.dk.Namespace})
```
0sewa0, line 94: Why not use just `return query.Delete(...)` ? As long as the secret has a correct metadata section it will work

pkg/controllers/dynakube/extension/eec/reconciler.go
0sewa0: also will this secret be cleaned-up in case `Extensions` get disabled?

this leads me to: 
wouldn't it be cleaner to have this secret reconciliation in a separate reconciler that is called somewhere at: https://github.com/Dynatrace/dynatrace-operator/blob/da71084911b1db6fab4d02ab1ca1e3b71507e4fe/pkg/controllers/dynakube/extension/reconciler.go#L47 ?

you can ignore my first comment of 
> why create a secret if the status is not even ready so we will fail either way
I thought this reconciler is the "main" `Extensions` reconciler, but this is only for the statefulset, so IMO the secret reconcile shouldn't be here

pkg/controllers/dynakube/extension/consts/consts.go
albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/f78d0db05eff5ec59242a155f958d6917857c392)

andriisoldatenko: thanks!

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/effefb722de190fad5f3e2e58bdde6121b9e64dc)

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/fd1af2ce0cfdf8f64e04761bf426598e57afa58e)
Adding more unit tests now

pkg/controllers/dynakube/extension/tls.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/certificates"
+	k8slabels "github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/labels"
+	k8ssecret "github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/secret"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (r *reconciler) reconcileTLSSecret(ctx context.Context) error {
```
0sewa0, line 17: why don't you have a `tls` package, that has a `Reconciler` that can handle the reconcile of this secret following the pattern we already have in the codebase?

albertogdd: Good idea, [fixed here](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/0b3b87c16ea21c9127cbe963a26c54ad21410e2e)
Unit tests incoming

albertogdd: [Unit tests added](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/7c824e188f80a032a4823a0e0aabb88002f66b51)

pkg/controllers/dynakube/extension/eec/reconciler.go
albertogdd: [Unit tests added](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/7c824e188f80a032a4823a0e0aabb88002f66b51)

aorcholski (approved):
LGTM, I tested it with custom secret and automatically created secret

pkg/controllers/dynakube/extension/tls/reconciler.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	if k8serrors.IsNotFound(err) {
+		err = r.createOrUpdateTLSSecret(ctx)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}
```
0sewa0, line 61: if I understand correctly, the scenario of "update" will never happen, only create

because `k8serrors.IsNotFound(err)` will be `false` if the secret already exists, so you will just `return nil`

you don't need that if
```suggestion
	if err != nil && !k8serrors.IsNotFound(err) {
		return err
	}

	err = r.createOrUpdateTLSSecret(ctx)
	if err != nil {
		return err
	}
```

albertogdd: [Fixed as discussed in Slack](https://github.com/Dynatrace/dynatrace-operator/pull/3778/commits/687ec32290257b5e583fafaead823bc9e427c06a)

StefanHauth (approved):

aorcholski (approved):

andriisoldatenko (approved):

PR #3783, opened 2024-09-12
Return early if owner is not well-known

0sewa0:
## Description

If we don't return early, when the owner is NOT well-known, then we will have a permission error when trying to get it with the client.

## How can this be tested?

tbd 😅

### DIFF
--- a/pkg/webhook/mutation/pod/metadata/workload.go
+++ b/pkg/webhook/mutation/pod/metadata/workload.go
@@ -65,6 +65,10 @@ func findRootOwner(ctx context.Context, clt client.Client, childObjectMetadata *
 				},
 			}
 
+			if !isWellKnownWorkload(parentObjectMetadata) {
+				return childObjectMetadata, nil
+			}
+
 			err = clt.Get(ctx, client.ObjectKey{Name: owner.Name, Namespace: objectMetadata.Namespace}, parentObjectMetadata)
 			if err != nil {
 				log.Error(err, "failed to query the object",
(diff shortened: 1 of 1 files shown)


pkg/webhook/mutation/pod/metadata/workload_test.go
```diff
@@ -211,3 +236,13 @@ func createTestWorkloadInfo() *workloadInfo {
 		name: "test",
 	}
 }
+
+func createFailK8sClient() client.Client {
```
andriisoldatenko, line 240: i think it's good practice to mark helpers as `t.Helper` https://pkg.go.dev/testing#T.Helper
>Helper marks the calling function as a test helper function. When printing file and line information, that function will be skipped. Helper may be called simultaneously from multiple goroutines.

andriisoldatenko: example https://github.com/dynatrace/dynatrace-operator/blob/cba0da7700f4adcec9fd44fb52bd0b81ed2005f2/test/helpers/registry/registry.go#L28-L36

0sewa0: surprised the linter is not complaining about it, but IMO it's fine here, it just creates an object 🤷

andriisoldatenko: it's up to you 👍 not blocking comment.

andriisoldatenko (approved):

pkg/webhook/mutation/pod/metadata/workload_test.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/3783/commits/218c5a125d1cbb6b714042d24751ba5aa5232f0b

I think its a good idea 👍

PR #3793, opened 2024-09-16
Fixup ondemand pipeline branch

andriisoldatenko:
Currently we use default branch, but we should use input.

image

### DIFF
--- a/.github/workflows/e2e-tests-ondemand.yaml
+++ b/.github/workflows/e2e-tests-ondemand.yaml
@@ -56,13 +56,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
         with:
-          ref: ${{ github.event.inputs.target }}
+          ref: ${{ github.event.inputs.target || 'release-1.3' }}
       - name: Run e2e test
         uses: ./.github/actions/run-e2e
         with:
           flc-namespace: ${{ format('dto-{0}-ondemand', matrix.platform ) }}
           flc-environment: ${{ format('dto-{0}-{1}',  matrix.platform, matrix.version ) }}
-          target-branch: ${{ github.event.inputs.target }}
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-tests-ondemand.yaml
```diff
@@ -55,12 +55,14 @@ jobs:
...
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          ref: ${{ github.event.inputs.target || 'release-1.3' }}
       - name: Run e2e test
         uses: ./.github/actions/run-e2e
         with:
           flc-namespace: ${{ format('dto-{0}-ondemand', matrix.platform ) }}
           flc-environment: ${{ format('dto-{0}-{1}',  matrix.platform, matrix.version ) }}
-          target-branch: ${{ github.event.inputs.target }}
+          target-branch: ${{ github.event.inputs.target || 'release-1.3' }}
```
chrismuellner, line 65: can the default target somehow be reused? having it hard coded in 2 places could lead to it not being updated correctly

chrismuellner (changes requested):

.github/workflows/e2e-tests-ondemand.yaml
andriisoldatenko: I agree, let me research...

andriisoldatenko: @chrismuellner https://github.com/orgs/community/discussions/26322
tl;dr it's not possible ...

andriisoldatenko: nope, that's the problem. Default value is used when it's triggered from UI, otherwise it's empty. wdyt https://github.com/Dynatrace/dynatrace-operator/pull/3793/commits/f6558b80652151981d078d6a5ebc9a0ba0e0a2f3 ?

chrismuellner (approved):

PR #3801, opened 2024-09-17
Validate collisions of logmodules and OneAgents in the validation webhook

gkrenn:
## Description

It is not possible to deploy the logmodule on the same node that the OneAgent.
To prevent misconfigurations we should add a validation to our validation webhook. 

[Jira](https://dt-rnd.atlassian.net/browse/K8S-10908)

## How can this be tested?

- Deploy a logmodule dynakube
- Deploy a CNFS dynakube => should fail


pkg/api/v1beta1/dynakube/validation/oneagent.go
```diff
@@ -17,14 +18,18 @@ const (
 `
 	errorImageFieldSetWithoutCSIFlag = `The DynaKube's specification tries to enable ApplicationMonitoring mode and get the respective image, but the CSI driver is not enabled.`
 
-	errorNodeSelectorConflict = `The DynaKube's specification tries to specify a nodeSelector conflicts with an another Dynakube's nodeSelector, which is not supported.
-The conflicting Dynakube: %s
-`
+	errorNodeSelectorConflict = `The DynaKube specification attempts to deploy a %s, which conflicts with another DynaKube's %s. Only one per namespace is supported.
```
waodim, line 21:
```suggestion
	errorNodeSelectorConflict = `The DynaKube specification attempts to deploy a %s, which conflicts with another DynaKube's %s. Only one Agent per node is supported.
```

waodim: Minor nitpick to add Agent here

The other thing is that only one agent is allowed per node and not namespace, or am I wrong there?

0sewa0 (dismissed):
other than @waodim 's comment, it works and looks "as ok as it can" considering the need to validate for each old version as well
IMO, we should rethink how we validate when multiple versions are used, as this can get out of hand quickly. WDYT?

My testing setup:
```
kind: DynaKube
metadata:
  name: zib50933zib50933zib50933zib50933zib50933
  namespace: dynatrace
spec:
  apiUrl: https://zib50933.dev.dynatracelabs.com/api
  tokens: zib50933
  dynatraceApiRequestThreshold: 15
  logModule:
    enabled: true
  templates:
    logModule:
      nodeSelector:
        dk-log: deploy
      imageRef:
        tag: latest

---
kind: DynaKube
metadata:
  name: zib
  namespace: dynatrace
spec:
  apiUrl: https://zib50933.dev.dynatracelabs.com/api
  tokens: zib50933
  enableIstio: true
  dynatraceApiRequestThreshold: 15

  oneAgent:
    cloudNativeFullStack:
      namespaceSelector:
        matchLabels:
          monitored/by: v2
      nodeSelector:
          dk-v2: deploy
      tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
         operator: Exists
       - effect: NoSchedule
         key: node-role.kubernetes.io/control-plane
         operator: Exists
---
kind: DynaKube
metadata:
  name: zib50933-old
  namespace: dynatrace
spec:
  apiUrl: https://zib50933.dev.dynatracelabs.com/api
  tokens: zib50933

  namespaceSelector:
     matchLabels:
       monitored/by: v1

  oneAgent:
   cloudNativeFullStack:
     nodeSelector:
      dk-v1: deploy
     tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
         operator: Exists
       - effect: NoSchedule
         key: node-role.kubernetes.io/control-plane
         operator: Exists

```

tried different order of application, and if no `nodeSelector` is specified, the DynaKube gets rejected as expected

pkg/api/v1beta1/dynakube/validation/oneagent.go
gkrenn: Nice catch, done

waodim (dismissed):
code lgtm, and it also worked for me, but I agree with @0sewa0 that IF possible we should think about other ways of validation without having to touch every api version's validation when adding something.

0sewa0 (approved):

pkg/api/v1beta1/dynakube/validation/oneagent.go
```diff
@@ -17,14 +18,18 @@ const (
 `
 	errorImageFieldSetWithoutCSIFlag = `The DynaKube's specification tries to enable ApplicationMonitoring mode and get the respective image, but the CSI driver is not enabled.`
 
-	errorNodeSelectorConflict = `The DynaKube's specification tries to specify a nodeSelector conflicts with an another Dynakube's nodeSelector, which is not supported.
-The conflicting Dynakube: %s
-`
+	errorNodeSelectorConflict = `The DynaKube specification attempts to deploy a %s, which conflicts with another DynaKube's %s. Only one Agent per node is supported.
```
chrismuellner, line 21: this does not give the information anymore, that the `nodeSelector` is the problem.

pkg/api/v1beta1/dynakube/validation/oneagent.go
```diff
@@ -17,14 +18,18 @@ const (
...
+Conflicting DynaKube: %s`
```
chrismuellner, line 22: can more than one Dynakube be conflicting? Example: 2 Dynakubes with node selector and I try to apply one without a node selector

chrismuellner (dismissed):

pkg/api/v1beta1/dynakube/validation/oneagent.go
gkrenn: That was kind of on purpose, because the `nodeSelector` is not really true. It also pops up if no `nodeSelectors` are specified and 2 dynakubes are colliding. 

So I would rather state that the dynakubes are colliding and which type of agent is the problem. So WDYF, should we still change it back? 🤔

pkg/api/v1beta1/dynakube/validation/oneagent.go
gkrenn: The error will state the first collision detected, just as we did only for the OneAgent before

pkg/api/v1beta1/dynakube/validation/oneagent.go
chrismuellner: only way to fix it is by specifying a node selector, so that should be in the message somewhere

pkg/api/v1beta1/dynakube/validation/oneagent.go
chrismuellner: I know, it's more confusing with oneagent/logmodule both existing now and only the first of them included in the warning

gkrenn: So what do you suggest?

pkg/api/v1beta1/dynakube/validation/oneagent.go
gkrenn: Updated

pkg/api/v1beta1/dynakube/validation/oneagent.go
0sewa0: couldn't just have something like:

```
var conflicts []string
for _, item := range validDynakubes.Items {
...

	if hasConflictingMatchLabels(oneAgentNodeSelector, item.OneAgentNodeSelector()) {
		...
		conflicts = append(conflicts, fmt.Sprintf(errorNodeSelectorConflict, oneAgentComponentName, oneAgentComponentName, item.Name))
	}
...
}

if len(conflicts)>0 {
	...	// reject
}
```

be careful not to just spit out a wall of text though 😅

gkrenn: Will try to update it tomorrow to consider multiple dynakubes and features 👍

pkg/api/v1beta3/dynakube/validation/logmodule.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	errorConflictingLogModule = "The DynaKube's specification tries to enable LogModule in a namespace where another DynaKube already deploys the OneAgent, which is not supported. The conflicting DynaKube: %s"
```
waodim, line 12: Is this error message here so specific because we know that there is a LogModule enabled?

Because I was a little confused when I first did it this way:
- apply v1beta2 and then v1beta3

and after that:

- apply v1beta3 and then v1beta2

as I got different error messages.

pkg/api/v1beta3/dynakube/validation/logmodule.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+		if item.NeedsOneAgent() {
+			if hasConflictingMatchLabels(dk.LogModuleNodeSelector(), item.OneAgentNodeSelector()) {
+				log.Info("requested dynakube has conflicting LogModule nodeSelector", "name", dk.Name, "namespace", dk.Namespace)
+
+				return fmt.Sprintf(errorConflictingLogModule, item.Name)
+			}
+		}
+	}
```
0sewa0, line 45: here you could also collect all the conflicting DynaKubes 😅

chrismuellner: just make sure to only list the conflicting dynakubes. if we output the full error for each dynakube we could easily hit the warning limit

pkg/api/v1beta3/dynakube/validation/logmodule.go
gkrenn: Yes, that's indeed the case, this PR adds basically 2 checks:
- Block if the log agent collides with another dynakube: That's the error message you marked. 
- Block if the one agent collides with another dynakube: That's the other one. But you are right, in the other one we can also specify OneAgent, as we know this one is checked. => Will update the other message, thx

pkg/api/v1beta3/dynakube/validation/logmodule.go
gkrenn: We don't need that, as there can only be 1 conflicting dynakube.

pkg/api/v1beta3/dynakube/validation/logmodule.go
gkrenn: Updated

pkg/api/v1beta3/dynakube/validation/logmodule.go
0sewa0: I could be many, but it's not very realistic.

Example
```
kind: DynaKube
metadata:
  name: cloudnative
  namespace: dynatrace
spec:
  apiUrl: https://zib50933.dev.dynatracelabs.com/api
  tokens: zib50933
  enableIstio: true
  dynatraceApiRequestThreshold: 15

  oneAgent:
    cloudNativeFullStack:
      namespaceSelector:
        matchLabels:
          monitored/by: v2
      nodeSelector:
          dk-v2: deploy
      tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
         operator: Exists
       - effect: NoSchedule
         key: node-role.kubernetes.io/control-plane
         operator: Exists
---
kind: DynaKube
metadata:
  name: cloudnative-old
  namespace: dynatrace
spec:
  apiUrl: https://zib50933.dev.dynatracelabs.com/api
  tokens: zib50933

  namespaceSelector:
     matchLabels:
       monitored/by: v1

  oneAgent:
   cloudNativeFullStack:
     nodeSelector:
      dk-v1: deploy
     tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
         operator: Exists
       - effect: NoSchedule
         key: node-role.kubernetes.io/control-plane
         operator: Exists
---
kind: DynaKube
metadata:
  name: log-module
  namespace: dynatrace
spec:
  apiUrl: https://zib50933.dev.dynatracelabs.com/api
  tokens: zib50933
  dynatraceApiRequestThreshold: 15
  logModule:
    enabled: true
```

gkrenn: 😮  - That's good, an edge case I didn't think of, will adapt the code and add it as unit test. Thx

gkrenn: Done

waodim (approved):
lgtm

PR #3802, opened 2024-09-18
Revert "Move CR name length validation to kubebuilder markers (#3665)"

0sewa0:
This reverts commit cc4809062d8933060c6ae8547111cea7ec66dbf7.

## Description

Unfortunately, the x-validation properties in the CRD is not supported in OCP 4.10 and 4.11, but the operator does support these versions. So we have to go back to checking the length by hand.

## How can this be tested?

On OCP 4.10 or 4.11 cluster you can successfully apply the CRD
And on every version the name length limit is enforced by the validation


andriisoldatenko (approved):
practically LGTM, tested in OCP-4-10

PR #3806, opened 2024-09-18
Allow to configure the annotations of service accounts within helm values

gkrenn:
## Description

The operator deploys service accounts for all components deployed in the cluster. 
This PR enables adding labels to these service accounts.

## How can this be tested?

Set the rbac values in the values.yaml and check if the labels are there in the service accounts.

### DIFF
--- a/config/helm/chart/default/tests/Common/activegate/serviceaccount-activegate_test.yaml
+++ b/config/helm/chart/default/tests/Common/activegate/serviceaccount-activegate_test.yaml
@@ -1,0 +1,15 @@
+suite: test serviceaccount for activegate
+templates:
+  - Common/activegate/serviceaccount-activegate.yaml
+tests:
+  - it: should exist
+    set:
+      rbac.activeGate.annotations:
+        test: test
+    asserts:
+        - isKind:
+            of: ServiceAccount
+        - equal:
+            path: metadata.annotations
+            value:
+              test: test
--- a/config/helm/chart/default/values.yaml
+++ b/config/helm/chart/default/values.yaml
@@ -194,3 +194,16 @@ csidriver:
         cpu: 20m
         memory: 30Mi
 
+rbac:
+  activeGate:
+    annotations:
+    kubernetesMonitoring:
+      annotations:
+  edgeConnect:
+    annotations:
+  extensions:
+    annotations:
+  logModule:
+    annotations:
+  oneAgent:
+    annotations:
--- a/config/helm/chart/default/values.yaml
+++ b/config/helm/chart/default/values.yaml
@@ -197,8 +197,6 @@ csidriver:
 rbac:
   activeGate:
     annotations:
-    kubernetesMonitoring:
-      annotations:
   edgeConnect:
     annotations:
   extensions:
--- a/config/helm/chart/default/values.yaml
+++ b/config/helm/chart/default/values.yaml
@@ -196,12 +196,12 @@ csidriver:
 
 rbac:
   activeGate:
-    annotations:
+    annotations: {}
   edgeConnect:
-    annotations:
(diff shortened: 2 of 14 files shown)


config/helm/chart/default/values.yaml
```diff
@@ -194,3 +194,16 @@ csidriver:
         cpu: 20m
         memory: 30Mi
 
+rbac:
+  activeGate:
+    annotations:
+    kubernetesMonitoring:
+      annotations:
```
0sewa0, line 201: would it make sense to have a separate section for the "base" `activeGate` ServiceAccount ?
because currently, it is kinda confusing. 
It could either work like:
- `rbac.activeGate.annotations` be on all `activeGate` ServiceAccounts and `rbac.activeGate.kubernetesMonitoring.annotation` would overrule it IF defined
- both `annotations` sections are handled independently are totally separate


```suggestion
  activeGate:
     default: # horrible name, I know
    	annotations:
    kubernetesMonitoring:
      annotations:
```

gkrenn: I think it is best if we keep the structure as is, but just implement the override logic. 

So the KubernetesMonitoring sa always uses the ActiveGate if not defined.

Not really a fan of the default section

0sewa0: me neither
the problem kinda comes from the fact that the "default" AG serviceAccount has 0 roles assigned to it :) 
as only the `kubernetesMonitoring` capability is interacting with the kube-api

0sewa0: btw: the override logic will not make it better, probably worse 😅

gkrenn: Hmm, then lets separate them completely. 🤔

chrismuellner: does it even make sense to separate the `annotations`? You'll use either of the service accounts, but both should have the same annotations.

So keep `rbac.activegate.annotations` and put it on activegate and kubernetes monitoring service account.

gkrenn: Done

config/helm/chart/default/values.yaml
```diff
@@ -194,3 +194,14 @@ csidriver:
         cpu: 20m
         memory: 30Mi
 
+rbac:
+  activeGate:
+    annotations:
```
chrismuellner, line 199:
```suggestion
    annotations: {}
```
can we set an explicit empty value so it doesn't look so forgotten?

config/helm/chart/default/tests/Common/activegate/serviceaccount-activegate_test.yaml
```diff
@@ -0,0 +1,15 @@
+suite: test serviceaccount for activegate
+templates:
+  - Common/activegate/serviceaccount-activegate.yaml
+tests:
+  - it: should exist
+    set:
+      rbac.activeGate.annotations:
```
chrismuellner, line 7: missing a test for kubernetes monitoring service account

gkrenn: nice catch 👍

config/helm/chart/default/values.yaml
gkrenn: Done

config/helm/chart/default/tests/Common/activegate/serviceaccount-activegate_test.yaml
gkrenn: Done

0sewa0 (approved):

chrismuellner (approved):

PR #3809, opened 2024-09-19
Do not inject into ignored namespaces after adding a namespace to the annotation

wepudt:
[K8S-11291](https://dt-rnd.atlassian.net/browse/K8S-11291)

## Description

Problem: When updating a DK with the annotation `feature.dynatrace.com/ignored-namespaces` to include a namespace that should NOT be injected we still inject if it was injected before.

This PR corrects this behaviour.

## How can this be tested?

1. Apply a DK.
2. Deploy a sample application into a namespace `sample-ns` and make sure it get's injected.
3. Update the DK:
   ```
   metadata:
     annotations:
       feature.dynatrace.com/ignored-namespaces: '["sample-ns.*", "^kube-.*", "^gke-.*", "^gmp-.*",  "dynatrace"]'
   ```
4. Restart the pod of the sample application.
   Make sure the sample application is not injected.

### DIFF
--- a/pkg/injection/namespace/mapper/mapper_test.go
+++ b/pkg/injection/namespace/mapper/mapper_test.go
@@ -211,7 +211,7 @@ func TestUpdateNamespace(t *testing.T) {
 		require.True(t, updated)
 		assert.Empty(t, namespace.Labels)
 	})
-	t.Run("Do not inject into pods when adding namespace to ignored namespaces", func(t *testing.T) {
+	t.Run("Remove injection label from ignored-namespaces if present from previous setup", func(t *testing.T) {
 		labels := map[string]string{"test": "selector"}
 		dk := createDynakubeWithAppInject("dk-test", convertToLabelSelector(labels))
 		namespace := createNamespace("test-namespace", labels)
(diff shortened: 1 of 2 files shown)


waodim (approved):
code lgtm - and it worked for me - samples were not injected.

pkg/injection/namespace/mapper/mapper_test.go
```diff
@@ -211,4 +211,22 @@ func TestUpdateNamespace(t *testing.T) {
 		require.True(t, updated)
 		assert.Empty(t, namespace.Labels)
 	})
+	t.Run("Do not inject into pods when adding namespace to ignored namespaces", func(t *testing.T) {
```
0sewa0, line 214: nitpick:
```suggestion
	t.Run("Remove injection label from ignored-namespaces if present from previous setup", func(t *testing.T) {
```

0sewa0 (approved):
Just a minor complaint about test naming, otherwise looks good, small and smart code change

pkg/injection/namespace/mapper/mapper_test.go
wepudt: fixed in 7ddca951561e58b96cc169fbc6bcda71eb7857d0

PR #3813, opened 2024-09-19
Fix ondemand pipeline target branch if run scheduled

luhi-DT:
## Description
https://dt-rnd.atlassian.net/browse/K8S-11301

On-demand pipelines are running for main branch if they are run on schedule.

## How can this be tested?

Check if e2e tests are run for correct release branch on schedule

### DIFF
--- a/.github/workflows/e2e-tests-ondemand.yaml
+++ b/.github/workflows/e2e-tests-ondemand.yaml
@@ -4,12 +4,9 @@ on:
   schedule:
     # every Mon/Wed/Fri at 08:00 UTC
     - cron: 0 8 * * 1,3,5
-  workflow_dispatch:
-    inputs:
-      target:
-        description: Target branch to run E2E tests over
-        required: true
-        default: release-1.3
+
+env:
+  branch: 'release-1.3'
 
 permissions:
   checks: write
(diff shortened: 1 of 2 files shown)


andriisoldatenko (approved):

.github/workflows/e2e-tests-ondemand.yaml
```diff
@@ -4,12 +4,9 @@ on:
...
     - cron: 0 8 * * 1,3,5
-  workflow_dispatch:
-    inputs:
-      target:
-        description: Target branch to run E2E tests over
-        required: true
-        default: release-1.3
+
+env:
+  branch: 'release-1.3'
```
andriisoldatenko, line 9: the only problem with this approach, if I want to trigger hotfix, which is not yet merged to release branch, i can't.

luhi-DT: you should not run hot fix on all pipelines imho

andriisoldatenko: okey anyway let's try with this fix.

0sewa0 (approved):

PR #3825, opened 2024-09-20
Add logmodule daemonset reconciler

0sewa0:
## Description
[K8S-10617](https://dt-rnd.atlassian.net/browse/K8S-10617)

[Adds the logmodule reconciler that will create a `DaemonSet` based on the `spec.templates.logModule`.](https://github.com/Dynatrace/dynatrace-operator/commit/bc994dc4d1e376438424b608d0aa892b815cbc6e)

Has some minor adjustments for the `configsecret` reconciler:
- [Fix Server propertie in deployment.conf](https://github.com/Dynatrace/dynatrace-operator/commit/d4c057b8d1a76ea5d68f66951d2735a2d4bc55c3)
- [Add no-communication-hosts handling](https://github.com/Dynatrace/dynatrace-operator/commit/d3a2cc54bed856540fa6419b93e361b95deb3db9)

#### What to configure in the `DynaKube`
1. `spec.logModule.enabled: true`
2. Because the final default image location is not known, we have an image in our private gcr, that GKE clusters can use without issues:
```
templates:
    logModule:
      imageRef:
        tag: latest
```
3. ✨ <Whatever else you want to configure in the `templates`> ✨ 

#### What to configure in the tenant
1. Configure log-ingest rules:
   - This will be done by the operator in the future. (upcoming ticket)
   - My tenant has it at `https://zib50933.dev.dynatracelabs.com/ui/settings/builtin:logmonitoring.log-storage-settings`
   - Press `Add rule`, this is a "catch-all" example:
image

#### What to check to see it "work"
If you see logs such as:
```

Process log content for /var/log/pods/kube-system_container-watcher-54sft_4a77a473-759c-4cf3-9db7-378637be0c30/container-watcher/#.log#


Successfully parsed 2 entries for /var/log/pods/kube-system_konnectivity-agent-8667cbffbd-sxwjk_da91698c-245c-4f49-9da5-725025937cc6/konnectivity-agent/#.log # and advanced file pointer by 306 after reading 612 bytes, current entry count: 2. Last file size: 515949, new position: 515643 
```


pkg/controllers/dynakube/logmodule/daemonset/container.go
```diff
@@ -0,0 +1,83 @@
+package daemonset
+
+import (
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/address"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	defaultImageRepo = "registry.lab.dynatrace.org/oneagent/dynatrace-logmodule-amd64" // TODO: finalize
```
waodim, line 10: what do you mean by finalize?

pkg/api/v1beta3/dynakube/extensions.go
```diff
@@ -98,3 +98,15 @@ type ImageRefSpec struct {
 	// Indicates a tag of the image to use
 	Tag string `json:"tag,omitempty"`
 }
+
+func (ref ImageRefSpec) String(defaultRepo, defaultTag string) string {
```
waodim, line 102: do we really need to pass default repo and tag as parameters? Not that it matters a lot but it somehow looks weird because of the word "default"

pkg/controllers/dynakube/logmodule/daemonset/reconciler_test.go
```diff
@@ -0,0 +1,276 @@
...
+		dk.Spec.Templates.LogModule.Labels = customLabels
+
+		reconciler := NewReconciler(nil,
+			nil, dk)
+		daemonset, err := reconciler.generateDaemonSet()
+		require.NoError(t, err)
+		require.NotNil(t, daemonset)
+
+		assert.Subset(t, daemonset.Spec.Template.Labels, customLabels)
+		// assert.Subset(t, daemonset.Labels, customLabels) TODO: Should this be the case?
```
chrismuellner, line 148: should it? Checking the activegate, custom labels are only put on the pods

chrismuellner (dismissed):

pkg/controllers/dynakube/logmodule/daemonset/container.go
chrismuellner: that's not the public registry, so it needs to be updated to the correct image before release/merge

pkg/api/v1beta3/dynakube/extensions.go
0sewa0: `ImageRefSpec` is not only used by the LogModule, but I agree, there is a better solution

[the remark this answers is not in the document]
0sewa0: I will move the `ImageRefSpec` elsewhere, it is not extensions specific

pkg/controllers/dynakube/logmodule/daemonset/reconciler.go
```diff
@@ -0,0 +1,119 @@
...
+
+	err = hasher.AddAnnotation(ds)
+	if err != nil {
+		return nil, err
+	}
+
+	return ds, nil
+}
+
+func GetName(dkName string) string {
```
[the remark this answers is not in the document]
0sewa0: we are in the `daemonset` package, that would be imported (probably with an alias) like `lmdaemonset`, then it would read `lmdaemonset.GetName`. So we avoid stuttering.

pkg/controllers/dynakube/logmodule/daemonset/volumes.go
```diff
@@ -0,0 +1,151 @@
...
+	dtLibVolumePathTemplate = "/tmp/dynatrace-logmodule-%s"
+	dtLogVolumeName         = "var-log-dynatrace"
+	dtLogVolumeMountPath    = "/var/log/dynatrace"
+
+	// for the logs that the logmodule will ingest
+	podLogsVolumeName       = "var-log-pods"
+	podLogsVolumePath       = "/var/log/pods"
+	dockerLogsVolumeName    = "var-lib-docker-containers"
+	dockerLogsVolumePath    = "/var/lib/docker/containers"
+	containerLogsVolumeName = "var-log-containers"
```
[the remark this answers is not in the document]
0sewa0: maybe more human readable version: `container-logs` ?

pkg/controllers/dynakube/logmodule/daemonset/reconciler_test.go
0sewa0: that is why I am asking :D

pkg/controllers/dynakube/logmodule/daemonset/reconciler.go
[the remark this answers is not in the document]
chrismuellner: the current usage _in_ the log module package confused me during review

we already have the name for other components in properties: https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/api/v1beta2/dynakube/properties.go#L100

pkg/controllers/dynakube/logmodule/daemonset/volumes.go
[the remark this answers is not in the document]
chrismuellner: either is fine. I was going from least specific to most specific information for my suggestion

pkg/controllers/dynakube/logmodule/daemonset/reconciler_test.go
chrismuellner: let's keep consistency with other components

pkg/controllers/dynakube/logmodule/daemonset/reconciler.go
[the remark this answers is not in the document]
0sewa0: the fact that we did it a certain way doesn't make it correct 😅

the problem with putting everything in the properties that it just becomes a bloated mess (it already is)
- pretty sure the properties you brought up as example have been there since the oneagent-operator days
- there are several examples for not doing it like this:
   - https://github.com/Dynatrace/dynatrace-operator/blob/266389ab053979d6935dfd140e360b6b72ac8b3f/pkg/controllers/dynakube/deploymentmetadata/reconciler.go#L86-L89
   - https://github.com/Dynatrace/dynatrace-operator/blob/266389ab053979d6935dfd140e360b6b72ac8b3f/pkg/controllers/dynakube/processmoduleconfigsecret/reconciler.go#L213
   - https://github.com/Dynatrace/dynatrace-operator/blob/266389ab053979d6935dfd140e360b6b72ac8b3f/pkg/controllers/dynakube/proxy/reconciler.go#L128-L131
the problem with putting such things closest to where it's used could be that it's harder to discover
- you would have to look at the different packages to see, instead of the single big package
   - I personally find it more readable, that there is a `logmodule` package that has a subpackage `daemonset`, so if I want to get something `logmodule-daemonset` specific, I will just look at that package, and it will have no noise from the ActiveGate stuff, for example.

also, if we want to have each bigger section of the spec in their own package, then such properties like `OneAgentDaemonSetName` should no longer be in the properties:
- because, you would like a syntax like `dk.OneAgent().DaemonSetName()`, but this is not possible, as it would depend on the name of the `dk`, but the `DaemonSetName` method can't access that
  - you can do some tricks like, that each section has a hidden `ParentName` field that is not part of the yaml and is set in the `dk.OneAgent()` func for example


TLDR: The way we go about this is a bit inconsistent, but having everything in the properties IMO is not how we should do it, as it "doesn't scale", properties would be an ever growing package.

pkg/api/v1beta3/dynakube/extensions.go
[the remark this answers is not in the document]
0sewa0: done

0sewa0: done

pkg/controllers/dynakube/logmodule/daemonset/reconciler.go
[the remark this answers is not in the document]
0sewa0: done, moved it to the props for now

pkg/controllers/dynakube/logmodule/daemonset/container_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+func TestSecurityContext(t *testing.T) {
+	t.Run("get base securityContext", func(t *testing.T) {
+		dk := dynakube.DynaKube{}
+		sc := getBaseSecurityContext(dk)
+
+		require.NotNil(t, sc)
+		require.NotEmpty(t, sc)
+
+		assert.NotNil(t, sc.Privileged)
```
aorcholski, line 88: Could we verify specific values True/False? It's hard requirement that Privileged is False. The same regarding other SecurityContext properties. We will be warned if want to change the default setting.

wepudt (approved):
lgtm

also ran on my cluster- all working as advertised

pkg/controllers/dynakube/logmodule/daemonset/container_test.go
0sewa0: IMO, that doesn't help more

I agree these tests are "tests for test's sake".
What these tests check is: "do we explicitly set the fields in the securityContext?"
- this is the "most sensible" test I could come up with
- I would be more OK with removing these tests(to reduce noise), then adding more exact ones, as that feels like having a test for checking the value of a `const`, which IMO is totally unnecessary 😅 

"dynamic" fields have a separate tests, to see if they are actually dynamic. (these are actual tests)

aorcholski (approved):
LGTM  , during test I found both messages

PR #3853, opened 2024-09-26
Add rbac turn-off option in Helm values.yaml

0sewa0:
[K8S-10824](https://dt-rnd.atlassian.net/browse/K8S-10824)

The following options have been added to the helm values.yaml:
```
rbac:
  activeGate:
    create: true
  edgeConnect:
    create: true
  extensions:
    create: true
  logModule:
    create: true
  oneAgent:
    create: true
```

Turning any of them to `false` will result in the following:
1. The related RBAC components (`Role`/`ClusterRole`/`ServiceAccount`) will not be deployed
2. The validation webhook will deny any `DynaKube`/`EdgeConnect` CR that tries to configure a feature relying on such RBAC components
   - This is done by creating a `ConfigMap` called `operator-config`, the content's of which is added to the envs of the webhook
     - This env is ONLY checked 1 time at startup

## How can this be tested?

Configure the `rbac.<module>.create` fields to your heart's content.
- Check the `install-config` `ConfigMap` reflect what you configred
- Try to deploy a CR that would use the disabled "module"

### DIFF
--- a/config/helm/chart/default/templates/Common/activegate/clusterrole-activegate.yaml
+++ b/config/helm/chart/default/templates/Common/activegate/clusterrole-activegate.yaml
@@ -1,5 +1,4 @@
-{{- if (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
-{{- if .Values.rbac.activeGate.create }}
+{{- if and .Values.rbac.activeGate.create (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
 # Copyright 2021 Dynatrace LLC
 
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -44,5 +43,4 @@ roleRef:
   kind: ClusterRole
   name: dynatrace-activegate
   apiGroup: rbac.authorization.k8s.io
-{{ end }}
 {{ end }}
--- a/config/helm/chart/default/templates/Common/activegate/clusterrole-activegate.yaml
+++ b/config/helm/chart/default/templates/Common/activegate/clusterrole-activegate.yaml
@@ -1,5 +1,5 @@
 {{- if (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
-
+{{- if .Values.rbac.activeGate.create }}
 # Copyright 2021 Dynatrace LLC
 
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -44,4 +44,5 @@ roleRef:
   kind: ClusterRole
   name: dynatrace-activegate
   apiGroup: rbac.authorization.k8s.io
-{{- end -}}
+{{ end }}
+{{ end }}
--- a/config/helm/chart/default/templates/Common/extensions-controller/role-extensions-controller.yaml
+++ b/config/helm/chart/default/templates/Common/extensions-controller/role-extensions-controller.yaml
@@ -1,5 +1,4 @@
-{{- if (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
-{{- if .Values.rbac.extensions.create }}
+{{- if and .Values.rbac.extensions.create (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
 # Copyright 2021 Dynatrace LLC
 
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -40,5 +39,4 @@ roleRef:
   kind: Role
   name: dynatrace-extensions-controller
   apiGroup: rbac.authorization.k8s.io
-{{ end }}
 {{ end }}
--- a/config/helm/chart/default/templates/Common/extensions-controller/role-extensions-controller.yaml
+++ b/config/helm/chart/default/templates/Common/extensions-controller/role-extensions-controller.yaml
@@ -1,4 +1,5 @@
 {{- if (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
+{{- if .Values.rbac.extensions.create }}
 # Copyright 2021 Dynatrace LLC
 
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -39,4 +40,5 @@ roleRef:
   kind: Role
   name: dynatrace-extensions-controller
   apiGroup: rbac.authorization.k8s.io
+{{ end }}
 {{ end }}
--- a/pkg/util/operatorconfig/modules.go
+++ b/pkg/util/operatorconfig/modules.go
@@ -1,0 +1,59 @@
+package operatorconfig
+
+import (
+	"encoding/json"
+	"os"
+	"sync"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
+)
+
+const (
+	modulesJsonEnv = "modules.json"
+)
+
+var (
+	once sync.Once
+
+	modules Modules
+
+	fallbackModules = Modules{
+		ActiveGate:  true,
+		OneAgent:    true,
+		Extensions:  true,
+		LogModule:   true,
+		EdgeConnect: true,
+	}
+
+	log = logd.Get().WithName("operator-config")
+)
+
+type Modules struct {
+	ActiveGate  bool `json:"activeGate"`
+	OneAgent    bool `json:"oneAgent"`
+	Extensions  bool `json:"extensions"`
+	LogModule   bool `json:"logModule"`
+	EdgeConnect bool `json:"edgeConnect"`
+}
+
+func GetModules() Modules {
+	once.Do(func() {
+		modulesJson := os.Getenv(modulesJsonEnv)
+		if modulesJson == "" {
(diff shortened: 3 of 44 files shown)


pkg/api/v1beta3/dynakube/validation/modules.go
```diff
@@ -0,0 +1,46 @@
+package validation
+
+import (
+	"context"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
+)
+
+const (
+	errorOneAgentModuleDisabled   = `OneAgent features has been disabled during Operator install. The necessary resources for deploying a OneAgent DaemonSet to work are not present on the cluster. Redeploy the Operator with all the necessary resources`
```
chrismuellner, line 10: can we reuse the constant and only change the component name? Would be easier to update the wording then.

chrismuellner: I prefer the way you phrased it for Edgeconnect:
> EdgeConnect has been disabled during Operator install. The necessary resources for EdgeConnect to work are not present on the cluster. Redeploy the Operator with all the necessary resources

No mention of features or workload necessary I think

pkg/util/operatorconfig/modules.go
```diff
@@ -0,0 +1,59 @@
+package operatorconfig
```
chrismuellner, line 1: `componentconfig`? We are not directly configuring the Operator, but configuring what components can be enabled.

waodim: I do not have the strongest opinion here, but I like `componentconfig` more tbh.

pkg/api/v1beta3/dynakube/validation/modules.go
waodim: Should we mention that it should be done in the `values.yaml`? Or do we trust that everyone knows this?

pkg/util/operatorconfig/modules.go
```diff
@@ -0,0 +1,59 @@
...
+	Extensions  bool `json:"extensions"`
+	LogModule   bool `json:"logModule"`
+	EdgeConnect bool `json:"edgeConnect"`
+}
+
+func GetModules() Modules {
+	once.Do(func() {
+		modulesJson := os.Getenv(modulesJsonEnv)
+		if modulesJson == "" {
+			log.Info("operator config envvar not set, using default", "envvar", modulesJsonEnv)
```
waodim, line 43:
```suggestion
			log.Info("operator-config envvar not set, using default", "envvar", modulesJsonEnv)
```

waodim: just for consistency, as you used a "-" in the next log message. (i know way too nitpicky)

pkg/api/v1beta1/dynakube/validation/modules.go
```diff
@@ -0,0 +1,28 @@
+package validation
+
+import (
+	"context"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta1/dynakube" //nolint:staticcheck
+)
+
+const (
+	errorOneAgentModuleDisabled   = `OneAgent features has been disabled during Operator install. The necessary resources for deploying a OneAgent DaemonSet to work are not present on the cluster. Redeploy the Operator with all the necessary resources`
```
waodim, line 10:
```suggestion
	errorOneAgentModuleDisabled   = `OneAgent features have been disabled during Operator install. The necessary resources for deploying a OneAgent DaemonSet to work are not present on the cluster. Redeploy the Operator with all the necessary resources`
```

waodim: same goes for other error messages

config/helm/chart/default/templates/Common/activegate/clusterrole-activegate.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
-
+{{- if .Values.rbac.activeGate.create }}
```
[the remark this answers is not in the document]
0sewa0: IMO its far cleaner to read than 
```
{{- if and ((eq (include "dynatrace-operator.openshiftOrOlm" .) "true") .Values.rbac.activeGate.create)}}
```
(would look something like this, but there can be typos 😅 )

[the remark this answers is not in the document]
chrismuellner: feel free to create functions to check these conditions

pkg/util/operatorconfig/modules.go
0sewa0: it is a config for the operator in general, to decide which components are "available"

you could argue that we only provide it to the webhook currently, so `webhookconfig`, but that may change in the future

Another I thought about is `installinfo`/`installconfig` as it is information related to the installation of the operator

`componentconfig` tells me nothing
- which component is the config for?

waodim: `operatorcomponentconfig` probably is too long right?

`installconfig` sounds good in my opinion.

config/helm/chart/default/templates/Common/activegate/clusterrole-activegate.yaml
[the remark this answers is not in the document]
0sewa0: you would just move the confusing logic to a different place

and in a way that you would have to introduce new functions for every component, that are very specific:
- `dynatrace-operator.activeGateRBACforOpenshiftOrOlm`
- `dynatrace-operator.oneAgentRBACforOpenshiftOrOlm`
- ...

and that is just a naming problem, but we are also in helm templates, I can't really create nice looking functions/abstractions, they will be just hacky and ugly. (also ALL of these would be single use functions)

If we don't like this "stacked if" I would just remove the `dynatrace-operator.openshiftOrOlm`, as it only is needed for `ClusterRoles` that only define the use for built-in SCCs

If we still need `dynatrace-operator.openshiftOrOlm` and don't like the "stacked if", I would much rather do the "1 ugly line way" over the "template-function for each singular case way" 😅

pkg/util/operatorconfig/modules.go
0sewa0: renamed it to`installconfig`

pkg/api/v1beta3/dynakube/validation/modules.go
0sewa0: In [Standardize the validation error message](https://github.com/Dynatrace/dynatrace-operator/pull/3853/commits/99409d2d33e6ed576a8200dfd1b1242a73e478e0) I tried to fix this, but depending if this goes in before or after https://github.com/Dynatrace/dynatrace-operator/pull/3856, I might do it differently

pkg/api/v1alpha1/edgeconnect/validation/module.go
```diff
@@ -4,10 +4,11 @@ import (
 	"context"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha1/edgeconnect"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/installconfig"
 )
 
-const (
-	errorModuleDisabled = `EdgeConnect has been disabled during Operator install. The necessary resources for EdgeConnect to work are not present on the cluster. Redeploy the Operator with all the necessary resources`
+var (
+	errorModuleDisabled = installconfig.GetModuleValidationErrorMessage("EdgeConnect")
```
waodim, line 11: very nice! 👍

pkg/util/operatorconfig/modules.go
0sewa0: made it nicer

config/helm/chart/default/templates/Common/activegate/clusterrole-activegate.yaml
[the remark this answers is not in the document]
0sewa0: I have seen the light, and fixed it in [Combine helm ifs](https://github.com/Dynatrace/dynatrace-operator/pull/3853/commits/8a9bf0a98f13cca266d3a871b72c90381cd1a13c) (also made some stuff more consistent)

config/helm/chart/default/templates/Common/extensions-controller/role-extensions-controller.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
+{{- if .Values.rbac.extensions.create }}
```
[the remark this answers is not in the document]
0sewa0: done 😉

waodim (dismissed):
tested it and it works - code lgtm

aorcholski (dismissed):
LGTM, works as expected

chrismuellner (approved):

waodim (approved):

PR #3855, opened 2024-09-26
Fix e2e framework not stopping if setup fails

luhi-DT:
TIcket: https://dt-rnd.atlassian.net/browse/K8S-11312

## Description

In the latests version of the e2e framework (0.4.0 ) has a bug in it that when a setup step fails it will only skip the other setup steps and not the whole test.

Which is unexpected behaviour for us, as logically, why would I want to run tests against an incorrectly setup env?

## How can this be tested?

Check e2e and check if the setup fails by having a wrong CRD e.g tests should stop


waodim (approved):
lgtm

.github/renovate.json5
```diff
@@ -154,6 +154,12 @@
       pinVersions: true,
       automerge: true,
     },
+    {
+      matchPackageNames: [
+        "sigs.k8s.io/e2e-framework",
```
chrismuellner, line 159: how will we notice the new version is out? we could disable automerge, as well, but that would mean we have to keep the PR open for a while...

luhi-DT: thats the point @0sewa0 and me discussed, we thought that not having this up to date for a while is not a problem, since the e2e tests are running anyway also on an older version. But we can also just disable auto-update i am also fine with this @0sewa0 your take on this?

0sewa0: the e2e framework has a single purpose, run our e2e tests reliably, if they do that on the current version, then I see 0 reason for upgrading
- we don't have to worry about security vulnerabilities (IMO the MAIN reason we update our dependencies), as it is not production code
- if the new version of the framework has "cool new stuff" in it, getting the new version via renovate will not let us know about the "cool new stuff" and it's value, somebody will have to search/discover it on their own anyway.
   - Sidenote: most e2e tests I saw written tend to be "copy-pasta" from other e2e tests, so I don't see people here being adventurous and searching for new ways to use the framework 😅 🙈

0sewa0 (approved):

.github/renovate.json5
andriisoldatenko: Can we just disable auto merge for this specific package?

andriisoldatenko: but i'm agree we can ignore for now upgrading since it's just for e2e tests.

andriisoldatenko (approved):

PR #3856, opened 2024-09-26
Use 1 validator per CR

0sewa0:
## Description

[K8S-11317](https://dt-rnd.atlassian.net/browse/K8S-11317)

We initially wanted to have the validation logic per version, which makes sense on paper, but not very realistic in reality, as most new validation logic needs to be added to all versions, and sometimes have to think about cross-version checks.

So now instead we use 1 custom-validator per CR, that can handle multiple version.
- For older versions, just use `ConvertTo`

We still need the 1 validation endpoint per version in the `ValidationWebhookConfiguration`
- It's how the framework works:
   - You define the webhook for a given version of the CR, so when the webhook receives the request it will convert it to that version.
     - So if you define only 1 webhook and point all the versions there, all but the 1 version will succeed, others will just fail due to a conversion error within the framework.

Sidenote: Will break https://github.com/Dynatrace/dynatrace-operator/pull/3853 but I can live with that, should be an easy merge conflict. (it's also my PR :P )

## How can this be tested?

Deploy 1 dk/ec for each version (broken or not) and check if the webhook reacts accordingly

Example: 3 different dk versions, all intentionally broken, and the webhook rejects them accordingly
```
Error from server (Forbidden): error when creating "./.vscode/deploy/dynakube_zib50933.yaml": admission webhook "v1beta3.dynakube.webhook.dynatrace.com" denied the request:
 1. The length limit for the name of a DynaKube is 40, because it is the base for the name of resources related to the DynaKube. (example: dkName-activegate-<some-hash>)
	The limit is necessary because kubernetes uses the name of some resources (example: StatefulSet) for the label value, which has a limit of 63 characters. (see https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)
Error from server (Forbidden): error when creating "./.vscode/deploy/dynakube_zib50933.yaml": admission webhook "v1beta2.dynakube.webhook.dynatrace.com" denied the request:
 1. The DynaKube's specification has an invalid API URL value set.
	Make sure you correctly specify the URL in your custom resource (including the /api postfix).

Error from server (Forbidden): error when creating "./.vscode/deploy/dynakube_zib50933.yaml": admission webhook "v1beta1.dynakube.webhook.dynatrace.com" denied the request:
 1. The DynaKube's specification has an invalid API URL value set.
	Make sure you correctly specify the URL in your custom resource (including the /api postfix).
```


cmd/webhook/builder.go
```diff
@@ -7,15 +7,12 @@ import (
...
 	edgeconnectv1alpha2 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha2/edgeconnect"
-	edgeconnectv1alpha2validation "github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha2/edgeconnect/validation"
 	dynakubev1beta1 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta1/dynakube" //nolint:staticcheck
-	dynakubev1beta1validation "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta1/dynakube/validation"
 	dynakubev1beta2 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube" //nolint:staticcheck
-	dynakubev1beta2validation "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta2/dynakube/validation"
 	dynakubev1beta3 "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
-	dynakubev1beta3validation "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube/validation"
+	dkvalidation "github.com/Dynatrace/dynatrace-operator/pkg/api/validation/dynakube"
+	ecvalidation "github.com/Dynatrace/dynatrace-operator/pkg/api/validation/edgeconnect"
```
andriisoldatenko, line 15: why not full name as it was before?
`dynakubevalidation`?

cmd/webhook/builder.go
```diff
@@ -7,15 +7,12 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/cmd/config"
```
chrismuellner, line 7: is there any loss in functionality by no longer validating older versions? do we need more safety checks during conversion now?

andriisoldatenko: I think we need to test it, it's hard to say now :)

pkg/api/validation/dynakube/validation.go
```diff
@@ -96,3 +106,24 @@ func (v *Validator) runValidators(ctx context.Context, validators []validatorFun
...
 }
+
+func getDynakube(obj runtime.Object) (dk *dynakube.DynaKube, err error) {
+	dk = &dynakube.DynaKube{}
+
+	switch v := obj.(type) {
+	case *dynakube.DynaKube:
+		dk = v
+	case *v1beta2.DynaKube:
+		err = v.ConvertTo(dk)
```
andriisoldatenko, line 117: Would it be useful to log an error, at least, that we can't convert?

andriisoldatenko: oh good catch, i missed named return var :)

andriisoldatenko:
code LGTM, i've tested EdgeConnect and it worked.

andriisoldatenko (approved):

waodim (approved):
tested it with various `dks`and `ecs` - worked for me.

andriisoldatenko (approved):

PR #3863, opened 2024-09-30
Add debug documentation and make commands for easier debugging

gkrenn:
## Description

This PR adds documentation how to debug the different components of the operator.
In addition some makefile commands are added to simplify the process.

[Jira](https://dt-rnd.atlassian.net/browse/K8S-10742)

## How can this be tested?

Try to follow the guide and start debugging.


hack/make/debug/csi-provisioner.patch
```diff
@@ -0,0 +1,81 @@
+From: A happy user <happy@example.com>
```
0sewa0, line 1: Couldn't this be done using helm instead + a `debug.Dockerfile`? (I guarantee that someone will commit the changes)

Like have a little `_debug.tpl` that have template functions that will add the same things to the manifests if a flag `debug` is `true`

IMO, these git patches can get easily outdated

hack/make/debug/debug.mk
```diff
@@ -0,0 +1,24 @@
+## Prepares all necessary CSI server resources for debugging (e.g. set limits, change build flags and startup command, ...)
+debug/prepare/csi-server:
+	git apply hack/make/debug/csi-server.patch
+
+## Prepares all necessary CSI provisioner resources for debugging (e.g. set limits, change build flags and startup command, ...)
+debug/prepare/csi-provisioner:
+	git apply hack/make/debug/csi-provisioner.patch
+
+## Forwards the CSI server port to localhost:40000
+debug/tunnel/csi-driver:
```
0sewa0, line 10: maybe I am incorrect but wouldn't you need to rebuild/push the operator image? (you are not really mentioning that, but the patch messes with the dockerfile)

So IMO there should be a make command that does the "build part"
- and it would make the git patches safer in general (these patches are only needed for build and deploy)
   - in another comment I mention an alternative for the git patches, but this comment is still valid 😅  

wouldn't make targets like these make more sense:
- `make debug/build`: 
   1. apply patches to the Dockerfile
   2. build/push debug image
   3. remove patches from the Dockerfile
- `make debug/deploy`:
   1. apply patches to the yamls
   2. deploy the operator via helm
   3. remove patches from the yamls

hack/make/debug/csi-provisioner.patch
andriisoldatenko: Also patch won't work as soon as we change something in this file. So better to introduce DEBUG env var or so. or use template.

gkrenn: Did it, as we once agreed on that we won't use a second Dockerfile in the repo, but yes, I agree that it would make more sense with a flag. Will change it 👍  - Thx for the input

gkrenn: Added a debug flag

hack/make/debug/debug.mk
gkrenn: Updated the docu to make it more clear and also added a `make debug/build` command

config/helm/chart/default/templates/Common/csi/daemonset.yaml
```diff
@@ -131,6 +142,19 @@ spec:
       - name: provisioner
         image: {{ include "dynatrace-operator.image" . }}
         imagePullPolicy: Always
+        {{- if eq .Values.debug true }}
```
waodim, line 145: If we adress the debug flag from the `values` file, shouldn't we add a default debug flag value to `values.yaml`? Or am I missing something that I do not get?

gkrenn: If it is not set in the values file, helm will use the default value which is `false`. Didn't want to specify it to not confuse the user, but maybe it's a better to add it to the `values.yaml` to be more explicit. - Will add it 👍

0sewa0: you don't have to put it into the `values.yaml`, that is just the "default" values for the chart, and we put user-facing flags there

the `debug` flag is not user-facing, so it shouldn't be there
similar to `.Values.manifests`

waodim: got it, I am also fine with not putting it to the `values` file 👍

hack/make/debug/debug.mk
```diff
@@ -0,0 +1,40 @@
...
+	DEBUG=true make deploy/helm
+
+## install and start telepresence in the cluster
+debug/telepresence/install:
+	telepresence helm install --create-namespace
+	telepresence connect -n dynatrace
+	telepresence intercept dynatrace-webhook -p 8443 --env-file local/telepresence.env
+
+## stop telepresence
+debug/telepresence/stop:
```
0sewa0, line 26: nitpick:
either call it `debug/telepresence/install` and `debug/telepresence/uninstall`
or `debug/telepresence/start` and `debug/telepresence/stop`

hack/make/debug/debug.mk
```diff
@@ -0,0 +1,40 @@
...
+	telepresence quit
+	telepresence helm uninstall
+
+## Forwards the CSI server ports to localhost:40000 and localhost:40001 (required for debugging the CSI driver)
+debug/tunnel:
+	kubectl port-forward -n dynatrace $$(kubectl get pod -n dynatrace -l app.kubernetes.io/component=csi-driver -o jsonpath='{.items[0].metadata.name}') 40000:40000 40001:40001 & echo $$! > /tmp/csi-driver-port-forward.pid
+
+## Cancels the port-forward
+debug/tunnel/stop:
```
0sewa0, line 35: why not `debug/tunnel/start`?

doc/debug.md
```diff
@@ -0,0 +1,183 @@
...
+
+### CSI-Driver-server
+
+- **Run CSI driver executables on the node** for file system operations.
+- **Makefile commands**:
+  - `make debug/build`: Build image with Delve included.
+  - `make debug/deploy`: Install image with necessary changes to deployments.
+  - `make debug/tunnel`: Open tunnel from local machine to CSI driver pod.
+- **IntelliJ setup**: Configure "Go Remote" with `localhost` and port `40000`, and set "On disconnect" to "Leave it running".
+- **VSCode:** Add a debug configuration to "Connect to Server" with `127.0.0.1` as the host and `40000` as the port, and set "remotePath" to `github.com/Dynatrace/dynatrace-operator`.
```
0sewa0, line 15: IMO: Having a "copy-pastable" json, is the most "TLDR"  😄 (you already kinda to this for the "none" TLDR sections)
```
        {
            "name": "Debug CSI server",
            "type": "go",
            "request": "attach",
            "mode": "remote",
            "remotePath": "github.com/Dynatrace/dynatrace-operator",
            "port": 40000,
            "host": "127.0.0.1"
        },
```

doc/debug.md
```diff
@@ -0,0 +1,183 @@
...
+            "envFile": "${workspaceFolder}/local/telepresence.env",
+            "args": [
+                "webhook-server",
+                "--certs-dir=./local/certs/"
+            ]
+        },
+        ```
+
+      - Set breakpoints and start debugging.
+    - When you are done, stop Telepresence with `telepresence quit`.
```
0sewa0, line 178: you have the `make debug/telepresence/stop`, why not reference it ?

0sewa0 (changes requested):
The structure of the document is a bit all over the places:
- I think the "Longer version" and `TLDR` version is just annoying to maintain. (2 things to maintain instead of 1)
- You describe multiple times what the `make` helpers do, sometimes with more, sometimes with less detail

To hopefully fix these, I would structure it like: (the name of the sections are up to you 😅 )
```
# Debugging the operator
...

## Makefile Helpers

### <1. helper make command>
<describe what it does, and maybe why>

### <2. helper make command>
...

## Debug Instructions 
### <component-1-to-debug>
- Context: <some description on what needs to be done>
- Setup-Steps: <list of commands to run>
- Config: <Steps for IntelliJ, just the json for VSCode>
- Teardown-Steps: <list of commands to run>

### <component-2-to-debug>
...
```

gkrenn:
Updated the `debug.md` as well as the `make` commands a little bit. Should be easier to read and use now.

chrismuellner (approved):

PR #3880, opened 2024-10-03
Remove proxy deprecation filter on OneAgent for argument and env var

albertogdd:
## Description

[K8S-11311](https://dt-rnd.atlassian.net/browse/K8S-11311)

When removing a proxy from our DynaKube, we need to add `--set-proxy=""` argument to OneAgent pods to trigger its proxy configuration deletion. This PR removes a logic that only adds this arg and env var on specific OneAgent versions and set these always.

## How can this be tested?

- Deploy DynaKube with proxy
- Check that proxy is configured in OA pod (run on pod shell `cat /mnt/volume_storage_mount/var_log/os/ruxitagent_* | grep Proxy`)
- Remove proxy from DynaKube
- Check that `--set-proxy=""` is set in OA manifest
- Check that proxy configuration is gone (run on pod shell `cat /mnt/volume_storage_mount/var_log/os/ruxitagent_* | grep Proxy`)

NOTE: While testing it was noticed that proxy config may sometimes not be getting properly set in pod and not shown in the shell command even when should be configured (maybe based on tenants or different setups?)


pkg/controllers/dynakube/oneagent/daemonset/arguments.go
```diff
@@ -14,23 +14,14 @@ const argumentPrefix = "--"
...
-	if err != nil {
-		return []string{}, err
-	}
-
-	if !isProxyAsEnvDeprecated {
-		// deprecated
-		b.appendProxyArg(argMap)
-	}
-
+	b.appendProxyArg(argMap)
```
chrismuellner, line 24: should we always set `--proxy=` even if proxy is set? That would mean it's configured via file and args at the same time for the oneagent.

albertogdd:
```golang
// deprecated
func (b *builder) appendProxyArg(argMap *prioritymap.Map) {
	if b.hasProxy() {
		argMap.Append(argumentPrefix+"set-proxy", "$(https_proxy)")
	}
	// if no proxy is set, we still have to set it as empty to clear proxy settings the OA might have cached
	argMap.Append(argumentPrefix+"set-proxy", "")
}
```
`--set-proxy=` to be correct. The information I got is that to definitely delete the container's configured proxy you have to unmount and also set the argument, if not, its deletion is not forced.

chrismuellner: do we need to enforce the deletion if there is already a different proxy set via file or could that cause issues? We disable proxy via args, but set it to a value via the file.

albertogdd: Pinging @StefanHauth , he knows more details

StefanHauth: `--set-proxy=""` should only be set if no proxy is configured. If a proxy is configured it should be set via file (on agent versions that support it. 
I don't know if updating the file content is automatically considered by the agent. I'll check that out.

StefanHauth: Update: `--set-proxy` always takes precedence. So we have to make sure that we either use `--set-proxy=""`or mount the files. If files are updated, OneAgent will pick up the changes and use the new proxy config on restart.

PR #3895, opened 2024-10-07
Move the ActiveGate structs to separate package

0sewa0:
[K8S-11245](https://dt-rnd.atlassian.net/browse/K8S-11245)

## Description
`ActiveGate` and related structs have been moved to their own package, as a first step (and proof of concept) to make the api packages more organized.

Look at commits for more details:
1. [Extract common structs to common package](https://github.com/Dynatrace/dynatrace-operator/commit/6d88759bcbd8cfdacf540208c9796edc51d6ac69)
2. [Remove dead code](https://github.com/Dynatrace/dynatrace-operator/commit/b12a3deebdb45ef13ed9c234f295a422d7ab4671)
3. [Update CRD generation](https://github.com/Dynatrace/dynatrace-operator/commit/b48c0181222d36a093e727c77d237b58b5f2e259)
4. [Move the ActiveGate structs to separate package](https://github.com/Dynatrace/dynatrace-operator/commit/85f9d3b4ffe2279e0c48511b430b343da5433f48)

Still TODO: Add tests for `pkg/api/v1beta3/dynakube/activegate/props.go`

## How can this be tested?

No change in actual behaviour, mainly look at the changes in the usage of the `activegate` package and `common` package, ask questions if something is hard to understand.


cmd/troubleshoot/dynakube_test.go
```diff
@@ -4,8 +4,10 @@ import (
 	"context"
 	"testing"
 
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/common"
```
andriisoldatenko, line 7: General Q: why is it common instead of a bunch of smaller packages, each with a different name and purpose?

0sewa0: We could do a bunch of small sub packages like:
- `"github.com/Dynatrace/dynatrace-operator/pkg/api/common/image"`
- `"github.com/Dynatrace/dynatrace-operator/pkg/api/common/connectioninfo"`
- ...

But it felt a bit overkill right now, because all the `structs` in it are very simple and small

0sewa0: +1

can do, do you have a better name than `common` for the folder (where these little packages will live)?

cmd/troubleshoot/dynakube_test.go
```diff
@@ -259,28 +261,28 @@ func (builder *testDynaKubeBuilder) withCustomPullSecret(secretName string) *tes
 }
 
 func (builder *testDynaKubeBuilder) withProxy(proxyURL string) *testDynaKubeBuilder {
-	builder.dynakube.Spec.Proxy = &dynakube.DynaKubeProxy{
+	builder.dynakube.Spec.Proxy = &common.ValueSource{
```
andriisoldatenko, line 264: here for example: common.ValueSource and I prefet to name it: `proxy` or something like it, prev it was clear 
Proxy = dynakube.DynakubeProxy

cmd/troubleshoot/dynakube_test.go
andriisoldatenko: let's think together:
- what do you do:  we create substructs for all APIs? or shared substructs?

andriisoldatenko: to be even more percise it's just `structs/` so either substructs or structs is good name IMO
`github.com/Dynatrace/dynatrace-operator/pkg/api/substructs` or `github.com/Dynatrace/dynatrace-operator/pkg/api/structs` or `github.com/Dynatrace/dynatrace-operator/pkg/api/sharedstructs`

0sewa0: we could call it `shared`

cmd/troubleshoot/dynakube_test.go
0sewa0: the struct itself has 0 relation to proxy :D 

`ValueSource` is the "old" `DynaKubeProxy` and `DynaKubeValueSource`, as these where the exact same

there is the `ProxySpec` that is 100% proxy related, and is used in EdgeConnect, and hopefully will be used in the DynaKube in the future

cmd/troubleshoot/dynakube_test.go
andriisoldatenko: good with me, it's more close what we do there.

cmd/troubleshoot/dynakube_test.go
0sewa0: `DynaKubeProxy` only exists for "historical reasons", which I would prefer not to keep. 😅

andriisoldatenko: okey, you can add a comment if you want. but we can resolve it.

pkg/api/v1beta3/dynakube/activegate/props.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (ag *Spec) IsMetricsIngestEnabled() bool {
+	return ag.IsMode(MetricsIngestCapability.DisplayName)
+}
+
+func (ag *Spec) NeedsService() bool {
+	return ag.IsRoutingEnabled() ||
+		ag.IsApiEnabled() ||
+		ag.IsMetricsIngestEnabled() ||
+		ag.enabledDependencies.extensions
```
waodim, line 86: should we abstract this also further?

pkg/api/v1alpha1/edgeconnect/convert_from_test.go
```diff
@@ -73,13 +74,13 @@ func getCurrentSpec() edgeconnect.EdgeConnectSpec {
 		KubernetesAutomation: &edgeconnect.KubernetesAutomationSpec{
 			Enabled: true,
 		},
-		Proxy: &common.ProxySpec{
+		Proxy: &proxy.Spec{
```
andriisoldatenko, line 77: @0sewa0 excellent work :)

andriisoldatenko:
there are still some uncommitted changes in your branch:
```
➜  dynatrace-operator git:(refactor/api-package-v2) ✗ gs
On branch refactor/api-package-v2
Your branch is up to date with 'origin/refactor/api-package-v2'.

  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   config/crd/bases/dynatrace.com_edgeconnects.yaml
```

pkg/api/v1beta3/dynakube/activegate/props.go
0sewa0: IMO, that would be overdoing it

waodim (approved):
to me it looks good - nothing further to add from my side.

andriisoldatenko (approved):

aorcholski (approved):
LGTM

PR #3897, opened 2024-10-07
Add EEC_DS_TOKEN env var to OTelC StatefulSet.

StefanHauth:
## Description

Add env var `EEC_DS_TOKEN` to OTelC StatefulSet.

## How can this be tested?

1. Deploy Dynakube with extensions
2. check that env var is set on OTelC pods
  ```
  EEC_DS_TOKEN: <set to the key 'otelc.token' in secret 'ext-dk-extensions-token'>  Optional: false 
  ```

### DIFF
--- a/pkg/controllers/dynakube/extension/otel/statefulset.go
+++ b/pkg/controllers/dynakube/extension/otel/statefulset.go
@@ -42,6 +42,7 @@ const (
 	envOTLPgrpcPort       = "OTLP_GRPC_PORT"
 	envOTLPhttpPort       = "OTLP_HTTP_PORT"
 	envOTLPtoken          = "OTLP_TOKEN"
+	envEECDStoken         = "EEC_DS_TOKEN"
 	envTrustedCAs         = "TRUSTED_CAS"
 	envK8sClusterName     = "K8S_CLUSTER_NAME"
 	envK8sClusterUuid     = "K8S_CLUSTER_UID"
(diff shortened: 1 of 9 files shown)


pkg/controllers/dynakube/extension/otel/statefulset.go
```diff
@@ -214,6 +215,13 @@ func buildContainerEnvs(dk *dynakube.DynaKube) []corev1.EnvVar {
 			},
 		},
 		},
+		{Name: envEECDStoken, ValueFrom: &corev1.EnvVarSource{
+			SecretKeyRef: &corev1.SecretKeySelector{
+				LocalObjectReference: corev1.LocalObjectReference{Name: dk.Name + consts.SecretSuffix},
```
luhi-DT, line 220: dk.Name + consts.SecretSuffix is often used, would it make sense to store it in a combined const instead?

luhi-DT (changes requested):

pkg/controllers/dynakube/extension/otel/statefulset.go
StefanHauth: I don't understand. How would you store `dk.Name` in const? It's a variable value. But I can see if it makes sense to put it into a function.

andriisoldatenko: What's @luhi-DT probably wants to create a method of DK instance/type to calculate this string?

luhi-DT: correct

luhi-DT: dk.Name + consts.SecretSuffix is used 3 times, you could create one consts and use it instead, so if the secretname changes you only need to change it in one place

andriisoldatenko (dismissed):

andriisoldatenko (commented):
practically LGTM

aorcholski (approved):
LGTM

andriisoldatenko (approved):

luhi-DT (approved):

PR #3906, opened 2024-10-08
Add node collector to DynaKube

waodim:
## Description

Adds the node collector config to the dynakube crd.

https://dt-rnd.atlassian.net/browse/K8S-11246

## How can this be tested?

Deploy a dynakube with `kspm` enabled and configure the collector with some test values.


pkg/api/v1beta3/dynakube/kspm.go
```diff
@@ -0,0 +1,57 @@
+package dynakube
+
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+type KspmSpec struct {
```
andriisoldatenko, line 8: what is Kspm? :)

pkg/api/v1beta3/dynakube/kspm.go
```diff
@@ -0,0 +1,57 @@
+package dynakube
```
0sewa0, line 1: should we have it in a separate package, similar to https://github.com/Dynatrace/dynatrace-operator/pull/3895?

pkg/api/v1beta3/dynakube/kspm.go
0sewa0: is it [Kubernetes Security Posture Management](https://www.sentinelone.com/blog/what-is-kspm/) ?

pkg/api/v1beta3/dynakube/dynakube_types.go
```diff
@@ -198,6 +202,8 @@ type TemplatesSpec struct {
 	// Low-level configuration options for the LogModule feature.
 	// +kubebuilder:validation:Optional
 	LogModule LogModuleTemplateSpec `json:"logModule,omitempty"`
+	// +kubebuilder:validation:Optional
+	KspmNodeConfigurationCollector KspmNodeConfigurationCollectorSpec `json:"kspmNodeConfigurationCollector,omitempty"`
```
0sewa0, line 206: nitpick: this name is bit wrong (IMO)

According to the epic:
> Deploy the K8s Node Configuration Collector Daemon Set with the Dynatrace Operator to enable Kubernetes Security Posture Management (KSPM) use cases.

So KSPM is the feature/usecase  -> so the `KspmSpec` makes sense, as it's the feature you turn off/on

But the **component** we deploy is the `K8s Node Configuration Collector` -> so in the `templates` section, we should name it after the **component** the user will configure, something like `K8sNodeConfigurationCollector`

wdyt?

cc @chrismuellner

0sewa0: Examples:
- We don't call the `OpenTelemetryCollector` the `ExtensionsOpenTelemetryCollector`, even though it only affects the `extensions` feature
- `LogModule` is fine, because we deploy the **component** called `LogModule`
- Same for  `ExtensionExecutionController`, as that is the name of the **component**

chrismuellner: I agree, but if we want this changes we need to discuss again with Product so we are all on the same page. These are customer facing changes you are proposing after all

0sewa0: outcome: leave it as is for now, will discuss it further with stakeholders, and rename if necessary

pkg/api/v1beta3/dynakube/kspm.go
waodim: Yes, I would assume so as well 👍 
I will add a comment

pkg/api/v1beta3/dynakube/kspm.go
waodim: valid point - let us merge your PR first and I will resolve

waodim: (hopefully correctly) done in https://github.com/Dynatrace/dynatrace-operator/pull/3906/commits/997c5369ec80a21be073ceb450105adf9053152f

0sewa0: 🧑‍🍳 💋

0sewa0 (approved):

aorcholski (changes requested):
please run `make doc/api-ref`  command
```
	modified:   doc/api/dynakube-api-ref.md
	modified:   doc/api/edgeconnect-api-ref.md
```

aorcholski (approved):
thx for py fix

0sewa0 (approved):

PR #3915, opened 2024-10-10
Add retryTimeout as a volume attribute

0sewa0:
## Description

Introduce a new volume-attribute for the CSI volumes, called `retryTimeout`.
- Not really used yet, just set on volumes and converted correctly

Introduce the `feature.dynatrace.com/max-csi-mount-timeout` feature-flag to configure the `retryTimeout` volume-attribute.
- This is meant to replace the `feature.dynatrace.com/max-csi-mount-attempts` feature-flag
   - This is why I introduced conversion logic for it
     - Due to the conversion logic I need to reintroduce `max-csi-mount-attempts` feature-flag logic to previous DynaKube versions

## How can this be tested?

Unittest are straight forward IMO.
Otherwise, 2 things can be tested:
1. Deploy an older version of the DynaKube, with `feature.dynatrace.com/max-csi-mount-attempts: <something not default>` configured, and check that the resulting latest DynaKube has the `feature.dynatrace.com/max-csi-mount-timeout: <something not default>`
2. Check that the injected app pods AND the deployed host agent pods have the `retryTimeout` volume-attribute
  - (they will only have it if they use the CSI driver 😅 )


pkg/api/v1beta3/dynakube/feature_flags.go
```diff
@@ -247,6 +251,36 @@ func (dk *DynaKube) FeatureMaxFailedCsiMountAttempts() int {
...
+	return duration
+}
+
+// MountAttemptsToTimeout converts the (old) number of csi mount attempts into a time.Duration string.
+// The converted value is based on the exponential backoff's algorithm.
+// The output is string because it's main purpose is to convert the value of an annotation to another annotation.
+func MountAttemptsToTimeout(maxAttempts int) string {
+	var power float64 = 2
+
+	var baseDelay = 1 * time.Second / 2
```
andriisoldatenko, line 271:
```suggestion
	var baseDelay = 500 * time.Millisecond
```

0sewa0: or `time.Second/2` 😛

andriisoldatenko: metally easy to read, since you don't need to calculate it.

pkg/api/v1beta3/dynakube/feature_flags.go
```diff
@@ -247,6 +251,36 @@ func (dk *DynaKube) FeatureMaxFailedCsiMountAttempts() int {
...
+
+	var delaySum time.Duration
+
+	for i := range maxAttempts {
+		secRetry := math.Pow(power, float64(i))
+		delay := time.Duration(secRetry) * baseDelay
+		delaySum += delay
```
andriisoldatenko, line 278: are you sure you need loop and can't rewrite using formula?

pkg/api/v1beta3/dynakube/feature_flags.go
0sewa0: IMO "half of a second" is easier to read than 500 Millisecond (my monkey brain doesn't like milliseconds 🐵 )

pkg/api/v1beta3/dynakube/feature_flags.go
0sewa0: looking at
```
	baseDelay := time.Second / 2
	delay := time.Duration(math.Exp2(float64(maxAttempts))) * baseDelay
```

seems to work

0sewa0: image

0sewa0: (fixed)

andriisoldatenko (approved):

andriisoldatenko (approved):

pkg/api/v1beta3/dynakube/feature_flags.go
waodim: I am confused, do we now need a loop for this calculation or not? You committed a solution without a loop, but reading this conversation I would assume we need a loop

andriisoldatenko: you don't need loop, as I originally suggested you can rewrite with formula, and @0sewa0 did it.

waodim (approved):
calculation lgtm

PR #3926, opened 2024-10-14
Skip OneAgent injection in case status is not ready 

gkrenn:
## Description

To prevent pods from getting stuck the webhook does no longer inject into user applications in the scenario when the state is unknown.

[Jira](https://dt-rnd.atlassian.net/browse/K8S-11367)

## How can this be tested?

- Apply CNFS DynaKube 
- Delete the `tenantUUID` from the status
  - Did not manage to delete it using `kubectl`. Instead I used `kubectl proxy` and `curl -XPATCH -H "Content-Type: application/merge-patch+json" \
--data '{"status": {"oneAgent": {"connectionInfoStatus": {"tenantUUID": null}}, "activeGate": {"connectionInfoStatus": {"tenantUUID": null}}}}' \
http://localhost:8001/apis/dynatrace.com/v1beta3/namespaces/dynatrace/dynakubes/dynakube/status`
- Create some pods that should be monitored
- Check if they were OneAgent injected and their annotations


pkg/webhook/mutation/pod/oneagent/mutator.go
```diff
@@ -112,6 +112,37 @@ func (mut *Mutator) ensureInitSecret(request *dtwebhook.MutationRequest) error {
...
+		reasons = append(reasons, EmptyTenantUUIDReason)
+	}
+
+	if !request.DynaKube.IsOneAgentCommunicationRouteClear() {
+		log.Info("OneAgent communication route is not clear, OneAgent cannot be injected", "pod", request.PodName())
+
+		reasons = append(reasons, EmptyConnectionInfoReason)
+	}
+
+	if status.CodeModules.Version == "" {
```
0sewa0, line 133: What happens in case of `codeModulesImage`?

0sewa0: if `codeModulesImage` is configured then the `status.CodeModules.Version` is empty and `status.CodeModules.ImageID` is set

0sewa0: also, you could use:
https://github.com/Dynatrace/dynatrace-operator/blob/a69eb0220442afcac5d06fbb9be1eb22ec787486/pkg/api/v1beta3/dynakube/oneagent_props.go#L145-L155

pkg/webhook/mutation/pod/oneagent/mutator.go
```diff
@@ -112,6 +112,37 @@ func (mut *Mutator) ensureInitSecret(request *dtwebhook.MutationRequest) error {
...
+		log.Info("code modules version is not available, OneAgent cannot be injected", "pod", request.PodName())
```
waodim, line 134:
```suggestion
		log.Info("codemodules version is not available, OneAgent cannot be injected", "pod", request.PodName())
```

waodim: minor nitpick because of consistency

Apart from that my only complaint would align with @0sewa0 comment about the codemodules case.

pkg/webhook/mutation/pod/oneagent/mutator.go
gkrenn: You are right, nice catch 👍

pkg/webhook/mutation/pod/oneagent/mutator.go
gkrenn: Updated

pkg/webhook/mutation/pod/oneagent/config.go
```diff
@@ -18,6 +18,10 @@ const (
 	releaseStageEnv        = "DT_RELEASE_STAGE"
 	releaseBuildVersionEnv = "DT_RELEASE_BUILD_VERSION"
 
+	EmptyConnectionInfoReason             = "EmptyConnectionInfo"
+	EmptyCodeModulesVersionAndImageReason = "EmptyCodeModulesVersionAndImage"
```
0sewa0, line 22:
```suggestion
	UnknownCodeModuleReason = "UnknownCodeModuleReason"
```

pkg/webhook/mutation/pod/oneagent/mutator.go
```diff
@@ -112,6 +112,37 @@ func (mut *Mutator) ensureInitSecret(request *dtwebhook.MutationRequest) error {
...
+	}
+
+	if !dk.IsOneAgentCommunicationRouteClear() {
+		log.Info("OneAgent communication route is not clear, OneAgent cannot be injected", "pod", request.PodName())
+
+		reasons = append(reasons, EmptyConnectionInfoReason)
+	}
+
+	if dk.CodeModulesVersion() == "" && dk.CodeModulesImage() == "" {
+		log.Info("codemodules version and image are not available, OneAgent cannot be injected", "pod", request.PodName())
```
0sewa0, line 134:
```suggestion
		log.Info("information about the codemodules (version or image) is not available, OneAgent cannot be injected", "pod", request.PodName())
```

0sewa0 (commented):
logic looks good, just some naming suggestions

waodim (commented):
Apart from @0sewa0 s naming nitpicks, code lgtm - tested it and it works for me. will approve as soon as everything is resolved.

pkg/webhook/mutation/pod/oneagent/mutator.go
gkrenn: Done

0sewa0 (approved):

waodim (approved):

PR #3929, opened 2024-10-14
WIP Improves error logging for Dynatrace Client

aorcholski:


pkg/clients/dynatrace/server_error.go
```diff
@@ -0,0 +1,42 @@
...
+	ParameterLocation string `json:"parameterLocation"`
+	Path              string `json:"path"`
+}
+
+// ServerError represents an error returned from the server (e.g. authentication failure).
+type ServerError struct {
+	Message              string                `json:"message"`
+	ConstraintViolations []ConstraintViolation `json:"constraintViolations,omitempty"`
+	Code                 int                   `json:"code"`
+}
```
0sewa0, line 25: is this correct?
[the docs](https://developer.dynatrace.com/plan/platform-services/error-handling/) show that it should be in a `details` section:
```
{
  "error": {
    "code": 400,
    "message": "Constraints violated.",
    "details": {
      "constraintViolations": [
        {
          "path": "detectionRules[0].filterConfig.pattern",
          "message": "may not be null",
          "parameterLocation": "PAYLOAD_BODY"
        }
      ]
    }
  }
}
```

the docs also say "These are commonly used fields in the `details` object", which reads to me that it is not always there, could that be a problem?

pkg/clients/dynatrace/dynatrace_client_test.go
```diff
@@ -267,3 +281,121 @@ func createTestDynatraceClientWithFunc(t *testing.T, handler http.HandlerFunc) (
 
 	return dynatraceServer, dynatraceClient
 }
+
+func testServerErrors(t *testing.T) {
```
0sewa0, line 285: these tests feel a bit overcomplicated, you are basically testing the fact that `json.Unmarshal(response, &se)` works for a `struct` that you created via `json.Marshal(&message)`, with a mocked webserver in between 😅 

Valid tests for this IMO would be:
1. Have a valid `json` object from the API (with the error you want), and try to `json.Unmarshal` that, to check that your `struct` matches the `json` object
2.  Unittests for the method `func (e ServerError) Error() string `,

0sewa0 (changes requested):

pkg/clients/dynatrace/server_error.go
chrismuellner: let's make sure to not break handling of error responses!

@aorcholski  please check with cluster teams for the best way to handle errors if we are unsure of edge cases.

pkg/clients/dynatrace/dynatrace_client_test.go
aorcholski: right, raw strings are returned by mocked server

pkg/clients/dynatrace/server_error.go
0sewa0: did a double check, and IMO [the docs](https://developer.dynatrace.com/plan/platform-services/error-handling/) are just wrong or not meant for the tenant-apis, their is never a `details` section in the error response, even the examples in the swagger UI don't have them

0sewa0 (approved):

albertogdd (approved):

PR #3940, opened 2024-10-16
Add symlink that points to the latest codemodule version

waodim:
## Description

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-1361)

**Attention** 
This PR will not be merged into `main` but into `epic/csi-nodb` - our feature branch for the csi no db epic.

With this a symlink is added at the location: `/data/<dk>/latest-codemodule` that points to the codemodule directory (`/data/codemodules/<base64encodedImage/version>`) in the `provisioner` container.

It is maintained in a way that the symlink is always removed and immediatly after created again if:

- agent is being newly installed
- csi driver pod starts but agent is already installed from previous runs

This is done in order to update the codemodule directory and with that also the symlink and to avoid errors because of a mismatch between symlink and codemodule dir (happened to me! I updated the `codeModulesImage` -> the directory got "updated" accordingly, but the symlink still pointed to the old directory). 

I added no unit tests as I do not think the things I introduced are properly unit testable, the other method in `symlink.go` is also not unit tested and `installAgent` also isn't.

## How can this be tested?

Deploy operator with `codeModulesImage` set.

shell into the `provisioner` container and run:

`bash-5.1# ls -lR /data | grep '^l'`

The result should contain this line which indicates that there is a symlink from latest-codemodule to the dir: 

`lrwxrwxrwx 1 root root  114 Oct 16 11:26 latest-codemodule -> /data/codemodules/cHVibGljLmVjci5hd3MvZHluYXRyYWNlL2R5bmF0cmFjZS1jb2RlbW9kdWxlczoxLjMwMS40OC4yMDI0MTAwNy0xMDMzMzY=`

Update the codeModulesImage and when the agent is installed run the command again. The symlink should be updated accordingly.

You can also cd into `latest-codemodule` and look if the linking works. You should see the same directory as under the codemodules directory. You can move into the `/agent/bin/` dir and check the version there.

### DIFF
--- a/pkg/injection/codemodule/installer/symlink/symlink.go
+++ b/pkg/injection/codemodule/installer/symlink/symlink.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"regexp"
 
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
 	"github.com/pkg/errors"
 	"github.com/spf13/afero"
 )
@@ -44,6 +45,40 @@ func CreateSymlinkForCurrentVersionIfNotExists(fs afero.Fs, targetDir string) er
 		log.Info("symlinking failed", "version", relativeSymlinkPath)
 
 		return errors.WithStack(err)
+	}
+
+	return nil
+}
+
+func CreateSymlinkForLatestVersion(fs afero.Fs, targetDir string, dk dynakube.DynaKube, symLinkPath, imageDigest string) error {
+	// MemMapFs (used for testing) doesn't comply with the Linker interface
+	linker, ok := fs.(afero.Linker)
+	if !ok {
+		log.Info("symlinking not possible", "targetDir", targetDir, "fs", fs)
+
+		return nil
+	}
+
+	relativeSymlinkPath := "/data/codemodules/" + imageDigest
+
+	log.Info("creating symlink", "points-to(relative)", relativeSymlinkPath, "location", symLinkPath)
+
+	if err := linker.SymlinkIfPossible(relativeSymlinkPath, symLinkPath); err != nil {
+		log.Info("symlinking failed", "codemodules-version", relativeSymlinkPath)
+
+		return errors.WithStack(err)
+	}
+
+	return nil
+}
+
+func RemoveSymLink(fs afero.Fs, symLinkPath string) error {
+	if info, _ := fs.Stat(symLinkPath); info != nil {
+		log.Info("symlink to codemodule directory exists, removing it due to the possibility of the agent being installed again")
+
+		if err := fs.RemoveAll(symLinkPath); err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/pkg/injection/codemodule/installer/symlink/symlink.go
+++ b/pkg/injection/codemodule/installer/symlink/symlink.go
@@ -10,7 +10,7 @@ import (
 	"github.com/spf13/afero"
 )
 
-func CreateSymlinkForCurrentVersionIfNotExists(fs afero.Fs, targetDir string) error {
+func CreateForCurrentVersionIfNotExists(fs afero.Fs, targetDir string) error {
 	var relativeSymlinkPath string
 
 	var err error
@@ -50,7 +50,7 @@ func CreateSymlinkForCurrentVersionIfNotExists(fs afero.Fs, targetDir string) er
 	return nil
 }
 
-func CreateSymlinkForLatestVersion(fs afero.Fs, dk dynakube.DynaKube, latestVersionDir, symlinkDir string) error {
+func CreateForLatestVersion(fs afero.Fs, dk dynakube.DynaKube, latestVersionDir, symlinkDir string) error {
 	// MemMapFs (used for testing) doesn't comply with the Linker interface
 	linker, ok := fs.(afero.Linker)
 	if !ok {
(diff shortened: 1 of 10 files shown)


pkg/injection/codemodule/installer/symlink/symlink.go
```diff
@@ -49,6 +50,38 @@ func CreateSymlinkForCurrentVersionIfNotExists(fs afero.Fs, targetDir string) er
...
+	if err := linker.SymlinkIfPossible(latestVersionDir, symlinkDir); err != nil {
+		log.Info("symlinking failed", "codemodules-version", latestVersionDir)
+
+		return errors.WithStack(err)
+	}
+
+	return nil
+}
+
+func RemoveSymLink(fs afero.Fs, symLinkPath string) error {
```
andriisoldatenko, line 73: I think it should be better to use `symlink` instead of `SymLink`, even it's acronym, but after quick look at go source code I see they also follow this standard https://go.dev/src/path/filepath/symlink.go

0sewa0: @andriisoldatenko you mean?
```suggestion
func RemoveSymlink(fs afero.Fs, symlinkPath string) error {
```

andriisoldatenko: yes, in our code base we use only lowercase, I just checked, so it's important to stay in touch
```
pkg/injection/codemodule/installer/image/installer.go
13:     "github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer/symlink"
88:     if err := symlink.CreateSymlinkForCurrentVersionIfNotExists(installer.fs, targetDir); err != nil {
91:             log.Info("failed to create symlink for agent installation", "err", err)

pkg/injection/codemodule/installer/url/installer.go
13:     "github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer/symlink"
80:     if err := symlink.CreateSymlinkForCurrentVersionIfNotExists(installer.fs, targetDir); err != nil {
82:             log.Info("failed to create symlink for agent installation", "targetDir", targetDir)

pkg/injection/codemodule/installer/symlink/symlink_test.go
1:package symlink

pkg/injection/codemodule/installer/symlink/config.go
1:package symlink
14:     log = logd.Get().WithName("oneagent-symlink")

pkg/injection/codemodule/installer/symlink/symlink.go
1:package symlink
22:             log.Info("symlinking not possible", "targetDir", targetDir, "fs", fs)
34:     symlinkTargetPath := filepath.Join(targetBindDir, "current")
35:     if fileInfo, _ := fs.Stat(symlinkTargetPath); fileInfo != nil {
36:             log.Info("symlink already exists", "location", symlinkTargetPath)
41:     log.Info("creating symlink", "points-to(relative)", relativeSymlinkPath, "location", symlinkTargetPath)
43:     if err := linker.SymlinkIfPossible(relativeSymlinkPath, symlinkTargetPath); err != nil {
44:             log.Info("symlinking failed", "version", relativeSymlinkPath)
```

0sewa0: ok, then your original comment had a typo 😛

andriisoldatenko: you mean `Symlink` ? yes the idea is to use as whole word instead of SymLink or symLink etc.

0sewa0: I like it 😘  , but the original message confused me a bit 😄

andriisoldatenko: it's Friday :) all good

waodim: will do

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/3940/commits/8c1329c701ea50ed9e22fedbe62b8a4c7df182bb

pkg/injection/codemodule/installer/symlink/symlink.go
```diff
@@ -49,6 +50,38 @@ func CreateSymlinkForCurrentVersionIfNotExists(fs afero.Fs, targetDir string) er
...
+	if err := linker.SymlinkIfPossible(latestVersionDir, symlinkDir); err != nil {
+		log.Info("symlinking failed", "codemodules-version", latestVersionDir)
+
+		return errors.WithStack(err)
+	}
+
+	return nil
+}
+
+func RemoveSymlink(fs afero.Fs, symlinkPath string) error {
```
andriisoldatenko, line 73: since it's already `symlink` package, i think we can simplify for `symlink.Remove`

pkg/injection/codemodule/installer/symlink/symlink.go
```diff
@@ -49,6 +50,38 @@ func CreateSymlinkForCurrentVersionIfNotExists(fs afero.Fs, targetDir string) er
 	return nil
 }
 
+func CreateSymlinkForLatestVersion(fs afero.Fs, dk dynakube.DynaKube, latestVersionDir, symlinkDir string) error {
```
andriisoldatenko, line 53:
```suggestion
func CreateForLatestVersionDir(fs afero.Fs, dk dynakube.DynaKube, latestVersionDir, symlinkDir string) error {
```

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/3940/commits/ee269cbac29dd6e645b309cc11b88a7e35d0dcfe

pkg/injection/codemodule/installer/symlink/symlink.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/3940/commits/ee269cbac29dd6e645b309cc11b88a7e35d0dcfe

I did it for all symlink exported methods

0sewa0 (approved):

andriisoldatenko (approved):

PR #3941, opened 2024-10-16
Use the secret.Query in the AuthToken reconciler 

0sewa0:
## Description
[DAQ-1456](https://dt-rnd.atlassian.net/browse/DAQ-1456)

Not much has changed, as this secret is rather unique in the codebase, so only small adjustments made sense,

## How can this be tested?

no big noticeable (biggest one would be that the Auth-token secret now has an ownerReference)

### DIFF
--- a/pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
+++ b/pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
@@ -10,7 +10,7 @@ import (
 	dtclient "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/conditions"
-	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/secret"
+	k8ssecret "github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/secret"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


waodim (dismissed):
code lgtm - works for me and there is an ownerreference

waodim (dismissed):

pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
```diff
@@ -128,7 +131,7 @@ func (r *Reconciler) createSecret(ctx context.Context, secretData map[string][]b
 		return errors.WithStack(err)
 	}
 
-	err = r.client.Create(ctx, secret)
+	err = r.query().WithOwner(r.dk).Create(ctx, secret)
```
andriisoldatenko, line 134: TBH, i don't get why `r.query()` is better then `r.cleint`? Could you please clarify?

pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
```diff
@@ -162,3 +165,7 @@ func (r *Reconciler) conditionSetSecretCreated(secret *corev1.Secret) {
 
 	setAuthSecretCreated(r.dk.Conditions(), ActiveGateAuthTokenSecretConditionType, "secret created "+days+" day(s) ago, token:"+tokenPublicPart)
 }
+
+func (r *Reconciler) query() k8ssecret.QueryObject {
```
andriisoldatenko, line 169: Technically, it's only about `querySecret` and not more generic `query`.

pkg/util/kubeobjects/secret/query.go
```diff
@@ -9,24 +9,30 @@ import (
...
-func Query(kubeClient client.Client, kubeReader client.Reader, log logd.Logger) query.Generic[*corev1.Secret, *corev1.SecretList] {
-	return query.Generic[*corev1.Secret, *corev1.SecretList]{
-		Target:     &corev1.Secret{},
-		ListTarget: &corev1.SecretList{},
-		ToList: func(sl *corev1.SecretList) []*corev1.Secret {
-			out := []*corev1.Secret{}
-			for _, s := range sl.Items {
-				out = append(out, &s)
-			}
+type QueryObject struct {
```
andriisoldatenko, line 12: this one I like :)

pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
0sewa0: It was started out as "refactor it more", but because this `AuthToken` is a bit special (we never update, just delete and recreate), so I couldn't use the nicer parts of the `query`
- only the `Delete` is a tiny bit better
- and now we have an owner-reference 

it is mainly for consistency, we should use our abstraction everywhere, so we handle the secrets in a consistent way
- Example: Have a owner-reference on it
- Maybe future example: Have the hash-annotation on it

pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
0sewa0: true true, but I wanted something short, I would call it `secretQuery`?

andriisoldatenko (commented):
code LGTM

andriisoldatenko (approved):

PR #3943, opened 2024-10-17
Delete extensions self-signed TLS Secret when `spec.extensions.enabled: false`

albertogdd:
## Description

[DAQ-1680](https://dt-rnd.atlassian.net/browse/K8S-11326)

Extensions self-signed TLS Secret was getting reconciled when `spec.extensions.enabled: false`, which is wrong.

Now we check `r.dk.IsExtensionsEnabled()`, like in the rest of extension reconcilers.

I also renamed the functions for clarity.

## How can this be tested?

- I added a new unit test with this case. It fails if run on main, and passes on this branch :heavy_check_mark: 
- To test manually, apply dynakube with `spec.extensions.enabled: false` and check that the secret `<namespace>-extensions-controller-tls` doesn't exist.

### DIFF
--- a/pkg/controllers/dynakube/extension/tls/reconciler.go
+++ b/pkg/controllers/dynakube/extension/tls/reconciler.go
@@ -36,14 +36,14 @@ func NewReconciler(clt client.Client, apiReader client.Reader, dk *dynakube.Dyna
 }
 
 func (r *reconciler) Reconcile(ctx context.Context) error {
-	if r.dk.ExtensionsNeedsSelfSignedTLS() {
-		return r.reconcileSelfSigned(ctx)
+	if !r.dk.IsExtensionsEnabled() || !r.dk.ExtensionsNeedsSelfSignedTLS() {
+		return r.deleteSelfSignedTLSSecret(ctx)
 	}
 
-	return r.reconcileTLSRefName(ctx)
+	return r.reconcileSelfSignedTLSSecret(ctx)
 }
 
-func (r *reconciler) reconcileSelfSigned(ctx context.Context) error {
+func (r *reconciler) reconcileSelfSignedTLSSecret(ctx context.Context) error {
 	query := k8ssecret.Query(r.client, r.client, log)
 
 	_, err := query.Get(ctx, types.NamespacedName{
@@ -58,7 +58,7 @@ func (r *reconciler) reconcileSelfSigned(ctx context.Context) error {
 	return err
 }
 
-func (r *reconciler) reconcileTLSRefName(ctx context.Context) error {
+func (r *reconciler) deleteSelfSignedTLSSecret(ctx context.Context) error {
 	query := k8ssecret.Query(r.client, r.client, log)
 
 	return query.Delete(ctx, &corev1.Secret{
--- a/pkg/controllers/dynakube/extension/tls/reconciler_test.go
+++ b/pkg/controllers/dynakube/extension/tls/reconciler_test.go
@@ -97,6 +97,23 @@ func TestReconcile(t *testing.T) {
 		require.True(t, k8serrors.IsNotFound(err))
 		require.Empty(t, secret)
 	})
+	t.Run("self-signed tls secret is deleted if spec.extensions is not enabled", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/controllers/dynakube/extension/tls/reconciler.go
```diff
@@ -36,14 +36,14 @@ func NewReconciler(clt client.Client, apiReader client.Reader, dk *dynakube.Dyna
 }
 
 func (r *reconciler) Reconcile(ctx context.Context) error {
-	if r.dk.ExtensionsNeedsSelfSignedTLS() {
-		return r.reconcileSelfSigned(ctx)
+	if !r.dk.IsExtensionsEnabled() || !r.dk.ExtensionsNeedsSelfSignedTLS() {
```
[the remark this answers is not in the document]
albertogdd: Good idea :+1: 

FYI @StefanHauth is it OK to add a new DynaKube condition about the extensions TLS secret? This wasn't written on the feature ticket but I could add it on this PR quickly.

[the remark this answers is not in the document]
StefanHauth: Yes, IMHO it's useful. It also quickly let's us see if the secret has been created succesfully.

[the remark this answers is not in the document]
albertogdd: [Fixed here](https://github.com/Dynatrace/dynatrace-operator/pull/3943/commits/0dc01eb3857fbc433402cdcda6a06189ef50918f)

pkg/controllers/dynakube/extension/tls/reconciler.go
```diff
@@ -52,13 +59,18 @@ func (r *reconciler) reconcileSelfSigned(ctx context.Context) error {
 	})
 
 	if err != nil && k8serrors.IsNotFound(err) {
-		return r.createSelfSignedTLSSecret(ctx)
+		err = r.createSelfSignedTLSSecret(ctx)
+		if err != nil {
+			return err
+		}
+
+		conditions.SetSecretCreated(r.dk.Conditions(), extensionsSelfSignedTLSSecretConditionType, getSelfSignedTLSSecretName(r.dk.Name))
```
StefanHauth, line 67: Shouldn't  the condition also be set accordingly if creating the certificate failed for whatever reason?

StefanHauth (changes requested):

pkg/controllers/dynakube/extension/tls/reconciler.go
albertogdd: True :+1: My brain is still coming from vacations. Im updating it

pkg/controllers/dynakube/extension/tls/conditions.go
```diff
@@ -0,0 +1,3 @@
+package tls
+
+const extensionsSelfSignedTLSSecretConditionType string = "ExtensionsSelfSignedTLSSecret"
```
andriisoldatenko, line 3: I think it would be better to have one condition type for both secrets . @StefanHauth wdyt?

StefanHauth: But then we can't see the status of each secret individually, right?

andriisoldatenko: Do you really need it? For me it sounds overkill, high level secret for both sounds good compromise.

andriisoldatenko: Because if we have two types, @albertogdd  should use both to check, and it is a bit painful and, again, overkill.

albertogdd: When you say both secrets, you mean the self-signed TLS and the TLSRef one?
IMO the TLSRef one is created by the clients, so we shouldn't have a condition about it. That's why I just added one condition just for the self-signed TLS secret.

albertogdd: Resolution: I will create a small ticket about the conditions (1 condition for the 2 secrets) and in this PR I will revert some changes and just fix the bug

PR #3944, opened 2024-10-17
Add node collector token secret

luhi-DT:
## Description
https://dt-rnd.atlassian.net/browse/DAQ-1402

Token for communication between AG and nodeConfigCollector is reconciled and created. 

## How can this be tested?

Deploy the operator and a Dynakube with `kspm.enabled: true`, check if the token is created
Later check if the dynakube is changed to  `kspm.enabled: false`, that the secret is removed


pkg/controllers/dynakube/kspm/token.go
```diff
@@ -0,0 +1,85 @@
...
+	err = query.Delete(ctx, secret)
+
+	if err != nil {
+		log.Info("could not delete kspm token", "name", secret.Name)
+
+		return err
+	}
+
+	return nil
+}
```
0sewa0, line 65: could introduce a condition, so you only try to delete when there is a condition

also `query.Delete(ctx, secret)` is safe for `k8serrors.IsNotFound(err)`, you would only need to create a secret object with valid name and namespace for it to work

pkg/controllers/dynakube/kspm/token.go
```diff
@@ -0,0 +1,85 @@
+package kspm
+
+import (
+	"context"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/kspm/consts"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/dttoken"
+	k8ssecret "github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/secret"
+	"k8s.io/api/core/v1"
```
0sewa0, line 10:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

pkg/controllers/dynakube/kspm/token.go
```diff
@@ -0,0 +1,85 @@
...
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func ensureKSPMSecret(ctx context.Context, client client.Client, apiReader client.Reader, dk *dynakube.DynaKube) error {
+	kspmSecretName := dk.Name + "-" + consts.KSPMSecretKey
```
0sewa0, line 17: we will need this in the future, to mount it, should have a function either in this package called `GetSecretName` or as a method for the `DynaKube`

pkg/controllers/dynakube/kspm/reconciler.go
```diff
@@ -0,0 +1,41 @@
...
+	dk        *dynakube.DynaKube
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context) error {
+	if r.dk.IsKSPMEnabled() {
+		return ensureKSPMSecret(ctx, r.client, r.apiReader, r.dk)
+	} else {
+		return removeKSPMSecret(ctx, r.client, r.apiReader, r.dk)
+	}
+}
```
0sewa0, line 41: There should be a `kspm` reconciler, but there will be other sub-reconcilers here as well, not just the token

the structure should be that the `kspm` reconciler just calls the other sub-reconcilers.
you could have a `pkg/controllers/dynakube/kspm/token` package, where this code is, and you just call that `token` reconciler in the `kspm` reconciler

0sewa0: example: https://github.com/Dynatrace/dynatrace-operator/blob/acade6bbb33c5a0f6444c1230d5091d7521e0700/pkg/controllers/dynakube/logmodule/reconciler.go#L28-L62

pkg/controllers/dynakube/kspm/token.go
0sewa0: example: https://github.com/Dynatrace/dynatrace-operator/blob/acade6bbb33c5a0f6444c1230d5091d7521e0700/pkg/controllers/dynakube/logmodule/configsecret/reconciler.go#L46-L61

0sewa0 (changes requested):

pkg/controllers/dynakube/kspm/reconciler.go
luhi-DT: done

pkg/controllers/dynakube/kspm/token.go
luhi-DT: done

pkg/controllers/dynakube/kspm/token.go
luhi-DT: done

pkg/controllers/dynakube/kspm/token.go
luhi-DT: done, thx for the input ❤️

0sewa0 (approved):
tested it

when `kspm.enabled: true`:
1. Secret gets created 
2. Condition is present

when `kspm.enabled: false`
1. Secret gets removed
2. Condition is removed

🍪

waodim (approved):
can confirm that it works, both secret and condition are handled correctly.

andriisoldatenko (approved):
code LGTM

PR #3947, opened 2024-10-18
Fix proxy content logged in init-container logs

luhi-DT:
## Description
https://dt-rnd.atlassian.net/browse/DAQ-1677

When using a proxy with auths, the content of username and password is logged in the initContainer logs

## How can this be tested?

1. Deploy Dynatrace the Operator
2. Create Dynakube with a proxy which contains username+password
4. Apply Dynakube
5. Restart monitored application
6. Check the oneagent-init container logs and see that both username and password is logged


0sewa0 (approved):

waodim (approved):
works 👍

PR #3950, opened 2024-10-18
Do not reconcile extensions self-signed TLS secret when `spec.extensions.enabled: false`

albertogdd:
## Description

[DAQ-1680](https://dt-rnd.atlassian.net/browse/DAQ-1680)

Extensions self-signed TLS Secret was getting reconciled when `spec.extensions.enabled: false`, which is wrong.

Now we check `r.dk.IsExtensionsEnabled()`, like in the rest of extension reconcilers.

I also renamed the functions for clarity.

## How can this be tested?

- I added a new unit test with this case. It fails to execute it on main, and passes on this branch :heavy_check_mark: 
- To test manually, apply dynakube with `spec.extensions.enabled: false` and check that the secret `<namespace>-extensions-controller-tls` doesn't exist.

[DAQ-1680]: https://dt-rnd.atlassian.net/browse/DAQ-1680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controllers/dynakube/extension/tls/reconciler.go
+++ b/pkg/controllers/dynakube/extension/tls/reconciler.go
@@ -36,11 +36,11 @@ func NewReconciler(clt client.Client, apiReader client.Reader, dk *dynakube.Dyna
 }
 
 func (r *reconciler) Reconcile(ctx context.Context) error {
-	if !r.dk.IsExtensionsEnabled() || !r.dk.ExtensionsNeedsSelfSignedTLS() {
-		return r.deleteSelfSignedTLSSecret(ctx)
+	if r.dk.IsExtensionsEnabled() && r.dk.ExtensionsNeedsSelfSignedTLS() {
+		return r.reconcileSelfSignedTLSSecret(ctx)
 	}
 
-	return r.reconcileSelfSignedTLSSecret(ctx)
+	return r.deleteSelfSignedTLSSecret(ctx)
 }
 
 func (r *reconciler) reconcileSelfSignedTLSSecret(ctx context.Context) error {
--- a/pkg/controllers/dynakube/extension/tls/reconciler.go
+++ b/pkg/controllers/dynakube/extension/tls/reconciler.go
@@ -36,14 +36,14 @@ func NewReconciler(clt client.Client, apiReader client.Reader, dk *dynakube.Dyna
 }
 
 func (r *reconciler) Reconcile(ctx context.Context) error {
-	if r.dk.ExtensionsNeedsSelfSignedTLS() {
-		return r.reconcileSelfSigned(ctx)
+	if !r.dk.IsExtensionsEnabled() || !r.dk.ExtensionsNeedsSelfSignedTLS() {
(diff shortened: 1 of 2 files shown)


pkg/controllers/dynakube/extension/tls/reconciler_test.go
```diff
@@ -55,13 +56,14 @@ func TestReconcile(t *testing.T) {
...
 
-		reconciler.Reconcile(context.Background())
+		err := reconciler.Reconcile(context.Background())
+		require.NoError(t, err)
 
 		var secret corev1.Secret
-		err := fakeClient.Get(context.Background(), SelfSignedTLSSecretObjectKey, &secret)
+		fakeClient.Get(context.Background(), SelfSignedTLSSecretObjectKey, &secret)
 
 		require.NoError(t, err)
```
andriisoldatenko, line 65: if you don;t return error on line 61, why we need to check error here? 🤔

albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/3950/commits/ad38cb1a1bdfe0821d57f20acd42c9e18c03f49d), missed it when reverting other changes

aorcholski (approved):
LGTM, Secret is created when needed and removed if extensions are disabled.

pkg/controllers/dynakube/extension/tls/reconciler.go
```diff
@@ -36,14 +36,14 @@ func NewReconciler(clt client.Client, apiReader client.Reader, dk *dynakube.Dyna
 }
 
 func (r *reconciler) Reconcile(ctx context.Context) error {
-	if r.dk.ExtensionsNeedsSelfSignedTLS() {
-		return r.reconcileSelfSigned(ctx)
+	if !r.dk.IsExtensionsEnabled() || !r.dk.ExtensionsNeedsSelfSignedTLS() {
```
andriisoldatenko, line 39: if I remember correctly (based on De Morgan's Thereom) we can refactor it using:

```
	if r.dk.IsExtensionsEnabled() && r.dk.ExtensionsNeedsSelfSignedTLS() {
		return r.reconcileSelfSignedTLSSecret(ctx)
	}

	return r.deleteSelfSignedTLSSecret(ctx)
}
```
I tested and it seems it works

albertogdd: It will be refactored next week anyways with the condition, so I wouldn't worry too much

albertogdd: [Done](https://github.com/Dynatrace/dynatrace-operator/pull/3950/commits/4d7f36c21f4399238e40cf0fa34e99e71cc422de)

andriisoldatenko (approved):

PR #3953, opened 2024-10-21
Implement conditions for EC config secret

andriisoldatenko:
## Description

This PR introduce new conditon type and project reconciliation reason of it.

ticket -> [DAQ-1388](https://dt-rnd.atlassian.net/browse/DAQ-1388)

## How can this be tested?

`make test/e2e/edgeconnect` or unit-test or manually via creating Edgeconnect CR and check condition type and status.

[DAQ-1388]: https://dt-rnd.atlassian.net/browse/DAQ-1388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controllers/edgeconnect/conditions.go
```diff
@@ -0,0 +1,4 @@
+package edgeconnect
+
+// SecretConfigConditionType identifies the secret config condition.
+const SecretConfigConditionType = "SecretConfigConditionType"
```
0sewa0, line 4: why is this not in `pkg/controllers/edgeconnect/consts/consts.go`?

IMO, the structure of the `edgeconnect` package is a bit messy(or just doesn't follow what we do elsewhere),because there is only 1 BIG reconciler, that handles everything

andriisoldatenko: Yeah, I'm aware of it, and we will refactor it soon to sync with the dynakube structure.

pkg/controllers/edgeconnect/consts/consts.go
```diff
@@ -23,4 +22,7 @@ const (
 	KeyEdgeConnectId                = "id"
 
 	AnnotationEdgeConnectContainerAppArmor = "container.apparmor.security.beta.kubernetes.io/" + EdgeConnectContainerName
+
+	// consts.SecretConfigConditionType identifies the secret config condition.
+	SecretConfigConditionType = "consts.SecretConfigConditionType"
```
0sewa0, line 27: the `consts.` in the type doesn't really serve any purpose IMO

pkg/controllers/edgeconnect/controller.go
```diff
@@ -810,19 +817,31 @@ func (controller *Controller) createOrUpdateEdgeConnectConfigSecret(ctx context.
...
 	if err != nil {
+		conditions.SetSecretGenFailed(ec.Conditions(), consts.SecretConfigConditionType, err)
+
 		return "", "", errors.WithStack(err)
 	}
 
 	query := k8ssecret.Query(controller.client, controller.apiReader, log)
 
-	_, err = query.CreateOrUpdate(ctx, secretConfig)
+	created, err := query.CreateOrUpdate(ctx, secretConfig)
```
0sewa0, line 827: the returned `bool` will be `true` on both the update and create scenario, it is `false` IF there was no change or an error happened

pkg/controllers/edgeconnect/controller_test.go
```diff
@@ -219,6 +223,87 @@ func TestReconcile(t *testing.T) {
...
+
+		err = controller.apiReader.Get(context.TODO(), client.ObjectKey{Name: ec.Name, Namespace: ec.Namespace}, ec)
+		require.NoError(t, err)
+		require.NotEmpty(t, ec.Conditions())
+
+		condition := meta.FindStatusCondition(*ec.Conditions(), consts.SecretConfigConditionType)
+		assert.Equal(t, metav1.ConditionFalse, condition.Status)
+		assert.Equal(t, conditions.SecretGenerationFailed, condition.Reason)
+		assert.Equal(t, "Failed to generate secret: get secret error", condition.Message)
+	})
```
0sewa0, line 306: Isn't this what you want?
```suggestion
	t.Run("SecretConfigConditionType is set SecretGenFailed", func(t *testing.T) {
		ec := &edgeconnect.EdgeConnect{
			ObjectMeta: metav1.ObjectMeta{
				Name:      testName,
				Namespace: testNamespace,
			},
			Spec: edgeconnect.EdgeConnectSpec{
				ApiServer: "abc12345.dynatrace.com",
				OAuth: edgeconnect.OAuthSpec{
					Endpoint:     "https://test.com/sso/oauth2/token",
					Resource:     "urn:dtenvironment:test12345",
					ClientSecret: testOauthClientSecret,
					Provisioner:  false,
				},
			},
		}

		controller := createFakeClientAndReconciler(t, ec,
			createKubeSystemNamespace(),
		)

		err := controller.reconcileEdgeConnectRegular(context.Background(), ec)
		require.Error(t, err)
		require.NotEmpty(t, ec.Conditions())

		condition := meta.FindStatusCondition(*ec.Conditions(), consts.SecretConfigConditionType)
		assert.Equal(t, metav1.ConditionFalse, condition.Status)
		assert.Equal(t, conditions.SecretGenerationFailed, condition.Reason)
		assert.Contains(t, condition.Message, "Failed to generate secret: failed to get clientSecret from client-secret secret")
	})
```

Reading the test is super confusing:
1. Here you configure a `controller` with an `apiReader` that will have the client-secret
```
		controller := createFakeClientAndReconciler(t, ec,
			createClientSecret(testOauthClientSecret, ec.Namespace),
			createKubeSystemNamespace(),
		)
```
2. Here you override the `apiReader`, to always die in case someone wants to get an error, and `return nil` other wise which is incorrect
```
		fakeClient := fake.NewClientWithInterceptors(interceptor.Funcs{...})
		controller.apiReader = fakeClient
```
- You made the previous call`createFakeClientAndReconciler` basically useless, as the whole purpose of the func is to setup the `apiReader` mock
- When overwriting a `Get` with an interceptor, you should make sure that all other cases still work (ie.: that is realistic)
   - Your `Get` interceptor will do impossible things when getting an edgeconnect resource for example (you do this in the next step)
     - What is impossible? : Doing nothing(ie.: not touching the `obj client.Object`) and returning `nil` error as a result of a `Get`. This will never-ever happen when using the actual `client`
3. Here you get the edgeconnect from the API, even though the function you are testing will not update the edgeconnect CR in the API, but only locally
```
		err = controller.apiReader.Get(context.TODO(), client.ObjectKey{Name: ec.Name, Namespace: ec.Namespace}, ec)
		require.NoError(t, err)
		require.NotEmpty(t, ec.Conditions())
```
- But why does the test pass?
  - Because of the weird `interceptor.Funcs` you defined for the `Get` action will do nothing, therefore the `ec` object here is actually left untouched, so the local changes done in the `reconcileEdgeConnectRegular` func will be present

0sewa0: What my suggestion basically does:
1. In the `createFakeClientAndReconciler` call, I leave out the config-secret, so it wont be present, so the logic will fail, where you want it to fail
2. Removed the `interceptor.Funcs` part, as it is unnecessary
3. Removed the `get` call for the `ec` as the function we are testing modifies the object, and doesn't update it on the API
4. Not checking the exact error message as that is annoying (used `assert.Contains`)
5. Used `context.Background` instead of TODO

0sewa0: I am pretty sure you could simplify the other tests for this function, as it reads very much like copy paste

0sewa0 (changes requested):

pkg/controllers/edgeconnect/consts/consts.go
andriisoldatenko: lol it was by mistake :)

pkg/controllers/edgeconnect/controller.go
andriisoldatenko: fixed in 746c065f3

StefanHauth (approved):

pkg/controllers/edgeconnect/controller_test.go
```diff
@@ -219,6 +222,111 @@ func TestReconcile(t *testing.T) {
...
+			Spec: edgeconnect.EdgeConnectSpec{
+				ApiServer: "abc12345.dynatrace.com",
+				OAuth: edgeconnect.OAuthSpec{
+					Endpoint:     "https://test.com/sso/oauth2/token",
+					Resource:     "urn:dtenvironment:test12345",
+					ClientSecret: testOauthClientSecret,
+					Provisioner:  false,
+				},
+			},
+		}
```
albertogdd, line 308: We could move this to a function to not repeat it in the code, WDYT?

andriisoldatenko: Repeating in tests is okey. Anyway we need big refactoring of EC. So we will refactor tests as well.

PR #3954, opened 2024-10-21
Change LogModule to LogMonitoring

gkrenn:
## Description

The requirements for `LogModule` have changed: 

- `LogModule` should be called `LogMonitoring`
- `spec.logModule.enabled: true` ==> `spec.logMonitoring: {}`
- `spec.logModule.enabled: false` ==> `spec.logMonitoring: nil`

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-1400)

## How can this be tested?

- Turn `LogMonitoring` on as described above => the DaemonSet should be created


pkg/api/v1beta3/dynakube/logmonitoring_props.go
```diff
@@ -0,0 +1,18 @@
...
+		Spec: dk.Spec.LogMonitoring,
+	}
+	lm.SetName(dk.Name)
+
+	return lm
+}
+
+func (dk *DynaKube) LogMonitoringTemplates() logmonitoring.TemplateSpec {
+	return dk.Spec.Templates.LogMonitoring
+}
```
0sewa0, line 18: do it more like 

in `pkg/api/v1beta3/dynakube/logmonitoring/spec.go`
```
type LogMonitoring struct {
	*Spec
	*Templates

	name string
}
```

in `pkg/api/v1beta3/dynakube/logmonitoring_props.go`

```
func (dk *DynaKube) LogMonitoring() *logmonitoring.LogMonitoring {
	lm := &logmonitoring.LogMonitoring{
		Spec: dk.Spec.LogMonitoring,
		Templates: dk.Spec.Templates.LogMonitoring
	}
	lm.SetName(dk.Name)

	return lm
}
```

So you can do the following `dk.LogMonitoring().DNSPolicy` and such.
IMO, it is much nicer than `dk. LogMonitoringTemplates().DNSPolicy`
Or you can also do `dk.LogMonitoring().Templates.DNSPolicy` if there is naming conflict

0sewa0 (changes requested):
Looks great, left a small suggestion

pkg/api/v1beta3/dynakube/logmonitoring_props.go
0sewa0: if there ever be a `logmonitoring.Status` you would also put that into the `type LogMonitoring struct`, so we truly handle it as their own CR

gkrenn: 👍  - Done

pkg/api/v1beta3/dynakube/logmonitoring/props.go
```diff
@@ -0,0 +1,21 @@
...
+
+const (
+	logModuleDaemonSetSuffix = "-logmodule"
+)
+
+func (logMonitoring *Spec) SetName(name string) {
+	logMonitoring.name = name
+}
+
+func (logMonitoring *Spec) Needed() bool {
```
waodim:
```suggestion
func (logMonitoring *Spec) IsNeeded() bool {
```

waodim: Don't have the strongest opinion for this one - I just think it fits the rest of our codebase better. You could also go with `Enabled()` but this is just my opinion, maybe you and/or the others are fine with `Needed()`

waodim (commented):
only one nitpick, otherwise lgtm

waodim (approved):
code lgtm - logmonitoring ds starts as expected.

0sewa0 (approved):
Tested it works, small leftover https://github.com/Dynatrace/dynatrace-operator/blob/caf163e3ac7ad427bac6191c2d019a7ff657cddc/pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go#L60 (due to change that happened after this PR was published)

PR #3955, opened 2024-10-22
More excessively check if the oneagent injection happened

0sewa0:
## Description

**Problem**: The `ActiveGate` deployed by the Operator can't be self-monitored. (ie.: injected by the same Operator)

**Reason**:
- We set the DT_DEPLOYMENT_METADATA env in the ActiveGate StatefulSet , 
- AND also set the DT_DEPLOYMENT_METADATA env as part of the oneAgent-injection, 
- AND we use this DT_DEPLOYMENT_METADATA envvar to “know” if a container was already injected or not so we do not inject it again, and we do not add it to the CONTAINER_INFO multiple times

So this will result in that: the AG container is NEVER added to the CONTAINER_INFO because it is “already injected”, yet it is not.

## How can this be tested?

Deploy an ActiveGate via the Operator and inject it with the same Operator.
- You will have to mess with the `feature.dynatrace.com/ignored-namespaces` feature-flag
- You may have to restart the ActiveGate to get it injected (or use multiple dynakubes)


pkg/util/kubeobjects/volumes/volumes.go
```diff
@@ -16,3 +16,13 @@ func GetVolumeMountByName(mounts []corev1.VolumeMount, volumeName string) (*core
 		volumeName, len(mounts),
 	)
 }
+
+func IsIn(mounts []corev1.VolumeMount, volumeName string) bool {
```
luhi-DT, line 20: `containsVolume`

PR #3956, opened 2024-10-22
Fix dynakube always being set to error phase

luhi-DT:
## Description
https://dt-rnd.atlassian.net/browse/DAQ-1842

Currently the Dynakube is always set to error phase if there is no kubernetesMEID (e.g because no automatic api monitoring) since the logModule Reconciler always returned an error in this case

## How can this be tested?

Create a cloudNative Dynakube without `kubernetes-monitoring` and without `automatic-api-monitoring` -> check if the phase is correctly set to Runninng


0sewa0 (approved):

waodim (approved):
lgtm

PR #3957, opened 2024-10-22
Remove semver validator for custom OneAgent version

albertogdd:
## Description

[DAQ-1817](https://dt-rnd.atlassian.net/browse/DAQ-1817)

Custom OneAgent version on the DynaKube spec (e.g. `spec.oneAgent.cloudNativeFullStack.version`) doesn't work at the moment. We validate the CR expecting a semver (1.0.0) but the OA DaemonSet fails to start with a semver (with error: version malformed).

The quick fix for it is to remove the semver validator for the CR and then the field is usable again, only with timestamped versions, like 1.200.0.20240101-000000.

## How can this be tested?

- Apply a DynaKube with the following version, everything should work fine:
```yaml
spec:
  oneAgent:
    cloudNativeFullStack:
```

[DAQ-1817]: https://dt-rnd.atlassian.net/browse/DAQ-1817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


luhi-DT:
Removing the validator is a workaround imho, the real problem is here: https://github.com/Dynatrace/dynatrace-operator/blob/8d5a86fc8d2beb36f825f5d911be0ba5b7658652/pkg/version/semantic.go#L49 Because this line expects a 5 segment version, have you tried changing this code?

albertogdd:
> Removing the validator is a workaround imho, the real problem is here:
> 
> https://github.com/Dynatrace/dynatrace-operator/blob/8d5a86fc8d2beb36f825f5d911be0ba5b7658652/pkg/version/semantic.go#L49
> 
> Because this line expects a 5 segment version, have you tried changing this code?

Just tried it, I get the same error `version malformed`

0sewa0:
Have you tried to use this: https://github.com/Dynatrace/dynatrace-operator/blob/45444304346eb08ca88aa3f3b61404936074ea0c/pkg/util/dtversion/dtversion.go#L26

luhi-DT (approved):

waodim (approved):
tested it, works 👍

PR #3966, opened 2024-10-23
Add OneAgent version validator

albertogdd:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-1817

Add validator for OA version. Now it works only for timestamped versions.

We removed it on a previous PR but got the requirement to have it in place for 1.3.2.

## How can this be tested?

Only DynaKubes with versions like this should pass the CR validation when applying:
```yaml
spec:
  oneAgent:
    cloudNativeFullStack:
```

### DIFF
--- a/pkg/api/validation/dynakube/oneagent.go
+++ b/pkg/api/validation/dynakube/oneagent.go
@@ -29,7 +29,7 @@ Use a nodeSelector to avoid this conflict. Conflicting DynaKubes: %s`
 
 	versionRegex = `^\d+.\d+.\d+.\d{8}-\d{6}$`
 
-	versionInvalidMessage = "Only versions in the form of major.minor.patch.timestamp (e.g. 1.0.0.20240101-000000) are allowed"
+	versionInvalidMessage = "The OneAgent's version is only valid in the format 'major.minor.patch.timestamp', e.g. 1.0.0.20240101-000000"
 )
 
 func conflictingOneAgentConfiguration(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
--- a/pkg/api/validation/dynakube/oneagent.go
+++ b/pkg/api/validation/dynakube/oneagent.go
@@ -3,9 +3,11 @@ package validation
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/dtversion"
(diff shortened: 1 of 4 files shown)


pkg/api/validation/dynakube/oneagent_test.go
```diff
@@ -477,3 +477,49 @@ func createDynakubeWithHostGroup(args []string, hostGroup string) *dynakube.Dyna
...
+		Spec: dynakube.DynaKubeSpec{
+			APIURL: testApiUrl,
+			OneAgent: dynakube.OneAgentSpec{
+				ClassicFullStack: &dynakube.HostInjectSpec{},
+			},
+		},
+	}
+
+	t.Run("valid OneAgent custom versions are allowed", func(t *testing.T) {
+		versions := []string{"", "1.0.0.20240101-000000"}
```
chrismuellner, line 493: please add a test for a short version, e.g. `1.0.0`, as well!

albertogdd: Short versions are invalid. They fail on OA DaemonSet level (with `version malformed`), that's why this validator rejects them too.

albertogdd: https://github.com/Dynatrace/dynatrace-operator/pull/3966/files#diff-101115bf2e0d1f10e1c5efea14cf1d120ff0300df964b0883577c6fbbe84f25eR503

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -24,6 +26,10 @@ Use a nodeSelector to avoid this conflict. Conflicting DynaKubes: %s`
 	warningOneAgentInstallerEnvVars = `The environment variables ONEAGENT_INSTALLER_SCRIPT_URL and ONEAGENT_INSTALLER_TOKEN are only relevant for an unsupported image type. Please ensure you are using a supported image.`
 
 	warningHostGroupConflict = `The DynaKube specification sets the host group using the --set-host-group parameter. Instead, specify the new spec.oneagent.hostGroup field. If both settings are used, the new field takes precedence over the parameter.`
+
+	versionRegex = `^\d+.\d+.\d+.\d{8}-\d{6}$`
+
+	versionInvalidMessage = "Only versions in the form of major.minor.patch.timestamp (e.g. 1.0.0.20240101-000000) are allowed"
```
chrismuellner, line 32:
```suggestion
	versionInvalidMessage = "The OneAgent's version is only valid in the format 'major.minor.patch.timestamp', e.g. 1.0.0.20240101-000000"
```

from the message it's not clear what `version` is the problem.

luhi-DT (approved):
tested, works as expected

aorcholski (approved):
LGTM

PR #3967, opened 2024-10-23
[release-1.3] Add OneAgent version validator

albertogdd:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-1817

Same as https://github.com/Dynatrace/dynatrace-operator/pull/3966

But adding the validator also for beta1 (because different apiVersion validators are only merged on main)

## How can this be tested?

Only DynaKubes with versions like this should pass the CR validation when applying:
```yaml
spec:
  oneAgent:
    cloudNativeFullStack:
```
:warning:  Test on v1beta1 and v1beta2

### DIFF
--- a/pkg/api/v1beta1/dynakube/validation/oneagent.go
+++ b/pkg/api/v1beta1/dynakube/validation/oneagent.go
@@ -3,11 +3,11 @@ package validation
 import (
 	"context"
 	"fmt"
-	"strings"
+	"regexp"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta1/dynakube" //nolint:staticcheck
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/dtversion"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/env"
-	"golang.org/x/mod/semver"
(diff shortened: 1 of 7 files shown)


luhi-DT (approved):

pkg/api/v1beta1/dynakube/validation/oneagent.go
```diff
@@ -143,15 +147,20 @@ func conflictingHostGroupSettings(_ context.Context, _ *Validator, dk *dynakube.
...
 		return ""
 	}
 
-	version := "v" + agentVersion
-	if !(semver.IsValid(version) && semver.Prerelease(version) == "" && semver.Build(version) == "" && len(strings.Split(version, ".")) == 3) {
-		return "Only semantic versions in the form of major.minor.patch (e.g. 1.0.0) are allowed!"
+	_, err := dtversion.ToSemver(agentVersion)
+	if err != nil {
+		return versionInvalidMessage
+	}
```
0sewa0, line 159: is this necessary? if you have the regexp?

albertogdd: In there we do semver.isValid() so its useful to check this too.

0sewa0 (approved):

PR #3968, opened 2024-10-23
Redeploy ActiveGate incase the KSPM token changes

0sewa0:
`SetHash` is an inferior version of `hasher.AddAnnotation` as it only works for StatefulSets and is not nil safe

## Description

In the unfortunate case when the `<dk>-kspm-token` secret's content changes, we want to redeploy the ActiveGate, to make sure it doesn't get out of sync with components that will also us `<dk>-kspm-token`.
- As the `<dk>-kspm-token` is only create once (or when it's deleted by the user), this is just a failsafe.

See commits/commit-description for more details

## How can this be tested?

1. Deploy kspm:
```
kind: DynaKube
metadata:
  name: dk
  namespace: dynatrace
spec:
  kspm:
    enabled: true
  activeGate:
    capabilities:
      - kubernetes-monitoring
```
2. Delete `<dk>-kspm-token` Secret
3. See that the ActiveGate gets restarted.


pkg/controllers/dynakube/kspm/token/reconciler.go
```diff
@@ -55,6 +56,13 @@ func ensureKSPMSecret(ctx context.Context, client client.Client, apiReader clien
 			return err
 		}
 
+		tokenHas, err := hasher.GenerateHash(secretConfig.Data)
```
luhi-DT, line 59:
```suggestion
		tokenHash, err := hasher.GenerateHash(secretConfig.Data)
```

waodim (approved):
code lgtm and it also worked for me

andriisoldatenko (approved):

albertogdd (approved):
It works

PR #3983, opened 2024-10-24
Move hash generation and adding to generic query implementation

waodim:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-1391

With this the generation and adding of the `template-hash` annotation for objects is added to the generic query implementation. The hash is always generated and set when the object is created or updated.

**NOTE:** I removed the previous way of setting the annotation where it is not needed anymore because I moved it into the query implementation. If you find places where I missed something to remove please let me know.

## How can this be tested?

Deploy the operator and have a look if everything runs, you can also check if the `kspm` secret for example has a `template-hash` annotation set. Other example you might want to check: 

- edgeconnect secret
- logmonitoring config secret
- activeGate auth token
- ...


pkg/controllers/dynakube/activegate/internal/statefulset/reconciler_test.go
```diff
@@ -230,9 +230,6 @@ func TestManageStatefulSet(t *testing.T) {
 
 		require.NoError(t, err)
 
-		err = r.manageStatefulSet(ctx)
-		require.NoError(t, err)
-
 		actualStatefulSet, err = statefulset.Query(r.client, r.apiReader, log).Get(ctx, client.ObjectKeyFromObject(desiredStatefulSet))
```
0sewa0, line 233: removing it, to make the tests pass, will not make the test (and therefore, the change) correct 😅

pkg/util/kubeobjects/internal/query/query.go
```diff
@@ -63,6 +69,11 @@ func (c Generic[T, L]) Update(ctx context.Context, object T) error {
 		}
 	}
 
+	err := hasher.AddAnnotation(object)
+	if err != nil {
+		return errors.WithStack(err)
+	}
```
0sewa0, line 75: some hints:

in the case you call `CreateOrUpdate`, will you ever reach this?
- you will not
- that is why the tests you updated failed

why?
- does the "newly generated" object has a hash, that will be used check equality in `CreateOrUpdate`?

would it be safe to use `hasher.AddAnnotation(object)` multiple times in a row?
- yes, if done correctly

0sewa0 (changes requested):
The implementation is not correct.

No tests should fail after the change, except specific test for the `query` clients (like, 
1 test `pkg/util/kubeobjects/statefulset/query_test.go`)

You are changing the behaviour of how the `query` works, and not how the StatefulSet `reconciler` works, so the tests of the `query` failing and needing update is expected, however `reconciler`'s test failing is a sign that something is not working properly.

pkg/util/kubeobjects/internal/query/query.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/3983/commits/5ec6fddaef04bf927b108bf4ff40b3843ffc2a47 finally found a fix (with lots of your help, thx). Please also read my description. I removed a test that I added before. If you think this test should be readded (in a version that works) please let me know.

0sewa0 (approved):

pkg/util/hasher/hasher.go
```diff
@@ -65,13 +65,15 @@ func AddAnnotation(object metav1.Object) error {
...
-	objectHash, err := GenerateHash(object)
-	if err != nil {
-		return err
+	if _, ok := annotations[AnnotationHash]; !ok {
+		objectHash, err := GenerateHash(object)
+		if err != nil {
+			return err
+		}
+
+		annotations[AnnotationHash] = objectHash
```
albertogdd, line 74: As far as I see in my cluster, this is setting the annotation on the STS metadata level, but it should be on the STS.ObjectMeta.Annotations level in order to work well. (work well = pods get restarted if the annotation changes)

albertogdd:
Also, extensions make use of hasher.GenerateHash, so I believe it also requires removing some lines over there.
https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/controllers/dynakube/extension/eec/statefulset.go#L106
https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/controllers/dynakube/extension/otel/statefulset.go#L98

pkg/util/hasher/hasher.go
waodim: so you tested it and the pods that get created by the sts do not get annotated?

andriisoldatenko:
late for the party, I run e2e tests for edgeconnect and it worked.
```
SS
out: helm uninstall dynatrace-operator \
                        --namespace dynatrace
release "dynatrace-operator" uninstalled

ok      github.com/Dynatrace/dynatrace-operator/test/scenarios/standard 1074.653s
```

PR #3991, opened 2024-10-25
Align ProcessModuleConfig secret reconciler with other reconcilers

gkrenn:
## Description

This PR refactors the process module config secret reconciler to use a Query object. 
It as well aligns the structure with other reconcilers and adds a cleanup for the secret created.

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-1455)

## How can this be tested?

- Create a CloudNativeFullStack DynaKube and check if the pmc secret exists.
- Check the conditions in the DynaKube for pmc as well


pkg/controllers/dynakube/processmoduleconfigsecret/reconciler.go
```diff
@@ -30,120 +30,95 @@ type Reconciler struct {
 	dtClient     dtclient.Client
 	dk           *dynakube.DynaKube
 	timeProvider *timeprovider.Provider
+	secretQuery  k8ssecret.QueryObject
```
albertogdd, line 33: Do we want to start adding query types into reconciler structs? I'm fine with it, just asking because we don't do it in other places.

pkg/controllers/dynakube/processmoduleconfigsecret/reconciler.go
```diff
@@ -30,120 +30,95 @@ type Reconciler struct {
...
 		}
+
+		return nil
 	}
 
-	if err := r.ensureSecret(ctx); err != nil {
-		return errors.WithMessage(err, "could not update secret")
+	err := r.reconcileSecret(ctx)
+	if err != nil {
+		return errors.WithMessage(err, "failed to create processModuleConfig secret")
```
waodim, line 77:
```suggestion
		return errors.WithMessage(err, "failed to create or update processModuleConfig secret")
```

waodim: just think that this is more detailed. if you your original message is enough, feel free to keep it.

pkg/controllers/dynakube/processmoduleconfigsecret/reconciler.go
```diff
@@ -30,120 +30,95 @@ type Reconciler struct {
...
-
-		return r.createSecret(ctx)
-	} else if err != nil {
+func (r *Reconciler) createOrUpdateSecret(ctx context.Context, secret *corev1.Secret) error {
+	_, err := r.secretQuery.WithOwner(r.dk).CreateOrUpdate(ctx, secret)
+	if err != nil {
 		conditions.SetKubeApiError(r.dk.Conditions(), pmcConditionType, err)
 
-		return err
+		return errors.Errorf("failed to create secret '%s': %v", secret.Name, err)
```
waodim, line 104:
```suggestion
		return errors.Errorf("failed to create or update secret '%s': %v", secret.Name, err)
```

pkg/controllers/dynakube/processmoduleconfigsecret/reconciler.go
waodim: just my 2 cents: I wouldn't mind adding it as well, I would just like to keep it consistent and add the query to the struct on all other locations.

gkrenn: Changed it back - Will create a small ticket to edit it everywhere, don't want to do it in this PR though.

pkg/controllers/dynakube/processmoduleconfigsecret/reconciler.go
gkrenn: Done

waodim (approved):
lgtm and works

albertogdd (approved):

PR #4001, opened 2024-10-29
Update go version, livenessprobe and CSI node driver registrar.

StefanHauth:
## Description

Manual dependency update to mitigate some vulnerabilities in v1.2.3

## How can this be tested?

* Just run the operator and see if normal operation is fine.


aorcholski (approved):
LGTM

PR #4004, opened 2024-10-29
Add default PersistentVolumeClaim to EEC statefulset

aorcholski:
[DAQ-1662](https://dt-rnd.atlassian.net/browse/DAQ-1662)

## Description

CRD changes:
- removed `spec.templates.extensionExecutionController.persistentVolumeClaimRetentionPolicy`
- added `spec.templates.extensionExecutionController.useEphemeralVolume`

EEC statefulset changes:
- PVC retention policy is set to `{ DELETE, DELETE }` and can't be changed
- dynakube validator doesn't allow to use `useEphemeralVolume` and `persistentVolumeClaim` at the same time
- the default PVC is added to EEC statefulset when required

`StatefulSet.PersistentVolumeClaim` field can't be updated so EEC statefulset is automatically recreated.

`spec.templates.extensionExecutionController.persistentVolumeClaim` spec is not validated. 

## How can this be tested?
1) unittests
```
make go/test
```
2) e2e tests
```
make test/e2e/extensions
```
3) deploy dynakube with EEC enabled and different PVC settings
```
spec:
  templates:
    extensionExecutionController:
      useEphemeralVolume: ( removed | true | false )
      persistentVolumeClaim: ( removed | {} |
        accesModes:
        - ReadWrite
        resources:
          requests:
            storage: 2Gi )
```      

[DAQ-1662]: https://dt-rnd.atlassian.net/browse/DAQ-1662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


0sewa0 (commented):
related to:
> Dynakube CR has to be deleted and re-created to make a change in templates.extensionExecutionController.useEphemeralVolume or templates.extensionExecutionController.persistentVolumeClaim field.

that is kinda bad UX, you could update this: (to also consider if the PersistentVolume related fields are different or not)
https://github.com/Dynatrace/dynatrace-operator/blob/a22de1b1756719a4051a61042851f2b028e16b0e/pkg/util/kubeobjects/statefulset/query.go#L37-L39

and then we will magically recreate it 😉

pkg/api/validation/dynakube/eec.go
```diff
@@ -6,7 +6,8 @@ import (
 )
 
 const (
-	errorExtensionExecutionControllerImageNotSpecified = `DynaKube's specification enables the Prometheus feature, make sure you correctly specify the ExtensionExecutionController image.`
+	errorExtensionExecutionControllerImageNotSpecified       = `DynaKube's specification enables the Prometheus feature, make sure you correctly specify the ExtensionExecutionController image.`
+	errorExtensionExecutionControllerInvalidPVCConfiguration = "You provided a PVC and demanded to use an ephemeral volume at the same time for extension controller. These settings are mutually exclusive, please choose only one."
```
0sewa0, line 10: The message sounds kinda weird:
```suggestion
	errorExtensionExecutionControllerInvalidPVCConfiguration = `DynaKube specifies a PVC for the extension controller while ephemeral volume is also enabled. These settings are mutually exclusive, please choose only one.`
```

pkg/api/validation/dynakube/eec.go
```diff
@@ -22,3 +23,17 @@ func extensionControllerImage(_ context.Context, _ *Validator, dk *dynakube.Dyna
 
 	return ""
 }
+
+func extensionControllerPVCStorageDevice(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
+	if !dk.IsExtensionsEnabled() {
+		return ""
+	}
+
+	if dk.Spec.Templates.ExtensionExecutionController.UseEphemeralVolume != nil && *dk.Spec.Templates.ExtensionExecutionController.UseEphemeralVolume && dk.Spec.Templates.ExtensionExecutionController.PersistentVolumeClaim != nil {
```
0sewa0, line 32: kind hard to read, could you make it more readable ? (like put it into a function?)

pkg/controllers/dynakube/extension/eec/statefulset.go
```diff
@@ -420,19 +422,39 @@ func setVolumes(dk *dynakube.DynaKube) func(o *appsv1.StatefulSet) {
...
+				Resources: corev1.VolumeResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceStorage: resource.MustParse("1Gi"),
+					},
+				},
+			}
+
+			if !persistentvolumeclaimspec.IsEmpty(dk.Spec.Templates.ExtensionExecutionController.PersistentVolumeClaim) {
+				pVCSpec = *dk.Spec.Templates.ExtensionExecutionController.PersistentVolumeClaim
+			}
```
0sewa0, line 439: I wouldn't do anything like this, just `if not nil` just set it, let the customer be dumb

0sewa0 (changes requested):

pkg/controllers/dynakube/extension/eec/statefulset.go
aorcholski: the idea is to use empty PersistentVolumeClaimSpec{} as kind of useDefaultPVCSpec flag (`{}` as `true`). This method of configuration is in line with the following [changes](https://github.com/Dynatrace/dynatrace-operator/pull/4012)

StefanHauth: It has been agreed with PA & PM, that we always set a PVC (default or custom) unless `useEphemeralStorage` is explicitly set to true. Not having a PVC could lead to data loss which is not acceptable.

chrismuellner: Noone is questioning your requirements, but `persistentvolumeclaimspec.IsEmpty` is checking each individual field, which could be simplified.

`...PersistentVolumeClaim != nil`

aorcholski: `PersistentVolumeClaim != nil` condition doesn't allow to distinguish :
1) dynakube.spec.PersistentVolumeClaimCpec{} 
2) dynakube.spec.PersistentVolumeClaimCpec{a=1} 

The default PVC spec  should be used in the 1) case only

0sewa0: that is IMO confusing:

Because both `pvc == nil` and `pvc == {}` will result in the same thing (if no ephemeral storage should be used)
Only difference is that the validation will kick in for `pvc == {}`

StefanHauth: I think the confusion is because this code suggests that ephemeral storage takes precedence. In fact, it doesn't because at this point it is already made sure that either ephemeral storage or a PVC is configured, never both.

And `IsEmpty` returns true if either `pvc == nil` or `pvc == {}`. If `pvc != nil && pvc != {}` it's considered as not empty and configured as provided in Dynakube.

One question that comes to my mind: some of the PVCS fields are mandatory. Do we need to check that those are provided correctly (in the validation webhook probably)?

0sewa0: > And IsEmpty returns true if either pvc == nil or pvc == {}

This makes having the `nil` value meaningless

We use the `nil` value for fields to determine if they are set or not, we use the defaults if the field is not set.
- This is to make our life easier and less error-prone 

Having 2 states that mean "field is not set" is just redundant and confusing.

Checking for `pvc == {}` would make sense if `useEphemeralStorage` was not a thing. (I believe I mentioned this in the meeting)
- The scenarios would look like:
    -  `pvc == nil` => use ephemeral storage
    - `pvc == {}` => use default PVC
    - `pvc == {...}` => use what is configured
- however I am not in favour of this due to how annoying is to check `pvc == {}` properly.

>One question that comes to my mind: some of the PVCS fields are mandatory. Do we need to check that those are provided correctly (in the validation webhook probably)?

Keep in mind the following: We shouldn't try to reimplement the validation logic of the Kubernetes API, it's not something we want on our plate, and we will probably miss some edge-cases.
- We should validate for things that are in our control
   - Example-1: DynaKube name length, we are the ones that will add certain postfixes to the name of the dynakube, this is logic specific to us, we should let the user know
   - Example-2: This very PR implements a validation that makes sense, having both `useEphemeralStorage: true` and a `pvc` configured is incorrect according to our logic, we should let the user know.

StefanHauth: If we use `pvc == nil` as `useEphemeralStorage,` it can happen that setting a PVC is accidentally forgotten, which can lead to data loss. That's an absolute no-go. Ephemeral storage should only be used in environments where data loss is acceptable (testing, dev). And it has to be a conscious decision to do so. This should not happen just because someone forgot to add it.

0sewa0: > however I am not in favour of this due to how annoying is to check pvc == {} properly.
I am not for `pvc == nil` to be `useEphemeralStorage: true`

What I'm for is the following:
- `useEphemeralStorage: true` + `pvc == nil` ===> use an ephemeral storage (as pvc was not configured)
- `useEphemeralStorage: true` + `pvc == {...}` ===> incorrect config (`pvc == {...}` can be `pvc == {}`)
- `useEphemeralStorage: false` + `pvc == nil` ===> use default pvc
- `useEphemeralStorage: false` + `pvc == {...}` ===> use the pvc that the user has defined, whatever it might be (correct or not)
  - we shouldn't validate what `pvc` the user provided, it is on the user to provide something that is correct
     - Why? We can't/shouldn't validate everything, we don't validate if the `nodeSelector` will result in successful deployments, same for the `tolerations` or for the `namespaceSelector`, and I could go on.

StefanHauth: Ok, then I misunderstood that. In your propsal `useEphemeralStorage: false` + `pvc: {}` would prevent EEC from starting because of an incorrect StatefulSet (because missing mandatory fields in the PVC).

chrismuellner: > useEphemeralStorage: false + pvc: {} would prevent EEC from starting because of an incorrect StatefulSet

which is exactly what we want. current `IsEmpty` validation is _trying_ to reimplement this validation, but only checks if _any_ field is set, which could still lead to an invalid resource. If the user has configured a PVC we should let Kubernetes handle any potential issues and not reimplement part of the validation ourselves.

0sewa0: > would prevent EEC from starting because of an incorrect StatefulSet
but that is fine, there are other things that could go wrong, that we can't anticipate like "is the persistent-volume available" and stuff like that. 

Why I am saying that we can't/shouldn't validate it, is because it is not fully in our control, like every single `image` related field, if something incorrect is put there then there will be a `ImagePullBackoff`.

What I would focus on, is to handle the error from the API nicely. (when our StatefulSet gets rejected)
- Make it visible in the status of the DynaKube
- Make sure that the `phase` of the DynaKube will be `Error` if such things happen
- etc...

And if we still want to do validation for `pvc == {}` then `pvc == nil` has no purpose, because I can do the same list of "expected behaviour" without the `pvc == nil`:
- `useEphemeralStorage: true` + `pvc == {}` ===> use an ephemeral storage
- `useEphemeralStorage: true` + `pvc == {...}` ===> incorrect config
- `useEphemeralStorage: false` + `pvc == {}` ===> use default pvc
- `useEphemeralStorage: false` + `pvc == {...}` ===> use the pvc that the user has defined, whatever it might be (correct or not)

baichinger: The entry point for validating PVCs is in https://github.com/kubernetes/kubernetes/blob/87c1ca88d45ac7ec2b37d46187f40ff70d36aee1/pkg/registry/core/persistentvolumeclaim/strategy.go with its `Validate` method. If validation is not strictly required at that stage, I suggest to defer it. It can always be added later.

0sewa0 (approved):

andriisoldatenko (approved):

PR #4009, opened 2024-11-04
Simplify URL parsing for CSI driver endpoint argument

gkrenn:
## Description

The CSI driver currently relies on a custom implementation for `URI` parsing. This pull request updates the logic to use the built-in `net/url` package and modifies the endpoint defined in the CSI `daemonset` to be an absolute path.

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-19)

## How can this be tested?

- deploy the operator 
- the CSI driver running is proof enough, as the livenessprobe relies on the socket to be created.


albertogdd (approved):
Works well :+1:

PR #4010, opened 2024-11-04
Remove kubebuilder defaults

0sewa0:
## Description

[DAQ-1435](https://dt-rnd.atlassian.net/browse/DAQ-1435)

Removes every usage of `// +kubebuilder:default` in the latest versions of the CRDs.

Reasoning:
- These defaults are not always respected, and can cause some weird behaviour.
- Resulting CR is confusing to users 
   - "why did a field in the CR I didn't define?"
- In unittest you have to "imitate" this default setting.
   - because the defaults were enforced by the api-server and not our code

Sidenote:
- There are some "weird" ordering changes in `structs`, this is because of using pointer values for fields that used to have the `// +kubebuilder:default`. (`betteralign -apply ./...` went crazy)
- `address.Of` function is no longer restricted to a random set of types

⚠️ Important change: [Make DynatraceApiRequestThreshold uint16](https://github.com/Dynatrace/dynatrace-operator/pull/4010/commits/40166183a3f9cb6d7f242053a9cb5e5d118cf463)

## How can this be tested?
Deploy DK/EC without setting fields that have defaults, make sure everything works as expected.
- Also check that when overwriting the defaults is also respected.

[DAQ-1435]: https://dt-rnd.atlassian.net/browse/DAQ-1435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/api/v1beta3/dynakube/dynakube_props.go
```diff
@@ -120,12 +120,16 @@ func (dk *DynaKube) TenantUUIDFromConnectionInfoStatus() (string, error) {
 	return "", errors.New("tenant UUID not available")
 }
 
-func (dk *DynaKube) ApiRequestThreshold() time.Duration {
-	if dk.Spec.DynatraceApiRequestThreshold < 0 {
-		dk.Spec.DynatraceApiRequestThreshold = DefaultMinRequestThresholdMinutes
+func (dk *DynaKube) GetDynatraceApiRequestThreshold() int {
+	if dk.Spec.DynatraceApiRequestThreshold == nil || *dk.Spec.DynatraceApiRequestThreshold < 0 {
```
waodim, line 124: If I am not mistaken there happens to be a little issue here:

I applied a dynakube with a threshold of -1, which should fall back then to the default 15. This happens rightly so, however, if you describe the dynakube it states that -1 is stored as a threshold. Lemme show you:

the dynakube describe:

```
│ Name:         dynakube                                                                                                                                                                                           │
│ Namespace:    dynatrace                                                                                                                                                                                          │
│ Labels:       <none>                                                                                                                                                                                             │
│ Annotations:  <none>                                                                                                                                                                                             │
│ API Version:  dynatrace.com/v1beta3                                                                                                                                                                              │
│ Kind:         DynaKube                                                                                                                                                                                           │
│ Metadata:                                                                                                                                                                                                        │
│   Creation Timestamp:  2024-11-05T12:33:39Z                                                                                                                                                                      │
│   Generation:          7                                                                                                                                                                                         │
│   Resource Version:    56374999                                                                                                                                                                                  │
│   UID:                 a8d8132a-135a-41b9-bf0a-72bc85a09d5e                                                                                                                                                      │
│ Spec:                                                                                                                                                                                                            │
│   API URL:                          https://asj34817.dev.dynatracelabs.com/api                                                                                                                                   │
│   Dynatrace API Request Threshold:  -1    
```


the proof that in code it is set to 15:

`│ {"level":"info","ts":"2024-11-06T08:49:05.559Z","logger":"oneagent-connectioninfo","msg":"skipping OneAgent connection info update, last request was made less than 15 minutes ago, 14 minutes remaining until n │`

0sewa0: actually, it works as expected, kinda, we should probably deny this input in validation

0sewa0: > Also if you do not set a threshold at all, it won't show up in the describe at all (in code again it works)

that is what we want

chrismuellner: don't we allow a value of `-1` in some places to disable the field, i.e. no request threshold, or do we just say it should be set to `0`?

0sewa0: just set it to 0

waodim: Alright but isn't it somehow misleading if someone for whatever (accidentally for example) reason applies a dynakube with a negative value then describes the dynakube sees the negative value but in code 15 is used? Shouldn't then the threshold show 15 in the describe (maybe with a validation warning, "threshold set to an invalid value, fall to default 15" or something along those lines)?

waodim: even nicer 👍 I know this probably not even such an important finding but I was still somewhat confused. Thanks for clearing that up

chrismuellner: >  I will make it an uint

are negative values rejected then or does it error if one is configured?

0sewa0: we will see :D

0sewa0: now you should get 
```
```

pkg/api/v1alpha2/edgeconnect/edgeconnect_types.go
```diff
@@ -37,35 +36,37 @@ type EdgeConnectSpec struct { //nolint:revive
 	// +kubebuilder:validation:Optional
 	Proxy *proxy.Spec `json:"proxy,omitempty"`
 
+	// ServiceAccountName that allows EdgeConnect to access the Kubernetes API
+	ServiceAccountName *string `json:"serviceAccountName,omitempty"`
+
+	// Enables automatic restarts of EdgeConnect pods in case a new version is available (the default value is: true)
+	AutoUpdate *bool `json:"autoUpdate"`
+
```
0sewa0, line 44: @chrismuellner I update the EC CRD aswell, but it will have no new version in the next release, should I undo it, or is it fine? (and create a follow up to track it)

I can just revert [Remove kubebuilder:default from EdgeConnect](https://github.com/Dynatrace/dynatrace-operator/pull/4010/commits/f74846fbfb3c909d5dca5518466572defb1f888e), so it's not much work

chrismuellner: do we consider this a customer facing change? I think a customer shouldn't notice this change

0sewa0: ok ok

waodim (commented):
Nothing further to add from my side, it works and if you fix the unit tests I will approve

waodim (dismissed):

chrismuellner (approved):
Looks good!

`v1beta3` without defaults:
```yaml
spec:
  activeGate:
    capabilities:
    - routing
    - kubernetes-monitoring
    - dynatrace-api
  apiUrl: ...
  oneAgent:
    cloudNativeFullStack: {}
```

Same Dynakube when applying `v1beta2`:
```yaml
spec:
  activeGate:
    capabilities:
    - routing
    - kubernetes-monitoring
    - dynatrace-api
    replicas: 1
    resources: {}
  apiUrl: ...
  dynatraceApiRequestThreshold: 15
  extensions: {}
  metadataEnrichment:
    enabled: false
    namespaceSelector: {}
  oneAgent:
    cloudNativeFullStack:
      autoUpdate: true
      namespaceSelector: {}
      oneAgentResources: {}
  templates:
    extensionExecutionController:
      imageRef: {}
      resources: {}
      imageRef: {}
      nodeAffinity: {}
      resources: {}
    openTelemetryCollector:
      imageRef: {}
      resources: {}
```

waodim (approved):

albertogdd (approved):

PR #4011, opened 2024-11-05
Fix automatic index.yaml PRs to not remove existing releases

albertogdd:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-1941

Every time we are releasing a new version we have to manually update the resulting automatic PR that updates the helm index.

Example: https://github.com/Dynatrace/dynatrace-operator/pull/4005

The issue is that we always update `main` branch index (because it's the one that is used) but we never update the release branch ones. So the `helm repo index --merge` command is executed over a index file that is outdated.

Now we do git restore of index.yaml from main branch to use the latest version.

## How can this be tested?

Test git restore with a single file in a GH action - [I did it here](https://github.com/albertogdd/alberto-dynatrace-operator/actions/runs/11776910088/job/32800241548)


.github/workflows/release.yaml
```diff
@@ -424,3 +424,25 @@ jobs:
             Helm upgrade to and install of `${{ github.ref }}` works.
             ## Checklist
             - [x] PR is labeled accordingly
+      - name: Create pull request for adding helm index to ${{ github.ref }} branch
+        if: ${{ !contains(github.ref, '-rc.') }}
+        uses: peter-evans/create-pull-request@v7
```
gkrenn, line 429: I know we don't do that consistently, but would be nice to use the sha value here

gkrenn (approved):
Does what it should do. - Is it really necessary though? - Wouldn't it be safer and easier to just pull the lastest `index.yaml` within `generate-new-helm-index-yaml.sh` before updating the yaml?

If not see the PR as approved.

luhi-DT (commented):
Do we need to update both index files (on main and release branch) or is there a way to execute helm repo index --merge in a different way? I dont like having index.yamls on both release and main branch and need to keep them in sync :(

luhi-DT:
> @gkrenn @luhi-DT @chrismuellner I will have a look... we would need to checkout both branches and copy the file from one to another. Easier said than done with github actions 🤡

I agree, this is not easy, but we should look into that, thanks @albertogdd :)

albertogdd:
Just realized a sparse checkout could work. I have to test it somehow but let me know your thoughts about these changes
@gkrenn @luhi-DT @chrismuellner

.github/workflows/release.yaml
albertogdd: Good idea :+1: done!

gkrenn:
This solution looks promising! While a single file checkout might have sufficed, using sparse checkout seems to work well too. Please test it first to ensure it functions as expected. 👍

.github/workflows/release.yaml
```diff
@@ -395,6 +395,17 @@ jobs:
           prerelease: false
           draft: true
           fail_on_unmatched_files: true
+      - name: Checkout index.yaml from main branch
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: main
+          sparse-checkout: config/helm/repos/stable/index.yaml
+          sparse-checkout-cone-mode: false
```
gkrenn, line 403: "non-cone mode is deprecated" => https://git-scm.com/docs/git-sparse-checkout#_internalsnon_cone_problems

gkrenn: I’m not sure if it’s a good idea to introduce deprecated features, even though the GitHub Actions documentation mentions them.

I would prefer to do this without using an action:

```
git restore --source main -- config/helm/repos/stable/index.yaml
```

This way, you can also skip the “Copy index.yaml file” step.

albertogdd: Awesome idea! It works :+1: I just updated the PR

andriisoldatenko (approved):

luhi-DT (approved):

PR #4012, opened 2024-11-06
Remove enabled field for extensions

StefanHauth:
## Description

Remove the `.enabled` flag from the `.extensions` section. Extensions shall be enabled by using

```
 extensions: {}
```

If the extensions section is missing, no extension-related components shall be deployed.

## How can this be tested?

* Deploy a Dynakube with extensions enabled with the new schema. Check that EEC and collector are deployed
* Deploy a Dynakube without an extensions section. Check that EEC and collector are not deployed


pkg/api/v1beta3/dynakube/extensions.go
```diff
@@ -6,8 +6,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+// +kubebuilder:validation:Optional
 type ExtensionsSpec struct {
-	// +kubebuilder:validation:Optional
 	Enabled bool `json:"enabled,omitempty"`
```
chrismuellner, line 11: `Enabled` flag is still in the spec here

chrismuellner: you could add some description on what the `extensions` section does as well!

pkg/api/v1beta3/dynakube/dynakube_types.go
```diff
@@ -76,6 +76,9 @@ type DynaKubeSpec struct { //nolint:revive
 	// +kubebuilder:validation:Optional
 	Kspm *kspm.Spec `json:"kspm,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	Extensions *ExtensionsSpec `json:"extensions,omitempty"`
```
chrismuellner, line 80: why was this reordered?

chrismuellner (changes requested):

pkg/api/v1beta3/dynakube/dynakube_types.go
StefanHauth: Because of the field alignment in the struct. Seems like all the pointer members have to cuddle on the top.

pkg/api/v1beta3/dynakube/extensions.go
StefanHauth: Oha, fixed in 37ac723

chrismuellner (approved):

aorcholski (approved):
LGTM

Issue #4008, opened 2024-10-31
Webhook PodDistributionBudget created when HighAvailability is set to False via Helm Deployment


coolmancgfs:
**Describe the bug**
When deploying the cloud native fullstack operator v1.3.2 via helm the webhook resource is created with a PodDistributionBudget that has `minAvailable: 1` even when you set highAvailability: false in the values.yaml.  The webhook then runs as a single replica with PDB that does not allow eviction.  This prevents eviction and blocks nodes from being drained for shutdown, resulting in blocking k8s upgrades.

**To Reproduce**
1.  Deploy the DT K8s Operator Cloud Native via Helm with the webhook values.yaml `highAvailability: false`
2. Attempt to drain node with webhook deployment running.

**Expected behavior**
The webhook deployment with `highAvailability: false` does not have a PDB that prevents eviction of the single pod

**Screenshots**
![image](https://github.com/user-attachments/assets/53434617-64c3-4da3-a74b-e24056c4e770)

**Additional context**
This appears to be a regression from this commit:
https://github.com/Dynatrace/dynatrace-operator/commit/83801f69cadfcb43509fe765bef4b692975609e5#diff-67f1e210b8d311b90567c94206d9d6b9d8bfa16967742f8f19fc0ffbeeb3bd17

As a workaround you can deploy the webhook in HA mode to allow multiple replicas and the PDB will not prevent eviction of a 1 pod.

luhi-DT:
Hi @coolmancgfs,

thanks for the report, will create an internal bug for this. A fix for this will be available in Operator v1.4.0

PR #4013, opened 2024-11-06
Fix PodDisruptionBudget not obeying highAvailability flag

luhi-DT:
## Description

Ticket: https://dt-rnd.atlassian.net/browse/DAQ-2140

Fixes https://github.com/Dynatrace/dynatrace-operator/issues/4008

Currently the highAvailability flag does not control the creation of the PodDisruptionBudget

## How can this be tested?

set highAvailability to false in values file -> make manifest -> check if the PDB is there


waodim (approved):
tested it - works 👍

chrismuellner (approved):

PR #4018, opened 2024-11-07
Implement rate limiter for GRPC server

andriisoldatenko:
## Description

This PR introduce:

and fixes https://dt-rnd.atlassian.net/browse/DAQ-2252

- switch journal=WAL for SQLite connection string (to avoid lock database error)
- implement rate limiter for Publish/Unpublish grpc server methods
- increase liveness probe to up-to `4s` https://github.com/Dynatrace/dynatrace-operator/pull/4018/files#diff-7e50509ea7a8d1b31acdb20b0d93c4dbb51cb3e5940a654756793cf3b4485462R216 (because under heavy load performance and small limit of CPU usage, response time increases from GRPC server).

## How can this be tested?

1. Deploy Operator release-1.3 version without patch and deploy dk:
```
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
spec:
  activeGate:
    capabilities:
      - routing
  oneAgent:
    applicationMonitoring:
      useCSIDriver: true
```
2. Create sample-app namespace and deploy sample apps to inject oneAgent:
```
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  name: sample-app
  namespace: sample-app
spec:
  replicas: 200
  selector:
    matchLabels:
      tier: sample-app
  template:
    metadata:
      labels:
        tier: sample-app
    spec:
      containers:
        - name: pause
          image: gcr.io/google_containers/pause-amd64:3.0
```
3. It's easy to run 100 pods using regular dev GKE cluster.
4. Monitor csi driver logs, and expose metrics endpoint and monitor for example number of `gorotines` per pod:
```
kubectl port-forward -n dynatrace dynatrace-oneagent-csi-driver-fh8gx 8080:8080
```

```
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 26
```

If you start collecting metric and visualise you we get something like it, but keep in mind ( we tested using bigger cluster and ~ 450-460 pods per node.)
Screenshot 2024-11-08 at 07 51 18

### DIFF
--- a/pkg/controllers/csi/driver/server.go
+++ b/pkg/controllers/csi/driver/server.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	dtcsi "github.com/Dynatrace/dynatrace-operator/pkg/controllers/csi"
@@ -39,6 +40,10 @@ import (
 	mount "k8s.io/mount-utils"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
+
+const MaxGrpcRequests = 20
+
+var counter atomic.Int32
 
 type Server struct {
 	csi.UnimplementedIdentityServer
@@ -96,7 +101,7 @@ func (svr *Server) Start(ctx context.Context) error {
 		return fmt.Errorf("failed to start server: %w", err)
 	}
 
-	server := grpc.NewServer(grpc.UnaryInterceptor(logGRPC()))
+	server := grpc.NewServer(grpc.UnaryInterceptor(grpcLimiter()))
 
 	go func() {
 		ticker := time.NewTicker(memoryMetricTick)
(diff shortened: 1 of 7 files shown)


0sewa0 (approved):

pkg/controllers/csi/driver/server.go
```diff
@@ -233,25 +238,37 @@ func (svr *Server) NodeExpandVolume(context.Context, *csi.NodeExpandVolumeReques
...
+		default:
+			resp, err := handler(ctx, req)
+			if err != nil {
+				log.Error(err, "GRPC failed", "full_method", info.FullMethod)
+			}
+
+			return resp, err
+		}
+
+		if counter.Load() > MaxGrpcRequests {
```
gkrenn, line 261: Like the rework into a switch statement. 

Would rather implement the rate limiter as an own interceptor though, it's not really related to `logGRPC`.

gkrenn (changes requested):

pkg/controllers/csi/driver/server.go
0sewa0: I think just renaming it would make sense, the "log" interceptor is kinda redundant at that point.
We are rate limiting, and log as a side thing

0sewa0: For you to have 2 interceptors, you would (AFAIK) need an `grpc_middleware.WithUnaryServerChain` for it to be called in order. And that would require a new dependency AND just overcomplicate things

0sewa0: just call it `grpcLimiter` or something like that instead of `logGRPC`

andriisoldatenko: agree with @0sewa0 about renaming.

pkg/controllers/csi/driver/server.go
```diff
@@ -233,25 +238,37 @@ func (svr *Server) NodeExpandVolume(context.Context, *csi.NodeExpandVolumeReques
...
+			return nil, status.Error(codes.ResourceExhausted, fmt.Sprintf("rate limit exceeded, current value %d more than max %d", counter.Load(), MaxGrpcRequests))
 		}
 
+		counter.Add(1)
```
gkrenn, line 265: Be aware that if you increase the counter after the if, the rate limiter will never be precise. 

It can happen that multiple go routines are already executing after the if and then increase the counter. Would put it in front to ensure that the request limit is actually enforced.

pkg/controllers/csi/driver/server.go
andriisoldatenko: the problem that we don't have another logic :) except to log and limit, we want log when we limit as well. So create 2 functions sounds overkill.

pkg/controllers/csi/driver/server.go
andriisoldatenko: @gkrenn  could you please be more percise? after which "if"?

andriisoldatenko: fixed in 87ab28812

wepudt (approved):
tested in my cluster and rate limit works as found in pod describe of sample app. no restarts of the CSI driver or any other anomalies found!

aorcholski (approved):
LGTM tested on my GKE cluster with prometheus enabled and the default `server` resources
- resources.limits.cpu: 50m,
- resources.limits.memory: 100Mi
No restarts.

gkrenn (approved):

PR #4028, opened 2024-11-08
Update CSI manifest to include all resources

gkrenn:
## Description

It was agreed that the `*-csi.yaml` manifests should contain all necessary objects for the operator.

This PR includes the following changes:
- Adds the missing objects.
- Removes all `*-all.yaml` files, as they are no longer needed and would be identical to the `*-csi.yaml` files.
- Simplifies the manifest make commands.
- Sets `OLM=true` when `/olm` make commands are called.
  - Note: This previously generated incorrect files in the release actions, but they are not yet in use, so this is not an issue.

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-1408)

## How can this be tested?

- `make manifests`


0sewa0 (approved):

aorcholski (approved):
LGTM I compared 
- openshift.yaml with openshift-csi.yaml
- kubernetes.yaml with kubernetes-csi.yaml

PR #4038, opened 2024-11-11
Update description for Dynakube version field

chrismuellner:
## Description
Align description for `version`/`image`/`codeModulesImage` fields.

https://dt-rnd.atlassian.net/browse/DAQ-1990

## How can this be tested?

Check CRD fields:
```
kubectl explain dynakubes.dynatrace.com.spec.oneAgent.cloudNativeFullStack
```


pkg/api/v1beta3/dynakube/oneagent_types.go
```diff
@@ -133,7 +133,7 @@ type AppInjectionSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Resource Requirements",order=15,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:resourceRequirements"}
 	InitResources *corev1.ResourceRequirements `json:"initResources,omitempty"`
 
-	// The OneAgent image that is used to inject into Pods.
+	// Use a custom OneAgent image to download binaries
```
0sewa0, line 136: this sounds a bit too general to me

chrismuellner: I didn't update the wording too much, just tried to get the format in line with version/image fields.

In our docs we use the term "OneAgent CodeModules" so I'll go with that. Still not sure if "downloading binaries" is the correct short form though:
```suggestion
	// Use a custom OneAgent CodeModules image to download binaries
```

andriisoldatenko:
@chrismuellner could you please update docs?
just run `make doc/api-ref` and commit changes, thanks

chrismuellner:
@andriisoldatenko should those files even be checked into the repository? I think they are mainly used to update the official docs and could just be generated when needed.

Running them shows that they were not updated for previous changes either and the defaults in the generated tables seem to be incorrect now that the kubebuilder markers are gone. I think we need to redo the script 🙂

0sewa0 (approved):

andriisoldatenko:
> @andriisoldatenko should those files even be checked into the repository? I think they are mainly used to update the official docs and could just be generated when needed.
> 
> Running them shows that they were not updated for previous changes either and the defaults in the generated tables seem to be incorrect now that the kubebuilder markers are gone. I think we need to redo the script 🙂

the problem if you not update them during PR, sometimes it can not work, so when you need them you need to debug who did change which brokes scripts. By updating in each PR it's much easier and less changes.

chrismuellner:
The script is broken and non-functional in its current state. It needs a follow up to fix it and not me running it again and pretending the output is correct.

andriisoldatenko:
> The script is broken and non-functional in its current state. It needs a follow up to fix it and not me running it again and pretending the output is correct.

thats okey, what's wrong with the script I'll create follow up ticket?

andriisoldatenko (approved):

PR #4039, opened 2024-11-11
Infer csi-usage based on csi-driver-availability

0sewa0:
## Description

[DAQ-1438](https://dt-rnd.atlassian.net/browse/DAQ-1438)

Removes the following:
- `spec.oneAgent.applicationMonitoring.useCSIDriver: true/false`:
   - Was used to configure if the CSI-driver should be used for application-monitoring (for injected monitored apps) 	
- `oneagent-readonly-host-fs` feature-flag
    - Was used to configure if the CSI-driver should be used for host-monitoring and cloud-native OS agents

So the expected behavior is:
- CSI-driver is used for `applicationMonitoring` IF the csi-driver was installed
   - in case the CSI-driver is missing, then the injected init-container will do the download  
- CSI-driver is **required** for `cloudNativeFullstack`
   - in case the CSI-driver is missing, the validation webhook will deny the Dl
- CSI-driver is used for `hostMonitoring` IF the csi-driver was installed
   - in case the CSI-driver is missing, then the deployed OS agents will have an empty dir mounted instead.

What if I have 4 nodes, and only 2 have CSI drivers on them, and I want to use the csi-driver on the 2 nodes where it is available and not use it where it is not? (the "hybrid" configuration)
- This scenario is NOT supported

## How can this be tested?

I tested the following 3 configs, in 2 scenarios:

### With CSI deployed (`make deploy/helm`)

a. Application monitoring -> injected apps have csi-volume
```
  oneAgent:
    applicationMonitoring: {}
```

b. Host monitoring -> deployed OS agents have csi-volume
```
  oneAgent:
    hostMonitoring: {}
```

b. Cloud Native Fullstack  -> injected apps and deployed OS agents have csi-volume
```
  oneAgent:
    cloudNativeFullstack: {}
```

### With CSI deployed (`make deploy/helm/no-csi`)

a. Application monitoring -> injected apps have NO csi-volume, does the downloading in the init-container
```
  oneAgent:
    applicationMonitoring: {}
```

b. Host monitoring -> deployed OS agents have a not-readonly `host-root` mount
```
  oneAgent:
    hostMonitoring: 
      env:
          - name: ONEAGENT_ENABLE_VOLUME_STORAGE # needed on GKE
            value: "true"
```

b. Cloud Native Fullstack  -> validation webhook denies it
```
  oneAgent:
    cloudNativeFullstack: {}
```

[DAQ-1438]: https://dt-rnd.atlassian.net/browse/DAQ-1438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/helm/chart/default/templates/Common/operator/install-config.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
 data:
   modules.json: |
     {
+      "csidriver": {{ .Values.csidriver.enabled }},
```
andriisoldatenko, line 11:
```suggestion
      "csiDriver": {{ .Values.csidriver.enabled }},
```
just curious, why `activeGate` and `oneAgent` :)

config/helm/chart/default/templates/Common/csi/daemonset.yaml
```diff
@@ -170,6 +170,9 @@ spec:
           - name: MAX_UNMOUNTED_VOLUME_AGE
             value: "{{ .Values.csidriver.maxUnmountedVolumeAge}}"
           {{- end }}
+        envFrom:
```
chrismuellner, line 173: why is this necessary? if `csiDriver` is not enabled, the daemonset does not exist

pkg/util/installconfig/modules.go
```diff
@@ -44,27 +50,46 @@ type Modules struct {
...
+		override = nil
+	})
+}
+
 func GetModuleValidationErrorMessage(moduleName string) string {
 	return fmt.Sprintf(validationErrorTemplate, moduleName, moduleName)
 }
+
+func GetCSIValidationErrorMessage() string {
+	return fmt.Sprintf(validationErrorTemplate, "csi-driver", "CloudNative")
```
chrismuellner, line 94: > csi-driver has been disabled during Operator install. The necessary resources for CloudNative to work are not present on the cluster. Redeploy the Operator via Helm with all the necessary resources enabled.

CSI Driver was disabled during Operator install. The necessary resources for Cloud Native Fullstack to work are not present on the cluster. Redeploy the Operator via Helm with all the necessary resources enabled.

chrismuellner: I appreciate the attempt to make this as generic as possible, but it's kinda hard to judge what I'm supposed to do as a user. Can we somehow fit the relevant `values.yaml` section in here or is that too detailed?

pkg/api/v1beta3/dynakube/oneagent_props.go
```diff
@@ -77,30 +84,20 @@ func (dk *DynaKube) OneAgentConnectionInfoConfigMapName() string {
 	return dk.Name + OneAgentConnectionInfoConfigMapSuffix
 }
 
-func (dk *DynaKube) NeedsReadOnlyOneAgents() bool {
-	return (dk.HostMonitoringMode() || dk.CloudNativeFullstackMode()) &&
-		dk.FeatureReadOnlyOneAgent()
+func (dk *DynaKube) NeedsCSIDriver() bool {
```
chrismuellner, line 87: Naming with `Needs`, `CanUse` and `Use` for the CSI driver is confusing to me and could lead to incorrect usage. Can we simplify this?

pkg/api/v1beta3/dynakube/oneagent_props.go
```diff
@@ -77,30 +84,20 @@ func (dk *DynaKube) OneAgentConnectionInfoConfigMapName() string {
...
-func (dk *DynaKube) NeedsCSIDriver() bool {
-	isAppMonitoringWithCSI := dk.IsAppMonitoringWithCSI()
-
-	isHostMonitoringWithCSI := dk.HostMonitoringMode() && dk.FeatureReadOnlyOneAgent()
+func (dk *DynaKube) UseCSIDriver() bool {
+	return dk.NeedsCSIDriver() || (dk.CanUseCSIDriver() && dk.isEnabledModules().CSIDriver)
+}
 
-	return dk.CloudNativeFullstackMode() || isAppMonitoringWithCSI || isHostMonitoringWithCSI
+func (dk *DynaKube) NeedsReadOnlyOneAgents() bool {
```
chrismuellner, line 99: Since we are dropping the feature flag, do we still need this check or can it just be a check if CSI driver is used in case of host agent?

chrismuellner (changes requested):

pkg/util/installconfig/modules.go
0sewa0: on it

pkg/api/v1beta3/dynakube/oneagent_props.go
0sewa0: It's a mess because it is kinda a mess in general :D 
- There is 1 feature (cloudNative) that will not work without the CSI-driver, so that should be validated
- There is 1 feature that can't use the csi-driver (classic), so you have to keep in mind
- There are 2 features (app/host-mon) that will use it if available, another thing to keep in mind

I tried (still trying) to just have an `IsCSIEnabled` check everywhere, but the problem is that you have to check more, as the features of cloudnative/classic/hostmon/appmon have shared code.
- Examples where using only `IsCSIEnabled` works
    - in the mutation webhook (csi is available => use it, easy)
- Examples where using only `IsCSIEnabled` doesn't work
    - oneagent-daemonset generation, because of classic, that is why we need `NeedsReadOnlyOneAgents`
    - in the validation webhook
       - the `disabledCSIForReadonlyCSIVolume` check requires the very elaborate `UseCSIDriver` check, because of the validation logic is a bit weird (don't really see why we even have this 🤷 )
       - the `isCSIModuleDisabled` check needs to know if the CSI-driver is a hard requirement or not 

I am tyring to come up with better names AND moving validation-specific logic to the validation package, and use an `IsCSIEnabled` check where possible

waodim (commented):
```
Volumes:                                                                                                                                                                                                       │
│   host-root:                                                                                                                                                                                                   │
│     Type:          HostPath (bare host directory volume)                                                                                                                                                       │
│     Path:          /                                                                                                                                                                                           │
│     HostPathType:                                                                                                                                                                                              │
│   connection-info-secret:                                                                                                                                                                                      │
│     Type:        Secret (a volume populated by a Secret)                                                                                                                                                       │
│     SecretName:  dynakube-test-oneagent-tenant-secret                                                                                                                                                          │
│     Optional:    false     
```                        

These are the volumes I see in the oneAgent pods when running this scenario:

```
  oneAgent:
    hostMonitoring: 
      env:
          - name: ONEAGENT_ENABLE_VOLUME_STORAGE # needed on GKE
            value: "true"
```

Shouldn't there be an emptyDir? Am I missing something?

cmd/troubleshoot/dynakube_test.go
```diff
@@ -8,7 +8,6 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/shared/value"
```
waodim, line 8:
```
Volumes:                                                                                                                                                                                                       │
│   host-root:                                                                                                                                                                                                   │
│     Type:          HostPath (bare host directory volume)                                                                                                                                                       │
│     Path:          /                                                                                                                                                                                           │
│     HostPathType:                                                                                                                                                                                              │
│   connection-info-secret:                                                                                                                                                                                      │
│     Type:        Secret (a volume populated by a Secret)                                                                                                                                                       │
│     SecretName:  dynakube-test-oneagent-tenant-secret                                                                                                                                                          │
│     Optional:    false    
``` 

These are the volumes I see in the oneAgent pods when running this scenario:

```
  oneAgent:
    hostMonitoring: 
      env:
          - name: ONEAGENT_ENABLE_VOLUME_STORAGE # needed on GKE
            value: "true"
```

Shouldn't there be an emptyDir? Am I missing something?

pkg/util/installconfig/modules.go
0sewa0: Adding the helm value in the message can get out of sync

cmd/troubleshoot/dynakube_test.go
0sewa0: no, `ONEAGENT_ENABLE_VOLUME_STORAGE ` will cause the agent do it's container magic AFAIK

pkg/api/v1beta3/dynakube/oneagent_props.go
```diff
@@ -77,30 +82,8 @@ func (dk *DynaKube) OneAgentConnectionInfoConfigMapName() string {
...
-	return *dk.Spec.OneAgent.ApplicationMonitoring.UseCSIDriver
-}
-
-func (dk *DynaKube) NeedsCSIDriver() bool {
-	isAppMonitoringWithCSI := dk.IsAppMonitoringWithCSI()
-
-	isHostMonitoringWithCSI := dk.HostMonitoringMode() && dk.FeatureReadOnlyOneAgent()
-
-	return dk.CloudNativeFullstackMode() || isAppMonitoringWithCSI || isHostMonitoringWithCSI
+func (dk *DynaKube) UseReadOnlyOneAgents() bool {
```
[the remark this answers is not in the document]
0sewa0: it is used both in validation and daemonset generation, so yes
- this is mainly annoying due to having classic in the daemonset generation, otherwise it could be moved to the validation package

also `UseCSIVolume` is a bit too generic, as it does not consider the application monitoring (I used to have a check for that but you considered it too convoluted 😅)

cmd/troubleshoot/dynakube_test.go
waodim: Then pls update the review section as you wrote there that there should be an emptyDir in this scenario

Otherwise I tested everything and it LGTM 👍

pkg/api/validation/dynakube/modules.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 var (
-	errorCSIModuleDisabled           = installconfig.GetCSIValidationErrorMessage()
+	errorCSIModuleRequired           = "CSI Driver was disabled during Operator install. It is a necessary resource for Cloud Native Fullstack to work. Redeploy the Operator via Helm with the CSI-driver enabled."
```
waodim, line 11:
```suggestion
	errorCSIModuleRequired           = "CSI Driver was disabled during Operator install. It is a necessary resource for Cloud Native Fullstack to work. Redeploy the Operator via Helm with the CSI Driver enabled."
```

waodim: nitpick just for consistency

waodim (approved):
lgtm and it also worked for me

chrismuellner (approved):

chrismuellner (approved):

waodim (approved):

PR #4047, opened 2024-11-12
Cleanup Dynakube's config and enrichment secrets when deleting it

albertogdd:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-2266

Currently, `dynatrace-dynakube-config` and `dynatrace-metadata-enrichment-endpoint` secrets doesn't get deleted when the Dynakube gets deleted.

This PR adds the cleanup of those.

## How can this be tested?

- On main branch: deploy a DK and delete it, check that these secrets doesn't get deleted
- On this branch: do the same and check that they get deleted


pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -59,7 +61,20 @@ func (dm DynakubeMapper) UnmapFromDynaKube(namespaces []corev1.Namespace) error
 		}
 	}
 
-	return nil
+	return dm.cleanup(namespaces)
+}
+func (dm DynakubeMapper) cleanup(namespaces []corev1.Namespace) error {
+	nsList := make([]string, 0, len(namespaces))
```
0sewa0, line 67: would add a check that if `len(namespaces) == 0` => _do nothing_

0sewa0: probably could do it in `UnmapFromDynaKube`

it wasn't needed before because the `for ... ` did nothing if `namespaces` were empty

albertogdd: Good idea! I also added some nice refactorings also based on your input ;) 
https://github.com/Dynatrace/dynatrace-operator/pull/4047/commits/e6ec109d74338fa9c8862575457e6b864ffa3f9c

waodim (approved):
tested it and the secrets get deleted - codewise nothing further to add from me

pkg/util/kubeobjects/internal/query/query.go
```diff
@@ -218,16 +218,22 @@ func (c Generic[T, L]) createOrUpdateForNamespaces(ctx context.Context, object T
...
 }
 
+func (c Generic[T, L]) DeleteForNamespace(ctx context.Context, objectName string, namespace string) error {
+	c.Log.Info("deleting object from namespace", "name", objectName, "namespace", namespace)
+
+	c.Target.SetName(objectName)
+	c.Target.SetNamespace(namespace)
+
+	return c.Delete(ctx, c.Target)
+}
```
0sewa0, line 228: I like it, but IMO what we should do (in a follow up, as it will touch many things), is to replace the `Delete` with this `DeleteForNamespace`, as it's much nicer to use

Reasoning:
- `c.Delete` is annoying to use, because you have to create a whole object, eventhough the only thing that matters is the namespace and name, all the rest is useless info
- "Unpacking" an object is far simpler then creating a dummy object
   - what we have now:
      - when no object was built: `query.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: GetSecretName(r.dk.Name), Namespace: r.dk.Namespace}})` 🤢 
      - when an object was built: `query.Delete(ctx, secret)` (this is RARE)
   - what we should have: 
     - when no object was built:`query.Delete(ctx, someName, someNamespace)` 	
     - when an object was built:`query.Delete(ctx, secret.Name, secret.Namespace)`

Could you create a follow up for this? 🥺

0sewa0 (approved):

waodim (approved):

PR #4053, opened 2024-11-13
Add builtin:logmonitoring.log-storage-settings to CR and reconciler

waodim:
## Description

[Ticket](https://dt-rnd.atlassian.net/browse/DAQ-1342)

With this it is possible to define ingest rule matchers in the dynakube. This configuration needs to be stored in settings (schema `builtin:logmonitoring.log-storage-settings`) in the scope of the current cluster ( = `KubernetesClusterMEID`).

This logic queries inside the logmonitoring reconciler all settings with this schemaID, and if there is none, we create new ones based on the defined ingest rule matchers in the dynakube (default is empty).

IF we query the settings and there are already some defined, I decided to add a condition to the dynakube that says exactly that.

## How can this be tested?

Go to the Environment API v2 page in your tenant and try it out. It is in the Settings section. At first you'll need to enable `com.compuware.apm.webuiff.config.core.hierarchy.resolution.pg.k8workload.pgwlhr.feature`

Deploy a dynakube with a valid logmonitoring section for example: 

```
  logMonitoring:
    ingestRuleMatchers:
    - attribute: "k8s.deployment.name"
      values:
        - "dynatrace"
        - "anothertest"
    - attribute: "k8s.container.name"
      values:
        - "testContainerName"
  templates:
      logMonitoring:
        imageRef:
          tag: latest
```

Wait a little bit then run in the environment API page the get query for the settings with the schema. See if it is created properly.

It should look like this: 

```
{
  "items": [
    {
      "objectId": "vu9U3hXa3q0AAAABACpidWlsdGluOmxvZ21vbml0b3JpbmcubG9nLXN0b3JhZ2Utc2V0dGluZ3MAEktVQkVSTkVURVNfQ0xVU1RFUgAQMDFGNjZGMjkxRjBENjU0MAAkMzcxM2RjMDktMDYzNC0zMjM0LTlhNTAtYTg4MTA4OGY2YmI0vu9U3hXa3q0",
      "value": {
        "enabled": true,
        "config-item-title": "dynakube-test",
        "send-to-storage": true,
        "matchers": [
          {
            "attribute": "k8s.deployment.name",
            "operator": "MATCHES",
            "values": [
              "dynatrace",
              "anothertest"
            ]
          },
          {
            "attribute": "k8s.container.name",
            "operator": "MATCHES",
            "values": [
              "testContainerName"
            ]
          }
        ]
      }
    }
  ],
  "totalCount": 1,
  "pageSize": 100
}
```

You can then grab the objectID, delete the setting and reapply a dynakube with an empty logmonitoring section and see if default settings gets created.

That looks like this:

```
[
  {
    "schemaId": "builtin:logmonitoring.log-storage-settings",
    "scope": "KUBERNETES_CLUSTER-D3A3C5A146830A79",
    "value": {
      "enabled": true,
      "config-item-title": "my-cluster-name",
      "send-to-storage": true,
      "matchers": []
    }
  }
]
```


pkg/clients/dynatrace/settings_logmonitoring.go
```diff
@@ -0,0 +1,127 @@
...
+	Title    string               `json:"config-item-title,omitempty"`
+	Matchers []IngestRuleMatchers `json:"matchers,omitempty"`
+	Enabled  bool                 `json:"enabled,omitempty"`
+	Storage  bool                 `json:"send-to-storage,omitempty"`
+}
+
+type Item struct {
+	ObjectID string `json:"objectId,omitempty"`
+	Val      Value  `json:"value,omitempty"`
+}
```
0sewa0, line 31: we are in the big dtclient package, so these names are a bit too general (also do they need to be public?)

you can solve this 2 ways:
1. Come up with more specific names (will also help with the naming of `Item.Val`)
2. I saw a pattern that you could use (we don't really do it)
   - You can define `structs` within a function, so it can be only used in it
   - This is useful when you only need the struct to create the body for a POST request
   - So you just use it in a function to create a `[]byte` using `json.Marshal`, and then just pass it around as a `[]byte`
      - This is nice to not pollute the package with throw-away types. 
      -  Be careful, as not all types may be that `throw-away`

pkg/controllers/dynakube/logmonitoring/reconciler.go
```diff
@@ -65,5 +68,31 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
...
+
+		objectId, err := r.dtc.CreateLogMonitoringSetting(ctx, dtclient.LogMonitoringSettingsSchemaId, r.dk.Status.KubernetesClusterMEID, r.dk.Status.KubernetesClusterName, matchers)
+
+		if err != nil {
+			return errors.WithMessage(err, "error when creating log monitoring setting")
+		}
+
+		log.Info("logmonitoring setting created", "settings", objectId)
+	}
+
```
0sewa0, line 96: you should put this into it's own function

0sewa0 (changes requested):

pkg/clients/dynatrace/settings_logmonitoring.go
0sewa0: Either way is fine by me. Way 1. is simpler and faster, but 2. is probably the "proper way".

However, we might want to look into using the 2. way all around the dtclient. (or to just restructure it a bit, as it's ever growing)

waodim: I would gladly try to do 2. But then again, wouldn't this be better a whole refactoring story, where we update the whole client package? (I am not sure how big of a change this would be)

pkg/controllers/dynakube/logmonitoring/reconciler.go
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/4053/commits/a4767e415761f112227f6ec964bd1ae7549845e4

pkg/clients/dynatrace/settings_logmonitoring.go
0sewa0: Just do option 1. for now
And create a follow up for possible improvements for option 2.

waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/4053/commits/98b868536e9d5168d1d3db5548f1d2afa67eef45

Also: I needed to alter the maxDescLen for our manifests a little bit because it got too large. I do not know why I did not get this error before.

hack/make/manifests/config.mk
```diff
@@ -1,4 +1,4 @@
-CRD_OPTIONS ?= "crd:crdVersions=v1,maxDescLen=350,ignoreUnexportedFields=true"
+CRD_OPTIONS ?= "crd:crdVersions=v1,maxDescLen=300,ignoreUnexportedFields=true"
```
0sewa0, line 1: 😢

pkg/clients/dynatrace/settings_logmonitoring.go
```diff
@@ -0,0 +1,127 @@
...
+	Enabled         bool                 `json:"enabled,omitempty"`
+	SendToStorage   bool                 `json:"send-to-storage,omitempty"`
+}
+
+type logMonitoringItem struct {
+	ObjectID           string             `json:"objectId,omitempty"`
+	LogMonitoringValue logMonitoringValue `json:"value,omitempty"`
+}
+
+type posLogMonSettingsBody struct {
```
0sewa0, line 33: Try to have a more consistent names:
- `logMonitoringValue` -> `logMonSettingsValue`
- `logMonitoringItem` -> `logMonSettingsItem`
- `posLogMonSettingsBody` -> `postLogMonSettingsBody`

just some ideas

pkg/clients/dynatrace/settings_logmonitoring.go
```diff
@@ -0,0 +1,127 @@
...
+	SchemaId      string             `json:"schemaId"`
+	SchemaVersion string             `json:"schemaVersion"`
+	Scope         string             `json:"scope,omitempty"`
+	Value         logMonitoringValue `json:"value"`
+}
+
+const (
+	LogMonitoringSettingsSchemaId = "builtin:logmonitoring.log-storage-settings"
```
0sewa0, line 41: does it have to be pulbic?

pkg/clients/dynatrace/settings_logmonitoring.go
waodim: tried my best: https://github.com/Dynatrace/dynatrace-operator/pull/4053/commits/5a091144b9e1fa2cbaff2d918020a6902f83aeae

pkg/clients/dynatrace/settings_logmonitoring.go
waodim: no, you're right, changed it here: https://github.com/Dynatrace/dynatrace-operator/pull/4053/commits/6a09f0c3111f5ba96d35d0bd25daf61e2824a777

pkg/controllers/dynakube/logmonitoring/reconciler.go
```diff
@@ -65,5 +68,40 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 		return err
 	}
 
+	err = r.checkLogMonitoringSettings(ctx)
```
0sewa0, line 71: Shouldn't this be a separate reconciler?
- as its rather simple this is prooobably fine as is.

You should try to limit how often we query for it, as now the logs are a bit spammy (we check it everytime)
- also we should remove the condition when logmonitoring is turned off (the setting does not need to be deleted, as that is not a requirement as of right now)

Try to do something similar that we already do for connection-info (just an example)
- Cleanup condition if necessary
- Only check the setting if condition is outdated

And if you add all this fun stuff, then having it in a separate package/reconciler makes more sense 😉

0sewa0 (changes requested):
settings are created and only created once, so works

but we are constantly (ie.: every reconcile) check if the settings exist this is a bit of an overkill

pkg/controllers/dynakube/logmonitoring/reconciler.go
waodim: thanks for the input. I agree totally. Will do that. 👍

pkg/clients/dynatrace/settings.go
```diff
@@ -118,6 +123,40 @@ func (dtc *dynatraceClient) GetSettingsForMonitoredEntity(ctx context.Context, m
 	return resDataJson, nil
 }
 
+func (dtc *dynatraceClient) GetSettingsForLogModule(ctx context.Context, monitoredEntity string) (GetLogMonSettingsResponse, error) {
+	if len(monitoredEntity) < 1 {
```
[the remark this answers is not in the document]
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/4053/commits/13961d08a7523bef40715806129282fe0312e631

pkg/clients/dynatrace/settings_logmonitoring.go
```diff
@@ -0,0 +1,127 @@
...
+	SendToStorage   bool                 `json:"send-to-storage,omitempty"`
+}
+
+type logMonSettingsItem struct {
+	ObjectID           string              `json:"objectId,omitempty"`
+	LogMonitoringValue logMonSettingsValue `json:"value,omitempty"`
+}
+
+type posLogMonSettingsBody struct {
+	SchemaId      string              `json:"schemaId"`
```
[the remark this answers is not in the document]
waodim: done in https://github.com/Dynatrace/dynatrace-operator/pull/4053/commits/13961d08a7523bef40715806129282fe0312e631

I removed every `omitempty` besides scope

a. because an empty scope can be passed
b. we also did it the same in other settings files

aorcholski (commented):
I removed `ingestRuleMatchers` from `logMonitoring` section to get empty `logMinitoring: {}`. Then I updated my dynakube and had to wait ~9 minutes for the default matcher to be added to the settings object on the server.
Is it OK? I guess we need limit the number of queries we sent to the server. On the other hand it is a little bit annoying.

waodim:
> I removed `ingestRuleMatchers` from `logMonitoring` section to get empty `logMinitoring: {}`. Then I updated my dynakube and had to wait ~9 minutes for the default matcher to be added to the settings object on the server. Is it OK? I guess we need to limit the number of queries we sent to the server. On the other hand it is a little bit annoying.

Just a sidenote which is important IMO: we not only removed the `ingestRuleMatchers` in the dynakube but we also removed the settings in the Environment API manually while the condition was stored in the dynakube that there are already settings, then edited the dynakube -> this led the waiting time. IMO i think this is ok, as I do not think it should be a normal use case to delete the settings in the API manually while everything is up and running. But maybe @chrismuellner has more context there and knows more.

chrismuellner:
Operator currently has no good way of reacting to changes done on the Dynatrace API side. We are listening and reacting to changes in the Kubernetes cluster and need to find a way to better/quicker react to external changes.

This is not something we can quickly fix here, but we'll need to properly think about options we have. For the initial feature the delay is acceptable.

aorcholski (commented):
Could you please

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (r *reconciler) Reconcile(ctx context.Context) error {
+	if !conditions.IsOutdated(r.timeProvider, r.dk, conditionType) {
+		return nil
+	}
+
+	if !r.dk.LogMonitoring().IsEnabled() {
+		meta.RemoveStatusCondition(r.dk.Conditions(), conditionType)
+	}
```
aorcholski, line 42: I think we should add `return nil`here. Operator shouldn't send LM queries to the DT server if LogMonitoring is Not enabled?

waodim: I agree, you are right. https://github.com/Dynatrace/dynatrace-operator/pull/4053/commits/2320f89c12bb9bfa9a873d7c75f1fa45105f9707

pkg/clients/dynatrace/settings_logmonitoring.go
```diff
@@ -0,0 +1,127 @@
...
+		return "", fmt.Errorf("response is not containing exactly one entry %s", resData)
+	}
+
+	return resDataJson[0].ObjectId, nil
+}
+
+func createBaseLogMonSettings(clusterName, schemaId string, schemaVersion string, scope string, ingestRuleMatchers []logmonitoring.IngestRuleMatchers) posLogMonSettingsBody {
+	matchers := []IngestRuleMatchers{}
+
+	if len(ingestRuleMatchers) > 0 {
```
aorcholski, line 89: condition can be removed. `range` handles nil, []{} values

0sewa0 (approved):

PR #4069, opened 2024-11-18
Always add ActiveGate ports no matter the capability

gkrenn:
## Description

This PR changes the handling of the ActiveGate ports. They will be activated no matter the enabled capability, the only requirement is that at least 1 capability has to be activated.

## How can this be tested?

- install to cluster 
- create ActiveGate DynaKube 
- service `dynakube-activegate` should be created as well as the ports on the ActiveGate container


pkg/api/v1beta3/dynakube/activegate/props.go
```diff
@@ -93,9 +93,7 @@ func (ag *Spec) IsMetricsIngestEnabled() bool {
 }
 
 func (ag *Spec) NeedsService() bool {
```
chrismuellner, line 95: why keep the check at all? if there's an activegate there should be a service

chrismuellner (changes requested):

aorcholski (commented):
I just try to understand the logic related to AG service.

I think we need NeedsService() function because of KSPM.

going up the call stack :
```
func (r *Reconciler) deleteCapability(ctx context.Context, agCapability capability.Capability) error {
-> if err := r.deleteService(ctx, agCapability); err != nil {
-> if r.dk.ActiveGate().NeedsService() {
		return nil
     }
-> func (ag *Spec) NeedsService() bool {
	return len(ag.Capabilities) > 0 ||
		ag.enabledDependencies.extensions ||
		ag.enabledDependencies.kspm      <---- depends on KSPM
    }     
```
So KSPM needs the AG service to exist.

**In that case how do we create the AG service if KSPM is enabled?**

dynakube/activegate/reconciler.go:Reconcile()
```
for _, agCapability := range capability.GenerateActiveGateCapabilities(r.dk) {
	if agCapability.Enabled() {
			return r.createCapability(ctx, agCapability)           
```		
going up the call stack :
```
for _, agCapability := range capability.GenerateActiveGateCapabilities(r.dk) {
-> func NewMultiCapability(dk *dynakube.DynaKube) Capability {
-> if dk == nil || !dk.ActiveGate().IsEnabled() {
-> return len(ag.Capabilities) > 0 || ag.enabledDependencies.Any()
-> func (d dependencies) Any() bool {
	return d.extensions   <---- doesn't depend on KSPM
    }
```
we get disabled MultiCapability because KSPM has no meaning from AG capability point of view.

Then :
```
if agCapability.Enabled() {
-> func (capability *capabilityBase) Enabled() bool {
	return capability.enabled
    }
```		
returns `false` and instance of AG is not created - good. **Unfortunately this also means the AG service is not created - KSPM is unhappy, right?**

pkg/api/v1beta3/dynakube/activegate/props.go
gkrenn: Ok, will adapt it so it is always created

gkrenn:
@aorcholski  Enabling KSPM standalone should not be possible. It is blocked by the validation webhook, as it requires enabled `Kubernetes-monitoring` capability.

If kubemon is enabled, the service will be created.

We could add `kspm` to the `Any()` function in the ActiveGate spec just to make sure in case we accidentally delete the check in the validation webhook. What is your opinion? @chrismuellner  - As long as the check in the validation webhook is there everything would be fine, but modeling the dependency in the ActiveGate spec would be cleaner in my opinion: 
```
func (d dependencies) Any() bool {
	return d.extensions || d.kspm
}
```

chrismuellner:
> We could add kspm to the Any() function in the ActiveGate spec just to make sure in case we accidentally delete the check in the validation webhook. 

I think (hope) that's an unlikely scenario. If there are benefits to adding this validation in the Activegate spec as well then let's do it, otherwise let's not overcomplicate something that should either not be possible or handled by the validation webhook.

pkg/api/v1beta3/dynakube/activegate/props.go
gkrenn: Removed it

pkg/controllers/dynakube/activegate/reconciler_test.go
```diff
@@ -478,7 +483,7 @@ func TestServiceCreation(t *testing.T) {
 		}
 	})
 
-	t.Run("service exposes correct ports for multiple capabilities", func(t *testing.T) {
+	t.Run("service exposes all ports for all capabilities", func(t *testing.T) {
```
aorcholski, line 486: this test duplicates `t.Run("service exposes all ports for every capabilities", func(t *testing.T)` line 438.
It checks open ports for Routing capability - so can be deleted.

aorcholski (approved):
LGTM

pkg/api/v1beta3/dynakube/activegate/spec.go
```diff
@@ -64,7 +64,7 @@ type dependencies struct {
 }
 
 func (d dependencies) Any() bool {
-	return d.extensions
+	return d.extensions // kspm is a dependency too, but blocked by validation webhook to not run standalone
```
0sewa0, line 67: you can even remove the `kspm` dependency as it is no longer used/needed (the creation of the Service was the only reason for it to exist)

gkrenn: I don't quite get what you mean - As far as I got it, KSPM requires an ActiveGate and to communicate with the ActiveGate it needs a service. Why shouldn't there be a dependency and what would you remove? - The unused `kspm` variable?

0sewa0: KSPM does need an activeGate, but it has to be explicitly defined in the `DynaKube`, otherwise our validation will complain.

This `kspm` dependency was introduced only to create the `Service` (if you look at where it's used, you will see 😉 ).
- This was needed because the "required" activeGate capability was `kubernetes-monitoring`, this didn't result in the creation of a `Service`
- but the node-collector wanted to send stuff to the activeGate, so the `Service` was needed
- at that point we didn't know if it's ok to "just create the Service" every time, so this hacky solution was added.

gkrenn: Isn't that exactly what I explained above in Adams question 😅 

That's why I created the comment, that the dependency is actually not required. - It's still a dependency though and we just rely on the validation webhook to block if the ActiveGate capability is not there. That's why I would leave the comment in there.

0sewa0: but you don't need the `bool` :D (you can leave the comment if you want)

0sewa0 (approved):

aorcholski (approved):
LGTM

PR #4070, opened 2024-11-18
Simplify GetSettingsForMonitoredEntity and fix timestamp comparision bug incase of multiple MEs

luhi-DT:
Ticket: https://dt-rnd.atlassian.net/browse/DAQ-2355

## Description

We only use GetSettingsForMonitoredEntites with a single entity, so i reworked it to reflect this and renamed it to GetSettingsForMonitoredEntity as this allows us to simplify some code. Also I fixed the bug that was already fixed on the release branch with https://github.com/Dynatrace/dynatrace-operator/pull/4052

## How can this be tested?

Same as in #4052 (Have two MEs for the same clusterID -> check if the one with the bigger timestamp is used for metadata enrichment)


waodim (approved):
lgtm

albertogdd (approved):

PR #4076, opened 2024-11-20
Makes ApiUrl immutable

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-2606)

## Description
Value of dynakube.spec.ApiUrl can't be changed.

```
[]$ kubectl -n dynatrace patch dynakube.dynatrace.com/dynakube  -p '{"spec":{"apiUrl":"https://<modified apiurl>"}}' --type=merge
Error from server (Forbidden): admission webhook "v1beta3.dynakube.webhook.dynatrace.com" denied the request: 
 1. The DynaKube's specification has mutated APIURL, APIURL is immutable, please delete the CR and then apply new one
```

## How can this be tested?

Apply a dynakube. Try to change apiUrl.

### DIFF
--- a/pkg/api/validation/dynakube/api_url.go
+++ b/pkg/api/validation/dynakube/api_url.go
@@ -6,8 +6,6 @@ import (
 	"strings"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
@@ -82,19 +80,8 @@ func IsThirdGenAPIUrl(_ context.Context, _ *Validator, dk *dynakube.DynaKube) st
 	return ""
 }
 
-func IsMutatedApiUrl(ctx context.Context, dv *Validator, dk *dynakube.DynaKube) string {
-	oldDk := &dynakube.DynaKube{}
-	if err := dv.apiReader.Get(ctx, client.ObjectKey{Name: dk.Name, Namespace: dk.Namespace}, oldDk); err != nil {
-		if k8serrors.IsNotFound(err) {
-			return ""
-		}
-
-		log.Info("error occurred while getting dynakube", "err", err.Error())
-
-		return ""
-	}
-
-	if oldDk.Spec.APIURL != dk.Spec.APIURL {
+func IsMutatedApiUrl(_ context.Context, _ *Validator, oldDk *dynakube.DynaKube, newDk *dynakube.DynaKube) string {
+	if oldDk.Spec.APIURL != newDk.Spec.APIURL {
 		return errorMutatedApiUrl
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/api/validation/dynakube/api_url.go
```diff
@@ -20,6 +22,8 @@ const (
 
 	errorThirdGenApiUrl = `The DynaKube's specification has an 3rd gen API URL. Make sure to remove the 'apps' part
 	out of it. Example: ` + ExampleApiUrl
+
+	errorMutatedApiUrl = "The DynaKube's specification has mutated APIURL, APIURL is immutable, please delete the CR and then apply new one"
```
waodim, line 26:
```suggestion
	errorMutatedApiUrl = "The DynaKube's specification mutated the API URL although it is immutable. Please delete the CR and then apply a new one"
```

waodim: just a suggestion. if you don't agree it's also fine for me.

waodim (dismissed):

pkg/api/validation/dynakube/api_url.go
```diff
@@ -77,3 +81,22 @@ func IsThirdGenAPIUrl(_ context.Context, _ *Validator, dk *dynakube.DynaKube) st
...
+	oldDk := &dynakube.DynaKube{}
+	if err := dv.apiReader.Get(ctx, client.ObjectKey{Name: dk.Name, Namespace: dk.Namespace}, oldDk); err != nil {
+		if k8serrors.IsNotFound(err) {
+			return ""
+		}
+
+		log.Info("error occurred while getting dynakube", "err", err.Error())
+
+		return ""
+	}
```
0sewa0, line 95: instead of doing this

try updating the `ValidateUpdate`, as we already get the "oldObject" there, but we ignored it so far.

The "old" way of running the validators is just kinda outdated, as it never checked for Update specific validations
- the current single list of [validatorErrorFuncs](https://github.com/Dynatrace/dynatrace-operator/blob/9c64c66cd898f972407a55f30a42a4211aaaae7f/pkg/api/validation/dynakube/validation.go#L25) is not going to cut it.

we should have logic like:
```go
func (v *Validator) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (warnings admission.Warnings, err error) {
	newDk, err := getDynakube(newObj)
	if err != nil {
		return
	}

	oldDk, err := getDynakube(oldObj)
	if err != nil {
		return
	}

	errMessages := v.runValidators(ctx, validatorErrorFuncs, newDk)
	warnings = v.runValidators(ctx, validatorWarningFuncs, newDk)

	errMessages = append(errMessages, v.runUpdateValidators(ctx, updateValidatorFuncs, oldDk, newDk)) // THIS

	if len(errMessages) != 0 {
		err = errors.New(validation.SumErrors(errMessages, "DynaKube"))
	}

	return
}
```

wepudt (dismissed):
lgtm

pkg/api/validation/dynakube/api_url.go
wepudt: I agree, but should this change be part of this, or should we have a refactoring that unifies this behaviour?

My thinking: for features I want the minimal change that makes this feature work within the current patterns.
If we want to change patterns (and I think we should) let's do that in a different PR.

wdyt?

0sewa0 (approved):

wepudt (approved):
lgtm

PR #4094, opened 2024-11-25
Fix incorrect e2e test in support-archive for new files

andriisoldatenko:
- init
- test
- polish code

### DIFF
--- a/test/features/support_archive/support_archive.go
+++ b/test/features/support_archive/support_archive.go
@@ -195,10 +195,6 @@ func assertFileSize(t *testing.T, requiredFiles []string, zipFile zip.File) []st
 		requiredFiles = slices.Delete(requiredFiles, index, index+1)
 	}
 
-	//// we need to skip fileSize check for specific files
-	//if strings.Contains(zipFileName, "oneagent-logmon-") {
-	//	return requiredFiles
-	//}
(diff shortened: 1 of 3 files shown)


test/features/support_archive/support_archive.go
```diff
@@ -187,7 +187,7 @@ func executeSupportArchiveCommand(ctx context.Context, t *testing.T, envConfig *
 	return executionResult
 }
 
-func assertFile(t *testing.T, requiredFiles []string, zipFile zip.File) []string {
+func assertFileSize(t *testing.T, requiredFiles []string, zipFile zip.File) []string {
```
0sewa0, line 190: This function is just weird, and the rename didn't make it better 😅 
- it checks if the "file is not empty" -> `assertFileSize` reads to me that there is a specific none-zero size we wish to assert for
- `assert...` name is already misleading as the "asserting" part is not it's main/only job, this function will filter out the present files from the `requiredFiles` so later the not-present `requiredFiles` can be logged (and fail the test).

andriisoldatenko: correct, but i don't have better name, and in general I prefer to refactor it.

i'll revert change, and we can discuss how to refactor this part.

0sewa0 (approved):

albertogdd (approved):
Passing in my cluster

PR #4124, opened 2024-11-29
Consider logmonitoring in the node-selector check

0sewa0:
## Description

[DAQ-2191](https://dt-rnd.atlassian.net/browse/DAQ-2191)

Enhances the check previously removed in https://github.com/Dynatrace/dynatrace-operator/pull/4109
- With the caveat, that we use the`apiUrl` for checking if DynaKubes use the same tenant
  - this is not perfect, because of tenant aliases, but it is considered misconfiguration to use a tenant alias for 1 DynaKube and not for another one. 	

How the nodeSelector validation changed:
- 2 CloudNative (or any OneAgent) on same node with different tenant ==> ⛔  (this is not new)
- 2 LogMonitoring on same node with different tenant ==> 🆗 
- 1 LogMonitoring and 1 CloudNative(or any OneAgent) on same node with different tenant ==> 🆗 

## How can this be tested?

Try recreating the scenarios from the unit-tests

[DAQ-2191]: https://dt-rnd.atlassian.net/browse/DAQ-2191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/api/validation/dynakube/logmonitoring_test.go
+++ b/pkg/api/validation/dynakube/logmonitoring_test.go
@@ -11,33 +11,44 @@ import (
 
 func TestIgnoredLogMonitoringTemplate(t *testing.T) {
 	t.Run("no warning if logMonitoring template section is empty", func(t *testing.T) {
-		dk := createStandaloneLogMonitoringDynakube(testName, "")
+		dk := createStandaloneLogMonitoringDynakube(testName, testApiUrl, "")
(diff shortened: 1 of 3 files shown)


waodim (approved):
tested it and works! 👍

PR #4125, opened 2024-11-29
Create kubernetes-monitoring rbac's if only KSPM is enabled

luhi-DT:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-2788

Currently the AG kubernetes-monitoring service account will not exist and validation will prevent KSPM from being enabled, if activeGate rbac is disabled in the helm chart. We should instead create all the resources need if someone wants to only have KSPM enabled in the helm chart, and still allow a Dynakube with a kubeMon AG, to be created.

## How can this be tested?

Install the helm chart with only KSPM rbac enabled (no activeGate), try to apply a Dynakube with KSPM and activeGate with kubernetes Monitoring enabled. Check if AG is started correctly and if SA/cluster roles exist.

### DIFF
--- a/pkg/api/validation/dynakube/modules.go
+++ b/pkg/api/validation/dynakube/modules.go
@@ -25,7 +25,7 @@ func isOneAgentModuleDisabled(_ context.Context, v *Validator, dk *dynakube.Dyna
 }
 
 func isActiveGateModuleDisabled(_ context.Context, v *Validator, dk *dynakube.DynaKube) string {
-	if dk.ActiveGate().IsEnabled() && !v.modules.ActiveGate {
+	if dk.ActiveGate().IsEnabled() && !v.modules.ActiveGate && !v.modules.KSPM {
 		return errorActiveGateModuleDisabled
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/api/validation/dynakube/modules.go
```diff
@@ -25,7 +25,7 @@ func isOneAgentModuleDisabled(_ context.Context, v *Validator, dk *dynakube.Dyna
 }
 
 func isActiveGateModuleDisabled(_ context.Context, v *Validator, dk *dynakube.DynaKube) string {
-	if dk.ActiveGate().IsEnabled() && !v.modules.ActiveGate {
+	if dk.ActiveGate().IsEnabled() && !v.modules.ActiveGate && !v.modules.KSPM {
```
0sewa0, line 28: Bit confusing, but necessary 🥲

0sewa0 (approved):

pkg/api/validation/dynakube/modules.go
waodim: should we abstract it into a method with a meaningful name to avoid confusion? Or would that overdo it?

0sewa0: It's fine as is

PR #4128, opened 2024-11-29
[CRD] Warning if multiple Dyankubes with extensions are enabled for same ApiURL

andriisoldatenko:
## Description

Ticket [DAQ-1617](https://dt-rnd.atlassian.net/browse/DAQ-1617)

Currently having multiple Dynakubes with extensions enabled for the same tenant in the same K8S cluster will have unwanted side-effects. OTOH we don’t want to completely prevent this scenario.

This PR introduce warning validation check for this scenario.

## How can this be tested?

unit tests/ or create 2 dks extensions enabled with same APIURLs and see no-blocking warning:
```
You are already using a Dynakube that enables extensions. Having multiple Dynakubes with '.spec.extensions' enabled can have severe side-effects on “sum” and “count” metrics and cause double-billing.
```

[DAQ-1617]: https://dt-rnd.atlassian.net/browse/DAQ-1617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/api/validation/dynakube/eec.go
+++ b/pkg/api/validation/dynakube/eec.go
@@ -3,11 +3,13 @@ package validation
 import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
 	"golang.org/x/net/context"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
 	errorExtensionExecutionControllerImageNotSpecified       = `DynaKube's specification enables the Prometheus feature, make sure you correctly specify the ExtensionExecutionController image.`
 	errorExtensionExecutionControllerInvalidPVCConfiguration = `DynaKube specifies a PVC for the extension controller while ephemeral volume is also enabled. These settings are mutually exclusive, please choose only one.`
+	warningIfmultiplyDKwithExtensionsEnabled                 = `You are already using a Dynakube that enables extensions. Having multiple Dynakubes with '.spec.extensions' enabled can have severe side-effects on “sum” and “count” metrics and cause double-billing.`
 )
 
 func extensionControllerImage(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
@@ -19,6 +21,31 @@ func extensionControllerImage(_ context.Context, _ *Validator, dk *dynakube.Dyna
 		log.Info("requested dynakube doesn't specify the ExtensionExecutionController image.", "name", dk.Name, "namespace", dk.Namespace)
 
 		return errorExtensionExecutionControllerImageNotSpecified
+	}
+
+	return ""
+}
+
+func warnIfmultiplyDKwithExtensionsEnabled(ctx context.Context, dv *Validator, dk *dynakube.DynaKube) string {
+	if !dk.IsExtensionsEnabled() {
+		return ""
+	}
+
+	validDynakubes := &dynakube.DynaKubeList{}
+	if err := dv.apiReader.List(ctx, validDynakubes, &client.ListOptions{Namespace: dk.Namespace}); err != nil {
+		log.Info("error occurred while listing dynakubes", "err", err.Error())
+
+		return ""
+	}
+
+	for _, item := range validDynakubes.Items {
+		if item.Name == dk.Name {
+			continue
+		}
+
+		if item.IsExtensionsEnabled() && (dk.ApiUrl() == item.ApiUrl()) {
+			return warningIfmultiplyDKwithExtensionsEnabled
+		}
 	}
 
 	return ""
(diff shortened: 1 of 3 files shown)


StefanHauth (approved):
LGTM

pkg/api/validation/dynakube/eec.go
```diff
@@ -24,6 +26,31 @@ func extensionControllerImage(_ context.Context, _ *Validator, dk *dynakube.Dyna
 	return ""
 }
 
+func warnIfmultiplyDKwithExtensionsEnabled(ctx context.Context, dv *Validator, dk *dynakube.DynaKube) string {
```
0sewa0, line 29:
```suggestion
func conflictingTenantForExtensions(ctx context.Context, dv *Validator, dk *dynakube.DynaKube) string {
```

pkg/api/validation/dynakube/eec.go
```diff
@@ -3,11 +3,13 @@ package validation
 import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
 	"golang.org/x/net/context"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
 	errorExtensionExecutionControllerImageNotSpecified       = `DynaKube's specification enables the Prometheus feature, make sure you correctly specify the ExtensionExecutionController image.`
 	errorExtensionExecutionControllerInvalidPVCConfiguration = `DynaKube specifies a PVC for the extension controller while ephemeral volume is also enabled. These settings are mutually exclusive, please choose only one.`
+	warningIfmultiplyDKwithExtensionsEnabled                 = `You are already using a Dynakube that enables extensions. Having multiple Dynakubes with '.spec.extensions' enabled can have severe side-effects on “sum” and “count” metrics and cause double-billing.`
```
0sewa0, line 12:
```suggestion
	warningConflictingTenantForExtensions                 = `You are already using a Dynakube that enables extensions. Having multiple Dynakubes with '.spec.extensions' enabled can have severe side-effects on “sum” and “count” metrics and cause double-billing.`
```

pkg/api/validation/dynakube/eec.go
0sewa0: you could also call it `conflictingApiUrlForExtensions`, to be more precious

pkg/api/validation/dynakube/eec.go
0sewa0: I like the message, but it doesn't mention that the problems is that both DynaKube use the same tenant (`apiUrl`), should we mention that for clarity? (if we mention it, then I would stick to calling out the `apiUrl` being the same, as maybe a customer will be not familiar with the "tenant" naming)

pkg/api/validation/dynakube/eec.go
0sewa0: [(based it on already existing validation func names)](https://github.com/Dynatrace/dynatrace-operator/blob/f85df8f8f7fb174496f39f32bc94cba84c8656cd/pkg/api/validation/dynakube/validation.go#L37-L42)

0sewa0 (commented):
Logic looks good, just some naming things

pkg/api/validation/dynakube/eec.go
andriisoldatenko: @StefanHauth wdyt?

pkg/api/validation/dynakube/eec.go
andriisoldatenko: on it!

pkg/api/validation/dynakube/eec.go
andriisoldatenko: renamed in bab9b2f5e

StefanHauth: Good catch, that should definitely be in the message. Also the name of the already existing DK should be in there.

```
`You are already using a Dynakube ('<name of existing DK>') that enables extensions. Having multiple Dynakubes with same '.spec.apiUrl' and '.spec.extensions' enabled can have severe side-effects on “sum” and “count” metrics and cause double-billing.`
```

andriisoldatenko: fixed in 40c4c4e5d531e55fe236236dad3b487658bffcfd

andriisoldatenko: @0sewa0, @StefanHauth  PTAL!

0sewa0 (approved):

andriisoldatenko:
tested in my cluster:
```
kaf ~/work/samples/dk-eec.yaml
Warning: You are already using a Dynakube ('dynakube') that enables extensions. Having multiple Dynakubes with same '.spec.apiUrl' and '.spec.extensions' enabled can have severe side-effects on “sum” and “count” metrics and cause double-billing.
dynakube.dynatrace.com/dynakube1 created
```

PR #4131, opened 2024-11-29
Recreate EEC statefulset for PVC changes only

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-2893)

## Description

Hash value calculated for desired statefulset.Spec.VolumeClaimTemplates is preserved as an annotation in Annotations map.

Comparing preserved value with value calculated for new desired statefulset we can make a decision to update or recreate EEC statefulset (PVC definition was changed or not).

It's hard to compare statefulset.PVC manifests directly because for example K8S fills in PVC.VolumeMode field so Applied and then Get manifest are never equal.

## How can this be tested?
1) unittest
```
make go/test
```
2) deploy and keep updating dynakube changing EEC-PVC related properties
`useEphemeralVolume` = true/false
`tag` = 1.305.. / 1.306..
`persistentVolumeClaim` = defined / not defined 

EEC statefulset should be recreated only if `useEphemeralVolume` or `persistentVolumeClaim` is changed.
```
  templates:
    extensionExecutionController:
      #useEphemeralVolume: true
      useEphemeralVolume: false
      imageRef:
        repository: ....
        #tag: 1.306.0.20241128-144422
        tag: 1.305.0.20241029-011002
      #persistentVolumeClaim:
      #  accessModes:
      #  - ReadWriteOnce
      #  storageClassName: "standard"
      #  resources:
      #    requests:
      #      storage: 2G 
```

### DIFF
--- a/pkg/util/kubeobjects/statefulset/query.go
+++ b/pkg/util/kubeobjects/statefulset/query.go
@@ -1,8 +1,6 @@
 package statefulset
 
 import (
-	"reflect"
-
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/hasher"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/internal/query"
@@ -10,6 +8,8 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+type MustRecreateFunc func(current, desired *appsv1.StatefulSet) bool
 
 func Query(kubeClient client.Client, kubeReader client.Reader, log logd.Logger) query.Generic[*appsv1.StatefulSet, *appsv1.StatefulSetList] {
 	return query.Generic[*appsv1.StatefulSet, *appsv1.StatefulSetList]{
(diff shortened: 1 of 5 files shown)


pkg/util/kubeobjects/statefulset/query.go
```diff
@@ -32,10 +32,33 @@ func Query(kubeClient client.Client, kubeReader client.Reader, log logd.Logger)
 	}
 }
 
+func RecreateQuery(kubeClient client.Client, kubeReader client.Reader, log logd.Logger, customMustRecreate MustRecreateFunc) query.Generic[*appsv1.StatefulSet, *appsv1.StatefulSetList] {
+	return query.Generic[*appsv1.StatefulSet, *appsv1.StatefulSetList]{
```
StefanHauth, line 36: There's lot's of duplicate code. Maybe this can go into a common function that accepts a "recreate" function.

```
func query(kubeClient client.Client, kubeReader client.Reader, log logd.Logger, mustRecreate MustRecreateFunc) ... {
  return query.Generic[*appsv1.StatefulSet, *appsv1.StatefulSetList]{ 
    ... 
    MustRecreate: mustRecreate
    ...
  }
}

func RecreateQuery(...) ... {
  recreateFunc := func(current, desired *appsv1.StatefulSet) bool {
			return mustRecreate(current, desired) || customMustRecreate(current, desired)
		}
  return query(..., recreateFunc)
}

func Query(...) ... {
  return query(..., mustRecreate)
}
```

andriisoldatenko (approved):

PR #4140, opened 2024-12-02
Add logmonitoring rbac to OA ServiceAccount

0sewa0:
## Description
[DAQ-2891](https://dt-rnd.atlassian.net/browse/DAQ-2891)

Changes:
- The `dynatrace-logmonitoring` `ClusterRole` (and binding) is always created not only on Openshift
- LogMonitoring can work in the fullstack OneAgent `DaemonSet`
   - The `dynatrace-logmonitoring` `ClusterRole` is also bound to the OneAgent `ServiceAccount`(`dynatrace-dynakube-oneagent`)
   - The LogMonitoring related Envs are also added to the fullstack OneAgent `DaemonSet`

## How can this be tested?

Deploy LogMonitoring (standalone and with fullstack):
- The `dynatrace-logmonitoring` `ClusterRole` should always exist and be bound to ServiceAccounts
- The API envs (`KUBELET_API_NODENAME`, `KUBELET_API_ADRESS`) should always be present if LogMonitoring is enabled

### DIFF
--- a/config/helm/chart/default/templates/Common/logmonitoring/clusterrole-logmonitoring.yaml
+++ b/config/helm/chart/default/templates/Common/logmonitoring/clusterrole-logmonitoring.yaml
@@ -1,4 +1,4 @@
-{{- if and .Values.rbac.logMonitoring.create (eq (include "dynatrace-operator.openshiftOrOlm" .) "true") }}
+{{- if .Values.rbac.logMonitoring.create }}
 # Copyright 2021 Dynatrace LLC
 
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -48,4 +48,21 @@ subjects:
 - kind: ServiceAccount
   name: dynatrace-logmonitoring
   namespace: {{ .Release.Namespace }}
+{{ if .Values.rbac.oneAgent.create }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: dynatrace-logmonitoring-fullstack
+  labels:
+    {{- include "dynatrace-operator.logMonitoringLabels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: dynatrace-logmonitoring
+subjects:
+- kind: ServiceAccount
+  name: dynatrace-dynakube-oneagent
+  namespace: {{ .Release.Namespace }}
+{{ end }}
 {{ end }}
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/env.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/env.go
@@ -17,8 +17,8 @@ const (
 	entityEnv        = "DT_ENTITY_KUBERNETES_CLUSTER"
 
 	// main container envs
-	apiNodeNameEnv  = "KUBELET_API_NODENAME"
-	apiIPAddressEnv = "KUBELET_API_ADDRESS"
+	ApiNodeNameEnv  = "KUBELET_API_NODENAME"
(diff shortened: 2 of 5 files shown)


pkg/controllers/dynakube/logmonitoring/daemonset/env.go
```diff
@@ -17,8 +17,8 @@ const (
 	entityEnv        = "DT_ENTITY_KUBERNETES_CLUSTER"
 
 	// main container envs
-	apiNodeNameEnv  = "KUBELET_API_NODENAME"
-	apiIPAddressEnv = "KUBELET_API_ADDRESS"
+	ApiNodeNameEnv  = "KUBELET_API_NODENAME"
```
andriisoldatenko, line 20:
```suggestion
	APInodeNameEnv  = "KUBELET_API_NODENAME"
```

andriisoldatenko: API is abbreviation.

0sewa0: if I do that then I have to do `APIIPAddressEnv` which is horrible (or more horrible than `ApiIPAddressEnv` 😛 )

andriisoldatenko: nope, `APIipAddressEnv `

0sewa0: that is not better, maybe even worse xd

chrismuellner: just call it `kubeletNodeName` then

andriisoldatenko: https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)?redirectedfrom=MSDN

>When using acronyms, use Pascal case or camel case for acronyms more than two characters long. For example, use HtmlButton or htmlButton. However, you should capitalize acronyms that consist of only two characters, such as System.IO instead of System.Io.

I'm okey to rename it as @chrismuellner  suggested.

0sewa0: that is a C# thing, at that point we could just use snake_case (which is the best IMO 😉 )

(but I will rename it)

andriisoldatenko: i know i know :) just posted as alternative opinion.

config/helm/chart/default/templates/Common/logmonitoring/clusterrole-logmonitoring.yaml
```diff
@@ -48,4 +48,21 @@ subjects:
 - kind: ServiceAccount
   name: dynatrace-logmonitoring
   namespace: {{ .Release.Namespace }}
+{{ if .Values.rbac.oneAgent.create }}
```
chrismuellner, line 51: this will work, but maybe we should start thinking about separate files soon

0sewa0: I did it like that, because it felt more "natural" than a separate file with a double `if` (ie.: with an `if and` 😅 )

pkg/controllers/dynakube/logmonitoring/daemonset/env.go
0sewa0: [fixed](https://github.com/Dynatrace/dynatrace-operator/pull/4140/commits/2acd00c27cb0287b355e80d2bf478fcc9d5a8582)

waodim (approved):
Tested it and it works. code also lgtm

chrismuellner (approved):

PR #4146, opened 2024-12-03
Feat/add ff for oa livenessprobe

luhi-DT:
## Description

Currently it can happen, that the agent is not getting ready in time (20 mins) and therefore is being killed by the livenessprobe, as the installer is still running. For such situations to still allow agent startup, we should introduce a feature flag to skip the probe.

## How can this be tested?

Create a dynakube with the feature flag `feature. dynatrace. com/oneagnet-skip-liveness-probe` set to true, rollout an agent daemonset, check if the daemonset has no livenessprobe configured


pkg/api/v1beta3/dynakube/feature_flags.go
```diff
@@ -58,6 +58,7 @@ const (
 	AnnotationFeatureOneAgentMaxUnavailable         = AnnotationFeaturePrefix + "oneagent-max-unavailable"
 	AnnotationFeatureOneAgentInitialConnectRetry    = AnnotationFeaturePrefix + "oneagent-initial-connect-retry-ms"
 	AnnotationFeatureRunOneAgentContainerPrivileged = AnnotationFeaturePrefix + "oneagent-privileged"
+	AnnotationFeatureOneAgentSkipLivenessProbe      = AnnotationFeaturePrefix + "oneagnet-skip-liveness-probe"
```
0sewa0, line 61:
```suggestion
	AnnotationFeatureOneAgentSkipLivenessProbe      = AnnotationFeaturePrefix + "oneagent-skip-liveness-probe"
```

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -224,8 +224,11 @@ func (b *builder) podSpec() (corev1.PodSpec, error) {
 		TerminationGracePeriodSeconds: address.Of(defaultTerminationGracePeriod),
 	}
 
-	if b.dk.NeedsOneAgentProbe() {
+	if b.dk.NeedsOneAgentReadinessProbe() {
 		podSpec.Containers[0].ReadinessProbe = b.getReadinessProbe()
+	}
+
+	if b.dk.NeedsOneAgentLivenessProbe() {
```
andriisoldatenko, line 231: can you please extend unit-tests to cover new condition?
```
func TestPodSpecProbes(t *testing.T) {
```

0sewa0 (changes requested):
the title of the PR is a bit weird 😅

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
luhi-DT: on it :)

pkg/api/v1beta3/dynakube/feature_flags.go
luhi-DT: done, thx

0sewa0 (approved):

PR #4171, opened 2024-12-06
Implement app volume migration logic

0sewa0:
## Description
[DAQ-1407](https://dt-rnd.atlassian.net/browse/DAQ-1407)

The main change:
[Implement volume migration logic](https://github.com/Dynatrace/dynatrace-operator/commit/a417719197f9d350a2c430294d9873ae9ea46ec9) 
	- Removes testing code that is sqlite specific, because it will be removed later either way
	- Update the entrypoint of the csi-init (the command that does  the migration)
	- Removed the "migration of deprecated binaries" as that have been introduce several versions ago (0.13) and we don't support skipping releases during upgrades.

For details about the other changes check the commits.

### Prep
1. Deploy an older Operator version (`main` is also fine)
2. Deploy a DynaKube (cloudnative)
3. Deploy and inject into sample apps

⚠️  Sidenote: It is preferable that you have a clean `/data/.../run` folder, so you see what is happening better

### Test
1. Upgrade to this Operator
2. In the csi-driver `server` check that `/data/appmounts` now has something in it, and they are pointing to `/data/.../run/...`
   - (Migration happened) 
3. Remove the sample apps, make sure they terminate "peacefully" (so don't force delete them 😅 )
4. Check that the `server` has no errors
5. Check that the `mount | grep "overlay on /data"` command in the server gives you nothing back
   - (Managed to unmount)
   - ⚠️ Sidenote: if you run  `mount | grep "overlay on /csi"`, that will give you back entries, but that is a [bug](https://dt-rnd.atlassian.net/browse/DAQ-3049) I can't fix here, just mitigate.
6. Check that the`/data/.../run` folders ONLY have their `mapped` folder remain:
   - this is yet again because of the [bug](https://dt-rnd.atlassian.net/browse/DAQ-3049)
7. If you deploy the sample apps again
   - You should only see new entries in `/data/appmounts`
   - These should fully go away, if you remove the sample apps (considering you didn't restart the CSI-driver in the meantime)

[DAQ-1407]: https://dt-rnd.atlassian.net/browse/DAQ-1407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andriisoldatenko (approved):
tested in my cluster + run some standard e2e tests

PR #4174, opened 2024-12-09
Change logMonitoring error handling in case of no kubernetesClusterMEID

waodim:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-3138

In case we have no `kuberentesClusterMEID` and kubernetes monitoring is not enabled, we error with a stacktrace, which pollutes the logs a lot.

With this I changed the way we behave: We only log that the missing `kubernetesClusterMEID` is the problem and and then return nil, as this is expected to not work.

## How can this be tested?

Deploy operator, make sure that your cluster has no `kubernetesClusterMEID`, enable `logMonitoring` and **no** kubernetes monitoring. Have a look at the dk status and the logs


pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
```diff
@@ -79,7 +79,16 @@ func (r *reconciler) checkLogMonitoringSettings(ctx context.Context) error {
 	if err != nil {
 		setLogMonitoringSettingError(r.dk.Conditions(), conditionType, err.Error())
 
-		return errors.WithMessage(err, "error when creating log monitoring setting")
+		if !r.dk.ActiveGate().IsKubernetesMonitoringEnabled() && r.dk.Status.KubernetesClusterMEID == "" {
```
0sewa0, line 82: IMO you should just simply check [in the beginning](https://github.com/Dynatrace/dynatrace-operator/blob/0b6cc5e10b1140eedc7385cfca607032e06ea8ad/pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go#L40) that  `r.dk.Status.KubernetesClusterMEID != ""` otherwise do nothing.

Your request will never go through if `r.dk.Status.KubernetesClusterMEID == ""`

0sewa0: you can/should still log something in that case

waodim: I agree this is nicer. 

Do you mean by 

> you can/should still log something in that case

to add a log message [here](https://github.com/Dynatrace/dynatrace-operator/blob/0b6cc5e10b1140eedc7385cfca607032e06ea8ad/pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go#L42)? Or am I wrong?

waodim: proposed solution: https://github.com/Dynatrace/dynatrace-operator/pull/4174/commits/f4c3bb120f1bcd3de07004fae437f1f56236ea12

0sewa0 (approved):
for now we will do it like this for consistency (the DaemonSet reconciler also errors if the MEID is missing)
we are still looking into handling this in a different way

PR #4190, opened 2024-12-11
Remove the db from the provisioner logic (and codebase)

0sewa0:
## Description
[DAQ-1412](https://dt-rnd.atlassian.net/browse/DAQ-1412)
[DAQ-1471](https://dt-rnd.atlassian.net/browse/DAQ-1471)

Removes the database code from the `provisioner` logic, and refactors it a bit, to hopefully make it more understandable.

Also removes the database code from the whole codebase:
- The `provisioner` was the last thing still using it
   - The linter complained after I removed the database code from there
     -   It was easier to remove the whole code than messing with the linter

## How can this be tested?

The provisioner refactor:
- It should work as normal, just no more database creation
- Look at code, see if it's clear

The database-code/CGO removal:
- Image still builds, but faster 😅 

[DAQ-1412]: https://dt-rnd.atlassian.net/browse/DAQ-1412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DAQ-1471]: https://dt-rnd.atlassian.net/browse/DAQ-1471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controllers/csi/provisioner/install.go
```diff
@@ -3,118 +3,119 @@ package csiprovisioner
...
+		}
+
+		imageInstaller, err := provisioner.imageInstallerBuilder(ctx, provisioner.fs, props)
+		if err != nil {
+			return nil, err
+		}
+
+		return imageInstaller, nil
+	default:
+		return nil, errors.New("not enough information to download CodeModule")
```
chrismuellner, line 79: what information is missing?

pkg/controllers/csi/provisioner/install.go
```diff
@@ -3,118 +3,119 @@ package csiprovisioner
...
-		eventRecorder.sendFailedInstallAgentVersionEvent(targetVersion, dk.GetName())
+}
 
-		return err
+func (provisioner *OneAgentProvisioner) getTargetDir(dk dynakube.DynaKube) string {
+	var dirName string
+
+	switch {
+	case dk.CodeModulesImage() != "":
+		// An image URI often contains one or several /-s, which is problematic when trying to use it as a folder name.
```
chrismuellner, line 88: replace `/-s` with `slashes`?

pkg/controllers/csi/provisioner/install.go
```diff
@@ -3,118 +3,119 @@ package csiprovisioner
...
+		// Easiest to just base64 encode it
+		dirName = base64.StdEncoding.EncodeToString([]byte(dk.CodeModulesImage()))
+	case dk.CodeModulesVersion() != "":
+		dirName = dk.CodeModulesVersion()
+	default:
+		dirName = "unknown"
```
chrismuellner, line 94: will this cause issues in case the same directory name is reused for different runs?

pkg/controllers/csi/provisioner/controller.go
```diff
@@ -53,28 +49,20 @@ type imageInstallerBuilder func(context.Context, afero.Fs, *image.Properties) (i
...
 
 // NewOneAgentProvisioner returns a new OneAgentProvisioner
-func NewOneAgentProvisioner(mgr manager.Manager, opts dtcsi.CSIOptions, db metadata.Access) *OneAgentProvisioner {
+func NewOneAgentProvisioner(mgr manager.Manager, opts dtcsi.CSIOptions) *OneAgentProvisioner {
 	return &OneAgentProvisioner{
-		client:                 mgr.GetClient(),
 		apiReader:              mgr.GetAPIReader(),
-		opts:                   opts,
 		fs:                     afero.NewOsFs(),
-		recorder:               mgr.GetEventRecorderFor("OneAgentProvisioner"),
```
chrismuellner, line 76: why do we completely remove the events? shouldn't we try to make them more useful or at least keep the same ones instead?

pkg/controllers/csi/provisioner/install.go
0sewa0: will update 👍

pkg/controllers/csi/provisioner/controller.go
0sewa0: IMO they are half-baked as of right now, I don't believe they are documented in any way, I have yet to see them being referenced in an RFA (or anywhere)

I would remove them for simplicity's sake, and add them back with more thought put into it.
- This can be done in context of the database removal

andriisoldatenko:
just an idea or rather a question: do we want to have some kind of cleanup script to remove old sqlite.db file? or it will be trashed anyway?

0sewa0:
> just an idea or rather a question: do we want to have some kind of cleanup script to remove old sqlite.db file? or it will be trashed anyway?

it will be part of the GC [DAQ-1383](https://dt-rnd.atlassian.net/browse/DAQ-1383) (follow up story)

pkg/injection/codemodule/installer/symlink/symlink.go
```diff
@@ -74,7 +74,7 @@ func Remove(fs afero.Fs, symlinkPath string) error {
 	if info, _ := fs.Stat(symlinkPath); info != nil {
 		log.Info("symlink to codemodule directory exists, removing it due to the possibility of the agent being installed again")
 
-		if err := fs.RemoveAll(symlinkPath); err != nil {
+		if err := fs.Remove(symlinkPath); err != nil {
```
waodim, line 77: Why the change here? Isn't `RemoveAll` better because of potential contents of the directory?

0sewa0: Do not remove what the symlink is pointing to, that is for the GC to do, you have to be sure that nothing is using it

waodim:
codewise I do not have anything to add. LGTM

It also works for me. Building is faster and csi still works as it should.

If all conversations are resolved, I will approve.

pkg/injection/codemodule/installer/symlink/symlink.go
gkrenn: Be aware that `RemoveAll` would not have failed if the file does not exist, `Remove` does - As we check in the if above, it should be fine 👍

Dockerfile
```diff
@@ -19,7 +19,7 @@ ARG GO_BUILD_TAGS
 
 COPY pkg ./pkg
 COPY cmd ./cmd
```
gkrenn, line 21: The `apt-get` commands can be removed too

pkg/controllers/csi/provisioner/install.go
gkrenn: What about panicing? - As this should anyway never happen, no need to obscure the error

pkg/controllers/csi/provisioner/controller.go
```diff
@@ -92,187 +80,58 @@ func (provisioner *OneAgentProvisioner) SetupWithManager(mgr ctrl.Manager) error
...
+
+	err := provisioner.apiReader.Get(ctx, request.NamespacedName, &dk)
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
-			return reconcile.Result{}, provisioner.db.DeleteDynakube(ctx, request.Name)
+			return reconcile.Result{}, nil // TODO: Call GC here
 		}
 
 		return reconcile.Result{}, err
 	}
```
gkrenn, line 93: We no longer check if the download of the package is required. Even in appliation monitoring mode without csi driver we download the binaries.

gkrenn (changes requested):

pkg/controllers/csi/provisioner/controller.go
0sewa0: this is not really a problem, as in `main` we use the CSI-driver if it is there
the feature-branch is not up to date in this respect

pkg/controllers/csi/provisioner/install.go
0sewa0: I definitely don't want to panic 😅 
will just return error

0sewa0: fixed

Dockerfile
0sewa0: fixed

pkg/controllers/csi/provisioner/controller.go
gkrenn: Don't we still have the `useCSIDriver` field and want to support it, even though the CSI driver is installed? 🤔

Fine for me then

gkrenn (approved):

pkg/controllers/csi/provisioner/controller.go
0sewa0: not in main (already in 1.4)

The "rebase" will be fuuuun 😅

PR #4215, opened 2024-12-17
Add pod info based symlink during app mount

waodim:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-1458

With this we now introduce the following:

- we add the pod's namespace to the `volumeConfig`, it will be used in the publishing of the volume
- a `symlink` that points to `/data/appmounts/<volume-id>` from `/data/<dk>/appmounts/<pod-namespace>/<pod-name>`.

## How can this be tested?

Deploy operator and set up some sample pods to inject into them. Have a look into the csi driver logs, and also into the container. Run this command: `ls -lR /data | grep '^l'` to see all `symlinks` inside the container and you should see something like this:

`lrwxrwxrwx 1 root root 84 Dec 17 12:39 demo-nginx-7b54d6884d-wvff5 -> /data/appmounts/csi-434efd556407e51d29d40f95226757193a6835327fcd6fd10f49226e88784439`

Which shows that the `symlink` is there and points to the `appmount` directory.

### DIFF
--- a/pkg/injection/codemodule/installer/symlink/symlink.go
+++ b/pkg/injection/codemodule/installer/symlink/symlink.go
@@ -70,9 +70,29 @@ func CreateForLatestVersion(fs afero.Fs, dk dynakube.DynaKube, latestVersionDir,
 	return nil
 }
 
+func CreateForPodInfoDir(fs afero.Fs, targetDir, symlinkDir string) error {
+	// MemMapFs (used for testing) doesn't comply with the Linker interface
+	linker, ok := fs.(afero.Linker)
+	if !ok {
+		log.Info("symlinking not possible", "targetDir", targetDir, "fs", fs)
+
+		return nil
+	}
+
+	log.Info("creating symlink based on pod info", "points-to(relative)", targetDir, "location", symlinkDir)
+
+	if err := linker.SymlinkIfPossible(targetDir, symlinkDir); err != nil {
+		log.Info("symlinking failed", "podinfodir", targetDir)
+
+		return errors.WithStack(err)
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/injection/codemodule/installer/symlink/symlink.go
```diff
@@ -70,9 +70,29 @@ func CreateForLatestVersion(fs afero.Fs, dk dynakube.DynaKube, latestVersionDir,
 	return nil
 }
 
+func CreateForPodInfoDir(fs afero.Fs, targetDir, symlinkDir string) error {
```
0sewa0, line 73: Can't we have 1 function like `Create`?

The only difference is the log lines, and only a bit.
If we log the paths we wanted to use for the symlink creation, we can figure out which symlink it is.

waodim: fair enough, on it

waodim: have a look for yourself, maybe it is overkill but at least we only have one function 😅 

https://github.com/Dynatrace/dynatrace-operator/pull/4215/commits/c99035075c8164702a9f71eb830f555ec3d83b6b

waodim: went with option b. as it seems cleaner to me.

I tested it but if you find time I would appreciate if you could test it as well

https://github.com/Dynatrace/dynatrace-operator/pull/4215/commits/8b10eaf288a9f5cd3272bb5b110df59897929fe2

0sewa0 (approved):

PR #4220, opened 2024-12-18
Remove tenant image registry restriction for image fields

0sewa0:
## Description

[DAQ-3502](https://dt-rnd.atlassian.net/browse/DAQ-3502)

The `Dynatrace server registry not allowed` part of the validation was causing problems with URIs that were in the format of `ip:port`. (look at the added unit-tests to see examples)

The current decision is to remove this part of the check, as the `url.Parse` part of the check was not fully correct (example: `my.images.com/test:tag` has no `.Host` field according to it) and does not fully work with correct image URIs (example: `127.0.0.1:5000/test:tag`). Furthermore, this is crude check, if a customer replicates the (incorrect) image from the tenant-registry this check will not protect them. 
- "Incorrect" image: The tenant-registry only has `amd64` images, but in case of `customImage` we set the node affinity to be broader (`arm64` etc) as the public-registry images support them and we expect the customer to replicate these in their private registries.

## How can this be tested?

deploy a DynaKube where the `image` field is set to `127.0.0.1:5000/test:tag` (or similar)

[DAQ-3502]: https://dt-rnd.atlassian.net/browse/DAQ-3502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/api/validation/dynakube/image.go
```diff
@@ -47,18 +42,12 @@ func imageFieldHasTenantImage(_ context.Context, _ *Validator, dk *dynakube.Dyna
...
 	if image != "" {
-		ref, err := name.ParseReference(image)
+		_, err := name.ParseReference(image)
 		if err != nil {
 			return fmt.Sprintf(errorUnparsableImageRef, section)
 		}
-
-		refUrl, _ := url.Parse(ref.Name())
-
-		if refUrl.Host == disallowedHost {
```
andriisoldatenko, line 59: just to make sure we all aware that we open gate for some security problems in future :) 
cc @aorcholski

0sewa0: what security problems?

andriisoldatenko: i mean you can use any registry which you want and we will pull from it and try to run it :) 

it's only potential problem :) nothing against implementation.

0sewa0: > i mean you can use any registry which you want and we will pull from it and try to run it :)

you could already do this, the only registry you couldn't use is the official Dynatrace Environment's registry ([look at how it was used](https://github.com/Dynatrace/dynatrace-operator/pull/4220/files#diff-a003644eb8d11e98c18628b0ffbdbd45e53e3a35ed4783c9985e52b1a794babbL41))

andriisoldatenko: oh i misread code, my bad.

andriisoldatenko (approved):

aorcholski (approved):
We could add a scheme if missing to avoid some issues of `url.Parse`. See [here](https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/controllers/dynakube/proxy/reconciler.go#L116)

PR #4231, opened 2024-12-19
Move install-config from `envFrom` to a direct env via templates

0sewa0:
## Description

[DAQ-3474](https://dt-rnd.atlassian.net/browse/DAQ-3474)

Problems with `envFrom`:
- Changes to the `ConfigMap` referenced is not "updated" in the env of the container.
- Changes in the `ConfigMap` will not restart the pod, as the yaml does not change

This results in:
- For the `ConfigMap`'s content be taken into account, the workloads need to be restarted manually
  - This would be needed if certain features would be turned on/off via manifests (example: CSI)

Fix:
- Directly put it into the `env` of the container
- Use a template for this, so we don't have to repeat ourselves too much

## How can this be tested?
1. Check that the manifests contain the env correctly
2. Check that changes in the `values.yaml` (`rbac` section) are reflected correctly in the yamls.
3. Check that the Webhook/Operator/CSI driver takes the env into account

[DAQ-3474]: https://dt-rnd.atlassian.net/browse/DAQ-3474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/helm/chart/default/templates/Common/csi/daemonset.yaml
```diff
@@ -170,9 +170,7 @@ spec:
           - name: MAX_UNMOUNTED_VOLUME_AGE
```
waodim, line 170: Looks good to me. It works for me, one only has to wait until webhook pods get restarted. Should we document that somewhere? I am thinking of the case that someone applies `kubernetes.yaml`, then immediately after `kubernetes-csi.yaml` and then immediately after the respective dynakube, one still gets this error until the webhook restarted. Is this a realistic problem, such that we need to give the advise, that one has to at least wait for the webhook pods?

waodim: Apart from my question I would approve.

Another thing. Do you think we run into problems when keeping the env var in lower case?

waodim (commented):
edit: see comments below

config/helm/chart/default/templates/Common/csi/daemonset.yaml
0sewa0: Great question, could you write that on the ticket so PO/PM can think about it, as that will not change the implementation.

But this will not fix the problem of the scripts that are applying `kubernets.yaml` and `...-csi.yaml` in sequence followed by some wait.
It will fix the fact that the user will not have to uninstall the Operator to fix the issue.

waodim: Yeah sure, will do 👍

waodim (approved):

gkrenn (approved):

PR #4236, opened 2024-12-19
Add telemetryService field to the DynakubeSpec

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-3214)

## Description
Adds `telemetryService` field to the Dynakube spec.

The default list of protocols is used if empty `telemetryService` or empty `protocols` list is provided in the dynakube spec:
```
telemetryService: {}
```
```
telemetryService:
    protocols:
```  
The default list of protocols contains:
- otlp
- zipkin
- jaeger
- statsd

## How can this be tested?
1) unittests

2) Apply dynakube
```
spec:
  telemetryService: {}
```  
ActiveGate pod is not created.

3) Apply dynakube
```
spec:
  activeGate:
    capabilities:
    - routing
  telemetryService: {}
```  
ActiveGate pod is created with PV/PVC.

`kubectl -n dynatrace describe pod/dynakube-activegate-0`
```
DT_CAPABILITIES:            MSGrouter,log_analytics_collector,generic_ingest_enabled,otlp_ingest
```


pkg/api/v1beta3/dynakube/telemetryservice_props.go
```diff
@@ -0,0 +1,33 @@
+package dynakube
+
+const (
+	TelemetryServiceOtlpProtocol   = "otlp"
```
andriisoldatenko, line 4: you can use golang-like enums (similar what k8s does)
```
type Protocol string

type Protocols []Protocol

const (
        JagerProtocol  Protocol = "jager"
        ZipkinProtocol Protocol = "zipkin"
        OtlpProtocol   Protocol = "otlp"
        StatsdProtocol Protocol = "statsd"
)
```

andriisoldatenko: or https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1524-L1534

pkg/api/v1beta3/dynakube/telemetryservice_props.go
```diff
@@ -0,0 +1,39 @@
+package dynakube
+
+type TelemetryServiceProtocol string
+
+const (
+	TelemetryServiceOtlpProtocol   TelemetryServiceProtocol = "otlp"
```
andriisoldatenko, line 6: it's hard to read, should we name it <prtocol_name><type>: eg. `OtlpTelemetryServiceProtocol `

0sewa0 (changes requested):

pkg/api/validation/dynakube/telemetryservice.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+
+	var duplicatedProtocols []string
+
+	for protocol, count := range protocolsOccurrences {
+		if count > 1 {
+			duplicatedProtocols = append(duplicatedProtocols, string(protocol))
+		}
+	}
```
andriisoldatenko, line 74: these all can me simplified:
```
protocols := copy(dk.TelemetryServiceProtocols()) // or tmp := append([]int(nil), dk.TelemetryServiceProtocols()...)
slices.Sort(protocols)
if len(slices.Compact(protocols))) != len(dk.TelemetryServiceProtocols()) {return error}
```
https://pkg.go.dev/slices#Compact

pkg/api/v1beta3/dynakube/dynakube_types.go
```diff
@@ -86,6 +86,11 @@ type DynaKubeSpec struct { //nolint:revive
 	// +kubebuilder:validation:Optional
 	Extensions *ExtensionsSpec `json:"extensions,omitempty"`
 
+	// When an (empty) TelemetryServiceSpec is provided, the OpenSignal related components (OpenSignal collector)
+	// is deployed by the operator.
```
waodim, line 90:
```suggestion
	// are deployed by the operator.
```

waodim: nitpick, feel free to ignore

pkg/api/validation/dynakube/telemetryservice.go
aorcholski: true but then we can't tell which protocols are duplicated

andriisoldatenko: do you really need to report it back to user? or just generic message would be enough?

pkg/api/v1beta3/dynakube/telemetryservice_props.go
```diff
@@ -0,0 +1,23 @@
...
+	var protocols []telemetryservice.Protocol
+	for _, proto := range dk.Spec.TelemetryService.Protocols {
+		protocols = append(protocols, telemetryservice.Protocol(proto))
+	}
+	return protocols
+}
+
+func (dk *DynaKube) IsTelemetryServiceEnabled() bool {
+	return dk.Spec.TelemetryService != nil
+}
```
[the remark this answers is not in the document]
0sewa0: so you can go `dk.TelemetryService().IsEnabled()` and `dk.TelemetryService().GetProtocols()`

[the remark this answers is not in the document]
aorcholski: I thought about it but regarding dynakube spec
- activeGate is a field
- telemetryService is a pointer

```
dk.TelemetryService() -> nil -> nil.IsEnabled() <--- it is a problem
```

pkg/api/validation/dynakube/telemetryservice.go
StefanHauth: If in doubt always be as verbose as possible. The list of duplicated protocols has some value for the user to quickly identify the boo-boo.

pkg/api/v1beta3/dynakube/telemetryservice/props.go
```diff
@@ -0,0 +1,39 @@
...
+	var protocols []Protocol
+	for _, proto := range spec.Protocols {
+		protocols = append(protocols, Protocol(proto))
+	}
+	return protocols
+}
+
+func (spec *Spec) IsEnabled() bool {
+	return spec != nil
+}
```
0sewa0, line 39: This will go boom, as the `Spec` is nillable, make it the method of the `TelemetryService` struct (similar to [KSPM](https://github.com/Dynatrace/dynatrace-operator/blob/ef1cf622916eea0875725c83d18a70b4829160c5/pkg/api/v1beta3/dynakube/kspm/props.go#L7-L10))
```
func (ts *TelemetryService) IsEnabled() bool {
	return ts.Spec != nil
}
``` 

you can have `IsEnabled` defined for the `Spec` in case of the ActiveGate as that section is not nillable

0sewa0 (changes requested):

pkg/api/v1beta3/dynakube/telemetryservice_props.go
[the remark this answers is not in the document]
0sewa0: you assumed correctly, but there is a way around it :D

pkg/api/v1beta3/dynakube/telemetryservice/props.go
0sewa0: FYI: so funny thing, it (the original that seems totally impossible) works (Adam tested it, I double checked)
```
func TestBoom(t *testing.T) {
	boom := TelemetryService{
		Spec: nil,
	}

	require.False(t, boom.IsEnabled())
}
```

just to not be confusing, we will still move the method to `TelemetryService`

aorcholski: TelemetryServiceSpec.IsEnabled replaced with TelemetryService.IsEnabled

0sewa0 (approved):

PR #4288, opened 2025-01-14
Add default initContainer resources in case of only metadata enrichment

waodim:
## Description

[Ticket](https://dt-rnd.atlassian.net/browse/DAQ-2774)

With this we add default initContainer resource to the metadata enrichment initContainer in the case we ONLY have metadata enrichment enabled without the CSI driver.

## How can this be tested?

- Unit tests
- Deploy `operator` without `CSI driver`, apply a `dynakube` with `metadata enrichment` enabled and inject into sample pods. The `initContainer` should now have the default resource values. This was not the case before.


pkg/webhook/mutation/pod/init_container.go
```diff
@@ -44,7 +44,7 @@ func initContainerResources(dk dynakube.DynaKube) corev1.ResourceRequirements {
 		return *customInitResources
 	}
 
-	if !dk.OneAgent().IsCSIAvailable() {
+	if !dk.OneAgent().IsCSIAvailable() && !dk.MetadataEnrichmentEnabled() {
```
0sewa0, line 47: this is not fully correct

scenarios you want to cover:
- no-csi + metadata + oa ==> no limit (unless custom is set)
  - why?: we have to download the archive and unzip it, the resource limits will not work 
- no-csi + metadata only ==> default limit
- no-csi + oa only ==> no limit
- csi + anything ==> default limit (unless custom is set `dk.OneAgent().GetInitResources()`)

0sewa0 (changes requested):

pkg/webhook/mutation/pod/init_container.go
waodim: Thanks for the feedback!

I think I covered all cases now: https://github.com/Dynatrace/dynatrace-operator/pull/4288/commits/72b3d43ee7600d36c9889ba01fbc5a96787764ef

pkg/webhook/mutation/pod/init_container.go
```diff
@@ -45,6 +45,10 @@ func initContainerResources(dk dynakube.DynaKube) corev1.ResourceRequirements {
 	}
 
 	if !dk.OneAgent().IsCSIAvailable() {
+		if dk.MetadataEnrichmentEnabled() {
+			return defaultInitContainerResources()
+		}
+
```
0sewa0, line 51: This is the:
no-csi + metadata only ==> default limit

what about the ?
no-csi + metadata + oa ==> no limit (unless custom is set)

```
	t.Run("should have no if OA is enabled and csi is disabled", func(t *testing.T) {
		installconfig.SetModulesOverride(t, installconfig.Modules{CSIDriver: false})

		dk := getTestDynakubeDefaultAppMon()

		dk.Spec.MetadataEnrichment.Enabled = ptr.To(true)
                dk.Spec.OneAgent.CloudNativeFullStack = &oneagent.CloudNativeFullStackSpec{}

		initResources := initContainerResources(*dk)

                require.Empty(t, initResources)
	})
```

waodim: should be covered here https://github.com/Dynatrace/dynatrace-operator/pull/4288/commits/94f387c3dae92d16be3327b5ee15399b9c438de2

pkg/webhook/mutation/pod/init_container.go
```diff
@@ -45,6 +45,10 @@ func initContainerResources(dk dynakube.DynaKube) corev1.ResourceRequirements {
 	}
 
 	if !dk.OneAgent().IsCSIAvailable() {
+		if dk.MetadataEnrichmentEnabled() && !dk.OneAgent().IsDaemonsetRequired() {
```
0sewa0, line 48:
```suggestion
		if dk.MetadataEnrichmentEnabled() && !dk.OneAgent().IsAppInjectionNeeded() {
```

pkg/webhook/mutation/pod/init_container_test.go
```diff
@@ -238,4 +239,35 @@ func TestInitContainerResources(t *testing.T) {
...
+
+		assert.Equal(t, defaultInitContainerResources(), initResources)
+	})
+	t.Run("should have no limit if OA is enabled and csi is disabled", func(t *testing.T) {
+		installconfig.SetModulesOverride(t, installconfig.Modules{CSIDriver: false})
+
+		dk := getTestDynakubeDefaultAppMon()
+
+		dk.Spec.MetadataEnrichment.Enabled = ptr.To(true)
+		dk.Spec.OneAgent.CloudNativeFullStack = &oneagent.CloudNativeFullStackSpec{}
```
0sewa0, line 267: kinda my mistake, but this is not a realistic scenario, as no-csi + cloudnative is not possible, it should be appmon

waodim: of course - I should've question this. But I get it now and understood it. Thanks!

waodim: fixed unit test https://github.com/Dynatrace/dynatrace-operator/pull/4288/commits/1d37ee6e3c6c51fff3e8f6e41a2da3e53aec6012

0sewa0 (approved):

andriisoldatenko (approved):

PR #4290, opened 2025-01-14
Remove command descriptions in case the whole application is failing

gkrenn:
## Description

Most of the operator subcommands are invoked by hardcoded pod commands. When these commands fail, a usage description is printed. This is unnecessary and causes confusion in the logs, so the print statements should be removed for subcommands that are not directly called by the user.

Example of confusing log: 
```
{"level":"info","ts":"2024-10-09T13:16:50.592Z","logger":"injection-startup","msg":"CONTAINER_INFO environment variable is missing"}
Usage:
  dynatrace-operator init [flags]

Flags:
  -h, --help   help for init
  ```

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-2912)

## How can this be tested?

Make the init container fail with an error, it should only log the error.

### DIFF
--- a/cmd/csi/init/cmd.go
+++ b/cmd/csi/init/cmd.go
@@ -25,8 +25,9 @@ var nodeId, endpoint string
 
 func New() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:  use,
-		RunE: run(),
+		Use:          use,
+		RunE:         run(),
+		SilenceUsage: true,
 	}
 
 	return cmd
--- a/cmd/csi/provisioner/cmd.go
+++ b/cmd/csi/provisioner/cmd.go
@@ -35,8 +35,9 @@ var probeAddress string
 
 func New() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:  use,
-		RunE: run(),
+		Use:          use,
+		RunE:         run(),
+		SilenceUsage: true,
 	}
 
 	addFlags(cmd)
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by andriisoldatenko** [comment]
tested and it works as expected.

**Review by andriisoldatenko — DISMISSED** [review]

**Comment by andriisoldatenko** [comment]
Just quick note: toplevel command will still print --help in case of error (I think it's exepected and waht we want:

```
./cmd
{"level":"info","ts":"2025-01-14T20:01:53.192+0100","logger":"install-config","msg":"envvar not set, using default","envvar":"modules.json"}
{"level":"info","ts":"2025-01-14T20:01:53.192+0100","logger":"install-config","msg":"problem unmarshalling envvar content, using default","envvar":"modules.json","err":"unexpected end of JSON input"}
{"level":"info","ts":"2025-01-14T20:01:53.192+0100","logger":"install-config","msg":"envvar content read and set","envvar":"modules.json","value":""}
Error: operator binary must be called with one of the subcommands
Usage:
  dynatrace-operator [flags]
  dynatrace-operator [command]

Available Commands:
  completion      Generate the autocompletion script for the specified shell
  csi-init
  csi-provisioner
  csi-server
  help            Help about any command
  init
  operator
  startup-probe
  support-archive
  troubleshoot
  webhook-server

Flags:
  -h, --help   help for dynatrace-operator

Use "dynatrace-operator [command] --help" for more information about a command.

{"level":"info","ts":"2025-01-14T20:01:53.193+0100","logger":"main","msg":"operator binary must be called with one of the subcommands"}
```

0sewa0 (dismissed):

andriisoldatenko (approved):

0sewa0 (approved):

PR #4292, opened 2025-01-15
Migrate CSI to fs from db

0sewa0:
## Description
[DAQ-4098](https://dt-rnd.atlassian.net/browse/DAQ-4098)

A rebase of [epic/csi-nodb](https://github.com/Dynatrace/dynatrace-operator/tree/epic/csi-nodb) to `main`
- didn't want to mess up the feature-branch, so did it in this branch
- it's a single commit, to keep my sanity

Some minor changes that I introduced in the rebase:
- the `RetryTimeout` is optional so we do not break during an upgrade
  - it is also not added to the OneAgent DaemonSet to not cause an redeploy due to it (the default should be fine here)
- removed cgo steps from the CSI
- made the CSI cleanup more consistent when it comes to error handling 

## How can this be tested?

CSI-driver should function without any issue.

[DAQ-4098]: https://dt-rnd.atlassian.net/browse/DAQ-4098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


waodim (dismissed):
Codewise lgtm and I tested it out on my cluster. Everything seems to work as intended 👍

pkg/controllers/csi/driver/server.go
```diff
@@ -62,39 +61,36 @@ type Server struct {
...
+	return mgr.Add(srv)
 }
 
-func (svr *Server) Start(ctx context.Context) error {
-	defer metadata.LogAccessOverview(svr.db)
-
-	endpoint, err := url.Parse(svr.opts.Endpoint)
+func (srv *Server) Start(ctx context.Context) error {
+	endpoint, err := url.Parse(srv.opts.Endpoint)
 	addr := endpoint.Host + endpoint.Path
```
andriisoldatenko, line 79: I think it was before like this, but code will panic if Parse return an err, 
```
invalid operation: operator + not defined on nil

Go build failed.
```

we should move check error before we do:
`addr := endpoint.Host + endpoint.Path` somethig like it
```
	endpoint, err := url.Parse(srv.opts.Endpoint)
	
	if err != nil {
		return fmt.Errorf("failed to parse endpoint '%s': %w", srv.opts.Endpoint, err)
	}
	
	addr := endpoint.Host + endpoint.Path

```

Dockerfile
```diff
@@ -38,6 +40,7 @@ RUN dnf install --installroot /tmp/rootfs-dependency \
       /tmp/rootfs-dependency/var/log/dnf* \
       /tmp/rootfs-dependency/var/log/yum.*
 
+# platform is required, otherwise the copy command will copy the wrong architecture files
```
andriisoldatenko, line 43: is it wrong place for comment? I assume we should move before line 2?

go.mod
```diff
@@ -75,6 +72,7 @@ require (
 	github.com/mxk/go-flowrate v0.0.0-20140419014527-cca7078d478f // indirect
 	github.com/opencontainers/image-spec v1.1.0-rc4 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
+	github.com/prometheus/client_model v0.6.1 // indirect
```
andriisoldatenko, line 75: we don't need this line; please remove it.

pkg/controllers/csi/provisioner/cleanup/run_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (c *Cleaner) createHostDirs(t *testing.T, name string) {
+	t.Helper()
+
+	hostDir := c.path.OsAgentDir(name)
+	err := c.fs.MkdirAll(hostDir, os.ModePerm)
+	require.NoError(t, err)
+}
+
+// func mockK8sObjects(t *testing.T, cleaner *Cleaner, objects ...client.Object) {
```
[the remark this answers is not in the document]
0sewa0: done

pkg/controllers/csi/driver/volumes/host/publisher.go
```diff
@@ -18,9 +18,7 @@ package hostvolumes
 
 import (
```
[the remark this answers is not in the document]
0sewa0: not really :D 
it is a `hostvolume.Publisher`
renamed it to `host` as `volume` is already in the path, so I don't stutter

Dockerfile
0sewa0: removed it, there was a `--platform` there previously but the Github CI was showing warnings saying "this is the default behaviour you dummy" so I removed it, but not the comment :D

pkg/controllers/csi/driver/server.go
0sewa0: done

go.mod
0sewa0: ran `go mod tidy`

0sewa0: (it's till there, something is using it, but not us directly)

andriisoldatenko: it's wired i did same and go mod tidy removed it 🤔

0sewa0: yes from the first `require` block, and moved it into the second `require` block (to the `// indirect` ones)

andriisoldatenko: okey now it's gone.

andriisoldatenko (approved):

waodim (approved):

PR #4309, opened 2025-01-17
Fix Log monitoring handling without Kubernetes settings available

waodim:
## Description

[Ticket](https://dt-rnd.atlassian.net/browse/DAQ-4004)

We have reached the decision, that Log monitoring without either

- `ActiveGate` with `Kubernetes monitoring` enabled, or
- The `automatic-kubernetes-monitoring` feature flag disabled

so in summary: Every `DynaKube` without either of those should be rejected. Also, if there is a `DynaKube` running, and for some reason Kubernetes settings are not yet available, we now do not error brutally but rather requeue and wait until the settings are there.

## How can this be tested?

Try a `DynaKube` with `Log monitoring` enabled and without `ActiveGate` with `kubemon` AND explicitly disable the `automatic-kubernetes-monitoring` feature flag.


pkg/controllers/dynakube/controller.go
```diff
@@ -357,6 +358,10 @@ func (controller *Controller) reconcileComponents(ctx context.Context, dynatrace
 			// this situation will clear itself after AG has been started
 			controller.setRequeueAfterIfNewIsShorter(fastUpdateInterval)
 
+			return goerrors.Join(componentErrors...)
+		} else if errors.Is(err, logmondaemonset.KubernetesSettingsNotAvailableError) {
```
gkrenn, line 362: Would combine with previous `if`

pkg/api/validation/dynakube/logmonitoring.go
```diff
@@ -7,10 +7,23 @@ import (
...
+	errorLogMonitoringWithoutK8SMonitoring = "The Dynakube's specification specifies Log monitoring without an ActiveGate with kubernetes-monitoring enabled or the automatic-kubernetes-monitoring feature flag. This is not allowed as Kubernetes settings are needed for Log monitoring."
 )
 
+func logMonitoringWithoutK8SMonitoring(ctx context.Context, dv *Validator, dk *dynakube.DynaKube) string {
+	if dk.LogMonitoring().IsEnabled() && !dk.ActiveGate().IsKubernetesMonitoringEnabled() {
+		if !dk.FeatureAutomaticKubernetesApiMonitoring() {
+			return errorLogMonitoringWithoutK8SMonitoring
+		}
+
+		return ""
```
gkrenn, line 21: Don't think this is right. 

What is assumed here is that `FeatureAutomaticKubernetesApiMonitoring` will always create the kubernetes settings if turned on. In the current implementation this only happens if kubemon capability is turned on.

If kubernetes monitoring is disabled, automatic kubemon ff does nothing - The setting will never be created.

gkrenn (changes requested):
See comment above. 

In addition, can we change the description of the feature flag `FeatureAutomaticKubernetesApiMonitoring` to tell that it's purpose is to create the Kubernetes settings object?

pkg/api/validation/dynakube/logmonitoring.go
waodim: I assumed that `FeatureAutomaticKubernetesApiMonitoring` ensures that there is a Kubernetes setting. Isn't it what it is doing?

If this is not the case we may think about changing the description

pkg/controllers/dynakube/controller.go
waodim: nice catch https://github.com/Dynatrace/dynatrace-operator/pull/4309/commits/3d8e30c8de0beba27643d316c29221de1966167b

pkg/api/validation/dynakube/logmonitoring.go
gkrenn: Only if the ActiveGate kubemon cap is turned on, look [here](https://github.com/Dynatrace/dynatrace-operator/blob/6900fe53167898c1a32e41e8bb650a5423aa6747/pkg/controllers/dynakube/activegate.go#L28).

Problem is that a MEID is required before the setting can be created I think. And the MEID is only created in case the kubemon ActiveGate is running. - Not 100% sure about that though 🤔

waodim: Yes I think you are right, AFAIK the MEID is required to create settings. Then I guess I interpreted the description of the FF wrong. 

Now that I think about it, it can be left out completely in the validation, right?

gkrenn: I think we have to look at the following 4 cases: 
- Kubemon enabled, ff enabled: Settings object will be there, logmonitoring validation should pass
- Kubemon enabled, ff disabled: Settings object will be created by customer (hopefully), logmonitoring validation should pass
- kubemon disabled, ff enabeld: Settings object will not be created, logmonitoring validation should fail
- kubemon disabled, ff disabled: Settings object will be created by customer, logmonitoring should pass

If the ff can be interpreted from customer perspective as: "Yes I created settings, let me do stuff on my own"

waodim: I am not sure about `kubemon disabled, ff disabled` as this would contradict the decision that was made to couple `logMonitoring` and `kubernetes-monitoring` capability together and make the requirement more obvious. 

On the other hand maybe a nicely formulated validation warning might do the trick here. I am unsure and would suggest to wait for @0sewa0 's feedback here, as he had discussions with PM's that led to the decision to make the requirement `kubernetes-monitoring `for `logMonitoring` more strict.

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/4309/commits/e77ae8f5161cf5ae44e6edb5ab3d2bb3c37e01bf

gkrenn (dismissed):

gkrenn (approved):

PR #4341, opened 2025-01-23
Add flag and permissions to enable pod provisioning

gkrenn:
## Description

This PR adds `codemodules.oneagent.dynatrace.com/remoteImageDownload: true` as feature flag. 
In addition permissions for creating jobs are added to the CSI driver. 

These changes are the first step to enabled remote docker image downloads.

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-3990)

## How can this be tested?

Just look at the permissions and the feature flags code

### DIFF
--- a/config/helm/chart/default/tests/Common/csi/role-csi_test.yaml
+++ b/config/helm/chart/default/tests/Common/csi/role-csi_test.yaml
@@ -47,6 +47,7 @@ tests:
                 - ""
               resources:
                 - secrets
+                - configmaps
               verbs:
                 - get
                 - list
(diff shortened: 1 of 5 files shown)


config/helm/chart/default/tests/Common/csi/role-csi_test.yaml
```diff
@@ -47,18 +47,20 @@ tests:
                 - ""
               resources:
                 - secrets
+                - configmaps
```
waodim, line 50: do I understand that correctly that you moved configmpas here because it has the same permissions like secrets?

gkrenn: Yes, it is unnecessary to write two sections with the same permissions

waodim (approved):
lgtm

PR #4346, opened 2025-01-24
Reject dynakubes that use public image and not read only oneAgents

waodim:
## Description

[Ticket](https://dt-rnd.atlassian.net/browse/DAQ-4337)

Previously when a dynakube was applied that had a mode that does not use read only oneagents (so everything that is not cloudNative or hostMon with csi) and a custom image (which we interpret as public images) then the oneAgent pods would run into a CrashLoopBackOff error.

In the pods you'd find this message:

```
❯ k logs daemonsets/cfg-oneagent
Found 3 pods, using pod/cfg-oneagent-2ghqj
[2025-01-22 13:41:58.584 UTC] [000a1650] [info   ] [bootstrapper] Started agent deployment as a container, PID 661072
[2025-01-22 13:41:58.586 UTC] [000a1650] [error  ] [bootstrapper] Volume storage should be mounted on /mnt/volume_storage_mount with write permissions
```

With this PR this configuration is now rejected in the first place.

## How can this be tested?

Try those modes in combination with a custom image 

example:

`helm install dynatrace-operator -n dynatrace oci://public.ecr.aws/dynatrace/dynatrace-operator --version 1.4.0 -n dynatrace --set csidriver.enabled=false`

```
spec:
  oneAgent:
    hostMonitoring:
      image: public.ecr.aws/dynatrace/dynatrace-oneagent:1.305.98.20250114-172617
```

### DIFF
--- a/pkg/api/validation/dynakube/oneagent.go
+++ b/pkg/api/validation/dynakube/oneagent.go
@@ -23,6 +23,8 @@ Use a nodeSelector to avoid this conflict. Conflicting DynaKubes: %s`
 
 	errorVolumeStorageReadOnlyModeConflict = `The DynaKube specification specifies a read-only host file system while OneAgent has volume storage enabled.`
 
+	errorPublicImageWithWrongConfig = `The DynaKube specification specifies a custom (and therefor probably a public) image in combination with a mode that needs write permissions for volume mounts.`
+
 	warningOneAgentInstallerEnvVars = `The environment variables ONEAGENT_INSTALLER_SCRIPT_URL and ONEAGENT_INSTALLER_TOKEN are only relevant for an unsupported image type. Please ensure you are using a supported image.`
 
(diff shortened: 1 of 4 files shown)


0sewa0 (approved):

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -23,6 +23,8 @@ Use a nodeSelector to avoid this conflict. Conflicting DynaKubes: %s`
 
 	errorVolumeStorageReadOnlyModeConflict = `The DynaKube specification specifies a read-only host file system while OneAgent has volume storage enabled.`
 
+	errorPublicImageWithWrongConfig = `The DynaKube specification specifies a custom (and therefor probably a public) image in combination with a mode that needs write permissions for volume mounts.`
```
chrismuellner, line 26:
```suggestion
	errorPublicImageWithWrongConfig = `Custom OneAgent image is only supported when CSI Driver is used.`
```
It would be much nicer to have a guide in the official documentation to link to explaining the feature flag, but for now I think we can boil it down to this.

PR #4353, opened 2025-01-27
Add debug messages to CSI identity services handlers

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-4525)
## Description

Adds debug messages to the CSI Server handlers
- GetPluginCapabilities
- GetPluginInfo
- Probe

## How can this be tested?

Add `LOG_LEVEL=debug` environment variable to the CSI Server container and check logs.

PR #4355, opened 2025-01-27
Add debug messages to CSI identity services handlers (main branch)

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-4525)

[PR for release-1.4](https://github.com/Dynatrace/dynatrace-operator/pull/4353)

## Description

Adds debug messages to the CSI Server handlers
- GetPluginCapabilities
- GetPluginInfo
- Probe

## How can this be tested?

Add `LOG_LEVEL=debug` environment variable to the CSI Server container and check logs.


pkg/controllers/csi/driver/server.go
```diff
@@ -141,14 +141,20 @@ func (srv *Server) Start(ctx context.Context) error {
 }
```
waodim, line 141: There was a comment left unanswered in the other PR which also sparks my interest tbh:

https://github.com/Dynatrace/dynatrace-operator/pull/4353#pullrequestreview-2574897633

Could someone answer that?

0sewa0 (approved):

pkg/controllers/csi/driver/server.go
aorcholski: `Probe` request is send every few seconds and is not rejected by Server rate limiter so we will see confirmation 
the Server process is running in a given period of time.

andriisoldatenko: It would be great to have this explanation inside the comment before the debug log. It will simplify future maintenance IMO.

andriisoldatenko (approved):

PR #4357, opened 2025-01-27
Fix OneAgent command line arguments

StefanHauth:
## Description

Multiple occasions of command line arguments for OneAgent shall be avoided. Exceptions are only `--set-host-property` and `--set-host-tag`. Host tags shall not be provided multiple times with the same value.

## How can this be tested?

Deploy a Dynakube with custom OneAgent arguments and make sure that the custom arguments take precedence in the OneAgent daemon set and that no duplications show up (besidd the exceptions)

### DIFF
--- a/pkg/util/prioritymap/map.go
+++ b/pkg/util/prioritymap/map.go
@@ -3,7 +3,7 @@ package prioritymap
 import (
 	"strings"
 
-	"golang.org/x/exp/slices"
+	"slices"
 	corev1 "k8s.io/api/core/v1"
 )
 
--- a/pkg/util/prioritymap/map.go
+++ b/pkg/util/prioritymap/map.go
@@ -1,9 +1,9 @@
 package prioritymap
 
 import (
+	"slices"
 	"strings"
 
-	"slices"
 	corev1 "k8s.io/api/core/v1"
 )
 
--- a/pkg/util/prioritymap/map.go
+++ b/pkg/util/prioritymap/map.go
@@ -3,6 +3,7 @@ package prioritymap
 import (
 	"strings"
 
+	"golang.org/x/exp/slices"
 	corev1 "k8s.io/api/core/v1"
 )
 
@@ -18,6 +19,7 @@ type Map struct {
 
 type entry struct {
 	value           any
+	key             string
 	delimiter       string
 	priority        int
 	allowDuplicates bool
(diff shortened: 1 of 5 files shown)


pkg/util/prioritymap/map.go
```diff
@@ -37,13 +39,34 @@ func WithSeparator(separator string) Option {
...
+	}
+}
+
+func WithAvoidDuplicatesFor(key string) Option {
+	return func(a *entry) {
+		if a.key == key {
+			a.allowDuplicates = false
+		}
+	}
+}
```
0sewa0, line 60: nitpick:
I guess this was added for the sake of completeness, but I don't see an actual scenario where we want to avoid duplicates for only a few entries 😅

StefanHauth: It was actually part of the first approach to fix it, but I see no harm in keeping it. Makes the map more versatile.

pkg/util/prioritymap/map.go
```diff
@@ -3,6 +3,7 @@ package prioritymap
 import (
 	"strings"
 
+	"golang.org/x/exp/slices"
```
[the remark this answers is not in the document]
StefanHauth: fixed in [4533214](https://github.com/Dynatrace/dynatrace-operator/pull/4357/commits/4533214bbd0b0dd070ad5fb3abd45b20c1084f7d)

andriisoldatenko:
@StefanHauth still linting is failing

pkg/util/prioritymap/map_test.go
```diff
@@ -158,3 +158,132 @@ func TestArgumentSlice(t *testing.T) {
...
+		"--set-host-id-source=auto",
+		"--set-server=localhost",
+		"--set-host-property=prop1",
+		"--set-host-property=prop2",
+		"--set-host-property=prop3",
+		"--set-host-property=prop3",
+		"--set-host-property=prop3",
+	}
+
+	custArgs := []string{
```
0sewa0, line 174: nitpick:
my 🐒 🧠  needed like 0.5 sec to understand that `cust` == `custom`, I think just calling them `customArgs` would be better

also you did not shorten `defaultArgs` but `len("default") > len("custom")` 😛

pkg/util/prioritymap/map_test.go
```diff
@@ -158,3 +158,132 @@ func TestArgumentSlice(t *testing.T) {
...
+		"--set-host-property=cust-prop3",
+		"--set-host-property=cust-prop3",
+		"--set-host-property=cust-prop3",
+	}
+
+	var tests = []struct {
+		title          string
+		expectedArgs   []string
+		defaultArgPrio int
+		custArgPrio    int
```
0sewa0, line 188: same as previous comment

pkg/util/prioritymap/map_test.go
StefanHauth: Fixed in [59eac56](https://github.com/Dynatrace/dynatrace-operator/pull/4357/commits/59eac56f63442a4fcf2850af4cf456725dd37ac0)

pkg/util/prioritymap/map_test.go
StefanHauth: Fixed in [59eac56](https://github.com/Dynatrace/dynatrace-operator/pull/4357/commits/59eac56f63442a4fcf2850af4cf456725dd37ac0)

0sewa0 (approved):

PR #4358, opened 2025-01-27
Removes a log message that corrupts support_archive zip package

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-4465)

## Description

At startup, the operator immediately issues a log line :
```
{"level":"info","ts":"2025-01-27T14:02:26.877Z","logger":"install-config","msg":"envvar content read and set","envvar":"modules.json","value":"..."} 
PK^C^D^T...
```
which precedes the first ZIP header so the package can't be unpacked by some implementations of zip command. 

Changes: 
- removed global initialization of the Modules from v1beta1(2)/convert_from packages to fix support_archive command
- GetModules function  is called by all operator subcommands to put the message at the beginning of a log file

## How can this be tested?
```
kubectl exec -n dynatrace deployment/dynatrace-operator -- dynatrace-operator support-archive --stdout > operator-support-archive.zip
```

`operator-support-archive.zip` should be decompressed with no errors.

supportarchive_console.log contains the message :
```
[support-archive]       envvar content read and set     {"envvar": "modules.json", "value": "{\n  \"csiDriver\": true,\n  \"activeGate\": true,\n  \"oneAgent\": true,\n  \"extensions\": true,\n  \"logMonitoring\": true,\n  \"edgeConnect\": true,\n  \"supportability\": true,\n  \"kspm\": true\n}\n"}
[support-archive]       dynatrace-operator      {"version": "snapshot-bugfix-corrupted-sup-arch", "gitCommit": "f00e14a06a6c72f33d7906ca07e441546802f412", "buildDate": "2025-01-27T15:29:01+00:00", "goVersion": "go1.23.5", "platform": "linux/amd64"} 
...
```

### DIFF
--- a/cmd/support_archive/builder.go
+++ b/cmd/support_archive/builder.go
@@ -98,7 +98,7 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
 
 		logBuffer := bytes.Buffer{}
 		log := newSupportArchiveLogger(getLogOutput(archiveToStdoutFlagValue, &logBuffer))
-		installconfig.GetModulesToLogger(log)
+		installconfig.ReadModulesToLogger(log)
 		version.LogVersionToLogger(log)
 
 		archiveTargetFile, err := createZipArchiveTargetFile(archiveToStdoutFlagValue, defaultSupportArchiveTargetDir)
--- a/cmd/support_archive/builder.go
+++ b/cmd/support_archive/builder.go
@@ -11,6 +11,7 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/cmd/remote_command"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/scheme"
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/installconfig"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/env"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/labels"
 	"github.com/Dynatrace/dynatrace-operator/pkg/version"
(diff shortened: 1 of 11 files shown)


cmd/support_archive/builder.go
```diff
@@ -97,6 +98,7 @@ func (builder CommandBuilder) buildRun() func(*cobra.Command, []string) error {
 
 		logBuffer := bytes.Buffer{}
 		log := newSupportArchiveLogger(getLogOutput(archiveToStdoutFlagValue, &logBuffer))
+		installconfig.GetModulesToLogger(log)
```
0sewa0, line 101: This change is enough. 
All the rest is unnecessary.  (obviously, except the changes in `pkg/util/installconfig/modules.go` 😅)

0sewa0 (changes requested):

cmd/support_archive/builder.go
aorcholski: I double checked.
Global initialization of the Modules in v1beta1(2)/convert_from packages corrupts support_archive package.

`isEnabledModules` are global variables so are initialized (`DoOnce` is executed and prints to `stdout`) at the very beginning of execution of support-archive command (before `main()`).

0sewa0: so if we add the changes in `pkg/api/v1betaX/dynakube/convert_from.go`  it should be enough?

aorcholski: yes 

`GetModules()` is called from every sub-command  to print the modules-configuration message in a predictable place (at the beginning)

0sewa0: makes sense

Could we do a tiiiny rework so it "reads nicer"

instead of calling `GetModules` randomly, I would just have a `SetModules(log)` (or `ReadModules(log)`) and call that?


and update the `GetModules` to be something like:
```
func GetModules() Modules {
	if override != nil {
		return *override
	}

	ReadModules(log)

	return modules
}

func ReadModules(log logd.Logger) Modules {
	once.Do(func() {
		modulesJson := os.Getenv(ModulesJsonEnv)
		if modulesJson == "" {
			log.Info("envvar not set, using default", "envvar", ModulesJsonEnv)
			modules = fallbackModules
		}
		err := json.Unmarshal([]byte(modulesJson), &modules)
		if err != nil {
			log.Info("problem unmarshalling envvar content, using default", "envvar", ModulesJsonEnv, "err", err)
			modules = fallbackModules
		}
		log.Info("envvar content read and set", "envvar", ModulesJsonEnv, "value", modulesJson)
	})
}
```

aorcholski: done

PR #4361, opened 2025-01-27
Add automatic restarts of agents if token is rotated

gkrenn:
## Description

The tenant and ActiveGate tokens can be rotated using the [tenant's rotation API](https://docs.dynatrace.com/docs/shortlink/tenant-token).

This process updates the `oneagent-tenant-secret` and ActiveGate secret. To activate these changes, the OneAgents and ActiveGates need to be restarted.

To accomplish this, this PR adds a hash as an annotation to the DaemonSet and StatefulSet. If the hash changes, the pods will be restarted.

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-4432)

## How can this be tested?

- Deploy dynakube with oneagents
- Navigate to the api docs in your tenant
- Start the token rotation
- the oneagents will be updated after some time.

Sample DynaKube: 
```yaml
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
spec:

  apiUrl: https://url/api
  oneAgent:
    cloudNativeFullStack:
      tolerations:
        - effect: NoSchedule
          key: node-role.kubernetes.io/master
          operator: Exists
        - effect: NoSchedule
          key: node-role.kubernetes.io/control-plane
          operator: Exists
  activeGate:
    capabilities:
      - kubernetes-monitoring
```


0sewa0 (dismissed):
I see that you always:
1. get tenant-secret
2. hash the value of the token inside
3. put it into an annotation

I think you could greatly simplify this.
- **Why not store the hash of the token in the Status?**
   - We will put it into an annotation anyway, so I see no reason against it.
- You could do some nice "magic" [here](https://github.com/Dynatrace/dynatrace-operator/blob/dfba3bb02f0a5d6d13283edbe30a53e414880e0d/pkg/controllers/dynakube/connectioninfo/common.go#L30-L59) to handle everything, and then just use the Status of the dk in everyplace you need it 🚀 .

gkrenn:
Sure, also possible, will rework 👍

config/crd/bases/dynatrace.com_dynakubes.yaml
```diff
@@ -2290,6 +2290,9 @@ spec:
                         description: Time of the last connection request
```
waodim, line 2290: - It works for the `oneAgent`, I guess in the worst case we have to wait a whole reconciliation cycle, don't we? I'm asking because if that *is* the case, can/should we improve? Maybe somehow with conditions?
- for `activeGate` it does not work for me. Does it work for you? It just did not restart ever.

gkrenn: In the worst case we have to wait the time till the cache times out (the conditions we define) and then till the next reconciliation. But that's ok in my opinion, as we want to limit the number of requests to the cluster.

ActiveGate does not update due to a cluster bug, a fix is already in the pipeline. Currently it always returns the same token, even after the rotation started, that's the reason why the ActiveGate does not restart.

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
```diff
@@ -96,7 +97,7 @@ func (r *Reconciler) generateDaemonSet() (*appsv1.DaemonSet, error) {
 	ds, err := daemonset.Build(r.dk, r.dk.LogMonitoring().GetDaemonSetName(), getContainer(*r.dk, tenantUUID),
 		daemonset.SetInitContainer(getInitContainer(*r.dk, tenantUUID)),
 		daemonset.SetAllLabels(labels.BuildLabels(), labels.BuildMatchLabels(), labels.BuildLabels(), r.dk.LogMonitoring().Template().Labels),
-		daemonset.SetAllAnnotations(nil, r.dk.LogMonitoring().Template().Annotations),
+		daemonset.SetAllAnnotations(nil, maputils.MergeMap(r.dk.LogMonitoring().Template().Annotations, r.buildAnnotations(r.dk))),
```
andriisoldatenko, line 100: do we still need it?

gkrenn: Ye, we have to add an annotation to the daemonset so the pods will restart.

pkg/controllers/dynakube/activegate/internal/statefulset/statefulset_test.go
```diff
@@ -167,6 +174,7 @@ func TestGetBaseSpec(t *testing.T) {
 		assert.Equal(t, &testReplicas, stsSpec.Replicas)
 		require.NotNil(t, stsSpec.Template.Annotations)
 		assert.Equal(t, testConfigHash, stsSpec.Template.Annotations[consts.AnnotationActiveGateConfigurationHash])
+		assert.Equal(t, testTokenHash, stsSpec.Template.Annotations[consts.AnnotationActiveGateTokenHash])
```
andriisoldatenko, line 177:
```suggestion
		assert.Equal(t, testTokenHash, stsSpec.Template.Annotations[consts.AnnotationActiveGateTenantTokenHash])
```
just for consistency

pkg/controllers/dynakube/connectioninfo/activegate/reconciler_test.go
```diff
@@ -108,8 +109,11 @@ func TestReconcile(t *testing.T) {
 		assert.Equal(t, metav1.ConditionTrue, condition.Status)
 		assert.Equal(t, conditions.SecretCreatedReason, condition.Reason)
 
+		tenantTokenHash, _ := hasher.GenerateHash(testTenantToken)
```
andriisoldatenko, line 112:
```suggestion
		tenantTokenHash, err := hasher.GenerateHash(testTenantToken)
		require.NoError(t, err)
```

pkg/controllers/dynakube/connectioninfo/oneagent/reconciler.go
```diff
@@ -128,6 +128,11 @@ func (r *reconciler) reconcileConnectionInfo(ctx context.Context) error {
 		return err
 	}
 
+	r.dk.Status.OneAgent.ConnectionInfoStatus.TenantTokenHash, err = hasher.GenerateHash(connectionInfo.ConnectionInfo.TenantToken)
+	if err != nil {
+		return err
```
andriisoldatenko, line 133:
```suggestion
		return errors.Wrap(err, "failed to generate TenantTokenHash")
```
just a little bit more context, please ;)

pkg/controllers/dynakube/connectioninfo/oneagent/reconciler_test.go
```diff
@@ -131,9 +132,12 @@ func TestReconcile(t *testing.T) {
 		err := r.Reconcile(ctx)
 		require.NoError(t, err)
 
+		tenantTokenHash, _ := hasher.GenerateHash(testTenantToken)
```
andriisoldatenko, line 135: same as about, add require.NoError(t, err)

pkg/controllers/dynakube/logmonitoring/daemonset/annotations.go
```diff
@@ -0,0 +1,16 @@
+package daemonset
+
+import (
+	"github.com/Dynatrace/dynatrace-operator/pkg/api"
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
+)
+
+const annotationTenantTokenHash = api.InternalFlagPrefix + "tenant-token-hash"
+
+func (r *Reconciler) buildAnnotations(dk *dynakube.DynaKube) map[string]string {
```
andriisoldatenko, line 10: the name of method is too generic, you are doing concrete thing here. What about
```
annotationsWithTokenHash()
```

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
andriisoldatenko: check my other comment with renaming, then this one would be much easier to udnerstand.

pkg/controllers/dynakube/activegate/consts/consts.go
```diff
@@ -28,6 +28,7 @@ const (
 	EnvDtHttpPort        = "DT_HTTP_PORT"
 
 	AnnotationActiveGateConfigurationHash = api.InternalFlagPrefix + "activegate-configuration-hash"
+	AnnotationActiveGateTokenHash         = api.InternalFlagPrefix + "activegate-token-hash"
```
aorcholski, line 31:
```suggestion
	AnnotationActiveGateTenantTokenHash         = api.InternalFlagPrefix + "activegate-tenant-token-hash"
```
I agree with @andriisoldatenko

pkg/controllers/dynakube/logmonitoring/daemonset/annotations.go
gkrenn: That was on purpose as this place should as the one place where all the annotations are built, will make it more clear though

pkg/controllers/dynakube/activegate/internal/statefulset/statefulset_test.go
gkrenn: done

pkg/controllers/dynakube/connectioninfo/activegate/reconciler_test.go
gkrenn: done

pkg/controllers/dynakube/connectioninfo/oneagent/reconciler_test.go
gkrenn: done

andriisoldatenko (approved):

PR #4377, opened 2025-01-28
Add validations for OneAgent arguments (release-1.4)

StefanHauth:
## Description

Makes sure that there are
* no duplicate arguments for OneAgent are allowed (except for `--set-host-property` and `--set-host-tag`
* no duplicate host-tags are allowed
* `--set-host-id-source` is not allowed for cloud native fullstack

### Webhook response samples
```
  oneAgent:
    cloudNativeFullStack:
      args:
      - --set-host-property=prop1
      - --set-host-property=prop2
      - --set-host-tag=tag1
      - --set-server=foo
      - --set-server=bar

Error from server (Forbidden): error when creating "cloudNativeFullStack-v1beta3.yaml": admission webhook "v1beta3.dynakube.webhook.dynatrace.com" denied the request:
 1. --set-server has been provided multiple times. Only --set-host-property and --set-host-tag arguments may be provided multiple times.
```

```
  oneAgent:
    cloudNativeFullStack:
      args:
      - --set-host-property=prop1
      - --set-host-property=prop2
      - --set-host-tag=tag1
      - --set-host-id-source=fqdn

Error from server (Forbidden): error when creating "cloudNativeFullStack-v1beta3.yaml": admission webhook "v1beta3.dynakube.webhook.dynatrace.com" denied the request:
 1. Setting --set-host-id-source in CloudNativFullstack mode is not allowed.
```

```
  oneAgent:
    cloudNativeFullStack:
      args:
      - --set-host-property=prop1
      - --set-host-property=prop2
      - --set-host-tag=tag1
      - --set-host-tag=tag1

The DynaKube "dynakube" is invalid: spec.oneAgent.cloudNativeFullStack.args[3]: Duplicate value: "--set-host-tag=tag1"
```

## How can this be tested?

Deploy Dynakubes that violate above conditions, they should be rejected

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-4488)

### DIFF
--- a/pkg/api/v1beta3/dynakube/oneagent_props.go
+++ b/pkg/api/v1beta3/dynakube/oneagent_props.go
@@ -228,6 +228,35 @@ func (dk *DynaKube) DefaultOneAgentImage(version string) string {
 	return apiUrlHost + DefaultOneAgentImageRegistrySubPath + ":" + tag
 }
 
+func (dk *DynaKube) OneAgentArgumentsMap() map[string][]string {
+	var args []string
+
+	switch {
+	case dk.CloudNativeFullstackMode() && dk.Spec.OneAgent.CloudNativeFullStack.Args != nil:
+		args = dk.Spec.OneAgent.CloudNativeFullStack.Args
+	case dk.ClassicFullStackMode() && dk.Spec.OneAgent.ClassicFullStack.Args != nil:
+		args = dk.Spec.OneAgent.ClassicFullStack.Args
+	case dk.HostMonitoringMode() && dk.Spec.OneAgent.HostMonitoring.Args != nil:
+		args = dk.Spec.OneAgent.HostMonitoring.Args
+
+	default:
+		return nil
+	}
+
+	argMap := make(map[string][]string)
+
+	for _, arg := range args {
+		key, value := splitArg(arg)
+		if _, exists := argMap[key]; !exists {
+			argMap[key] = []string{value}
+		} else {
+			argMap[key] = append(argMap[key], value)
+		}
+	}
+
+	return argMap
+}
+
 func (dk *DynaKube) HostGroup() string {
 	if dk.Spec.OneAgent.HostGroup != "" {
 		return dk.Spec.OneAgent.HostGroup
--- a/pkg/api/validation/dynakube/oneagent.go
+++ b/pkg/api/validation/dynakube/oneagent.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"slices"
 	"strings"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
@@ -32,6 +33,12 @@ Use a nodeSelector to avoid this conflict. Conflicting DynaKubes: %s`
 	versionRegex = `^\d+.\d+.\d+.\d{8}-\d{6}$`
 
 	versionInvalidMessage = "The OneAgent's version is only valid in the format 'major.minor.patch.timestamp', e.g. 1.0.0.20240101-000000"
+
+	errorDuplicateOneAgentArgument = "%s has been provided multiple times. Only --set-host-property and --set-host-tag arguments may be provided multiple times."
+
+	errorHostIdSourceArgumentInCloudNative = "Setting --set-host-id-source in CloudNativFullstack mode is not allowed."
+
+	errorSameHostTagMultipleTimes = "Providing the same tag(s) (%s) multiple times with --set-host-tag is not allowed."
 )
 
 func conflictingOneAgentConfiguration(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
@@ -200,4 +207,54 @@ func isOneAgentVersionValid(_ context.Context, _ *Validator, dk *dynakube.DynaKu
 	}
 
 	return ""
+}
+
+func duplicateOneAgentArguments(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
+	args := dk.OneAgentArgumentsMap()
+	if args == nil {
+		return ""
+	}
+
+	for key, values := range args {
+		if key != "--set-host-property" && key != "--set-host-tag" && len(values) > 1 {
+			return fmt.Sprintf(errorDuplicateOneAgentArgument, key)
+		} else if key == "--set-host-tag" {
+			if duplicatedTags := findDuplicates(values); len(duplicatedTags) > 0 {
+				return fmt.Sprintf(errorSameHostTagMultipleTimes, duplicatedTags)
+			}
+		}
+	}
+
+	return ""
+}
+
+func forbiddenHostIdSourceArgument(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
+	args := dk.OneAgentArgumentsMap()
+	if args == nil {
+		return ""
+	}
+
+	for key := range args {
+		if dk.CloudNativeFullstackMode() && key == "--set-host-id-source" {
+			return errorHostIdSourceArgumentInCloudNative
+		}
+	}
+
+	return ""
+}
+
+func findDuplicates[S ~[]E, E comparable](s S) []E {
+	seen := make(map[E]bool)
+
+	duplicates := make([]E, 0)
+
+	for _, val := range s {
+		if _, ok := seen[val]; !ok {
+			seen[val] = true
+		} else if !slices.Contains(duplicates, val) {
+			duplicates = append(duplicates, val)
+		}
+	}
+
+	return duplicates
 }
--- a/pkg/api/validation/dynakube/oneagent_test.go
+++ b/pkg/api/validation/dynakube/oneagent_test.go
@@ -598,7 +598,7 @@ func TestOneAgentArguments(t *testing.T) {
 				"--set-host-tag=bar",
 				"--set-host-tag=foo",
 			},
-			expectedError: fmt.Sprintf(errorSameHostTagMultipleTimes, "[bar foo]"),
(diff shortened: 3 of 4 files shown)


pkg/api/validation/dynakube/oneagent.go
```diff
@@ -191,3 +197,51 @@ func isOneAgentVersionValid(_ context.Context, _ *Validator, dk *dynakube.DynaKu
...
+		return ""
+	}
+
+	for key, value := range args {
+		if key != "--set-host-property" && key != "--set-host-tag" && len(value) > 1 {
+			return fmt.Sprintf(errorDuplicateOneAgentArgument, key)
+		} else if key == "--set-host-tag" {
+			if tag := hasDuplicates(value); tag != nil {
+				return fmt.Sprintf(errorSameHostTagMultipleTimes, *tag)
+			}
```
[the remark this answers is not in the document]
0sewa0: only downside is that it doesn't give you the exact tag that is repeated, however your `hasDuplicates` only gives you the first duplicate which is not the complete answer aswell 😅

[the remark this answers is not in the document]
0sewa0: you could just print the values of the tags, for simplicities sake
```
return fmt.Sprintf(errorSameHostTagMultipleTimes, values)
```

pkg/api/v1beta3/dynakube/oneagent_props.go
```diff
@@ -228,6 +228,35 @@ func (dk *DynaKube) DefaultOneAgentImage(version string) string {
 	return apiUrlHost + DefaultOneAgentImageRegistrySubPath + ":" + tag
 }
 
+func (dk *DynaKube) OneAgentArguments() map[string][]string {
```
0sewa0, line 231: I would expect this to just return the `[]string` as the `Args` is a `[]string` 

Trying to come up with a better name, but failing 😅 
(I like what it's doing, I just don't like the name)

0sewa0 (changes requested):

pkg/api/v1beta3/dynakube/oneagent_props.go
StefanHauth: The idea is to have a single location where the arguments are properly "parsed". And the function returns the OneAgent arguments. It's about what is returned not in which format.

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -191,3 +197,51 @@ func isOneAgentVersionValid(_ context.Context, _ *Validator, dk *dynakube.DynaKu
...
+	for _, val := range s {
+		if _, ok := seen[val]; !ok {
+			seen[val] = true
+		} else {
+			return &val
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
StefanHauth: reworked the function in [1431322](https://github.com/Dynatrace/dynatrace-operator/pull/4377/commits/143132226d6602639d2d92727e6fffd6080210d2)

pkg/api/validation/dynakube/oneagent.go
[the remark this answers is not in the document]
StefanHauth: reworked the function in [1431322](https://github.com/Dynatrace/dynatrace-operator/pull/4377/commits/143132226d6602639d2d92727e6fffd6080210d2)

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -191,3 +197,51 @@ func isOneAgentVersionValid(_ context.Context, _ *Validator, dk *dynakube.DynaKu
...
 	return ""
 }
+
+func hasNoDuplicatOneAgentArguments(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
+	args := dk.OneAgentArguments()
+	if args == nil {
+		return ""
+	}
+
+	for key, value := range args {
```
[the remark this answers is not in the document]
StefanHauth: fixed in [1431322](https://github.com/Dynatrace/dynatrace-operator/pull/4377/commits/143132226d6602639d2d92727e6fffd6080210d2)

pkg/api/v1beta3/dynakube/oneagent_props.go
0sewa0: > the arguments are properly "parsed"

that is kinda the problem, other function that is just basically name of the field you are getting does no parsing (OneAgentNodeSelector, OneAgentSecCompProfile, OneAgentNamespaceSelector)
- So my main problem is that you are going against the current pattern

on `main` this would read `dk.OneAgent().Arguments()`, and someone might use it to get the that inner field `dk.Spec.OneAgent.CloudNativeFullStack.Args`, which is a `[]string` but they will weirdly receive a `map[string]string`
- you could call it `dk.OneAgentArgumentsMap()` which is a mouthful, but considering that in `main`it would be `dk.OneAgent().ArgumentsMap()` which is IMO not that bad

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -191,3 +197,53 @@ func isOneAgentVersionValid(_ context.Context, _ *Validator, dk *dynakube.DynaKu
 
 	return ""
 }
+
+func hasNoDuplicatOneAgentArguments(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
```
0sewa0, line 201:
```suggestion
func duplicateOneAgentArguments(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
```

I tried to follow the naming of other functions

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -191,3 +197,53 @@ func isOneAgentVersionValid(_ context.Context, _ *Validator, dk *dynakube.DynaKu
...
+			if duplicatedTags := findDuplicates(values); len(duplicatedTags) > 0 {
+				return fmt.Sprintf(errorSameHostTagMultipleTimes, duplicatedTags)
+			}
+		}
+	}
+
+	return ""
+}
+
+func hasNoHostIdSourceArgument(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
```
0sewa0, line 220:
```suggestion
func forbiddenHostIdSourceArgument(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
```

I tried to follow the naming of other functions

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -191,3 +197,53 @@ func isOneAgentVersionValid(_ context.Context, _ *Validator, dk *dynakube.DynaKu
...
+	for _, val := range s {
+		if _, ok := seen[val]; !ok {
+			seen[val] = true
+		} else {
+			duplicates = append(duplicates, val)
+		}
+	}
+
+	return duplicates
+}
```
0sewa0, line 249: you are collecting the duplicates in a way that you duplicating them in the output 😅
- which is IMO not that helpful

so if you have:
- `[]string{"foo", "foo", "foo", "foo"}` the output will be `[]string{"foo", "foo", "foo"}` -> OK? kinda confusing
- `[]string{"foo", "foo", "bar", "bar"}` the output will be `[]string{"foo", "bar"}` -> IMO this is what you want (or this is the least confusing output)
- `[]string{"foo", "foo", "bar", "bar", "foo"}` the output will be `[]string{"foo", "bar", "foo"}` -> well, this is confusing 😅

you could just do:
```suggestion
func findDuplicates[S ~[]E, E comparable](s S) []E {
	seen := make(map[E]bool)

	duplicates := make([]E, 0)

	for _, val := range s {
		if _, ok := seen[val]; !ok {
			seen[val] = true
            duplicates = append(duplicates, val)
		}
	}

	return duplicates
}
```

but IMO, this is still a bit of overengineering, I think it would be enough to tell the user that `you defined duplicate --set-host-tag args` the rest they can figure out easily

andriisoldatenko: I agree with @0sewa0. It's overengeniring for mostly nothing.

StefanHauth: Telling the user as exactly as possible what the problem is improves UX, so this clearly has value to the customer.
But the duplication of tags in the error message needs some fixing, ofc.

pkg/api/validation/dynakube/oneagent.go
StefanHauth: Fixed in [5d093cf](https://github.com/Dynatrace/dynatrace-operator/pull/4377/commits/5d093cf3fe7092cda67d5f0f73208e02a81e2716)

pkg/api/validation/dynakube/oneagent.go
StefanHauth: Fixed in [5d093cf](https://github.com/Dynatrace/dynatrace-operator/pull/4377/commits/5d093cf3fe7092cda67d5f0f73208e02a81e2716)

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -201,3 +208,53 @@ func isOneAgentVersionValid(_ context.Context, _ *Validator, dk *dynakube.DynaKu
...
+
+	for _, val := range s {
+		if _, ok := seen[val]; !ok {
+			seen[val] = true
+		} else if !slices.Contains(duplicates, val) {
+			duplicates = append(duplicates, val)
+		}
+	}
+
+	return duplicates
```
0sewa0, line 259: A possible simplification
```suggestion 
	for _, val := range s {
		if _, ok := seen[val]; !ok {
			seen[val] = true
		} else {
			duplicates = append(duplicates, val)
		}
	}

	return slices.Compact(duplicates)
```

0sewa0 (approved):
just added a small (optional) suggestion

PR #4392, opened 2025-02-03
Add logic to download CodeModules in separate Job

0sewa0:
## Description
[DAQ-3994](https://dt-rnd.atlassian.net/browse/DAQ-3994)

If the `codemodules.oneagent.dynatrace.com/remoteImageDownload: "true"` feature flag is set, the CSI `provisioner` will schedule Jobs to download the CodeModule
- For details, look at the commit messages/descriptions

## How can this be tested?

Example `DynaKube`:
```
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
  annotations:
    codemodules.oneagent.dynatrace.com/remoteImageDownload: "true"
spec:
  apiUrl: https://<mytenant>.dev.dynatracelabs.com/api
  customPullSecret: <my-pull-secret>
  oneAgent:
    cloudNativeFullStack:
      codeModulesImage: quay.io/dynatrace/dynatrace-bootstrapper:snapshot-testing-add-testing-parts
```

Things to note:
- You have to use `codeModulesImage:  quay.io/dynatrace/dynatrace-bootstrapper:snapshot-testing-add-testing-parts` for testing, as there are no official images available yet with the bootstrapper in them
   ~- This quay repo is private so you should have a pull-secret configured to use it (also, nice test)~
   - You can also build https://github.com/Dynatrace/dynatrace-bootstrapper if you don't like this :D 

You should see that Jobs/Pods get created, that after succeeding get deleted, and on the fs `/data/codemodules` a codemodule got successfully copied
Works with multiple DynaKubes(1 with job, 1 without job for example), as the GC leaves the `/data/work` folder alone. (probably could be renamed) 

[DAQ-3994]: https://dt-rnd.atlassian.net/browse/DAQ-3994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/api/v1beta3/dynakube/feature_flags.go
```diff
@@ -70,10 +70,10 @@ const (
 	AnnotationFeatureEnforcementMode       = AnnotationFeaturePrefix + "enforcement-mode"
 
 	// CSI.
-	AnnotationFeatureMaxFailedCsiMountAttempts      = AnnotationFeaturePrefix + "max-csi-mount-attempts"
-	AnnotationFeatureMaxCsiMountTimeout             = AnnotationFeaturePrefix + "max-csi-mount-timeout"
-	AnnotationFeatureReadOnlyCsiVolume              = AnnotationFeaturePrefix + "injection-readonly-volume"
-	AnnotationFeatureCodeModulesRemoteImageDownload = AnnotationFeatureCodeModulesPrefix + "remoteImageDownload"
```
andriisoldatenko, line 76: do you want remove old feature flag?

0sewa0: they were not removed ... 😬

pkg/controllers/csi/provisioner/install.go
```diff
@@ -3,18 +3,25 @@ package csiprovisioner
...
 	"github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer/image"
+	"github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer/job"
 	"github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer/symlink"
 	"github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer/url"
 	"github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/processmoduleconfig"
 )
 
+const notReadyRequeueDuration = 30 * time.Second
+
+var notReady = errors.New("download job is not ready yet")
```
andriisoldatenko, line 23: Fixed error variables, usually named `Err...`.

pkg/injection/codemodule/installer/job/installer.go
```diff
@@ -0,0 +1,121 @@
...
+
+func NewInstaller(ctx context.Context, fs afero.Fs, props *Properties) installer.Installer {
+	return &Installer{
+		fs:       fs,
+		props:    props,
+		nodeName: env.GetNodeName(),
+	}
+}
+
+type Installer struct {
```
andriisoldatenko, line 39: why do you need to duplicate here installer.Installer? am I miss something?

andriisoldatenko: can we embed one type into another to add extra methods?

0sewa0: the name of the package is `job`

when you import and use it:
```
"github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer/job"

job.NewInstaller()
```

0sewa0: > can we embed one type into another to add extra methods?

there is a common `interface` that is the `installer.Installer` 
- this is old code, didn't want to change it

andriisoldatenko: gotcha

pkg/api/v1beta3/dynakube/feature_flags.go
andriisoldatenko: oh... this git diff L) i need 👓

pkg/injection/codemodule/installer/job/installer.go
```diff
@@ -0,0 +1,121 @@
...
+	fs       afero.Fs
+	props    *Properties
+	nodeName string
+}
+
+func (inst *Installer) InstallAgent(ctx context.Context, targetDir string) (bool, error) {
```
andriisoldatenko, line 45:
```suggestion
func (i *Installer) InstallAgent(ctx context.Context, targetDir string) (bool, error) {
```
there is no strict rule :) but short name what we do everywhere (if you don't want to confuse with `i` what about `in` ?
it's not strict suggestion just more like question.

0sewa0: `i` is "iterator" so nonono

`in` is (to me) a python keyword :D

`inst` is to shortest I liked 😅

pkg/injection/codemodule/installer/job/installer.go
```diff
@@ -0,0 +1,121 @@
...
+
+		log.Info("failed to create symlink for agent installation", "err", err)
+
+		return false, errors.WithStack(err)
+	}
+
+	return true, nil
+}
+
+func (inst *Installer) isReady(ctx context.Context, targetDir, jobName string) (bool, error) {
```
andriisoldatenko, line 77: I'm just curious: jobName is something we can get from `inst.GetJobName()`; why do we want to pass it as an argument?

and also: IMO it's better to rename for something like `jobIsReady`, cause it's what it does.

pkg/injection/codemodule/installer/job/installer.go
```diff
@@ -0,0 +1,121 @@
...
+	job, err := inst.query().Get(ctx, types.NamespacedName{Name: jobName, Namespace: inst.props.Owner.GetNamespace()})
```
andriisoldatenko, line 78: for example, Namespace you get using `inst.props.Owner.GetNamespace()`

pkg/injection/codemodule/installer/job/installer.go
0sewa0: `getJobName` does hashing, didn't want to do that multiple times, might make sense

andriisoldatenko: agree, yes let's keep it.

doc/api/feature-flags.md
```diff
@@ -54,10 +54,10 @@ const (
     AnnotationFeatureEnforcementMode       = AnnotationFeaturePrefix + "enforcement-mode"
 
     // CSI.
-    AnnotationFeatureMaxFailedCsiMountAttempts      = AnnotationFeaturePrefix + "max-csi-mount-attempts"
-    AnnotationFeatureMaxCsiMountTimeout             = AnnotationFeaturePrefix + "max-csi-mount-timeout"
-    AnnotationFeatureReadOnlyCsiVolume              = AnnotationFeaturePrefix + "injection-readonly-volume"
-    AnnotationFeatureCodeModulesRemoteImageDownload = AnnotationFeatureCodeModulesPrefix + "remoteImageDownload"
```
andriisoldatenko, line 60:
```
➜  dynatrace-operator git:(feature/download-via-job) rg AnnotationFeatureCodeModulesRemoteImageDownload
➜  dynatrace-operator git:(feature/download-via-job)
```
return nothing so we removed/renamed old feature flag using new one? is it accurate?

0sewa0: only the name of the `var` and related function was renamed, this function/ff was not yet used

pkg/controllers/csi/provisioner/install.go
0sewa0: done

doc/api/feature-flags.md
```diff
@@ -54,10 +54,10 @@ const (
...
+    AnnotationFeatureMaxFailedCsiMountAttempts = AnnotationFeaturePrefix + "max-csi-mount-attempts"
+    AnnotationFeatureMaxCsiMountTimeout        = AnnotationFeaturePrefix + "max-csi-mount-timeout"
+    AnnotationFeatureReadOnlyCsiVolume         = AnnotationFeaturePrefix + "injection-readonly-volume"
+    AnnotationFeatureDownloadViaJob            = AnnotationFeatureCodeModulesPrefix + "standalone"
```
chrismuellner, line 60: replace back to `codemodules.oneagent.dynatrace.com/remoteImageDownload` please

chrismuellner (dismissed):

doc/api/feature-flags.md
0sewa0: done

andriisoldatenko (changes requested):
Found interesting bug:

In logs all good, 
```
codemodule-download time="2025-02-04T18:55:13Z" level=info msg="Copying file /opt/dynatrace/oneagent/log/ignore_me.txt to /data/work/codemodule-
download-2396602260/copy-tmp/log/ignore_me.txt"
codemodule-download time="2025-02-04T18:55:13Z" level=info msg="Copying file /opt/dynatrace/oneagent/manifest.json to /data/work/codemodule-down
load-2396602260/copy-tmp/manifest.json"
codemodule-download time="2025-02-04T18:55:13Z" level=info msg="Successfully copied from /opt/dynatrace/oneagent to /data/codemodules/cXVheS5pby
9keW5hdHJhY2UvZHluYXRyYWNlLWJvb3RzdHJhcHBlcjpzbmFwc2hvdC10ZXN0aW5nLWFkZC10ZXN0aW5nLXBhcnRz"
```

and container completed:
image
^^
but because I injected isito by mistake to dynatrace NS:

Job still running:

```
k get jobs -n dynatrace
NAME                             STATUS    COMPLETIONS   DURATION   AGE
codemodule-download-1067618131   Running   0/1           5m2s       5m2s
codemodule-download-2396602260   Running   0/1           5m2s       5m2s
codemodule-download-633334064    Running   0/1           5m2s       5m2s
```

This check seems inaccurate:

https://github.com/dynatrace/dynatrace-operator/blob/b8ad5ebf28ac2c2f35a5fbb40f79e60d199c89d2/pkg/injection/codemodule/installer/job/installer.go#L92


```
k get jobs codemodule-download-1067618131 -n dynatrace -o=jsonpath="{.status}"
{"active":1,"ready":0,"startTime":"2025-02-04T18:54:16Z","terminating":0,"uncountedTerminatedPods":{}}%
```

may be we could check that our container completed, and ignore all other SideCars (like istio OR security).

pkg/injection/codemodule/installer/job/bootstrapper.go
```diff
@@ -0,0 +1,89 @@
...
+		Image:           inst.props.ImageUri,
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		Command:         []string{bootstrapCommand},
+		VolumeMounts: []corev1.VolumeMount{
+			{
+				Name:      volumeName,
+				MountPath: inst.props.PathResolver.RootDir,
+			},
+		},
+		SecurityContext: &corev1.SecurityContext{ // TODO: Use the SecurityContext from the `provisioner` container
```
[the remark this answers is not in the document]
0sewa0: it will be done in a later story (that I didn't create yet 😅)

pkg/injection/codemodule/installer/job/installer.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	if k8serrors.IsNotFound(err) && inst.isAlreadyPresent(targetDir) {
+		log.Info("agent already installed", "image", inst.props.ImageUri, "target dir", targetDir)
+
+		return true, nil
+	}
+
+	if !k8serrors.IsNotFound(err) {
+		if job.Status.Succeeded > 0 {
```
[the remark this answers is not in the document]
0sewa0: there are many problems here, that kinda makes using a `Job` redundant:
1. The`Job` does not give me any information about a given `container`, in this istio scenario, according to the `Job`, it is active
   - there are no relevant `Events` nor `Conditions` on the `Job`
2. To find the relevant `Pod` for my `Job` I have list all of them, and filter according to the owner-reference (annoying)
3. After that I have to go digging in `containerStatuses` and find my container in a `terminated` state with an errorCode of 0 (annoying)

What I will do instead is to just check on the FS, as the this download should be atomic
- so if the expected thing is there, then I will just delete the job and call it a day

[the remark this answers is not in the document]
andriisoldatenko: > so if the expected thing is there, then I will just delete the job and call it a day
I like the approach,

aorcholski (commented):
can I call my dynakube
- work
- codemodules
- appmounts
?

"codemodules":
```
bash-5.1# ls -l /data/            
total 284
drwxrwxrwx 5 root root   4096 Feb  5 16:55 appmounts
drwxr-xr-x 6 root root   4096 Feb  5 16:56 codemodules
-rw-r--r-- 1 root root   4096 Feb  5 05:02 csi.db
-rw-r--r-- 1 root root  32768 Feb  5 10:04 csi.db-shm
-rw-r--r-- 1 root root 238992 Feb  5 10:04 csi.db-wal
drwxr-xr-x 5 root root   4096 Feb  5 16:54 work

bash-5.1# ls -l /data/codemodules/
total 20
drwxrwxrwx 3 root root 4096 Feb  5 16:55 appmounts
drwxr-xr-x 6 root root 4096 Feb  5 16:54 cXVheS5pby9keW5hdHJhY2UvZHluYXRyYWNlLWJvb3RzdHJhcHBlcjpzbmFwc2hvdC10ZXN0aW5nLWFkZC10ZXN0aW5nLXBhcnRz
drwxr-xr-x 3 root root 4096 Feb  5 16:55 config
lrwxrwxrwx 1 root root  118 Feb  5 16:56 latest-codemodule -> /data/codemodules/cXVheS5pby9keW5hdHJhY2UvZHluYXRyYWNlLWJvb3RzdHJhcHBlcjpzbmFwc2hvdC10ZXN0aW5nLWFkZC10ZXN0aW5nLXBhcnRz
drwxrwxrwx 9 root root 4096 Feb  5 16:54 osagent
```

0sewa0:
> can I call my dynakube ...

@aorcholski good point, but that should be fixed different place 😅

waodim (approved):
Tested it several times and it lgtm. New approach of checking for the jobs success also works.

andriisoldatenko (approved):

PR #4390, opened 2025-02-03
misleading declaration of healthz port between server and liveness-probe

andriisoldatenko:
## Description

This PR fixes bug [DAQ-4645](https://dt-rnd.atlassian.net/browse/DAQ-4645)

## How can this be tested?

1. helm unit tests
2. deploy dk with CSI driver enabled:
```
apiVersion: dynatrace.com/v1beta2
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
spec:
  apiUrl: https://*****.dev.dynatracelabs.com/api
  oneAgent:
    applicationMonitoring:
      useCSIDriver: true
```

If you run, you will see that after this PR 

server **doesn't have** liveness probe port declaration
```
{
  "args": [
    "csi-server",
    "--endpoint=unix:/csi/csi.sock",
    "--node-id=$(KUBE_NODE_NAME)"
  ],
...
  "name": "server",
  "ports": [
    {
      "containerPort": 8080,
      "name": "metrics",
      "protocol": "TCP"
    }
  ],
...
}

```

and liveness-probe container has it:

```
k get pods dynatrace-oneagent-csi-driver-44888 -n dynatrace -o jsonpath='{.spec.
containers[3]}' | jq
{
  "args": [
    "--csi-address=/csi/csi.sock",
    "--health-port=9808",
    "--probe-timeout=4s"
  ],
  "command": [
    "livenessprobe"
  ],
  "image": "quay.io/dynatrace/dynatrace-operator:snapshot-misleading-declaration-healthz-port",
  "imagePullPolicy": "Always",
  "name": "liveness-probe",
  "ports": [
    {
      "containerPort": 9808,
      "name": "healthz",
      "protocol": "TCP"
    }
  ],
```

[DAQ-4645]: https://dt-rnd.atlassian.net/browse/DAQ-4645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4397, opened 2025-02-04
Revert "misleading declaration of healthz port between server and liveness-probe"

andriisoldatenko:
Reverts Dynatrace/dynatrace-operator#4390


0sewa0 (approved):
revert it for now, and we will discuss possible improvements on clearly communicating what is happening with the ports (which container is listening on it, but which container is actually using it)

PR #4413, opened 2025-02-05
Add `/var/log` mount to log agent

waodim:
## Description

With this PR we add a new mount to `/var/log` to the `logAgent`. Also other mounts' type got changed to `Directory` as we do not want to create a directory if it does not exist on host filesystem.

[Ticket](https://dt-rnd.atlassian.net/browse/DAQ-4838)

## How can this be tested?

Deploy logAgent and have a look at the `volumes` and the `volumeMounts`

Should look like this:

```
volumeMounts:
- mountPath: /var/log/journal
  name: var-log-journal
  readOnly: true
...
volumes:
- name: var-log-journal
  hostPath:
    path: /var/log
    type: Directory
```

### DIFF
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
@@ -22,12 +22,15 @@ const (
 	dtLogVolumeMountPath = "/var/log/dynatrace"
 
 	// for the logs that the logmonitoring will ingest
-	podLogsVolumeName       = "var-log-pods"
-	podLogsVolumePath       = "/var/log/pods"
-	dockerLogsVolumeName    = "docker-container-logs"
-	dockerLogsVolumePath    = "/var/lib/docker/containers"
-	containerLogsVolumeName = "container-logs"
-	containerLogsVolumePath = "/var/log/containers"
+	podLogsVolumeName         = "var-log-pods"
+	podLogsVolumePath         = "/var/log/pods"
+	dockerLogsVolumeName      = "docker-container-logs"
+	dockerLogsVolumePath      = "/var/lib/docker/containers"
+	containerLogsVolumeName   = "container-logs"
+	containerLogsVolumePath   = "/var/log/containers"
+	journalLogsVolumeName     = "var-log-journal"
+	journalLogsVolumePath     = "/var/log/journal"
+	journalLogsVolumeHostPath = "/var/log"
 )
 
 // getConfigVolumeMount provides the VolumeMount for the deployment.conf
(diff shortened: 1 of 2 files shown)


pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
```diff
@@ -74,7 +77,7 @@ func getDTVolumes() []corev1.Volume {
 			VolumeSource: corev1.VolumeSource{
 				HostPath: &corev1.HostPathVolumeSource{
 					Path: dtLibVolumePath,
-					Type: ptr.To(corev1.HostPathDirectoryOrCreate),
+					Type: ptr.To(corev1.HostPathDirectory),
```
0sewa0, line 80: is this OK to do?

as this is for the dynatrace logs, so it can be not there without us

pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
```diff
@@ -130,6 +135,33 @@ func getIngestVolumes() []corev1.Volume {
...
+		{
+			Name:      journalLogsVolumeName,
+			MountPath: journalLogsVolumePath,
+			ReadOnly:  true,
+		},
+	}
+}
+
+// getIngestVolumeMounts provides the VolumeMounts for the log folders that will be ingested by the logmonitoring
+func getJournalLogsVolumes() []corev1.Volume {
```
0sewa0, line 157: technically, this is also an "ingest volume" as it is mounted so that log-monitoring can ingest its contents 😅 

also, the comments are copy-pasted 😄

pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
waodim: I agree I changed it back.

@chrismuellner do you also agree?

done https://github.com/Dynatrace/dynatrace-operator/pull/4413/commits/226f5a9a17bf7b71a17fec9dd75a90d322124032

pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
waodim: I moved it into the ingest volumes https://github.com/Dynatrace/dynatrace-operator/pull/4413/commits/226f5a9a17bf7b71a17fec9dd75a90d322124032

pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
chrismuellner: yes, directories for log agent specific configuration should be created if they don't exist!

0sewa0 (approved):

chrismuellner (approved):

PR #4419, opened 2025-02-06
Move csi dk dirs into sub dir

0sewa0:
## Description

[DAQ-4933](https://dt-rnd.atlassian.net/browse/DAQ-4933)

To avoid `DynaKube` names (such as `work`, `codemodules`, `appmounts`) to conflict with hard-coded paths in the FS of our CSI driver, we put move them into the ~`dyna.kubes`~ `_dynakubes` dir.
- why the goofy dir name:
   - wanted to make sure, that no older/future `DynaKube` names can cause conflicts (a `DynaKube` can't have `.` in it's na,e)
      - (I am open to name suggestions 😅 ) 

## How can this be tested?

- Migrate from an older version to this branch: To test the cleanup is safe
- Use the csi-driver to mount volumes: To test the it actually works

It should behave as previously, but now calling your `DynaKube` something like `codemodules` shouldn't cause problems

⚠️ ⚠️ ⚠️ 
Use the `codeModulesImage`: `quay.io/dynatrace/dynatrace-bootstrapper:snapshot-fix-atomic` for testing, otherwise you might experience undefined behavior caused by the bootstrapper's logic not being atomic ([which is a bug fixed here](https://github.com/Dynatrace/dynatrace-bootstrapper/pull/7))
⚠️ ⚠️ ⚠️ 

[DAQ-4933]: https://dt-rnd.atlassian.net/browse/DAQ-4933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controllers/csi/config.go
```diff
@@ -31,6 +31,7 @@ const (
 	OverlayWorkDirPath   = "work"
 	SharedAgentBinDir    = "codemodules"
 	SharedAppMountsDir   = "appmounts"
+	SharedDynaKubesDir   = "dyna.kubes"
```
waodim, line 34: How about `shared-dk` or `shared-dynakube` or something along those lines?

I mean I am not completely against this name but just wanted to give some ideas

0sewa0: you can create a DynaKube with name `shared-dk` and `shared-dynakube`

[the validation will not catch it](https://github.com/Dynatrace/dynatrace-operator/blob/1ade90a1a28f3216c48420f3c1c48fc97e48ce69/pkg/api/validation/dynakube/dynakube_name.go#L20)

waodim: fair enough 😅 I guess `dyna.kubes` is the best possible way

pkg/controllers/csi/config.go
```diff
@@ -31,6 +31,7 @@ const (
 	OverlayWorkDirPath   = "work"
```
waodim, line 31: It works for me: I had a `dk` called `codemodules` and ran into problems regarding `oneAgent` pods (some files where not there), then I updated the operator version to this branch. I had to manually restart the `oneAgent` pods and then it worked. Do we want it this way?

pkg/controllers/csi/config.go
andriisoldatenko: what about `_dynakubes` or `_shared`?

pkg/controllers/csi/driver/server.go
```diff
@@ -223,6 +224,18 @@ func (srv *Server) unmount(volumeInfo csivolumes.VolumeInfo) {
 		}
 
 		for _, path := range needsCleanUp {
+			podInfoSymlinkPath := srv.findPodInfoSymlink(volumeInfo) // cleaning up the pod-info symlink here is far more efficient instead of having to walk the whole fs during cleanup
+			if podInfoSymlinkPath != "" {
+				_ = srv.fs.Remove(podInfoSymlinkPath)
+
+				podInfoSymlinkDir := filepath.Dir(podInfoSymlinkPath)
+
+				isEmpty, _ := srv.fs.IsEmpty(podInfoSymlinkDir)
```
andriisoldatenko, line 233:
```suggestion
				if isEmpty, _ := srv.fs.IsEmpty(podInfoSymlinkDir); isEmpty {
				    _ = srv.fs.Remove(podInfoSymlinkDir)
				}
```
you just need it once, so worth to combine IMO together.

0sewa0: done

aorcholski (commented):
csi image :
```
    image: quay.io/dynatrace/dynatrace-operator:snapshot-fix-csi-dk-name-conflicts
```

I deleted `codemodules`  dynakube

filesystem :
```
bash-5.1# ls -l /data
total 12
drwxrwxrwx 2 root root 4096 Feb 11 12:17 appmounts
drwxr-xr-x 4 root root 4096 Feb 11 10:21 dyna.kubes
drwxr-xr-x 2 root root 4096 Feb 11 11:06 work
```

provisioner log :
```
│ {"level":"info","ts":"2025-02-11T12:21:36.557Z","logger":"csi-provisioner","msg":"reconciling DynaKube","namespace":"dynatrace","dynakube":"codemodules"}│
│ {"level":"info","ts":"2025-02-11T12:21:36.574Z","logger":"csi-cleanup","msg":"removing unknown path","path":"/data/codemodules"}│
│ {"level":"info","ts":"2025-02-11T12:21:36.685Z","logger":"csi-cleanup","msg":"removing unknown path","path":"/data/dynakube"}│
│ {"level":"info","ts":"2025-02-11T12:21:36.692Z","logger":"csi-cleanup","msg":"failed to list the shared binaries directory, skipping unused binaries cleanup"}
```

Is "/data/codemodules" `unknown` or `shared binaries` directory?

pkg/controllers/csi/metadata/path_resolver.go
```diff
@@ -40,7 +48,7 @@ func (pr PathResolver) LatestAgentBinaryForDynaKube(dynakubeName string) string
 }
 
 func (pr PathResolver) AgentTempUnzipRootDir() string {
-	return filepath.Join(pr.RootDir, "tmp_zip")
+	return pr.Base("tmp_zip")
```
andriisoldatenko, line 51: wdyt if we move it to constants?

0sewa0: I would leave it as is for now, this supposed to be a "quick" bugfix 😅

andriisoldatenko (dismissed):

andriisoldatenko (approved):

PR #4386, opened 2025-01-30
Cherry-pick: Change readonly oneAgent + custom image validation to a warning #4346

waodim:
## Description

same as #4346

## How can this be tested?

same as in #4346

PR #4426, opened 2025-02-07
Change readonly oneAgent + custom image validation to a warning

waodim:
## Description

In #4346 we implemented a validation that rejected `dynakubes` that have configured a custom image and a mode that uses `readOnly` `oneAgents`. After some discussion we came to the conclusion that an error is too harsh and now we only give a warning.

## How can this be tested?

Same as in #4346 with the difference that now the `dk` should not be rejected but allowed with a warning.


0sewa0 (approved):

0sewa0:
don't forget to update https://github.com/Dynatrace/dynatrace-operator/pull/4386

PR #4435, opened 2025-02-10
Remove obsolete logAgent mounts under `var/log/` (release-1.4)

waodim:
## Description

Fix #4431 as there are some volume mounts that are obsolete, as we have a mount at the parent directory `var/log/`  

## How can this be tested?

same as in #4431

### DIFF
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
@@ -94,7 +94,6 @@ func getIngestVolumeMounts() []corev1.VolumeMount {
 		{
 			Name:      varLogsVolumeName,
 			MountPath: varLogsVolumePath,
-			ReadOnly:  true,
 		},
 	}
 }
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
@@ -22,14 +22,10 @@ const (
 	dtLogVolumeMountPath = "/var/log/dynatrace"
 
 	// for the logs that the logmonitoring will ingest
-	podLogsVolumeName       = "var-log-pods"
-	podLogsVolumePath       = "/var/log/pods"
-	dockerLogsVolumeName    = "docker-container-logs"
-	dockerLogsVolumePath    = "/var/lib/docker/containers"
-	containerLogsVolumeName = "container-logs"
-	containerLogsVolumePath = "/var/log/containers"
-	journalLogsVolumeName   = "var-log"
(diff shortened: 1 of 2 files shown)


0sewa0 (changes requested):
Doesn't work, probably the [previous PR](https://github.com/Dynatrace/dynatrace-operator/pull/4431) already didn't work

```
Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process:
 error during container init: error mounting "/var/lib/kubelet/pods/0d1fff74-b76b-4eaf-9f5e-4d20f21f9950/volumes/kubernetes.io~empty-dir/dynatrace-logs" to rootfs at "/var/log/dynatrace": create mount destination f
or /var/log/dynatrace mount: mkdirat /run/containerd/io.containerd.runtime.v2.task/k8s.io/main/rootfs/var/log/dynatrace: read-only file system: unknown
```

0sewa0:
Problem:
```
    volumeMounts:
    - mountPath: /var/log/dynatrace
      name: dynatrace-logs
    - mountPath: /var/log
      name: var-log
      readOnly: true
```

these 2 conflict...

`/var/log/dynatrace` is a subdir of `/var/log`, but `/var/log` is `readonly` so `/var/log/dynatrace` can't be created ...

this was not a problem previously as all other log dirs (so `/var/log` subdirs) were mounted 1-by-1

andriisoldatenko:
Same here for release-1.4 (92c2f08a4)
```

  Warning  Failed     6s (x2 over 7s)  kubelet            Error: failed to create containerd task: failed to create shim task: OCI runtime crea
te failed: runc create failed: unable to start container process: error during container init: error mounting "/var/lib/kubelet/pods/b6753a79-1
ee4-41dc-ae26-56e10933e6c4/volumes/kubernetes.io~empty-dir/dynatrace-logs" to rootfs at "/var/log/dynatrace": mkdir /run/containerd/io.containe
rd.runtime.v2.task/k8s.io/main/rootfs/var/log/dynatrace: read-only file system: unknown

```

pkg/controllers/dynakube/logmonitoring/daemonset/volumes.go
```diff
@@ -22,14 +22,10 @@ const (
 	dtLogVolumeMountPath = "/var/log/dynatrace"
```
waodim, line 22: Can confirm this way it does not work. Anyone has an idea how we should proceed from now on? Because I think this mount is a requirement that was given to usw

waodim:
Closing it as requirements changed. Will go back to the state before dc46a3de6b2034e9809582f698e0b73c2835e313

PR #4494, opened 2025-02-20
Creates AG TLS certificate (pem) if trustedCAs are used and tlsSecretName is empty

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-1357)

## Description

### Operator

Operator creates ActiveGate TLS certificate (using PEM format) if TrustedCAs certificates are used. IP address of the AG service is also added to the AlternativeNames.

It solves the problem with OneAgents which use single `custom.pem` file to store TrustedCAs and AG certificates. A common scenario is that custom DynatraceServer certificate is  provided forcing the creation of AG certificate because OA verifies all TLS certificates if `custom.pem` file exists.

OSAgent (running in the OneAgent POD) needs IP address to connect to AG because is not able to resolve `<name>.<namespace>` hostname.

The `feature.dynatrace.com/automatic-tls-certificate` opt-in feature flag has to be used because AG>=1.307 only supports PEM format.

### Pipeline
regarding
- hack/doc/gen_e2e_features.sh
- hack/doc/gen_feature_flags.sh

`echo "..\n.."` is not interpreted as new line on all linux distributions so a multiline string is defined in one line using quotes and line continuation.


## How can this be tested?

### e2e test: `codemodules.WithProxyCA`
```
make test/e2e/istio
```

### Manual test

1) create a configmap with TrustedCAs certificate (any certificate, only existence is important)

2) install CloudNative dynakube with 
```
metadata:
  annotations:
    feature.dynatrace.com/automatic-tls-certificate: "true"
spec:
  apiUrl: ...
  activeGate:                                                                                                                                                                                                                                                                                                                                                                                 
    capabilities:                                                                                                                                                                                                                                                                                                                                                                             
    - routing  
  extensions: {}
  networkZone: <abc>
  oneAgent:
    cloudNativeFullStack: {}
  templates:
    extensionExecutionController:
      imageRef:
        repository: public.ecr.aws/dynatrace/dynatrace-eec
        tag: 1.305.112.20250217-161735
  trustedCAs: <configmap> 
```

3) check if `dynakube-activegate-tls-secret` secret exists

4) check if AG cert is mounted to AG pod
```
 - name: server-certs
      secretName: dynakube-activegate-tls-secret 
```

5) Check if AG is using the PEM certificate (internally AG converts it to p12)
```
2025-02-24 15:37:19 UTC INFO	[entrypoint.sh] Setting custom SSL PEM certificate
2025-02-24 15:37:20 UTC INFO	[agctl] Creating keystore from PEM certificate
...
2025-02-24 15:37:24 UTC INFO    [<tec31546>] [<WebServer>, WebServerFactory] Starting Web Server...
2025-02-24 15:37:25 UTC INFO    [<tec31546>] [<WebServer>, WebServerFactory] Using SSL keystore file /var/lib/dynatrace/gateway/config/../ssl/server.p12
```

6) check if AG cert is mounted to Extensions pod
```
 - name: server-certs
      secretName: dynakube-activegate-tls-secret 
```

7) check if Extensions talk to AG
```
│      AG host:                       dynakube-activegate.dynatrace.svc.cluster.local 
...
│ 2025-02-24 15:26:51.766 UTC [00000010] info    [native] Logs ingest limits configuration arrived:
│      -MaxContentLength: 65536
│      -MaxAttributeLength: 250 
```

8) check osagent log in OA pod (choose the newest ruxitagent _host file)
```
bash-4.4$ cat /mnt/volume_storage_mount/host_root/var/log/dynatrace/oneagent/os/ruxitagent_host_310786.0.log | grep "\[comm  \]"
2025-02-24 16:00:59.927 UTC [0004be16] info    [comm  ] Connected to https://<AG serviceIP>:443/communication
```

9) install sample app

10) check agent log in the app pod (for my java app)
```
cat /opt/dynatrace/oneagent-paas/log/java/ruxitagent-....log | grep "\[comm  \]"
2025-02-24 16:15:20.566 UTC [00000007] info    [comm  ] Connected to https://dynakube-activegate.dynatrace:443/communication
```

### DIFF
--- a/doc/api/feature-flags.md
+++ b/doc/api/feature-flags.md
@@ -35,7 +35,12 @@ const (
     AnnotationFeatureAutomaticK8sApiMonitoringClusterName = AnnotationFeaturePrefix + "automatic-kubernetes-api-monitoring-cluster-name"
     AnnotationFeatureK8sAppEnabled                        = AnnotationFeaturePrefix + "k8s-app-enabled"
 
+    AnnotationFeatureActiveGateAutomaticTLSCertificate = AnnotationFeaturePrefix + "automatic-tls-certificate"
+
     AnnotationFeatureNoProxy = AnnotationFeaturePrefix + "no-proxy"
+
+    AnnotationFeatureOneAgnetIgnoreAgTlsCertificate = AnnotationFeaturePrefix + "oneagent-ignore-ag-tls-certificate"
+    AnnotationFeatureOneAgnetIgnoreTrustedCAs       = AnnotationFeaturePrefix + "oneagent-ignore-trusted-cas"
 
     // Deprecated: AnnotationFeatureOneAgentIgnoreProxy use AnnotationFeatureNoProxy instead.
     AnnotationFeatureOneAgentIgnoreProxy = AnnotationFeaturePrefix + "oneagent-ignore-proxy"
@@ -74,7 +79,7 @@ const (
 
 <a name="MountAttemptsToTimeout"></a>
 
-## func [MountAttemptsToTimeout](<https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/api/v1beta3/dynakube/tmp/feature_flags.go#L261>)
+## func [MountAttemptsToTimeout](<https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/api/v1beta3/dynakube/tmp/feature_flags.go#L271>)
 
 ```go
 func MountAttemptsToTimeout(maxAttempts int) string
--- a/doc/api/feature-flags.md
+++ b/doc/api/feature-flags.md
@@ -39,9 +39,6 @@ const (
 
     AnnotationFeatureNoProxy = AnnotationFeaturePrefix + "no-proxy"
 
-    AnnotationFeatureOneAgnetIgnoreAgTlsCertificate = AnnotationFeaturePrefix + "oneagent-ignore-ag-tls-certificate"
-    AnnotationFeatureOneAgnetIgnoreTrustedCAs       = AnnotationFeaturePrefix + "oneagent-ignore-trusted-cas"
-
     // Deprecated: AnnotationFeatureOneAgentIgnoreProxy use AnnotationFeatureNoProxy instead.
     AnnotationFeatureOneAgentIgnoreProxy = AnnotationFeaturePrefix + "oneagent-ignore-proxy"
 
@@ -79,7 +76,7 @@ const (
 
 <a name="MountAttemptsToTimeout"></a>
 
-## func [MountAttemptsToTimeout](<https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/api/v1beta3/dynakube/tmp/feature_flags.go#L271>)
+## func [MountAttemptsToTimeout](<https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/api/v1beta3/dynakube/tmp/feature_flags.go#L268>)
 
 ```go
 func MountAttemptsToTimeout(maxAttempts int) string
--- a/pkg/api/v1beta3/dynakube/activegate/props.go
+++ b/pkg/api/v1beta3/dynakube/activegate/props.go
@@ -102,7 +102,7 @@ func (ag *Spec) IsMetricsIngestEnabled() bool {
 }
 
 func (ag *Spec) IsAutomaticTlsSecretEnabled() bool {
-	return ag.TlsSecretName == "" && ag.trustedCAs != "" && ag.automaticTLSCertificateEnabled
+	return ag.automaticTLSCertificateEnabled
 }
 
 func (ag *Spec) HasCaCert() bool {
--- a/pkg/api/v1beta3/dynakube/activegate/props.go
+++ b/pkg/api/v1beta3/dynakube/activegate/props.go
@@ -24,10 +24,6 @@ func (ag *Spec) SetName(name string) {
 	ag.name = name
 }
 
-func (ag *Spec) SetTrustedCAs(name string) {
-	ag.trustedCAs = name
-}
-
 func (ag *Spec) SetAutomaticTLSCertificate(enabled bool) {
 	ag.automaticTLSCertificateEnabled = enabled
 }
--- a/pkg/api/v1beta3/dynakube/activegate/props.go
+++ b/pkg/api/v1beta3/dynakube/activegate/props.go
@@ -10,6 +10,7 @@ import (
 
 const (
 	TenantSecretSuffix            = "-activegate-tenant-secret"
+	TlsSecretSuffix               = "-activegate-tls-secret"
 	ConnectionInfoConfigMapSuffix = "-activegate-connection-info"
 	AuthTokenSecretSuffix         = "-activegate-authtoken-secret"
 	DefaultImageRegistrySubPath   = "/linux/activegate"
@@ -21,6 +22,10 @@ func (ag *Spec) SetApiUrl(apiUrl string) {
 
 func (ag *Spec) SetName(name string) {
 	ag.name = name
+}
+
+func (ag *Spec) SetTrustedCAs(name string) {
+	ag.trustedCAs = name
 }
 
 func (ag *Spec) SetExtensionsDependency(isEnabled bool) {
@@ -93,7 +98,7 @@ func (ag *Spec) IsMetricsIngestEnabled() bool {
 }
 
 func (ag *Spec) HasCaCert() bool {
-	return ag.IsEnabled() && ag.TlsSecretName != ""
+	return ag.IsEnabled() && (ag.TlsSecretName != "" || ag.TlsSecretName == "" && ag.trustedCAs != "")
 }
 
 // ActivegateTenantSecret returns the name of the secret containing tenant UUID, token and communication endpoints for ActiveGate.
@@ -104,6 +109,14 @@ func (ag *Spec) GetTenantSecretName() string {
 // ActiveGateAuthTokenSecret returns the name of the secret containing the ActiveGateAuthToken, which is mounted to the AGs.
 func (ag *Spec) GetAuthTokenSecretName() string {
 	return ag.name + AuthTokenSecretSuffix
+}
+
+func (ag *Spec) GetTlsSecretName() string {
+	if ag.TlsSecretName != "" {
+		return ag.TlsSecretName
+	}
+
+	return ag.name + TlsSecretSuffix
 }
 
 func (ag *Spec) GetConnectionInfoConfigMapName() string {
--- a/pkg/api/v1beta3/dynakube/activegate/props.go
+++ b/pkg/api/v1beta3/dynakube/activegate/props.go
@@ -28,6 +28,10 @@ func (ag *Spec) SetTrustedCAs(name string) {
 	ag.trustedCAs = name
 }
 
+func (ag *Spec) SetAutomaticTLSCertificate(enabled bool) {
+	ag.automaticTLSCertificateEnabled = enabled
+}
+
 func (ag *Spec) SetExtensionsDependency(isEnabled bool) {
 	ag.enabledDependencies.extensions = isEnabled
 }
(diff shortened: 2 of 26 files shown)


doc/api/feature-flags.md
```diff
@@ -35,8 +35,13 @@ const (
     AnnotationFeatureAutomaticK8sApiMonitoringClusterName = AnnotationFeaturePrefix + "automatic-kubernetes-api-monitoring-cluster-name"
     AnnotationFeatureK8sAppEnabled                        = AnnotationFeaturePrefix + "k8s-app-enabled"
 
+    AnnotationFeatureActiveGateAutomaticTLSCertificate = AnnotationFeaturePrefix + "automatic-tls-certificate"
+
     AnnotationFeatureNoProxy = AnnotationFeaturePrefix + "no-proxy"
 
+    AnnotationFeatureOneAgnetIgnoreAgTlsCertificate = AnnotationFeaturePrefix + "oneagent-ignore-ag-tls-certificate"
+    AnnotationFeatureOneAgnetIgnoreTrustedCAs       = AnnotationFeaturePrefix + "oneagent-ignore-trusted-cas"
```
0sewa0, line 43: why are these needed?

what scenarios are there?

0sewa0: my main reasoning: why have certificates if they are just ignored?

pkg/api/v1beta3/dynakube/activegate/props.go
```diff
@@ -92,20 +101,37 @@ func (ag *Spec) IsMetricsIngestEnabled() bool {
 	return ag.IsMode(MetricsIngestCapability.DisplayName)
 }
 
+func (ag *Spec) IsAutomaticTlsSecretEnabled() bool {
+	return ag.TlsSecretName == "" && ag.trustedCAs != "" && ag.automaticTLSCertificateEnabled
```
0sewa0, line 105: why do you need `trustedCAs` for creating the automatic TLS cert creation?

doc/api/feature-flags.md
StefanHauth: We had lot's of troubles with this certificate conundrum in the past. Having troubleshooting tools available is very valuable.

pkg/api/v1beta3/dynakube/activegate/props.go
StefanHauth: `trustedCAs` cause OAs to be picky about certificates, that's when the use of AG certificates becomes mandatory for OA communication to work. If no trustedCAs are used, we probably still could create the certs, but it would also work without.

pkg/api/v1beta3/dynakube/activegate/props.go
```diff
@@ -92,20 +101,37 @@ func (ag *Spec) IsMetricsIngestEnabled() bool {
...
 }
 
-// ActiveGateAuthTokenSecret returns the name of the secret containing the ActiveGateAuthToken, which is mounted to the AGs.
+// GetAuthTokenSecretName returns the name of the secret containing the ActiveGateAuthToken, which is mounted to the AGs.
 func (ag *Spec) GetAuthTokenSecretName() string {
 	return ag.name + AuthTokenSecretSuffix
 }
 
+// GetTlsSecretName returns the name of the AG TLS secret.
+func (ag *Spec) GetTlsSecretName() string {
```
andriisoldatenko, line 123:
```suggestion
func (ag *Spec) GetTLSSecretName() string {
```
i think we decided in past to use TLS instead of Tls.

pkg/api/v1beta3/dynakube/activegate/spec.go
```diff
@@ -114,6 +115,8 @@ type Spec struct {
 
 	enabledDependencies dependencies
 
+	automaticTLSCertificateEnabled bool
```
andriisoldatenko, line 118: for example here you are using TLS, not cameCase Tls.

pkg/controllers/dynakube/activegate/internal/tls/conditions.go
```diff
@@ -0,0 +1,5 @@
+package tls
+
+const (
+	conditionType = "TlsSecret"
```
andriisoldatenko, line 4:
```suggestion
	conditionType = "TLSSecret"
```

andriisoldatenko: or `TLSsecret`

pkg/controllers/dynakube/activegate/internal/tls/reconciler.go
```diff
@@ -0,0 +1,171 @@
...
+	return &Reconciler{
+		client:       client,
+		dk:           dk,
+		apiReader:    apiReader,
+		timeProvider: timeprovider.New(),
+	}
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context) error {
+	if r.dk.ActiveGate().IsEnabled() && r.dk.ActiveGate().IsAutomaticTlsSecretEnabled() {
```
andriisoldatenko, line 53:
```suggestion
	if r.dk.ActiveGate().IsEnabled() && r.dk.ActiveGate().IsAutomaticTlsSecretEnabled() {
```
naming a bit confusing, we use IsEnabled, and then IssomethingEnabled, I prefer to stay with SomethingIsEnabled

wdyt?

pkg/controllers/dynakube/activegate/internal/tls/reconciler.go
```diff
@@ -0,0 +1,171 @@
...
+
+		return err
+	}
+
+	conditions.SetSecretCreated(r.dk.Conditions(), conditionType, secret.Name)
+
+	return nil
+}
+
+func getCertificateAltNames(dkName string) []string {
```
andriisoldatenko, line 155: we have a similar function in extensions, maybe we can move it to util or so, https://github.com/dynatrace/dynatrace-operator/blob/c827bd288b91d9ef79f1a60fac19e34f39ed56e9/pkg/controllers/dynakube/extension/tls/reconciler.go#L142
not a blocker just an idea.

pkg/controllers/dynakube/activegate/internal/tls/reconciler.go
```diff
@@ -0,0 +1,171 @@
...
+		dkName + activeGateSelfSignedTLSCommonNameSuffix + ".svc",
+		dkName + activeGateSelfSignedTLSCommonNameSuffix + ".svc.cluster.local",
+	}
+}
+
+func getCertificateAltIPs(ips []string) []net.IP {
+	altIPs := []net.IP{}
+
+	for _, ip := range ips {
+		altIPs = append(altIPs, net.ParseIP(ip))
```
andriisoldatenko, line 167: net.ParseIP returns nil if it can't parse it, may be it's better to add small check here?

test/features/cloudnative/container.go
```diff
@@ -63,11 +63,11 @@ func checkSampleContainer(sampleApp *sample.App, agCrt []byte, trustedCAs []byte
 	}
 }
 
-func AssessOneAgentContainer(builder *features.FeatureBuilder, agCrt []byte, trustedCAs []byte) {
+func AssessOneAgentContainer(builder *features.FeatureBuilder, agCrt func() []byte, trustedCAs []byte) {
 	builder.Assess("certificates are propagated to OneAgent containers", checkOneAgentContainer(agCrt, trustedCAs))
 }
 
-func checkOneAgentContainer(agCrt []byte, trustedCAs []byte) features.Func {
+func checkOneAgentContainer(agCrt func() []byte, trustedCAs []byte) features.Func {
```
andriisoldatenko, line 70:
```suggestion
func checkOneAgentContainer(agCrtFunc func() []byte, trustedCAs []byte) features.Func {
```
wdyt if we name it as agCrtFunc since we changed type, and it's confusing. Why `trustedCAs` is []byte and agCrt is func.

pkg/api/v1beta3/dynakube/activegate/props.go
chrismuellner: let's not tie this to `trustedCAs` with no other way to use the automatic tls cert.

doc/api/feature-flags.md
chrismuellner: this is not troubleshooting, but just another feature flag that will blow up the support matrix and complicate troubleshooting

0sewa0: there is already enough confusion around the certificates in the DynaKube, this just adds to it.

the confusion:
- `skipCertCheck`: only affects the Operator's communication, not passed to any other deployed component
- `trustedCAs`: affects Operator, OneAgent, ActiveGate.... (probably everything)
- `activeGate.TlsSecretName`: affects ActiveGate, OneAgent

the confusion will grow, the more components we deploy

**these new feature-flags feel to me that they only enable incorrect setups:**
- `automatic-tls-certificate: true` + `oneagent-ignore-ag-tls-certificate: false`
   - wont this just not work? The AG has the CERT the OA does not, can they communicate?
- `trustedCAs: <something>` +   `oneagent-ignore-trusted-cas: true`
  -   wont this just not work? The AG has the CERT the OA does not, can they communicate?

pkg/controllers/dynakube/activegate/internal/tls/reconciler.go
aorcholski: I think we don't want to duplicate ActiveGate name in condition `r.dk.ActiveGate().IsActiveGateEnabled()`.

But regarding AG feature I prefer `IsAutomaticTlsSecretEnabled` (Is<feature>Enabled) 
```
func (ag *Spec) IsRoutingEnabled() bool {
func (ag *Spec) IsApiEnabled() bool {
func (ag *Spec) IsMetricsIngestEnabled() bool {
func (ag *Spec) IsAutomaticTlsSecretEnabled() bool {
```

test/features/cloudnative/container.go
aorcholski: Because agCrt is created and read in runtime.
It shouldn't be evaluated when builder.Assess is called but later when the test-step is actually executed.

pkg/api/v1beta3/dynakube/activegate/props.go
aorcholski: renamed

pkg/controllers/dynakube/activegate/internal/tls/reconciler.go
aorcholski: moved to `certificates` package

pkg/controllers/dynakube/activegate/internal/tls/reconciler.go
aorcholski: added.
It's a serious error (blocks creation of the cert) because it means a bug in Golang libraries or K8S (the source of the IP string is K8S service). There is no point to create the certificate without IP address because OSAgent needs it.

pkg/api/v1beta3/dynakube/activegate/props.go
aorcholski: logic changed. `IsAutomaticTlsSecretEnabled()` doesn't depend on TrustedCAs.

doc/api/feature-flags.md
aorcholski: oneagent-ignore feature flags removed

andriisoldatenko:
@aorcholski could you please update PR description (remove ff flags)?

pkg/controllers/dynakube/extension/eec/statefulset.go
```diff
@@ -312,7 +312,7 @@ func buildContainerVolumeMounts(dk *dynakube.DynaKube) []corev1.VolumeMount {
 		})
 	}
 
-	if dk.Spec.ActiveGate.TlsSecretName != "" {
+	if dk.ActiveGate().HasCaCert() {
```
0sewa0, line 315: Question:
`dk.ActiveGate().HasCaCert()` is for which Cert, it is(currently) only for the `tlsCert`? 
- A saw you added `SetTrustedCAs` to the `activegate` but you don't use `ag.trustedCAs` (other then setting it)

Did you wanted to be something like:
```go
func (ag *Spec) HasCaCert() bool {
	return ag.IsEnabled() && (ag.TlsSecretName != "" || ag.IsAutomaticTlsSecretEnabled() || ag.ag.trustedCAs != "")
}
```

(because you are setting a volume called `activeGateTrustedCertVolumeName`, so I am guessing, that it should also exist if only `trustedCAs` were given)

aorcholski: Extensions needs `tlsSecretName` only. So no changes are needed.
But `trustedCAs` will be removed from `activegate.Spec` because is not needed.

0sewa0 (approved):

PR #4502, opened 2025-02-24
Fix/multiple set host property arguments

StefanHauth:
## Description

Dynakubes with multiple arguments to set host properties for OneAgent are erroneously rejected by the validation webhook:
For example:
```
oneAgent:
  cloudNativeFullstack:
    args:
    - --set-host-property=prop1=val1    
    - --set-host-property=prop2=val2
```
is considerd illegal.

This PR fixes the validation webhook behaviour so that multiple host properties can be set again.

## How can this be tested?

Deploy various Dynakubes with different argument configurations:

These variations shall be accepted:

```
oneAgent:
  cloudNativeFullstack:
    args:
    - --set-host-property=prop1=val1    
    - --set-host-property=prop2=val2

oneAgent:
  cloudNativeFullstack:
    args:
    - --set-host-property=prop1=val1    
    - --set-host-property=prop2=val2
    - --set-host-tag=prop1
    - --set-host-tag=prop2
   

oneAgent:
  cloudNativeFullstack:
    args:
    - --set-server=foobar.com
    - --set-host-property=prop1=val1    
    - --set-host-property=prop2=val2
    - --set-host-tag=prop1
    - --set-host-tag=prop2
```

These variations shall be denied:
```
oneAgent:
  cloudNativeFullstack:
    args:
    - --set-server=foobar.com
    - --set-server=johndoe.com
    - --set-host-property=prop1=val1    
    - --set-host-property=prop2=val2
    - --set-host-tag=prop1
    - --set-host-tag=prop2
   

oneAgent:
  cloudNativeFullstack:
    args:
    - --set-server=foobar.com
    - --set-host-property=prop1=val1    
    - --set-host-property=prop2=val2
    - --set-host-tag=prop1
    - --set-host-tag=prop1
```

Be creative in your testing, the only arguments that are allowed multiple times are `--set-host-property` and `--set-host-tag`, where the latter is not allowed to have the same value twice. All other arguments are allowed only once.

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-5603)


pkg/api/v1beta3/dynakube/oneagent_props.go
```diff
@@ -292,16 +292,14 @@ func (dk *DynaKube) HostGroupAsParam() string {
 }
 
 func splitArg(arg string) (key, value string) {
-	split := strings.Split(arg, "=")
```
0sewa0, line 295: why not just use `split := strings.SplitN(arg, "=", 2)`

0sewa0 (changes requested):

pkg/api/v1beta3/dynakube/oneagent_props.go
StefanHauth: What's wrong with `Cut`? IMHO its a bit more explicit.

0sewa0: would have been less code change 😅 

but also, doing `if !found` is not different then just doing a `return argName, argValue`, because the `if !found` is true, then `argName` and `argValue` will already be empty, no?

StefanHauth: No, `If sep does not appear in s, cut returns s, "", false.`

0sewa0: Ah, thanks for clarifying.

Are we sure that there are no arguments in the format for "--turn-on-something" ? (ie.: no "=" at all but still valid)

StefanHauth: But there's another issue with this code, it doesn't support arguments without values. Will fix this.

StefanHauth: Fixed in [7c082dd](https://github.com/Dynatrace/dynatrace-operator/pull/4502/commits/7c082ddb28b8627fdfdd9e8e539cc676c4ad0a67)

StefanHauth:
Closing this for now, will open a new one, once we've approval on main.

PR #4541, opened 2025-03-04
Improve CSI server logs

0sewa0:
## Description

An error log of the csi-server, that is logged by the GRPC server is unclear, as it doesn't have the details such as:
- which pod, in which ns has the problem
- which volume-id
- which DK

The problem is that we are bound to use(not really) the `status.Error` to format our msg, so I tried to come up with a "prefix" for such errors based on the `VolumeConfig`.
- Now the error logs are a bit shouty, mainly due to the `volumeID`, but also if the name of the ns/pod/dk is long, the log line will also be.

## How can this be tested?

Deploy a sample app, while the provisioner is still downloading, should see the message.
In a unit-test there is an example how it would look like in a vacuum, here is a live example: 
```
{"level":"error","ts":"2025-03-04T13:20:43.333Z","logger":"csi-driver","msg":"GRPC call failed","method":"NodePublishVolume","full_method":"/csi.v1.Node/NodePublishVolume","error":"rpc error: code = Unavailable desc
 = (codemodules) php-zib50933-oa/php-glibc-oa-7c4c5485b9-246xr:csi-a8c06cc0e0e8c3f9a131573f59834805fe52d7cb254922d98065d78f372b5569 version or digest is not yet set, csi-provisioner hasn't finished setup injection i
s delayed","stacktrace":"github.com/Dynatrace/dynatrace-operator/pkg/controllers/csi/driver.(*Server).Start.grpcLimiter.func2\n\tgithub.com/Dynatrace/dynatrace-operator/pkg/controllers/csi/driver/server.go:322\ngith
ub.com/container-storage-interface/spec/lib/go/csi._Node_NodePublishVolume_Handler\n\tgithub.com/container-storage-interface/spec@v1.11.0/lib/go/csi/csi_grpc.pb.go:1352\ngoogle.golang.org/grpc.(*Server).processUnary
RPC\n\tgoogle.golang.org/grpc@v1.70.0/server.go:1400\ngoogle.golang.org/grpc.(*Server).handleStream\n\tgoogle.golang.org/grpc@v1.70.0/server.go:1810\ngoogle.golang.org/grpc.(*Server).serveStreams.func2.1\n\tgoogle.g
olang.org/grpc@v1.70.0/server.go:1030"}
```

I can't control the stack-trace 😬 , unless we want to move away from the `status.Error` for creating the erros in the `server`

### DIFF
--- a/pkg/controllers/csi/driver/volumes/app/publisher.go
+++ b/pkg/controllers/csi/driver/volumes/app/publisher.go
@@ -60,14 +60,18 @@ func (pub *Publisher) PublishVolume(ctx context.Context, volumeCfg *csivolumes.V
 	}
 
 	if !pub.isCodeModuleAvailable(volumeCfg) {
+		msg := "version or digest is not yet set, csi-provisioner hasn't finished setup injection is delayed"
+
 		return nil, status.Error(
 			codes.Unavailable,
-			"version or digest is not yet set, csi-provisioner hasn't finished setup yet for DynaKube: "+volumeCfg.DynakubeName,
+			volumeCfg.GetErrorMessage(msg),
 		)
 	}
 
 	if err := pub.mountCodeModule(volumeCfg); err != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("failed to mount oneagent volume: %s", err))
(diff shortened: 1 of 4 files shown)


0sewa0:
The log is horrible to read, so will close it and revisit it later with a better designed approach

PR #4554, opened 2025-03-05
[OtelC] Add volume with TLS certificate(s) and customize telemetryService configuration.

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-5928)

## Description

Adds volume with trustedCAs certificates or AG certificate depending on AG type (in-cluster / public).
Adjusts telemetryIngest configuration  so that the Collector uses certificate(s).

## How can this be tested?

unitests


pkg/controllers/dynakube/otelc/certificates/certificates.go
```diff
@@ -0,0 +1,31 @@
+package certificates
+
+import "github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta3/dynakube"
+
+func IsAGCertificateNeeded(dk *dynakube.DynaKube) bool {
+	if isInClusterActiveGate(dk) {
```
andriisoldatenko, line 6:
```suggestion
	if isInClusterActiveGate(dk) && k.ActiveGate().HasCaCert() {
	   return true
	}
	return false
```
no need to use extra if IMO.

pkg/controllers/dynakube/otelc/certificates/certificates.go
```diff
@@ -0,0 +1,31 @@
...
+		if dk.ActiveGate().HasCaCert() {
+			return true
+		}
+	}
+
+	return false
+}
+
+func IsCACertificateNeeded(dk *dynakube.DynaKube) bool {
+	if !isInClusterActiveGate(dk) {
```
andriisoldatenko, line 16: same as previous

pkg/controllers/dynakube/otelc/certificates/certificates.go
andriisoldatenko: why not use otelc_props for this purpose, no need to create new package and pass dk instance.

here -> https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/api/v1beta3/dynakube/otelc_props.go

pkg/otelcgen/exporters.go
```diff
@@ -10,12 +10,13 @@ var (
...
 func (c *Config) buildExporters() map[component.ID]component.Config {
 	serverConfig := &ServerConfig{
-		Endpoint:   c.buildExportersEndpoint(),
-		TLSSetting: &TLSSetting{},
+		Endpoint: c.buildExportersEndpoint(),
 	}
 
 	if c.caFile != "" {
-		serverConfig.TLSSetting.CAFile = c.caFile
+		serverConfig.TLSSetting = &TLSSetting{
```
andriisoldatenko, line 17: it can panic, if TLSSetting is not set, AFAIK, since it pointer.

pkg/controllers/dynakube/otelc/certificates/certificates.go
aorcholski: agree but I would like to emphasize that 

IsAG depends on `isInClusterActiveGate(dk)`
IsCA depends on `! isInClusterActiveGate(dk)`

because of that it never happans that both returns TRUE

pkg/otelcgen/exporters.go
andriisoldatenko: potentially we can update test/ or write new one.

pkg/controllers/dynakube/otelc/certificates/certificates.go
andriisoldatenko: i don't get you here,

andriisoldatenko: if a {
   if b {
   }
}
same as
if a && b {}

Creating a certificate package, in my opinion, is too much. It isn't very clear. It's just what we call few "_props". You don't do anything with certificates or so.

pkg/otelcgen/exporters.go
StefanHauth: We tried that with nil, it seems to work fine. It is used to avoid `tls: {}` in the receiver as well as in the exporter configuration.

StefanHauth: Its the same as here: https://github.com/Dynatrace/dynatrace-operator/blob/6ce9df993a1e259a560145d2a7375ec1191fc1a7/pkg/otelcgen/otelcgen.go#L126

StefanHauth: And the parent struct has an `omitempty` annotation: https://github.com/Dynatrace/dynatrace-operator/blob/6ce9df993a1e259a560145d2a7375ec1191fc1a7/pkg/otelcgen/otelcgen.go#L41

StefanHauth (approved):
LGTM, also tested in real setup

pkg/otelcgen/exporters.go
andriisoldatenko: or i forgot that i did it.

andriisoldatenko (approved):

PR #4562, opened 2025-03-07
Fix exporter endpoint and add unit tests.

StefanHauth:
## Description

The endpoint for the OpenTelemeteryCollector in case of public AG was created wrong. It contained the tenantUUID twice.
Also if `telemetryIngest` was deployed standalone without AG or OA, it was still depending on the status field of AG or OA although it's not needed because the tenantUUID is already part of the apiURL set in the DynaKube

## How can this be tested?

Deploy Dynakube that has only `telemeterIngest` activated and nothing else.

### DIFF
--- a/pkg/controllers/dynakube/otelc/secret/reconciler.go
+++ b/pkg/controllers/dynakube/otelc/secret/reconciler.go
@@ -83,16 +83,17 @@ func (r *Reconciler) ensureTelemetryIngestApiCredentialsSecret(ctx context.Conte
 }
 
 func (r *Reconciler) getDtEndpoint() ([]byte, error) {
-	tenantUUID, err := r.dk.TenantUUID()
-	if err != nil {
-		return nil, err
-	}
 
 	if r.dk.ActiveGate().IsApiEnabled() {
+		tenantUUID, err := r.dk.TenantUUID()
+		if err != nil {
+			return nil, err
+		}
+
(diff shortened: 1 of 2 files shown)


pkg/controllers/dynakube/otelc/secret/reconciler.go
```diff
@@ -83,16 +83,17 @@ func (r *Reconciler) ensureTelemetryIngestApiCredentialsSecret(ctx context.Conte
...
+		tenantUUID, err := r.dk.TenantUUID()
+		if err != nil {
+			return nil, err
+		}
+
 		return []byte(fmt.Sprintf("https://%s-activegate.dynatrace.svc/e/%s/api/v2/otlp", r.dk.Name, tenantUUID)), nil
 	}
 
-	return []byte(fmt.Sprintf("https://%s.%s/api/v2/otlp", tenantUUID, r.dk.ApiUrlHost())), nil
+	return []byte(fmt.Sprintf("https://%s/api/v2/otlp", r.dk.ApiUrlHost())), nil
```
aorcholski, line 95: we should check if ApiUrlHost returns empty string

aorcholski (changes requested):

pkg/controllers/dynakube/otelc/secret/reconciler.go
StefanHauth: The validation webhook makes sure we have a valid API URL, so this should be fine.

aorcholski (dismissed):
LGTM

PR #4584, opened 2025-03-12
Update securityContext for node config collector container

waodim:
## Description

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-4042)

Node config collector should no longer run with a privileged user, therefor we need to update the securityContext of the container.

## How can this be tested?

Deploy kspm and see if everything runs and works.

check the securityContext of the kspm pods, should look like this:

```
securityContext:
  capabilities:
    add:
      - DAC_OVERRIDE
    drop:
      - ALL
  privileged: false
  readOnlyRootFilesystem: true
  runAsGroup: 65532
  runAsNonRoot: true
  runAsUser: 65532
```

### DIFF
--- a/pkg/controllers/dynakube/kspm/daemonset/container.go
+++ b/pkg/controllers/dynakube/kspm/daemonset/container.go
@@ -12,7 +12,7 @@ const (
 	defaultImageTag  = "latest"
 
 	containerName       = "node-config-collector"
-	runAs         int64 = 0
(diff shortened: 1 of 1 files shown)


aorcholski (approved):
LGTM
```
│ 2025/03/12 15:04:12.329183 Info: Collection uploaded to "https://dynakube-activegate.dynatrace/....node-config....".   
```

PR #4593, opened 2025-03-13
Add validation to deny init bootstrapper without code modules image set

gkrenn:
## Description

The init bootstrapper without CSI driver requires a code modules image to be set. 
This PR adds the validation in the webhook.

## How can this be tested?

Apply and should fail:
```
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
  annotations:
    feature.dynatrace.com/remote-image-download: "true"
spec:
  apiUrl: https://ppn33804.dev.dynatracelabs.com/api
  oneAgent:
    applicationMonitoring: {}
      #codeModulesImage: "quay.io/dynatrace/dynatrace-bootstrapper:gktest"
```

Should work: 
```
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
  annotations:
    feature.dynatrace.com/remote-image-download: "true"
spec:
  apiUrl: https://ppn33804.dev.dynatracelabs.com/api

  oneAgent:
    applicationMonitoring:
      codeModulesImage: "quay.io/dynatrace/dynatrace-bootstrapper:gktest"
```

### DIFF
--- a/pkg/api/validation/dynakube/oneagent.go
+++ b/pkg/api/validation/dynakube/oneagent.go
@@ -19,6 +19,8 @@ const (
 
 	errorImageFieldSetWithoutCSIFlag = `The DynaKube specification attempts to enable ApplicationMonitoring mode and retrieve the respective image, but the CSI driver and/or remote image download is not enabled.`
 
+	errorBootstrapperRemoteImageDownloadRequiresCodeModulesImage = `The DynaKube specification enables remote image download, but the code modules image is not set.`
+
 	errorNodeSelectorConflict = `The Dynakube specification conflicts with another Dynakube's OneAgent or Standalone-LogMonitoring. Only one Agent per node is supported.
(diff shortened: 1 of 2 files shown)


0sewa0 (approved):

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -153,6 +155,14 @@ func imageFieldSetWithoutCSIFlag(_ context.Context, v *Validator, dk *dynakube.D
 	return ""
 }
 
+func missingCodeModulesImage(_ context.Context, v *Validator, dk *dynakube.DynaKube) string {
```
waodim, line 158: Not that it is very important in this case but what about unit tests? Should we add one?

gkrenn: Checked for the other features, we don't have any for the oneliners. + We would need to rewrite the csi.isavailable() function to be mockable, as currently it reads the env variables to see which components are enabled.

PR #4599, opened 2025-03-14
Add bootstrapper (CSI-Driver variant) e2e test

waodim:
## Description

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-5587)

With this we add a e2e test that tests the `bootstrapper` in the `CSI-Driver` variant. 

What happens: After the `DynaKube` withe the necessary feature flag is applied and created, `Pods` spawn that are controlled by a `Job`. Those `Pods` download necessary data and configures the `CSI-Driver`. After that is done the Pods are deleted again. I test exactly this.

After that the test is basically the same like the classic `codemodules` test.

## How can this be tested?

run standard e2e tests or `run make test/e2e/applicationmonitoring/bootstrapper-csi`


test/features/bootstrapper/remote_image_download_csi.go
```diff
@@ -0,0 +1,101 @@
...
+				// let's wait not only for the pods to be deleted but rather exactly check if they really reached the state Completed
+				ctx = pod.WaitForCondition(currentPod.Name, currentPod.Namespace, func(object k8s.Object) bool {
+					p, isPod := object.(*corev1.Pod)
+
+					return isPod && len(p.Status.ContainerStatuses) > 0 && p.Status.ContainerStatuses[0].State.Terminated != nil && p.Status.ContainerStatuses[0].State.Terminated.Reason == "Completed"
+				}, time.Minute*5)(ctx, t, envConfig)
+
+				ctx = pod.WaitForPodsDeletionWithOwner(job.Name, job.Namespace)(ctx, t, envConfig)
+			}
+		}
```
0sewa0, line 86: Is this not flacky?

asking because the `provisioner` will just shoot down the `Job` if it sees the folder it wants to see
so it may happen that you do not see Complete but that is not necessarily a problem

also if you accidentally run this test in an istio env, which would only happen in "local" testing, you will never see Complete (I don't know if which container will our container be (I think istio-proxy is the first))

waodim: I was also thinking about only waiting for the pods to be deleted, but I thought that may not be precise enough. WDYT does this suffice? To wait only for the pods to be deleted?

0sewa0: I think it is enough to check that they appear, and that they go away.
and a sample app gets injected successfully

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/4599/commits/90ef1349e62426f2ca489b64ae614d460dad451e

StefanHauth:
`make test/e2e/applicationmonitoring/bootstrapper-csi` ran successfully on my machine.

test/features/bootstrapper/remote_image_download_csi.go
```diff
@@ -0,0 +1,89 @@
...
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+const (
+	bootstrapperImage = "quay.io/dynatrace/dynatrace-bootstrapper:snapshot"
+)
+
+func InstallWithCSI(t *testing.T) features.Feature {
+	builder := features.New("remote-image-download-with-csi")
```
StefanHauth, line 29: The feature has been renamed to "node-image-pull" recently (#4601). This should be reflected in the tests as well.

StefanHauth (changes requested):

test/features/bootstrapper/remote_image_download_csi.go
waodim: yup, already talked to @0sewa0 and I am fixing currently, just needed to merge main to my branch.

0sewa0 (approved):

PR #4639, opened 2025-03-21
Add volume-type annotation

0sewa0:
## Description

[DAQ-6540](https://dt-rnd.atlassian.net/browse/DAQ-6540)

New pod level annotation:
- `oneagent.dynatrace.com/volume-type: "ephemeral" / "csi"`

This will make it that the injection will overrule the `dk.OneAgent().IsCSIAvailable()` check.
- This currently only feasible in the `v2` webhook, because in `v1` we tell the init-container that it will use CSI in a Secret generated by the Operator

However, now we have to generate both init-secrets if: 
- `feature.dynatrace.com/node-image-pull: "true"` is present
- and CSI is available
because with-csi we need the old init-secret, without-csi we need the new init-secret
- we generate and maintain these secrets in the Operator, so we can't do it on a per pod base.

it is a draft, because I want to add unittests:
- they are not failing, but the new scenarios are not tested

## How can this be tested?

- `make deploy` (so you have csi)

- apply dk:
```
kind: DynaKube
metadata:
  name: dk
  namespace: dynatrace
  annotations:
spec:
  apiUrl: https://zib50933.dev.dynatracelabs.com/api
  tokens: dk

  metadataEnrichment:
    enabled: true

  oneAgent:
    cloudNativeFullStack:
```

- apply sample app, with annotation:
```
oneagent.dynatrace.com/volume-type: "ephemeral"
```

Outcome: Your sample app should use the bootstrapper image as its init-container, eventhough there is 

[DAQ-6540]: https://dt-rnd.atlassian.net/browse/DAQ-6540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controllers/dynakube/processmoduleconfigsecret/reconciler.go
```diff
@@ -49,7 +49,11 @@ func NewReconciler(clt client.Client,
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context) error {
-	if !(r.dk.OneAgent().IsCloudNativeFullstackMode() || r.dk.OneAgent().IsApplicationMonitoringMode()) {
+	isNeeded := r.dk.OneAgent().IsCSIAvailable() &&
```
StefanHauth, line 52: Nitpick: move this to a function with a more meaningful name.

pkg/webhook/mutation/pod/common/oneagent/config.go
```diff
@@ -17,6 +17,8 @@ const (
 	// defaults to DefaultInstallPath if not set.
 	AnnotationInstallPath = "oneagent.dynatrace.com/install-path"
 
+	AnnotationVolumeType = "oneagent.dynatrace.com/volume-type"
```
StefanHauth, line 20: Nitpick: isn't there a const for the `oneagent.dynatrace.com` prefix?

pkg/controllers/dynakube/injection/reconciler_test.go
```diff
@@ -444,6 +445,196 @@ func TestSetupEnrichmentInjection(t *testing.T) {
 	})
 }
 
+func TestGenerateCorrectInitSecret(t *testing.T) {
```
StefanHauth, line 448: There's a lot of code duplication in this tests. Would be nice if there's a way to factor parts of it to functions and re-use it (or alternatively do table tests ;-) )

waodim (commented):
Tested it it works. Correct image is used for the initContainer. If all points are adressed I will approve.

pkg/webhook/mutation/pod/common/oneagent/config.go
0sewa0: only for the `oneagent` part

will use that

pkg/controllers/dynakube/injection/reconciler_test.go
0sewa0: I agree, I previously tried to make it into a table test, but it became too convoluted for it's own good. So a scrapped it.
I will try to reduce the duplication.

pkg/controllers/dynakube/processmoduleconfigsecret/reconciler.go
StefanHauth: From the function name I would expect a bit more than "IsNeeded", e.g. what stating what it is that is needed.

StefanHauth (approved):
Did not have the time to test it, otherwise LGTM.

pkg/controllers/dynakube/processmoduleconfigsecret/reconciler.go
0sewa0: the `isNeeded` is "scoped" to the `Reconcile` func, so it asnwers the question `is the Reconcile needed`

"Reconcile needs to be done if csi is available and oneagent is configured accordingly"

(or this is my logic behind it 😅 )

PR #4647, opened 2025-03-24
Update dynatrace-bootstrapper to v1.0.2

0sewa0:
## Description

It is mainly a cosmetic change.
Somethings were fixed in v1.0.2 of the dynatrace-bootstrapper, and I knew that it would break a few imports, so I didn't wait for renovate.

## How can this be tested?

No change in actual behavior, we just use less consts, because we import them

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/Dynatrace/dynatrace-operator
 go 1.23.4
 
 require (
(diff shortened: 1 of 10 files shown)


StefanHauth (approved):

waodim (approved):
lgtm

andriisoldatenko (approved):

PR #4651, opened 2025-03-24
Creates AG TLS certificate by default

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-6519)

## Description

AG TLS cert (PEM) is created by default if AG is enabled.

`feature.dynatrace.com/automatic-tls-certificate=false` annotation disables auto-cert feature. 

## How can this be tested?

1) unittests
make go/test

2) e2e tests (istio required)
test/e2e/cloudnative/codemodules-with-proxy-no-certs
test/e2e/cloudnative/codemodules-with-proxy-and-ag-cert
test/e2e/cloudnative/codemodules-with-proxy-and-auto-ag-cert
test/e2e/cloudnative/codemodules-with-proxy-custom-ca-ag-cert
test/e2e/cloudnative/codemodules-with-proxy-custom-ca-auto-ag-cert


doc/api/feature-flags.md
```diff
@@ -34,7 +34,7 @@ const (
     AnnotationFeatureAutomaticK8sApiMonitoringClusterName = AnnotationFeaturePrefix + "automatic-kubernetes-api-monitoring-cluster-name"
     AnnotationFeatureK8sAppEnabled                        = AnnotationFeaturePrefix + "k8s-app-enabled"
 
-    AnnotationFeatureActiveGateAutomaticTLSCertificate = AnnotationFeaturePrefix + "automatic-tls-certificate"
+    AnnotationFeatureActiveGateDisableAutomaticTLSCertificate = AnnotationFeaturePrefix + "disable-automatic-tls-certificate"
```
0sewa0, line 37: do not do this, we used to have these "negative" annotations, but we moved away from it

pkg/api/v1beta3/dynakube/feature_flags.go
```diff
@@ -129,9 +129,9 @@ func (dk *DynaKube) FeatureNoProxy() string {
 	return dk.getFeatureFlagRaw(AnnotationFeatureNoProxy)
 }
 
-// FeatureActiveGateAutomaticTLSCertificate is a feature flag to enable automatic creation of AG TLS certificate if TrustedCAs are used.
-func (dk *DynaKube) FeatureActiveGateAutomaticTLSCertificate() bool {
-	return dk.getFeatureFlagRaw(AnnotationFeatureActiveGateAutomaticTLSCertificate) == truePhrase
+// FeatureActiveGateDisableAutomaticTLSCertificate is a feature flag to disable automatic creation of AG TLS certificate.
+func (dk *DynaKube) FeatureActiveGateDisableAutomaticTLSCertificate() bool {
+	return dk.getFeatureFlagRaw(AnnotationFeatureActiveGateDisableAutomaticTLSCertificate) == truePhrase
 }
```
0sewa0, line 135:
```suggestion
func (dk *DynaKube) FeatureActiveGateAutomaticTLSCertificate() bool {
	return dk.getFeatureFlagRaw(AnnotationFeatureActiveGateAutomaticTLSCertificate) != falsePhrase
}
```

probably, this is the PR (+ test update)
- this just makes the FF `true` by default, and therefore an opt out

similar example:
```go
func (dk *DynaKube) FeatureAutomaticKubernetesApiMonitoring() bool {
	return dk.getFeatureFlagRaw(AnnotationFeatureAutomaticK8sApiMonitoring) != falsePhrase
}
```

0sewa0 (changes requested):

doc/api/feature-flags.md
chrismuellner: @aorcholski please check the updated ticket description

test/features/cloudnative/codemodules/codemodules.go
```diff
@@ -203,7 +206,7 @@ func WithProxyCA(t *testing.T, proxySpec *value.Source) features.Feature {
 		dynakubeComponents.WithIstioIntegration(),
 		dynakubeComponents.WithProxy(proxySpec),
 		dynakubeComponents.WithAnnotations(map[string]string{
-			dynakube.AnnotationFeatureActiveGateAutomaticTLSCertificate: "true",
```
0sewa0, line 206: shouldn't this stay `true`? (or it can be removed)

OR `WithProxyAndAutomaticAGCert` supposed to "replace" it

hack/make/tests/e2e.mk
```diff
@@ -64,20 +64,28 @@ test/e2e/cloudnative/codemodules: manifests/crd/helm
...
 ## Runs CloudNative codemodules-with-proxy e2e test only
 test/e2e/cloudnative/codemodules-with-proxy: manifests/crd/helm
-	go test -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 20m -count=1  ./test/scenarios/istio -args --feature "codemodules-with-proxy" $(SKIPCLEANUP)
+	go test -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 20m -count=1  ./test/scenarios/istio -args --feature "codemodules-with-proxy$$" $(SKIPCLEANUP)
 
 ## Runs CloudNative codemodules-with-proxy-custom-ca e2e test only
-test/e2e/cloudnative/codemodules-with-custom-ca: manifests/crd/helm
-	go test -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 20m -count=1  ./test/scenarios/istio -args --feature "codemodules-with-proxy-custom-ca" $(SKIPCLEANUP)
+test/e2e/cloudnative/codemodules-with-proxy-custom-ca: manifests/crd/helm
+	go test -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 20m -count=1  ./test/scenarios/istio -args --feature "codemodules-with-proxy-custom-ca$$" $(SKIPCLEANUP)
```
0sewa0, line 71: why the `$$`

hack/make/tests/e2e.mk
```diff
@@ -64,20 +64,28 @@ test/e2e/cloudnative/codemodules: manifests/crd/helm
 
 ## Runs CloudNative codemodules-with-proxy e2e test only
 test/e2e/cloudnative/codemodules-with-proxy: manifests/crd/helm
-	go test -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 20m -count=1  ./test/scenarios/istio -args --feature "codemodules-with-proxy" $(SKIPCLEANUP)
+	go test -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 20m -count=1  ./test/scenarios/istio -args --feature "codemodules-with-proxy$$" $(SKIPCLEANUP)
```
0sewa0, line 67: why the `$$`

0sewa0 (commented):
LGTM

only few clarifying questions :D

hack/make/tests/e2e.mk
aorcholski: `$` regex symbol matching the null string at the end of a line - I want single test to be executed only (`A$` for A only, without AB, AC)

`$$` because `$` is "special character" for make comamnd (`$<`, `$%`,...) and make tries to parse `$"` instead of `$` and `"` separately.

In case of single `$`: 
```
bash: -c: line 1: unexpected EOF while looking for matching `"'
```

hack/make/tests/e2e.mk
aorcholski: [see above ](https://github.com/Dynatrace/dynatrace-operator/pull/4651#discussion_r2015916837)

hack/make/tests/e2e.mk
0sewa0: why not rename the 
- `codemodules-with-proxy-and-auto-ag-cert` to `codemodules-auto-ag-cert-with-proxy`
- `codemodules-with-proxy-custom-ca-auto-ag-cert` to `codemodules-auto-ag-cert-with-proxy-custom-ca`

so we don't need the magic :D

test/features/cloudnative/codemodules/codemodules.go
aorcholski: WithProxyCA is not needed, removed

pkg/api/v1beta3/dynakube/feature_flags.go
```diff
@@ -129,9 +129,9 @@ func (dk *DynaKube) FeatureNoProxy() string {
 	return dk.getFeatureFlagRaw(AnnotationFeatureNoProxy)
 }
 
-// FeatureActiveGateAutomaticTLSCertificate is a feature flag to enable automatic creation of AG TLS certificate if TrustedCAs are used.
+// FeatureActiveGateAutomaticTLSCertificate is a feature flag to disable automatic creation of ActiveGate TLS certificate.
```
waodim, line 132: Out of curiosity: Why is this needed here? Wouldn't be a conversion the better approach here?

aorcholski: The FF is present in v1beta3 on the main branch only. It was a "temporary" home for it until v1beta4 was introduced. 
I would like them to be consistent with each other to avoid potential confusion.

0sewa0 (approved):
run the e2e tests, they passed 👍

waodim (approved):
tests ran through. LGTM 👍

PR #4654, opened 2025-03-25
Cleanup /cmd

0sewa0:
## Description
[DAQ-6628](https://dt-rnd.atlassian.net/browse/DAQ-6628)

Removes bunch of bloat in the `/cmd` packages

No more `Provider` this and `Builder` that.
- these were introduced to test the `/cmd`, however, it tested really nothing, we have e2e tests to check if we break something in that area.

Also moved around some packages, closer to where they belong.

I didn't:
- cleanup the `troubleshoot` package's structure
- cleanup the `support_archive` package's structure
both could use some structural improvement

The testing coverage will drop, but that is fine, the tests I removed were testing nothing. (they were testing that we can write mocks in a way so that nothing happens 👏 )

## How can this be tested?

No change in behavior should be experienced 

[DAQ-6628]: https://dt-rnd.atlassian.net/browse/DAQ-6628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/operator/certs.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/pkg/errors"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+func runCertBootstrapper(cfg *rest.Config, namespace string) error {
+	bootstrapManager, err := createCertBootstapperManager(cfg, namespace)
```
waodim, line 17: Is `bootstrap` correct? I always only saw `bootstrapper` as a name. Not that it is that important, just asking out of curiosity.

cmd/operator/certs.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return err
+	}
+
+	ctx, cancelFn := context.WithCancel(context.Background())
+
+	err = certificates.AddBootstrap(bootstrapManager, namespace, cancelFn)
```
waodim, line 24: same as above

cmd/operator/certs.go
0sewa0: it used to be called that, I tried to rename it so it doesn't conflict with the new bootstrapper

0sewa0: renamed it. hopefully better

waodim (approved):
Tested it - operator worked. Also ran some e2e tests.

PR #4658, opened 2025-03-25
Updated golangci to v2

0sewa0:
## Description

golangci-lint has a new v2, the change that affects us is that the `staticcheck` now has `stylecheck` builtin which we didn't have before, this is the `- "-ST*" # this is stylecheck, but has been rolled into staticcheck` line about. 
- I also turned off other new linters

also updated the excluded path according to the new apiVersions:
```
      - pkg/api/v1alpha1/dynakube
      - pkg/api/v1alpha1/edgeconnect
      - pkg/api/v1beta1/dynakube
      - pkg/api/v1beta2/dynakube
      - pkg/api/v1beta3/dynakube
```

Actual code changes: Mainly changes in tests (how to use `assert` lib), and some other smaller fixes

## How can this be tested?

Linters pass and everything works as expected
~(I know about the merge conflict)~


.golangci.yml
```diff
@@ -1,343 +1,318 @@
...
+    generated: lax
+    paths:
+      - pkg/api/v1alpha1/dynakube
+      - pkg/api/v1alpha1/edgeconnect
+      - pkg/api/v1beta1/dynakube
+      - pkg/api/v1beta2/dynakube
+      - pkg/api/v1beta3/dynakube
+      - third_party$
+      - builtin$
+      - examples$
```
andriisoldatenko, line 318: what is examples?

0sewa0: it was migrated, dunno

I only added the `/pkg/api..` stuff (and the `staticcheck` parts), everything else was migrated via `golangci-lint migrate`

andriisoldatenko (approved):

PR #4666, opened 2025-03-26
Fix Settings API server error handler of EdgeConnect client.

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-6704)

## Description

EdgeConnect client properly process Settings API server error. 

## How can this be tested?

unittest


aorcholski:
Will be fixed in v1.5.0

PR #4674, opened 2025-03-31
Uses ConfigMap instead of Secret for DT API endpoint.

StefanHauth:
## Description

The value for `DT_ENDPOINT` doesn't need to be stored in a secret as it's public anyway. With storing it in a ConfigMap we get the benefit, that it's picked up by the support archive automatically.

Also E2E tests are updated to check that the ConfigMap is created.

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-6815)

## How can this be tested?

* deploy a DK with telemetryIngest:
```
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace

spec:
  apiUrl: https://..../api

  telemetryIngest:
    protocols:
    - zipkin
```

* check OTel collector pod, value for env var `DT_ENDPOINT` shall be taken from ConfigMap `dynatrace-telemetry-api-endpoint`
```
    - name: DT_ENDPOINT
      valueFrom:
        configMapKeyRef:
          key: DT_ENDPOINT
          name: dynatrace-telemetry-api-endpoint
```
* pull support archive, make sure the ConfigMap `dynatrace-telemetry-api-endpoint` is contained
* `make test/e2e/telemetryingest`

### DIFF
--- a/pkg/controllers/dynakube/otelc/endpoint/conditions.go
+++ b/pkg/controllers/dynakube/otelc/endpoint/conditions.go
@@ -1,4 +1,4 @@
-package secret
+package endpoint
 
 const (
 	configMapConditionType = "TelemetryIngestApiEndpointConfigMap"
--- a/pkg/controllers/dynakube/otelc/endpoint/conditions.go
+++ b/pkg/controllers/dynakube/otelc/endpoint/conditions.go
@@ -1,5 +1,5 @@
 package endpoint
 
 const (
-	configMapConditionType = "TelemetryIngestApiEndpointConfigMap"
+	configMapConditionType = "OtelpApiEndpointConfigMap"
 )
--- a/pkg/controllers/dynakube/otelc/endpoint/reconciler_test.go
+++ b/pkg/controllers/dynakube/otelc/endpoint/reconciler_test.go
@@ -46,11 +46,11 @@ func TestSecretCreation(t *testing.T) {
 
 		r := NewReconciler(clt, clt, &dk)
 
-		err = r.ensureTelemetryIngestApiEndpointConfigMap(ctx)
+		err = r.ensureOtlpApiEndpointConfigMap(ctx)
 		require.NoError(t, err)
 
 		var apiEndpointConfigMap corev1.ConfigMap
-		err = clt.Get(ctx, types.NamespacedName{Name: consts.TelemetryApiEndpointConfigMapName, Namespace: dk.Namespace}, &apiEndpointConfigMap)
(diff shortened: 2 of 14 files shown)


pkg/controllers/dynakube/otelc/secret/reconciler.go
```diff
@@ -36,101 +36,101 @@ func NewReconciler(client client.Client, apiReader client.Reader, dk *dynakube.D
 
 func (r *Reconciler) Reconcile(ctx context.Context) error {
 	if r.dk.TelemetryIngest().IsEnabled() {
```
0sewa0, line 38: the package is still called `secret`

and the methods in it are a mouthful, almost all are `....TelemetryIngestApiEndpointConfigMap`

so questions:
- is this a `telemetryIngest` specific logic?
  - if so, why not have it under `otelc/telemetryingest/endpoint` or `telemetryingest/configmap`
    - then you could simplify the method names => not everything would need to say `TelemetryIngest` (`ensureConfigMap` or `ensureEndpointConfigMap`)
- should the same be done with `otelc/service/`?  => `otelc/telemetryingest/service/`

0sewa0 (changes requested):

aorcholski (commented):
the configmap is contained in support-archive package
e2e test PASS

pkg/controllers/dynakube/otelc/secret/reconciler.go
StefanHauth: For now it is telemetryIngest specific, but it could get used more generic in the future. What about renaming it to `...OtlpApiEndpoint...`?

pkg/controllers/dynakube/otelc/endpoint/conditions.go
```diff
@@ -0,0 +1,5 @@
+package endpoint
+
+const (
+	configMapConditionType = "TelemetryIngestApiEndpointConfigMap"
```
0sewa0, line 4: nitpick
```suggestion
	configMapConditionType = "OtelpApiEndpointConfigMap"
```

0sewa0 (dismissed):

pkg/controllers/dynakube/otelc/endpoint/reconciler_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+const (
+	testApiToken       = "apiTokenValue"
+	testTenantUUID     = "abc12345"
+	testKubeSystemUUID = "12345"
+)
+
+func TestSecretCreation(t *testing.T) {
```
aorcholski, line 34:
```suggestion
func TestConfigMapCreation(t *testing.T) {
```

aorcholski (approved):

0sewa0 (approved):

PR #4694, opened 2025-04-02
Update samples

chrismuellner:
## Description

* Enable `metadataEnrichment` by default for log monitoring sample
* Enable `metadataEnrichment` by default in all webhook injection samples
* `v1beta3` samples now have same detail (all possible fields) as `v1beta4` samples

## How can this be tested?

Samples can be applied and all fields are valid.


waodim (approved):
lgtm

PR #4703, opened 2025-04-03
Remove noise from CI

0sewa0:
## Description

It removes things from the CI, that are there mainly because of historical reasons
- The `docker-metada` action does exactly 0 work for us, we overrule what it wants to do
   - tried to add it in https://github.com/Dynatrace/dynatrace-operator/commit/ca24754162a3aabbf7ddfdf64756999bcbef561e properly, but might remove if my hack doesn't work
- go build tags are not used, we never need to build with `e2e` as those are for e2e testing, that we never lunch from the binary

**This is mainly a cosmetic change.**

## How can this be tested?

tbd


.github/workflows/release.yaml
```diff
@@ -255,7 +235,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [ prepare ]
     env:
-      IMAGE: gcr.io/${{ secrets.GCR_REPOSITORY_DEPLOYER }}:${{ needs.prepare.outputs.version_with_major_minor }}
+      IMAGE: gcr.io/${{ secrets.GCR_REPOSITORY_DEPLOYER }}:${{ needs.prepare.outputs.version }}
```
andriisoldatenko, line 238: out of curiosity, why you switch back to full version?

andriisoldatenko: For context, we discussed previously and decided it would be nice to have a full version with no prefix (like line 239 and version_with_major_minor).

PR #4721, opened 2025-04-07
WIP: Add built-in GKE Autopilot Allowlistsynchronizer

meibensteiner:
## Description

DAQ-1491
Customers using GKE-Autopilot can apply exceptions to GKE-Autopilot allowing workloads with higher privileges like the standalone logagent and csi driver.
The PR adds an AllowlistSynchronizer (if supported) which automatically applies the Allowlist.
Some assertion were also fixed along the way.

## How can this be tested?

If the helm chart is applied on a GKE Autopilot cluster >= 1.32 an AllowlistSynchronizer CR should be deployed additionally.
This also allows the csi-driver and a standalone logagent to be rolled out, if enabled.


config/helm/chart/default/tests/Common/edgeconnect/serviceaccount-edgeconnect_test.yaml
```diff
@@ -18,4 +18,4 @@ tests:
       rbac.edgeConnect.create: false
     asserts:
       - hasDocuments:
-        count: 0
+          count: 0
```
andriisoldatenko, line 21: I think this change is redundant.

chrismuellner: try it with `count: 1` and check if it fails

config/helm/chart/default/templates/_platform.tpl
```diff
@@ -25,6 +25,15 @@ Auto-detect the platform (if not set), according to the available APIVersions
     {{- end -}}
 {{- end }}
 
+{{/*
+Auto-detect wether or not we need allowlisting on GKE-Autopilot
```
s-quatres, line 29:
```suggestion
Auto-detect whether or not we need allowlisting on GKE-Autopilot
```

config/helm/chart/default/tests/Common/edgeconnect/serviceaccount-edgeconnect_test.yaml
andriisoldatenko: my point was, why do we want to change format of all files?

chrismuellner: Because it's currently wrong and only works by chance. 🙂

@andriisoldatenko You can verify that locally by changing the value of `count` to something other than `0`, running the unit tests and seeing that they still succeed. If you want to not do these changes in this PR feel free to create a separate PR with these fixes - I'm fine with that as well.

andriisoldatenko: nah i'm fine, just a bit surprised about yaml :)

andriisoldatenko:
@meibensteiner build images fails seems bc you're using fork?

chrismuellner (approved):

PR #4742, opened 2025-04-09
Remove dependency to livenessprobe and CSI node driver registrar

aorcholski:
[JIRA livenessprobe](https://dt-rnd.atlassian.net/browse/DAQ-2161)
[JIRA registrar](https://dt-rnd.atlassian.net/browse/DAQ-2162)

## Description

Native `livenessprobe` and  `csi-node-driver-registrar` executable files are part of the operator image.

Operator `csi-livenessprobe` and `csi-registrar` sub-commands are used by default.

## How can this be tested?
1) Unittests
2) Deployment
Compare memory usage of native command and  operator sub-command. According to my tests there is no need to increase the limit.

2a) the default deployment
make deploy
add LOG_LEVEL env vars to CSI containers (daemonset/dynatrace-oneagent-csi-driver):
```
     containers:
    - name: server
      env:
      - name: LOG_LEVEL
        value: debug
    - name: registrar
      env:
      - name: LOG_LEVEL
        value: debug
    - name: liveness-probe
      env:                                                                                                                                                                                                                                                                                                                                                                                  
      - name: LOG_LEVEL                                                                                                                                                                                                                                                                                                                                                                     
        value: debug                                                                                                                                                                                                                                                                                                                                                                        
```
check server/registrar/livenessprobe logs

2b) native commands
 change values.yaml
```
csidriver:
  registrar:
    operatorcmd: false
  livenessprobe:
    operatorcmd: false    
```
make deploy
add LOG_LEVEL env vars to CSI containers
check server/registrar/livenessprobe logs

2c) cross compatibility
```
csidriver:
  registrar:
    operatorcmd: false | true
  livenessprobe:
    operatorcmd: true | false    
```


pkg/controllers/csi/livenessprobe/livenessprobe.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return &Server{
+		csiAddress:   csiAddress,
+		healthPort:   healthPort,
+		driverName:   driverName,
+		probeTimeout: probeTimeoutDuration,
+	}
+}
+
+func (srv *Server) Start(ctx context.Context) error {
```
andriisoldatenko, line 45:
```suggestion
func (s *Server) Start(ctx context.Context) error {
```

pkg/controllers/csi/livenessprobe/livenessprobe_test.go
```diff
@@ -0,0 +1,158 @@
...
+type testCSIServer struct {
+	csi.UnimplementedIdentityServer
+
+	probeTimeout time.Duration
+
+	pluginInfoReqCount int
+	probeReqCount      int
+}
+
+func TestNormalResponse(t *testing.T) {
```
andriisoldatenko, line 34: why we don't use similar approach with mock like original project https://github.com/kubernetes-csi/livenessprobe/blob/master/cmd/livenessprobe/livenessprobe_test.go#L31 ?

pkg/controllers/csi/livenessprobe/livenessprobe.go
```diff
@@ -0,0 +1,138 @@
+package livenessprobe
```
andriisoldatenko, line 1: TBH i don't see any benefits to put it to `pkg/controllers` it's nothing related to controllers IMO.

cmd/csi/livenessprobe/cmd.go
```diff
@@ -0,0 +1,44 @@
...
+		RunE: run(),
+	}
+
+	addFlags(cmd)
+
+	return cmd
+}
+
+func addFlags(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVar(&probeTimeout, "probe-timeout", "9s", "probe timeout")
```
andriisoldatenko, line 29: why string and not time.Second, and why do we want to hardcode it here and not inside helm?

andriisoldatenko:
I did quick look and IMO what's missing:
- metrics and instrumentation - we used it a lot for performance tests in past, in this PR it is missing 
- unit tests similar to original repo based on mock server.

pkg/controllers/csi/livenessprobe/livenessprobe.go
aorcholski: right, moved to `pkg/csi`

pkg/csi/csitest/driver/driver.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"net"
+	"sync"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+// CSIDriverServers is a unified driver component with both Controller and Node
```
andriisoldatenko, line 12: please add link here where we copied / or inspired so in future we can find it easy.

andriisoldatenko (approved):

PR #4745, opened 2025-04-10
Use conditions properly in bootstrapper config reconciler

0sewa0:
## Description

partial fix for [DAQ-7199](https://dt-rnd.atlassian.net/browse/DAQ-7199)

Use a unique Condition for the bootstarpper config secret.
- Only ping the Dynatrace API every now and then (or if the secret is missing)

## How can this be tested?
Deploy dk with `node-image-pull: true` and see that we do not ping the Dynatrace API every reconcile.
Also should fix the "continues reconcile"

[DAQ-7199]: https://dt-rnd.atlassian.net/browse/DAQ-7199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controllers/dynakube/injection/reconciler.go
+++ b/pkg/controllers/dynakube/injection/reconciler.go
@@ -206,7 +206,7 @@ func (r *reconciler) cleanupOneAgentInjection(ctx context.Context) {
 			return
 		}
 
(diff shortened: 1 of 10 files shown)


andriisoldatenko (approved):

andriisoldatenko:
tested in my cluster!

PR #4749, opened 2025-04-10
Rename Helm switch kubernetes monitoring full

StefanHauth:
## Description

The Helm switch `rbac.kubernetesMonitoringFull` is renamed to `preview.fullObjectCoverage`

## How can this be tested?

* Helm unit tests
* Deploy operator with Helm and `--set preview.fullObjectCoverage.enabled=true` and check if ClusterRole(Binding) `kubernetes-monitoring-full` is created
* Deploy operator with Helm without the switch and the switch set to false and check that the ClusterRole is not created

### DIFF
--- a/config/helm/chart/default/values.yaml
+++ b/config/helm/chart/default/values.yaml
@@ -220,7 +220,4 @@ rbac:
   kspm:
     create: true
     annotations: {}
-  kubernetesMonitoringFull:
-    create: false
-    annotations: {}
   supportability: true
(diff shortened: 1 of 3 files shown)


config/helm/chart/default/values.yaml
```diff
@@ -220,7 +220,4 @@ rbac:
   kspm:
     create: true
     annotations: {}
-  kubernetesMonitoringFull:
-    create: false
-    annotations: {}
```
0sewa0, line 225: just to clarify, we don't want the `preview` section in the `values.yaml` for a reason, right?

aorcholski (approved):
LGTM

0sewa0:
tested it:
- `make deploy` -> everything works, no `dynatrace-kubernetes-monitoring-full` ClusterRole
- using `--set preview.fullObjectCoverage.enabled=true` -> everything works, `dynatrace-kubernetes-monitoring-full` ClusterRole is also present

PR #4757, opened 2025-04-14
Rework feature-flag handling

0sewa0:
## Description

Feature-flags are no longer per-version, as that just adds code duplication with no benefit.
- moved it under `api/exp`, ie "Experimental" 

Tried to come up with some naming conventions:
- grouped FFs related to a given feature in separate files: `oneagent`, `activegate`, `injection`,  `csi`
  - each "group" has there annotaion keys public, with the naming format: `OA...Key`, `AG...Key`, `Injection...Key`, `CSI...Key`
  -  tried to format for the method names
- Did this grouping instead of having separate packages, because IMO that would just overcomplicate things, and I would prefer to focus on keeping this package small...

## How can this be tested?

no change in behavior*

(*) I added the `getFeatureFlagBool` helper method, that uses `strconv.ParseBool`, so "true/false" flags can be set according to that.
> It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False


pkg/api/exp/injection.go
```diff
@@ -0,0 +1,91 @@
...
+
+// IgnoreUnknownState is a feature flag that makes the operator inject into applications even when the dynakube is in an UNKNOWN state,
+// this may cause extra host to appear in the tenant for each process.
+func (ff *FeatureFlags) IgnoreUnknownState() bool {
+	return ff.getFeatureFlagBool(InjectionIgnoreUnknownStateKey, false)
+}
+
+// IsInjectionAutomatic controls OneAgent is injected to pods in selected namespaces automatically ("automatic-injection=true" or flag not set)
+// or if pods need to be opted-in one by one ("automatic-injection=false").
+func (ff *FeatureFlags) IsInjectionAutomatic() bool {
```
waodim, line 36:
```suggestion
func (ff *FeatureFlags) IsAutomaticInjection() bool {
```

waodim: Reads better IMO

pkg/api/exp/oneagent.go
```diff
@@ -0,0 +1,68 @@
...
+	OAPrivilegedKey          = FFPrefix + "oneagent-privileged"
+	OASkipLivenessProbeKey   = FFPrefix + "oneagent-skip-liveness-probe"
+
+	// Deprecated: Dedicated field since v1beta2.
+	OASecCompProfileKey = FFPrefix + "oneagent-seccomp-profile"
+
+	OANodeImagePullKey = FFPrefix + "node-image-pull"
+	// OANodeImagePullTechnologiesKey can be set on a Pod or DynaKube to configure which code module technologies to download. It's set to
+	// "all" if not set.
+	OANodeImagePullTechnologiesKey = "oneagent.dynatrace.com/technologies"
```
waodim, line 18: I don't know if the prefix `"oneagent.dynatrace.com"` will be used more often in the future, but if so, it would make sense to create a const for it like you did for `FFPrefix`

waodim (dismissed):
Apart from my nitpicks it otherwise LGTM. Tested it and it works for me so far. You can decide if you want to apply my feedback, otherwise feel free to merge.

pkg/api/exp/injection.go
andriisoldatenko: but i like original version more, it's easy to read to me 🙈

0sewa0: which is the original now 😅

andriisoldatenko:
@0sewa0 could you please update CONTRIBUTING.md with what you just wrote in PR description, I prefer to keep it.

pkg/api/exp/injection.go
waodim: I don't have a strong opinion there - I can live with both :)

pkg/api/exp/flag.go
```diff
@@ -0,0 +1,97 @@
+package exp
+
+import (
+	"strconv"
+	"time"
+)
+
+const (
+	FFPrefix = "feature.dynatrace.com/"
```
andriisoldatenko, line 9:
```suggestion
	FFDynatracePrefix = "feature.dynatrace.com/"
```
or if you saying this prefix is everywhere across codebase? AFAIK we have few others  don't we?

pkg/api/exp/flag.go
```diff
@@ -0,0 +1,97 @@
...
+
+	val, err := strconv.ParseBool(raw)
+	if err != nil {
+		return defaultVal
+	}
+
+	return val
+}
+
+func (ff *FeatureFlags) getFeatureFlagInt(annotation string, defaultVal int) int {
```
andriisoldatenko, line 85:
```suggestion
func (ff *FeatureFlags) getIntOrDefault(annotation string, defaultVal int) int {
```

andriisoldatenko: https://github.com/golang/go/blob/786e62bcd3f03d73ddf0c999780ffe6f1a0319ea/src/net/http/httputil/dump.go#L189 for example, just to me more explicit.

pkg/api/exp/injection.go
andriisoldatenko: based on go.doc:
>**is** **injected** to pods in selected namespaces **automatically**

pkg/api/exp/flag.go
0sewa0: this is the "only" FF prefix as of right now, there is 1 oneagent, but that is an outlier, I would leave this as is for now

pkg/api/exp/oneagent.go
andriisoldatenko: kk.

andriisoldatenko (approved):
LGTM, i ran few manual tests, and e2e tests and it works.

PR #4763, opened 2025-04-15
Prevent to run `helm upgrade` with wrong/missing CRD versions

waodim:
## Description

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-5399)

With this PR we prevent the successful usage of helm upgrade, if no CRD version is available on the cluster and the installCRD flag is set to false.

## How can this be tested?

- run helm upgrade with installCRD set to false. Operator pod should error.


cmd/operator/cmd.go
```diff
@@ -88,6 +91,35 @@ func runLocally(kubeCfg *rest.Config) error {
...
+			versionFound = true
+
+			break
+		}
+	}
+
+	if !versionFound {
+		return errors.New("DynaKube kind not found in group version")
+	}
+
```
0sewa0, line 122:
```suggestion
	groupKind := schema.GroupKind{
		Group: v1beta4.GroupVersion.Group,
		Kind: "DynaKube", // maybe use reflect to get the name of the DynaKube stuct, shrug
	}

	_, err = operatorManager.GetRESTMapper().RESTMapping(groupKind, v1beta4.GroupVersion.Version)
	if err != nil {
		log.Info("missing CRD")
		return err
	}
```

didn't test it, also this only considers the `DynaKube`

you could also do similar with `EdgeConnect`

0sewa0 (changes requested):

aorcholski (commented):
I performed the following steps
- make deploy
- deleted  dynakube CRD
- deleted operator pod
No "DynaKube kind not found in group version" error message in the operator log after it is re-started..

cmd/operator/cmd.go
andriisoldatenko: i'm about to propose same, i like this idea, bc when we bump version we will update this file easy, rather than change string.

Also you can revisit what we do here https://github.com/dynatrace/dynatrace-operator/blob/02ee3154eb669436c4df8f49aaa84486d6f8ed1f/cmd/support_archive/resources.go#L142-L165

andriisoldatenko: +1 for EdgeConnect

andriisoldatenko: should be:

```
	groupVersion := schema.GroupVersion{
		Group:   v1beta4.GroupVersion.Group,
		Version: v1beta4.GroupVersion.Version,
	}
```

andriisoldatenko: regarding name +1 https://pkg.go.dev/reflect#TypeOf

andriisoldatenko:
Also I prefet to move this check and fail somewhere https://github.com/dynatrace/dynatrace-operator/blob/ad4b9942f142321a4cf40939383216bcded73923/cmd/operator/cmd.go#L48-L47 so we can check it before both cases runInPod and run locally as well.

cmd/operator/cmd.go
waodim: Done https://github.com/Dynatrace/dynatrace-operator/pull/4763/commits/15e868540889ad0ee13ac834d9dd72d662226b7c

cmd/operator/cmd.go
```diff
@@ -12,14 +17,17 @@ import (
...
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
 const (
-	use = "operator"
+	use             = "operator"
+	expectedVersion = "v1beta4"
```
0sewa0, line 30: no longer necessary

waodim:
> Also I prefet to move this check and fail somewhere https://github.com/dynatrace/dynatrace-operator/blob/ad4b9942f142321a4cf40939383216bcded73923/cmd/operator/cmd.go#L48-L47 so we can check it before both cases runInPod and run locally as well.
> 
> or even as first step before we even run sub-command inside main.go for example.

I am unsure on how to do that to be honest. And as I error the Pod before starting at all in case of no matching crd version I think it works as we want it. Or do you have a strong opinion there?

cmd/operator/cmd.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/4763/commits/327a4450af6b6b09e14af1f36d36846561090a13

0sewa0 (changes requested):
tested it without applying any CRD, I don't know what is the expected outcome

because [this](https://github.com/Dynatrace/dynatrace-operator/blob/df7c2e68316e25c511e64209b50605e08b29b410/pkg/controllers/certificates/webhook_cert_controller.go#L83) will not allow us to reach the code

which causes the operator to be `Running` but with a bunch of errors

0sewa0:
> > Also I prefet to move this check and fail somewhere [dynatrace/dynatrace-operator@`ad4b994`/cmd/operator/cmd.go#L48-L47](https://github.com/dynatrace/dynatrace-operator/blob/ad4b9942f142321a4cf40939383216bcded73923/cmd/operator/cmd.go#L48-L47) so we can check it before both cases runInPod and run locally as well.
> > or even as first step before we even run sub-command inside main.go for example.
> 
> I am unsure on how to do that to be honest. And as I error the Pod before starting at all in case of no matching crd version I think it works as we want it. Or do you have a strong opinion there?

it is an entrypoint, bit of simple repetition is fine, and [considering my comment](https://github.com/Dynatrace/dynatrace-operator/pull/4763#pullrequestreview-2772841747) you have to move the check either way

```diff
diff --git a/cmd/operator/cmd.go b/cmd/operator/cmd.go
index 14746f48..be366bdf 100644
--- a/cmd/operator/cmd.go
+++ b/cmd/operator/cmd.go
@@ -73,6 +73,17 @@ func runInPod(kubeCfg *rest.Config) error {
 	}
 
 	isOLM := kubesystem.IsDeployedViaOlm(*operatorPod)
+
+	operatorManager, err := createOperatorManager(kubeCfg, namespace, isOLM)
+	if err != nil {
+		return err
+	}
+
+	err = checkCRDs(operatorManager)
+	if err != nil {
+		return err
+	}
+
 	if !isOLM {
 		err = runCertInit(kubeCfg, namespace)
 		if err != nil {
@@ -80,7 +91,7 @@ func runInPod(kubeCfg *rest.Config) error {
 		}
 	}
 
-	return runOperator(kubeCfg, namespace, isOLM)
+	return errors.WithStack(operatorManager.Start(ctrl.SetupSignalHandler()))
 }
 
 func runLocally(kubeCfg *rest.Config) error {
@@ -91,11 +102,7 @@ func runLocally(kubeCfg *rest.Config) error {
 		return err
 	}
 
-	return runOperator(kubeCfg, namespace, false)
-}
-
-func runOperator(kubeCfg *rest.Config, namespace string, isOLM bool) error {
-	operatorManager, err := createOperatorManager(kubeCfg, namespace, isOLM)
+	operatorManager, err := createOperatorManager(kubeCfg, namespace, false)
 	if err != nil {
 		return err
 	}
@@ -105,10 +112,7 @@ func runOperator(kubeCfg *rest.Config, namespace string, isOLM bool) error {
 		return err
 	}
 
-	ctx := ctrl.SetupSignalHandler()
-	err = operatorManager.Start(ctx)
-
-	return errors.WithStack(err)
+	return errors.WithStack(operatorManager.Start(ctrl.SetupSignalHandler()))
 }
 
 func checkCRDs(operatorManager manager.Manager) error {

```

cmd/operator/cmd.go
```diff
@@ -84,17 +102,43 @@ func runLocally(kubeCfg *rest.Config) error {
...
+	_, err := operatorManager.GetRESTMapper().RESTMapping(groupKind, v1beta4.GroupVersion.Version)
 	if err != nil {
+		log.Info("missing expected CRD version for DynaKube", "version", v1beta4.GroupVersion.Version)
+
 		return err
 	}
 
-	ctx := ctrl.SetupSignalHandler()
-	err = operatorManager.Start(ctx)
+	groupKind = schema.GroupKind{
```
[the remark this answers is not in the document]
0sewa0: for `edgeConnect` to be fully "turn-off-able", (CRD and all) we would need a few changes elsewhere, which is out of scope, however, to avoid potential future confusion I think this change is more complete with the check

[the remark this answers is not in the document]
waodim: good catch.

done https://github.com/Dynatrace/dynatrace-operator/pull/4763/commits/75ecfc57d207573ad5fc4f5076d610c8d7c7e0ea

aorcholski (commented):
```
                --install \
                --namespace dynatrace \
                --create-namespace \
                --atomic \
                --set installCRD=false \
                ...
```

```
│ {"level":"info","ts":"2025-04-22T12:34:29.579Z","logger":"operator-command","msg":"missing expected CRD version for DynaKube","version":"v1beta4"}
│ {"level":"info","ts":"2025-04-22T12:34:29.579Z","logger":"main","msg":"no matches for kind \"DynaKube\" in version \"dynatrace.com/v1beta4\""} 
```

aorcholski (commented):
make deploy
```
│ {"level":"info","ts":"2025-04-22T12:49:07.459Z","logger":"install-config","msg":"envvar content read and set","envvar":"modules.json","value":"{\n  \"csiDriver\": true,\n  \"activeGate\": true,\n  \"oneAgent\": true,\n  \"extensions\": true,\n  \"logMonitoring\": true,\n  \"edgeConnect\": true,\n  \"supportability\": true,\n  \"kspm\": tr │
│ {"level":"info","ts":"2025-04-22T12:49:07.459Z","logger":"version","msg":"dynatrace-operator","version":"snapshot-feature-err-no-expected-crd-vers","gitCommit":"f001b0d3a6c2edb8cd59780c116867067e5ace56","buildDate":"2025-04-18T07:10:39+00:00","goVersion":"go1.24.2","platform":"linux/amd64"}                                                  │
│ {"level":"info","ts":"2025-04-22T12:49:07.459Z","msg":"logging level","logLevel":"info"}                                                                                                                                                                                                                                                             │
│ {"level":"info","ts":"2025-04-22T12:49:07.864Z","logger":"main","msg":"error listening on :10080: listen tcp :10080: bind: address already in use"}                                                                                                                                                                                                  │
│ Error: error listening on :10080: listen tcp :10080: bind: address already in use 
```

0sewa0 (approved):
tested with an upgrade
1.0.0 -> this branch + `installCRDs: false` == failed upgrade
1.0.0 -> this branch + `installCRDs: true` == successful upgrade

(used an old version so I don't have to think about which CRD version was introduced when 😄 )

andriisoldatenko (approved):

PR #4770, opened 2025-04-15
Remove dynamic parts from args section in job container

waodim:
## Description

In order to be able to run the bootstrapper with csi driver (which spawns privileged pods - think about allow listing) on gke autopilot, there are no dynamic string sections allowed. This is the case in the `args` section. This PR should fix this.

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-7359)

## How can this be tested?

- Deploy operator with bootstrapper + csi and have a look at the args and env section:
- new envs should be there
- args of `work path` and `target path` should reference to env vars

Sample dynaKube:

```
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
  annotations:
spec:
  apiUrl: https://tenantuuid.dev.dynatracelabs.com/api

  oneAgent:
    applicationMonitoring:
```

**NOTE: as the allowlist for the job is not working yet, you will need to run the bootstrapper WITHOUT CSI driver on GKE Autopilot!**


StefanHauth:
Can't test it on GKE Autopilot, because the created Job requires privileges but we do not yet have an allow-list for it. Otherwise, LGTM.

StefanHauth (approved):

StefanHauth:
Also tested it with CSI driver and manual allow-list now on GKE autopilot. Jobs are created fine. ->LGTM

PR #4775, opened 2025-04-16
Remove obsolete validation warning in case of custom image is used

waodim:
## Description

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-7423)

In 1.4.0 we had the problem that if we had a DynaKube that ran a mode that supports OneAgents that are not readOnly (like for example hostMonitoring + custom image without CSI Driver) the OneAgents would error and stay in an error state because of:

```
    [2025-04-16 13:56:38.463 UTC] [00035ef2] [info   ] [bootstrapper] Started agent deployment as a container, PID 220914
    [2025-04-16 13:56:38.466 UTC] [00035ef2] [error  ] [bootstrapper] Volume storage should be mounted on /mnt/volume_storage_mount with write permissions
```

This is why this validation was added in the first place. By adding `hostPath` Volumes to the DynaKube this problem is fixed and does not need a validation anymore. Which is why I removed it in this PR.

## How can this be tested?

Deploy the operator without CSI Driver and a DynaKube with hostMon and a custom image:

```
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
  annotations:
spec:
  apiUrl: https://tenantuuid.dev.dynatracelabs.com/api
  oneAgent:
    hostMonitoring: 
      image: public.ecr.aws/dynatrace/dynatrace-oneagent:1.305.98.20250114-172617
```

### DIFF
--- a/pkg/api/validation/dynakube/oneagent.go
+++ b/pkg/api/validation/dynakube/oneagent.go
@@ -26,8 +26,6 @@ Use a nodeSelector to avoid this conflict. Conflicting DynaKubes: %s`
 
 	errorVolumeStorageReadOnlyModeConflict = `The DynaKube specification specifies a read-only host file system while OneAgent has volume storage enabled.`
 
-	warningPublicImageWithWrongConfig = `You are using a custom OneAgent image without CSI driver. Only images from the tenant registry are supported."`
-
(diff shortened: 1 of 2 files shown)


0sewa0 (approved):
no code best code

PR #4811, opened 2025-04-25
Setup the cert controller properly

0sewa0:
## Description

There is bug in the webhook certificate management.

We don't renew them in the default scenario.
- This is bad, because they get outdated after 7 days
We renew them in the OLM scenario.
- This is bad, because in case of OLM, the webhook expects different certs, which we break with our renewed ones.

## How can this be tested?

This only comes up in 2 cases:
- OLM deployments (here it will come up quickly)
- longer running deployments, when the generated certs gets outdated

### How to simulate each scenario
These are ways to configure the Operator deployment to let the error surface.

#### OLM
in the `values.yaml` add the olm-annotation to the Operator
```yaml
operator:
  annotations:
   olm.operatorNamespace: beep
```
__IMPORTANT__: in a correct OLM scenario, both the Webhook and Operator would have this annotation, and the Webhook would also behave differently, but setting it on the Webhook now will just fail it constantly. (as we are not actually deploying with OLM)

- this will cause the Operator to not create certs
  - so the `helm install` should fail (as we are not actually deploying via OLM that would handle the certs)
      - on 1.5 this is not that case -> baaad

#### Longer running envs

1. Deploy the Operator normally.
2. Look for the `dynatrace-webhook-certs`
3. Delete `dynatrace-webhook-certs`
4. Delete a pod of the webhook
5. See that the `dynatrace-webhook-certs` was recreated.
   - on 1.5, this doesn't happen -> baad

### According to the Logs

based on @andriisoldatenko's test:

run the following `k logs -n dynatrace deployments/dynatrace-operator | grep "Starting EventSource"`
- look for `webhook-cert-controller` and `webhook-boostrap-controller`
- Incase of OLM install, neither must be present
- Incase of default install, both must be present

### DIFF
--- a/cmd/operator/manager.go
+++ b/cmd/operator/manager.go
@@ -34,7 +34,6 @@ func getControllerAddFuncs(isOLM bool) []controllerSetupFunc {
 	}
 
 	return funcs
-
 }
 
 func createOperatorManager(cfg *rest.Config, namespace string, isOLM bool) (manager.Manager, error) {
--- a/cmd/operator/manager.go
+++ b/cmd/operator/manager.go
@@ -31,7 +31,7 @@ func createOperatorManager(cfg *rest.Config, namespace string, isOLM bool) (mana
 		return nil, errors.WithStack(err)
 	}
 
-	if isOLM {
+	if !isOLM {
 		err = certificates.Add(mgr, namespace)
 		if err != nil {
 			return nil, err
--- a/cmd/operator/manager.go
+++ b/cmd/operator/manager.go
@@ -20,6 +20,23 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 )
 
+type controllerSetupFunc func(manager.Manager, string) error
+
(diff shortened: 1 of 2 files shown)


cmd/operator/manager.go
```diff
@@ -31,7 +31,7 @@ func createOperatorManager(cfg *rest.Config, namespace string, isOLM bool) (mana
 		return nil, errors.WithStack(err)
 	}
 
-	if isOLM {
+	if !isOLM {
```
andriisoldatenko, line 34: Is it possible to unit test it ?

andriisoldatenko (approved):
tested these 4 scenarios:
```
1.5.0 no OLM ( webhook-cert-controller is missing ) bug
1.5.0 with OLM ( webhook-cert-controller is exists ) bug
1.5.1 (patch) no OLM ( webhook-cert-controller must be there missing ) correct
1.5.1 (patch) with OLM ( webhook-cert-controller should not be there ) correct
```
practically LGTM

PR #4818, opened 2025-04-28
Add `cmd/bootstrapper/download` package

0sewa0:
## Description

[DAQ-7404](https://dt-rnd.atlassian.net/browse/DAQ-7404)

Basically, extracts the current `startup` [init-container's zip download logic](https://github.com/Dynatrace/dynatrace-operator/blob/30bb962a3b71abb1cf5ef83d4cf6393717bba467/pkg/injection/startup/run.go#L60-L90) into its own package.
- no more dealing with `env` and some shared `secretConfig`
   - it has its own config, that is expected to be on the FS (mounted from a Secret) 

Will be used in the future replacement of the `startup` package.

## How can this be tested?

If the unittests are not enough, then you can run it locally, by adding this test(fill it with your credentials):
```go
	t.Run("live test", func(t *testing.T) {
		ctx := context.Background()
		inputDir := "./input"
		targetDir := "./target"
		config := Config{
			URL:      "https://***.dev.dynatracelabs.com/api",
			ApiToken: "***",
		}
		fs := afero.Afero{Fs: afero.NewOsFs()}
		props := &url.Properties{
			Os:            dtclient.OsUnix,
			Type:          dtclient.InstallerTypePaaS,
			Flavor:        arch.Flavor,
			Arch:          arch.Arch,
			TargetVersion: url.VersionLatest,
			SkipMetadata:  false,
			PathResolver:  metadata.PathResolver{RootDir: "./"},
		}
		setupConfig(t, &fs, inputDir, config)
		client := New()

		err := client.Do(ctx, fs, inputDir, targetDir, *props)
		require.NoError(t, err)
	})
```

[DAQ-7404]: https://dt-rnd.atlassian.net/browse/DAQ-7404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/bootstrapper/download/client.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer"
+	"github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer/url"
+	"github.com/spf13/afero"
+)
+
+type InstallerBuilder func(fs afero.Fs, dtc dtclient.Client, props *url.Properties) installer.Installer
+type DTClientBuilder func(url, apiToken, paasToken string, opts ...dtclient.Option) (dtclient.Client, error)
+
+type Client struct {
+	newInstaller InstallerBuilder
```
andriisoldatenko, line 19:
```suggestion
	installBuilder InstallerBuilder
```
IMO, struct attribute with `new` a bit confusing.

andriisoldatenko: or may be we can addd importAlias to `codemoduleInstaller codemodule/installer` and use
```
type CLient struct {
     installer InstallerFunc
     dtClient DTClientFunc

}
```
or I saw somewhere pattern like so:
```
type CLient struct {
     installer InstallerOption
     dtClient DTClientOption

}
``` 
bc on line 23 you already have `Option func(*Client)` which is good.

cmd/bootstrapper/download/client.go
```diff
@@ -0,0 +1,97 @@
...
+		client,
+		&props,
+	)
+
+	_, err = oneAgentInstaller.InstallAgent(ctx, targetDir)
+
+	return err
+}
+
+func (cl *Client) createClientFromFs(fs afero.Afero, inputDir string) (dtclient.Client, error) {
```
andriisoldatenko, line 67: this method name a bit confusing to me, because it's method of Client struct and it returns dtclient.Client, so IMO better to name it: `createDTClient`. Regarding "from FS" ( IMO arguments already explaining to us that it's from FS) but I'm okey if we keep it as is: `createDTClientFromFS`

cmd/bootstrapper/download/client.go
0sewa0: in golang, constructors tend to be called `func New() Type` or `func NewType() Type` calling it "builder" sounds wrong to me 
- I know we do it elsewhere, that doesn't mean we should keep doing that
- Considering we have actual "builders", to me calling this a "builder" is just confusing

Also it reads nicer to say:
```go
	oneAgentInstaller := cl.newInstaller(
		fs,
		client,
		&props,
	)
```
than
```go
	oneAgentInstaller := cl. installBuilder(
		fs,
		client,
		&props,
	)
```
it would be different if it looked more like (this is a builder)
```go
	oneAgentInstaller := cl. installBuilder().WithFS(fs).WithDT(client).WithProps(props).Build()
```

to call it `installer` and be correct(so it is not a constuctor func, but an actual `installer.Installer`), then I would need to re-arch the whole thing, because then the caller of `download.New()` would need to create the `dtclient`, so the caller would need to define the `download.Config` and so on.
- IMO, the current way is nicer, because it compartmentalizes more, and reflects the relation between packages  more
  - `cmd/bootstrapper/download` is a sub package of `cmd/bootstrapper`, and at the end of the day, it is optional (only necessary if we want to download the CodeModule as a ZIP) 

I will define a `type NewFunc func(afero.Fs, dtclient.Client, *Properties) installer.Installer` in the `installer/url` package to make it more reusable.

andriisoldatenko: I agree with what you wrote here, the concern was more about struct which has attribute which starts with `new`.
but let's use current impl.

andriisoldatenko (approved):

PR #4832, opened 2025-04-29
Make FIPS image locally buildable

andriisoldatenko:
## Description

resolves https://dt-rnd.atlassian.net/browse/DAQ-6756

Also this change is based on [what we had before](https://github.com/Dynatrace/dynatrace-operator/commit/35bc3675aeda430a8aae112b8f12858edf4e3a4f) for CGO=1 builds.

> !Note: Warning
> By default it will build FIPS images using local platform, bc cross-compile takes long and useless for local debug.

Locally, you can build fips image via:

```
# this will build correct image name with `-fips` suffix for tag
OPERATOR_DEV_BUILD_PLATFORM="linux/arm64" make image/build/fips

# and will deploy this image to your cluster
make deploy/fips
```

first time it will be slow, and then when podman will cache layer, we are good and quick.

> **_NOTE:_**  It took 1h:05mins or my machine to build and image without cache by removing all cache 
> using `podman system prune --all`

## How can this be tested?

build/deploy and run some standard tests, for example `test/e2e/standard`

also you can exec to operator pod and check envs:

```
bash-5.1$ env | grep -i fips
GOFIPS=1
LIBGCRYPT_FORCE_FIPS_MODE=1
bash-5.1$
```


hack/build/build_image.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -x
+
```
andriisoldatenko, line 3:
```suggestion
```

fips.Dockerfile
```diff
@@ -1,6 +1,6 @@
 # check=skip=RedundantTargetPlatform
 # setup build image
-FROM --platform=$TARGETPLATFORM mcr.microsoft.com/oss/go/microsoft/golang:1.24.2-fips-bullseye@sha256:28ab4742d3b5feb0b3c7450629b4e105128d8709dd6ca22898472ab302140c37 AS operator-build
+FROM mcr.microsoft.com/oss/go/microsoft/golang:1.24.2-fips-bookworm@sha256:28ab4742d3b5feb0b3c7450629b4e105128d8709dd6ca22898472ab302140c37 AS operator-build
```
0sewa0, line 3: you could probably remove all the `--platform=$TARGETPLATFORM` stuff, no?

so you can probably also remove 
```
# check=skip=RedundantTargetPlatform
# setup build image
```

andriisoldatenko: good catch, will do.

andriisoldatenko: @0sewa0 fixed in 706e31852

hack/make/images.mk
```diff
@@ -28,6 +28,13 @@ endif
 images/build: ensure-tag-not-snapshot
 	./hack/build/build_image.sh "${IMAGE}" "${TAG}" "${DEBUG}"
 
+## Build an Operator FIPS image with a give IMAGE and TAG
+# because cross-compile takes ~1h, we want to build fips locally only for local architecture
+# so that's why we run it with no_platform=true to build image without platform and fallback decision to podman
+.PHONY:
```
andriisoldatenko, line 34:
```suggestion
```

andriisoldatenko:
> Tried it, it works but wouldn't it make sense to:
> 
> * have a `make` target for it, like `make images/build/fips` or something
>   
>   * similar to https://github.com/Dynatrace/dynatrace-operator/blob/f5c6938cd84f1f602c592df99c544bdf682d0de4/hack/make/deploy/deploy.mk#L8-L9
> * As cross-compiling takes an AGE (1h), I don't see anyone actually using it like that. So I would make the default platfrom the same as the system when building FIPS locally

@0sewa0 tried to fix in 251d2da50

hack/make/images.mk
```diff
@@ -28,6 +28,12 @@ endif
 images/build: ensure-tag-not-snapshot
 	./hack/build/build_image.sh "${IMAGE}" "${TAG}" "${DEBUG}"
 
+## Build an Operator FIPS image with a give IMAGE and TAG
+# because cross-compile takes ~1h, we want to build fips locally only for local architecture
+# so that's why we run it with no_platform=true to build image without platform and fallback decision to podman
```
0sewa0, line 33: `no_platform=true` is no longer a thing

hack/make/images.mk
```diff
@@ -28,6 +28,12 @@ endif
 images/build: ensure-tag-not-snapshot
 	./hack/build/build_image.sh "${IMAGE}" "${TAG}" "${DEBUG}"
 
+## Build an Operator FIPS image with a give IMAGE and TAG
+# because cross-compile takes ~1h, we want to build fips locally only for local architecture
+# so that's why we run it with `OPERATOR_DEV_BUILD_PLATFORM="--platform=linux/arm64" make image/build/fips
```
andriisoldatenko, line 33:
```suggestion
# so that's why the recomended way to run it (assuming local platfrom is arm64) is `OPERATOR_DEV_BUILD_PLATFORM="--platform=linux/arm64" make image/build/fips
```

hack/build/build_image.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -x
```
andriisoldatenko, line 3:
```suggestion
```

hack/build/build_image.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+
```
andriisoldatenko, line 3:
```suggestion
```

0sewa0 (dismissed):
ran `OPERATOR_DEV_BUILD_PLATFORM="linux/arm64" make images/build/fips`
worked

0sewa0 (approved):

PR #4861, opened 2025-05-07
Remove metadata check in v2

0sewa0:
## Description

[DAQ-8230](https://dt-rnd.atlassian.net/browse/DAQ-8230)

In the v2 webhook, if we inject the OneAgent we always do metadata-enrichment
- right now v2 webhook is only called if we inject the OneAgent, so the check is not needed

## How can this be tested?

DK with `node-image-pull` without CSI (or `volume-type: ephemeral` )
- metadata-enrichment is done regardless if `metadata-enrichment` is enabled or not

[DAQ-8230]: https://dt-rnd.atlassian.net/browse/DAQ-8230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


waodim (approved):
`metadata-enrichment.dynatrace.com/injected: true                                                                                                                                               `

Annotation is there

```
/var/lib/dynatrace/enrichment$ ls
dt_metadata.json        dt_metadata.properties  endpoint
```

enrichment dir is there.

Although metdataEnrichment is not enabled.

PR #4869, opened 2025-05-09
Add `enable-ds-eviction` annotation in OneAgent DaemonSet

albertogdd:
https://dt-rnd.atlassian.net/browse/DAQ-5871

## Description

Adds `cluster-autoscaler.kubernetes.io/enable-ds-eviction=false` to OneAgent DaemonSet Pods.

## How can this be tested?

- Deploy DynaKube and check that the OneAgent DaemonSet Pod template has the annotation, and also the Pods.

### DIFF
--- a/pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -151,14 +151,11 @@ func (b *builder) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 	)
 	maxUnavailable := intstr.FromInt(dk.FF().GetOneAgentMaxUnavailable())
 
-	daemonSetAnnotations := map[string]string{
-		annotationEnableDaemonSetEviction: "false",
-	}
-
 	templateAnnotations := map[string]string{
 		annotationUnprivileged:            annotationUnprivilegedValue,
 		webhook.AnnotationDynatraceInject: "false",
 		annotationTenantTokenHash:         dk.Status.OneAgent.ConnectionInfoStatus.TenantTokenHash,
+		annotationEnableDaemonSetEviction: "false",
 	}
 
 	templateAnnotations = maputils.MergeMap(templateAnnotations, b.hostInjectSpec.Annotations)
@@ -168,7 +165,7 @@ func (b *builder) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 			Name:        dk.Name,
 			Namespace:   dk.Namespace,
 			Labels:      labels,
-			Annotations: daemonSetAnnotations,
+			Annotations: map[string]string{},
 		},
 		Spec: appsv1.DaemonSetSpec{
 			Selector: &metav1.LabelSelector{
--- a/pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -151,7 +151,7 @@ func (b *builder) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 	)
 	maxUnavailable := intstr.FromInt(dk.FF().GetOneAgentMaxUnavailable())
 
-	daemonsetAnnotations := map[string]string{
+	daemonSetAnnotations := map[string]string{
 		annotationEnableDaemonSetEviction: "false",
 	}
 
@@ -168,7 +168,7 @@ func (b *builder) BuildDaemonSet() (*appsv1.DaemonSet, error) {
 			Name:        dk.Name,
 			Namespace:   dk.Namespace,
 			Labels:      labels,
-			Annotations: daemonsetAnnotations,
+			Annotations: daemonSetAnnotations,
 		},
 		Spec: appsv1.DaemonSetSpec{
 			Selector: &metav1.LabelSelector{
--- a/pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
+++ b/pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
@@ -20,13 +20,10 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


0sewa0:
I think this should be on the `DaemonSet` and not the Pods the `DaemonSet` creates.
- In `enable-ds-eviction` the `ds` is for `DaemonSet`

pkg/controllers/dynakube/oneagent/daemonset/daemonset_test.go
```diff
@@ -344,7 +344,7 @@ func TestHostMonitoring_SecurityContext(t *testing.T) {
 		dk := dynakube.DynaKube{
 			ObjectMeta: metav1.ObjectMeta{
 				Annotations: map[string]string{
-					exp.OAPrivilegedKey: "true",
+					exp.OAPrivilegedKey: annotationTrueValue,
```
andriisoldatenko, line 347: We discussed before that we want to use string literals in tests, so pls undo this change.

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -20,9 +20,13 @@ import (
...
 
 const (
-	annotationUnprivileged      = "container.apparmor.security.beta.kubernetes.io/dynatrace-oneagent"
+	annotationUnprivileged     = "container.apparmor.security.beta.kubernetes.io/dynatrace-oneagent"
+	annotationTenantTokenHash  = api.InternalFlagPrefix + "tenant-token-hash"
+	annotationEnableDsEviction = "cluster-autoscaler.kubernetes.io/enable-ds-eviction"
+
 	annotationUnprivilegedValue = "unconfined"
-	annotationTenantTokenHash   = api.InternalFlagPrefix + "tenant-token-hash"
+	annotationTrueValue         = "true"
```
andriisoldatenko, line 28:
```suggestion
	StringTrueValue         = "true"
```
IMO it's nothing about annotations, yes we use it there only, but it's just confusing to me

andriisoldatenko: cc @0sewa0 maybe it's better to fix separately, I did a quick search and we have all combinations:
- tokenSecretHashAnnotation
- AnnotationXXX
- justWithoutAnnotation

https://github.com/dynatrace/dynatrace-operator/blob/8c0742c8e8bfd3b25c5eb2bb004ef8f71b387fc6/pkg/controllers/dynakube/kspm/daemonset/volumes.go#L13

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -20,9 +20,13 @@ import (
 )
 
 const (
-	annotationUnprivileged      = "container.apparmor.security.beta.kubernetes.io/dynatrace-oneagent"
+	annotationUnprivileged     = "container.apparmor.security.beta.kubernetes.io/dynatrace-oneagent"
+	annotationTenantTokenHash  = api.InternalFlagPrefix + "tenant-token-hash"
+	annotationEnableDsEviction = "cluster-autoscaler.kubernetes.io/enable-ds-eviction"
```
andriisoldatenko, line 25: what is DS? can we use `annotationEnableDaemonSetEviction`

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
0sewa0: (I always found putting `"true"` in a const to be a bit of an overkill :) )

andriisoldatenko: yes, i did search in k8s/api codebase - they never do it.

so I'm thinking let's be consistent, since we are not changing entire codebase just few files, let's revert back to "true".

albertogdd: I will remove the change then :+1: thanks for the input guys

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
albertogdd: Fixed https://github.com/Dynatrace/dynatrace-operator/pull/4869/commits/3e0096a920b7a5e2f47cf044009feab2e222c54e

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
albertogdd: Fixed https://github.com/Dynatrace/dynatrace-operator/pull/4869/commits/3e0096a920b7a5e2f47cf044009feab2e222c54e

andriisoldatenko:
> I think this should be on the `DaemonSet` and not the Pods the `DaemonSet` creates.
> 
> * In `enable-ds-eviction` the `ds` is for `DaemonSet`

just for reference, based on ticket description: 

https://github.com/dynatrace/dynatrace-operator/blob/a280166bd6fc27e978a5d9e173667446dbc8e2a8/config/helm/chart/default/templates/Common/csi/daemonset.yaml#L39

albertogdd:
> > I think this should be on the `DaemonSet` and not the Pods the `DaemonSet` creates.
> > 
> > * In `enable-ds-eviction` the `ds` is for `DaemonSet`
> 
> just for reference, based on ticket description:
> 
> https://github.com/dynatrace/dynatrace-operator/blob/a280166bd6fc27e978a5d9e173667446dbc8e2a8/config/helm/chart/default/templates/Common/csi/daemonset.yaml#L39

True! Also on the docs, it mentions the annotations belongs to the pod
https://kubernetes.io/docs/reference/labels-annotations-taints/#cluster-autoscaler-kubernetes-io-enable-ds-eviction

0sewa0 (approved):

PR #4897, opened 2025-05-20
Add default mappedHostPaths to kspm sample (v1beta5)

waodim:
## Description

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-5199)

A default list for `mappedHostPaths` for `KSPM` was worked out. With this PR this list is added to the `v1beta5` `KSPM` sample. As I already needed to create a new folder for this sample, I also added all other sample to the `v1beta5` folder.

## How can this be tested?

read

Try to apply the `KSPM` sample and see if it works and if the field is correctly set.


assets/samples/dynakube/v1beta5/kubernetesSecurityPostureManagement.yaml
```diff
@@ -0,0 +1,206 @@
...
+    # namespaceSelector:
+    #   matchLabels:
+    #     app: my-app
+    #   matchExpressions:
+    #   - key: app
+    #     operator: In
+  #     values: [my-frontend, my-backend, my-database]
+
+  kspm:
+    # default path list
```
chrismuellner, line 64: This comment should be the same as the comment on the field: https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/api/latest/dynakube/kspm/spec.go#L25

The comment on the field is not clear though, I'd suggest `Define the host paths that are mounted to the container.`

waodim: Thanks for the input. Updated the comment: https://github.com/Dynatrace/dynatrace-operator/pull/4897/commits/8848b4241756db3b03569b616900c562857585aa

gkrenn (dismissed):
Looks good, field in DynaKube is set, implementation to be actually used still has to be done

assets/samples/dynakube/v1beta5/kubernetesSecurityPostureManagement.yaml
```diff
@@ -0,0 +1,208 @@
...
+    #     app: my-app
+    #   matchExpressions:
+    #   - key: app
+    #     operator: In
+  #     values: [my-frontend, my-backend, my-database]
+
+  kspm:
+    #  Define the host paths that are mounted to the container.
+    #
+    #  Default list of paths:
```
aorcholski, line 66: "Default" is a little bit confusing. 

According to my ticket "By default no directories are mounted". This file is an example (so also kind of an instruction) so it should be clearly stated that this is rather a suggestion (smallest set of directories to get valuable knowledge about the host) or just a random example...

IMHO Default == kspm:{} -> all these directories are mounted
but it's not true.

waodim: Thanks for the input. I agree, done: https://github.com/Dynatrace/dynatrace-operator/pull/4897/commits/4d37be28c3ff12b437b803a701fbe7f975176f79

aorcholski (approved):
LGTM

chrismuellner (approved):

PR #4927, opened 2025-06-04
Cleanup image build related make targets and related scripts.

0sewa0:
## Description
(related to) [DAQ-9325](https://dt-rnd.atlassian.net/browse/DAQ-9325)
- created these changes during this research

Did some updates to the local image build scripts (mainly scripts related to `hack/make/images.mk`)
- 1 exceptions is the update to `hack/build/ci/create-image-index.sh` (formerly `hack/build/ci/create-fips-image-index.sh`)

Now when pushing local dev images (ie.: `make build`) the scripts will always: 
- build an image `quay.io/dynatrace/dynatrace-operator:snapshot-<sanitized-branch-name>-<arch>`
- push the image
- create an image index `quay.io/dynatrace/dynatrace-operator:snapshot-<sanitized-branch-name>`

Also added a new make targets:
- `images/build/multi`: builds an arm64 and amd64 image
- `images/push/multi`: pushes the arm64 and amd64 image, creates an image-index containing them
- `images/push/build/multi`: calls the 2 previous targets 

### Reasoning for the change:
1. Our CI builds both arm64 and amd64 images
  - Publishes them as part of an image-index
2. Our local build only builds for 1 arch
  - Publishes them as a single image manifest, with the same name as the CI created image-index
     -  ❌ possible conflict

I wanted to test my changes on both a local ARM cluster, and on a cloud AMD cluster, without having to create PR and rely on the CI.

## How can this be tested?

All of these scripts should work using `podman`.
But the changes were used to test `Rancher Desktop` (on Mac) to use `docker` instead of `podman`.

Targets to test:
```bash
# defaults
make build
make images/build/push/multi
make images/build/push/fips

# customized
OPERATOR_BUILD_PLATFORM="linux/arm64" make build
OPERATOR_BUILD_PLATFORM="linux/arm64" make images/build/push/fips
```

[DAQ-9325]: https://dt-rnd.atlassian.net/browse/DAQ-9325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hack/build/build_image.sh
```diff
@@ -31,13 +31,7 @@ else
   CONTAINER_CMD=docker
 fi
 
-OPERATOR_BUILD_PLATFORM="--platform=linux/amd64"
-if [ -n "${OPERATOR_DEV_BUILD_PLATFORM}" ]; then
-  echo "overriding platform to ${OPERATOR_DEV_BUILD_PLATFORM}"
```
andriisoldatenko, line 36: just to be super pedantic, IMO better to print that we don't use it anymore and print WARNING if it's set
`echo "WARNING!!! please remove env variable `OPERATOR_DEV_BUILD_PLATFORM `, we don't use it anymore"`

0sewa0: the likelihood of someone reading this is minimal

PR #4949, opened 2025-06-08
WIP Add csidriver.job settings to values.yaml

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-6488)

## How can this be tested?

helm tests
unittests


0sewa0:
messed around with the `values.yaml` got the following config:
```yaml
    - name: TOLERATIONS
      value: '[{"effect":"NoSchedule","key":"node-role.kubernetes.io/master","operator":"Exists"},{"effect":"NoSchedule","key":"node-role.kubernetes.io/control-plane","operator":"Exists"}]'
    - name: ANNOTATIONS
      value: '{"test-annotation":"my-annotation"}'
    - name: LABELS
      value: '{"test-label":"my-label"}'
    - name: CSI_DATA_DIR
      value: /var/lib/kubelet/plugins/csi.oneagent.dynatrace.com/data
    - name: modules.json
      value: |
        {
          "csiDriver": true,
          "activeGate": true,
          "oneAgent": true,
          "extensions": true,
          "logMonitoring": true,
          "edgeConnect": true,
          "supportability": true,
          "kspm": true
        }
    - name: job.json
      value: |
        {
          "securityContext": {"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"privileged":false,"readOnlyRootFilesystem":true,"runAsNonRoot":false,"runAsUser":0,"seLinuxOptions":{"level":"s0"},"se
ccompProfile":{"type":"RuntimeDefault"}},
          "resources": {"requests":{"cpu":"100m","memory":"100Mi"}},
        }
```

that land on the job: (the resource seems to be not propagated)
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  annotations:
	...
    test-annotation: my-annotation
  labels:
	...
    test-label: my-label
  name: codemodule-download-2693506667
  namespace: dynatrace
spec:
	...
  template:
    metadata:
      annotations:
		...
        test-annotation: my-annotation
      labels:
		...
        test-label: my-label
    spec:
      containers:
		...
        name: codemodule-download
        resources:
          requests:
            cpu: 300m // incorrect
            memory: 100Mi
        securityContext:
          capabilities:
            drop:
            - ALL
          privileged: false
          readOnlyRootFilesystem: true
          runAsNonRoot: false
          runAsUser: 0
          seLinuxOptions:
            level: s0
          seccompProfile:
            type: RuntimeDefault
```

config/helm/chart/default/values.yaml
```diff
@@ -204,6 +204,24 @@ csidriver:
...
+        drop:
+          - ALL
+      seLinuxOptions:
+        level: s0
+      seccompProfile:
+        type: RuntimeDefault
+    resources:
+      requests:
+        cpu: 300m
+        memory: 100Mi
```
0sewa0: should be the same as the `provisioner`'s, I know the ticket has a comment that mentions a possible better config, however we should test and verify that that actually works, before making it default

the customer can configure it on their own if they wish, but otherwise, lets stay on the safe side

config/helm/chart/default/tests/Common/csi/daemonset_test.yaml
```diff
@@ -239,6 +239,10 @@ tests:
                         fieldPath: spec.nodeName
                   - name: TOLERATIONS
                     value: '[{"effect":"NoSchedule","key":"node-role.kubernetes.io/master","operator":"Exists"},{"effect":"NoSchedule","key":"node-role.kubernetes.io/control-plane","operator":"Exists"}]'
+                  - name: ANNOTATIONS
+                    value: '{}'
+                  - name: LABELS
+                    value: '{}'
```
0sewa0: related to this and the `job.json`:

wouldn't it make sense to have a more consistent feeling config?
you can either go with:
a) everything has it's own separate env, like `ANNOTATIONS` (so `JOB_SECURITY_CONTEXT` and so on)
b) or everything is in this kind of json, `jobs.json` would be too specific, could be something like `helm.json`, that would have something like:
```json
{
	"annotations": {},
   "labels": {},
   "tolerations": [{"effect":"NoSchedule","key":"node-role.kubernetes.io/master","operator":"Exists"},{"effect":"NoSchedule","key":"node-role.kubernetes.io/control-plane","operator":"Exists"}],
    "job": {
					"securityContext": {...},
          "resources": {...}
		}
}
```
c) something totally different :D 

all have their up/downsides, I don't have a strong preference either way, I just want something consistent


foreseen questions about the `modules.json`:
- having that separate is nice, due to it being reused in other components

aorcholski: my first commit is almost B)
so let's make a step back

aorcholski: https://github.com/Dynatrace/dynatrace-operator/pull/4949/commits/0d935409e58c1a29594ec9c4adbef9b0d3d75920

config/helm/chart/default/values.yaml
aorcholski: https://github.com/Dynatrace/dynatrace-operator/pull/4949/commits/4002c0150319b944090630f210542f95839e62c2

PR #4952, opened 2025-06-10
Do not set v2.initContainer.RunAsUser on OpenShift cluster

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-8351)

## Description

securityContext.RunAsUser of the `dynatrace-operator` initContainer is not set by webhook if it runs in OpenShift cluster.

e2e test scenarios:
- SampleApp
  No SecurityContext, no ClusterRole - it works because k8s automatically sets RunAsUser to a valid value.
```
	sampleApp := sample.NewApp(t, &dk,
		sample.AsDeployment(),
		sample.WithSecurityContext(corev1.PodSecurityContext{}),
		sample.WithoutClusterRole(),
	)
	builder.Assess("install sample app", sampleApp.Install())
```
This case fails without the fix
```
+       if !isOpenShift {
+               initSecurityCtx.RunAsUser = ptr.To(oacommon.DefaultUser)
+       }
```

- RandomUserSampleApp
  RunAsUser set to an invalid value, ClusterRole(Binding) defined - it works because ClusterRole provided
```
	randomUserSample := sample.NewApp(t, &dk,
		sample.WithName("random-user"),
		sample.AsDeployment(),
		sample.WithSecurityContext(corev1.PodSecurityContext{
			RunAsUser:  ptr.To[int64](1234),
			RunAsGroup: ptr.To[int64](1234),
		}),
	)
	builder.Assess("install sample app with random users set", randomUserSample.Install())
```	
- RandomUserSampleFail
  RunAsUser set to an invalid value, no ClusterRole - installation fails as expected
```
		randomUserSampleFail := sample.NewApp(t, &dk,
			sample.WithName("random-user-fail"),
			sample.AsDeployment(),
			sample.WithSecurityContext(corev1.PodSecurityContext{
				RunAsUser:  ptr.To[int64](1234),
				RunAsGroup: ptr.To[int64](1234),
			}),
			sample.WithoutClusterRole(),
		)
		builder.Assess("try to install sample app with random users set", randomUserSampleFail.InstallFail())
```

## How can this be tested?

``` make test/e2e/applicationmonitoring/bootstrapper-no-csi```

### DIFF
--- a/cmd/webhook/cmd.go
+++ b/cmd/webhook/cmd.go
@@ -37,6 +37,8 @@ const (
 	FlagCertificateDirectory   = "certs-dir"
 	FlagCertificateFileName    = "cert"
 	FlagCertificateKeyFileName = "cert-key"
+
+	openshiftSecurityGVR = "security.openshift.io/v1"
 )
 
 var (
@@ -65,8 +67,6 @@ func New() *cobra.Command {
 
 func run() func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
-		const openshiftSecurityGVR = "security.openshift.io/v1"
-
 		installconfig.ReadModules()
 		version.LogVersion()
 		logd.LogBaseLoggerSettings()

### CODE REVIEW

File: cmd/webhook/cmd.go [hunk 1]
```diff
@@ -75,6 +79,23 @@ func run() func(*cobra.Command, []string) error {
 			return err
 		}
 
+		isOpenShift := false
+
+		client, err := discovery.NewDiscoveryClientForConfig(kubeConfig)
+		if err != nil {
+			logd.Get().WithName("platform").Info("detected platform", "platform", "unknown")
```
0sewa0: the message should be `failed to detect platform, due to discovery client issues`, and you could even log the error 
so `log.Error(...)`

aorcholski: done https://github.com/Dynatrace/dynatrace-operator/pull/4952/commits/ad4cbaec98a5337510e2f83a8aa0876c2e946262

cmd/webhook/cmd.go
```diff
@@ -63,6 +65,8 @@ func New() *cobra.Command {
 
 func run() func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
+		const openshiftSecurityGVR = "security.openshift.io/v1"
```
andriisoldatenko: could you please move it outside of the function, to the top level where we usually declare constants?

andriisoldatenko (dismissed):

andriisoldatenko (approved):

PR #4955, opened 2025-06-11
Improve error verbosity in the Dynatrace client

Incredible-O:
## Description
If Dynatrace (or some proxy) returns non-JSON objects to the client (like `<!doctype html><html>hi</html>` for example), the following error message is currently returned:
```
response error, can't unmarshal json response 403: invalid character '<' looking for beginning of value
```

Now, this error message is more descriptive:
```
Server returned status code 403; can't unmarshal response (content-type: text/html): <!doctype html><html>hi</html>
```

Addresses [DAQ-2762](https://dt-rnd.atlassian.net/browse/DAQ-2762)

## How can this be tested?
* Run unit tests in `pkg/clients/dynatrace/dynatrace_client_test.go`

[DAQ-2762]: https://dt-rnd.atlassian.net/browse/DAQ-2762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/clients/dynatrace/dynatrace_client.go
+++ b/pkg/clients/dynatrace/dynatrace_client.go
@@ -9,6 +9,8 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -40,8 +42,8 @@ type tokenType int
 const (
 	dynatraceApiToken tokenType = iota
 	dynatracePaaSToken
-	installerUrlToken // in this case we don't care about the token
-	MaxResponseLen    = 1000
+	installerUrlToken     // in this case we don't care about the token
+	defaultMaxResponseLen = 1000
 )
 
 // makeRequest does an HTTP request by formatting the URL from the given arguments and returns the response.
@@ -179,11 +181,22 @@ func (dtc *dynatraceClient) handleErrorResponseFromAPI(response []byte, statusCo
 			sb.WriteString(fmt.Sprintf(" (via proxy %s)", proxy))
 		}
 
-		responseLen := int(math.Min(MaxResponseLen, float64(len(response))))
+		responseLen := int(math.Min(float64(GetMaxResponseLen()), float64(len(response))))
 		sb.WriteString(fmt.Sprintf("; can't unmarshal response (content-type: %s): %s", contentType, response[:responseLen]))
 
 		return errors.New(sb.String())
 	}
 
 	return se.ErrorMessage
+}
+
+func GetMaxResponseLen() int {
+	envVar, set := os.LookupEnv("DT_CLIENT_API_ERROR_LOG_LEN")
+	if set {
+		maxResponseLen, err := strconv.Atoi(envVar)
+		if err == nil {
+			return maxResponseLen
+		}
+	}
+	return defaultMaxResponseLen
 }
--- a/pkg/clients/dynatrace/dynatrace_client.go
+++ b/pkg/clients/dynatrace/dynatrace_client.go
@@ -180,7 +180,7 @@ func (dtc *dynatraceClient) handleErrorResponseFromAPI(response []byte, statusCo
 			sb.WriteString(fmt.Sprintf(" (via proxy %s)", proxy))
 		}
 
-		responseLen := min(GetMaxResponseLen(), len(response))
+		responseLen := min(getMaxResponseLen(), len(response))
 		sb.WriteString(fmt.Sprintf("; can't unmarshal response (content-type: %s): %s", contentType, response[:responseLen]))
 
 		return errors.New(sb.String())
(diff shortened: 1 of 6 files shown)


pkg/clients/dynatrace/dynatrace_client.go
```diff
@@ -146,10 +149,39 @@ func (dtc *dynatraceClient) makeRequestForBinary(ctx context.Context, url string
...
-		return errors.WithMessagef(err, "response error, can't unmarshal json response %d", statusCode)
+		var sb strings.Builder
+
+		sb.WriteString(fmt.Sprintf("Server returned status code %d", statusCode))
+
+		if proxy != "" {
+			sb.WriteString(fmt.Sprintf(" (via proxy %s)", proxy))
+		}
+
+		responseLen := int(math.Min(1000, float64(len(response))))
```
andriisoldatenko: please use named constant for 1000 in the to of the file.

pkg/clients/dynatrace/dynatrace_client_test.go
```diff
@@ -89,6 +90,39 @@ func TestGetResponseOrServerError(t *testing.T) {
...
+		code := http.StatusForbidden
+
+		err := dc.handleErrorResponseFromAPI(response, code, http.Header{})
+		require.Error(t, err)
+
+		maxLen := 1000
+
+		var shortenedResponse []byte
+
+		if len(response) >= maxLen {
```
andriisoldatenko: could you please remove logic from tests, you use static response, why you need this if here?

pkg/clients/dynatrace/dynatrace_client_test.go
```diff
@@ -89,6 +90,39 @@ func TestGetResponseOrServerError(t *testing.T) {
...
+			shortenedResponse = response[:maxLen]
+		} else {
+			shortenedResponse = response
+		}
+
+		assert.EqualError(t, err, fmt.Sprintf("Server returned status code %d; can't unmarshal response (content-type: unknown): %s", code, shortenedResponse))
```
andriisoldatenko: it's better to check string against err so we know what to expect instead of formating it on the fly.

pkg/clients/dynatrace/dynatrace_client.go
andriisoldatenko: MaxResponseLen = 1000

pkg/clients/dynatrace/dynatrace_client_test.go
```diff
@@ -89,6 +90,39 @@ func TestGetResponseOrServerError(t *testing.T) {
...
+	})
+
+	t.Run("HTML response with proxy header set", func(t *testing.T) {
```
andriisoldatenko, line 127: i would add json response but without header set test case.

waodim (commented):
Apart from Andriis points, I have nothing to add and will approve as soon as you adressed those.

pkg/clients/dynatrace/dynatrace_client_test.go
Incredible-O: Done in [d1043d1](https://github.com/Dynatrace/dynatrace-operator/pull/4955/commits/d1043d139156f5af06769057758719153f7ac7f0)

pkg/clients/dynatrace/dynatrace_client_test.go
Incredible-O: Done in 2be7dc88

pkg/clients/dynatrace/dynatrace_client_test.go
Incredible-O: Done in c99988c4

pkg/clients/dynatrace/dynatrace_client.go
Incredible-O: Done in 4de068ab

pkg/clients/dynatrace/dynatrace_client.go
```diff
@@ -179,11 +181,22 @@ func (dtc *dynatraceClient) handleErrorResponseFromAPI(response []byte, statusCo
...
+
+func GetMaxResponseLen() int {
+	envVar, set := os.LookupEnv("DT_CLIENT_API_ERROR_LOG_LEN")
+	if set {
+		maxResponseLen, err := strconv.Atoi(envVar)
+		if err == nil {
+			return maxResponseLen
+		}
+	}
+	return defaultMaxResponseLen
```
andriisoldatenko: i think we need to improve it a little bit,
```
func getMaxResponseLen() int {
	if envVar, exists := os.LookupEnv("DT_CLIENT_API_MAX_ERROR_LOG_LEN"); exists {
		maxResponseLen, err := strconv.Atoi(envVar)
		if err == nil && maxResponseLen > 0 {
			return maxResponseLen
		}
	}
	
	return defaultMaxResponseLen
}
```

also it can be lowercase

pkg/clients/dynatrace/dynatrace_client.go
```diff
@@ -146,11 +152,51 @@ func (dtc *dynatraceClient) makeRequestForBinary(ctx context.Context, url string
...
-		return errors.WithMessagef(err, "response error, can't unmarshal json response %d", statusCode)
+		var sb strings.Builder
+
+		sb.WriteString(fmt.Sprintf("Server returned status code %d", statusCode))
+
+		if proxy != "" {
+			sb.WriteString(fmt.Sprintf(" (via proxy %s)", proxy))
+		}
+
+		responseLen := int(math.Min(float64(GetMaxResponseLen()), float64(len(response))))
```
andriisoldatenko:
```suggestion
		responseLen := min(GetMaxResponseLen(), len(response))
```
go supports generics ;)

andriisoldatenko (changes requested):

pkg/clients/dynatrace/dynatrace_client.go
Incredible-O: Done in 36ad638a

pkg/clients/dynatrace/dynatrace_client.go
Incredible-O: Done in 22ab8000

andriisoldatenko (approved):
well done!

PR #4956, opened 2025-06-12
Reconcile enrichment rules when node-image-pull is used

0sewa0:
## Description

[DAQ-9393](https://dt-rnd.atlassian.net/browse/DAQ-9393)

If OA gets injected into a pod via `node-image-pull` (no CSI), then full metadata enrichment happen, no matter the `metadataEnrichment.enabled`.

Reasoning:
- The self-extracting CodeModule usecase (`node-image-pull` no-csi) uses the [bootstrappers](https://github.com/Dynatrace/dynatrace-bootstrapper) CLI, and the arguments for these were designed with this limitation. (you can't turn off metadataEnrichment)

## How can this be tested?

1. Deploy operator with no csi
2. Apply dynakube, example:
```
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
  annotations:
spec:
  apiUrl: https://<my-tenant>.dev.dynatracelabs.com/api
  tokens: dynakube
  metadataEnrichment:
    enabled: false
  oneAgent:
    cloudNativeFullStack:
```
3. Check in the dynakube's Status, that the Enrichment Rules are there

### DIFF
--- a/pkg/controllers/dynakube/metadata/rules/reconciler.go
+++ b/pkg/controllers/dynakube/metadata/rules/reconciler.go
@@ -29,7 +29,7 @@ func NewReconciler(dtc dtclient.Client, dk *dynakube.DynaKube) controllers.Recon
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context) error {
-	if !r.dk.MetadataEnrichmentEnabled() {
+	if !r.dk.MetadataEnrichmentEnabled() && !r.dk.FF().IsNodeImagePull() {
 		if meta.FindStatusCondition(*r.dk.Conditions(), conditionType) == nil {
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


waodim (approved):
Rules are considered, status is set

```
│     Last Transition Time:  2025-06-12T14:51:13Z                                                                                                                                                                  │
│     Message:               Metadata-enrichment rules are up-to-date in the status                                                                                                                                │
│     Reason:                StatusUpdated                                                                                                                                                                         │
│     Status:                True                                                                                                                                                                                  │
│     Type:                  MetadataEnrichmentRules    
```

andriisoldatenko:
Tested, but first need to add rule to my tenant https://docs.dynatrace.com/docs/ingest-from/setup-on-k8s/guides/metadata-automation/k8s-metadata-telemetry-enrichment

and click save!

then I was able to see rule inside status not only conditions.
```
k get dks dynakube -n dynatrace -o yaml  | yq '.status.metadataEnrichment'
rules:
  - source: label-team
    target: dt.security_context
    type: ANNOTATION
```

PR #4970, opened 2025-06-17
Fix fips release ci

waodim:
## Description

Here the fips release failed: https://github.com/Dynatrace/dynatrace-operator/actions/runs/15681786204

## How can this be tested?

- fips release will have to be done with the next rc (or is there another way to test?)


waodim:
> At this point, I would just remove the descriptions or set maxDescLen=0, because they make no sense this short. What do you guys think?

Actually yes. I was just hoping that someone maybe finds an error on my side, because the jump form 350 to 25 (or 0) just does not feel right to me. But if there is no other way, I would also suggest to just go for 0

waodim:
Attention: The changes regarding the CRD description length got removed from this PR again as we came to the conclusion that it is fine for now. But please have a look at the FIPS stuff!

0sewa0 (approved):

PR #4972, opened 2025-06-17
Run whole ci on main and release branches

waodim:
## Description

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-9822)

Currently we skip some steps in the CI due to the detect changes step. We should not skip steps when on main or an release branch. This PR fixes this problem.

## How can this be tested?

- fork this repo and push some things to trigger the several scenarios

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,20 +28,20 @@ jobs:
       helm_changed: ${{ steps.helm.outputs.any_changed }}
       prerequisites_changed: ${{ steps.prerequisites.outputs.any_changed }}
       ci_changed: ${{ steps.ci.outputs.any_changed }}
-      run_anyway: ${{ steps.set-run-anyway.outputs.run_anyway }}
+      is_protected_branch: ${{ steps.set-run-anyway.outputs.is_protected_branch }}
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 2
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
-      - name: Set run anyway outputs
-        id: set-run-anyway
+      - name: Check if it is a protected branch
+        id: is-protected-branch
         run: |
           if [[ "$GITHUB_REF" == "refs/heads/main" || "$GITHUB_REF" == refs/heads/release-* ]]; then
-            echo "run_anyway=true" >> "$GITHUB_OUTPUT"
+            echo "is_protected_branch=true" >> "$GITHUB_OUTPUT"
           else
-            echo "run_anyway=false" >> "$GITHUB_OUTPUT"
+            echo "is_protected_branch=false" >> "$GITHUB_OUTPUT"
           fi
       - name: Detect CI file changes
         uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c # v46.0.5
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -28,12 +28,21 @@ jobs:
       helm_changed: ${{ steps.helm.outputs.any_changed }}
       prerequisites_changed: ${{ steps.prerequisites.outputs.any_changed }}
       ci_changed: ${{ steps.ci.outputs.any_changed }}
+      run_anyway: ${{ steps.set-run-anyway.outputs.run_anyway }}
```
andriisoldatenko:
```suggestion
      run_anyway: ((contains("release-"), github.base_ref) || github.ref == 'refs/heads/main'
```
if I'm not mistake you can just do it like this, without bash

.github/workflows/ci.yaml
```diff
@@ -81,7 +90,7 @@ jobs:
 
   helm-test:
     needs: detect-changes
-    if: needs.detect-changes.outputs.helm_changed == 'true' || needs.detect-changes.outputs.prerequisites_changed == 'true' || needs.detect-changes.outputs.ci_changed == 'true'
+    if: needs.detect-changes.outputs.run_anyway == 'true' || needs.detect-changes.outputs.helm_changed == 'true' || needs.detect-changes.outputs.prerequisites_changed == 'true' || needs.detect-changes.outputs.ci_changed == 'true'
```
andriisoldatenko: also i'm not sure that you need 
```suggestion
    if: needs.detect-changes.outputs.run_anyway || needs.detect-changes.outputs.helm_changed == 'true' || needs.detect-changes.outputs.prerequisites_changed == 'true' || needs.detect-changes.outputs.ci_changed == 'true'
```

andriisoldatenko: based on docs https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution

.github/workflows/ci.yaml
waodim: I don't think this works tbh, i tried it: https://github.com/waodim/dynatrace-operator/actions/runs/15705005595

.github/workflows/ci.yaml
waodim: As I set explicitly true and false in the set run anyways step, should I really remove it? and should I remove every else? because we always check for true

andriisoldatenko: I don't see any reason to have it, but I'm not 100% sure that it will work as expected, that's why I prefer to try it out.

.github/workflows/ci.yaml
andriisoldatenko: oh cool, so it only works for `if`, so yeah the way you implemented and if it works let's go this way.

.github/workflows/ci.yaml
waodim: ok I agree, let me try!

andriisoldatenko: so try it out, if it doesn't work just ignore my comment

waodim: done: a2213cb952c60923f7f3e62a9975aa6bd33e14e4

tested it here it seems to work: https://github.com/waodim/dynatrace-operator/actions/runs/15705406957

.github/workflows/ci.yaml
0sewa0: it could have a more understandable name, like `is_proctected_branch`

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/4972/commits/ceee2978db722b34e5c01db6ee732597971e087b

double check if I missed something pls

.github/workflows/ci.yaml
```diff
@@ -28,12 +28,21 @@ jobs:
       helm_changed: ${{ steps.helm.outputs.any_changed }}
       prerequisites_changed: ${{ steps.prerequisites.outputs.any_changed }}
       ci_changed: ${{ steps.ci.outputs.any_changed }}
+      is_protected_branch: ${{ steps.set-run-anyway.outputs.is_protected_branch }}
```
Incredible-O:
```suggestion
      is_protected_branch: ${{ steps.is-protected-branch.outputs.is_protected_branch }}
```

Other than that, LGTM 👍🏿

PR #4985, opened 2025-06-17
Additional transform processor to add pod IP.

StefanHauth:
## Description

An additional processor is needed to fix pod association for Jaeger compact thrift telemetry ingest.

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-9389)

## How can this be tested?

1. Deploy operator
2. Deploy DynaKube with telemetryIngest, Jaeger protocol enabled
3. Deploy Jager sample app (e.g. HotROD v1.40.0), configure for using the compact-thrift endpoint
4. Make sure that traces are visible in the Distributed Tracings app and that workload and pod metadata is correctly added


pkg/otelcgen/processors.go
```diff
@@ -27,6 +27,7 @@ type MemoryLimiter struct {
 var (
 	k8sattributes     = component.MustNewID("k8sattributes")
 	transform         = component.MustNewID("transform")
+	transformPodIp    = component.MustNewIDWithName("transform", "add-pod-ip")
```
andriisoldatenko:
```suggestion
	transformPodIP    = component.MustNewIDWithName("transform", "add-pod-ip")
```

pkg/otelcgen/processors.go
```diff
@@ -121,6 +123,18 @@ func (c *Config) buildTransform() map[string]any {
 	}
 }
 
+func (c *Config) buildTransformPodIp() map[string]any {
```
andriisoldatenko:
```suggestion
func (c *Config) buildTransformPodIP() map[string]any {
```

andriisoldatenko: @StefanHauth could you please update tests?

pkg/otelcgen/processors.go
StefanHauth: fixed

pkg/otelcgen/processors.go
StefanHauth: fixed

aorcholski (approved):
I see traces from hotROD app and pod ip

andriisoldatenko (approved):

PR #4999, opened 2025-06-18
Address Staticcheck `ST1005` error

Incredible-O:
## Description
* Fixes: ```ST1005: error strings should not end with punctuation or newlines```
* See https://staticcheck.dev/docs/checks/#ST1005

Addresses [DAQ-6691](https://dt-rnd.atlassian.net/browse/DAQ-6691)

## How can this be tested?
* `cmd.go` unit tests
* run `k -n dynatrace exec  deployment/dynatrace-operator -- dynatrace-operator support-archive` and check the error message

[DAQ-6691]: https://dt-rnd.atlassian.net/browse/DAQ-6691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/supportarchive/cmd.go
```diff
@@ -67,10 +67,10 @@ func New() *cobra.Command {
 				return nil
 			}
 
-			return fmt.Errorf("%s %s %s\n",
+			return fmt.Errorf("%s %s %s",
 				"The only option to retrieve the support archive is by using '--stdout=true'.",
 				"Please provide this parameter and make sure that you pipe the command output to a file.",
-				"Otherwise, your terminal will be flooded with binary data.")
+				"Otherwise, your terminal will be flooded with binary data")
```
0sewa0: kinda unrelated, this supposed to be the output instruction for a cli tool, so I don't mind the dot really

BUT, 

why are we using `fmt.Errorf` where everything is static?
why not use `errors.New(msg)`?

even using `strings.Builder` would be less weird:
like:
```go
sb := strings.Builder{}
sb.WriteString("The only option to retrieve the support archive is by using '--stdout=true'. ")
sb.WriteString("Please provide this parameter and make sure that you pipe the command output to a file. ")
sb.WriteString("Otherwise, your terminal will be flooded with binary data.")

return errors.New(sb.String())
```

but why not just a multiline string?: (this would have `\n` all over, but how does that look in the terminal?)
```go
msg := `The only option to retrieve the support archive is by using '--stdout=true'.
Please provide this parameter and make sure that you pipe the command output to a file.
Otherwise, your terminal will be flooded with binary data.`

return errors.New(msg)
```

andriisoldatenko: I prefer option 2.

Incredible-O: Addressed in d28c66dc

cmd/supportarchive/cmd.go
```diff
@@ -67,10 +68,12 @@ func New() *cobra.Command {
 				return nil
 			}
 
-			return fmt.Errorf("%s %s %s\n",
-				"The only option to retrieve the support archive is by using '--stdout=true'.",
-				"Please provide this parameter and make sure that you pipe the command output to a file.",
-				"Otherwise, your terminal will be flooded with binary data.")
+			sb := strings.Builder{}
+			sb.WriteString("The only option to retrieve the support archive is by using '--stdout=true'. ")
```
andriisoldatenko, line 72: why do u need extra space hee?
```suggestion
			sb.WriteString("The only option to retrieve the support archive is by using '--stdout=true'.")
```

cmd/supportarchive/cmd.go
```diff
@@ -67,10 +68,12 @@ func New() *cobra.Command {
...
+			sb.WriteString("Please provide this parameter and make sure that you pipe the command output to a file. ")
```
andriisoldatenko, line 73:
```suggestion
			sb.WriteString("Please provide this parameter and make sure that you pipe the command output to a file.")
```

cmd/supportarchive/cmd.go
```diff
@@ -67,10 +68,12 @@ func New() *cobra.Command {
 				return nil
 			}
 
-			return fmt.Errorf("%s %s %s\n",
-				"The only option to retrieve the support archive is by using '--stdout=true'.",
-				"Please provide this parameter and make sure that you pipe the command output to a file.",
-				"Otherwise, your terminal will be flooded with binary data.")
+			sb := strings.Builder{}
```
andriisoldatenko, line 71: IMO strings builder is too much here, I like multilines ` ``raw text`` ` raw string patter more for such stuff, but i'm OK with this approach as well.

cmd/supportarchive/cmd.go
Incredible-O: Because in line 76 (where `sb.String()` is invoked), three separate sentences are connected. Removing the spaces would make the output look like this, making it harder to recognize when new sentences (or words) begin/end:
```Error: The only option to retrieve the support archive is by using '--stdout=true'.Please provide this parameter and make sure that you pipe the command output to a file.Otherwise, your terminal will be flooded with binary data.```

andriisoldatenko (approved):

PR #5001, opened 2025-06-18
Properly cleanup CSI host mounts

0sewa0:
## Description

[DAQ-9735](https://dt-rnd.atlassian.net/browse/DAQ-9735)

CSI host mounts were not thoroughly cleaned up if they were migrated.

OneAgent DaemonSet Pods will be stuck in `ContainerCreation` if the following happens:
1. OA pods with CSI volumes deployed, with an older (1.4.2) Operator version
2. Upgrade to 1.5.1 Operator version (CSI volume migration happens in the background, via symlink)
     - set `csidriver.cleanupPeriod` to a low value, so the cleanup happens faster, so this error happens also faster 
4. Switch modes from CloudNative/HostMonitoring to Appmon, wait for the OA pods to go away
     - DO NOT DELETE YOUR DYNAKUBE (or it won't happen)
5.  Switch mode back to CloudNative/HostMonitoring 
     - DO NOT DELETE YOUR DYNAKUBE (or it won't happen)
7. See the new OA pods be stuck in `ContainerCreation`

The problem happens because the `IsMountPoint` call follows the link, even if it leads to no where
and it the link leads to nowhere, then `IsMountPoint` errors, and we don’t delete because of safety concerns
- “if we can’t determine if it is a MountPoint or not, we should do nothing“, however, `IsMountPoint` should be smarter about this error handling

## How can this be tested?

Run the above mentioned scenario.

Log lines you want to see:
```
,"ts":"2025-06-18T15:06:05.476Z","logger":"csi-cleanup","msg":"removed old host mount directory","path":"/data/zib50933/osagent"} // old but actual folder
,"ts":"2025-06-18T15:06:05.751Z","logger":"csi-cleanup","msg":"removed old host mount directory","path":"/data/_dynakubes/mydk/osagent"} // incase of a migrated folder, this is just a symlink
```


pkg/controllers/csi/provisioner/cleanup/hostmounts.go
```diff
@@ -25,7 +26,15 @@ func (c *Cleaner) isMountPoint(file string) (bool, error) {
 		}
 	}
 
-	return c.mounter.IsMountPoint(file)
+	isMountPoint, err := c.mounter.IsMountPoint(file)
+	if os.IsNotExist(err) {
```
andriisoldatenko, line 30: nice! potentially can we unit test it? 🤔

andriisoldatenko, line 30:
```suggestion
	if err != nil && os.IsNotExist(err) {
```

andriisoldatenko, line 30: never mind you can ignore it `os.IsNotExist(nil)` returns false

andriisoldatenko (approved):

PR #5017, opened 2025-06-23
Use consistent method receiver names

Incredible-O:
## Description
Addresses Staticcheck `ST1016` errors:
* See https://staticcheck.dev/docs/checks/#ST1016
* Example: `ST1016: methods on the same type should have the same receiver name (seen 3x "e", 6x "ec", 7x "edgeConnect") (staticcheck)`

Relevant ticket: [DAQ-6691](https://dt-rnd.atlassian.net/browse/DAQ-6691)

## How can this be tested?
* Unit tests

### DIFF
--- a/pkg/api/v1beta4/dynakube/convert_from.go
+++ b/pkg/api/v1beta4/dynakube/convert_from.go
@@ -13,48 +13,48 @@ import (
 )
 
 // ConvertFrom converts from the Hub version (latest) to this version (v1beta4).
-func (dst *DynaKube) ConvertFrom(srcRaw conversion.Hub) error {
+func (dk *DynaKube) ConvertFrom(srcRaw conversion.Hub) error {
 	src := srcRaw.(*dynakubelatest.DynaKube)
 
-	dst.fromStatus(src)
+	dk.fromStatus(src)
 
-	dst.fromBase(src)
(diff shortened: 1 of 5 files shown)


pkg/api/v1beta4/dynakube/convert_from.go
```diff
@@ -13,91 +13,91 @@ import (
 )
 
 // ConvertFrom converts from the Hub version (latest) to this version (v1beta4).
-func (dst *DynaKube) ConvertFrom(srcRaw conversion.Hub) error {
+func (dk *DynaKube) ConvertFrom(srcRaw conversion.Hub) error {
```
waodim: I am not sure about this one. Should we `nolint` this? I mean yes of course is violates this check, but on the other hand I am not sure how confusing it can become when trying to distinguish between the `dynakube` versions.

cc @andriisoldatenko

andriisoldatenko: IMO: in this case i would keep dst, so nolint.

waodim: Yes, tbh I agree, would also go with nolinting it

Incredible-O: Addressed in 471d4ae6

PR #5022, opened 2025-06-24
Revert "Improve how we wait for status when we create/destroy clusters"

0sewa0:
## Description
[DAQ-8528](https://dt-rnd.atlassian.net/browse/DAQ-8528)

Just a revert of https://github.com/Dynatrace/dynatrace-operator/pull/3805 with some extra fixes.
- We don't just wait till we reach `environment-deployed`, we wait until transition, then check if we are in the correct state

## How can this be tested?

Run e2e pipeline on this branch

### DIFF
--- a/.github/scripts/create-cluster.sh
+++ b/.github/scripts/create-cluster.sh
@@ -2,7 +2,7 @@
 
 set -x
 
-DEFAULT_TIMEOUT="60m" # K8s takes <10min, Openshift >40min
+DEFAULT_TIMEOUT="80m" # K8s takes <10min, Openshift >40min
 DESIRED_STATE="environment-deployed"
 
 kubectl version
@@ -12,7 +12,7 @@ echo "Creating environment '$FLC_ENVIRONMENT' in namespace '$FLC_NAMESPACE'"
 kubectl get flcenvironments --namespace "$FLC_NAMESPACE"
 
 echo "Patching environment '$FLC_ENVIRONMENT' to 'deployed'"
-kubectl patch --namespace "$FLC_NAMESPACE" --type merge --patch '{"spec": {"desiredState": "environment-deployed"}}' flcenvironment "$FLC_ENVIRONMENT"
+kubectl patch --namespace "$FLC_NAMESPACE" --type merge --patch "{\"spec\": {\"desiredState\": \"$DESIRED_STATE\"}}" flcenvironment "$FLC_ENVIRONMENT"
 
 echo "Waiting up to '$DEFAULT_TIMEOUT' for successful deployment of environment '$FLC_ENVIRONMENT'"
 kubectl wait --namespace "$FLC_NAMESPACE" --timeout="$DEFAULT_TIMEOUT" --for=condition=InTransition=false flcenvironment "$FLC_ENVIRONMENT"
--- a/.github/scripts/create-cluster.sh
+++ b/.github/scripts/create-cluster.sh
@@ -2,7 +2,7 @@
 
 set -x
 
-DEFAULT_TIMEOUT="80m" # K8s takes <10min, Openshift >40min
+DEFAULT_TIMEOUT="60m" # K8s takes <10min, Openshift >40min
 DESIRED_STATE="environment-deployed"
 
 kubectl version
@@ -15,4 +15,20 @@ echo "Patching environment '$FLC_ENVIRONMENT' to 'deployed'"
 kubectl patch --namespace "$FLC_NAMESPACE" --type merge --patch '{"spec": {"desiredState": "environment-deployed"}}' flcenvironment "$FLC_ENVIRONMENT"
 
 echo "Waiting up to '$DEFAULT_TIMEOUT' for successful deployment of environment '$FLC_ENVIRONMENT'"
-kubectl wait --namespace "$FLC_NAMESPACE" --timeout="$DEFAULT_TIMEOUT" --for jsonpath='{.status.currentState}'="$DESIRED_STATE" flcenvironment "$FLC_ENVIRONMENT"
+kubectl wait --namespace "$FLC_NAMESPACE" --timeout="$DEFAULT_TIMEOUT" --for=condition=InTransition=false flcenvironment "$FLC_ENVIRONMENT"
+
+if [[ "$FLC_ENVIRONMENT" == *"ocp"* ]]; then
+  echo "PLATFORM is set to 'ocp'. Waiting for 5 minutes..."
+  sleep 300  # Wait for 5 minutes (300 seconds)
+  echo "done"
+fi
+
+echo "Checking currentState='$DESIRED_STATE' for '$FLC_ENVIRONMENT'..."
+flc_state=$(kubectl get flcenvironment "$FLC_ENVIRONMENT" --namespace "$FLC_NAMESPACE" -ojsonpath="{.status.currentState}")
+if [[ "$flc_state" != "$DESIRED_STATE" ]]; then
+  echo "Pipeline deployment did not reach expected state '$DESIRED_STATE', currentState: ${flc_state}..."
+  exit 1
+  else
+    echo "successful..."
+fi
+echo "done"
--- a/.github/scripts/destroy-cluster.sh
+++ b/.github/scripts/destroy-cluster.sh
@@ -3,10 +3,30 @@
 set -x
 
 DEFAULT_TIMEOUT="20m"
+DESIRED_STATE="environment-not-deployed"
 
 echo "Destroying environment '$FLC_ENVIRONMENT' in namespace '$FLC_NAMESPACE'"
 
 kubectl get flcenvironments --namespace "$FLC_NAMESPACE"
 
 echo "Patching environment '$FLC_ENVIRONMENT' to 'not-deployed'"
-kubectl patch --namespace "$FLC_NAMESPACE" --type merge --patch '{"spec": {"desiredState": "environment-not-deployed"}}' flcenvironment "$FLC_ENVIRONMENT"
+kubectl patch --namespace "$FLC_NAMESPACE" --type merge --patch "{\"spec\": {\"desiredState\": \"$DESIRED_STATE\"}}" flcenvironment "$FLC_ENVIRONMENT"
+
+echo "Waiting up to '$DEFAULT_TIMEOUT' for successful destruction of environment '$FLC_ENVIRONMENT'"
+kubectl wait --namespace "$FLC_NAMESPACE" --timeout="$DEFAULT_TIMEOUT" --for=condition=InTransition=false flcenvironment "$FLC_ENVIRONMENT"
+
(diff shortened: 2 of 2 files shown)


.github/scripts/create-cluster.sh
```diff
@@ -12,7 +12,23 @@ echo "Creating environment '$FLC_ENVIRONMENT' in namespace '$FLC_NAMESPACE'"
 kubectl get flcenvironments --namespace "$FLC_NAMESPACE"
 
 echo "Patching environment '$FLC_ENVIRONMENT' to 'deployed'"
-kubectl patch --namespace "$FLC_NAMESPACE" --type merge --patch '{"spec": {"desiredState": "environment-deployed"}}' flcenvironment "$FLC_ENVIRONMENT"
+kubectl patch --namespace "$FLC_NAMESPACE" --type merge --patch "{\"spec\": {\"desiredState\": \"$DESIRED_STATE\"}}" flcenvironment "$FLC_ENVIRONMENT"
```
andriisoldatenko:
```suggestion
kubectl patch --namespace "$FLC_NAMESPACE" --type merge --patch "{\"spec\": {\"desiredState\": \"$DESIRED_STATE\"}}" flcenvironment "$FLC_ENVIRONMENT"
```
this change doesn't bring any value

.github/scripts/create-cluster.sh
```diff
@@ -12,7 +12,23 @@ echo "Creating environment '$FLC_ENVIRONMENT' in namespace '$FLC_NAMESPACE'"
...
 
 echo "Waiting up to '$DEFAULT_TIMEOUT' for successful deployment of environment '$FLC_ENVIRONMENT'"
-kubectl wait --namespace "$FLC_NAMESPACE" --timeout="$DEFAULT_TIMEOUT" --for jsonpath='{.status.currentState}'="$DESIRED_STATE" flcenvironment "$FLC_ENVIRONMENT"
+kubectl wait --namespace "$FLC_NAMESPACE" --timeout="$DEFAULT_TIMEOUT" --for=condition=InTransition=false flcenvironment "$FLC_ENVIRONMENT"
+
+if [[ "$FLC_ENVIRONMENT" == *"ocp"* ]]; then
+  echo "PLATFORM is set to 'ocp'. Waiting for 5 minutes..."
+  sleep 300  # Wait for 5 minutes (300 seconds)
```
andriisoldatenko: prev approach had lot's of issues, for example 
we wait for 80m on line 18, but instead, we want to send a request (kubectl get and wait in bash loop) but more frequently.

Also AFAIK `--for=condition=InTransition=false ` worked not well, that's why i switched to different approach. 
So IMO we should use as written in ticket every 1-2 mins or so get s.tatus.currentState and decide what we want to do. and if it's failed -> fail fast.

.github/scripts/create-cluster.sh
```diff
@@ -12,7 +12,23 @@ echo "Creating environment '$FLC_ENVIRONMENT' in namespace '$FLC_NAMESPACE'"
 kubectl get flcenvironments --namespace "$FLC_NAMESPACE"
```
waodim: As much as I love the idea to fix this issue, did we come to a conclusion on how to fix this? Is there a proven way on how to improve it? Just asking out of curiosity.

andriisoldatenko: check my most recent commit.

.github/scripts/create-cluster.sh
```diff
@@ -15,4 +15,29 @@ echo "Patching environment '$FLC_ENVIRONMENT' to 'deployed'"
...
 echo "Waiting up to '$DEFAULT_TIMEOUT' for successful deployment of environment '$FLC_ENVIRONMENT'"
-kubectl wait --namespace "$FLC_NAMESPACE" --timeout="$DEFAULT_TIMEOUT" --for jsonpath='{.status.currentState}'="$DESIRED_STATE" flcenvironment "$FLC_ENVIRONMENT"
+
+# wait until timeout or until the environment is in the desired state
+NEXT_WAIT_TIME=0
+while [[ $NEXT_WAIT_TIME -ne 80 ]]; do
+  # Check if the environment is in the desired state
+  current_state=$(kubectl --namespace "$FLC_NAMESPACE" --for jsonpath='{.status.currentState}' flcenvironment "$FLC_ENVIRONMENT")
+
+  if [[ "$current_state" == "$DESIRED_STATE" ]]; then
```
StefanHauth: is the additional var her required? The other if branches here use hard-coded strings. I'd suggest to make that consistent.

.github/scripts/create-cluster.sh
```diff
@@ -15,4 +15,29 @@ echo "Patching environment '$FLC_ENVIRONMENT' to 'deployed'"
...
+    echo "Environment '$FLC_ENVIRONMENT' is in desired state '$DESIRED_STATE'."
```
StefanHauth:
```suggestion
    echo "Environment '$FLC_ENVIRONMENT' has been deployed successfully."
```

.github/scripts/destroy-cluster.sh
```diff
@@ -11,4 +12,24 @@ kubectl get flcenvironments --namespace "$FLC_NAMESPACE"
...
 kubectl patch --namespace "$FLC_NAMESPACE" --type merge --patch '{"spec": {"desiredState": "environment-not-deployed"}}' flcenvironment "$FLC_ENVIRONMENT"
 
-kubectl wait --namespace "$FLC_NAMESPACE" --timeout="$DEFAULT_TIMEOUT" --for jsonpath='{.status.currentState}'=environment-not-deployed flcenvironment "$FLC_ENVIRONMENT"
+# wait until timeout or until the environment is in the desired state
+NEXT_WAIT_TIME=0
+while [[ $NEXT_WAIT_TIME -ne 20 ]]; do
+  # Check if the environment is in the desired state
+  current_state=$(kubectl --namespace "$FLC_NAMESPACE" --for jsonpath='{.status.currentState}' flcenvironment "$FLC_ENVIRONMENT")
+
+  if [[ "$current_state" == "$DESIRED_STATE" ]]; then
```
StefanHauth: Same here, use either hard coded strings or var consistently.

.github/scripts/destroy-cluster.sh
```diff
@@ -11,4 +12,24 @@ kubectl get flcenvironments --namespace "$FLC_NAMESPACE"
...
+    echo "Environment '$FLC_ENVIRONMENT' is in desired state '$DESIRED_STATE'."
```
StefanHauth:
```suggestion
    echo "Environment '$FLC_ENVIRONMENT' has been destroyed successfully."
```

StefanHauth (changes requested):

.github/scripts/destroy-cluster.sh
andriisoldatenko: i'll hardcode, since we don't reuse code.

.github/scripts/create-cluster.sh
```diff
@@ -15,4 +15,29 @@ echo "Patching environment '$FLC_ENVIRONMENT' to 'deployed'"
...
+    break
+  elif [[ "$current_state" == "environment-deploying" ]]; then
+    echo "Environment '$FLC_ENVIRONMENT' is currently deploying. Waiting for it to complete..."
+    let NEXT_WAIT_TIME += 1
+    sleep 60
+  elif [[ "$current_state" == "environment-deploy-failed" ]]; then
```
waodim: I am not sure if this state is valid: https://bitbucket.lab.dynatrace.org/projects/CPN/repos/flc-operator/browse/internal/state/states.go#11

Or am I missing something?

andriisoldatenko: nah you are absolutely right , it was made copilot 💩 fixing.

.github/scripts/destroy-cluster.sh
andriisoldatenko: fixed in b0a40921f

.github/scripts/destroy-cluster.sh
andriisoldatenko: fixed in b0a40921f

.github/scripts/create-cluster.sh
andriisoldatenko: fixed in b0a40921f

.github/scripts/create-cluster.sh
andriisoldatenko: fixed in b0a40921f

.github/scripts/create-cluster.sh
andriisoldatenko: fixed in 0cb05c67e

StefanHauth (approved):
LGTM

.github/scripts/create-cluster.sh
```diff
@@ -15,4 +15,25 @@ echo "Patching environment '$FLC_ENVIRONMENT' to 'deployed'"
...
+    exit 1
+  elif [[ -z "$current_state" ]]; then
+    echo "Environment '$FLC_ENVIRONMENT' does not exist or is not ready. Exiting."
+    exit 1
+  else
+    echo "Current state of environment '$FLC_ENVIRONMENT': '$current_state'. Waiting for desired state 'environment-deployed'..."
+    NEXT_WAIT_TIME=$((NEXT_WAIT_TIME+1))
+    sleep 60
+  fi
+done
```
aorcholski, line 39:
```
exit 1
``` 
should be added at the end of the script (line 40). Cluster is not ready after 80 iteration means an error (like in line 30).

.github/scripts/destroy-cluster.sh
```diff
@@ -11,4 +11,24 @@ kubectl get flcenvironments --namespace "$FLC_NAMESPACE"
...
+    exit 1
+  elif [[ -z "$current_state" ]]; then
+    echo "Environment '$FLC_ENVIRONMENT' does not exist or is not ready. Exiting."
+    exit 1
+  else
+    echo "Current state of environment '$FLC_ENVIRONMENT': '$current_state'. Waiting for desired state 'environment-not-deployed'..."
+    NEXT_WAIT_TIME=$((NEXT_WAIT_TIME+1))
+    sleep 60
+  fi
+done
```
aorcholski, line 34:
```
exit 1
``` 
should be added at the end of the script (line 35). Cluster is not deleted after 20 iteration means an error (like in line 25).

aorcholski (changes requested):

.github/scripts/create-cluster.sh
andriisoldatenko: or good catch!

andriisoldatenko: fixed in 26c27ed9c

.github/scripts/destroy-cluster.sh
andriisoldatenko: fixed in 26c27ed9c

PR #5049, opened 2025-06-30
Add deprecation warning v1beta1 DK version

waodim:
## Description

We decided to mark the oldest existing version of our DK CRD as deprecated, because this will make our lives easier when it comes to future releases.

## How can this be tested?

- read
- have a look at the CRD
- try to apply manifests

### DIFF
--- a/config/crd/bases/dynatrace.com_dynakubes.yaml
+++ b/config/crd/bases/dynatrace.com_dynakubes.yaml
@@ -32,165 +32,92 @@ spec:
     name: v1beta1
     schema:
       openAPIV3Schema:
-        description: DynaKube is the Schema for the DynaKube API
         properties:
           apiVersion:
-            description: |-
-              APIVersion defines the versioned schema of this representation of an object.
-              Servers should convert recognized schemas to the latest internal value, and
-              may reject unrecognized values.
-              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
             type: string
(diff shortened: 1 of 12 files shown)


config/crd/bases/dynatrace.com_dynakubes.yaml
```diff
@@ -29,6 +29,9 @@ spec:
     - jsonPath: .metadata.creationTimestamp
       name: Age
       type: date
+    deprecated: true
+    deprecationWarning: The dynakube API is deprecated and will be removed in a future
```
andriisoldatenko: can we ask someone to review message? because I think the plan is to remove in 1.7 so we need to state it explicitly.

waodim: added reference to 1.7 here: https://github.com/Dynatrace/dynatrace-operator/pull/5049/commits/2ea1e82b3a7cbe0b40a8f497a23a2d88028e9805

Would still be nice to get a review. I will reach out.

wepudt: Perhaps word it a bit differently: This version of the DynaKube API is deprecated ...?

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5049/commits/5224c4b6f7549985176f5f9a37a9aeaaee3893c7

PR #5072, opened 2025-07-04
feature(charts): add liveness probe config of csi daemonset in values

oihanitrz:
## Description

This PR introduces the ability to customize the `livenessProbe` for the `server` container of the `dynatrace-oneagent-csi-driver` DaemonSet via the Helm chart.

### Motivation

In certain scenarios, the CSI driver pods enter a `CrashLoopBackOff` state. After investigation, and based on this article from Dynatrace Community (https://community.dynatrace.com/t5/Troubleshooting/Kubernetes-Operator-CSI-driver-crashes-frequently/ta-p/272021), it appears that tuning the `livenessProbe` is one of the recommended mitigations.

While adjusting resource limits is currently supported through the Helm chart, the other two workarounds require manual patches, which are not ideal for industrialized or automated deployments. This change makes `livenessProbe` configuration manageable via values.yaml, enhancing flexibility and reducing operational overhead.

## How can this be tested?

- Install or upgrade the Helm chart with a custom `values.yaml` including a disabled `livenessProbe` under `csidriver.server.livenessProbe`.
- Confirm that the generated DaemonSet for `dynatrace-oneagent-csi-driver` reflects the expected probe configuration.

Example `values.yaml`:
```yaml
csidriver:
  server:
    livenessProbe:
      enabled: false


oihanitrz:
I also added support for customizing the --probe-timeout argument of the liveness-probe container, which I had initially overlooked.

PR #5076, opened 2025-07-07
Use DT error codes (WIP)

StefanHauth:
## Description

Sample of how error codes can be used in code.

## How can this be tested?

Check logs and see if error codes are there


albertogdd:
Can we close this PR?

PR #5103, opened 2025-07-14
Unify webhook mutation

0sewa0:
## Description

mainly [DAQ-7415](https://dt-rnd.atlassian.net/browse/DAQ-7415)

also these: (for linting and just sanity checking)
- [DAQ-9674](https://dt-rnd.atlassian.net/browse/DAQ-9674)
- [DAQ-7416](https://dt-rnd.atlassian.net/browse/DAQ-7416)

> [!TIP]
> Very quick summery of the change:
> - csi never configures, only provides binary
>   - related volume is always readonly (remove secrets related to it)
> - only 1 type of init secret
>   - the `dynatrace-bootstrapper-config` + `dynatrace-bootstrapper-certs` 
> - use new `bootstrap` instead of old `startup` package for init-container
> - only 1 version of injection
>   - no more v1 or v2
>   - the volume-mounts are always not more than 3   

<details><summary>Details</summary>
<p>

TODO

</p>
</details> 

> [!CAUTION]
> I still have to update the e2e tests, so running them is not useful 

## How can this be tested?

TODO


pkg/webhook/mutation/pod/common/interface.go
```diff
@@ -1,22 +1,16 @@
...
 
-import "context"
-
-type PodInjector interface {
-	Handle(context.Context, *MutationRequest) error
-}
-
-type PodMutator interface {
+type Mutator interface {
 	// Enabled returns true if the mutator needs to be executed for the given request.
```
andriisoldatenko: doc string should be updated, wrong format,  `IsEnabled returns...`

pkg/webhook/mutation/pod/oneagent/init.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/consts"
+	maputils "github.com/Dynatrace/dynatrace-operator/pkg/util/map"
+	dtwebhook "github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/common"
+	"github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/common/arg"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func mutateInitContainer(mutationRequest *dtwebhook.MutationRequest, installPath string) error {
+	isCSI := isCSIVolume(mutationRequest.BaseRequest)
```
andriisoldatenko: don't see any reason to use a new variable here. I would prefer to use the function directly inside if

pkg/webhook/mutation/pod/oneagent/init.go
```diff
@@ -0,0 +1,112 @@
...
+		addCSIBinVolume(
+			mutationRequest.Pod,
+			mutationRequest.DynaKube.Name,
+			mutationRequest.DynaKube.FF().GetCSIMaxRetryTimeout().String())
+		// incase of CSI, the CSI volume itself is already always readonly, so the mount should always be readonly, the init-container should just read from it
+		addInitBinMount(mutationRequest.InstallContainer, true)
+	} else {
+		log.Info("configuring init-container with emptyDir bin volume", "name", mutationRequest.PodName())
+		addEmptyDirBinVolume(mutationRequest.Pod)
+		// incase of no CSI, the the emptyDir can't be readonly for the init-container as it first has to download/move the agent into it
```
andriisoldatenko:
```suggestion
		// in case of no CSI, the emptyDir can't be readonly for the init-container, as it first has to download/move the agent into it
```

pkg/webhook/mutation/pod/oneagent/init.go
```diff
@@ -0,0 +1,112 @@
...
+		addInitBinMount(mutationRequest.InstallContainer, false)
+
+		if mutationRequest.DynaKube.FF().IsNodeImagePull() {
+			log.Info("configuring init-container with self-extracting image", "name", mutationRequest.PodName())
+			// The first element would be the "bootstrap" sub command, which is not needed incase of self-extracting image
```
andriisoldatenko:
```suggestion
			// The first element would be the "bootstrap" subcommand, which is not needed in case of self-extracting image
```

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -107,17 +107,24 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 	return nil
 }
 
-func (r *Reconciler) unmapDynakube(ctx context.Context) {
-	if meta.FindStatusCondition(*r.dk.Conditions(), codeModulesInjectionConditionType) != nil &&
-		meta.FindStatusCondition(*r.dk.Conditions(), metaDataEnrichmentConditionType) != nil {
+func (r *Reconciler) cleanup(ctx context.Context) {
+	if meta.FindStatusCondition(*r.dk.Conditions(), codeModulesInjectionConditionType) == nil &&
```
waodim, line 111: The inverted nil check makes total sense here. Out of curiosity: why was there a `not equal` check before? Am I missing something?

0sewa0: probably a bug, I see no logic in the previous one 😅

pkg/webhook/mutation/pod/oneagent/init.go
```diff
@@ -0,0 +1,112 @@
...
+	if technology, ok := pod.Annotations[AnnotationTechnologies]; ok {
+		return technology
+	}
+
+	technology := dk.FF().GetNodeImagePullTechnology()
+	if technology != "" {
+		return technology
+	}
+
+	return ""
```
aorcholski:
```suggestion
	return dk.FF().GetNodeImagePullTechnology()
```

pkg/webhook/mutation/pod/oneagent/init.go
0sewa0: leftover from a previous impl, fixing

pkg/webhook/mutation/pod/common/volumes/volumes.go
baichinger, line 1: It seems `InitConfigSubPath` can be removed. The `subPath` mount doesn't seem to be necessary.

0sewa0: true, used to be "necessary" when everything (bin + config) was put into the `dynatrace-config` emptyDir (during node-image-pull)

pkg/webhook/mutation/pod/oneagent/init.go
0sewa0: done

0sewa0: done

pkg/webhook/mutation/pod/oneagent/init.go
0sewa0: done

pkg/webhook/mutation/pod/oneagent/init.go
0sewa0: done

PR #5116, opened 2025-07-16
Custom version field should not be affected by autoUpdate

0sewa0:
## Description

[DAQ-11120](https://dt-rnd.atlassian.net/browse/DAQ-11120)

Fixes the following scenario:
1. User creates a `DynaKube` with `autoUpdate: false` + sets the custom version to `version: 1.2.3`
  - Only affects host `OneAgent`
2. Operator deploys/uses version 1.2.3
3. (🕐  clock  time passes 🕐 )
4. User updates the custom version to `version: 1.3.0`
5. **we do NOT deploy/use version 1.3.0 (we should)**

## How can this be tested?

Follow instructions in description ☝️ (just use actual versions 😅 )

### DIFF
--- a/pkg/controllers/dynakube/version/updater.go
+++ b/pkg/controllers/dynakube/version/updater.go
@@ -47,7 +47,7 @@ func (r *reconciler) run(ctx context.Context, updater StatusUpdater) error {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


andriisoldatenko (approved):

andriisoldatenko:
tested in my cluster!

PR #5119, opened 2025-07-16
WIP Makes settings.read/write, entities.read scopes optional

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-10269)


pkg/controllers/dynakube/dynatraceclient/builder.go
```diff
@@ -22,15 +28,17 @@ type Builder interface {
...
-	dk        dynakube.DynaKube
+	ctx                    context.Context
+	apiReader              client.Reader
+	tokens                 token.Tokens
+	newDynatraceClientFunc dtclient.NewFunc
+	dk                     dynakube.DynaKube
 }
 
-func NewBuilder(apiReader client.Reader) Builder {
+func NewBuilder(apiReader client.Reader, newDynatraceClientFunc dtclient.NewFunc) Builder {
```
0sewa0: do not do this, 
having to tell the "builder" which "builder" (constructor) it should use seems just wrong
- if you want to mock it in the `conroller_test.go` there is the [interface](https://github.com/Dynatrace/dynatrace-operator/pull/5119/files#diff-c660df24e3afd404994914d00c13b3ade035b5f7fd590a9c528876865d0b653fR22) create a mock for that

semi offtopic: (you don't have to fix this)
this whole `builder` here feels a bit wrong, could possibly simplified, like:
- `SetContext`: it is somewhat of an antipattern to make context be part of struct
- Have this "builder" be something that will give you a `dtclient` for a given `dynakube`
  - you could already "translate" the scopes needed according to the `dynakube` and provide it to the client
- Move the token verification to be part of the client

pkg/controllers/dynakube/dynatraceclient/builder.go
```diff
@@ -149,3 +169,23 @@ func lastErrorFromCondition(dkStatus *dynakube.DynaKubeStatus) error {
 
 	return nil
 }
+
+func setConditionOptionalScopeAvailable(dkStatus *dynakube.DynaKubeStatus, conditionType string, scope string) {
+	tokenCondition := metav1.Condition{
+		Type:    conditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  ReasonOptionalScope,
```
0sewa0: shouldn't there be another reason when scope is available? (I like the `Message` parts, only the `Reason` is a bit off)

If we read it aloud: "ApiTokenSettingsRead is available(ie. `true`) BECAUSE token is optional" is a bit off
Reason in the `true` state should be `ScopeAvailable` (or something) so it would sound like: 
"ApiTokenSettingsRead is available(ie. `true`) BECAUSE token has the scope" (or something)

the `OptionalScope` is good `Reason` for the `false` scenario:  "ApiTokenSettingsRead is not available(ie. `false`) BECAUSE the scope is optional"

aorcholski: [f507e8e](https://github.com/Dynatrace/dynatrace-operator/pull/5119/commits/f507e8e6b656fc3640443e56814a1134dec4b0b4)
```
Message:       settings.read is available
Reason:        OptionalScope
Status:        True
Type:          ApiTokenSettingsRead 
```
```
Message:       settings.read is not available, some features may not work
Reason:        OptionalScope
Status:        False
Type:          ApiTokenSettingsRead 
```

pkg/controllers/dynakube/controller.go
```diff
@@ -428,3 +441,82 @@ func (controller *Controller) createDynakubeMapper(ctx context.Context, dk *dyna
 
 	return &dkMapper
 }
+
+func (controller *Controller) tokenVerification(ctx context.Context, dynatraceClient dtclient.Client, dk *dynakube.DynaKube) error {
```
waodim:
```suggestion
func (controller *Controller) verifyToken(ctx context.Context, dynatraceClient dtclient.Client, dk *dynakube.DynaKube) error {
```

pkg/controllers/dynakube/dynatraceclient/builder.go
aorcholski: [f507e8e](https://github.com/Dynatrace/dynatrace-operator/pull/5119/commits/f507e8e6b656fc3640443e56814a1134dec4b0b4)
- context removed from the _builder_ struct
- token verification moved to the controller  
  I think dtclient is responsible for sending and receiving data, it depends on the caller how to use received data

pkg/controllers/dynakube/dynatraceclient/builder.go
0sewa0: shouldn't it be like?
```
Message:       settings.read is available
Reason:        ScopePresent
Status:        True
Type:          ApiTokenSettingsRead 
```

pkg/controllers/dynakube/dynatraceclient/builder_test.go
```diff
@@ -39,13 +40,13 @@ func TestBuildDynatraceClient(t *testing.T) {
 			},
 			dk: *dk,
 		}
-		dtc, err := dynatraceClientBuilder.Build()
+		dtc, err := dynatraceClientBuilder.Build(context.Background())
```
0sewa0: small nitpick:

you could just define the ctx in the beginning

```
func TestBuildDynatraceClient(t *testing.T) {
	ctx := context.Background()
...
}

File: pkg/controllers/dynakube/dynatraceclient/builder.go (hunk 2 above)
Comment (aorcholski) [reply pkg/controllers/dynakube/dynatraceclient/builder.go]: good idea
[a60377d](https://github.com/Dynatrace/dynatrace-operator/pull/5119/commits/a60377dcfaf5892113bc31ea2d1a1028b04b5a38)

File: pkg/controllers/dynakube/monitoredentities/reconciler.go [hunk 5]
```diff
@@ -40,6 +40,10 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 
 	conditions.SetStatusOutdated(r.dk.Conditions(), MEIDConditionType, "Kubernetes Cluster MEID is outdated in the status")
 
+	if !conditions.IsOptionalScopeAvailable(r.dk, dynatrace.ConditionTypeAPITokenEntitiesRead) {
+		log.Info(dynatrace.TokenScopeEntitiesRead + " optional scope not available, skipping reconciliation")
```
0sewa0: does it make sense to continue here?

why not just return?

I get that the `dtClient` should be able to handle it "soonTM" but IMO in this case, the whole reconciler is "turned off" if the scope is not provided

aorcholski: it's just a temporary workaround. 
deadcode linter rejects `IsOptionalScopeAvailable` definition if it is not used

The logic will be changed in scope of the following tickets.

pkg/controllers/dynakube/dynatraceclient/builder_test.go
aorcholski: right
[9c177cc](https://github.com/Dynatrace/dynatrace-operator/pull/5119/commits/9c177cc910b8e6fbe7405ae3271214c05440944b)

pkg/controllers/dynakube/monitoredentities/reconciler.go
0sewa0: yeah but I mean: 😄 

```go
if !conditions.IsOptionalScopeAvailable(r.dk, dynatrace.ConditionTypeAPITokenEntitiesRead) {
	log.Info(dynatrace.TokenScopeEntitiesRead + " optional scope not available, skipping reconciliation")
	return nil
}
```

aorcholski: you are right
`return nil` is in scope of the following ticket.
I added a message because it can't be an empty block. Empty blocks are rejected by a linter.
I don't want to change too much so I fixed the message. It is less confusing now (I hope).
[d4429a5](https://github.com/Dynatrace/dynatrace-operator/pull/5119/commits/d4429a540ce6870e0cfcb931e11ad73c8832eb5e)

0sewa0: you can add `// TODOS`, so it is clearer

like `// TODO: move this check inside the dtClient`, because I am worried it will be forgotten (if someone else does the following ticket)

aorcholski: done
[6c571cc](https://github.com/Dynatrace/dynatrace-operator/pull/5119/commits/6c571cc8e6df60e6eaf9734d74093004cb82f16a)

PR #5128, opened 2025-07-18
Generate sbom with correct arch inside the Dockerfile

albertogdd:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-8101

Generate sbom with correct build arch in the Dockerfile.

Renovate __should__ detect the cyclonemx-sbom tag and upgrade it properly

## How can this be tested?

make images/build
CI for FIPS


Dockerfile
```diff
@@ -14,12 +14,16 @@ RUN go mod download -x
...
 COPY pkg ./pkg
 COPY cmd ./cmd
+COPY .git /.git
 
 ARG GO_LINKER_ARGS
 ARG GO_BUILD_TAGS
 ARG TARGETARCH
 ARG TARGETOS
 
+RUN go install github.com/CycloneDX/cyclonedx-gomod/cmd/cyclonedx-gomod@v1.9.0
```
andriisoldatenko:
```suggestion
# renovate depName=github.com/CycloneDX/cyclonedx-gomod/cmd/cyclonedx-gomod
RUN go install github.com/CycloneDX/cyclonedx-gomod/cmd/cyclonedx-gomod@v1.9.0
```

and add this file to renovate list simialr to:
```
 {
      fileMatch: [
        "(^|/|\\.)prerequisites.mk$",
        "(^|/|\\.)third-party-licenses.sh$",
      ],
      matchStrings: [
        "depName=(?<depName>.*?)\\s.*?_version=(?<currentValue>.*)\\s",
      ],
      datasourceTemplate: "go",
    },
```
because Dockerfile matcher (IMO will only catch docker tags inside file )

Dockerfile
```diff
@@ -14,12 +14,16 @@ RUN go mod download -x
 
 COPY pkg ./pkg
 COPY cmd ./cmd
+COPY .git /.git
```
andriisoldatenko, line 17: why do we need .git folder inisdie docker layer?

albertogdd: This is because cyclonedx-gomod needs to have context of the existing tags.
You can see in [their Dockerfile examples](https://github.com/CycloneDX/cyclonedx-gomod/blob/main/Dockerfile.examples) that they even clone the repo
We already did this in the past ([see here](https://github.com/Dynatrace/dynatrace-operator/commit/086252211f901f6df4a4a2b2483ad6370802e0aa))

.github/workflows/release.yaml
```diff
@@ -47,10 +47,6 @@ jobs:
         uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
         with:
           go-version-file: "${{ github.workspace }}/go.mod"
-      - name: Prepare SBOM # Needs setup-go, uses a binary installed via `go install`
```
0sewa0, line 50: probably can remove the `Setup Golang` step, as the comment says

the only other step that will use golang (but probably not even that) is the `Build image` step, but that also has a `Setup Golang` inside it

Dockerfile
```diff
@@ -57,8 +62,6 @@ COPY --from=registry.k8s.io/sig-storage/livenessprobe:v2.16.0@sha256:88092d10090
 COPY ./third_party_licenses /usr/share/dynatrace-operator/third_party_licenses
 COPY LICENSE /licenses/
 
-COPY ./dynatrace-operator-bin-sbom.cdx.json ./dynatrace-operator-bin-sbom.cdx.json
```
0sewa0, line 60: you have to get the sbom from the build-container, otherwise it will be lost

this command would get it from the local FS, so you need something like `COPY --from=operator-build ...`

hack/make/release/release.mk
```diff
@@ -1,7 +1,3 @@
-## Generates SBOM of binary
-release/gen-sbom: prerequisites/cyclonedx-gomod
-	cyclonedx-gomod app -licenses -assert-licenses -json -main cmd/ -output dynatrace-operator-bin-sbom.cdx.json
-
 ## Generates SBOM of binary with CGO
 release/gen-sbom/fips: prerequisites/cyclonedx-gomod
 	CGO_ENABLED=1 cyclonedx-gomod app -licenses -assert-licenses -json -main cmd/ -output dynatrace-operator-bin-sbom.cdx.json
```
0sewa0: why not fix it for the fips image the same way?

.github/workflows/release.yaml
albertogdd: Fixed!

Dockerfile
albertogdd: Totally missed this, thanks for the hint :+1: 
Fixed!

hack/make/release/release.mk
albertogdd: Done! :sunglasses:

0sewa0 (approved):

PR #5151, opened 2025-07-25
Remove `entities.read` scope usage and replace it with `settings.read`

0sewa0:
## Description

[DAQ-10641](https://dt-rnd.atlassian.net/browse/DAQ-10641)

Replace the `GetMonitoredEntitiesForKubeSystemUUID` that used the `entities.read` scope with `GetK8sClusterME` that uses `settings.read` scope to get the same thing.
- It lists the `builtin:cloud.kubernetes` settings, that have a `scope` which is the ID of the entity we are looking for
- Only 1 such setting exists per tenant per Kubernetes cluster, and the `label` of the setting is the Name of the entity 

`GetK8sClusterME` is bit hard to parse at first, I added all the docs strings to help with that.
- Writing `GetKubernetesClusterMonitoredEntity` was IMO way worse
- The term "ME" is a common Dynatrace term we already use it in the status, see `kubernetesClusterMEID`

`GetK8sClusterME` (as the rename suggest) assumes a few things:
- The only type of entity we care about is a `KUBERNETES_CLUSTER_...` entity (so calling it `MonitoredEntity` would be too generic)
- There can only be 1 `KUBERNETES_CLUSTER_...` entity (per tenant) for a Kubernetes cluster, this is why it doesn't return a list.
- We only care about the ID of it and Label/Name, so that is the only things we store.

The `GetK8sClusterME` function is only used in the `monitoredentities.Reconciler`, that now will check for the `settings.read` scope.
Also the `entities.read` scope has been fully removed from the code.

## How can this be tested?
The unittests are pretty extensive... (mainly because some of them feel like integration tests...)

Manual test:
1. Create an `apiToken`, normal in every way BUT without the `entities.read` scope.
2. Use this `apiToken` in your DynaKube, and start deploying things that want to know the Kubernetes cluster entityID
 - (standalone logmon for example) 

[DAQ-10641]: https://dt-rnd.atlassian.net/browse/DAQ-10641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/clients/dynatrace/client.go
+++ b/pkg/clients/dynatrace/client.go
@@ -68,17 +68,17 @@ type Client interface {
 	// CreateOrUpdateKubernetesAppSetting returns the object id of the created k8s app settings if successful, or an api error otherwise
 	CreateOrUpdateKubernetesAppSetting(ctx context.Context, scope string) (string, error)
 
-	// GetKubernetesClusterEntity returns the KUBERNETES_CLUSTER entity for the give kubernetes cluster,
+	// GetK8sClusterME returns the Kubernetes Cluster Monitored Entity for the give kubernetes cluster.
 	// Uses the `settings.read` scope to list the `builtin:cloud.kubernetes` settings.
 	// - Only 1 such setting exists per tenant per kubernetes cluster
-	// - The `scope` for the setting is the ID of the KUBERNETES_CLUSTER entity
-	// - The `label` of the setting is the Name of the KUBERNETES_CLUSTER entity
-	// In case 0 settings are found, so no KUBERNETES_CLUSTER entity exist, we return an empty object, without an error.
-	GetKubernetesClusterEntity(ctx context.Context, kubeSystemUUID string) (KubernetesClusterEntity, error)
+	// - The `scope` for the setting is the ID (example: KUBERNETES_CLUSTER-A1234567BCD8EFGH) of the Kubernetes Cluster Monitored Entity
+	// - The `label` of the setting is the Name (example: my-dynakube) of the Kubernetes Cluster Monitored Entity
+	// In case 0 settings are found, so no Kubernetes Cluster Monitored Entity exist, we return an empty object, without an error.
+	GetK8sClusterME(ctx context.Context, kubeSystemUUID string) (K8sClusterME, error)
 
 	// GetSettingsForMonitoredEntity returns the settings response with the number of settings objects,
 	// or an api error otherwise
-	GetSettingsForMonitoredEntity(ctx context.Context, monitoredEntity KubernetesClusterEntity, schemaID string) (GetSettingsResponse, error)
+	GetSettingsForMonitoredEntity(ctx context.Context, monitoredEntity K8sClusterME, schemaID string) (GetSettingsResponse, error)
 
 	// GetSettingsForLogModule returns the settings response with the number of settings objects,
 	// or an api error otherwise
--- a/pkg/clients/dynatrace/client.go
+++ b/pkg/clients/dynatrace/client.go
@@ -68,6 +68,12 @@ type Client interface {
 	// CreateOrUpdateKubernetesAppSetting returns the object id of the created k8s app settings if successful, or an api error otherwise
 	CreateOrUpdateKubernetesAppSetting(ctx context.Context, scope string) (string, error)
 
+	// GetKubernetesClusterEntity returns the KUBERNETES_CLUSTER entity for the give kubernetes cluster,
+	// Uses the `settings.read` scope to list the `builtin:cloud.kubernetes` settings.
+	// - Only 1 such setting exists per tenant per kubernetes cluster
+	// - The `scope` for the setting is the ID of the KUBERNETES_CLUSTER entity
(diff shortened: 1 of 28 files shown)


pkg/clients/dynatrace/client.go
```diff
@@ -68,13 +68,11 @@ type Client interface {
 	// CreateOrUpdateKubernetesAppSetting returns the object id of the created k8s app settings if successful, or an api error otherwise
 	CreateOrUpdateKubernetesAppSetting(ctx context.Context, scope string) (string, error)
 
-	// GetMonitoredEntitiesForKubeSystemUUID returns a (possibly empty) list of k8s monitored entities for the given uuid,
-	// or an api error otherwise
-	GetMonitoredEntitiesForKubeSystemUUID(ctx context.Context, kubeSystemUUID string) ([]MonitoredEntity, error)
+	GetKubernetesClusterEntity(ctx context.Context, kubeSystemUUID string) (KubernetesClusterEntity, error)
```
andriisoldatenko: pls update doc string, instead of delete it.

0sewa0: done

pkg/clients/dynatrace/settings_test.go
```diff
@@ -33,20 +85,19 @@ func TestDynatraceClient_GetMonitoredEntitiesForKubeSystemUUID(t *testing.T) {
 		require.NotNil(t, dtc)
 
 		// act
-		actual, err := dtc.GetMonitoredEntitiesForKubeSystemUUID(ctx, testUID)
+		actual, err := dtc.GetKubernetesClusterEntity(ctx, testUID)
```
waodim: Should we also rename the ME reconciler as well?

0sewa0: maybe, `k8sclusterentity.Reconciler` is a bit of a mouthful 😅

pkg/util/conditions/token.go
```diff
@@ -6,7 +6,8 @@ import (
 )
 
 const (
-	DataIngestTokenMissing string = "DataIngestTokenMissing"
+	DataIngestTokenMissing = "DataIngestTokenMissing"
+	ScopeMissingReason     = "ScopeMissing"
```
andriisoldatenko, line 10:
```suggestion
	ScopeIsMissingReason     = "ScopeIsMissing"
```

pkg/clients/dynatrace/settings_test.go
waodim: don't have a strong opinion there. I agree that this is not really a nice sounding name.

- `clusterentity.Reconciler`
- `kubecluster.Reconciler`
- `entity.Reconciler`

Some ideas but to be honest I do not think there is a good one. Probably the best is to stick with ME reconciler

pkg/util/conditions/token.go
andriisoldatenko: also, i don't see unit-test which checks that we set status ScopeMissing, only this test found
```
	t.Run("no error + no run if no scope in status", func(t *testing.T) {
		clt := dtclientmock.NewClient(t)
		dk := createDynaKube()
		dk.Status.Conditions = []metav1.Condition{}

		reconciler := NewReconciler(clt, &dk)

		err := reconciler.Reconcile(ctx)

		require.NoError(t, err)
		require.Empty(t, dk.Status.KubernetesClusterMEID)

		condition := meta.FindStatusCondition(*dk.Conditions(), MEIDConditionType)
		require.NotNil(t, condition)
		assert.Equal(t, conditions.ScopeMissingReason, condition.Reason)
		assert.Equal(t, metav1.ConditionFalse, condition.Status)
		assert.Contains(t, condition.Message, dtclient.TokenScopeSettingsRead)
	})
```

andriisoldatenko:
tested in my cluster with scope/without scopes with logMon dunakube and it works. Only issue found that we don't set status is missing for missing scope.

pkg/util/conditions/token.go
andriisoldatenko: nvm

andriisoldatenko (approved):

pkg/clients/dynatrace/settings_test.go
0sewa0: renamed it to `k8sentity` added a doc for the package, also did bunch of renames everywhere related to it

cc @andriisoldatenko

waodim (approved):
Unit tests ran through and I also saw the condition (this was pre your renaming, so please fix unit tests at first, there are some failing ones now)

```
│     Last Transition Time:  2025-07-25T11:12:24Z                                                                                                                                                                                                                            │
│     Message:               settings.read is not available, some features may not work                                                                                                                                                                                      │
│     Reason:                OptionalScope                                                                                                                                                                                                                                   │
│     Status:                False                                                                                                                                                                                                                                           │
│     Type:                  ApiTokenSettingsRead  
```

andriisoldatenko (approved):

PR #5152, opened 2025-07-25
Make cluster creation failures more visible in e2e tests

gkrenn:
## Description

This PR increases the visibility of failures due to cluster creation. 
It does this by: 
- Creating a step after the e2e tests run that fails if the cluster creation failed => we see a red step in the job if the cluster creation failed
- Updating the slack notifications accordingly

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-10023)

## How can this be tested?

- Create a new branch based on this one
- Revert commit 7c7dcd32f06ae7845c0708718797f29e161173b2
- Trigger the e2e testing workflows manually in GitHub


.github/actions/run-e2e/action.yaml
```diff
@@ -72,11 +76,21 @@ runs:
...
       run: ref/.github/scripts/create-cluster.sh
       env:
         FLC_NAMESPACE: ${{ inputs.flc-namespace }}
         FLC_ENVIRONMENT: ${{ inputs.flc-environment }}
+    - name: Check cluster status
+      id: check_cluster_status
+      shell: bash
+      run: |
+        echo "Cluster status: ${{ steps.create_cluster.outputs.cluster_status }}"
+        if [[ "${{ steps.create_cluster.outputs.cluster_status }}" == "skipped" ]]; then
```
waodim, line 90: Why are we not checking for failure as well? Or is this done implicitly and I am missing it?

gkrenn: If there would be an error in create-cluster.sh the`cluster_status` is set to skipped, no extra need to check for error.

.github/scripts/create-cluster.sh
```diff
@@ -36,4 +42,6 @@ while [[ $NEXT_WAIT_TIME -ne 80 ]]; do
 done
 
 echo "Timeout reached while waiting for environment '$FLC_ENVIRONMENT' to be deployed."
+cluster_status="skipped"
```
0sewa0, line 45: add some big log, the "screams" what happened, so we can easily identify in the failed actions, like
```
|-------------------------|
|                         |
| CLUSTER CREATION FAILED |
|                         |
|-------------------------|
```

gkrenn: Created a separate step ("Cluster created successfully"), that tells you exactly that: https://github.com/Dynatrace/dynatrace-operator/actions/runs/16523974227/job/46732419218

gkrenn: Would assume that's enough

0sewa0: it is not enough; it is actually better than enough 🥳

0sewa0 (approved):

PR #5170, opened 2025-07-30
Migrate to wsl_v5 and fix linting issues

albertogdd:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-10415

Migrate from wsl to wsl_v5.

**I also removed duplicated linters (because they are already inside golangci-lint): gci, gofmt, govet and wsl.**

## How can this be tested?
- CI
- Run make go/lint


hack/make/prerequisites.mk
```diff
@@ -3,7 +3,7 @@ kustomize_version=v5.7.1
 #renovate depName=sigs.k8s.io/controller-tools/cmd
 controller_gen_version=v0.18.0
 # renovate depName=github.com/golangci/golangci-lint/v2
-golang_ci_cmd_version=v2.2.2
+golang_ci_cmd_version=v2.3.0
```
0sewa0: Shouldn't we also update the `wsl` dependency, in the `prerequisites/go-linting` target?

aorcholski (approved):
LGTM. no errors
```
...
go vet -copylocks=false ./...
wsl -fix -allow-trailing-comment ./pkg/...
betteralign -apply ./...
golangci-lint run --build-tags "e2e," --timeout 300s
0 issues.
# we add `tee` in the end to make it fail if it finds dead code, by default deadcode always return exit code 0
deadcode -test -tags="e2e," ./... | tee deadcode.out && [ ! -s deadcode.out ]
```

hack/make/prerequisites.mk
albertogdd: Removed

aorcholski (approved):
LGTM, no errors,  errors reported by `go fmt` are still reported
```
go install github.com/dkorunic/betteralign/cmd/betteralign@latest
betteralign -apply ./...
golangci-lint run --build-tags "e2e," --timeout 300s
0 issues.
# we add `tee` in the end to make it fail if it finds dead code, by default deadcode always return exit code 0
deadcode -test -tags="e2e," ./... | tee deadcode.out && [ ! -s deadcode.out ]
```

.github/workflows/ci.yaml
```diff
@@ -161,7 +161,7 @@ jobs:
         uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
         with:
           # renovate depName=github.com/golangci/golangci-lint
-          version: v2.2.2
+          version: v2.3.0
```
andriisoldatenko: we have separate ticket to bump this problem. https://github.com/Dynatrace/dynatrace-operator/pull/5132#pullrequestreview-3066924737 I prefer to have a seperate PR.

hack/make/go.mk
```diff
@@ -1,39 +1,13 @@
...
-## Runs linters that format/change the code in place
-go/format: go/fmt go/gci
-
-## Runs go vet
-go/vet:
-	go vet -copylocks=false $(LINT_TARGET)
-
-## Runs go wsl linter
-go/wsl:
-	wsl -fix -allow-trailing-comment ./pkg/...
```
andriisoldatenko, line 26: question: these linters/lines are here because they not only validate but also fix problems, so that's why we have a flag `--write` by removing them we need to do it manually or we need to introduce: `golangci-lint run --fix`
where  **--fix** - Fix found issues (if it's supported by the linter)`

andriisoldatenko (changes requested):

.github/workflows/ci.yaml
andriisoldatenko: Is it blocking us from making migration to **wsl_v5** without upgrading **golangci-lint?**

albertogdd: Thanks I didn't know about that ticket :+1: 
[Reverted](https://github.com/Dynatrace/dynatrace-operator/pull/5170/commits/14c31c675f83034630ff5ef4b5d3f51bb9b7e8b2)

hack/make/prerequisites.mk
albertogdd: [Version bump reverted](https://github.com/Dynatrace/dynatrace-operator/pull/5170/commits/14c31c675f83034630ff5ef4b5d3f51bb9b7e8b2)

hack/make/go.mk
albertogdd: [Added --fix to make go/golangci](https://github.com/Dynatrace/dynatrace-operator/pull/5170/commits/14c31c675f83034630ff5ef4b5d3f51bb9b7e8b2)

andriisoldatenko (approved):

PR #5197, opened 2025-08-05
chore: update code for newer go version

avorima:
## Description

Clean up code to use newer language features and stdlib functions.

### DIFF
--- a/pkg/api/exp/flag.go
+++ b/pkg/api/exp/flag.go
@@ -57,15 +57,7 @@ func (ff *FeatureFlags) getDisableFlagWithDeprecatedAnnotation(annotation string
 }
 
 func (ff *FeatureFlags) getRaw(annotation string) string {
-	if ff.annotations == nil {
-		return ""
-	}
-
-	if raw, ok := ff.annotations[annotation]; ok {
-		return raw
-	}
-
-	return ""
+	return ff.annotations[annotation]
 }
 
 func (ff *FeatureFlags) getBoolWithDefault(annotation string, defaultVal bool) bool {
--- a/pkg/controllers/csi/metadata/correctness.go
+++ b/pkg/controllers/csi/metadata/correctness.go
@@ -174,13 +174,13 @@ func GetRelevantOverlayMounts(mounter mount.Interface, baseFolder string) ([]Ove
 			}
 
 			for _, opt := range mountPoint.Opts {
-				switch key, value, _ := strings.Cut(opt, "="); key {
+				switch dirType, dirPath, _ := strings.Cut(opt, "="); dirType {
 				case "lowerdir":
-					overlayMount.LowerDir = value
+					overlayMount.LowerDir = dirPath
 				case "upperdir":
-					overlayMount.UpperDir = value
+					overlayMount.UpperDir = dirPath
 				case "workdir":
-					overlayMount.WorkDir = value
+					overlayMount.WorkDir = dirPath
 				}
 			}
 
--- a/pkg/controllers/csi/metadata/correctness.go
+++ b/pkg/controllers/csi/metadata/correctness.go
@@ -174,16 +174,13 @@ func GetRelevantOverlayMounts(mounter mount.Interface, baseFolder string) ([]Ove
 			}
 
 			for _, opt := range mountPoint.Opts {
-				switch {
-				case strings.HasPrefix(opt, "lowerdir="):
-					split := strings.SplitN(opt, "=", 2)
-					overlayMount.LowerDir = split[1]
-				case strings.HasPrefix(opt, "upperdir="):
-					split := strings.SplitN(opt, "=", 2)
-					overlayMount.UpperDir = split[1]
-				case strings.HasPrefix(opt, "workdir="):
-					split := strings.SplitN(opt, "=", 2)
-					overlayMount.WorkDir = split[1]
+				switch key, value, _ := strings.Cut(opt, "="); key {
+				case "lowerdir":
+					overlayMount.LowerDir = value
+				case "upperdir":
+					overlayMount.UpperDir = value
+				case "workdir":
+					overlayMount.WorkDir = value
 				}
 			}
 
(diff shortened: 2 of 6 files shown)


pkg/api/exp/flag.go
```diff
@@ -57,15 +57,7 @@ func (ff *FeatureFlags) getDisableFlagWithDeprecatedAnnotation(annotation string
...
-	if ff.annotations == nil {
-		return ""
-	}
-
-	if raw, ok := ff.annotations[annotation]; ok {
-		return raw
-	}
-
-	return ""
+	return ff.annotations[annotation]
```
andriisoldatenko, line 60: I'm afraid that prev implementation is not 100% eqvivalent to this change, for example I tried to add unit tests:
```
func TestNewFlags(t *testing.T) {
	ff := NewFlags(nil)
	assert.Nil(t, ff.annotations)

	assert.Nil(t, ff.annotations["test"])

	assert.Equal(t, ff.getRaw("test"), "")
}
```

and it fails, because we don't check or initialise properly map:

```
=== RUN   TestNewFlags
    flag_test.go:14: 
        	Error Trace:	/Users/andrii.soldatenko/work/dynatrace-operator/pkg/api/exp/flag_test.go:14
        	Error:      	Expected nil, but got: ""
        	Test:       	TestNewFlags
--- FAIL: TestNewFlags (0.00s)

FAIL

Process finished with the exit code 1

```

pkg/controllers/csi/metadata/correctness.go
```diff
@@ -174,16 +174,13 @@ func GetRelevantOverlayMounts(mounter mount.Interface, baseFolder string) ([]Ove
...
-				case strings.HasPrefix(opt, "lowerdir="):
-					split := strings.SplitN(opt, "=", 2)
-					overlayMount.LowerDir = split[1]
-				case strings.HasPrefix(opt, "upperdir="):
-					split := strings.SplitN(opt, "=", 2)
-					overlayMount.UpperDir = split[1]
-				case strings.HasPrefix(opt, "workdir="):
-					split := strings.SplitN(opt, "=", 2)
-					overlayMount.WorkDir = split[1]
+				switch key, value, _ := strings.Cut(opt, "="); key {
```
andriisoldatenko: I prefer to name it: `before, after, _ := strings.Cut(opt, "=")`

pkg/webhook/mutation/pod/attributes.go
```diff
@@ -84,30 +84,23 @@ func createImageInfo(imageURI string) containerattr.ImageInfo { // TODO: move to
...
 
-	repoPart := ""
-
-	registrySplit := strings.SplitN(imageURI, "/", 2)
-	if len(registrySplit) == 1 {
-		repoPart = registrySplit[0]
-	} else if len(registrySplit) == 2 {
-		imageInfo.Registry = registrySplit[0]
-		repoPart = registrySplit[1]
+	registry, repo, ok := strings.Cut(imageURI, "/")
```
andriisoldatenko:
```suggestion
	registry, repo, found := strings.Cut(imageURI, "/")
```

pkg/webhook/mutation/pod/attributes.go
```diff
@@ -84,30 +84,23 @@ func createImageInfo(imageURI string) containerattr.ImageInfo { // TODO: move to
...
+	if ok {
```
andriisoldatenko:
```suggestion
	if found {
```

pkg/webhook/mutation/pod/attributes.go
```diff
@@ -84,30 +84,23 @@ func createImageInfo(imageURI string) containerattr.ImageInfo { // TODO: move to
...
+		repo = registry
 	}
 
-	digestSplit := strings.SplitN(repoPart, "@", 2)
-	if len(digestSplit) == 1 {
-		repoPart = digestSplit[0]
-	} else if len(digestSplit) == 2 {
-		imageInfo.ImageDigest = digestSplit[1]
-		repoPart = digestSplit[0]
+	repo, digest, ok := strings.Cut(repo, "@")
```
andriisoldatenko:
```suggestion
	repo, digest, found := strings.Cut(repo, "@")
```

pkg/webhook/mutation/pod/attributes.go
```diff
@@ -84,30 +84,23 @@ func createImageInfo(imageURI string) containerattr.ImageInfo { // TODO: move to
...
 
-	tagSplit := strings.SplitN(repoPart, ":", 2)
-	if len(tagSplit) == 1 {
-		imageInfo.Repository = tagSplit[0]
-	} else if len(tagSplit) == 2 {
-		imageInfo.Tag = tagSplit[1]
-		imageInfo.Repository = tagSplit[0]
+	var tag string
+
+	imageInfo.Repository, tag, ok = strings.Cut(repo, ":")
```
andriisoldatenko:
```suggestion
	imageInfo.Repository, tag, found = strings.Cut(repo, ":")
```

pkg/webhook/mutation/pod/attributes_test.go
```diff
@@ -27,9 +27,9 @@ func TestAddPodAttributes(t *testing.T) {
 		rawArgs := []string{}
 
 		for _, arg := range request.InstallContainer.Args {
-			splitArg := strings.SplitN(arg, "=", 2)
-			require.Len(t, splitArg, 2)
-			rawArgs = append(rawArgs, splitArg[1])
+			_, rawArg, ok := strings.Cut(arg, "=")
```
andriisoldatenko:
```suggestion
			_, rawArg, found := strings.Cut(arg, "=")
```

pkg/api/exp/flag.go
avorima: reading from a nil map returns the zero value, so in case of string it's the empty string. the assertion is wrong in that test case

pkg/controllers/csi/metadata/correctness.go
0sewa0: you could call it:
`dirType` and `path` or `dirPath`

`before` and `after` would tell you the "position" of the strings in the cut string, which I don't care that much about

in other places we also name the "cut parts" what they represent (`registry, repo, found := strings.Cut(imageURI, "/")`)

pkg/api/exp/flag.go
andriisoldatenko: true:
>A nil map is equivalent to an empty map except that no elements may be added.

found it, it will panic ONLY on write .. uhhh.

pkg/controllers/csi/metadata/correctness.go
andriisoldatenko: i'm OK with what @0sewa0 proposed. I liked idea to name parts.

PR #5206, opened 2025-08-07
Pipe internal `http.Server` logs from the webhook into the debug logs

0sewa0:
## Description
[DAQ-11838](https://dt-rnd.atlassian.net/browse/DAQ-11838)

We don't want to spam the normal logs with logs from the internal `http.Server` started by the webhook.
- We can't set the logger for it directly, through an Option or such via the framework
  - So we have to "go around it"  

I also added a way to "easily" enable the debug logs for the Operator components in the `value.yaml`.
- Off by default, so users are not getting spammed
- On by default for `make deploy` so we don't have to turn it on during development
   - (Can easily change this behavior) 

## How can this be tested?

Use `make deploy`
Ping the webhook "insecurely", so it shows up in the debug logs

[DAQ-11838]: https://dt-rnd.atlassian.net/browse/DAQ-11838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/logd/logger.go
```diff
@@ -83,3 +83,9 @@ func (l Logger) WithValues(keysAndValues ...any) Logger {
 func (l *Logger) debugLog(message string, keysAndValues ...any) {
 	l.Logger.V(debugLogLevelElevation).Info(message, keysAndValues...)
 }
+
+func (l *Logger) Write(p []byte) (n int, err error) {
+	l.debugLog("HTTP-Server log", "msg", string(p))
```
0sewa0: maybe too specific message, or needs some doc string

cmd/main.go
```diff
@@ -15,6 +15,7 @@ limitations under the License.
 package main
 
 import (
+	"log"
```
andriisoldatenko:
```suggestion
	stdLog "log"
```

cmd/main.go
```diff
@@ -52,7 +53,8 @@ func rootCommand(_ *cobra.Command, _ []string) error {
 }
 
 func main() {
-	ctrl.SetLogger(log.Logger)
+	ctrl.SetLogger(_log.Logger)
+	log.SetOutput(&_log)
```
andriisoldatenko: I would keep how it was, and just name stdLog alias. It just less changes.
i mean `stdLog.SetOutput(&log)`

hack/make/deploy/deploy.mk
```diff
@@ -19,6 +20,7 @@ deploy: manifests/crd/helm
 			--set csidriver.enabled=$(ENABLE_CSI) \
 			--set manifests=true \
 			--set image="$(IMAGE_URI)" \
+			--set debugLogs=$(DEBUG_LOGS) \
 			--set debug=$(DEBUG)
```
andriisoldatenko, line 24: Just out of curiosity, I can't find `debug` value inside `values.yaml`. why we have second one?

0sewa0: that is more of a "hidden" value, it is for setting the webhook up for local debug, probably would be safer to rename it to `localDebug` or something

PR #5221, opened 2025-08-11
Move extension structs to separate package

avorima:
## Description

Move Extension types into separate package. Follow up of [#3895](https://github.com/Dynatrace/dynatrace-operator/pull/3895)

[DAQ-1369](https://dt-rnd.atlassian.net/browse/DAQ-1369)

## How can this be tested?

The change should have no impact on functionality. Run e2e tests to check.

[DAQ-1369]: https://dt-rnd.atlassian.net/browse/DAQ-1369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/api/latest/dynakube/extension/spec.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (e *Extensions) SetNamespace(namespace string) {
+	e.namespace = namespace
+}
+
+func (e *Extensions) SetEnabled(enabled bool) {
+	e.enabled = enabled
+}
+
+func (e *Extensions) Enabled() bool {
```
0sewa0:
```suggestion
func (e *Extensions) IsEnabled() bool {
```

to follow the pattern in other packages

pkg/api/latest/dynakube/extension/spec.go
```diff
@@ -0,0 +1,144 @@
...
+// +kubebuilder:object:generate=true
+
+type OpenTelemetryCollectorSpec struct {
+
+	// Number of replicas for your OtelCollector
+	// +kubebuilder:validation:Optional
+	Replicas *int32 `json:"replicas"`
+
+	CommonProperties `json:",inline"`
+}
```
0sewa0: otel is a special things:
it could be its own package, as it is also considered when `telemtryIngest` is used

pkg/api/latest/dynakube/extension/spec.go
```diff
@@ -0,0 +1,144 @@
...
+
+// +kubebuilder:object:generate=true
+
+// CommonProperties encapsulates common properties shared by all extensions.
+type CommonProperties struct {
```
0sewa0: In general, I am kind of against this.

Code reuse is great, but in this case it binds 2 things together unnecessarily

also, fun fact `TLSRefName string` is never used for the OpenTelemetryCollectorSpec part (not because of your change, but in general)

pkg/api/latest/dynakube/extension/spec.go
```diff
@@ -0,0 +1,144 @@
...
+	return "dynatrace" + consts.ExtensionsControllerSuffix + "-" + consts.ExtensionsCollectorTargetPortName
+}
+
+func (e *Extensions) ServiceNameFQDN() string {
+	return e.ServiceName() + "." + e.namespace
+}
+
+func (e *Extensions) ServiceName() string {
+	return e.name + consts.ExtensionsControllerSuffix
+}
```
0sewa0: we use the `Get` prefix for getting some field or value

which is "very Java of us", but we found it more readable in general

for consistency, lets stick to them

pkg/api/latest/dynakube/extension/spec.go
```diff
@@ -0,0 +1,144 @@
...
+	e.name = name
+}
+
+func (e *Extensions) SetNamespace(namespace string) {
+	e.namespace = namespace
+}
+
+func (e *Extensions) SetEnabled(enabled bool) {
+	e.enabled = enabled
+}
```
0sewa0: methods should be in `props.go`, the `spec.go` should only have the struct definition.

0sewa0 (commented):
Mainly commented to make it consistent with the other packages

If we want to change the general pattern, we can but we should agree on it

pkg/api/latest/dynakube/extension/spec.go
0sewa0: context for "why":
it often happen to use that a member of the `struct` clashed with the name of the method

example `spec.ServiceName` would clash with `func (s *Spec) ServiceName() string`

this caused the we would use `Get...` in such cases, which in turned cause that we had inconsistent method naming.

The same is the case for boolean fields, such as `IsEnabled()` vs `Enabled()`

pkg/api/latest/dynakube/extension/spec.go
avorima: so should i keep otel in the dynakube package then?

0sewa0: for now yes, that is the easiest, we can handle it separatly later

pkg/api/latest/dynakube/dynakube_types.go
```diff
@@ -91,7 +93,7 @@ type DynaKubeSpec struct { //nolint:revive
 	// When an (empty) ExtensionsSpec is provided, the extensions related components (extensions controller and extensions collector)
 	// are deployed by the operator.
 	// +kubebuilder:validation:Optional
-	Extensions *ExtensionsSpec `json:"extensions,omitempty"`
+	Extensions *extension.Spec `json:"extensions,omitempty"`
```
andriisoldatenko:
```suggestion
	Extensions *extensions.Spec `json:"extensions,omitempty"`
```
wdyt if we name it `extensionS` instead?

pkg/api/latest/dynakube/extension/props.go
```diff
@@ -0,0 +1,61 @@
+package extension
```
andriisoldatenko:
```suggestion
package extensions
```

pkg/api/latest/dynakube/extension/spec.go
```diff
@@ -1,15 +1,26 @@
...
+package extension
 
 import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/shared/image"
 	corev1 "k8s.io/api/core/v1"
 )
 
-// +kubebuilder:validation:Optional
-type ExtensionsSpec struct {
+type Extensions struct {
```
andriisoldatenko: to be consisten, here we call it Extensions, so I prefer to do same everywhere, already commented.

pkg/api/latest/dynakube/extension/spec.go
```diff
@@ -24,7 +35,7 @@ type ExtensionExecutionControllerSpec struct {
 
 	// Overrides the default image
 	// +kubebuilder:validation:Optional
-	ImageRef image.Ref `json:"imageRef,omitempty"`
+	ImageRef image.Ref `json:"imageRef"`
```
andriisoldatenko: out of curiosity, why did we decide to make it required here?

pkg/api/latest/dynakube/extension/spec.go
```diff
@@ -39,7 +50,7 @@ type ExtensionExecutionControllerSpec struct {
 
 	// Define resources' requests and limits for single ExtensionExecutionController pod
 	// +kubebuilder:validation:Optional
-	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+	Resources corev1.ResourceRequirements `json:"resources"`
```
andriisoldatenko: same here, I don't think we want to change behavior now.

pkg/api/latest/dynakube/extension/spec.go
avorima: i just removed the omitempty, because it has no effect

pkg/api/latest/dynakube/dynakube_types.go
avorima: i think in general package names should be singular, but in this case plural works.

andriisoldatenko: you are right, that's how go comunity suggests to do, but in our case it's extensions not single extension.

0sewa0 (approved):

PR #5222, opened 2025-08-11
Skip apimonitoring reconciler if missing optional scope

albertogdd:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-10639

ApiMonitoring reconciler now is skipped if `settings.read` or `settings.write` token scopes are missing.

I also refactored a bit the unit tests, because to test the reconcile through now we need to have both conditions set to true (check `NewDynakube()` function). I also removed dynakubes and ctx.Background() variables that were being reused on the tests. 

## How can this be tested?

- Create an API token without settings.read or settings.write (or both)
- Apply a DynaKube using the API token
- No errors should pop on the logs. But you should see this new line on "automatic-api-monitoring" logger: "api token missing optional scope, skipping reconciliation"

### DIFF
--- a/pkg/controllers/dynakube/activegate_test.go
+++ b/pkg/controllers/dynakube/activegate_test.go
@@ -64,7 +64,7 @@ func TestReconcileActiveGate(t *testing.T) {
 		require.Error(t, err)
 		require.Equal(t, "failed to reconcile ActiveGate: BOOM", err.Error())
 	})
-	t.Run(`reconcile disabled automatic kubernetes api monitoring`, func(t *testing.T) {
+	t.Run("reconcile disabled automatic kubernetes api monitoring", func(t *testing.T) {
 		dk := &dynakube.DynaKube{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      testName,
@@ -100,7 +100,7 @@ func TestReconcileActiveGate(t *testing.T) {
 		err := controller.reconcileActiveGate(ctx, dk, nil, nil)
 		require.NoError(t, err)
 	})
-	t.Run(`reconcile automatic kubernetes api monitoring`, func(t *testing.T) {
+	t.Run("reconcile automatic kubernetes api monitoring", func(t *testing.T) {
 		dk := &dynakube.DynaKube{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      testName,
@@ -138,7 +138,7 @@ func TestReconcileActiveGate(t *testing.T) {
 		err := controller.reconcileActiveGate(ctx, dk, nil, nil)
 		require.NoError(t, err)
 	})
-	t.Run(`reconcile automatic kubernetes api monitoring with custom cluster name`, func(t *testing.T) {
+	t.Run("reconcile automatic kubernetes api monitoring with custom cluster name", func(t *testing.T) {
 		const clusterLabel = "..blabla..;.🙃"
 
 		dk := &dynakube.DynaKube{
--- a/doc/coding-style-guide.md
+++ b/doc/coding-style-guide.md
@@ -326,6 +326,7 @@ func (controller *Controller) reconcileEdgeConnectDeletion(ctx context.Context, 
 - Use `<...>mock` as package import alias, in all cases, even if no alias would strictly be necessary.
   - Examples: `dtclientmock`, `controllermock`, `dtbuildermock`, `injectionmock`, `registrymock`
 - Use this structure: (or table-tests)
+  - The usage of `"` instead of ``` ` ``` in `t.Run` is important, as in VSCode you can't run individual tests if they are defined as ``` t.Run(`test`, ...) ```, but can when defined as``` t.Run("test", ...) ```.
 
 ```go
 func TestMyFunction(t *testing.T) {
@@ -333,7 +334,7 @@ func TestMyFunction(t *testing.T) {
     testString := "test"
 
     // Each test case of a function gets a t.Run
-    t.Run(`useful title`, func(t *testing.T) {
+    t.Run("useful title", func(t *testing.T) {
         // Arrange/Setup
         testInt := 1
 
@@ -345,7 +346,7 @@ func TestMyFunction(t *testing.T) {
         assert.Equal(t, out, testString)
     })
 
-    t.Run(`other useful title`, func(t *testing.T) {
+    t.Run("other useful title", func(t *testing.T) {
         // Arrange
         testInt := 4
 
--- a/pkg/api/latest/dynakube/dynakube_props_test.go
+++ b/pkg/api/latest/dynakube/dynakube_props_test.go
@@ -30,14 +30,14 @@ func TestTokens(t *testing.T) {
 	testName := "test-name"
 	testValue := "test-value"
 
-	t.Run(`GetTokensName returns custom token name`, func(t *testing.T) {
+	t.Run("GetTokensName returns custom token name", func(t *testing.T) {
 		dk := DynaKube{
 			ObjectMeta: metav1.ObjectMeta{Name: testName},
 			Spec:       DynaKubeSpec{Tokens: testValue},
 		}
 		assert.Equal(t, dk.Tokens(), testValue)
 	})
-	t.Run(`GetTokensName uses instance name as default value`, func(t *testing.T) {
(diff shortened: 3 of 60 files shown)


pkg/controllers/dynakube/activegate_test.go
```diff
@@ -64,7 +64,7 @@ func TestReconcileActiveGate(t *testing.T) {
 		require.Error(t, err)
 		require.Equal(t, "failed to reconcile ActiveGate: BOOM", err.Error())
 	})
-	t.Run(`reconcile disabled automatic kubernetes api monitoring`, func(t *testing.T) {
+	t.Run("reconcile disabled automatic kubernetes api monitoring", func(t *testing.T) {
```
0sewa0: vscode 💌

pkg/controllers/dynakube/apimonitoring/reconciler_test.go
```diff
@@ -179,28 +180,35 @@ func TestReconcile(t *testing.T) {
...
 		assert.Empty(t, actual)
 	})
+
+	t.Run("optional scope settings.write not available", func(t *testing.T) {
+		r := createReconcilerWithError(t, nil, errors.New("Unauthorized, missing token scopes"), nil, nil)
+		conditions.SetOptionalScopeMissing(&r.dk.Status.Conditions, dtclient.ConditionTypeAPITokenSettingsWrite, "not available")
+		// act
+		err := r.Reconcile(context.Background())
+
+		// assert
```
andriisoldatenko:
```suggestion
```

pkg/controllers/dynakube/apimonitoring/reconciler_test.go
```diff
@@ -179,28 +180,35 @@ func TestReconcile(t *testing.T) {
...
 
 		// assert
 		require.NoError(t, err)
 		assert.Empty(t, actual)
 	})
+
+	t.Run("optional scope settings.write not available", func(t *testing.T) {
+		r := createReconcilerWithError(t, nil, errors.New("Unauthorized, missing token scopes"), nil, nil)
+		conditions.SetOptionalScopeMissing(&r.dk.Status.Conditions, dtclient.ConditionTypeAPITokenSettingsWrite, "not available")
+		// act
```
andriisoldatenko:
```suggestion
```

pkg/controllers/dynakube/apimonitoring/reconciler_test.go
albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/5222/commits/96668ca81a9c2c974a74e66b93565cdd24f003f9)

pkg/controllers/dynakube/apimonitoring/reconciler_test.go
albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/5222/commits/96668ca81a9c2c974a74e66b93565cdd24f003f9)

pkg/controllers/dynakube/apimonitoring/reconciler_test.go
```diff
@@ -137,193 +141,150 @@ func createMonitoredEntities() dtclient.K8sClusterME {
...
-		_, err := r.createObjectIDIfNotExists(ctx)
+		_, err := r.createObjectIDIfNotExists(context.Background())
 
-		// assert
 		require.NoError(t, err)
 	})
 }
 
 func TestHandleKubernetesAppEnabled(t *testing.T) {
-	ctx := context.Background()
```
0sewa0, line 260: pls don't remove this

having to say `context.Background()` in every call is annoying
- also, it is kind of a pattern we do in "newer" test

using `context.Background()` is totaly safe to use between tests, as it is a not-cancelable context, and we do not use it in any specific way.
- so it is just kind of a "placeholder"

if we modified the context, or relied on it (for getting the logger or something) then `context.Background()` wouldn't cut it

if we want to move away from this, then we should do it everywhere (right now, this would just make it more inconsistent)

pkg/controllers/dynakube/activegate_test.go
```diff
@@ -100,7 +100,7 @@ func TestReconcileActiveGate(t *testing.T) {
 		err := controller.reconcileActiveGate(ctx, dk, nil, nil)
 		require.NoError(t, err)
 	})
-	t.Run(`reconcile automatic kubernetes api monitoring`, func(t *testing.T) {
+	t.Run("reconcile automatic kubernetes api monitoring", func(t *testing.T) {
```
andriisoldatenko: i posted in slack discussion, could you please revert this change, and we will fix it in 1 big PR

pkg/controllers/dynakube/apimonitoring/reconciler_test.go
```diff
@@ -99,7 +102,7 @@ func createReadOnlyReconciler(t *testing.T, dk *dynakube.DynaKube, monitoredEnti
 	return r
 }
 
-func createReconcilerWithError(t *testing.T, dk *dynakube.DynaKube, monitoredEntitiesError error, getSettingsResponseError error, createSettingsResponseError error, createAppSettingsResponseError error) Reconciler { //nolint:revive
+func createReconcilerWithError(t *testing.T, monitoredEntitiesError error, getSettingsResponseError error, createSettingsResponseError error, createAppSettingsResponseError error) Reconciler {
```
0sewa0, line 105: the change make sense, but there are bigger problems in this func, if you want to refactor things 
- you don't have to implement it, but would be nice :D

Example big red-flag:
```go
	for _, call := range mockClient.ExpectedCalls {
		call.Maybe()
	}
```
- this basically means, that the mock allows almost everything, therefore it is not much use

IMO, we should move away from these singular "magic create big object func", as you can't properly mock things this way (so you do `call.Maybe()`)
- we should init mocks per test case, not per package
- to avoid bloating the package with all the testing funcs you can define such testing funcs within the test-case, see [here](https://github.com/Dynatrace/dynatrace-operator/blob/8c3152487fba66f263521dab68f9174878980fde/pkg/clients/dynatrace/settings_test.go#L22-L23).
   - this would also avoid people randomly reusing such funcs without thinking and in consequence  binding test-cases together for no good reason

pkg/controllers/dynakube/activegate_test.go
albertogdd: Fixed

0sewa0 (commented):
functionally LGTM, just added some notes for the tests

pkg/controllers/dynakube/apimonitoring/reconciler_test.go
avorima: tests should use [t.Context](https://pkg.go.dev/testing#T.Context) unless the lifetime should exceed that of the surrounding test (which shouldn't be the case for unit tests).

andriisoldatenko: yeah it's a good catch, it's a feature of go 1.24 :) also again, let;s use it everywhere, so @avorima you can make a PR to refactor it in all places not only here. 


replace `context.Background()` to -> `t.Context()`

0sewa0: (i will add this to the style guide 💋 )

pkg/controllers/dynakube/apimonitoring/reconciler_test.go
albertogdd: Agree! can we do that in a separate ticket though? It will require some hours of debugging and switching stuff...
The dk refactoring was needed because the same instance was being reused + now updating conditions -> failed tests

pkg/controllers/dynakube/apimonitoring/reconciler_test.go
albertogdd: SGTM thanks for the input @avorima 
Can we then leave it as it is in the PR @0sewa0 and replace it with t.Context() in the future? :angel:

0sewa0: it is fine, created a PR for the style guide about it https://github.com/Dynatrace/dynatrace-operator/pull/5231

we can create follow ups to clean things up according to that

PR #5227, opened 2025-08-12
chore(e2e): fixup some targets from previous refactoring

andriisoldatenko:


aorcholski (approved):
LGTM I checked all names in the code and tested `extensions`, `telemetryingest`, `telemetryingest/local-active-gate-and-cleanup` targets

PR #5228, opened 2025-08-12
Use private struct field for k8ssecret.QueryObject

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-2124)

## Description

Adds `secretQuery` field to Reconciler structs if the query if performed more than once.

Removes `clt` and `apiReader` fields if they are not used. 

tls/reconciler and customproperties/reconciler use `clt` and `apiReader` instead of duplicated `clt`.

## How can this be tested?

unittests

### DIFF
--- a/pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
+++ b/pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
@@ -29,16 +29,16 @@ const (
 var _ controllers.Reconciler = &Reconciler{}
 
 type Reconciler struct {
-	dk          *dynakube.DynaKube
-	dtc         dtclient.Client
-	secretQuery k8ssecret.QueryObject
+	dk      *dynakube.DynaKube
+	dtc     dtclient.Client
+	secrets k8ssecret.QueryObject
 }
 
 func NewReconciler(clt client.Client, apiReader client.Reader, dk *dynakube.DynaKube, dtc dtclient.Client) *Reconciler {
 	return &Reconciler{
-		dk:          dk,
-		dtc:         dtc,
-		secretQuery: k8ssecret.Query(clt, apiReader, log),
+		dk:      dk,
+		dtc:     dtc,
+		secrets: k8ssecret.Query(clt, apiReader, log),
 	}
 }
 
@@ -65,7 +65,7 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 }
 
 func (r *Reconciler) reconcileAuthTokenSecret(ctx context.Context) error {
-	secret, err := r.secretQuery.Get(ctx, client.ObjectKey{Name: r.dk.ActiveGate().GetAuthTokenSecretName(), Namespace: r.dk.Namespace})
+	secret, err := r.secrets.Get(ctx, client.ObjectKey{Name: r.dk.ActiveGate().GetAuthTokenSecretName(), Namespace: r.dk.Namespace})
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
 			log.Info("creating activeGateAuthToken secret")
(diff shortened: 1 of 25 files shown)


pkg/controllers/dynakube/connectioninfo/activegate/reconciler.go
```diff
@@ -21,12 +21,11 @@ import (
 )
 
 type reconciler struct {
-	client       client.Client
 	apiReader    client.Reader
```
albertogdd: Consider refactoring  `IsTenantSecretPresent` function to receive SecretQuery parameter instead of apiReader, so you could also remove this attribute from the reconciler!

pkg/controllers/dynakube/connectioninfo/oneagent/reconciler.go
```diff
@@ -25,8 +25,8 @@ type reconciler struct {
 	apiReader    client.Reader
```
albertogdd: Same here

aorcholski: [71cf3b5](https://github.com/Dynatrace/dynatrace-operator/pull/5228/commits/71cf3b5bb4f128a2a06b813ad0cc6b71bffda502)

pkg/controllers/dynakube/connectioninfo/activegate/reconciler.go
aorcholski: [71cf3b5](https://github.com/Dynatrace/dynatrace-operator/pull/5228/commits/71cf3b5bb4f128a2a06b813ad0cc6b71bffda502)

albertogdd (approved):
Deployed it on my cluster and works fine 👍

pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
```diff
@@ -29,18 +29,16 @@ const (
...
 
 func NewReconciler(clt client.Client, apiReader client.Reader, dk *dynakube.DynaKube, dtc dtclient.Client) *Reconciler {
 	return &Reconciler{
-		client:    clt,
-		apiReader: apiReader,
-		dk:        dk,
-		dtc:       dtc,
+		dk:          dk,
+		dtc:         dtc,
+		secretQuery: k8ssecret.Query(clt, apiReader, log),
```
andriisoldatenko: IMO it's better to call it `secret` or `secretObject` and since secret Object is a bit too much I prefer call it secret.

why? bc when you read carefully usage code (secret query get - sounds a bit odd):
```
r.secretQuery.Get(
```

but instead:

```
r.secret.Get(

r.secret.WithOwner(r.dk).Create
```
because "query" - says we're querying something, and on top getting or creating

andriisoldatenko: cc @0sewa0 since you are the "creator" of this fancy name - wdyt? why you called a thing "Query" which returns QueryObject - but in really it returns core.Secret

andriisoldatenko: (another idea from @0sewa0 call it `secrets` not `secret`).

aorcholski: The `QueryObject` is not a `secret` or `secretObject`. I thought about `secretRequest` but IMHO `secrets` is better.

andriisoldatenko: It's not a request - check of implementation details . Yeah ok with me to call it secrets.

aorcholski: [9f4a32e](https://github.com/Dynatrace/dynatrace-operator/pull/5228/commits/9f4a32e19e7a8cbd39f6ac1e393b5e945ae9e9c1)

PR #5232, opened 2025-08-12
Deprecate autoUpdate field

avorima:
## Description

The autoUpdate field is confusing for users and should not be used anymore. Add deprecation warnings when it is defined on DynaKubes and also let the linter detect any usages in code.

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-11398)

## How can this be tested?

I tested it by running a modified version of an e2e test that set the field explicitly and observed the output.
The test should still pass and produce a warning when the DynaKube is written.


pkg/api/validation/dynakube/oneagent.go
```diff
@@ -194,6 +195,23 @@ func conflictingHostGroupSettings(_ context.Context, _ *Validator, dk *dynakube.
 	return ""
 }
 
+func deprecatedAutoUpdate(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
```
andriisoldatenko, line 198: am i miss something or you don't have tests for this new code?

andriisoldatenko: you can check `validation_test.go` how we test them

avorima: my bad. i added some

PR #5239, opened 2025-08-13
Adapt LogMonitoring to optional scope handling

waodim:
## Description

With this the LogModule rollout is adapted to the optional scopes.

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-10495)

For the LogModuleSettings Reconciler:

- if the `settings.read` scope is there - `GetSettingsForLogModule` can be performed
- if the `settings.write` scope is there - `CreateLogMonitoringSetting` can be performed

Conditions are set respectively if one of those scopes are missing.

For the LogModule DaemonSet reconciler:

- if all scopes are there the reconciler waits for the `ClusterMEID` is set - to avoid unnecessary restart
- otherwise if one of the scopes is missing the DaemonSet is still created but without certain metadata -> without the ClusterMEID (field: `dt.entity.kubernetes_cluster`) set in the args of the init container of the LogModule pods! If you then use a token with necessary scopes the change in the manifest (e.g. the arg is added) will trigger the pods to restart.

## How can this be tested?

- create tokens in different variations (with scopes, only with read, only with write, without scopes) and deploy the operator + dynakube with LogMon enabled. Check the Conditions, check that the DaemonSet is still rolled out, check the restart triggering by first using a token without some scopes and then using a token with all scopes -> after some time the logmon pods should restart
- unit tests
- e2e tests

### DIFF
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/args.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/args.go
@@ -6,12 +6,11 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
 )
 
-func getInitArgs(dk dynakube.DynaKube) []string {
+func getInitArgs(dk dynakube.DynaKube, useMetadata bool) []string {
 	baseArgs := []string{
 		fmt.Sprintf("-p k8s.cluster.name=$(%s)", clusterNameEnv),
 		fmt.Sprintf("-p k8s.cluster.uid=$(%s)", clusterUIDEnv),
 		fmt.Sprintf("-p k8s.node.name=$(%s)", nodeNameEnv),
-		fmt.Sprintf("-p dt.entity.kubernetes_cluster=$(%s)", entityEnv),
 		fmt.Sprintf("-c k8s_fullpodname $(%s)", podNameEnv),
 		fmt.Sprintf("-c k8s_poduid $(%s)", podUIDEnv),
 		fmt.Sprintf("-c k8s_basepodname $(%s)", basePodNameEnv),
@@ -20,6 +19,10 @@ func getInitArgs(dk dynakube.DynaKube) []string {
 		fmt.Sprintf("-c k8s_cluster_id $(%s)", clusterUIDEnv),
 		"-c k8s_containername " + containerName,
 		"-l " + dtLogVolumeMountPath,
+	}
+
+	if useMetadata {
+		baseArgs = append(baseArgs, fmt.Sprintf("-p dt.entity.kubernetes_cluster=$(%s)", entityEnv))
 	}
 
 	return append(baseArgs, dk.LogMonitoring().Template().Args...)
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/args_test.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/args_test.go
@@ -9,14 +9,15 @@ import (
 )
 
 const (
-	expectedBaseInitArgsLen = 12
+	expectedBaseInitArgsLen            = 12
+	expectedBaseInitArgsLenWithoutMEID = 11
 )
 
 func TestGetInitArgs(t *testing.T) {
 	t.Run("get base init args", func(t *testing.T) {
 		dk := dynakube.DynaKube{}
 		dk.Name = "dk-name-test"
-		args := getInitArgs(dk)
+		args := getInitArgs(dk, true)
 
 		assert.Len(t, args, expectedBaseInitArgsLen)
 
@@ -34,12 +35,23 @@ func TestGetInitArgs(t *testing.T) {
 				"customArg2",
 			},
 		}
-		args := getInitArgs(dk)
+		args := getInitArgs(dk, true)
 
 		assert.Len(t, args, expectedBaseInitArgsLen+len(dk.LogMonitoring().Args))
 
 		for _, customArg := range dk.LogMonitoring().Args {
 			assert.Contains(t, args, customArg)
+		}
+	})
+	t.Run("get base init args when no MEID or not all necessary scopes are set", func(t *testing.T) {
+		dk := dynakube.DynaKube{}
+		dk.Name = "dk-name-test"
+		args := getInitArgs(dk, false)
+
+		assert.Len(t, args, expectedBaseInitArgsLenWithoutMEID)
+
+		for _, arg := range args {
+			assert.NotEmpty(t, arg)
 		}
 	})
 }
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/container_test.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/container_test.go
@@ -80,7 +80,7 @@ func TestGetInitContainer(t *testing.T) {
 
 	t.Run("get main container", func(t *testing.T) {
 		dk := dynakube.DynaKube{}
-		initContainer := getInitContainer(dk, tenantUUID)
+		initContainer := getInitContainer(dk, tenantUUID, true)
 
 		require.NotEmpty(t, initContainer)
 
@@ -107,7 +107,7 @@ func TestGetInitContainer(t *testing.T) {
 				Tag:        expectedTag,
 			},
 		}
-		initContainer := getInitContainer(dk, tenantUUID)
+		initContainer := getInitContainer(dk, tenantUUID, true)
 
 		require.NotEmpty(t, initContainer)
 		assert.NotEmpty(t, initContainer.Image)
@@ -129,12 +129,23 @@ func TestGetInitContainer(t *testing.T) {
 				Limits:   limits,
 			},
 		}
-		initContainer := getInitContainer(dk, tenantUUID)
+		initContainer := getInitContainer(dk, tenantUUID, true)
 
 		require.NotEmpty(t, initContainer)
 		assert.NotEmpty(t, initContainer.Resources)
 		assert.Equal(t, requests, initContainer.Resources.Requests)
 		assert.Equal(t, limits, initContainer.Resources.Limits)
+	})
+	t.Run("get main container without the use of metadata", func(t *testing.T) {
+		dk := dynakube.DynaKube{}
+		initContainer := getInitContainer(dk, tenantUUID, false)
+
+		require.NotEmpty(t, initContainer)
+
+		assert.NotEmpty(t, initContainer.Args)
+		assert.Len(t, initContainer.Args, expectedBaseInitArgsLen-1)
+		assert.NotEmpty(t, initContainer.Env)
+		assert.Len(t, initContainer.Env, expectedBaseInitEnvLen-1)
 	})
 }
 
@@ -175,7 +186,7 @@ func TestSecurityContext(t *testing.T) {
 
 	t.Run("main and init container securityContext differ only in capabilities", func(t *testing.T) {
 		dk := dynakube.DynaKube{}
-		initContainer := getInitContainer(dk, tenantUUID)
+		initContainer := getInitContainer(dk, tenantUUID, false)
 		mainContainer := getContainer(dk, tenantUUID)
 
 		require.NotNil(t, initContainer)
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/env.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/env.go
@@ -27,8 +27,8 @@ const (
 	ruxitConfigPath = "/var/lib/dynatrace/oneagent/agent/config/ruxitagentproc.conf"
 )
 
-func getInitEnvs(dk dynakube.DynaKube) []corev1.EnvVar {
-	return []corev1.EnvVar{
+func getInitEnvs(dk dynakube.DynaKube, useMetadata bool) []corev1.EnvVar {
+	envs := []corev1.EnvVar{
 		{
 			Name: nodeNameEnv,
 			ValueFrom: &corev1.EnvVarSource{
@@ -70,14 +70,19 @@ func getInitEnvs(dk dynakube.DynaKube) []corev1.EnvVar {
 			Value: dk.Status.KubernetesClusterName,
 		},
 		{
-			Name:  entityEnv,
-			Value: dk.Status.KubernetesClusterMEID,
-		},
-		{
 			Name:  basePodNameEnv,
 			Value: dk.LogMonitoring().GetDaemonSetName(),
 		},
 	}
+
+	if useMetadata {
+		envs = append(envs, corev1.EnvVar{
+			Name:  entityEnv,
+			Value: dk.Status.KubernetesClusterMEID,
+		})
+	}
+
+	return envs
 }
 
 func GetKubeletEnvs() []corev1.EnvVar {
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/env_test.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/env_test.go
@@ -22,12 +22,29 @@ func TestGetInitEnvs(t *testing.T) {
 		dk.Status.KubernetesClusterMEID = "test-me-id"
 		dk.Status.KubernetesClusterName = "test-cluster-name"
 
-		envs := getInitEnvs(dk)
+		envs := getInitEnvs(dk, true)
 
 		assert.Len(t, envs, expectedBaseInitEnvLen)
 
 		for _, env := range envs {
 			hasValueOrFieldPath(t, env)
+		}
+	})
+	t.Run("get base init envs without MEID but all scopes set", func(t *testing.T) {
+		dk := dynakube.DynaKube{}
+		dk.Name = "dk-name-test"
+		dk.Status.KubeSystemUUID = "test-cluster-uuid"
+		dk.Status.KubernetesClusterMEID = ""
+		dk.Status.KubernetesClusterName = "test-cluster-name"
+
+		envs := getInitEnvs(dk, false)
+
+		assert.Len(t, envs, expectedBaseInitEnvLen-1)
+
+		for _, env := range envs {
+			if env.Name == "DT_ENTITY_KUBERNETES_CLUSTER" {
+				require.Empty(t, env.Value)
+			}
 		}
 	})
 }
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
@@ -4,11 +4,13 @@ import (
 	"context"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	dtclient "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/conditions"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/hasher"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/daemonset"
 	k8slabels "github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/labels"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/node"
+
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -57,13 +59,17 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 		return nil // clean-up shouldn't cause a failure
 	}
 
-	if !r.isMEConfigured() {
-		log.Info("Kubernetes settings are not yet available, which are needed for LogMonitoring, will requeue")
+	hasAllScopes := r.hasReadScope() && r.hasWriteScope()
+
+	if hasAllScopes && !r.isMEConfigured() {
+		log.Info("Kubernetes settings are not yet available and both settings.write and settings.read token scopes are availaible, will requeue")
 
 		return KubernetesSettingsNotAvailableError
 	}
 
-	ds, err := r.generateDaemonSet()
+	useMetadata := r.isMEConfigured() && hasAllScopes
+
+	ds, err := r.generateDaemonSet(useMetadata)
 	if err != nil {
 		return err
 	}
@@ -83,7 +89,7 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 	return nil
 }
 
-func (r *Reconciler) generateDaemonSet() (*appsv1.DaemonSet, error) {
+func (r *Reconciler) generateDaemonSet(useMetadata bool) (*appsv1.DaemonSet, error) {
 	tenantUUID, err := r.dk.TenantUUID()
 	if err != nil {
 		return nil, err
@@ -94,7 +100,7 @@ func (r *Reconciler) generateDaemonSet() (*appsv1.DaemonSet, error) {
 	maxUnavailable := intstr.FromInt(r.dk.FF().GetOneAgentMaxUnavailable())
 
 	ds, err := daemonset.Build(r.dk, r.dk.LogMonitoring().GetDaemonSetName(), getContainer(*r.dk, tenantUUID),
-		daemonset.SetInitContainer(getInitContainer(*r.dk, tenantUUID)),
+		daemonset.SetInitContainer(getInitContainer(*r.dk, tenantUUID, useMetadata)),
 		daemonset.SetAllLabels(labels.BuildLabels(), labels.BuildMatchLabels(), labels.BuildLabels(), r.dk.LogMonitoring().Template().Labels),
 		daemonset.SetAllAnnotations(nil, r.getAnnotations()),
 		daemonset.SetServiceAccount(serviceAccountName),
@@ -121,6 +127,14 @@ func (r *Reconciler) generateDaemonSet() (*appsv1.DaemonSet, error) {
 	}
 
 	return ds, nil
+}
+
+func (r *Reconciler) hasReadScope() bool {
+	return conditions.IsOptionalScopeAvailable(r.dk, dtclient.ConditionTypeAPITokenSettingsRead)
+}
+
+func (r *Reconciler) hasWriteScope() bool {
+	return conditions.IsOptionalScopeAvailable(r.dk, dtclient.ConditionTypeAPITokenSettingsWrite)
 }
 
 func (r *Reconciler) isMEConfigured() bool {
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
@@ -2,6 +2,7 @@ package daemonset
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
@@ -9,12 +10,14 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube/oneagent"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/scheme/fake"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/shared/communication"
+	dtclient "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/conditions"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/hasher"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
+
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -128,7 +131,7 @@ func TestReconcile(t *testing.T) {
 		assert.Equal(t, metav1.ConditionFalse, condition.Status)
 	})
 
-	t.Run("returns an error if no clusterMEID set", func(t *testing.T) {
+	t.Run("does not return an error if no clusterMEID set", func(t *testing.T) {
 		dk := createDynakube(true)
 		dk.Status.KubernetesClusterMEID = ""
 
@@ -139,7 +142,7 @@ func TestReconcile(t *testing.T) {
 
 		err := reconciler.Reconcile(context.Background())
 
-		require.Error(t, err)
+		require.NoError(t, err)
 	})
 }
 
@@ -148,7 +151,7 @@ func TestGenerateDaemonSet(t *testing.T) {
 		dk := createDynakube(true)
 
 		reconciler := NewReconciler(nil, fake.NewClient(), dk)
-		daemonset, err := reconciler.generateDaemonSet()
+		daemonset, err := reconciler.generateDaemonSet(true)
 		require.NoError(t, err)
 		require.NotNil(t, daemonset)
 
@@ -185,7 +188,7 @@ func TestGenerateDaemonSet(t *testing.T) {
 		}
 
 		reconciler := NewReconciler(nil, fake.NewClient(), dk)
-		daemonset, err := reconciler.generateDaemonSet()
+		daemonset, err := reconciler.generateDaemonSet(true)
 		require.NoError(t, err)
 		require.NotNil(t, daemonset)
 
@@ -205,7 +208,7 @@ func TestGenerateDaemonSet(t *testing.T) {
 		dk.Status.OneAgent.ConnectionInfoStatus.TenantTokenHash = testTokenHash
 
 		reconciler := NewReconciler(nil, fake.NewClient(), dk)
-		daemonset, err := reconciler.generateDaemonSet()
+		daemonset, err := reconciler.generateDaemonSet(true)
 		require.NoError(t, err)
 		require.NotNil(t, daemonset)
 
@@ -222,7 +225,7 @@ func TestGenerateDaemonSet(t *testing.T) {
 		}
 
 		reconciler := NewReconciler(nil, fake.NewClient(), dk)
-		daemonset, err := reconciler.generateDaemonSet()
+		daemonset, err := reconciler.generateDaemonSet(true)
 		require.NoError(t, err)
 		require.NotNil(t, daemonset)
 
@@ -238,7 +241,7 @@ func TestGenerateDaemonSet(t *testing.T) {
 		}
 
 		reconciler := NewReconciler(nil, fake.NewClient(), dk)
-		daemonset, err := reconciler.generateDaemonSet()
+		daemonset, err := reconciler.generateDaemonSet(true)
 		require.NoError(t, err)
 		require.NotNil(t, daemonset)
 
@@ -252,7 +255,7 @@ func TestGenerateDaemonSet(t *testing.T) {
 		dk.Spec.CustomPullSecret = customPullSecret
 
 		reconciler := NewReconciler(nil, fake.NewClient(), dk)
-		daemonset, err := reconciler.generateDaemonSet()
+		daemonset, err := reconciler.generateDaemonSet(true)
 		require.NoError(t, err)
 		require.NotNil(t, daemonset)
 
@@ -273,7 +276,7 @@ func TestGenerateDaemonSet(t *testing.T) {
 			Tolerations: customTolerations,
 		}
 		reconciler := NewReconciler(nil, fake.NewClient(), dk)
-		daemonset, err := reconciler.generateDaemonSet()
+		daemonset, err := reconciler.generateDaemonSet(true)
 		require.NoError(t, err)
 		require.NotNil(t, daemonset)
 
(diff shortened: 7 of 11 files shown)


pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler_test.go
```diff
@@ -116,11 +116,49 @@ func TestCheckLogMonitoringSettings(t *testing.T) {
...
 
-		err := r.checkLogMonitoringSettings(ctx)
+		err := r.checkLogMonitoringSettings(ctx, true, true)
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "error when creating")
 
 		mockClient.AssertCalled(t, "GetSettingsForLogModule", ctx, "meid")
 		mockClient.AssertCalled(t, "CreateLogMonitoringSetting", ctx, "meid", "cluster-name", mock.Anything)
 	})
+	t.Run("read-only settings exist -> can not create setting", func(t *testing.T) {
```
aorcholski:
```suggestion

	t.Run("read-only settings exist -> can not create setting", func(t *testing.T) {
```

pkg/controllers/dynakube/logmonitoring/daemonset/container_test.go
```diff
@@ -129,13 +129,24 @@ func TestGetInitContainer(t *testing.T) {
...
 		}
-		initContainer := getInitContainer(dk, tenantUUID)
+		initContainer := getInitContainer(dk, tenantUUID, true)
 
 		require.NotEmpty(t, initContainer)
 		assert.NotEmpty(t, initContainer.Resources)
 		assert.Equal(t, requests, initContainer.Resources.Requests)
 		assert.Equal(t, limits, initContainer.Resources.Limits)
 	})
+	t.Run("get main container without the use of metadata", func(t *testing.T) {
```
aorcholski, line 144:
```suggestion

	t.Run("get main container without the use of metadata", func(t *testing.T) {
```
and please add an empty line between 115-116

pkg/controllers/dynakube/logmonitoring/daemonset/container_test.go
```diff
@@ -129,13 +129,24 @@ func TestGetInitContainer(t *testing.T) {
...
+		dk := dynakube.DynaKube{}
+		initContainer := getInitContainer(dk, tenantUUID, false)
+
+		require.NotEmpty(t, initContainer)
+
+		assert.NotEmpty(t, initContainer.Args)
+		assert.Len(t, initContainer.Args, expectedBaseInitArgsLen-1)
```
aorcholski:
```suggestion
		assert.Len(t, initContainer.Args, expectedBaseInitArgsLenWithoutMEID)
```

pkg/controllers/dynakube/logmonitoring/daemonset/env_test.go
```diff
@@ -22,14 +22,31 @@ func TestGetInitEnvs(t *testing.T) {
...
 		}
 	})
+	t.Run("get base init envs without MEID but all scopes set", func(t *testing.T) {
+		dk := dynakube.DynaKube{}
+		dk.Name = "dk-name-test"
+		dk.Status.KubeSystemUUID = "test-cluster-uuid"
+		dk.Status.KubernetesClusterMEID = ""
+		dk.Status.KubernetesClusterName = "test-cluster-name"
+
+		envs := getInitEnvs(dk, false)
```
aorcholski: should be TRUE, we useMetadata , just the value is empty.
Otherwise the condition in #45 is never True

pkg/controllers/dynakube/logmonitoring/daemonset/env_test.go
```diff
@@ -22,14 +22,31 @@ func TestGetInitEnvs(t *testing.T) {
...
+
+		assert.Len(t, envs, expectedBaseInitEnvLen-1)
+
+		for _, env := range envs {
+			if env.Name == "DT_ENTITY_KUBERNETES_CLUSTER" {
```
aorcholski:
```suggestion
			if env.Name == entityEnv {
```

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
```diff
@@ -57,13 +58,17 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
...
+	hasAllScopes := r.hasReadScope() && r.hasWriteScope()
+
+	if hasAllScopes && !r.isMEConfigured() {
+		log.Info("Kubernetes settings are not yet available and both settings.write and settings.read token scopes are available, will requeue")
 
 		return KubernetesSettingsNotAvailableError
 	}
 
-	ds, err := r.generateDaemonSet()
+	useMetadata := r.isMEConfigured() && hasAllScopes
```
aorcholski:
```suggestion
	useMetadata := r.isMEConfigured()
```

I think it's my fault, the Description of this case could be better, sorry.
The purpose of this change is to always deploy LogMon module. There is no reason to NOT use MEID if it is available.

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
```diff
@@ -42,28 +42,66 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+	hasReadScope := conditions.IsOptionalScopeAvailable(r.dk, dtclient.ConditionTypeAPITokenSettingsRead)
+	hasWriteScope := conditions.IsOptionalScopeAvailable(r.dk, dtclient.ConditionTypeAPITokenSettingsWrite)
+
+	if !hasReadScope {
+		setLogMonitoringSettingError(r.dk.Conditions(), ConditionType, "settings.read scope missing: cannot query existing settings")
+	}
+
+	if !hasWriteScope {
+		setLogMonitoringSettingError(r.dk.Conditions(), ConditionType, "settings.write scope missing: cannot create or update settings")
+	}
```
0sewa0: you are overwriting the condition here, losing the `read` one.

Just "build" the message, and set the condition within the `if !hasReadScope && !hasWriteScope { ... }`

0sewa0: also, do you ever "unset" the condition, in case the scopes change?

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
```diff
@@ -42,28 +42,66 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+	err := r.checkLogMonitoringSettings(ctx, hasReadScope, hasWriteScope)
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func (r *reconciler) checkLogMonitoringSettings(ctx context.Context) error {
+func (r *reconciler) checkLogMonitoringSettings(ctx context.Context, hasReadScope, hasWriteScope bool) error {
```
0sewa0: you don't need `hasReadScope, hasWriteScope bool`,  no?

you only want to read the setting to check if you want to create it. (so you need write AND read for this func to make sense)

also you never reach this func as you return early in line 61

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
```diff
@@ -42,28 +42,66 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+
+	if !hasReadScope && !hasWriteScope {
+		log.Info("LogMonitoring settings are not available due to missing scopes, will skip reconciliation")
+
+		return nil
+	} else if !hasReadScope && r.dk.Status.KubernetesClusterMEID == "" {
+		log.Info("LogMonitoring settings are not yet available and settings.read scope is missing, will skip reconciliation")
+
+		return nil
+	}
```
0sewa0: IMO not necessary check 

you only want to read the setting to check if you want to create it. (so you need write AND read for this reconciler to make sense)

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
```diff
@@ -57,13 +58,17 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 		return nil // clean-up shouldn't cause a failure
 	}
 
-	if !r.isMEConfigured() {
-		log.Info("Kubernetes settings are not yet available, which are needed for LogMonitoring, will requeue")
+	hasAllScopes := r.hasReadScope() && r.hasWriteScope()
+
+	if hasAllScopes && !r.isMEConfigured() {
```
0sewa0: you only "need" the `read` check here

You could simplify this a lot, `useMetadata` is not necessary.

1. Check that you have the `read` scope
2. If you have the `read` scope but the ME is not configured, return early `return KubernetesSettingsNotAvailableError`
3. Everywhere else, set the ME if present
  - You don't need to do the extra check every time, if you returned early, you know that the state of the ME status is correct, so you can just configure accordingly 
    - if you have the `read` scope then the ME will be set, because otherwise you would have returned early
    - if you have no`read` scope  then the ME won't be set, which is ok

waodim: yeah I had a quick sync with @aorcholski and came to the conclusion that the scope check is unecessary here. I will update it accordingly. Thanks

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
```diff
@@ -42,28 +42,66 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+
+	if hasReadScope {
+		logMonitoringSettings, err = r.dtc.GetSettingsForLogModule(ctx, r.dk.Status.KubernetesClusterMEID)
+		if err != nil {
+			setLogMonitoringSettingError(r.dk.Conditions(), ConditionType, err.Error())
+
+			return errors.WithMessage(err, "error trying to check if setting exists")
+		}
+
+		if logMonitoringSettings.TotalCount > 0 {
```
aorcholski: This condition statement can be removed. The following statement in 107-113 is enough.

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
waodim: done - please doublee check if i missed something.

https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/01b8896bb3a6fd0db3afc677b606df94db638d2f

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
waodim: done - please doublee check if i missed something.

https://github.com/Dynatrace/dynatrace-operator/commit/01b8896bb3a6fd0db3afc677b606df94db638d2f

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
waodim: done - please doublee check if i missed something.

https://github.com/Dynatrace/dynatrace-operator/commit/01b8896bb3a6fd0db3afc677b606df94db638d2f

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
waodim: done - please doublee check if i missed something.

https://github.com/Dynatrace/dynatrace-operator/commit/01b8896bb3a6fd0db3afc677b606df94db638d2f

pkg/controllers/dynakube/logmonitoring/daemonset/env_test.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/db94178a24b11a62e094a5568163d9be7723b07c

pkg/controllers/dynakube/logmonitoring/daemonset/env_test.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/db94178a24b11a62e094a5568163d9be7723b07c

pkg/controllers/dynakube/logmonitoring/daemonset/container_test.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/db94178a24b11a62e094a5568163d9be7723b07c

pkg/controllers/dynakube/logmonitoring/daemonset/container_test.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/b8f5abc073e8faf16d9258711260f6202609a0e5

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler_test.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/b8f5abc073e8faf16d9258711260f6202609a0e5

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/01b8896bb3a6fd0db3afc677b606df94db638d2f

pkg/controllers/dynakube/logmonitoring/daemonset/args.go
```diff
@@ -22,5 +21,9 @@ func getInitArgs(dk dynakube.DynaKube) []string {
 		"-l " + dtLogVolumeMountPath,
 	}
 
+	if dk.Status.KubernetesClusterMEID != "" && dk.Status.KubernetesClusterName != "" {
```
0sewa0, line 23:
```suggestion
	if dk.Status.KubernetesClusterMEID != ""  {
```
we don't care about the `KubernetesClusterName`

pkg/controllers/dynakube/logmonitoring/daemonset/container_test.go
```diff
@@ -80,6 +80,8 @@ func TestGetInitContainer(t *testing.T) {
 
 	t.Run("get main container", func(t *testing.T) {
 		dk := dynakube.DynaKube{}
+		dk.Status.KubernetesClusterMEID = "test-me-id"
+		dk.Status.KubernetesClusterName = "test-cluster-name"
```
0sewa0, line 84:
```suggestion

```
we don't care about the `KubernetesClusterName`

pkg/controllers/dynakube/logmonitoring/daemonset/args_test.go
```diff
@@ -9,12 +9,15 @@ import (
...
-	expectedBaseInitArgsLen = 12
+	expectedBaseInitArgsLen            = 12
+	expectedBaseInitArgsLenWithoutMEID = 11
 )
 
 func TestGetInitArgs(t *testing.T) {
 	t.Run("get base init args", func(t *testing.T) {
 		dk := dynakube.DynaKube{}
+		dk.Status.KubernetesClusterMEID = "test-me-id"
+		dk.Status.KubernetesClusterName = "test-cluster-name"
```
0sewa0, line 20:
```suggestion

```
we don't care about the `KubernetesClusterName`

pkg/controllers/dynakube/logmonitoring/daemonset/env.go
```diff
@@ -69,15 +69,20 @@ func getInitEnvs(dk dynakube.DynaKube) []corev1.EnvVar {
...
-			Name:  entityEnv,
-			Value: dk.Status.KubernetesClusterMEID,
-		},
 		{
 			Name:  basePodNameEnv,
 			Value: dk.LogMonitoring().GetDaemonSetName(),
 		},
 	}
+
+	if dk.Status.KubernetesClusterMEID != "" && dk.Status.KubernetesClusterName != "" {
```
0sewa0, line 74:
```suggestion
	if dk.Status.KubernetesClusterMEID != "" {
```
we don't care about the `KubernetesClusterName`

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
```diff
@@ -39,31 +41,65 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+		missingScopes = append(missingScopes, "settings.write")
+	}
+
+	missingScopeThreshold := 2
+	if len(missingScopes) == missingScopeThreshold {
+		message := strings.Join(missingScopes, ", ") + " scope(s) missing: cannot query/create or update existing settings"
+		setLogMonitoringSettingError(r.dk.Conditions(), ConditionType, message)
+		log.Info("LogMonitoring settings are not available due to missing scopes, will skip reconciliation")
+
+		return nil
```
0sewa0:
```suggestion
	if len(missingScopes) > 0 {
		message := strings.Join(missingScopes, ", ") + " scope(s) missing: cannot query existing log monitoring settings and/or safely create new one."
		setLogMonitoringSettingError(r.dk.Conditions(), ConditionType, message)
		log.Info(message)

		return nil
```

you need both, so if any of them are missing, you should do nothing (also tweaked the message a bit)

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
```diff
@@ -39,31 +41,65 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+	} else {
+		log.Info("LogMonitoring settings are available, proceeding with reconciliation")
+
+		_ = meta.RemoveStatusCondition(r.dk.Conditions(), ConditionType)
```
0sewa0: why are you removing the condition when it should be actually there, saying "all is good" 😅

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
```diff
@@ -39,31 +41,65 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
+	err := r.checkLogMonitoringSettings(ctx, missingScopes)
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func (r *reconciler) checkLogMonitoringSettings(ctx context.Context) error {
+func (r *reconciler) checkLogMonitoringSettings(ctx context.Context, missingScopes []string) error {
```
0sewa0:
```suggestion
func (r *reconciler) checkLogMonitoringSettings(ctx context.Context) error {
```

this logic will not, and should not, run if any of the scopes are missing. so you do not need to handle that.

"reading" on its own makes no sense, as we do not store it or give it to anything
"creation" on its own is dangerous, because we are never supposed to update, and we should only create if nothing is there already (this is why we must "read" first)

waodim: I think I probably misunderstood the requirement in the ticket because I though we should pass getting if the reading scope is there and vice versa for creating. Thanks for clearing that up

0sewa0: lets just say, this is not our "clearest" of workflows 😅

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
waodim: Goes into the same direction as in the other comment I probably misunderstood something. Thanks I will fix it.

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/0fa731347c17909a02fb34155c85b9ff7dbe934b

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/0fa731347c17909a02fb34155c85b9ff7dbe934b

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/0fa731347c17909a02fb34155c85b9ff7dbe934b

pkg/controllers/dynakube/logmonitoring/daemonset/env.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/0fa731347c17909a02fb34155c85b9ff7dbe934b

pkg/controllers/dynakube/logmonitoring/daemonset/args_test.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/0fa731347c17909a02fb34155c85b9ff7dbe934b

pkg/controllers/dynakube/logmonitoring/daemonset/args_test.go
```diff
@@ -28,6 +30,8 @@ func TestGetInitArgs(t *testing.T) {
 	t.Run("add user defined args to existing init args", func(t *testing.T) {
 		dk := dynakube.DynaKube{}
 		dk.Name = "dk-name-test"
+		dk.Status.KubernetesClusterMEID = "test-me-id"
+		dk.Status.KubernetesClusterName = "test-cluster-name"
```
0sewa0, line 36:
```suggestion

```

pkg/controllers/dynakube/logmonitoring/daemonset/env_test.go
```diff
@@ -30,6 +31,23 @@ func TestGetInitEnvs(t *testing.T) {
 			hasValueOrFieldPath(t, env)
 		}
 	})
+
+	t.Run("get base init envs without MEID but all scopes set", func(t *testing.T) {
+		dk := dynakube.DynaKube{}
+		dk.Name = "dk-name-test"
+		dk.Status.KubeSystemUUID = "test-cluster-uuid"
+		dk.Status.KubernetesClusterName = "test-cluster-name"
```
0sewa0:
```suggestion

```

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
```diff
@@ -295,6 +300,78 @@ func TestGenerateDaemonSet(t *testing.T) {
 
 		assert.Equal(t, daemonset.Spec.Template.Spec.NodeSelector, customNodeSelector)
 	})
+
+	t.Run("generate a daemonset with no kubernetes cluster name set in env and arg section if no MEID and all scopes set", func(t *testing.T) {
+		dk := createDynakube(true)
+		dk.Status.KubernetesClusterMEID = ""
+		dk.Status.KubernetesClusterName = ""
```
0sewa0:
```suggestion
```

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
```diff
@@ -295,6 +300,78 @@ func TestGenerateDaemonSet(t *testing.T) {
...
+			}
+		}
+
+		require.False(t, found)
+	})
+
+	t.Run("both scopes set, MEID missing - wait, DS not created", func(t *testing.T) {
+		dk := createDynakube(true)
+		dk.Status.KubernetesClusterMEID = ""
+		dk.Status.KubernetesClusterName = ""
```
0sewa0:
```suggestion

```

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
```diff
@@ -295,6 +300,78 @@ func TestGenerateDaemonSet(t *testing.T) {
...
+		require.Contains(t, initContainer.Args, fmt.Sprintf("-p dt.entity.kubernetes_cluster=$(%s)", entityEnv))
+
+		found := false
+		for _, env := range initContainer.Env {
+			if env.Name == entityEnv && env.Value == dk.Status.KubernetesClusterMEID {
+				found = true
+
+				break
+			}
+		}
```
0sewa0: could use https://github.com/Dynatrace/dynatrace-operator/blob/95f6d633ba1c67b53ab5f85e4fffb5595078a156/pkg/util/kubeobjects/env/env.go#L16

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
```diff
@@ -295,6 +300,78 @@ func TestGenerateDaemonSet(t *testing.T) {
...
+
+		mockK8sClient := fake.NewClient()
+		reconciler := NewReconciler(mockK8sClient, mockK8sClient, dk)
+
+		err := reconciler.Reconcile(context.Background())
+		require.Error(t, err)
+		require.ErrorContains(t, err, "missing information about the kubernetes monitored-entity")
+
+		var ds appsv1.DaemonSet
+		err = mockK8sClient.Get(context.Background(), types.NamespacedName{Name: dk.LogMonitoring().GetDaemonSetName(), Namespace: dk.Namespace}, &ds)
```
0sewa0:
```suggestion
		err = mockK8sClient.Get(t.Context(), types.NamespacedName{Name: dk.LogMonitoring().GetDaemonSetName(), Namespace: dk.Namespace}, &ds)
```

nitpick: just remembered, we should start using `t.Context()`

(if this was the only thing, we could let it in, as there is a ticket to fix it everywhere, which is just a string replace 🤷 )

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
```diff
@@ -59,7 +84,12 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
 func (r *reconciler) checkLogMonitoringSettings(ctx context.Context) error {
 	log.Info("start reconciling log monitoring settings")
 
-	logMonitoringSettings, err := r.dtc.GetSettingsForLogModule(ctx, r.dk.Status.KubernetesClusterMEID)
+	var (
+		err                   error
+		logMonitoringSettings dtclient.GetLogMonSettingsResponse
+	)
+
+	logMonitoringSettings, err = r.dtc.GetSettingsForLogModule(ctx, r.dk.Status.KubernetesClusterMEID)
```
0sewa0:
```suggestion
	logMonitoringSettings, err := r.dtc.GetSettingsForLogModule(ctx, r.dk.Status.KubernetesClusterMEID)
```

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/c40a302c1769fa14320cc994173bf898c70faf46

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/c40a302c1769fa14320cc994173bf898c70faf46

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/c40a302c1769fa14320cc994173bf898c70faf46

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/c40a302c1769fa14320cc994173bf898c70faf46

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/c40a302c1769fa14320cc994173bf898c70faf46

pkg/controllers/dynakube/logmonitoring/daemonset/env_test.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/c40a302c1769fa14320cc994173bf898c70faf46

pkg/controllers/dynakube/logmonitoring/daemonset/args_test.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/c40a302c1769fa14320cc994173bf898c70faf46

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
```diff
@@ -58,7 +58,7 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 	}
 
 	if !r.isMEConfigured() {
-		log.Info("Kubernetes settings are not yet available, which are needed for LogMonitoring, will requeue")
+		log.Info("Kubernetes settings are not yet available and both settings.write and settings.read token scopes are available, will requeue")
```
0sewa0: you have to check if we have the `settings.read` for this. to know if it is "ok" that the MEid is missing
- `settings.read` is needed to get the MEid (done in different reconciler)
- we use the MEid here to pass it to the daemonset  

It is a separate issue to log-settings, where you need `settings.read` and `settings.write`
- `settings.read` is needed to get the MEid (done in different reconciler)
- we use the MEid here as `scope` for checking the log-settings, needing `settings.read` again
- we create a log-setting scoped to the MEid, using `settings.write`

0sewa0: I see this used to be kinda like what I am asking for, but was changed because https://github.com/Dynatrace/dynatrace-operator/pull/5239#discussion_r2276234078

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
```diff
@@ -295,6 +301,63 @@ func TestGenerateDaemonSet(t *testing.T) {
...
+		init := daemonset.Spec.Template.Spec.InitContainers[0]
+		require.NotContains(t, init.Args, fmt.Sprintf("-p dt.entity.kubernetes_cluster=$(%s)", entityEnv))
+
+		require.Nil(t, env.FindEnvVar(init.Env, entityEnv))
+	})
+
+	t.Run("both scopes set, MEID missing - wait, DS not created", func(t *testing.T) {
+		dk := createDynakube(true)
+		dk.Status.KubernetesClusterMEID = ""
+		setScopes(dk, true, true)
```
andriisoldatenko:
```suggestion
		setScopes(dk, true, true)
```
TBH i prefer to no use this pattern, cause it's super unclear what does it mean `setScopes(true, true)`
Instead, just create 2 similar test functions, don't forget to mark them as t.Helper and 
- setReadScopes
- setWriteScopes

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
```diff
@@ -295,6 +301,63 @@ func TestGenerateDaemonSet(t *testing.T) {
...
+		require.NotNil(t, daemonset)
+
+		init := daemonset.Spec.Template.Spec.InitContainers[0]
+		require.NotContains(t, init.Args, fmt.Sprintf("-p dt.entity.kubernetes_cluster=$(%s)", entityEnv))
+
+		require.Nil(t, env.FindEnvVar(init.Env, entityEnv))
+	})
+
+	t.Run("both scopes set, MEID missing - wait, DS not created", func(t *testing.T) {
+		dk := createDynakube(true)
```
andriisoldatenko, line 323:
```suggestion
		dk := createDynakube(true)
```
here as well,

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
0sewa0: So,

if only `settings.read` is missing ->
- It is ok that `!r.isMEConfigured()`, so we don't need to skip the standalone logmon daemonset reconcile if ME is not set
- we can't check the log-settings, so we skip the log-setting reconciler

if only `settings.write` is missing ->
- It is NOT ok that `!r.isMEConfigured()`, so we skip the standalone logmon daemonset reconcile if ME is not set
  - (this logic could be challenged)
- we can't create the log-settings, so we skip the log-setting reconciler

if both are missing ->
- It is ok that `!r.isMEConfigured()`, so we don't need to skip the standalone logmon daemonset reconcile if ME is not set
- we can't check the log-settings, so we skip the log-setting reconciler

0sewa0: > (this logic could be challenged)

We could remove this `!r.isMEConfigured()` early exit, and just say, that we provide the ME info to the daemonset if it is there, or not when it is not

If the logmon daemonset works without giving the info about the ME to it, then this logic is not necessary

however the ticket kinda has a point:

> if all optional scopes are assigned - the Log Module daemon set is not created if ClusterMEID or ClusterName is unknown - just to avoid unnecessary restart. ClusterMEID and ClusterName are set after a short period of time.

(however, you it is a bit incorrect as you don't need `settings.write` for this part)

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/3b451aaf990ee0ad0fb0949713a390a932305d32

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
waodim: fair enough, done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/3b451aaf990ee0ad0fb0949713a390a932305d32

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
waodim: as much as I like the thought to remove the check at all, I agree with you that the ticket has a point regarding the unnecessary restart.

Let's stick with the check for .isMEConfigured() and the read scope. I will do it now and await your feedback.

waodim: This is my attempt. Please have a look https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/04853ac59223ef2949cfa5e7543b2b7f487fcdd8

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler_test.go
```diff
@@ -344,3 +411,13 @@ func createBOOMK8sClient() client.Client {
...
+
+func setReadScope(t *testing.T, dk *dynakube.DynaKube) {
+	t.Helper()
+	meta.SetStatusCondition(dk.Conditions(), metav1.Condition{Type: dtclient.ConditionTypeAPITokenSettingsRead, Status: metav1.ConditionTrue})
+}
+
+func setWriteScope(t *testing.T, dk *dynakube.DynaKube) {
+	t.Helper()
+	meta.SetStatusCondition(dk.Conditions(), metav1.Condition{Type: dtclient.ConditionTypeAPITokenSettingsWrite, Status: metav1.ConditionTrue})
+}
```
0sewa0, line 423: ❤️

pkg/controllers/dynakube/logmonitoring/daemonset/args.go
```diff
@@ -11,7 +11,6 @@ func getInitArgs(dk dynakube.DynaKube) []string {
 		fmt.Sprintf("-p k8s.cluster.name=$(%s)", clusterNameEnv),
```
0sewa0: now I know why the `KubernetesClusterName` check was there

to be consistent, do not set this if `KubernetesClusterName` is empty (also not set the env)
- `!r.isMEConfigured()` already checks for its existence, so you can handle it the same way as `KubernetesClusterMEID`

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/41105c3339a2658168e6164fdda060da36651c8e

0sewa0 (dismissed):

pkg/controllers/dynakube/logmonitoring/daemonset/env.go
```diff
@@ -65,19 +65,25 @@ func getInitEnvs(dk dynakube.DynaKube) []corev1.EnvVar {
...
+		envs = append(envs, corev1.EnvVar{
```
andriisoldatenko, line 75: append support multi value append, nitpick but can be refactored as:
```
package main

import "fmt"

func main() {
	a := []int{}

	a = append(a, 1, 2)

	fmt.Println(a)
}
```

like this
```
envs  = append(envs, corev1.EnvVar{}, corev1.EnvVar{})
```

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
```diff
@@ -57,8 +58,8 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 		return nil // clean-up shouldn't cause a failure
 	}
 
-	if !r.isMEConfigured() {
-		log.Info("Kubernetes settings are not yet available, which are needed for LogMonitoring, will requeue")
+	if !r.isMEConfigured() && conditions.IsOptionalScopeAvailable(r.dk, dtclient.ConditionTypeAPITokenSettingsRead) {
+		log.Info("Kubernetes settings are not yet available and settings.read token scope is available, will requeue")
```
andriisoldatenko:
```suggestion
		log.Info("Kubernetes settings are not yet available and 'settings.read' token scope is available, will requeue")
```

pkg/controllers/dynakube/logmonitoring/logmonsettings/reconciler.go
```diff
@@ -39,11 +40,35 @@ func (r *reconciler) Reconcile(ctx context.Context) error {
...
-	} else if r.dk.Status.KubernetesClusterMEID == "" {
-		log.Info("Kubernetes settings are not yet available, which are needed for LogMonitoring, will requeue")
+	}
+
+	hasReadScope := conditions.IsOptionalScopeAvailable(r.dk, dtclient.ConditionTypeAPITokenSettingsRead)
+	hasWriteScope := conditions.IsOptionalScopeAvailable(r.dk, dtclient.ConditionTypeAPITokenSettingsWrite)
+
+	var missingScopes []string
+	if !hasReadScope {
+		missingScopes = append(missingScopes, "settings.read")
```
andriisoldatenko: we already have a constant https://github.com/dynatrace/dynatrace-operator/blob/2cc353d9e9820d1bbce663d5b66c976f2542c13b/pkg/clients/dynatrace/client.go#L120 can we resue it here?

waodim: thanks for mentioning, missed them! done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/4a3b7d1fd86a7edc826cceaf61f1e03c1b5cdc86

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/4a3b7d1fd86a7edc826cceaf61f1e03c1b5cdc86

pkg/controllers/dynakube/logmonitoring/daemonset/env.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5239/commits/4a3b7d1fd86a7edc826cceaf61f1e03c1b5cdc86

andriisoldatenko (approved):

0sewa0 (approved):

PR #5246, opened 2025-08-14
Temporarily fix GCR deployer image step

waodim:
## Description

Previously we always ran into the porblem that we built the GCR deployer image BEFORE updating the schema.yaml and creating the CSV bundle PR (example: https://github.com/Dynatrace/dynatrace-operator/pull/5060/commits/213d97c14394b2ceb85d3b4cbc8c54cba0b99cfe#diff-86e01e3f5885090a5bf9dcf296ef49cfd95097bfca4b251a40b866bb9966a22f) 

This is a temporary fix, that tries to solve that for at least releases `1.6.2` and `1.7.0`.

I commented out the original deployer step and created a new workflow file, that triggers if a PR is merged to `release-* `branches. There I added an if check that let's the job only run if the PR that got merged was from a branch with the prefix `csv-*`

This makes sure, that we only build the deployer image when the csv bundle PR is merged.

## How can this be tested?

You can have a look at the fork I created where I tried to simulate the triggering (from a csv branch to a release branch) https://github.com/waodim/dynatrace-operator/pull/20

Create your own fork and test it in a way that nothing gets pushed somewhere where it is not wanted.

### DIFF
--- a/.github/workflows/gcr-deployer-image.yaml
+++ b/.github/workflows/gcr-deployer-image.yaml
@@ -1,0 +1,63 @@
+name: GCR Deployer image
+on:
+  pull_request:
+    branches:
+      - release-*
+    types:
+      - closed
+
+env:
+  GOOGLE_MARKETPLACE_ANNOTATION: com.googleapis.cloudmarketplace.product.service.name=services/dynatrace-operator-dynatrace-marketplace-prod.cloudpartnerservices.goog
+  PLATFORMS: linux/amd64,linux/arm64,linux/ppc64le,linux/s390x
+  DEPLOYER_PLATFORMS: linux/amd64
+
+
+jobs:
+  prepare:
+    if: startsWith(github.event.pull_request.head.ref, 'csv-') && startsWith(github.event.pull_request.base.ref, 'release-')
+    name: Prepare properties
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Prepare build parameters
+        id: prep
+        run: |
+          hack/build/ci/prepare-build-variables.sh
+    outputs:
+      labels: ${{ steps.prep.outputs.docker_image_labels }}
+      version: ${{ steps.prep.outputs.docker_image_tag }}
+      version_without_prefix: ${{ steps.prep.outputs.docker_image_tag_without_prefix }}
+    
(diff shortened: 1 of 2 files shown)


.github/workflows/gcr-deployer-image.yaml
```diff
@@ -0,0 +1,64 @@
...
+        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # 6.18.0
+        with:
+          platforms: ${{ env.DEPLOYER_PLATFORMS }}
+          provenance: false
+          context: ./config/helm
+          file: ./config/helm/Dockerfile
+          push: true
+          tags: ${{ env.IMAGE }},${{ env.IMAGE_NO_PREFIX }}
+          annotations: |
+            ${{ env.GOOGLE_MARKETPLACE_ANNOTATION }}
```
andriisoldatenko:
```suggestion
            ${{ env.GOOGLE_MARKETPLACE_ANNOTATION }}

```

.github/workflows/release.yaml
```diff
@@ -163,38 +163,38 @@ jobs:
...
+  #       uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # 6.18.0
+  #       with:
+  #         platforms: ${{ env.DEPLOYER_PLATFORMS }}
+  #         provenance: false
+  #         context: ./config/helm
+  #         file: ./config/helm/Dockerfile
+  #         push: true
+  #         tags: ${{ env.IMAGE }},${{ env.IMAGE_NO_PREFIX }}
+  #         annotations: |
+  #          ${{ env.GOOGLE_MARKETPLACE_ANNOTATION }}
```
Incredible-O: I think we should just remove this code (for now) or provide context as to why it's commented out (maybe something like "moved to gcr-deployer-image.yaml as per DAQ-11982")

waodim: good idea, I added a comment to give some context https://github.com/Dynatrace/dynatrace-operator/pull/5246/commits/69426a9fe81fb6cb5cba6e6b600b950834fddcc8

andriisoldatenko: agree with @Incredible-O, remove code instead of commenting, if one needs it, he can check git history.

waodim: I am fine with that as well, removed: https://github.com/Dynatrace/dynatrace-operator/pull/5246/commits/76421463767936c99cb15e69a5f11ec8945f5d68

PR #4684, opened 2025-04-01
Revert "Migrate CI to build and distribute images using buildah...

wepudt:
fixing [DAQ-6857](https://dt-rnd.atlassian.net/browse/DAQ-6857)

## Description

Reverts https://github.com/Dynatrace/dynatrace-operator/pull/4617, and adds minor fixes like:
- use `type: oci`
- add `annotation....` for google marketplace

## How can this be tested?

Look at the built images, check that they are the correct arch, and that they have an annotation

[DAQ-6857]: https://dt-rnd.atlassian.net/browse/DAQ-6857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5247, opened 2025-08-14
Update contribution.md related to reverts and bugfixes

0sewa0:
## Description

Guideline idea about how to handle bug-fixes for unreleased feature from @andriisoldatenko.
- +some idea from about the general bug-fix process 

## How can this be tested?

Read and agree or disagree (or correct my English 😅 )

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,10 +18,35 @@ Make sure all the following are true when creating a pull-request:
 - The [coding style guide](doc/coding-style-guide.md) was followed when creating the change.
 - The PR has a meaningful title [guidelines](https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#use-imperative-mood-in-your-commit-message-subject).
 - The PR is labeled accordingly with a **single** label.
+- The PR has a link to a Jira ticket that requested the change. (If it exists)
 - Unit tests have been updated/added.
 - Relevant documentation has been updated/added.
   - [ARCHITECTURE.md](https://github.com/Dynatrace/dynatrace-operator/blob/main/ARCHITECTURE.md)
   - [Other docs](https://github.com/Dynatrace/dynatrace-operator/blob/main/doc)
+
+### Bug fixes
+
+There are 2 types of bugfixes, where the process differs.
+
+#### Bugfix for bugs in production
+
+1. Must be fixed first on the (release) branch where it will be first released. On this PR use the `bugfix` label.
+2. Cherry picked to `main` afterwards. On this PR use the `cherrypick` label.
+
+##### Reasoning
+
+We have the labels, but using them inconsistently makes them useless.
+
+#### Bugfix for bugs in main (ie.: not released), example: for new features
+
+1. The initial PR introducing the (buggy) feature must be **reverted**.
+2. (re)Implement feature with fix in it. The bugfix must be in separate commits in the PR, at the end, these commits should have the bugfix ticket in their description.
+
+##### Reasoning
+
+We should aim to have each commit as valid/working state of the repo. It can help us cherry-pick/revert things in the future.
+
+- Cherry-picking/reverting is rather error-prone when you need to find the 2 initial (flawed) implementation PRs and the 5 small bug-fix PR for each, and if you miss one, nothing will work.
 
 ## Quick start
 


CONTRIBUTING.md
```diff
@@ -18,11 +18,40 @@ Make sure all the following are true when creating a pull-request:
...
+We have the labels, but using them inconsistently makes them useless.
+
+#### Bugfix for bugs in main (ie.: not released), example: for new features
+
+1. The initial PR introducing the (buggy) feature must be **reverted**.
+2. (re)Implement feature with fix in it. The bugfix must be in separate commits in the PR, at the end, these commits should have the bugfix ticket in their description.
+
+##### Reasoning
+
+We should aim to have each commit as valid/working state of the repo. It can help us cherry-pick/revert things in the future.
```
andriisoldatenko:
```suggestion
We must aim to have each commit as valid/working state of the repo. It can help us cherry-pick/revert things in the future. 
> [!WARNING]  
> If the main is broken, the first step is to revert and unblock everyone, and then decide how to fix it.
```

albertogdd: I mean, even if you revert the commit (with a new commit, I assume), before the revert the state of the main branch is and will be 'broken' anyways, so I don't think this new approach improves this too much.

I agree that 'reimplementing' the feature makes it easier to find the right commit that introduces it, but IMO it is a bit overkill for the benefits we get... If we document properly which commit introduces a feature and which ones fix bugs about it there should be no problem.

albertogdd: Also, the PR/commit reimplementing the feature would contain the whole thing, so as a PR reviewer finding the key changes that fix the bug would be super annoying.

CONTRIBUTING.md
```diff
@@ -18,11 +18,40 @@ Make sure all the following are true when creating a pull-request:
...
+
+There are 2 types of bugfixes, where the process differs.
+
+But a general rule:
+
+- Always add/update unittest when doing a bugfix, to make sure it doesn't happen again.
+
+#### Bugfix for bugs in production
+
+1. Must be fixed first on the (release) branch where it will be first released. On this PR use the `bugfix` label.
```
waodim:
```suggestion
1. Must be fixed first on the (release) branch where it will be first released. On this PR use the `bug` label.
```

CONTRIBUTING.md
```diff
@@ -18,11 +18,40 @@ Make sure all the following are true when creating a pull-request:
...
+2. Cherry picked to `main` afterwards. On this PR use the `cherrypick` label.
+
+##### Reasoning
+
+We have the labels, but using them inconsistently makes them useless.
+
+#### Bugfix for bugs in main (ie.: not released), example: for new features
+
+1. The initial PR introducing the (buggy) feature must be **reverted**.
+2. (re)Implement feature with fix in it. The bugfix must be in separate commits in the PR, at the end, these commits should have the bugfix ticket in their description.
```
waodim:
```suggestion
2. (re)Implement feature with the fix in it. The bugfix must be in separate commits in the PR, at the end, these commits should have the bugfix ticket in their description.
```

CONTRIBUTING.md
0sewa0: > I mean, even if you revert the commit (with a new commit, I assume), before the revert the state of the main branch is and will be 'broken' anyways, so I don't think this new approach improves this too much.

true, so the reasoning should be cleared up :D 

It is more about the having clarity that a certain commit is good or not. If it was at some point reverted (which you would see in history), that means that this commit is incorrect state. If you just do bugfixes for it randomly then you muddy the water.

>  IMO it is a bit overkill for the benefits we get

Using `git revert` should be simple and we should use it more often. Mainly to achieve what @andriisoldatenko is mentioning [here](https://github.com/Dynatrace/dynatrace-operator/pull/5247#discussion_r2276120673)

And the "reimplementing" part is also rather simple git workflow: New branch -> cherry initial buggy change -> implement/commit fix -> create new PR
- It not as scary or annoying as it sounds, I guarantee you that "finding which quick fix is for which initial PR" a year later is far more annoying

It may sound as an overkill if we have never done it, don't worry, it is rather straight forward.
Also, we can/should still refine the process (on what exactly we need to do maybe in the PR description or like that)

0sewa0: > Also, the PR/commit reimplementing the feature would contain the whole thing, so as a PR reviewer finding the key changes that fix the bug would be super annoying.

That is why I said: `The bugfix must be in separate commit(s) in the PR, at the end, these commits should have the bugfix ticket in their description.`

albertogdd: But the commit that re-implements the feature doesn't contain the bug, so how are you going to make another commit fixing the bug? :nerd_face:

CONTRIBUTING.md
```diff
@@ -18,11 +18,40 @@ Make sure all the following are true when creating a pull-request:
...
+1. Must be fixed first on the (release) branch where it will be first released. On this PR use the `bugfix` label.
+2. Cherry picked to `main` afterwards. On this PR use the `cherrypick` label.
+
+##### Reasoning
+
+We have the labels, but using them inconsistently makes them useless.
+
+#### Bugfix for bugs in main (ie.: not released), example: for new features
+
+1. The initial PR introducing the (buggy) feature must be **reverted**.
```
avorima, line 48: This works well if the commit is recent, but becomes less viable with commits where a revert is not straightforward, e.g. if another feature builds on top of the feature that introduced the bug. I would refrain from demanding a revert take place in all circumstances and rather make this a recommendation to allow case-by-case decisions.

CONTRIBUTING.md
0sewa0: So how it should works:

1. You have a commit `Add my feature`
2. You revert it a few days later, then you will have the `(revert) Add my feature`
3. You create new PR with fixes, that has 2 commits. `1. Add my feature` and `2.Fix my feature` (unless you need total reimplementation, at which point IMO you need a full review again anyway)

but I can see your point that it can be a bit extreme in some cases (also mentioned by @avorima here https://github.com/Dynatrace/dynatrace-operator/pull/5247#discussion_r2276295550 )

So some easy rules we could define here:
- Breaks e2e tests -> revert
- (?) No other change rely on it at the time of finding the bug -> revert

wdyt?

Also, we could do something like "Original Jira ticket has to be linked to the bugs" or something to have the tracking.

andriisoldatenko: Just as a note, we're already following this process. There are a few examples in the history of PRs where we reverted and created a new PR + fix. It's a well-known approach in repos where few developers work simultaneously. 

Few examples: 

https://github.com/Dynatrace/dynatrace-operator/pull/4397
https://github.com/Dynatrace/dynatrace-operator/pull/4684

andriisoldatenko: you can find more by searching https://github.com/Dynatrace/dynatrace-operator/pulls?page=2&q=is%3Apr+revert+is%3Aclosed

albertogdd: @andriisoldatenko Yep, I'm fine with it in some cases where we want to revert the whole thing and rethink it. My problem is with doing it with every single bug we find.

@0sewa0 okay so then what you mean with re-implementating is:
- Cherry pick commit that got reverted
- Make a new commit fixing it on top of it
- Create PR (containing both commits)
Do I understand it right?

andriisoldatenko: >My problem is with doing it with every single bug we find.

@albertogdd maybe wording is not perfect idk, but I read it differently. It's says nothing about "every bug".
The purpose of documenting - because:
1) we already doing it, 
2) and scope is pretty narrow so something was not released and mostly merged to main.

0sewa0: > Do I understand it right?

yes

> revert the whole thing and rethink it.

most of the time it is not a big rethink, just 1 forgotten edge-case, should be rather clean


the main goal would be to recognize `revert` as a nice tool to quickly unblock `main` in case there is a problem.
- because right now it feels like it is a "forbidden" tool, and that "revert == bad" and kind of a "mark of shame"
  - It is great for preventing others from using and building their changes on know to be buggy code
  - It gives the original implementer time to properly fix and test the change in peace

0sewa0: reworked it

CONTRIBUTING.md
0sewa0: done

CONTRIBUTING.md
0sewa0: done

CONTRIBUTING.md
0sewa0: reworked it

CONTRIBUTING.md
```diff
@@ -18,11 +18,51 @@ Make sure all the following are true when creating a pull-request:
...
+
+1. Create a PR with the fix
+2. In the description of the PR, mention the original change that introduced the bug.
+3. Link the related Jira bug ticket to the original implementation Jira ticket.
+
+##### Reasoning
+
+- We must aim to have `main` to be in a correct state by end of day. So that we do not (re)test known to be buggy state.
+- Using `revert` quickly can quickly unblock others and allow the proper fix to be not rushed due to time pressure.
+- Cherry-picking 1 commit is easier than cherry-picking 2 commits where the 2. commit is bugfix for the 1.commit, but you have several unrelated commits between them.
```
avorima: AFAIK, in english ordinal numbers tend to be written with the specific numeric suffix.

```suggestion
- Cherry-picking 1 commit is easier than cherry-picking 2 commits where the 2nd commit is the bugfix for the 1st commit, but you have several unrelated commits between them.
```

PR #5249, opened 2025-08-14
Add warning if no otelcollector image is provided for telemetry ingest

chrismuellner:
## Description

[DAQ-12442](https://dt-rnd.atlassian.net/browse/DAQ-12442)

Telemetry ingest can currently be used without providing an OTel collector image, but it should be obvious to users via warning that that is not the recommended way!

## How can this be tested?

Apply the `telemetryIngest` sample for the Dynakube (in this PR) and check for the warning with/without `templates.otelCollector.imageRef...` section!

[DAQ-12442]: https://dt-rnd.atlassian.net/browse/DAQ-12442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/api/validation/dynakube/telemetryservice_test.go
+++ b/pkg/api/validation/dynakube/telemetryservice_test.go
@@ -316,7 +316,7 @@ func TestForbiddenSuffix(t *testing.T) {
 }
 
 func TestImages(t *testing.T) {
-	t.Run(`otel collector image`, func(t *testing.T) {
+	t.Run(`otel collector image missing`, func(t *testing.T) {
 		assertAllowedWithWarnings(t, 1,
 			&dynakube.DynaKube{
 				ObjectMeta: defaultDynakubeObjectMeta,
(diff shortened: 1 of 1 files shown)


pkg/api/validation/dynakube/telemetryservice_test.go
```diff
@@ -313,3 +314,35 @@ func TestForbiddenSuffix(t *testing.T) {
 			})
 	})
 }
+
+func TestImages(t *testing.T) {
+	t.Run(`otel collector image`, func(t *testing.T) {
```
andriisoldatenko: We decided recently to use double quotes here in t.Run

chrismuellner: is that also relevant for the 1.6 release branch? Rest of the file has the same quotes and I do not want to start changing unrelated code now.

andriisoldatenko: You are right it's related to main do ignore my comment.

waodim (approved):
Code lgtm - tested it and got the warning for both repo and tag

PR #5265, opened 2025-08-18
Fix deprecation warning for azure/setup-helm

avorima:
## Description

I see the following line in some workflow logs:

```
Warning: Input 'token' has been deprecated with message: GitHub token is no longer required
```

The source of this is the azure/setup-helm action.

## How can this be tested?

Run a workflow that includes the run-e2e or build-helm actions, e.g. `e2e-tests-ondemand.yaml` and check output.


avorima:
@Incredible-O you're right

PR #5280, opened 2025-08-21
refactor dtpull secret

andriisoldatenko:
## Description

This PR refactors dtpull secret reconciler and resolves [DAQ-1457](https://dt-rnd.atlassian.net/browse/DAQ-1457)

## How can this be tested?

a) unit-tests
b) deploy operator and simple DK,

- find `<dk-name>-pull-secret` via:
```
k get secrets | grep "\-pull-secret"
dynakube-pull-secret                                 kubernetes.io/dockerconfigjson   1      8m19s
```
so make sure it;s created;

- Also, you can check the condition on DK, which we use to make sure it gets updated in time:

```
Last Transition Time:  2025-08-21T07:45:40Z                                                                                              │
│     Message:               dynakube-pull-secret updated                                                                                      │
│     Reason:                SecretUpdated                                                                                                     │
│     Status:                True 
```
- and check age (wait 15 mins default )

```
     Last Transition Time:  2025-08-21T08:00:42Z                                                                                              │
│     Message:               dynakube-pull-secret updated                                                                                      │
│     Reason:                SecretUpdated                                                                                                     │
│     Status:                True                                                                                                              │
│     Type:                  PullSecret`
```

🎉 

[DAQ-1457]: https://dt-rnd.atlassian.net/browse/DAQ-1457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controllers/dynakube/dtpullsecret/reconciler.go
+++ b/pkg/controllers/dynakube/dtpullsecret/reconciler.go
@@ -2,12 +2,11 @@ package dtpullsecret
 
 import (
 	"context"
-	"reflect"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/token"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/conditions"
-	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/secret"
+	k8ssecret "github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/secret"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/timeprovider"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -21,20 +20,18 @@ const (
 )
 
 type Reconciler struct {
-	client       client.Client
-	apiReader    client.Reader
 	dk           *dynakube.DynaKube
 	tokens       token.Tokens
 	timeprovider *timeprovider.Provider
+	secrets      k8ssecret.QueryObject
 }
 
 func NewReconciler(clt client.Client, apiReader client.Reader, dk *dynakube.DynaKube, tokens token.Tokens) *Reconciler {
 	return &Reconciler{
-		client:       clt,
-		apiReader:    apiReader,
 		dk:           dk,
 		tokens:       tokens,
 		timeprovider: timeprovider.New(),
+		secrets:      k8ssecret.Query(clt, apiReader, log),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


albertogdd (approved):
Works fine, tested on my cluster

pkg/controllers/dynakube/dtpullsecret/reconciler.go
```diff
@@ -69,99 +65,52 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
...
+		conditions.SetKubeAPIError(r.dk.Conditions(), PullSecretConditionType, errors.WithMessage(err, "failed to create or update secret"))
 
-	pullSecret.Data = desiredPullSecretData
-	if err := r.client.Update(ctx, pullSecret); err != nil {
-		conditions.SetKubeAPIError(r.dk.Conditions(), PullSecretConditionType, err)
+		return errors.WithMessage(err, "failed to create or update secret")
+	}
 
-		return errors.WithMessagef(err, "failed to update secret %s", pullSecret.Name)
+	if created {
```
0sewa0: this is `true` in both the Update and Create scenarios

`CreateOrUpdate` will return true if it created or updated the secret, it returns false if it did nothing (or errored)

andriisoldatenko: oopsii

0sewa0: > oopsii

It is not the most obvious logic 😅 

I would just use the `SetSecretUpdated` conditions, but using only`SetSecretCreated` is also fine

andriisoldatenko: fixed in b4491659f

andriisoldatenko: i made it align with other places `SetSecretCreated`

https://github.com/dynatrace/dynatrace-operator/blob/dffd56671eedc03889c451dc6617b0feea5cec9b/pkg/controllers/dynakube/connectioninfo/oneagent/reconciler.go#L170

https://github.com/dynatrace/dynatrace-operator/blob/dffd56671eedc03889c451dc6617b0feea5cec9b/pkg/controllers/dynakube/extension/secret.go#L84

etc..


yes indeed semantic of CreateOrUpdate must retrun (true if created, and false if updated) atleast based on my prev Django experince https://docs.djangoproject.com/en/5.2/ref/models/querysets/#update-or-create anyway.

0sewa0 (dismissed):

pkg/controllers/dynakube/dtpullsecret/reconciler.go
0sewa0: we can change that if you want 👀 (in a separate change, in a separate ticket)

andriisoldatenko: For now, let's stay consistent. It's more about whether we want to separate the Condition for secret (create/update). If not, it's okay as is.

PR #5281, opened 2025-08-21
Skip flaky test

0sewa0:
## Description

[DAQ-12605](https://dt-rnd.atlassian.net/browse/DAQ-12605)

Skips the test that is currently inconsistent between envs.
We need to look into it more deeply, and maybe rework the tests in general.

## How can this be tested?

`make test/e2e/cloudnative/codemodules-with-proxy-custom-ca-ag-cert` is skipped
`make test/e2e/istio` runs through (expect the one above, should be still skipped)

### DIFF
--- a/test/features/cloudnative/codemodules/codemodules.go
+++ b/test/features/cloudnative/codemodules/codemodules.go
@@ -330,6 +330,8 @@ func WithProxyAndAutomaticAGCert(t *testing.T, proxySpec *value.Source) features
 }
 
 func WithProxyCAAndAGCert(t *testing.T, proxySpec *value.Source) features.Feature {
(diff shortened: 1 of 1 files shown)


test/features/cloudnative/codemodules/codemodules.go
```diff
@@ -330,6 +330,8 @@ func WithProxyAndAutomaticAGCert(t *testing.T, proxySpec *value.Source) features
 }
 
 func WithProxyCAAndAGCert(t *testing.T, proxySpec *value.Source) features.Feature {
+	t.Skip("skipping test, as it has inconsistent behavior, will be investigated/fixed in DAQ-12605")
```
andriisoldatenko: pls move this skip to test level not feature level.

0sewa0: done

PR #5289, opened 2025-08-25
Update to OCP 4.12+

StefanHauth:
## Description

Support for OCP 4.10 & 4.11 ended in March 2025. Up the version.

## How can this be tested?
n/a

### DIFF
--- a/config/helm/chart/default/templates/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring.yaml
+++ b/config/helm/chart/default/templates/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring.yaml
@@ -100,6 +100,22 @@ rules:
       - list
       - watch
       - get
+  - apiGroups:
+      - discovery.k8s.io
+    resources:
+      - endpointslices
+    verbs:
+      - list
+      - watch
+      - get
+  - apiGroups:
+      - autoscaling
+    resources:
+      - horizontalpodautoscalers
+    verbs:
+      - list
+      - watch
+      - get
   - nonResourceURLs:
       - /metrics
       - /version
--- a/config/helm/chart/default/tests/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring_test.yaml
+++ b/config/helm/chart/default/tests/Common/kubernetes-monitoring/clusterrole-kubernetes-monitoring_test.yaml
@@ -122,6 +122,28 @@ tests:
       - contains:
           path: rules
           content:
+            apiGroups:
+              - discovery.k8s.io
+            resources:
+              - endpointslices
+            verbs:
+              - list
+              - watch
+              - get
+      - contains:
+          path: rules
+          content:
+              apiGroups:
+              - autoscaling
+              resources:
+              - horizontalpodautoscalers
+              verbs:
+              - list
+              - watch
+              - get
+      - contains:
+          path: rules
+          content:
             nonResourceURLs:
               - /metrics
               - /version


hack/build/bundle.sh
```diff
@@ -67,13 +67,13 @@ grep -v '# Labels for testing.' "./config/olm/${PLATFORM}/bundle-${VERSION}.Dock
 mv "./config/olm/${PLATFORM}/bundle-${VERSION}.Dockerfile.output" "./config/olm/${PLATFORM}/bundle-${VERSION}.Dockerfile"
 if [ "${PLATFORM}" = "openshift" ]; then
   # shellcheck disable=SC2129
-	echo 'LABEL com.redhat.openshift.versions="v4.10"' >> "./config/olm/${PLATFORM}/bundle-${VERSION}.Dockerfile"
+	echo 'LABEL com.redhat.openshift.versions="v4.12"' >> "./config/olm/${PLATFORM}/bundle-${VERSION}.Dockerfile"
```
Incredible-O: Not specifically related to this PR, but what is the rationale behind ignoring the shellcheck warning? Maybe it could be addressed like so:

```bash
{
  echo 'LABEL com.redhat.openshift.versions="v4.12"'
  echo 'LABEL com.redhat.delivery.operator.bundle=true'
  echo 'LABEL com.redhat.delivery.backport=true'
} >> "./config/olm/${PLATFORM}/bundle-${VERSION}.Dockerfile"
```

hack/build/bundle.sh
```diff
@@ -67,13 +67,13 @@ grep -v '# Labels for testing.' "./config/olm/${PLATFORM}/bundle-${VERSION}.Dock
...
 	echo 'LABEL com.redhat.delivery.operator.bundle=true' >> "./config/olm/${PLATFORM}/bundle-${VERSION}.Dockerfile"
 	echo 'LABEL com.redhat.delivery.backport=true' >> "./config/olm/${PLATFORM}/bundle-${VERSION}.Dockerfile"
 	sed 's/\bkubectl\b/oc/g' "./config/olm/${PLATFORM}/${VERSION}/manifests/dynatrace-operator.v${VERSION}.clusterserviceversion.yaml" > "./config/olm/${PLATFORM}/${VERSION}/manifests/dynatrace-operator.v${VERSION}.clusterserviceversion.yaml.output"
 	mv "./config/olm/${PLATFORM}/${VERSION}/manifests/dynatrace-operator.v${VERSION}.clusterserviceversion.yaml.output" "./config/olm/${PLATFORM}/${VERSION}/manifests/dynatrace-operator.v${VERSION}.clusterserviceversion.yaml"
-        # allow installation on v4.10 and beyond
-	echo '  com.redhat.openshift.versions: v4.10' >> "./config/olm/${PLATFORM}/${VERSION}/metadata/annotations.yaml"
+        # allow installation on v4.12 and beyond
+	echo '  com.redhat.openshift.versions: v4.12' >> "./config/olm/${PLATFORM}/${VERSION}/metadata/annotations.yaml"
 fi
```
Incredible-O, line 78: Do you think introducing a constant for the OpenShift versions would make sense?

hack/build/bundle.sh
andriisoldatenko: @Incredible-O you can create a story and mark it as tech debt and we can discuss and implement all these ideas in one go.

hack/build/bundle.sh
StefanHauth: Good point. Fixed!

PR #5290, opened 2025-08-25
Update style-guide about secret handling

0sewa0:
## Description

Based on the outcome of [DAQ-2064](https://dt-rnd.atlassian.net/browse/DAQ-2064)

Once we agree, I can create a few follow ups, to implement this.
- (this pattern is not really used by us)

## How can this be tested?

1. Read
2. Agree/disagree

### DIFF
--- a/doc/coding-style-guide.md
+++ b/doc/coding-style-guide.md
@@ -116,6 +116,32 @@ Important characteristics:
   - The cleanup is "invisible" for the caller of the `Reconciler`, so its up to the `Reconciler` to decide if it needs to clean up or setup/update.
     - Example: `DynakubeController` should just call the `ActiveGateReconciler.Reconcile` function, and not worry about if it will create(i.e.: setup) a `StatefulSet` or delete(i.e.: cleanup) the no longer necessary one.
 
+## Secret/ConfigMap handling
+
+A significant role of the Operator is to create/update or just pass along Secrets/ConfigMaps to components it manages.
+
+The main problem that needs to be solved is: If a `Secret/ConfigMap` changes, that warrants the restart of a component that uses it, how do we go about this?
+
+General rules to follow:
+
+- Each `Secret/ConfigMap` we manage should have it's own package, that contain:
+  - The func to determine the name of the `Secret/ConfigMap`. (example: `GetName`)
+  - The "reconcile" logic
+    - This should be also responsible for calculating a hash from the content of the `Secret/ConfigMap`, and storing this value in the `status` of the `DynaKube`
+  - A unique annotation key ( public `const`) (example: `HashAnnotationKey`)
+  - A public `const` for each key used inside the `data` of the `Secret/ConfigMap`.
+- This package is used by other packages that would like to mount the `Secret/ConfigMap`
+  - Using the `GetName`, to reference the `Secret/ConfigMap` in a Volume or Env
+  - Using the "public `const` for each key" to what to part of the `Secret/ConfigMap` to use. (use `Items` instead of `SubPath`)
+  - Using the `HashAnnotationKey` to put the hashed content of the `Secret/ConfigMap` (from the `DynaKube`'s `status`) into the `template.metadata.annotations` of the component, IF it needs to be restarted when the `Secret/ConfigMap`.
+
+Reasoning for these rules:
+
+- As of writing this, we have 15+ `Secrets/ConfigMaps` that we manage and provide to 4+ components. (some components need more, some less)
+  - Putting all of this within the `api` package, when we already tend to have a package for each `Secret/ConfigMap` for "reconciling" is just breaking the "single-place-of-truth" for no good reason.
+  - Fetching and recalculating the hash for each component where a `Secret/ConfigMap` is used would be fine if we only had a few of them, but we have 15+.
+    - Why not use the cache? -> Using the built-in cache of the `Client` had caused problems in the past (the cache was not up to date), which could lead to inconsistency.
+
 ## Errors
 
 ### Do's


doc/coding-style-guide.md
```diff
@@ -116,6 +116,32 @@ Important characteristics:
   - The cleanup is "invisible" for the caller of the `Reconciler`, so its up to the `Reconciler` to decide if it needs to clean up or setup/update.
     - Example: `DynakubeController` should just call the `ActiveGateReconciler.Reconcile` function, and not worry about if it will create(i.e.: setup) a `StatefulSet` or delete(i.e.: cleanup) the no longer necessary one.
 
+## Secret/ConfigMap handling
+
+A significant role of the Operator is to create/update or just pass along Secrets/ConfigMaps to components it manages.
+
+The main problem that needs to be solved is: If a `Secret/ConfigMap` changes, that warrants the restart of a component that uses it, how do we go about this?
+
+General rules to follow:
```
andriisoldatenko, line 126: based on what i read, do you want a skeleton for all secrets/configmaps so we can create "common" reconcile logic?

and dynamically do it, so the client of "this common logic" MUST provide some public methods inside this package?

doc/coding-style-guide.md
```diff
@@ -116,6 +116,32 @@ Important characteristics:
   - The cleanup is "invisible" for the caller of the `Reconciler`, so its up to the `Reconciler` to decide if it needs to clean up or setup/update.
     - Example: `DynakubeController` should just call the `ActiveGateReconciler.Reconcile` function, and not worry about if it will create(i.e.: setup) a `StatefulSet` or delete(i.e.: cleanup) the no longer necessary one.
 
+## Secret/ConfigMap handling
+
+A significant role of the Operator is to create/update or just pass along Secrets/ConfigMaps to components it manages.
+
+The main problem that needs to be solved is: If a `Secret/ConfigMap` changes, that warrants the restart of a component that uses it, how do we go about this?
```
andriisoldatenko, line 124: I don't get how do we solve original problem?

doc/coding-style-guide.md
0sewa0: > want a skeleton for all secrets/configmaps so we can create "common" reconcile logic

our secrets are not consistent enough for this

What I want is to remove such methods:
```
func (ag *Spec) GetTenantSecretName() string {
	return ag.name + TenantSecretSuffix
}
```

And move them into the package that actually manages this secret, with funcs like: (just draft)
```
package tenantoken // this doesnt exits :D 

import (...)

const (
	secretSuffix = "-activegate-tenant-secret"

	HashAnnotationKey = "internal.dynatrace.com/ag-tenant-token-hash"
	TokenFieldKey = "tenant-token"
)

func GetSecretName(dk dynakube.Dynakube) string {
	return dk.Name + secretSuffix
}

func GetHashValue(dk dynakube.Dynakube) string {
	return dk.Status.ActiveGate.TenantTokenHash
}

func AddHashAnnotation(annotations map[string]string, dk dynakube.Dynakube) {
	if annotations == nil {
		annotations := map[string]string{}
	}
	annotations[HashAnnotationKey] = GetHashValue(dk)
}

// ... actual reconcile logic
```


IMO, this would be better, because currently, the things (const/funcs) I mention in the example, is spread all around, in different packages

> I don't get how do we solve original problem?

It "solves it" by organizing it better, currently it is just a mess.
- The problem is not a "complex" problem in its own.
- It gets complex in a codebase, where there are multiple patterns of handling this, and have to use 4 packages to collect all the necessary consts

andriisoldatenko: so i like your examples, so I don't see anything bad to add them to coding style-guide with sub-section `Do`.

doc/coding-style-guide.md
```diff
@@ -116,6 +116,32 @@ Important characteristics:
...
+  - The func to determine the name of the `Secret/ConfigMap`. (example: `GetName`)
+  - The "reconcile" logic
+    - This should be also responsible for calculating a hash from the content of the `Secret/ConfigMap`, and storing this value in the `status` of the `DynaKube`
+  - A unique annotation key ( public `const`) (example: `HashAnnotationKey`)
+  - A public `const` for each key used inside the `data` of the `Secret/ConfigMap`.
+- This package is used by other packages that would like to mount the `Secret/ConfigMap`
+  - Using the `GetName`, to reference the `Secret/ConfigMap` in a Volume or Env
+  - Using the "public `const` for each key" to what to part of the `Secret/ConfigMap` to use. (use `Items` instead of `SubPath`)
+  - Using the `HashAnnotationKey` to put the hashed content of the `Secret/ConfigMap` (from the `DynaKube`'s `status`) into the `template.metadata.annotations` of the component, IF it needs to be restarted when the `Secret/ConfigMap` changes.
+
```
andriisoldatenko, line 153: **For example:**
```
// put your example here
```

albertogdd (approved):

doc/coding-style-guide.md
```diff
@@ -116,6 +116,32 @@ Important characteristics:
   - The cleanup is "invisible" for the caller of the `Reconciler`, so its up to the `Reconciler` to decide if it needs to clean up or setup/update.
     - Example: `DynakubeController` should just call the `ActiveGateReconciler.Reconcile` function, and not worry about if it will create(i.e.: setup) a `StatefulSet` or delete(i.e.: cleanup) the no longer necessary one.
 
+## Secret/ConfigMap handling
```
andriisoldatenko, line 120: after discussing during standup:
Don't forget to add recommended hash algorithms.

andriisoldatenko (dismissed):

PR #5308, opened 2025-08-27
[release-1.7] Fix wrong init-container runAsNonRoot value on OCP

albertogdd:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-13021

init-container `runAsNonRoot` was set to false on OCP even if a different user and group was set on the Pod SecurityContext.

- Deploy DynaKube with init-container injection (CNFS)
- Apply a Pod deployment containing SecurityContext as, e.g: runAsUser: 1001, runAsGroup: 1001
- Check that init-container has field `runAsNonRoot: true`

Here is the Deployment manifest I used to replicate the issue
```yaml
# debug.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    app: debug
  annotations:
    metadata.dynatrace.com/alberto.test: "true"
  name: debug
  namespace: app
spec:
  replicas: 1
  selector:
    matchLabels:
      app: debug
  strategy: {}
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: debug
      annotations:
        metadata.dynatrace.com/alberto.test: "true"
    spec:
      containers:
      - args:
        - |-
          apk add curl jq openssl bash bind-tools
          sleep infinity
        command:
        - /bin/sh
        - -c
        image: alpine:3.18
        name: alpine
        securityContext:
          capabilities:
            drop:
            - ALL
          readOnlyRootFilesystem: false
          runAsUser: 9031
        resources:
          limits:
            memory: 128Mi
            cpu: 100m
```


pkg/webhook/mutation/pod/mutator/oneagent/init.go
```diff
@@ -95,7 +95,17 @@ func HasPodGroupSet(ctx *corev1.PodSecurityContext) bool {
 }
 
 func IsNonRoot(ctx *corev1.SecurityContext) bool {
```
andriisoldatenko: is any chance we can rename it? I know it was before, but very confusing name, I prefer to `secCtx` or `sc`

albertogdd: I agree, on it

albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/5308/commits/a42afeff7fafe01c0d5e37f7fdba4381cbe82b31)

pkg/webhook/mutation/pod/mutator/oneagent/init.go
```diff
@@ -86,16 +86,26 @@ func getTechnology(pod corev1.Pod, dk dynakube.DynaKube) string {
...
-	return ctx != nil && ctx.RunAsGroup != nil
+func HasPodGroupSet(psc *corev1.PodSecurityContext) bool {
+	return psc != nil && psc.RunAsGroup != nil
 }
 
-func IsNonRoot(ctx *corev1.SecurityContext) bool {
-	return ctx != nil &&
-		(ctx.RunAsUser != nil && *ctx.RunAsUser != RootUserGroup) &&
-		(ctx.RunAsGroup != nil && *ctx.RunAsGroup != RootUserGroup)
+func IsNonRoot(sc *corev1.SecurityContext) bool {
```
andriisoldatenko, line 97: ❤️

pkg/webhook/mutation/pod/mutator/oneagent/init_test.go
```diff
@@ -303,3 +304,51 @@ func TestGetTechnology(t *testing.T) {
 		})
 	}
 }
+
+func Test_isNonRoot(t *testing.T) {
+	t.Run("root user", func(t *testing.T) {
```
andriisoldatenko, line 309: very good example to use data driven test, since, it's just one function and lot's of inputs

andriisoldatenko:
```
tests := []struct {
		name   string
		ctx    *corev1.SecurityContext
		expect bool
	}{
		{
			name: "root user",
			ctx: &corev1.SecurityContext{
				RunAsUser:  ptr.To(int64(0)),
				RunAsGroup: ptr.To(int64(0)),
			},
			expect: false,
		},
		{
			name: "non-root user",
			ctx: &corev1.SecurityContext{
				RunAsUser:  ptr.To(int64(1000)),
				RunAsGroup: ptr.To(int64(1000)),
			},
			expect: true,
		},
		{
			name: "root user and nil group (OCP case)",
			ctx: &corev1.SecurityContext{
				RunAsUser:  ptr.To(int64(0)),
				RunAsGroup: nil,
			},
			expect: false,
		},
		{
			name: "non-root user and nil group (OCP case)",
			ctx: &corev1.SecurityContext{
				RunAsUser:  ptr.To(int64(1000)),
				RunAsGroup: nil,
			},
			expect: true,
		},
		{
			name: "nil user and root group (edge case)",
			ctx: &corev1.SecurityContext{
				RunAsUser:  nil,
				RunAsGroup: ptr.To(int64(0)),
			},
			expect: false,
		},
		{
			name: "nil user and non-root group (edge case)",
			ctx: &corev1.SecurityContext{
				RunAsUser:  nil,
				RunAsGroup: ptr.To(int64(1000)),
			},
			expect: true,
		},
		{
			name:   "nil context",
			ctx:    nil,
			expect: true,
		},
	}

	for _, tt := range tests {
		tt := tt
		t.Run(tt.name, func(t *testing.T) {
			result := IsNonRoot(tt.ctx)
			assert.Equal(t, tt.expect, result)
		})
	}
```

andriisoldatenko: but i'm okey to leave as is :)

albertogdd: True, I prefer to leave it as it is because you can trigger them one by one on VSCode :smiley_cat:

PR #5318, opened 2025-08-29
Fix confusing error message when OA connection info is not available and AG not configured

0sewa0:
## Description
[DAQ-13113](https://dt-rnd.atlassian.net/browse/DAQ-13113)

The problem:
1. The tenant doesn't give connection info for the OneAgent (for whatever reason)
  - In the OA reconciler we return early, and say "no connection info, no daemonset"
  - The dtpullsecret reconcile would happen after this, but wont due to the early return
2. AG is not configured, the AG reconciler will run, just to make sure "everything is cleaned up"
   - But BOOM, because it uses the same dtpullsecret reconcile
   - Due to No AG configured-> the AG connection-info reconciler will only clean up, and not get the tenantUUID
dtpullsecret reconcile would need the tenantUUID  and because it is a "shared" reconciler it will run, because the OA is configured

## How can this be tested?

Restricted networkzone + OA only deployment -> No pullsecret error

(most of the changes are test refactors)

[DAQ-13113]: https://dt-rnd.atlassian.net/browse/DAQ-13113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controllers/dynakube/logmonitoring/configsecret/reconciler.go
+++ b/pkg/controllers/dynakube/logmonitoring/configsecret/reconciler.go
@@ -3,8 +3,10 @@ package configsecret
 import (
 	"context"
 	"fmt"
+	"maps"
 	"strings"
 
+	"github.com/Dynatrace/dynatrace-operator/pkg/api"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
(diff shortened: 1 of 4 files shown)


aorcholski (approved):
LGTM no error in the log, OA daemonset is not created (no AG with NZ)

PR #5330, opened 2025-09-01
Pass the proxy/no-proxy to the standalone logmon

0sewa0:
## Description

[DAQ-13144](https://dt-rnd.atlassian.net/browse/DAQ-13144)

What is set in `spec.proxy` and `feature.dynatrace.com/no-proxy` should be provided to the standalone logmon.
Also the ServiceName+ServiceIPs of the AG should also be part of the `noProxy`, similar the way we do it for hostmon.

Added a new status field: `status.proxyURLHash`
- This is set in the proxy reconciler, so that I do not have to calc this more often than I have to.

## How can this be tested?

- `Proxy` and `noProxy` are set in the deployment conf of the standalone logmon
- The internal annotations are present on the daemonset

In the UI: (this is the `Logs` app)
image

[DAQ-13144]: https://dt-rnd.atlassian.net/browse/DAQ-13144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/util/hasher/hasher.go
+++ b/pkg/util/hasher/hasher.go
@@ -17,7 +17,7 @@ const AnnotationHash = api.InternalFlagPrefix + "template-hash"
 
 // GenerateHash creates a hash from the provided input.
 // This hash is meant to be used for simplifying detecting differences between 2 objects.
-// Uses uses FNV-1 hashing, should be used for hashing not sensitive data.
+// Uses FNV-1 hashing, should be used for hashing not sensitive data.
 func GenerateHash(input any) (string, error) {
 	data, err := json.Marshal(input)
 	if err != nil {
@@ -36,7 +36,7 @@ func GenerateHash(input any) (string, error) {
 
 // GenerateSecureHash creates a hash from the provided input.
 // This hash is meant to be used for simplifying detecting differences between 2 values.
-// Uses uses SHA256 hashing, can be used for hashing sensitive data.
+// Uses SHA256 hashing, can be used for hashing sensitive data.
 func GenerateSecureHash(input any) (string, error) {
 	data, err := json.Marshal(input)
 	if err != nil {


pkg/controllers/dynakube/logmonitoring/configsecret/reconciler.go
```diff
@@ -134,6 +139,22 @@ func (r *Reconciler) getSecretData(ctx context.Context) (map[string][]byte, erro
 		hostIDSourceKey: "k8s-node-name",
 	}
 
+	if r.dk.HasProxy() {
+		proxyURL, err := r.dk.Proxy(ctx, r.apiReader)
+		if err != nil {
+			conditions.SetSecretGenFailed(r.dk.Conditions(), LmcConditionType, err)
```
andriisoldatenko: secretGenFailed? why?

andriisoldatenko: apiErro instead?

0sewa0: fixed

andriisoldatenko (approved):

0sewa0:
@aorcholski 
> is not removed from daemonset.template.annotations when proxy is removed trom the dynakube

did you wait till it is reconciled? (change in annotation will not cause a reconcile)

andriisoldatenko:
tested and it works,

```
2025-09-02 08:59:59.006 UTC [00000001] info    [native] Proxy ....................... http://squid.proxy.svc.cluster.local:3128
[2025-09-02 09:01:00.255 UTC] [/logmonitoring.cpp] [debug] Loading /var/log/pods/kube-system_kube-proxy-gke-andrii-test-default-
pool-c1873b05-5rmg_1abff3a1db234a6df534e3edaa1e732b/kube-proxy/#.log from persistence
```

pkg/util/hasher/hasher.go
```diff
@@ -29,6 +34,21 @@ func GenerateHash(ds any) (string, error) {
 	return strconv.FormatUint(uint64(hasher.Sum32()), 10), nil
 }
 
+// GenerateSecureHash creates a hash from the provided input.
+// This hash is meant to be used for simplifying detecting differences between 2 values.
+// Uses uses SHA256 hashing, can be used for hashing sensitive data.
```
avorima: typo/wording: Double "uses" on both functions.

PR #5335, opened 2025-09-02
Handle dangling symlinks in the host publisher

0sewa0:
## Description

[DAQ-13275](https://dt-rnd.atlassian.net/browse/DAQ-13275)

Sometimes the symlink to the old host agent is not cleaned up properly or in time, so the host publisher wants to publisher something that doesn't exits.

So added a small check for this in the publisher to make sure that it doesn't happen.

Sidenote: not using `afero` here because the `afero.Fs` interface doesn't have `Lstat`

## How can this be tested?

Added unit tests for it.

The manual way:

Create the faulty symlink:
```bash
rm /data/_dynakubes/<dk-name>/osagent
mkdir /data/<tenantUUID>/osagent
ln -s /data/<tenantUUID>/osagent /data/_dynakubes/<dk-name>/osagent
rmdir /data/<tenantUUID>/osagent/
```

Create dk that has host agents and uses the CSI driver

[DAQ-13275]: https://dt-rnd.atlassian.net/browse/DAQ-13275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controllers/csi/driver/volumes/host/publisher_test.go
```diff
@@ -78,3 +114,52 @@ func getTestVolumeConfig(t *testing.T) csivolumes.VolumeConfig {
 		RetryTimeout: time.Microsecond, // doesn't matter
 	}
 }
+
+func Test_cleanupDanglingSymlink(t *testing.T) {
+	t.Run("removes dangling symlink", func(t *testing.T) {
+		base := t.TempDir()
+
+		// Create dir to be symlinked
+		missingDir := path.Join(base, "dir")
```
avorima: actual filesystem path operations should always use the filepath package

andriisoldatenko: it's a valid point:
>The path package should only be used for paths separated by forward slashes, such as the paths in URLs. This package does not deal with Windows paths with drive letters or backslashes; to manipulate operating system paths, use the [path/filepath](https://pkg.go.dev/path/filepath) package.

andriisoldatenko: i'm affraid @avorima  we need to replace it everywhere, wdyt to create and PR it independently?

0sewa0: most of the place we use `filepath` (just maybe not in tests)

because we use the `PathResolver` that uses the `filepath`

0sewa0: but yes, pls replace it everywhere

0sewa0: I will replace them in this

andriisoldatenko: kk

andriisoldatenko: @avorima could you please do me a favor? Also PR coding-style guide...

0sewa0: > I will replace them in this

ie.: in the packages I already touched not truly everywhere, do not trust my commit message

avorima: ok, i'll create a follow-up

avorima (approved):
code LGTM

PR #5343, opened 2025-09-03
Improve RunAsNonRoot check

0sewa0:
## Description

[DAQ-13355](https://dt-rnd.atlassian.net/browse/DAQ-13355)

There was a bug found, when somehow setting the `podSecurityContext`'s user to `0` causes a problem, that we still think it is "non root", the unit tests make it look that everything is fine.

- Added a lot more unittests, on different levels
- Fixed the `IsNonRoot` check, to check what is actually kubernetes will check
   - If the `RunAsUser` is `0`, it is `RunAsNonRoot: false`
   - If the `RunAsUser` is anything but `0`, it will be `RunAsNonRoot: true`
     -   `RunAsGroup` has not effect, user takes precedence
   - Only `RunAsGroup` is set, and it is `0`,  it will be `RunAsNonRoot: true`

## How can this be tested?

Inject into a Pod with a PodSecurityContext like:
```
      securityContext:
        runAsUser: 0
```

[DAQ-13355]: https://dt-rnd.atlassian.net/browse/DAQ-13355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/webhook/mutation/pod/config.go
+++ b/pkg/webhook/mutation/pod/config.go
@@ -15,4 +15,7 @@ const (
 
 	NoBootstrapperConfigReason = "NoBootstrapperConfig"
 	NoMutationNeededReason     = "NoMutationNeeded"
+
+	RootUser  int64 = 0
+	RootGroup int64 = 0
 )
--- a/pkg/webhook/mutation/pod/init.go
+++ b/pkg/webhook/mutation/pod/init.go
@@ -85,19 +85,31 @@ func securityContextForInitContainer(pod *corev1.Pod, dk dynakube.DynaKube, isOp
 
 	return combineSecurityContexts(initSecurityCtx, *pod)
 }
-
 func combineSecurityContexts(baseSecurityCtx corev1.SecurityContext, pod corev1.Pod) *corev1.SecurityContext {
+	containerSecurityCtx := &corev1.SecurityContext{}
+	if len(pod.Spec.Containers) > 0 {
+		containerSecurityCtx = pod.Spec.Containers[0].SecurityContext
+	}
+
 	podSecurityCtx := pod.Spec.SecurityContext
 
-	if oacommon.HasPodUserSet(podSecurityCtx) {
+	if hasPodUserSet(podSecurityCtx) {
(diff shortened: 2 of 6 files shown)


albertogdd:
Still I think we should merge this PR because its nice changes :+1:

andriisoldatenko:
> this can (maybe fix it)
> 
> ```diff
> diff --git a/pkg/webhook/mutation/pod/init.go b/pkg/webhook/mutation/pod/init.go
> index e5e6d6043..16466e7b9 100644
> --- a/pkg/webhook/mutation/pod/init.go
> +++ b/pkg/webhook/mutation/pod/init.go
> @@ -79,6 +79,9 @@ func securityContextForInitContainer(pod *corev1.Pod, dk dynakube.DynaKube, isOp
>  
>  	if !isOpenShift {
>  		initSecurityCtx.RunAsUser = ptr.To(oacommon.DefaultUser)
> +	} else if len(pod.Spec.Containers) > 0 && pod.Spec.Containers[0].SecurityContext != nil {
> +		// in Openshift, the runAsUser can be set via SCCs, and you can set it to root
> +		initSecurityCtx.RunAsUser = pod.Spec.Containers[0].SecurityContext.RunAsUser
>  	}
>  
>  	addSeccompProfile(&initSecurityCtx, dk)
> ```

TBH i prefer to may be log/warn it if we want to go this direction, otherwise we need document this limitation.

0sewa0:
I just reintroduced the [old logic](https://github.com/Dynatrace/dynatrace-operator/blob/8a1a3908465f5b0c35de4d779d23cf872650f551/pkg/webhook/mutation/pod/v1/init.go#L85) for this, from "v1"  

I tested that, in ocp, with my malicious SCC.
- I guess ocp uses a webhook, but only does it on the container level, so we need this to get changes done by webhooks

albertogdd:
I guess we merge this but leave the bug open for resolution for 1.8

andriisoldatenko (approved):
code LGTM.

tested again using example from bug and it works.

PR #5355, opened 2025-09-05
Handle dangling symlinks in the host publisher (#5335)

0sewa0:
## Description

[DAQ-13275](https://dt-rnd.atlassian.net/browse/DAQ-13275)

cherry pick of #5335

## How can this be tested?

same as #5335

[DAQ-13275]: https://dt-rnd.atlassian.net/browse/DAQ-13275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controllers/csi/driver/volumes/host/publisher.go
+++ b/pkg/controllers/csi/driver/volumes/host/publisher.go
@@ -54,9 +54,14 @@ func (pub *Publisher) PublishVolume(ctx context.Context, volumeCfg *csivolumes.V
 func (pub *Publisher) mountStorageVolume(volumeCfg *csivolumes.VolumeConfig) error {
 	oaStorageDir := pub.path.OsAgentDir(volumeCfg.DynakubeName)
 
-	cleanupDanglingSymlink(oaStorageDir)
+	err := cleanupDanglingSymlink(oaStorageDir)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controllers/csi/driver/volumes/host/publisher.go
```diff
@@ -75,3 +77,14 @@ func (pub *Publisher) mountStorageVolume(volumeCfg *csivolumes.VolumeConfig) err
...
 	return nil
 }
+
+func cleanupDanglingSymlink(hostDir string) {
+	linkInfo, err := os.Lstat(hostDir)
+	if err == nil && linkInfo.Mode()&os.ModeSymlink != 0 {
+		_, err := os.Stat(hostDir)
+		if os.IsNotExist(err) {
+			log.Debug("found dangling symlink", "path", hostDir)
+			os.Remove(hostDir)
```
gkrenn: You should return any error here. If the cleanup did not work the mount will fail again.

andriisoldatenko: Or at least log it? If we don't want to silently hide it.

0sewa0: fixed

andriisoldatenko (approved):

PR #5358, opened 2025-09-05
Add RBAC config for the database extensions

gkrenn:
## Description

This PR fixes the following: 
- The EEC requires additional permissions to look for pods. 
- A default ServiceAccount for the Java DB extensions executor is required.

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-11698)

## How can this be tested?

- `make deploy`
  - `dynatrace-database-extensions` Role and Rolebinding exist
  - `dynatrace-database-extensions-executor` ServiceAccount exists

### DIFF
--- a/config/helm/chart/default/templates/Common/opentelemetry-collector/serviceaccount-opentelemetry-collector.yaml
+++ b/config/helm/chart/default/templates/Common/opentelemetry-collector/serviceaccount-opentelemetry-collector.yaml
@@ -1,4 +1,4 @@
-{{- if or (.Values.rbac.extensions.create) (.Values.rbac.telemetryIngest.create) }}
+{{- if or (and .Values.rbac.extensions.create .Values.rbac.extensions.prometheus.create) (.Values.rbac.telemetryIngest.create) }}
 # Copyright 2021 Dynatrace LLC
 
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/config/helm/chart/default/tests/Common/opentelemetry-collector/serviceaccount-opentelemetry-collector_test.yaml
+++ b/config/helm/chart/default/tests/Common/opentelemetry-collector/serviceaccount-opentelemetry-collector_test.yaml
@@ -16,10 +16,36 @@ tests:
           value: NAMESPACE
       - isNotEmpty:
           path: metadata.labels
+  
+  - it: should exist with prometheus extension
+    set:
+      rbac.extensions.create: true
+      rbac.extensions.prometheus.create: true
+      rbac.telemetryIngest.create: false
+    asserts:
+      - isKind:
+          of: ServiceAccount
+      - equal:
+          path: metadata.name
+          value: dynatrace-opentelemetry-collector
+      - equal:
+          path: metadata.namespace
+          value: NAMESPACE
+      - isNotEmpty:
+          path: metadata.labels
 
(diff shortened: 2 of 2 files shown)


config/helm/chart/default/values.yaml
```diff
@@ -238,6 +238,10 @@ rbac:
   extensions:
     create: true
     annotations: {}
+    prometheus: 
+      create: true
```
0sewa0, line 242: question:

what is the difference between:

`extensions.create` and `extensions.prometheus.create` ?

AFAIK the "old" `extensions.create` is basically the "new" `extensions.prometheus.create`, no?

or we keeping `extensions.create` so we don't break upgrading users? is this necessary (extensions was not GA AFAIK)?

andriisoldatenko: @0sewa0 what you're looking for is in scope of my story https://dt-rnd.atlassian.net/browse/DAQ-11700
AFAIK we just want to rename

> The existing Helm switch `.rbac.extensions.create` shall be renamed to `.rbac.extensions.prometheus.create`.

0sewa0: then this PR is mixing the 2 tickets together,

because the ticket mentioned in the description says nothing about `rbac.extensions.prometheus.create`
it is only about adding `rbac.extensions.database.create`, and nothing more

andriisoldatenko: agree, wdyt @StefanHauth ?

0sewa0: we can do the "helm part" of the rename/move in this PR, just lets not have this "we introduce the new field, but remove the old one 1 day later"

StefanHauth: `extensions.create` will become `extensions.prometheus.create`, and we shall not keep `extensions.create`, the feature was never GA and it will only cause confusion.
The rename/move was intended to be part of the other story, so I would stall this PR until Prometheus extensions found their final place in CRD and Helm values.

StefanHauth (changes requested):
Stall this PR until prometheus stuff has been moved out of the way

config/helm/chart/default/values.yaml
gkrenn: Left the `extensions.create` in because of the extension controller role and service account. Otherwise we would need to check programmatically if any of the extensions is enabled? @StefanHauth 

Will undo the prometheus move as it is not part of this ticket, then the PR does not need to be blocked as it is only introducing a new database field

gkrenn: So the `extensions.create` would have been a general switch for extension, not related to the old prometheus functionality anymore

gkrenn: As an afterthough, maybe it would be better as Marcell suggested to do all the helm changes in this PR? Is that fine for you? @andriisoldatenko

0sewa0: I would be for it :P 

The CRD changes to `prometheus` doesn't have to be done at the same/before/after time as the helm changes.

They are not related from code perspective.

gkrenn: As discussed with @StefanHauth, it makes sense to keep the `extensions.create` flag as general opt-out flag

PR #5380, opened 2025-09-08
Expose additional pod mutator errors as annotation

avorima:
## Description

Add `oneagent.dynatrace.com/reason` and `metadata.dynatrace.com/reason` annotation that will be populated if the
respective pod mutators return errors.
Added integration tests to verify that the annotation handling works as expected.

Also did a few fly-by's to simplify/modernize code.

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-1715)

## How can this be tested?

Run webhook integration tests. I could not reproduce the error cases manually.

```
make go/test
```


pkg/webhook/mutation/pod/webhook_integration_test.go
```diff
@@ -0,0 +1,280 @@
...
+			}
+		})
+
+		assert.Contains(t, pod.Annotations, metadatamutator.AnnotationReason)
+	})
+}
+
+func createPodFromTemplate(t *testing.T, clt client.Client, mutateFn func(*corev1.Pod)) *corev1.Pod {
+	t.Helper()
+	podTemplatePath := filepath.Join(project.TestDataDir(), "sample-app/pod-base.yaml")
```
0sewa0: in certain ways I like that we use the same pod sample for integration and e2e tests.

but feels a bit icky as well, but that is probably because the `test` folder is a bit of a mess.

- there is no distinction between what is "e2e" specific  (like the `test/helpers`/ and `test/features` and `test/scenarios`) and what is more general

we should move things around in the `test` folder. WDYT? (in a separate PR/ticket 😅 )

pkg/webhook/mutation/pod/webhook_integration_test.go
```diff
@@ -0,0 +1,280 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+)
+
+const testNamespace = "dynatrace"
+
+func TestWebhook(t *testing.T) {
+	clt := integrationtests.SetupWebhookTestEnvironment(t,
+		envtest.WebhookInstallOptions{
+			// TODO(avorima): Load this from a file using Paths
```
0sewa0, line 38: interesting idea, no idea how you could achieve it nicely, due to the mutation webhook configuration is part of the helm manifests. 🤔

pkg/webhook/mutation/pod/webhook_integration_test.go
```diff
@@ -0,0 +1,280 @@
...
+							TenantUUID: uuid.NewString(),
+						},
+					},
+				},
+			},
+		}
+		createDynaKube(t, clt, dk)
+
+		pod := createPodFromTemplate(t, clt, func(pod *corev1.Pod) {
+			pod.Annotations[oneagentmutator.AnnotationInject] = "true" // metadata needs this as well
```
0sewa0: > // metadata needs this as well

huh, it shouldn't

also why do you even need these annotations?
- is it some integration tests weirdness?

I don't see you creating a Namespace, maybe that is the reason?

pkg/webhook/mutation/pod/webhook_integration_test.go
avorima: somewhat of a development leftover. i'll create an extra testadata dir with a pod spec local to the test. i'm not against reworking `test`, but i think it needs a bit more planning

pkg/webhook/mutation/pod/webhook_integration_test.go
avorima: usually webhook files exist in the config directory. it's one of the many things where the repo differs a lot from the kubebuilder starting point and it might be worth looking into aligning things a bit.

0sewa0: the problem there is that kubebuilder really wants to use kustomize, which we really don't want to :D

pkg/webhook/mutation/pod/webhook_integration_test.go
avorima: you're right i misinterpreted some errors while testing. removed oneagent completely from the metadata test case

pkg/webhook/mutation/pod/webhook_integration_test.go
avorima: you can make them coexist. we can talk about it later

pkg/util/integrationtests/environment.go
```diff
@@ -57,6 +67,101 @@ func SetupTestEnvironment(t *testing.T) client.Client {
 	return clt
 }
 
+func SetupWebhookTestEnvironment(t *testing.T, webhookOptions envtest.WebhookInstallOptions, webhookSetup func(ctrl.Manager) error) client.Client {
```
andriisoldatenko, line 56: can we somehow reuse 
```
func SetupTestEnvironment(t *testing.T) client.Client {
	// specify test environment configuration
	testEnv = &envtest.Environment{
		Scheme:                   scheme.Scheme,
		CRDDirectoryPaths:        []string{filepath.Join(projectpath.Root, "config", "crd", "bases")},
		ErrorIfCRDPathMissing:    true,
		AttachControlPlaneOutput: true,
	}
```
and add webhook setup below?

pkg/util/integrationtests/environment.go
```diff
@@ -57,6 +67,101 @@ func SetupTestEnvironment(t *testing.T) client.Client {
...
+		ErrorIfCRDPathMissing: true,
+		WebhookInstallOptions: webhookOptions,
+	}
+
+	// Retrieve the first found binary directory to allow running tests from IDEs
+	if getFirstFoundEnvTestBinaryDir() != "" {
+		testEnv.BinaryAssetsDirectory = getFirstFoundEnvTestBinaryDir()
+	}
+
+	// TODO: refactor and reuse from e2e tests or another place
```
andriisoldatenko: could youo please write a ticket for this TODO?

pkg/webhook/mutation/pod/mutator/errors.go
```diff
@@ -0,0 +1,23 @@
+package mutator
+
+import corev1 "k8s.io/api/core/v1"
+
+//nolint:revive // The mutator package is always aliased so this name doesn't stutter.
```
andriisoldatenko, line 5: i prefer to split comments, nolint and next or before your comment

pkg/webhook/mutation/pod/testdata/pod.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
andriisoldatenko: usually we construc test data using go structs, why we want here to use yaml?

pkg/webhook/mutation/pod/webhook.go
```diff
@@ -79,9 +80,15 @@ func (wh *webhook) Handle(ctx context.Context, request admission.Request) admiss
 
 	wh.recorder.Setup(mutationRequest)
 
-	err = wh.handle(mutationRequest)
-	if err != nil {
-		return silentErrorResponse(mutationRequest.Pod, err)
+	originalPod := mutationRequest.Pod.DeepCopy()
+	if err := wh.handle(mutationRequest); err != nil {
+		mutErr := new(dtwebhook.MutatorError)
```
andriisoldatenko, line 85: out of curiosity why not: `mutErr := &dtwebhook.MutatorError{}`?

pkg/webhook/mutation/pod/webhook_integration_test.go
andriisoldatenko: Please create reasearch about this TODO

pkg/webhook/mutation/pod/webhook_integration_test.go
```diff
@@ -0,0 +1,252 @@
...
+			pod := &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "dynatace-webhook",
+					Namespace: testNamespace,
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  dtwebhook.WebhookContainerName,
+							Image: "webhook:1.0.0",
```
andriisoldatenko: to not confuse i prefer to rename to dummy-webhook-img:1.0.0

pkg/util/integrationtests/environment.go
avorima: it's your TODO so i'm not 100% sure what you had in mind https://dt-rnd.atlassian.net/browse/DAQ-13987

avorima: it's your TODO so i'm not 100% sure what you had in mind https://dt-rnd.atlassian.net/browse/DAQ-13987

pkg/webhook/mutation/pod/mutator/errors.go
avorima: this comment is specifically about the nolint directive. putting it anywhere else would lead to the godoc of the struct mentioning it which is wrong. it's actually the recommended style for the linter https://github.com/ashanbrown/nolintlint

pkg/webhook/mutation/pod/testdata/pod.yaml
avorima: it's less code when re-using the YAML and also less code in general

pkg/webhook/mutation/pod/webhook.go
avorima: new is more versatile for zero initialization, because it can be used in an initializer statement. so i prefer to use it for errors for easier refactoring down the line

pkg/util/integrationtests/environment.go
andriisoldatenko: 🙈 sure, just remove line or add jira to comment

pkg/webhook/mutation/pod/mutator/errors.go
andriisoldatenko: pls check this example https://github.com/dynatrace/dynatrace-operator/blob/09e69c073126bb951e2be551f65563a2dab65b70/pkg/api/v1beta2/dynakube/dynakube_status.go#L16

pkg/webhook/mutation/pod/testdata/pod.yaml
andriisoldatenko: less gocode and more yaml code ;) we need to decide how we want to proceed, cause usally we prefer go.
I'm not against testdata, and i see where it can be useful, especially if you want to aplly it using kubectl. But i asked because it;s something new and next time we need to know how to do it.

pkg/webhook/mutation/pod/mutator/errors.go
avorima: yes, this style is not enforced but i think it's good practice. i wouldn't know from looking at this nolint directive why the author decided to add it

andriisoldatenko: i did it :) by trying to decopule godocs and nolint lines ;)

andriisoldatenko: IMO it's easy to read rahter than // nolint // comment

pkg/webhook/mutation/pod/testdata/pod.yaml
avorima: most pod fields aren't actually required to pass the test, so i switched to go code

pkg/webhook/mutation/pod/mutator/errors.go
avorima: but the // comment after //nolint is not godoc

andriisoldatenko: Sorry, I misunderstood you, is it a comment to explain nolint?

avorima: yes, exactly

pkg/webhook/mutation/pod/webhook_integration_test.go
avorima: https://dt-rnd.atlassian.net/browse/DAQ-14005

PR #5392, opened 2025-09-09
Fix failing test/e2e/release test (failing if launched locally).

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-13598)

## Description

PR adds functions to install dynakube and/or apiToken without settings scopes.

All tests use normal apiToken which has `settings` scopes granted.

The `logmonitoring.WithOptionalScopes` test first install dynakube and ApiToken without optional scopes. Next the token secret is updated and following steps are performed.

`ApiTokenNoSettings` token shouldn't have the following scopes:
- settings.read
- settings.write
- entities.read

We can remove `entities.read` from the official `ApiTokenNoSettings` token after this fix is back-ported to all branches that are still tested and use this [logic](https://github.com/Dynatrace/dynatrace-operator/blob/v1.7.0/test/helpers/tenant/secrets.go#L113).

## How can this be tested?
Create testdata/secrets/single-tenant.yaml
```
tenantUid: <tenant>
apiUrl: https:// <tenant>.dev.dynatracelabs.com/api
apiToken: dt...
apiTokenNoSettings: dt...
dataIngestToken: dt...
```
Launch tests
```
make test/e2e
```

I checked :
```
make test/e2e/no-csi
make test/e2e/release
make test/e2e/standard
```

### DIFF
--- a/test/helpers/tenant/secrets.go
+++ b/test/helpers/tenant/secrets.go
@@ -107,13 +107,8 @@ func GetEdgeConnectTenantSecret(t *testing.T) EdgeConnectSecret {
 	return result
 }
 
-func CreateTenantSecret(secretConfig Secret, name, namespace string) features.Func {
+func CreateTenantSecret(apiToken string, dataIngestToken string, name, namespace string) features.Func {
 	return func(ctx context.Context, t *testing.T, envConfig *envconf.Config) context.Context {
-		apiToken := secretConfig.APIToken
-		if secretConfig.APITokenNoSettings != "" {
-			apiToken = secretConfig.APITokenNoSettings
-		}
-
 		defaultSecret := corev1.Secret{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      name,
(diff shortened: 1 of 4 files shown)


test/helpers/tenant/secrets.go
```diff
@@ -107,13 +107,8 @@ func GetEdgeConnectTenantSecret(t *testing.T) EdgeConnectSecret {
 	return result
 }
 
-func CreateTenantSecret(secretConfig Secret, name, namespace string) features.Func {
+func CreateTenantSecret(apiToken string, dataIngestToken string, name, namespace string) features.Func {
```
avorima: can be simplified to
```suggestion
func CreateTenantSecret(apiToken, dataIngestToken, name, namespace string) features.Func {
```
this applies to all places where you added the token parameters

avorima (commented):
thank you fixing this! i never run these make targets, because some tests don't work in my env. got to get that fixed

0sewa0 (approved):

PR #5397, opened 2025-09-10
WIP: otlp env var injection

bacherfl:
## Description

This PR contains a initial implementation of the injection of OTLP exporter related env vars for the automatic configuration of OpenTelemetry SDKs. This PR should serve as a discussion base at this stage - tests and documentation will be added when we have agreed on an approach

## How can this be tested?

TODO

### DIFF
--- a/pkg/controllers/dynakube/otelc/endpoint/reconciler.go
+++ b/pkg/controllers/dynakube/otelc/endpoint/reconciler.go
@@ -2,10 +2,9 @@ package endpoint
 
 import (
 	"context"
-	"fmt"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/otlp"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
-	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/activegate/capability"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/otelc/consts"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/conditions"
 	k8sconfigmap "github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/configmap"
@@ -99,17 +98,9 @@ func (r *Reconciler) reconcileConfigMap(ctx context.Context) error {
 func (r *Reconciler) generateData() (map[string]string, error) {
 	data := make(map[string]string)
 
-	dtEndpoint := r.dk.APIURL() + "/v2/otlp"
-
-	if r.dk.ActiveGate().IsEnabled() {
-		tenantUUID, err := r.dk.TenantUUID()
-		if err != nil {
-			return data, err
-		}
-
-		serviceFQDN := capability.BuildServiceName(r.dk.Name) + "." + r.dk.Namespace + ".svc"
-
-		dtEndpoint = fmt.Sprintf("https://%s/e/%s/api/v2/otlp", serviceFQDN, tenantUUID)
+	dtEndpoint, err := otlp.GetOtlpIngestEndpoint(r.dk)
+	if err != nil {
+		return data, err
 	}
 
 	data["DT_ENDPOINT"] = dtEndpoint
--- a/pkg/injection/namespace/bootstrapperconfig/bootstrapperconfig.go
+++ b/pkg/injection/namespace/bootstrapperconfig/bootstrapperconfig.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"strconv"
+	"strings"
 
 	"github.com/Dynatrace/dynatrace-bootstrapper/pkg/configure/enrichment/endpoint"
 	"github.com/Dynatrace/dynatrace-bootstrapper/pkg/configure/oneagent/ca"
@@ -70,6 +71,21 @@ func (s *SecretGenerator) GenerateForDynakube(ctx context.Context, dk *dynakube.
 		if err != nil {
 			return errors.WithStack(err)
 		}
+
+		// TODO check if this can be refactored to be cleaner
+		// here we extract the token again from the 'endpoint.properties', which includes both the DT_METRICS_INGEST_URL and DT_METRICS_INGEST_API_TOKEN in a single string,
+		// while for the OTLP env vars we need the ingest token as a separate key within the secret.
+		// This way we do not have to look up the secret containing the token in the dynatrace namespace again (i.e. less kubernetes API calls),
+		// but this is also not ideal, as the extractApiToken relies on the endpoint.properties to be in a certain format.
+		// Ideally, we can refactor this to retrieve the endpoint and token at the beginning and then generate the content for the secrets,
+		// as they are required by the bootstrapper and the otlp env vars, respectively.
+		apiToken := extractApiToken(data)
+		if len(apiToken) > 0 {
+			err = s.createSecretForNSlist(ctx, consts.OtlpIngestTokenSecretName, ConfigConditionType, nsList, dk, extractApiToken(data))
+		}
+		if err != nil {
+			return errors.WithStack(err)
+		}
 	}
 
 	certs, err := s.generateCerts(ctx, dk)
@@ -88,9 +104,35 @@ func (s *SecretGenerator) GenerateForDynakube(ctx context.Context, dk *dynakube.
 		if err != nil {
 			return errors.WithStack(err)
 		}
+
+		// create cert secret for otlp configuration
+		// TODO is the creation of the OTLP related secrets tied to the same preconditions as the creation of the bootstrap secrets?
+		// TODO should we decouple the OTLP secret creation to be independent of any errors during the creation of the other secrets?
+		err = s.createSecretForNSlist(ctx, consts.OtlpIngestCertsSecretName, CertsConditionType, nsList, dk, certs)
+		if err != nil {
+			return errors.WithStack(err)
+		}
 	}
 
 	return nil
+}
+
+func extractApiToken(data map[string][]byte) map[string][]byte {
+	result := map[string][]byte{}
+	endpointInfo, ok := data[endpoint.InputFileName]
+	if !ok {
+		return result
+	}
+	split := strings.Split(string(endpointInfo), "\n")
+	for _, line := range split {
+		if strings.HasPrefix(line, MetricsTokenSecretField) {
+			lineSplit := strings.Split(line, "=")
+			if len(lineSplit) == 2 {
+				data[dtclient.DataIngestToken] = []byte(lineSplit[1])
+			}
+		}
+	}
+	return result
 }
 
 func (s *SecretGenerator) createSecretForNSlist( //nolint:revive // argument-limit
--- a/pkg/webhook/mutation/pod/otlp/mutator.go
+++ b/pkg/webhook/mutation/pod/otlp/mutator.go
@@ -54,6 +54,8 @@ func (OTLPMutator) IsEnabled(request *dtwebhook.BaseRequest) bool {
 }
 
 func (OTLPMutator) IsInjected(request *dtwebhook.BaseRequest) bool {
+	// TODO the check for already present env vars is also done in the Mutate() implementation, so I'm not sure we need this check here as well.
+	// Maybe a check for an annotation indicating that the OTLP vars have already been injected might be better here.
 	otlpEndpointEnvVars := []string{envVarOtlpExporterEndpoint, envVarOtlpTraceEndpoint, envVarOtlpMetricsEndpoint, envVarOtlpLogsEndpoint}
 	// if any of the OTEL_EXPORTER_* variables are already set in all containers, return false, as we do not want to overwrite them
 	// TODO check if also initContainers should be considered
--- a/pkg/webhook/mutation/pod/otlp/mutator.go
+++ b/pkg/webhook/mutation/pod/otlp/mutator.go
@@ -1,0 +1,231 @@
+package otlp
+
+import (
+	"fmt"
+	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
+	"github.com/Dynatrace/dynatrace-operator/pkg/consts"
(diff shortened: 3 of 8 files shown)


pkg/injection/namespace/bootstrapperconfig/bootstrapperconfig.go
```diff
@@ -88,11 +104,37 @@ func (s *SecretGenerator) GenerateForDynakube(ctx context.Context, dk *dynakube.
...
+		err = s.createSecretForNSlist(ctx, consts.OtlpIngestCertsSecretName, CertsConditionType, nsList, dk, certs)
+		if err != nil {
+			return errors.WithStack(err)
+		}
 	}
 
 	return nil
 }
 
+func extractApiToken(data map[string][]byte) map[string][]byte {
```
bacherfl, line 120: see https://github.com/Dynatrace/dynatrace-operator/pull/5397/files#diff-62036e9d0edf8b3704940783898f28f99f2c1611e1da5461331811aab7db2462R74-R81 for why this is currently done this way

pkg/injection/namespace/bootstrapperconfig/bootstrapperconfig.go
```diff
@@ -70,6 +71,21 @@ func (s *SecretGenerator) GenerateForDynakube(ctx context.Context, dk *dynakube.
...
 			return errors.WithStack(err)
 		}
+
+		// TODO check if this can be refactored to be cleaner
+		// here we extract the token again from the 'endpoint.properties', which includes both the DT_METRICS_INGEST_URL and DT_METRICS_INGEST_API_TOKEN in a single string,
+		// while for the OTLP env vars we need the ingest token as a separate key within the secret.
+		// This way we do not have to look up the secret containing the token in the dynatrace namespace again (i.e. less kubernetes API calls),
+		// but this is also not ideal, as the extractApiToken relies on the endpoint.properties to be in a certain format.
+		// Ideally, we can refactor this to retrieve the endpoint and token at the beginning and then generate the content for the secrets,
+		// as they are required by the bootstrapper and the otlp env vars, respectively.
```
0sewa0, line 81: you could use the "tokens" that is in the main reconciler https://github.com/Dynatrace/dynatrace-operator/blob/cee6877a2e4de9e12c8634ca9447fc7c9e6dd62d/pkg/controllers/dynakube/controller.go#L127-L128

This could be passed to the `bootstrapperconfig` as well, so it also doesn't query for them again. (separate improvment)

pkg/injection/namespace/bootstrapperconfig/bootstrapperconfig.go
```diff
@@ -88,11 +104,37 @@ func (s *SecretGenerator) GenerateForDynakube(ctx context.Context, dk *dynakube.
 		if err != nil {
 			return errors.WithStack(err)
 		}
+
+		// create cert secret for otlp configuration
+		// TODO is the creation of the OTLP related secrets tied to the same preconditions as the creation of the bootstrap secrets?
+		// TODO should we decouple the OTLP secret creation to be independent of any errors during the creation of the other secrets?
```
0sewa0, line 110: I would be for a separate otlp secret, created by a separate package

The whole idea behind the `bootstarpperconfig` is that it creates a secret with all the necessary config files for the injected init-container, the webhook doesn't read it really, just ensures the existence of the secret (as best it can) and mounts it to the ini-container

the interesting thing is the certs only secret, the maybe could be reused, depending on how you will use it

pkg/injection/namespace/bootstrapperconfig/bootstrapperconfig.go
0sewa0: can be solved an other way, see [comment](https://github.com/Dynatrace/dynatrace-operator/pull/5397/files#r2336269320)

pkg/webhook/mutation/pod/otlp/mutator.go
```diff
@@ -0,0 +1,234 @@
...
+func NewMutator() dtwebhook.Mutator {
+	return &OTLPMutator{}
+}
+
+func (OTLPMutator) IsEnabled(request *dtwebhook.BaseRequest) bool {
+	//TODO implement when structure for OTLP enrichment config in Dynakube has been defined.
+	return true
+}
+
+func (OTLPMutator) IsInjected(request *dtwebhook.BaseRequest) bool {
```
0sewa0, line 56: this is more of a "legacy" part of the interface

it is mainly used for differentiating between: 
- Mutate -> changes need to be made to the init-container (general setup, add volumes, etc) + mutate the user container
- Reinvoke -> Only (new) container needs some attention, the init-container is fine
   - History: It used to be that reinvoke still touched the init-container, but not the volumes and such, this has been changed. 

For otelp specifically, you only care if the envs are on all the containers you want to "inject", no init-container to worry about

pkg/webhook/mutation/pod/otlp/mutator.go
```diff
@@ -0,0 +1,234 @@
...
+	for i := range request.Pod.Spec.Containers {
+		envVarsToCheck := []string{envVarOtlpExporterEndpoint, envVarOtlpExporterHeaders, envVarOtlpExporterCertificate, envVarOtlpExporterProtocol}
+		for _, envVar := range envVarsToCheck {
+			// TODO how to handle EnvFrom? OTLP env vars could also be imported via a configMap
+			if isEnvVarSet(request.Pod.Spec.Containers[i].Env, envVar) {
+				continue
+			}
+		}
+		// TODO use secret specifically for otlp
+		// TODO the creation of this secret via the SecretGenerator used by the Dynakube reconciler can take quite some time (~5mins when tested locally), due to the reconciliation interval
```
0sewa0, line 116: This is the reason the ["replicate" logic exists](https://github.com/Dynatrace/dynatrace-operator/blob/cee6877a2e4de9e12c8634ca9447fc7c9e6dd62d/pkg/webhook/mutation/pod/handle.go#L156)

The idea is that the `SecretGenerator` maintains 2 "types of" Secrets:
- A "source" Secret, within the dynatrace namespace
- Secrets, in the namespaces that exist at time of reconcile

This "replicate" is for namespaces that are created within this time window between reconciles.
- So the webhook doesn't have to regen the secret on the fly, just copy it over from the "source" Secret

pkg/webhook/mutation/pod/otlp/mutator.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func injectOtlpTraceEnvVars(c *corev1.Container, apiURL string, injectCert bool) {
+	// check if any environment variable related to the otlp trace exporter is already set.
+	// If yes, do not set any related env var to not change any customer specific settings
+
+	envVarsToCheck := []string{envVarOtlpTraceEndpoint, envVarOtlpTraceHeaders, envVarOtlpTraceCertificate, envVarOtlpTraceProtocol}
+
+	for _, envVar := range envVarsToCheck {
+		// TODO how to handle EnvFrom? OTLP env vars could also be imported via a configMap
```
0sewa0, line 134: you can't, unless you want to check each and every configMap/Secret mounted as `EnvFrom`
- this would require (potentially) lots of requests
- this would require LOTS of extra permissions 

this should be documented as a known limitation

sidenote:
> // If yes, do not set any related env var to not change any customer specific settings

Personal opinion: I do not see a valid usecase where I want half of my OTLP envs to come from DT and another half from myself. Sounds like problems begging to happen. (keep in mind, my knowledge about OTLP is non existent)

I would just abort the enrichment if any of the OTLP envs are already present, or overwrite them.

0sewa0: > I would just abort the enrichment if any of the OTLP envs are already present, or overwrite them.

had a discussion about this, in more details, and we actually agreed
so the general gist:

there are 3 possible signals we can configure: traces, metrics, logs
- all of these have there separate, not conlicting env vars

So if the user has manually configured 1 of the signals, but wants Dynatrace to add the other 2, that is fine and possible, without override.

But if the customer has manually configured 1 of the signals, AND configrued Dynatrace to add the same signal, then it has to be specified by the customer that "it is ok" to override this. (override field in the CR)
- if we are not allowed to override, then we leave it alone

There is a more hidden conflict, which also requires the override. You can configure OTELP to send EVERY singnal to 1 place, with a single set of less specific envvars. Now, if we inject a set of more specific envvars for a given signal then that takes precedent, so we overruled the users config, so override needs to be specified by the user for this to be allowed.

pkg/controllers/dynakube/otelc/endpoint/reconciler.go
```diff
@@ -99,17 +98,9 @@ func (r *Reconciler) reconcileConfigMap(ctx context.Context) error {
...
-	if r.dk.ActiveGate().IsEnabled() {
-		tenantUUID, err := r.dk.TenantUUID()
-		if err != nil {
-			return data, err
-		}
-
-		serviceFQDN := capability.BuildServiceName(r.dk.Name) + "." + r.dk.Namespace + ".svc"
-
-		dtEndpoint = fmt.Sprintf("https://%s/e/%s/api/v2/otlp", serviceFQDN, tenantUUID)
+	dtEndpoint, err := otlp.GetOtlpIngestEndpoint(r.dk)
```
andriisoldatenko, line 101: since package already has otlp, it's a best practice to simplify the name of the function.
```
otlp.GetIngestEndpoint(r.dk)
```

pkg/webhook/mutation/pod/otlp/mutator.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
+	"github.com/Dynatrace/dynatrace-operator/pkg/consts"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/otlp"
+	dtwebhook "github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/mutator"
+	corev1 "k8s.io/api/core/v1"
+	"slices"
+)
+
+const (
+	envVarOtlpTraceEndpoint    = "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT"
```
andriisoldatenko, line 14: Better to **not** start from envVar, then it's hard to find it and hard to read. 
also you can create `env.go` and put all of them there.

So something like it is mroe readable and based on what we already did in past:

```
otlpTraceEndpointEnv
```

pkg/webhook/mutation/pod/otlp/mutator.go
```diff
@@ -0,0 +1,234 @@
...
+	activeGateTrustedCertSecretKeyPath = "activegate.pem"
+	activeGateTrustedCertMountPath     = "activegate-server-certs"
+	envActiveGateTrustedCert           = activeGateTrustedCertMountPath + "/" + activeGateTrustedCertSecretKeyPath
+
+	envVarDtApiToken = "DT_API_TOKEN"
+
+	otlpAPIPath = "/v2/otlp"
+)
+
+type OTLPMutator struct{}
```
andriisoldatenko, line 45:
```suggestion
type Mutator struct{}
```

`otlp.Mutator` much easy to read.

pkg/webhook/mutation/pod/otlp/mutator.go
```diff
@@ -0,0 +1,234 @@
...
+			gotOtlpExporterEnvVars++
+		}
+	}
+	if gotOtlpExporterEnvVars == len(request.Pod.Spec.Containers) {
+		return true
+	}
+	return false
+}
+
+func (OTLPMutator) Mutate(request *dtwebhook.MutationRequest) error {
```
andriisoldatenko, line 83: here we use `OTLP`, and in different places we use Otlp, so please stick with one name convention, we prefer and enforce `OTLP` since it's abbrv

pkg/webhook/mutation/pod/otlp/mutator.go
```diff
@@ -0,0 +1,234 @@
...
+	envVarsToCheck := []string{envVarOtlpLogsEndpoint, envVarOtlpLogsHeaders, envVarOtlpLogsCertificate, envVarOtlpLogsProtocol}
+
+	for _, envVar := range envVarsToCheck {
+		// TODO how to handle EnvFrom? OTLP env vars could also be imported via a configMap
+		if isEnvVarSet(c.Env, envVar) {
+			return
+		}
+	}
+
+	addEnvVarLiteralValue(c, envVarOtlpLogsEndpoint, fmt.Sprintf("%s/%s", apiURL, "logs"))
```
andriisoldatenko, line 184: Can we rethink and maybe reuse std?
```
slices.ContainsFunc()
```

to check if it exists and if not just add it in one go?

andriisoldatenko: For example:
```
contains := slices.ContainsFunc(envs, func(env corev1.EnvVar) bool {
			if env.Name == "env1" {
				return true
			}
			return false
		})
		if !contains {
			envs := append(envs, corev1.EnvVar{Name: "env1", Value: "value1"})
		}
```

pkg/webhook/mutation/pod/otlp/mutator.go
```diff
@@ -0,0 +1,234 @@
...
+	addEnvVarLiteralValue(c, envVarOtlpLogsProtocol, "http/protobuf")
+	addEnvVarLiteralValue(c, envVarOtlpLogsHeaders, fmt.Sprintf("Authorization $(%s)", envVarDtApiToken))
+
+	if injectCert {
+		addEnvVarLiteralValue(c, envVarOtlpLogsCertificate, envActiveGateTrustedCert)
+	}
+}
+
+func addEnvVarLiteralValue(c *corev1.Container, envVar string, value string) {
+	for _, e := range c.Env {
```
andriisoldatenko, line 194: it seems like duplication of `isEnvVarSet`

pkg/webhook/mutation/pod/otlp/mutator.go
StefanHauth: > Personal opinion: I do not see a valid usecase where I want half of my OTLP envs to come from DT and another half from myself. Sounds like problems begging to happen. (keep in mind, my knowledge about OTLP is non existent)

The main use-case is easy migration of existing setups.

PR #5404, opened 2025-09-11
Add possibility to use helm charts on E2E tests

albertogdd:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-8880

We want to be able to run our E2E tests using Helm charts from quay/AWS registries. 
**Our main use case is to run our E2E tests daily using the new Nightly chart from quay.**

## How can this be tested?

Make sure your env is clean
```sh
make undeploy
```

Run any E2E test (and export the new env var `HELM_CHART`)
```sh
export HELM_CHART=oci://quay.io/dynatrace/dynatrace-operator:0.0.0-nightly-chart \
make test/e2e/standard
```

The deployed operator should come from the provided helm chart

```yaml
apiVersion: v1
kind: Pod
metadata:
  labels:
    app.kubernetes.io/component: operator
    app.kubernetes.io/name: dynatrace-operator
    app.kubernetes.io/version: 0.0.0-nightly-2025-09-11
    name: dynatrace-operator
    pod-template-hash: 545d75d69c
```

### DIFF
--- a/hack/make/tests/e2e.mk
+++ b/hack/make/tests/e2e.mk
@@ -27,156 +27,156 @@ test/e2e-publish:
 	exit $$RC
 
 ## Run standard e2e test only
-test/e2e/standard:
+test/e2e/standard: manifests/crd/helm
 	$(GOTESTCMD) -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 200m -count=1 ./test/scenarios/standard -args $(SKIPCLEANUP)
 
 ## Run istio e2e test only
-test/e2e/istio:
+test/e2e/istio: manifests/crd/helm
 	$(GOTESTCMD) -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 200m -count=1 ./test/scenarios/istio -args $(SKIPCLEANUP)
 
 ## Run no-csi e2e test only
-test/e2e/no-csi:
+test/e2e/no-csi: manifests/crd/helm
 	$(GOTESTCMD) -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 200m -count=1 ./test/scenarios/nocsi -args $(SKIPCLEANUP)
 
 ## Run release e2e test only
-test/e2e/release:
+test/e2e/release: manifests/crd/helm
 	$(GOTESTCMD) -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 20m -count=1 ./test/scenarios/release -args $(SKIPCLEANUP)
 
 ## Runs ActiveGate e2e test only
-test/e2e/activegate:
+test/e2e/activegate: manifests/crd/helm
 	$(GOTESTCMD) -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 20m -count=1  ./test/scenarios/nocsi -run "activegate" -args $(SKIPCLEANUP)
 
 ## Runs ActiveGate proxy e2e test only
-test/e2e/activegate/proxy:
+test/e2e/activegate/proxy: manifests/crd/helm
 	$(GOTESTCMD) -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 20m -count=1  ./test/scenarios/istio -run "activegate" -args $(SKIPCLEANUP)
 
 ## Runs ClassicFullStack e2e test only
-test/e2e/classic:
+test/e2e/classic: manifests/crd/helm
 	$(GOTESTCMD) -v -tags "$(shell ./hack/build/create_go_build_tags.sh true)" -timeout 20m -count=1  ./test/scenarios/nocsi -run "classic" -args  $(SKIPCLEANUP)
 
 ## Runs ClassicFullStack switch mode e2e test only
-test/e2e/classic/switchmodes:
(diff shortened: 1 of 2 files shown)


test/scenarios/env.go
```diff
@@ -4,11 +4,10 @@ package scenarios
 
 import "os"
 
-const installViaHelmEnvVar = "INSTALL_VIA_HELM"
 const HelmChartTagEnvVar = "USE_HELM_CHART_TAG"
 
 func InstallViaHelm() bool {
-	if os.Getenv(installViaHelmEnvVar) == "true" && os.Getenv(HelmChartTagEnvVar) != "" {
+	if os.Getenv(HelmChartTagEnvVar) != "" {
```
Incredible-O: 🧐 Is `USE_HELM_CHART_TAG` ever unset or empty?

Please correct me if I'm wrong, but this is what _I think_ would (normally) happen:

1️⃣ The input variable `use_helm_chart_tag`, while not required, is set to `'0.0.0-nightly-chart'` by default:

https://github.com/Dynatrace/dynatrace-operator/blob/859646de6d6e36c581e92c61a764776ab721fa91/.github/workflows/e2e-tests.yaml#L13-L17

:two: `use_helm_chart_tag` is then passed to the composite action `run-e2e` as `use-helm-chart-tag` (btw: you may want to rename `use_helm_chart_tag`, for the sake of consistency)

https://github.com/Dynatrace/dynatrace-operator/blob/859646de6d6e36c581e92c61a764776ab721fa91/.github/workflows/e2e-tests.yaml#L70-L76

:three: The value of `use-helm-chart-tag` is then passed to the environment variable `USE_HELM_CHART_TAG`:

https://github.com/Dynatrace/dynatrace-operator/blob/859646de6d6e36c581e92c61a764776ab721fa91/.github/actions/run-e2e/action.yaml#L100-L106

This makes me think `InstallViaHelm()` always returns `true` 🤔

.github/workflows/e2e-tests.yaml
```diff
@@ -10,6 +10,11 @@ on:
         description: 'Target branch to run E2E tests over'
         required: true
         default: 'main'
+      use_helm_chart_tag:
```
Incredible-O: What about:
```suggestion
      helm-chart-tag:
```

test/scenarios/env.go
albertogdd: `USE_HELM_CHART_TAG` can be unset/empty when running locally via make commands (and not exporting the env var). Also, while triggering the GH action manually in the UI and setting an empty value.

I put the default value with the nightly because this workflow is scheduled daily with the default values, and we want to test the nightly chart every day!

Incredible-O: Ah, I see. So this is expected behavior for our scheduled CI builds then. Makes sense 👍🏿

test/scenarios/istio/istio_test.go
```diff
@@ -34,11 +36,21 @@ func TestMain(m *testing.M) {
 	nsWithoutIstio := *namespace.New(operator.DefaultNamespace)
 	testEnv.BeforeEachTest(istio.AssertIstioNamespace())
 	testEnv.BeforeEachTest(istio.AssertIstiodDeployment())
-	testEnv.Setup(
-		helpers.SetScheme,
-		namespace.CreateForEnv(nsWithIstio),
-		operator.InstallViaMake(true),
-	)
+
+	if scenarios.InstallViaHelm() {
```
andriisoldatenko: I prefer to do it in more clear way:
- `make deploy` under the hood also installs the operator using Helm, but we use a local Helm chart.
https://github.com/dynatrace/dynatrace-operator/blob/87e0cdb07199b8069ba79fa74050c300588badc7/hack/make/deploy/deploy.mk#L12-L24
- so the only new we need to handle is install via helm but use custom params and OCI chart.
- So i would refactor `operator.InstallViaMake(false),` a bit instead.

.github/workflows/e2e-tests.yaml
albertogdd: I simplified the code and renamed this variable :+1:

test/scenarios/istio/istio_test.go
albertogdd: Great idea! Code got way simpler this way. Please review again :+1:

albertogdd:
Successful E2E run using this changes
https://github.com/Dynatrace/dynatrace-operator/actions/runs/17828691997/job/50687932865

hack/make/tests/e2e.mk
```diff
@@ -27,156 +27,156 @@ test/e2e-publish:
 	exit $$RC
 
 ## Run standard e2e test only
-test/e2e/standard: manifests/crd/helm
```
andriisoldatenko: but if user uses local chart we need to build it, wdyt?

andriisoldatenko (dismissed):

Incredible-O (dismissed):

hack/make/tests/e2e.mk
albertogdd: Fixed, pls approve again!

Incredible-O: Wasn't that supposed to fix a bug?

albertogdd: Andrii disagrees, I will just remove it from this PR and discuss it on the side

PR #5411, opened 2025-09-15
Extract metadata and OneAgent injection into separate package

bacherfl:
## Description

This PR extracts the logic related to the metadata and oneAgent injection into a separate package. This refactoring is a preparation for also adding a separate handler for the injection of OTLP related environment variables, which will also live in its own package

Part of https://dt-rnd.atlassian.net/browse/DAQ-13038

## How can this be tested?

- unit tests
- e2e-tests (`make test/e2e/applicationmonitoring/metadataenrichment`)


pkg/webhook/mutation/pod/handler/injection/handle.go
```diff
@@ -7,29 +7,62 @@ import (
...
+
+	webhookPodImage string
+	isOpenShift     bool
+}
+
+func New( //nolint:revive
+	kubeClient client.Client,
+	apiReader client.Reader,
+	recorder events.EventRecorder,
+	webhookPodImage string,
```
andriisoldatenko, line 33: Any reason why we use different orders of inputs here and below in a struct?

pkg/webhook/mutation/pod/handler/injection/init_test.go
```diff
@@ -17,11 +20,22 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	testWebhookImage        = "test-wh-image"
+	testNamespaceName       = "test-namespace"
+	testClusterID           = "test-cluster-id"
+	testPodName             = "test-pod"
+	testDynakubeName        = "test-dynakube"
+	testUser          int64 = 420
```
andriisoldatenko:
```suggestion
	testUser         = int64(420)
```

andriisoldatenko: just easy to read

pkg/webhook/mutation/pod/handler/injection/handle.go
bacherfl: no particular reason, but it certainly makes sense to keep them in the same order. I have updated the PR accordingly with [d35704f](https://github.com/Dynatrace/dynatrace-operator/pull/5411/commits/d35704fd8efbf3e6dcf611a8b209db3700e92130)

pkg/webhook/mutation/pod/handler/injection/init_test.go
bacherfl: updated with [d35704f](https://github.com/Dynatrace/dynatrace-operator/pull/5411/commits/d35704fd8efbf3e6dcf611a8b209db3700e92130)

PR #4846, opened 2025-05-05
Remove nodes-controller

albertogdd:
https://dt-rnd.atlassian.net/browse/DAQ-3933

## Description

Remove nodes-controller from the operator.

## How can this be tested?

- Deploy a Dynakube and check that no `nodes-controller-cache` ConfigMap exists. (you may need to manually remove the old one)

PR #5422, opened 2025-09-19
revert Remove nodes-controller (#4846)

andriisoldatenko:

### DIFF
--- a/pkg/clients/dynatrace/dynatrace_client.go
+++ b/pkg/clients/dynatrace/dynatrace_client.go
@@ -8,9 +8,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"os"
-	"strconv"
-	"strings"
 	"time"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/clients/utils"
@@ -19,6 +16,19 @@ import (
 
 const ApiTokenHeader = "Api-Token "
 
+type HostNotFoundErr struct {
+	IP string
+}
+
+func (e HostNotFoundErr) Error() string {
+	return fmt.Sprintf("host not found for ip: %v", e.IP)
+}
+
+type hostInfo struct {
+	version  string
+	entityID string
+}
+
 // client implements the Client interface.
 type dynatraceClient struct {
 
@@ -26,6 +36,8 @@ type dynatraceClient struct {
 	now time.Time
 
 	httpClient *http.Client
+
+	hostCache map[string]hostInfo
 
 	url       string
 	apiToken  string
@@ -41,8 +53,7 @@ type tokenType int
 const (
 	dynatraceApiToken tokenType = iota
 	dynatracePaaSToken
-	installerUrlToken     // in this case we don't care about the token
-	defaultMaxResponseLen = 1000
+	installerUrlToken // in this case we don't care about the token
 )
 
 // makeRequest does an HTTP request by formatting the URL from the given arguments and returns the response.
@@ -104,7 +115,7 @@ func (dtc *dynatraceClient) getServerResponseData(response *http.Response) ([]by
 
 	if response.StatusCode != http.StatusOK &&
 		response.StatusCode != http.StatusCreated {
-		return responseData, dtc.handleErrorResponseFromAPI(responseData, response.StatusCode, response.Header)
+		return responseData, dtc.handleErrorResponseFromAPI(responseData, response.StatusCode)
 	}
 
 	return responseData, nil
@@ -151,7 +162,7 @@ func (dtc *dynatraceClient) makeRequestForBinary(ctx context.Context, url string
 	return hex.EncodeToString(hash.Sum(nil)), err
 }
 
-func (dtc *dynatraceClient) handleErrorResponseFromAPI(response []byte, statusCode int, headers http.Header) error {
+func (dtc *dynatraceClient) handleErrorResponseFromAPI(response []byte, statusCode int) error {
 	se := serverErrorResponse{}
 
 	contentType := "unknown"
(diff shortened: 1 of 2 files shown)


pkg/controllers/dynakube/token/feature.go
```diff
@@ -26,6 +26,13 @@ func (feature *Feature) IsScopeMissing(scopes []string) (bool, []string) {
 
 func getFeaturesForAPIToken(paasTokenExists bool) []Feature {
 	return []Feature{
+		{
+			Name:           "Access problem and event feed, metrics, and topology",
+			RequiredScopes: []string{dtclient.TokenScopeDataExport},
+			IsEnabled: func(dk dynakube.DynaKube) bool {
+				return true
```
aorcholski:
```suggestion
				return dk.FF().IsHostAvailabilityDetectionEnabled()
```
_DataExport_ scope is need if the FF is enabled, right?

andriisoldatenko: true

andriisoldatenko: fixed in 4ca939f0b

sowens81:
Watching for this fix to go live.

pkg/api/exp/flag_test.go
```diff
@@ -40,6 +40,39 @@ func TestGetNoProxy(t *testing.T) {
 	}
 }
 
+func TestIsHostAvailabilityDetectionEnable(t *testing.T) {
```
waodim:
```suggestion
func TestIsHostAvailabilityDetectionEnabled(t *testing.T) {
```

pkg/controllers/nodes/nodes_controller.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+}
+
+func (controller *Controller) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) { //nolint: revive
+	nodeName := request.NamespacedName.Name
+	dk, err := controller.determineDynakubeForNode(nodeName)
+
+	if dk != nil && !dk.FF().IsHostAvailabilityDetectionEnabled() {
+		// feature flag is disabled, skipping reconciliation
+		log.Info("skipping reconciling node name due to FF flag", "node", nodeName)
```
waodim:
```suggestion
		log.Info("skipping reconciling node name due to host availability detection feature flag being disabled", "node", nodeName)
```

pkg/api/exp/flag_test.go
```diff
@@ -40,6 +40,39 @@ func TestGetNoProxy(t *testing.T) {
 	}
 }
 
+func TestIsHostAvailabilityDetectionEnabled(t *testing.T) {
```
StefanHauth: This is a copy of GetNoProxy() test, is this intentional?

StefanHauth (changes requested):

pkg/controllers/nodes/nodes_controller.go
```diff
@@ -0,0 +1,366 @@
...
+	dk, err := controller.determineDynakubeForNode(nodeName)
+
+	if dk != nil && !dk.FF().IsHostAvailabilityDetectionEnabled() {
+		// feature flag is disabled, skipping reconciliation
+		log.Info("skipping reconciling node name due to host availability detection feature flag being disabled", "node", nodeName)
+
+		return reconcile.Result{}, nil
+	}
+
+	log.Info("reconciling node name", "node", nodeName)
```
aorcholski, line 69:
```suggestion
    if dk != nil {
		log.Info("reconciling node name", "node", nodeName, "dynakube", dk.Name)
	} else {
		log.Info("reconciling node name", "node", nodeName)
	}

```

pkg/api/exp/flag_test.go
andriisoldatenko: removed, origianlly i was thinkig to write new tests, but then i found we test bool FF so it's not nessesary. 
removed 04165a166

pkg/controllers/nodes/nodes_controller.go
andriisoldatenko: I don't want to change this @aorcholski let's stick with old implementation.

albertogdd:
> Hi @albertogdd @albertogdd We are a customer heavily impacted by this. We've seen increase in the number of alerts. Is there any regression testing in place to avoid a recurrence in the future?

@hibri our whole team is working on a new release to fix this issue. More information will come in the official docs.

cmd/operator/manager.go
```diff
@@ -30,7 +30,7 @@ func getControllerAddFuncs(isOLM bool) []controllerSetupFunc {
 		edgeconnect.Add,
 	}
 
-	if envvars.GetBool(consts.HostAvailabilityDetectionEnvVar, true) {
+	if envvars.GetBool(consts.HostAvailabilityDetectionEnvVar, false) {
```
StefanHauth: This would mean that if its not set we don't activate the feature. Shouldn't the default if not set be "enable"?

andriisoldatenko: do we want to enable it even without env var? and only disable if user explicitly set env var + false?

andriisoldatenko: fixed in d1d5a59f4

pkg/api/exp/flag.go
```diff
@@ -11,6 +11,8 @@ const (
 	PublicRegistryKey = FFPrefix + "public-registry"
 	NoProxyKey        = FFPrefix + "no-proxy"
 
+	HostAvailabilityDetection = FFPrefix + "host-availability-detection"
```
andriisoldatenko, line 14: removing this

aorcholski (approved):

albertogdd (approved):

PR #5423, opened 2025-09-19
Fix nil values on scheduled E2E workflows

albertogdd:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-14185

Latest E2E changes introduced a new input parameter with a default value. This value turned out to be nil when triggering it on a scheduled workflow (see: https://github.com/Dynatrace/dynatrace-operator/actions/runs/17844754882/job/50741870724#step:4:472)

This is because `default` values are not considered when workflows are scheduled (see this post https://stackoverflow.com/a/73495922)

This PR use expressions to properly set default values when they are nil. Only downside is that then the helm-chart input parameter becomes mandatory (if not set, it uses always the nightly chart)

## How can this be tested?

- Merging to main and waiting for overnight execution

### DIFF
--- a/cmd/operator/manager.go
+++ b/cmd/operator/manager.go
@@ -6,9 +6,12 @@ import (
 	"time"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/scheme"
+	"github.com/Dynatrace/dynatrace-operator/pkg/consts"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/certificates"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/edgeconnect"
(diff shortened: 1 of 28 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
         with:
           flc-namespace: dto-daily
           flc-environment: dto-k8s-latest-flc
-          helm-chart: ${{ github.event.inputs.helm-chart }}
+          helm-chart: ${{ github.event.inputs.helm-chart || 'oci://quay.io/dynatrace/dynatrace-operator:0.0.0-nightly-chart' }}
```
StefanHauth, line 40: Why is it needed to provide the default value here again? Feels redundant.

albertogdd: Because of this https://stackoverflow.com/a/73495922

andriisoldatenko (approved):

PR #5426, opened 2025-09-19
[release-1.7] Reintroduce “Mark for termination” event to improve graceful node shutdown detection for setups using autoscaling

andriisoldatenko:
## Description

same as https://github.com/Dynatrace/dynatrace-operator/pull/5422

## How can this be tested?

same as https://github.com/Dynatrace/dynatrace-operator/pull/5422


wepudt (approved):
checked active deployment with simple DK with enabled, disabled, and enabled FF again. 
nodecontroller was working as expected.

PR #5434, opened 2025-09-22
Add OTLP webhook handler 

bacherfl:
## Description

This PR adds the boilerplate code for executing the OTLP webhook handler, as described in https://dt-rnd.atlassian.net/browse/DAQ-13038. The implementation of the OTLP exporter configuration, as well as the OTEL resource attribute injection will be covered in separate PRs

## How can this be tested?

- Unit tests
- E2E tests: `make test/e2e/applicationmonitoring/metadataenrichment` - These will make sure that the existing functionality covered by the webhook works as it did previously. E2E tests specifically for the OTLP configuration will be added when the feature is implemented


pkg/webhook/mutation/pod/webhook_test.go
```diff
@@ -199,7 +238,7 @@ func getTestWebhookPod(t *testing.T) corev1.Pod {
 	}
 }
 
-func createTestWebhook(t *testing.T, oaMut, metaMut podwebhook.Mutator, objects ...client.Object) *webhook {
+func createTestWebhook(t *testing.T, oaMut, metaMut podwebhook.Mutator, otlpHandler handler.Handler, objects ...client.Object) *webhook {
```
avorima: I should've mentioned this in the refactoring PR, but this shouldn't initialize an injection handler using the actual package but just use a mock like you're doing with the otlp handler now.

pkg/webhook/mutation/pod/handler/otlp/config.go
```diff
@@ -0,0 +1,17 @@
...
+var (
+	log = logd.Get().WithName("pod-mutation-otlp")
+)
+
+const (
+	AnnotationPrefix = "otlp-exporter-configuration"
+	// AnnotationOTLPInjectionEnabled controls whether the automatic injection of OTLP env vars and resource attributes should happen for a pod
+	AnnotationOTLPInjectionEnabled = AnnotationPrefix + ".dynatrace.com/inject"
+	// AnnotationOTLPInjected indicates whether the OTLP env vars and resource attributes have already been injected into a pod
+	AnnotationOTLPInjected = AnnotationPrefix + ".dynatrace.com/injected"
```
avorima, line 16: should also have a `/reason` annotation for the error case

pkg/webhook/mutation/pod/webhook.go
```diff
@@ -87,12 +88,25 @@ func (wh *webhook) Handle(ctx context.Context, request admission.Request) admiss
 	wh.recorder.Setup(mutationRequest)
 
 	originalPod := mutationRequest.Pod.DeepCopy()
+
+	var mutErr *dtwebhook.MutatorError
 	if err := wh.injectionHandler.Handle(mutationRequest); err != nil {
```
avorima: this could be wrapped in a separate function to only extract the MutatorError once. don't know how many more handlers the webhook will have eventually

pkg/webhook/mutation/pod/webhook_test.go
bacherfl: that makes sense - I have adapted that accordingly in [036fdb6](https://github.com/Dynatrace/dynatrace-operator/pull/5434/commits/036fdb6dc5cdbabd17a9d8635e4050cd4d7e0fa2)

pkg/webhook/mutation/pod/handler/otlp/config.go
bacherfl: added in [036fdb6](https://github.com/Dynatrace/dynatrace-operator/pull/5434/commits/036fdb6dc5cdbabd17a9d8635e4050cd4d7e0fa2)

pkg/webhook/mutation/pod/webhook.go
bacherfl: i have restructured that section in [036fdb6](https://github.com/Dynatrace/dynatrace-operator/pull/5434/commits/036fdb6dc5cdbabd17a9d8635e4050cd4d7e0fa2) - should be more compact now

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
```diff
@@ -0,0 +1,40 @@
+package exporter
```
0sewa0, line 1: Just a note/idea:

I like this idea of `Handler`s and `Mutator`s.

But each is `Mutator` is directly bound/tied to their `Handler`. 
Which is kinda hidden from the package structure.

We should think about a different package structure (when all is done and still make sense) that reflects this:
```
pkg/webhook/mutation/pod/
|
|--- handlers/
|           |
|           |--- otlp/
|           |       |
|           |       |--- mutators/
|           |
|           |--- injection/
|                   |
|                   |--- mutators/
|
```
And have the interfaces and related helpers in some common place

bacherfl: i agree, that makes sense - would you prefer to include the suggested change in this PR, or in one of the follow ups when it comes to implementing the OTLP env var and resource attributes mutators? Probably the latter will make more sense as it will become more apparent on how everything should ideally be structured during the implementation, but I can also do that change as part of this PR - as you prefer

0sewa0: me say:
> when all is done and still make sense

you say:
> Probably the latter will make more sense as it will become more apparent on how everything should ideally be structured 

we have an agreement :D 

do it later, when we have a better understanding of the packages we will have
- the workload name detection logic will also have to land somewhere, so

PR #5437, opened 2025-09-22
WIP Add livenessprobe to the AG 'activegate' container 

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-13631)

## Description

Livenessprobe settings agreed with AG PA, see the ticket.

## How can this be tested?

make unittests


pkg/controllers/dynakube/activegate/internal/statefulset/statefulset.go
```diff
@@ -228,6 +228,20 @@ func (statefulSetBuilder Builder) buildBaseContainer() []corev1.Container {
 			FailureThreshold:    3,
 			TimeoutSeconds:      2,
 		},
+		LivenessProbe: &corev1.Probe{
+			ProbeHandler: corev1.ProbeHandler{
+				HTTPGet: &corev1.HTTPGetAction{
+					Path:   "/rest/state",
```
andriisoldatenko, line 234: does it work with all AG versions?

aorcholski: yes

PR #5452, opened 2025-09-25
Introduce script for sanitizing git branch names

Incredible-O:
## Description
Currently, we sanitize branch names (or reference them) in multiple places throughout our code base. This can cause bugs, like https://github.com/Dynatrace/dynatrace-operator/pull/5304, for example.

**Example:** If we were on a branch named `some-release-1.2`, what would be the name of the corresponding **image tag**?
- Our CI would build and push an image with the tag `snapshot-some-release-1.2`
  - the [Docker metadata action](https://github.com/docker/metadata-action) (used by our CI) _does_ sanitize the tag and its suffix, however, dots are valid characters. Thus, they aren't replaced with hyphens.
- Our build system would build an image with the tag `snapshot-some-release-1-2`
  - notice how that one dot _would_ be replaced in this case 🤠

💡 Solve this issue by introducing a script that is called from both our CI and our build scripts. 

## How can this be tested?
1. Create a branch, make sure to add dots or other special characters of your choice
2. Try one or multiple of the following use cases:
   - Run `make images/build/push`, check the tag of the pushed image
   - Invoke the new script directly, maybe try `hack/build/ci/sanitize-branch-name.sh release-1.2.3` or something
   - Test the CI in a fork


hack/build/ci/sanitize-branch-name.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+readonly BRANCH_NAME=${1}
+
+if [ -z "${BRANCH_NAME}" ]; then
+  echo "Usage: $0 <branch name>"
+  exit 1
+fi
+
+echo "${BRANCH_NAME//[^a-zA-Z0-9_-]/-}"
```
avorima: can we just include the . in our valid characters so that we have parity with docker metadata?
```suggestion
echo "${BRANCH_NAME//[^a-zA-Z0-9_.-]/-}"
```

Incredible-O: Thanks, addressed in ab1131cfc

hack/build/ci/sanitize-branch-name.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+readonly BRANCH_NAME=${1}
+
+if [ -z "${BRANCH_NAME}" ]; then
+  echo "Usage: $0 <branch name>"
```
avorima: probably better to write to stderr (`>&2`) to prevent accidentally capturing this as output

avorima (commented):
one last thing, LGTM otherwise

hack/build/ci/sanitize-branch-name.sh
Incredible-O: Good idea, thanks! Addressed in 6e33aba4b

avorima (approved):
verified results with make and direct calls to the script

PR #5453, opened 2025-09-25
Rename otlc.token file

StefanHauth:
## Description
The token file used for the token between EEC and datasources shall be renamed to `datasource.token`. This token will be used by other datasources not only by OTelC (as now for Prometheus). Therefore it shall get a more general name.

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-13637)

## How can this be tested?
* deploy a Dynakube with Prometheus extensions enabled
* check that OTelC starts correctly (initial CrashLoop Backoffs until AG is ready are to be expected)
* unit tests
* E2E tests for extensions


pkg/controllers/dynakube/extension/secret.go
```diff
@@ -15,81 +15,56 @@ import (
...
-			conditions.SetSecretGenFailed(r.dk.Conditions(), secretConditionType, errors.Wrap(err, "error generating otelc token"))
-
-			return err
-		}
-
-		newSecret, err := r.buildSecret(*newEecToken, *newOtelcToken)
-		if err != nil {
-			log.Info("failed to generate extension secret")
-			conditions.SetSecretGenFailed(r.dk.Conditions(), secretConditionType, err)
+		newSecret, err = r.newSecret()
```
aorcholski: could we use additional function `isMigrationNeeded` instead of "positive" ErrNoMigration error?
```
	var newSecret *corev1.Secret

      newSecret, err = r.newSecret()
    } else 
      if r.isMigrationNeeded(existingSecret) {
   		  newSecret, err = r.migrateSecret(existingSecret)
      } else {
		// no new secret to create, either because it already exists or migration not needed
		log.Info("extensions token secret no migration required")
		return nil    
      }
    }

	if err != nil {
		log.Info("migration of extension secret failed")
		return err
	}
	...
	
func (r *reconciler) isMigrationNeeded(existingSecret *corev1.Secret) (*corev1.Secret, error) {
	if _, exists := existingSecret.Data[consts.DatasourceTokenSecretKey]; exists {
		// no migration needed, datasource token file name already correct
		return false
	}
    return true
}	
```

pkg/controllers/dynakube/extension/secret.go
```diff
@@ -98,3 +75,82 @@ func (r *reconciler) buildSecret(eecToken dttoken.Token, otelcToken dttoken.Toke
...
+	if err != nil {
+		conditions.SetSecretGenFailed(r.dk.Conditions(), secretConditionType, err)
+
+		return nil, err
+	}
+
+	return migratedSecret, nil
+}
+
+func (r *reconciler) cleanupSecret(ctx context.Context) {
```
andriisoldatenko: this function name is not accurate, it does more then `cleanupSecret` it does:
- `remove status condition` and 
- `remove secret`

pkg/controllers/dynakube/extension/secret.go
```diff
@@ -98,3 +75,82 @@ func (r *reconciler) buildSecret(eecToken dttoken.Token, otelcToken dttoken.Toke
...
+
+	return newSecret, nil
+}
+
+func (r *reconciler) migrateSecret(existingSecret *corev1.Secret) (*corev1.Secret, error) {
+	const oTelcTokenKey = "otelc.token"
+
+	if _, exists := existingSecret.Data[consts.DatasourceTokenSecretKey]; exists {
+		// no migration needed, datasource token file name already correct
+		return nil, ErrNoMigration
```
andriisoldatenko: i don't think logically returnig error for case when migration is no needed is good idea, instead change function to:
```
migrateSecret(target *corev1.Secret) (*corev1.Secret, bool) {
isMigrated := false

...
return secret, isMigrated
```

pkg/controllers/dynakube/extension/secret.go
andriisoldatenko: I think we can do it much simplier, and easy to remove later:

```   
        if r.migrationNeed(ctx) {
              // all you need to do is remove condition here
              // then reconcileSecret will do the rest.
        }
	err := r.reconcileSecret(ctx)
	if err != nil {
		return err
	}
```

pkg/controllers/dynakube/extension/secret.go
StefanHauth: This sentinel error was actually enforced by some of our linters on me.

test/features/extensions/upgrade/upgrade.go
```diff
@@ -0,0 +1,56 @@
...
+		componentDynakube.WithActiveGate(),
+	}
+
+	testDynakube := *componentDynakube.New(options...)
+
+	previousVersionDynakube := &dynakubev1beta5.DynaKube{}
+	previousVersionDynakube.ConvertFrom(&testDynakube)
+	componentDynakube.InstallPreviousVersion(builder, helpers.LevelAssess, &secretConfig, *previousVersionDynakube)
+
+	//	componentDynakube.Install(builder, helpers.LevelAssess, &secretConfig, testDynakube)
```
andriisoldatenko: pls remove if not needed

test/features/extensions/upgrade/upgrade.go
```diff
@@ -0,0 +1,56 @@
...
+
+	//	builder.Assess("active gate pod is running", checkActiveGateContainer(&testDynakube))
```
andriisoldatenko: same

pkg/controllers/dynakube/extension/secret.go
```diff
@@ -15,6 +15,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const OtelcTokenSecretKey = "otelc.token"
```
andriisoldatenko:
```suggestion
// Remove it after release-1.7 will be deprecated.
const DeprecatedOtelcTokenSecretKey = "otelc.token"
```

pkg/controllers/dynakube/extension/secret.go
```diff
@@ -39,15 +41,18 @@ func (r *reconciler) reconcileSecret(ctx context.Context) error {
...
 	if err != nil && !k8serrors.IsNotFound(err) {
 		log.Info("failed to check existence of extension secret")
 		conditions.SetKubeAPIError(r.dk.Conditions(), secretConditionType, err)
 
 		return err
 	}
 
-	if k8serrors.IsNotFound(err) {
+	// TODO: Remove in future release when migration is no longer needed
+	migrationNeeded := r.removeOldSecretAndConditionIfNeeded(ctx, existingSecret)
```
andriisoldatenko:
```suggestion
	migrationNeeded := r.removeDeprecatedSecretAndConditionIfNeeded(ctx, existingSecret)
```

test/features/extensions/upgrade/upgrade.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/Dynatrace/dynatrace-operator/test/helpers"
+	componentDynakube "github.com/Dynatrace/dynatrace-operator/test/helpers/components/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/test/helpers/components/operator"
+	"github.com/Dynatrace/dynatrace-operator/test/helpers/kubeobjects/statefulset"
+	"github.com/Dynatrace/dynatrace-operator/test/helpers/tenant"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func Feature(t *testing.T) features.Feature {
+	builder := features.New("extensions-upgrade")
```
andriisoldatenko:
```suggestion
	builder := features.New("deprecated-secret-upgrade-operator")
```
we already inside extensions folder, but let's be more concrete what we test.

pkg/controllers/dynakube/extension/reconciler_test.go
```diff
@@ -124,7 +124,52 @@ func TestReconciler_Reconcile(t *testing.T) {
...
+
+		fakeClient := fake.NewClient(oldSecret)
+		r := NewReconciler(fakeClient, fakeClient, dk)
+
+		err = r.Reconcile(context.Background())
+		require.NoError(t, err)
+
+		// assert extensions token is generated
+		var secretFound corev1.Secret
+		err = fakeClient.Get(context.Background(), client.ObjectKey{Name: testName + "-extensions-token", Namespace: testNamespace}, &secretFound)
```
andriisoldatenko, line 155:
```suggestion
		err = fakeClient.Get(t.Context(), client.ObjectKey{Name: testName + "-extensions-token", Namespace: testNamespace}, &secretFound)
```

pkg/controllers/dynakube/extension/reconciler_test.go
```diff
@@ -124,7 +124,52 @@ func TestReconciler_Reconcile(t *testing.T) {
...
+
+		oldSecret, err := k8ssecret.Build(dk, testName+"-extensions-token", secretData)
+		require.NoError(t, err)
+
+		conditions.SetSecretCreated(dk.Conditions(), secretConditionType, oldSecret.Name)
+
+		fakeClient := fake.NewClient(oldSecret)
+		r := NewReconciler(fakeClient, fakeClient, dk)
+
+		err = r.Reconcile(context.Background())
```
andriisoldatenko, line 150:
```suggestion
		err = r.Reconcile(t.Context())
```

andriisoldatenko (approved):

PR #5474, opened 2025-09-30
Feat/otlp env var mutator

bacherfl:
## Description

This PR adds the logic for the OTLP exporter mutator. Right now, it will set the `OTLP_*_ENDPOINT` and `OTLP_*_PROTOCOL` environment variables in the matching pods. The env vars referring to the AG certificate, as well as the DataIngest token will be added int separate PRs, respectively. To prevent partial configuration of the OTLP exporter env vars, a feature flag (`otlp-exporter-configuration`) has been added. This flag is currently set to `false` by default, and will bypass the OTLP mutator logic.

Part of: https://dt-rnd.atlassian.net/browse/DAQ-1381 - As mentioned, the ingest token and certificates will need to be added later, but It makes sense to split this up, to not make this PR too big, and focus on the override logic for existing user env vars.

## How can this be tested?

- Unit tests
- E2E tests: `make test/e2e/applicationmonitoring/metadataenrichment` - These will make sure that the existing functionality covered by the webhook works as it did previously. E2E tests specifically for the OTLP configuration will be added when the feature is implemented


pkg/api/exp/otlp_exporter_configuration.go
```diff
@@ -0,0 +1,11 @@
...
+
+const (
+	OTLPExporterConfigurationKey = FFPrefix + "otlp-exporter-configuration"
+)
+
+// IsOTLPExporterConfiguration controls whether the automatic configuration of the OTLP exporter via environment variables
+// for user containers is enabled.
+func (ff *FeatureFlags) IsOTLPExporterConfiguration() bool {
+	return ff.getBoolWithDefault(OTLPExporterConfigurationKey, false)
+}
```
0sewa0: The `DynaKube.Spec.OTLPExporterConfiguration` is already a thing, why do we need a feature-flag?

Why not just add a little helper [here](https://github.com/Dynatrace/dynatrace-operator/blob/821525078a38a3d4c3234ecb98460aa18decb7bc/pkg/api/latest/dynakube/otlpexporterconfiguration/props.go#L19)
```go
func (s *Spec) IsEnabled() bool {
	return s.IsLogsEnabled() || s.IsTracesEnabled() || s.IsMetricsEnabled()
}
```

You could also use the same pattern we use for other sections like [here](https://github.com/Dynatrace/dynatrace-operator/blob/821525078a38a3d4c3234ecb98460aa18decb7bc/pkg/api/latest/dynakube/logmonitoring_props.go#L7-L17)
- Currently it would be a bit overkill, as OTLP is rather simple and self-reliant.
   - but then you could use the signature `dk.OTELPConfiguration().IsEnabled()` and be consistent with the codebase

bacherfl: > The `DynaKube.Spec.OTLPExporterConfiguration` is already a thing, why do we need a feature-flag?

Initially it seemed like having this explicit feature flag to enable the (currently partial) OTLP env var injection might be the best approach, but after giving it a second thought, I think you are right - having a `IsEnabled()` helper for the spec will also do the trick.

> You could also use the same pattern we use for other sections like [here](https://github.com/Dynatrace/dynatrace-operator/blob/821525078a38a3d4c3234ecb98460aa18decb7bc/pkg/api/latest/dynakube/logmonitoring_props.go#L7-L17)
> 
> * Currently it would be a bit overkill, as OTLP is rather simple and self-reliant.
>   
>   * but then you could use the signature `dk.OTELPConfiguration().IsEnabled()` and be consistent with the codebase

Thanks for the suggestion - I think consistency with the code base is already a strong enough argument to have this for the OTLPConfiguration as well. Will update the PR to add that

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
```diff
@@ -15,26 +23,284 @@ func New() dtwebhook.Mutator {
...
+	mutated, err := m.mutate(request.BaseRequest)
+	if err != nil {
+		log.Error(err, "error during reinvocation of OTLP env vars mutator", "podName", request.PodName(), "namespace", request.Namespace.Name)
+	}
+
+	return mutated
+}
+
+func (m Mutator) mutate(request *dtwebhook.BaseRequest) (bool, error) {
+	otlpExporterConfig := request.DynaKube.OTLPExporterConfiguration()
```
0sewa0, line 80: nit: 
you call this `otlpExporterSpec := request.DynaKube.OTLPExporterConfiguration()` in other places
I am fine with either, but consistency would be nice

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
```diff
@@ -15,26 +23,284 @@ func New() dtwebhook.Mutator {
...
+		}
+	}
+
+	addEnvVarLiteralValue(c, OTLPLogsEndpointEnv, fmt.Sprintf("%s/%s", apiURL, "logs"))
+	addEnvVarLiteralValue(c, OTLPLogsProtocolEnv, "http/protobuf")
+
+	return true
+}
+
+func addEnvVarLiteralValue(c *corev1.Container, name string, value string) {
```
0sewa0: could use https://github.com/Dynatrace/dynatrace-operator/blob/d2ff055c9ab0f08bdb0ff746fe508cc4406efca2/pkg/util/kubeobjects/env/env.go#L37

(or add a little wrapper next to it, if you don't like to create `corev1.EnvVar` structs)

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
```diff
@@ -15,26 +23,284 @@ func New() dtwebhook.Mutator {
...
+
+		serviceFQDN := capability.BuildServiceName(dk.Name) + "." + dk.Namespace + ".svc"
+
+		dtEndpoint = fmt.Sprintf("https://%s/e/%s/api/v2/otlp", serviceFQDN, tenantUUID)
+	}
+
+	return dtEndpoint, nil
+}
+
+func isEnvVarSet(env []corev1.EnvVar, envVar string) bool {
```
0sewa0: could use https://github.com/Dynatrace/dynatrace-operator/blob/d2ff055c9ab0f08bdb0ff746fe508cc4406efca2/pkg/util/kubeobjects/env/env.go#L27

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
```diff
@@ -15,26 +23,284 @@ func New() dtwebhook.Mutator {
...
+		// metrics exporter env var
+		OTLPMetricsEndpointEnv,
+		OTLPMetricsHeadersEnv,
+		OTLPMetricsCertificateEnv,
+		OTLPMetricsProtocolEnv,
+		// logs exporter env var
+		OTLPLogsEndpointEnv,
+		OTLPLogsHeadersEnv,
+		OTLPLogsCertificateEnv,
+		OTLPLogsProtocolEnv,
```
0sewa0, line 165: would it make sense to abstract this a bit?

like: (just a draft of an idea, don't take it as holy scripture 😅 )
```go
type envInjector interface {
   IsEnabled() bool
   GetEnvs() []string
   Inject(c *corev1.Container) bool
}
```

And an impl of this:
```go

type tracesInjector struct {
    dk dynakube.Dynakube // could be just the OtlpExporterConfig
	apiUrl string
}


func (ti *tracesInjector) IsEnabled() bool {
	return ti.dk.OtlpExporterConfig.IsTracesEnabled()
}

func (ti *tracesInjector) GetEnvs() []string {
	return []string{
		OTLPTraceEndpointEnv,
		OTLPTraceHeadersEnv,
		OTLPTraceCertificateEnv,
		OTLPTraceProtocolEnv,
	}
}

// ...
```

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
```diff
@@ -15,26 +23,284 @@ func New() dtwebhook.Mutator {
...
+		c.Name,
+	) {
+		return true
+	}
+
+	if override {
+		return false
+	}
+
+	envVarsToCheck := []string{
```
0sewa0, line 145: is this correct?

because in the scenario where:

 a container, that has the logs envs, but we don't want to inject the logs envs, only the metrics envs
this would skip the container

so you would need to do something like:

```go
envVarsToCheck := []string{
		// general exporter env var
		OTLPExporterEndpointEnv,
		OTLPExporterHeadersEnv,
		OTLPExporterCertificateEnv,
		OTLPExporterProtocolEnv
}

if otlpExporterConfig.IsTracesEnabled() {
	envVarsToCheck = append(envVarsToCheck, "<the traces envs>")
}

if otlpExporterConfig.IsMetricsEnabled() {
	envVarsToCheck = append(envVarsToCheck, "<the metrics envs>")
}

if otlpExporterConfig.IsLogsEnabled() {
	envVarsToCheck = append(envVarsToCheck, "<the logs envs>")
}
// ...
```

0sewa0: but does it even make sense to check the "feature specific" envs?

you already check those, when trying to inject them.

pkg/api/latest/dynakube/otlpexporterconfiguration/props_test.go
```diff
@@ -24,38 +24,43 @@ import (
 
 func TestSpec_IsOverrideEnabled(t *testing.T) {
 	t.Run("overrideEnvVars not set", func(t *testing.T) {
```
avorima: these tests could be implemented as table-driven tests like

```go
tests := []struct{ name string; spec spec; expect bool }{ ... }
foreach test {
       t.Run(test.name, func(t *testing.T) { assert.Equal(t, test.expect, spec.IsXEnabled()) }
}
```

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
```diff
@@ -15,26 +23,284 @@ func New() dtwebhook.Mutator {
 	return &Mutator{}
 }
 
-func (Mutator) IsEnabled(_ *dtwebhook.BaseRequest) bool {
-	log.Debug("checking if OTLP env var injection is enabled")
+func (m Mutator) IsEnabled(request *dtwebhook.BaseRequest) bool {
+	log.Debug("checking if OTLP env var injection is enabled", "podName", request.PodName(), "namespace", request.Namespace.Name)
```
avorima: if this is moved after the enabled check we get one less log line if it's disabled. the log message should then be changed to mention it being enabled

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
bacherfl: Yes I think this check is needed - the desired behaviour is that if `override` is set to `false`, then the webhook should set no exporter env var, if any of the otlp exporter env vars is already present - also e.g. if we have just enabled traces, whereas the user has provided the env var for metrics. But that being said, I believe the check for the specific env vars in the `inject[*]EnvVars()` functions are not required anymore at this point (earlier the logic was that e.g. traces are added if metrics are already defined by the user, regardless of the override setting, but since we now have the behaviour described above this check is now obsolete)

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
bacherfl: yes I like that idea - will adapt the PR to add that :)

pkg/api/latest/dynakube/otlpexporterconfiguration/props_test.go
bacherfl: makes sense, I will adapt the tests accordingly

pkg/api/latest/dynakube/otlpexporterconfiguration/props.go
```diff
@@ -16,18 +16,26 @@ limitations under the License.
...
-	return s.OverrideEnvVars != nil && *s.OverrideEnvVars
+func (s *OTLPExporterConfiguration) IsEnabled() bool {
+	if s.Spec == nil {
+		return false
+	}
+
+	return s.IsTracesEnabled() || s.IsMetricsEnabled() || s.IsLogsEnabled()
+}
+
+func (s *OTLPExporterConfiguration) IsOverrideEnabled() bool {
```
andriisoldatenko:
```suggestion
func (s *OTLPExporterConfiguration) IsOverrideEnvVarsEnabled() bool {
```
otherwise it's unclear what kind of override is enabled.

pkg/api/latest/dynakube/otlpexporterconfiguration/props_test.go
```diff
@@ -23,76 +23,165 @@ import (
...
+			name: "overrideEnvVars disabled",
+			config: OTLPExporterConfiguration{Spec: &Spec{
+				OverrideEnvVars: &disabled,
+			}},
+			expected: false,
+		},
+		{
+			name: "overrideEnvVars enabled",
+			config: OTLPExporterConfiguration{Spec: &Spec{
+				OverrideEnvVars: &enabled,
```
andriisoldatenko:
```suggestion
				OverrideEnvVars: ptr.To(true),
```

pkg/api/latest/dynakube/otlpexporterconfiguration_props.go
```diff
@@ -0,0 +1,13 @@
+package dynakube
+
+import (
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube/otlpexporterconfiguration"
+)
+
+func (dk *DynaKube) OTLPExporterConfiguration() *otlpexporterconfiguration.OTLPExporterConfiguration {
+	otlpExporterConfiguration := &otlpexporterconfiguration.OTLPExporterConfiguration{
```
andriisoldatenko, line 8: is any reason to duplicate package name and struct name?

can we try something like `otlpexporterconfiguration.Config`?

pkg/webhook/mutation/pod/handler/otlp/handle.go
```diff
@@ -21,18 +20,20 @@ func New(
...
 
-	log.Debug("OTLP injection enabled", "podName", mutationRequest.PodName(), "namespace", mutationRequest.Namespace.Name)
-
 	if h.envVarMutator.IsEnabled(mutationRequest.BaseRequest) {
-		err := h.envVarMutator.Mutate(mutationRequest)
-		if err != nil {
-			return err
+		if h.envVarMutator.IsInjected(mutationRequest.BaseRequest) {
+			if h.envVarMutator.Reinvoke(mutationRequest.ToReinvocationRequest()) {
+				log.Info("reinvocation policy applied", "podName", mutationRequest.PodName())
```
andriisoldatenko: to me it' sounds like debug log level

pkg/webhook/mutation/pod/mutator/otlp/exporter/injectors_test.go
```diff
@@ -0,0 +1,151 @@
...
+			assert.Equal(t, tt.expectInjected, injected)
+
+			for _, envName := range tt.expectEnvVars {
+				assert.True(t, env.IsIn(c.Env, envName), "expected env var %s to be injected", envName)
+			}
+		})
+	}
+}
+
+func TestMetricsInjector_IsEnabledAndInject(t *testing.T) {
```
andriisoldatenko:
```suggestion
func TestMetricsInjectorIsEnabledAndInject(t *testing.T) {
```
we use underscore for testing lowercase methods (aka private ones) and for public you don't need `_`

pkg/api/latest/dynakube/otlpexporterconfiguration_props.go
bacherfl: Here I tried to use the same style as for the other config struct, e.g. like the [MetadataEnrichment](https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/api/latest/dynakube/metadataenrichment/spec.go#L17) - but I agree that the naming is a bit clunky here, so we can also rename this to what you suggested

andriisoldatenko: @0sewa0 wdyt?

0sewa0: so, these `otlpexporterconfiguration.OTLPExporterConfiguration` are just "wrappers" for the spec/status/templates sections for a given part of the CR.

naming wise, I don't really know what I would call them.
I agree that `otlpexporterconfiguration.OTLPExporterConfiguration` and `metadataenrichment.MetadataEnrichment` are a mouthful. 😅

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
bacherfl: adapted with [d4aa721](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/d4aa721050dac55621b1895f4311d2b9a6a59179)

pkg/api/latest/dynakube/otlpexporterconfiguration/props_test.go
bacherfl: adapted with [d4aa721](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/d4aa721050dac55621b1895f4311d2b9a6a59179)

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
bacherfl: adapted with [d4aa721](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/d4aa721050dac55621b1895f4311d2b9a6a59179)

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
bacherfl: adapted with [d4aa721](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/d4aa721050dac55621b1895f4311d2b9a6a59179)

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
bacherfl: adapted with [d4aa721](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/d4aa721050dac55621b1895f4311d2b9a6a59179)

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
bacherfl: added with [5ef0da3](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/5ef0da3a31fcee47f8dc0441be9ebf1f51b04023)

pkg/api/latest/dynakube/otlpexporterconfiguration_props.go
0sewa0: `otlpexporterconfiguration.Config` and `metadataenrichment.Config` works for me, but make sure we remain consistent :D (upated styleguide 👀 )

pkg/webhook/mutation/pod/mutator/otlp/exporter/injectors_test.go
bacherfl: adapted with [e8fef18](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/e8fef18703729106aa1310bc5a1989924d575786)

pkg/webhook/mutation/pod/handler/otlp/handle.go
bacherfl: adapted with [e8fef18](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/e8fef18703729106aa1310bc5a1989924d575786)

pkg/api/latest/dynakube/otlpexporterconfiguration/props_test.go
bacherfl: adapted with [e8fef18](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/e8fef18703729106aa1310bc5a1989924d575786)

pkg/api/latest/dynakube/otlpexporterconfiguration/props.go
bacherfl: adapted with [e8fef18](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/e8fef18703729106aa1310bc5a1989924d575786)

pkg/api/latest/dynakube/otlpexporterconfiguration_props.go
bacherfl: in this case, should we keep the current naming for now, as this is currently done in pretty much all sub packages of the Dynakube API? Otherwise, this PR would need to change all those as well if we want to keep consistent

0sewa0: yes, @andriisoldatenko can you create a little techdept ticket to fix this? 👀

andriisoldatenko: on it 👮

andriisoldatenko: https://dt-rnd.atlassian.net/browse/DAQ-14905

pkg/webhook/mutation/pod/mutator/otlp/exporter/config.go
```diff
@@ -3,9 +3,11 @@ package exporter
 const (
 	AnnotationPrefix = "otlp-exporter-configuration"
 	// AnnotationOneAgentInject can be set at pod level to enable/disable OneAgent injection.
-	AnnotationInject   = AnnotationPrefix + ".dynatrace.com/inject"
-	AnnotationInjected = AnnotationPrefix + ".dynatrace.com/injected"
-	AnnotationReason   = AnnotationPrefix + ".dynatrace.com/reason"
+	AnnotationInject             = AnnotationPrefix + ".dynatrace.com/inject"
+	AnnotationInjected           = AnnotationPrefix + ".dynatrace.com/injected"
+	AnnotationReason             = AnnotationPrefix + ".dynatrace.com/reason"
+	AnnotationInjectedTrueValue  = "true"
```
avorima: hm, removing this doesn't give me any issues locally. also the linter complains about 5 uses but these are only used once so they're essentially redundant

bacherfl: yes that was a bit confusing, but in the CI checks the linter complained about this for some reason. But as they each are only used once, I agree that these should not be necessary. I will remove them again and see what happens

bacherfl: after removing the consts and using the literal values directly the linter was complaining again, but [884e0f4](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/884e0f4d2c7b61c8c8ebfd2ed8386f827d0f0ceb) seems to have solved the issue

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
```diff
@@ -15,26 +25,190 @@ func New() dtwebhook.Mutator {
...
+
+	// first, check if otlp injection is enabled explicitly on pod
+	enabledOnPod := maputils.GetFieldBool(request.Pod.Annotations, AnnotationInject, false)
+
+	if !enabledOnPod {
+		// if not enabled explicitly, check general injection setting via 'dynatrace.com/inject' annotation
+		enabledOnPod = maputils.GetFieldBool(request.Pod.Annotations, dtwebhook.AnnotationDynatraceInject, request.DynaKube.FF().IsAutomaticInjection())
+	}
+
+	namespaceEnabled := true
```
waodim:
```suggestion
	namespaceSelectorEnabled := true
```

waodim: or maybe `enabledOnNamespace`?

Really a minor nitpick, just for consistency maybe worth thinking about renaming

bacherfl: adapted with [381bc3a](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/381bc3a8e59cf7c72f5763e5cb335a03458c4004)

pkg/webhook/mutation/pod/mutator/otlp/exporter/config.go
avorima: sorry, to be insistent here but this strconv still doesn't sit well with me. i can't reproduce the failure even with a completely fresh install of golangci-lint with the same goconst version. could you add `true` and `false` to the ignored string values in [linter config](https://golangci-lint.run/docs/linters/configuration/#goconst)?

bacherfl: no worries - i reverted the change now and adapted the linter config ([7a7d37c](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/7a7d37cab1a8eb96e1e605a6671448d3bb07b038))

.golangci.yml
```diff
@@ -64,6 +64,10 @@ linters:
     - wsl_v5
     - zerologlint
   settings:
+    goconst:
+      ignore-string-values:
+        - "true"
+        - "false"
```
Incredible-O, line 70: Are you sure that change is still necessary? I've reverted it locally and `make go/lint` still works for me 🤔

bacherfl: locally it was working for me as well without this change, but for some reason the lint check in the CI pipeline fails without this setting

bacherfl: to be sure, I just tried to undo the linter config change again, but got the same issue: https://github.com/Dynatrace/dynatrace-operator/actions/runs/18458973490/job/52585823900?pr=5474

Incredible-O: I see, thanks for checking!

That's probably the reason:
```go: downloading github.com/jgautheron/goconst v1.8.2```

We should discuss/address this at some point; I'll create a follow-up ticket for that 😄

File: pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go [hunk 17]
```diff
@@ -15,26 +24,190 @@ func New() dtwebhook.Mutator {
...
+		OTLPLogsCertificateEnv,
+		OTLPLogsProtocolEnv,
+	}
+
+	for _, envVar := range envVarsToCheck {
+		if env.IsIn(c.Env, envVar) {
+			return true
+		}
+	}
+
```
0sewa0, line 173: Just an idea: (ignore if not to your liking)

possible (future) improvement

```go
for _, injector := range injectors {
	for _, envVar := range injector.GetEnvs() {
		if env.IsIn(c.Env, envVar) {
			return true
		}
	}
}
```

you would need to define a `general` injector, that is only used in this check, that the user can't make us actually inject.

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
```diff
@@ -15,26 +24,190 @@ func New() dtwebhook.Mutator {
...
+func getIngestEndpoint(dk *dynakube.DynaKube) (string, error) {
+	dtEndpoint := dk.APIURL() + "/v2/otlp"
+
+	if dk.ActiveGate().IsEnabled() {
+		tenantUUID, err := dk.TenantUUID()
+		if err != nil {
+			return "", err
+		}
+
+		serviceFQDN := capability.BuildServiceName(dk.Name) + "." + dk.Namespace + ".svc"
```
0sewa0: could be a helper func in the `capability` package, as it is [already used elsewhere](https://github.com/Dynatrace/dynatrace-operator/blob/ed3fbb740a600aea938bc1ddf399fbd3402eacd5/pkg/controllers/dynakube/otelc/endpoint/reconciler.go#L110) 

```suggestion
		serviceFQDN := capability.BuildServiceFQDN(dk)
```

0sewa0: looking more at it, it is just a straight copy of the whole func `generateData`. (you just don't put it into a map)

wouldn't it make sense to have that logic extracted into its own func so it can be reused?

like `BuildOtlpEndpoint` or something

bacherfl: that makes sense - any preferences on where to put this func?

0sewa0: it could be in that `otelc/endpoint` package, as the name fits the most

hopefully no import cycle will happen 😅

bacherfl: adapted with [34c4330](https://github.com/Dynatrace/dynatrace-operator/pull/5474/commits/34c433080e63a7724b857eb69d3356da06bc35c5)

PR #5475, opened 2025-09-30
Remove unnecessary quotes from make commands

albertogdd:
## Description

https://dt-rnd.atlassian.net/browse/DAQ-14710

I keep seeing double quotes on E2E executions, I believe we don't need these

e.g. (check `--set image`)
```
helm upgrade dynatrace-operator oci://quay.io/dynatrace/dynatrace-operator:0.0.0-nightly-chart \
		--install \
		--namespace dynatrace \
		--create-namespace \
		--atomic \
		--set installCRD=true \
		--set csidriver.enabled=true \
		--set manifests=true \
		--set image=""quay.io/dynatrace/dynatrace-operator:snapshot"" \
		--set debugLogs=true \
		--set debug=false
```

## How can this be tested?

`make deploy`
`make manifests`

### DIFF
--- a/hack/make/manifests/kubernetes.mk
+++ b/hack/make/manifests/kubernetes.mk
@@ -5,7 +5,7 @@ define generate_k8s_manifest
 		--set installCRD=true \
 		--set platform="kubernetes" \
 		--set manifests=true \
-		--set olm=${OLM} \
+		--set olm=$(OLM) \
 		--set image=$(IMAGE_URI) > $(2)
 endef
 
--- a/hack/make/manifests/openshift.mk
+++ b/hack/make/manifests/openshift.mk
@@ -5,7 +5,7 @@ define generate_openshift_manifest
 		--set installCRD=true \
 		--set platform="openshift" \
 		--set manifests=true \
-		--set olm=${OLM} \
+		--set olm=$(OLM) \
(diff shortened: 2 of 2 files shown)


hack/make/manifests/kubernetes.mk
```diff
@@ -5,8 +5,8 @@ define generate_k8s_manifest
 		--set installCRD=true \
 		--set platform="kubernetes" \
 		--set manifests=true \
-		--set olm="${OLM}" \
-		--set image="$(IMAGE_URI)" > $(2)
+		--set olm=${OLM} \
```
avorima: `${}` and `$()` are equivalent to my knowledge, so let's use only one form to be more consistent

andriisoldatenko: I think @avorima is right so please use one style.

albertogdd: [Fixed](https://github.com/Dynatrace/dynatrace-operator/pull/5475/commits/39c275c62b5e03fd0cdc39b48e454cb96acdfe3c)

andriisoldatenko (approved):

albertogdd:
Tested edge case when branch contains `/` to see if make deploy fails without double quotes. Image name transforms / to - so it works well

image

PR #5484, opened 2025-10-02
Store nsSelector matching namespaces in conditions

waodim:
<!--

Thanks for opening a pull request! Here are some tips to get everything merged smoothly:

1. Read our contributor guidelines: https://github.com/Dynatrace/dynatrace-operator/blob/main/CONTRIBUTING.md

2. If the PR is unfinished, raise it as a draft or prefix the title with "WIP:" so it's clear to everyone.

3. Be sure to allow edits from maintainers, so it's easier for us to help: https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

-->

## Description

[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-11997)

With this we want to improve trouble shooting. This introduces a check of which namespace matches the respective namespace selector (split by metadataEnrichment and oneAgent ns selector!)

We store all namespaces into the dk status and list at most 10 namespaces in order to avoid unreadable long lists. We also log how many namespaces are matching (again up to 10 at most)

<!--

Please include the following:
- The motivation for the change
    - Link to the Github issue or Jira ticket, if exists.
- The summary of the change

-->

## How can this be tested?

Test out a DK with both namespaces selectors enabled (metadataEnrichment and oneAgent)

Try out different scenarios:
- a lot of namespaces match the metadataEnrichment namespaceSelector
- a lot of namespaces match the oneAgent namespaceSelector
- a lot of namespaces match both
- little to no namespace match either of them or both

just play around and see how we react to it.

sample logs:

```
{"level":"info","ts":"2025-10-02T13:39:22.247Z","logger":"dynakube-injection","msg":"namespaces monitored","selector":"OneAgent","count (at most 10 are shown)":10,"namespaces":["bh-day-ns10","bh-day-ns11","bh-day-ns2","bh-day-ns3","bh-day-ns4","bh-day-ns5","bh-day-ns6","
h-day-ns7","bh-day-ns8","bh-day-ns9"[]}                                                                                                                                                                                                                                        
{"level":"info","ts":"2025-10-02T13:39:22.247Z","logger":"dynakube-injection","msg":"namespaces monitored","selector":"MetadataEnrichment","count (at most 10 are shown)":2,"namespaces":["bh-day-ns","default"]} 
```

sample status:

```
    Last Transition Time:  2025-10-02T12:10:52Z                                                                                                                                                                                                                                
    Message:               10 namespaces match: bh-day-ns10, bh-day-ns11, bh-day-ns2, bh-day-ns3, bh-day-ns4, bh-day-ns5, bh-day-ns6, bh-day-ns7, bh-day-ns8, bh-day-ns9                                                                                                       
    Reason:                MatchesFound                                                                                                                                                                                                                                        
    Status:                True                                                                                                                                                                                                                                                
    Type:                  OneAgentNamespacesMonitored                                                                                                                                                                                                                         
    Last Transition Time:  2025-10-02T12:10:52Z                                                                                                                                                                                                                                
    Message:               2 namespaces match: bh-day-ns, default                                                                                                                                                                                                              
    Reason:                MatchesFound                                                                                                                                                                                                                                        
    Status:                True                                                                                                                                                                                                                                                
    Type:                  MetadataEnrichmentNamespacesMonitored                                                                                                                                                                                                               
    Last Transition Time:  2025-10-02T12:09:06Z                                                                                                                                                                                                                                
    Message:               No namespaces match the configured selectors                                                                                                                                                                                                        
    Reason:                NoMatches                                                                                                                                                                                                                                           
    Status:                False                                                                                                                                                                                                                                               
    Type:                  NamespacesMonitored                                                                                                                                                                                                                                 
    Last Transition Time:  2025-10-02T13:39:21Z  
```

<!--

-->

### DIFF
--- a/pkg/controllers/dynakube/injection/conditions.go
+++ b/pkg/controllers/dynakube/injection/conditions.go
@@ -1,17 +1,31 @@
 package injection
 
 import (
+	"fmt"
+	"strings"
+
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
+	namespacesMonitoredConditionType = "NamespacesMonitored"
+
+	oneAgentNamespacesMonitoredConditionType           = "OneAgentNamespacesMonitored"
+	metadataEnrichmentNamespacesMonitoredConditionType = "MetadataEnrichmentNamespacesMonitored"
+
 	metaDataEnrichmentConditionType   = "MetadataEnrichment"
 	codeModulesInjectionConditionType = "CodeModulesInjection"
+
+	matchesFoundReason          = "MatchesFound"
+	noMatchesReason             = "NoMatches"
+	selectorNotConfiguredReason = "SelectorNotConfigured"
 
 	secretsCreatedReason  = "SecretsCreated"
 	secretsCreatedMessage = "Namespaces mapped and secrets created"
 )
+
+const maxNamesInMsg = 10
 
 func setMetadataEnrichmentCreatedCondition(conditions *[]metav1.Condition) {
 	condition := metav1.Condition{
@@ -31,4 +45,83 @@ func setCodeModulesInjectionCreatedCondition(conditions *[]metav1.Condition) {
 		Message: secretsCreatedMessage,
 	}
 	_ = meta.SetStatusCondition(conditions, condition)
+}
+
+func setNamespacesMonitoredSelectorCondition(conditions *[]metav1.Condition, selectorType string, configured bool, names []string) {
+	var condType string
+
+	switch selectorType {
+	case "OneAgent":
+		condType = oneAgentNamespacesMonitoredConditionType
+	case "MetadataEnrichment":
+		condType = metadataEnrichmentNamespacesMonitoredConditionType
+	default:
+		condType = namespacesMonitoredConditionType
+	}
+
+	cond := metav1.Condition{Type: condType}
+
+	switch {
+	case !configured:
+		cond.Status = metav1.ConditionFalse
+		cond.Reason = selectorNotConfiguredReason
+		cond.Message = "Selector not configured"
+	case len(names) == 0:
+		cond.Status = metav1.ConditionFalse
+		cond.Reason = noMatchesReason
+		cond.Message = "0 namespaces match"
+	default:
+		cond.Status = metav1.ConditionTrue
+		cond.Reason = matchesFoundReason
+		msg := formatMatchMessage(names, maxNamesInMsg)
+		cond.Message = msg
+	}
+
+	cond.LastTransitionTime = metav1.Now()
+	_ = meta.SetStatusCondition(conditions, cond)
+}
+
+func updateCollectedNamespacesMonitoredCondition(conditions *[]metav1.Condition) {
+	oa := meta.FindStatusCondition(*conditions, oneAgentNamespacesMonitoredConditionType)
+	me := meta.FindStatusCondition(*conditions, metadataEnrichmentNamespacesMonitoredConditionType)
+
+	collected := metav1.Condition{
+		Type:               namespacesMonitoredConditionType,
+		LastTransitionTime: metav1.Now(),
+	}
+
+	if oa != nil && oa.Status == metav1.ConditionTrue || me != nil && me.Status == metav1.ConditionTrue {
+		collected.Status = metav1.ConditionFalse
+		collected.Reason = noMatchesReason
+		collected.Message = "No namespaces match the configured selectors"
+		_ = meta.SetStatusCondition(conditions, collected)
+
+		return
+	}
+
+	if oa != nil && oa.Reason == selectorNotConfiguredReason && me != nil && me.Reason == selectorNotConfiguredReason {
+		collected.Status = metav1.ConditionFalse
+		collected.Reason = selectorNotConfiguredReason
+		collected.Message = "No selectors configured"
+		_ = meta.SetStatusCondition(conditions, collected)
+
+		return
+	}
+
+	collected.Status = metav1.ConditionFalse
+	collected.Reason = noMatchesReason
+	collected.Message = "No namespaces match the configured selectors"
+	meta.SetStatusCondition(conditions, collected)
+}
+
+func formatMatchMessage(names []string, limit int) string {
+	if len(names) == 0 {
+		return "no namespaces match"
+	}
+
+	if len(names) > limit {
+		return fmt.Sprintf("%d namespaces match: %s (at most %d shown)", len(names), strings.Join(names[:limit], ", "), limit)
+	}
+
+	return fmt.Sprintf("%d namespaces match: %s", len(names), strings.Join(names, ", "))
 }
--- a/pkg/controllers/dynakube/injection/reconciler.go
+++ b/pkg/controllers/dynakube/injection/reconciler.go
@@ -92,7 +92,22 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 			setupErrors = append(setupErrors, err)
 		}
 
-		err := r.generateInitSecret(ctx)
+		oneAgents, metadataEnrichment, err := mapper.ListMonitoredNamespaces(ctx, r.apiReader, r.dk)
+
+		log.Info("namespaces monitored", "selector", "OneAgent", "count (at most 10 are shown)", len(oneAgents), "namespaces", oneAgents)
+
+		log.Info("namespaces monitored", "selector", "MetadataEnrichment", "count (at most 10 are shown)", len(metadataEnrichment), "namespaces", metadataEnrichment)
+
+		setNamespacesMonitoredSelectorCondition(r.dk.Conditions(), "OneAgent", r.dk.OneAgent().IsAppInjectionNeeded(), oneAgents)
+		setNamespacesMonitoredSelectorCondition(r.dk.Conditions(), "MetadataEnrichment", r.dk.MetadataEnrichment().IsEnabled(), metadataEnrichment)
+
+		updateCollectedNamespacesMonitoredCondition(r.dk.Conditions())
+
+		if err != nil {
+			log.Error(err, "listing monitored namespaces failed")
+		}
+
+		err = r.generateInitSecret(ctx)
 		if err != nil {
 			setupErrors = append(setupErrors, err)
 		}
--- a/pkg/injection/namespace/mapper/mapper.go
+++ b/pkg/injection/namespace/mapper/mapper.go
@@ -3,6 +3,7 @@ package mapper
 import (
 	"context"
 	"regexp"
+	"sort"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
 	dtwebhook "github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/mutator"
@@ -174,4 +175,35 @@ func isIgnoredNamespace(dk *dynakube.DynaKube, namespaceName string) bool {
 	}
 
 	return false
+}
+
+func ListMonitoredNamespaces(ctx context.Context, clt client.Reader, dk *dynakube.DynaKube) (oneAgent []string, metaEnrich []string, _ error) {
+	nsList := &corev1.NamespaceList{}
+	if err := clt.List(ctx, nsList); err != nil {
+		return nil, nil, err
+	}
+
+	for i := range nsList.Items {
+		ns := &nsList.Items[i]
+		if isIgnoredNamespace(dk, ns.Name) {
+			continue
+		}
+
+		if ok, err := matchOneAgent(dk, ns); err != nil {
+			return nil, nil, err
+		} else if ok {
+			oneAgent = append(oneAgent, ns.Name)
+		}
+
+		if ok, err := matchMetadataEnrichment(dk, ns); err != nil {
+			return nil, nil, err
+		} else if ok {
+			metaEnrich = append(metaEnrich, ns.Name)
+		}
+	}
+
+	sort.Strings(oneAgent)
+	sort.Strings(metaEnrich)
+
+	return oneAgent, metaEnrich, nil
 }
--- a/pkg/injection/namespace/mapper/mapper_test.go
+++ b/pkg/injection/namespace/mapper/mapper_test.go
@@ -1,6 +1,7 @@
 package mapper
 
 import (
+	"context"
 	"testing"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/exp"
@@ -12,6 +13,9 @@ import (
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
 func createBaseDynakube(name string, appInjection bool, metadataEnrichment bool) *dynakube.DynaKube {
@@ -244,4 +248,56 @@ func TestUpdateNamespace(t *testing.T) {
 		require.True(t, updated)
 		assert.Len(t, namespace.Labels, 1)
 	})
+}
+
+func TestListMonitoredNamespaces_BasicMatching(t *testing.T) {
+	ctx := context.Background()
+
+	scheme := runtime.NewScheme()
+	require.NoError(t, corev1.AddToScheme(scheme))
+
+	nsObjs := []client.Object{
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "ns-a", Labels: map[string]string{"team": "core"}}},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "ns-b", Labels: map[string]string{"team": "edge"}}},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "ns-c"}},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "ns-d", Labels: map[string]string{"team": "x"}}},
+	}
+
+	cl := fake.NewClientBuilder().WithScheme(scheme).WithObjects(nsObjs...).Build()
+
+	oaSel := metav1.LabelSelector{
+		MatchExpressions: []metav1.LabelSelectorRequirement{
+			{Key: "team", Operator: metav1.LabelSelectorOpIn, Values: []string{"core", "edge"}},
+		},
+	}
+
+	meSel := metav1.LabelSelector{
+		MatchLabels: map[string]string{"team": "core"},
+	}
+
+	dk := createBaseDynakube("dk", true, true)
+	dk.Spec.OneAgent.ApplicationMonitoring.NamespaceSelector = oaSel
+	dk.Spec.MetadataEnrichment.NamespaceSelector = meSel
+	oa, me, err := ListMonitoredNamespaces(ctx, cl, dk)
+	require.NoError(t, err)
+
+	require.Equal(t, []string{"ns-a", "ns-b"}, oa, "OneAgent set mismatch")
+	require.Equal(t, []string{"ns-a"}, me, "MetadataEnrichment set mismatch")
+}
+
+func TestListMonitoredNamespaces_MatchAllAndNotConfigured(t *testing.T) {
+	ctx := context.Background()
+	scheme := runtime.NewScheme()
+	require.NoError(t, corev1.AddToScheme(scheme))
+	nsObjs := []client.Object{
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "x"}},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "y"}},
+	}
+	cl := fake.NewClientBuilder().WithScheme(scheme).WithObjects(nsObjs...).Build()
+
+	dk := createBaseDynakube("dk", true, false)
+	oa, me, err := ListMonitoredNamespaces(ctx, cl, dk)
+	require.NoError(t, err)
+	require.Equal(t, []string{"x", "y"}, oa, "OA should match all namespaces in alphabetical order")
+	require.Empty(t, me, "ME disabled should yield zero namespaces")
 }


pkg/controllers/dynakube/injection/conditions.go
```diff
@@ -32,3 +46,82 @@ func setCodeModulesInjectionCreatedCondition(conditions *[]metav1.Condition) {
...
+func setNamespacesMonitoredSelectorCondition(conditions *[]metav1.Condition, selectorType string, configured bool, names []string) {
+	var condType string
+
+	switch selectorType {
+	case "OneAgent":
+		condType = oneAgentNamespacesMonitoredConditionType
+	case "MetadataEnrichment":
+		condType = metadataEnrichmentNamespacesMonitoredConditionType
+	default:
+		condType = namespacesMonitoredConditionType
```
0sewa0: what is the purpose of this?

currently it will always be 0, because `default` only runs when non of the cases match

pkg/controllers/dynakube/injection/conditions.go
```diff
@@ -32,3 +46,82 @@ func setCodeModulesInjectionCreatedCondition(conditions *[]metav1.Condition) {
...
+	}
+
+	cond := metav1.Condition{Type: condType}
+
+	switch {
+	case !configured:
+		cond.Status = metav1.ConditionFalse
+		cond.Reason = selectorNotConfiguredReason
+		cond.Message = "Selector not configured"
```
0sewa0: not configured -> do not have it in the status (so do a remove)

pkg/controllers/dynakube/injection/conditions.go
```diff
@@ -1,18 +1,32 @@
...
 
+	matchesFoundReason          = "MatchesFound"
+	noMatchesReason             = "NoMatches"
+	selectorNotConfiguredReason = "SelectorNotConfigured"
+
 	secretsCreatedReason  = "SecretsCreated"
 	secretsCreatedMessage = "Namespaces mapped and secrets created"
 )
 
+const maxNamesInMsg = 10
```
0sewa0: why separate `const` section?

pkg/injection/namespace/mapper/mapper.go
```diff
@@ -175,3 +176,34 @@ func isIgnoredNamespace(dk *dynakube.DynaKube, namespaceName string) bool {
 
 	return false
 }
+
+func ListMonitoredNamespaces(ctx context.Context, clt client.Reader, dk *dynakube.DynaKube) (oneAgent []string, metaEnrich []string, _ error) {
```
0sewa0: so, here we are kind doing "double work", or introducing it atleast.

We already request the namespaces: https://github.com/Dynatrace/dynatrace-operator/blob/b57c523e610621b5cdef7cc4538ebbf98e6b57a7/pkg/injection/namespace/mapper/dynakubes.go#L48
- (as part of the mapping)

we also do this matching here https://github.com/Dynatrace/dynatrace-operator/blob/b57c523e610621b5cdef7cc4538ebbf98e6b57a7/pkg/injection/namespace/mapper/mapper.go#L66

SO, IMO we can either:
a) improve the mapper so we do not do double the work
b) do it how it is done now, because the "long term" plan is to remove this `mapper` because we don't want to need have namespace webhook and namespace labeling (due to permissions)

pkg/controllers/dynakube/injection/conditions.go
waodim: I thought about it being a general info to inform if namespaces match configured selectors, what do you mean by it always being 0? 

EDIT: I see it now I will remove it 😅

pkg/injection/namespace/mapper/mapper.go
waodim: done - check the commit and the message. I had to do some changes like alter some recevier methods of the DynakubeMapper to have a pointer because otherwise the newly introduced fields where I cache the namespaces that match are not stored properly. (call by value/call by reference issue it seems)

https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/71ff10790fe9f306564ef5d35b61c486002bf1f9

pkg/controllers/dynakube/injection/conditions.go
waodim: brainlag https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/71ff10790fe9f306564ef5d35b61c486002bf1f9

pkg/controllers/dynakube/injection/conditions.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/71ff10790fe9f306564ef5d35b61c486002bf1f9

pkg/controllers/dynakube/injection/conditions.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/71ff10790fe9f306564ef5d35b61c486002bf1f9

pkg/injection/namespace/mapper/mapper.go
```diff
@@ -62,22 +62,31 @@ func setUpdatedViaDynakubeAnnotation(ns *corev1.Namespace) {
 	ns.Annotations[UpdatedViaDynakubeAnnotation] = "true"
 }
 
-func match(dk *dynakube.DynaKube, namespace *corev1.Namespace) (bool, error) {
+func matchKinds(dk *dynakube.DynaKube, namespace *corev1.Namespace) (bool, bool, error) {
```
0sewa0: oh no, `(bool, bool, error)` is rather bad practice
if you do that you should at least have the var names in there `(matchOA bool, matchME bool, error)`
but I am still not a fan

why do you have to move it out of `match` ?

pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -20,23 +21,28 @@ type DynakubeMapper struct {
...
-		operatorNs: operatorNs,
-		dk:         dk,
-		secrets:    k8ssecret.Query(clt, apiReader, log),
+		ctx:         ctx,
+		client:      clt,
+		apiReader:   apiReader,
+		operatorNs:  operatorNs,
+		dk:          dk,
+		secrets:     k8ssecret.Query(clt, apiReader, log),
+		lastOANames: []string{},
```
0sewa0: "Names" is misleading, I know that it is referring to "Namespaces", but still

there is `operatorNs` which also does it differently. (I know that calling it `lastOANss` is just a no go 😅 )

"Last" of what? (the "previously matched namespaces")

why not just `matchedOANamespaces`

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -92,6 +92,25 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 			setupErrors = append(setupErrors, err)
 		}
 
+		oneAgentNamespaceNames := dkMapper.OneAgentNamespaceNames()
+		metadataEnrichmentNamespaceNames := dkMapper.MetadataEnrichmentNamespaceNames()
+
+		oaSelectorConfigured := r.dk.OneAgent().IsAppInjectionNeeded() && (r.dk.OneAgent().IsNamespaceSelectorDefined())
```
0sewa0: selector not defined == "everything" monitored

why do we need the `IsNamespaceSelectorDefined` ?

pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -131,3 +160,10 @@ func (dm DynakubeMapper) updateNamespaces(modifiedNs []*corev1.Namespace) error
 
 	return nil
 }
+
+func (dm *DynakubeMapper) OneAgentNamespaceNames() []string {
+	return dm.lastOANames
+}
+func (dm *DynakubeMapper) MetadataEnrichmentNamespaceNames() []string {
+	return dm.lastMENames
+}
```
0sewa0: why do we do it "outside" in a different reconciler, and why not as part of the `MapFromDynakube`?

and also clean it up in the `UnmapFromDynakube`?

pkg/controllers/dynakube/injection/reconciler.go
waodim: Because for if I do not also check if the namespace selector is configured we set the condition with all namespaces for both selectors. Do we really want that?

pkg/injection/namespace/mapper/dynakubes.go
waodim: on it

pkg/injection/namespace/mapper/mapper.go
waodim: I think in my world this made sense but actually you are right this seems like a unnecessary change, I will revert it again

pkg/injection/namespace/mapper/dynakubes.go
waodim: I placed that here such that we can call it in the injection reconciler in order to have them already grouped correctly. Is there a better way?

pkg/injection/namespace/mapper/dynakubes.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/17b181578162d04a9b7a2f41041f275f5be2c46b

pkg/injection/namespace/mapper/mapper.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/17b181578162d04a9b7a2f41041f275f5be2c46b

pkg/injection/namespace/mapper/dynakubes.go
0sewa0: but the injection reconciler will call the `MapFromDynakube`/`UnmapFromDynakube` methods, no?

pkg/controllers/dynakube/injection/reconciler.go
0sewa0: the "status" of the DynaKube is that "all namespaces" are monitored/enriched, I would not hide that.

right now you remove the condition:
```go
case !configured:
		_ = meta.RemoveStatusCondition(conditions, condType)

		return
```

IMO the condition should be removed IF the `namespaceSelector` has no effect (if no OA/Meta configured in general)


if we wish to overcomplicate it, and have a separate `Reason` just for "MatchAll" we can, but IMO it is not that necessary. 

wdyt @chrismuellner ?

waodim: I agree with a condition that all namespaces are monitored tbh. Lemme look into it 👍

pkg/injection/namespace/mapper/dynakubes.go
waodim: you're right, tried my best: https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/8608e32c8efa12a5de92d8f007d747762aac93cd

pkg/controllers/dynakube/injection/reconciler.go
waodim: Done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/8608e32c8efa12a5de92d8f007d747762aac93cd

pkg/controllers/dynakube/injection/conditions.go
```diff
@@ -32,3 +46,80 @@ func setCodeModulesInjectionCreatedCondition(conditions *[]metav1.Condition) {
 	}
 	_ = meta.SetStatusCondition(conditions, condition)
 }
+
+func setNamespacesMonitoredSelectorCondition(conditions *[]metav1.Condition, selectorType string, configured bool, names []string) {
```
0sewa0: why is this not done in the mapper?

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -86,13 +86,31 @@ func (r *Reconciler) Reconcile(ctx context.Context) error {
 		defer r.cleanup(ctx)
 	} else {
 		dkMapper := r.createDynakubeMapper(ctx)
-		if err := dkMapper.MapFromDynakube(); err != nil {
+
+		oaNames, meNames, err := dkMapper.MapFromDynakube()
```
0sewa0: why is this not done in the mapper,

here all you would need to say is `err := dkMapper.MapFromDynakube()`

all the rest with the conditions should be in the mapper, as it should "own" those conditions

pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -79,15 +94,17 @@ func (dm DynakubeMapper) UnmapFromDynaKube(namespaces []corev1.Namespace) error
 		}
 	}
 
+	dm.matchedMENamespaces = []string{}
+	dm.matchedOANamespaces = []string{}
```
0sewa0: does this do anything?

just remove the conditions, `UnmapFromDynaKube` is only called during cleanup, when no namespaces need to be monitored, due the features being turned of

pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -79,15 +94,17 @@ func (dm DynakubeMapper) UnmapFromDynaKube(namespaces []corev1.Namespace) error
...
 	return nil
 }
 
-func (dm DynakubeMapper) mapFromDynakube(nsList *corev1.NamespaceList, dkList *dynakube.DynaKubeList) ([]*corev1.Namespace, error) {
-	var updated bool
-
-	var err error
-
-	var modifiedNs []*corev1.Namespace
+func (dm *DynakubeMapper) mapFromDynakube(nsList *corev1.NamespaceList, dkList *dynakube.DynaKubeList) ([]*corev1.Namespace, matchedNamespaces, error) {
```
0sewa0: why return `matchedNamespaces`?

you have it as part of your `DynakubeMapper` struct, just set it there, and use it later (in `MapFromDynakube`)

pkg/controllers/dynakube/injection/conditions.go
```diff
@@ -32,3 +46,80 @@ func setCodeModulesInjectionCreatedCondition(conditions *[]metav1.Condition) {
...
+	}
+
+	if oa != nil && oa.Reason == selectorNotConfiguredReason && me != nil && me.Reason == selectorNotConfiguredReason {
+		collected.Status = metav1.ConditionFalse
+		collected.Reason = selectorNotConfiguredReason
+		collected.Message = "No selectors configured"
+		_ = meta.SetStatusCondition(conditions, collected)
+
+		return
+	}
```
0sewa0: This being a `false` status gives me the impression that "nothing matches" the selector, but the opposite is true.

I would be for to just not have this case, just have "we have matches, look at them" and "we dont have matches, something maybe off".

It would simplify the logic a bunch.

0sewa0 (changes requested):

pkg/controllers/dynakube/injection/conditions.go
waodim: agree totally. done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/f3111d79d5e3ffe94af64b01027b8ac1f6ef5ce3

pkg/injection/namespace/mapper/dynakubes.go
waodim: Oh I oversaw that, lemme do that quickly, you are right, when moving it here this makes more sense

pkg/injection/namespace/mapper/dynakubes.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/f3111d79d5e3ffe94af64b01027b8ac1f6ef5ce3

pkg/controllers/dynakube/injection/reconciler.go
waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/f3111d79d5e3ffe94af64b01027b8ac1f6ef5ce3

pkg/controllers/dynakube/injection/conditions.go
waodim: I tried to group it at the beginning of the story - but now I see this does not make any sense https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/f3111d79d5e3ffe94af64b01027b8ac1f6ef5ce3

pkg/injection/namespace/mapper/dynakubes.go
waodim: done now https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/5f272cd3fd2e7b8a53c65d616de31eee59dd4db2

pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -20,32 +22,73 @@ type DynakubeMapper struct {
...
+		matchedOANamespaces: []string{},
+		matchedMENamespaces: []string{},
 	}
 }
 
 // MapFromDynakube checks all the namespaces to all the dynakubes
 // updates the labels on the namespaces if necessary,
 // finds conflicting dynakubes (2 dynakube with codeModules on the same namespace)
-func (dm DynakubeMapper) MapFromDynakube() error {
-	modifiedNs, err := dm.MatchingNamespaces()
```
0sewa0, line 40: why?

now it is a copy-paste of what `dm.MatchingNamespaces()` does

pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -107,7 +149,21 @@ func (dm DynakubeMapper) mapFromDynakube(nsList *corev1.NamespaceList, dkList *d
...
+			} else if ok {
+				dm.matchedOANamespaces = append(dm.matchedOANamespaces, namespace.Name)
+			}
+
+			if ok, err := matchMetadataEnrichment(dm.dk, namespace); err != nil {
+				return nil, err
+			} else if ok {
+				dm.matchedMENamespaces = append(dm.matchedMENamespaces, namespace.Name)
+			}
+		}
```
0sewa0: why have this and the new `match` where you already check, if something matches one or the other?

pkg/api/latest/dynakube/metadataenrichment/props.go
```diff
@@ -22,6 +22,14 @@ func (enrichment *MetadataEnrichment) IsEnabled() bool {
 	return enrichment.Enabled != nil && *enrichment.Enabled
 }
 
-func (enrichment *MetadataEnrichment) GetNamespaceSelector() *v1.LabelSelector {
+func (enrichment *MetadataEnrichment) GetNamespaceSelector() *metav1.LabelSelector {
 	return &enrichment.NamespaceSelector
 }
+
+func (enrichment *MetadataEnrichment) IsNamespaceSelectorDefined() bool {
```
0sewa0: is this used?

pkg/api/latest/dynakube/oneagent/props.go
```diff
@@ -133,6 +133,14 @@ func (oa *OneAgent) GetNamespaceSelector() *metav1.LabelSelector {
 	}
 }
 
+func (oa *OneAgent) IsNamespaceSelectorDefined() bool {
```
0sewa0: is this used?

pkg/injection/namespace/mapper/dynakubes.go
waodim: As a matter of fact there is no new `match` as I changed it back after your feedback. Why I am here again checking is because I want to fill up the slices with the respective namespaces. Should I move that into `updateNamespaces`? Doesn't sound too nice to me tbh. What's your take on that?

pkg/api/latest/dynakube/oneagent/props.go
waodim: removed it https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/0db421736836691cba8d7bf4cee8d1365e3eafa2

pkg/api/latest/dynakube/metadataenrichment/props.go
waodim: removed it https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/0db421736836691cba8d7bf4cee8d1365e3eafa2

pkg/injection/namespace/mapper/dynakubes.go
waodim: Confused myself tbh, thx for pointing this out https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/0db421736836691cba8d7bf4cee8d1365e3eafa2

pkg/injection/namespace/mapper/dynakubes.go
0sewa0: yeah, I got confused myself.

the issue is that to do this in a "nice" way would require to rework this whole package, which I would avoid doing right now.

Because having a `match` func that would return `(bool, bool, error)` is already bad, but it would soon return `(bool, bool, bool, error)` (due to the otelp configuration feature), which is fully horrible 😅 

But this block just replicates the logic, which is also not good. (you would need to change it in 2 places in case of change -> error prone)

So we could so something like (maybe you also did something like this, I don't remember 🙈 )
```go
type matchResult struct {
	IsOA, IsME bool
}

func (m matchResult) IsAny() bool {
	return m.IsME || m.IsOA
}

func match(dk *dynakube.DynaKube, namespace *corev1.Namespace) (result matchResult, err error) {...}
```
Which is "semi easy" to extend. Still kinda ugly and a naming nightmare 😅

waodim: My apporach was a little bit different, but I see your approach is better, probably the "best" in the current situation. Should I create a refactor story? Such that we can work on it in the future?

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/da97c11bf3adfa28e3272e7f76e252e615983e36

pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -107,7 +139,22 @@ func (dm DynakubeMapper) mapFromDynakube(nsList *corev1.NamespaceList, dkList *d
 	for i := range nsList.Items {
 		namespace := &nsList.Items[i]
 
-		updated, err = updateNamespace(namespace, dkList)
+		if !isIgnoredNamespace(dm.dk, namespace.Name) {
```
0sewa0: ` match(dm.dk, namespace)` also does that check

pkg/injection/namespace/mapper/conditions.go
```diff
@@ -0,0 +1,96 @@
...
+		cond.Reason = matchesFoundReason
+		msg := formatMatchMessage(names, maxNamesInMsg)
+		cond.Message = msg
+	}
+
+	cond.LastTransitionTime = metav1.Now()
+	_ = meta.SetStatusCondition(conditions, cond)
+}
+
+func updateCollectedNamespacesMonitoredCondition(conditions *[]metav1.Condition) {
```
0sewa0: is this needed? is there a value for having a condition that says what 2 other condition say in more detail?

pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -20,32 +22,63 @@ type DynakubeMapper struct {
...
+			"selector", "MetadataEnrichment",
+			"count (at most 10 are displayed)", len(dm.matchedMENamespaces),
+			"namespaces", dm.matchedMENamespaces,
+		)
+	}
+
+	oaActive := dm.dk.OneAgent().IsAppInjectionNeeded()
+	meActive := dm.dk.MetadataEnrichment().IsEnabled()
+	setNamespacesMonitoredSelectorCondition(dm.dk.Conditions(), "OneAgent", oaActive, dm.matchedOANamespaces)
+	setNamespacesMonitoredSelectorCondition(dm.dk.Conditions(), "MetadataEnrichment", meActive, dm.matchedMENamespaces)
```
0sewa0: why do you have extra `if`s here for the logs?

you could log the same thing in `setNamespacesMonitoredSelectorCondition` as you have all the info

pkg/injection/namespace/mapper/dynakubes.go
waodim: 🫠

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/3ffed9da3144dff5bb80952713570e86ec0f4229

pkg/injection/namespace/mapper/conditions.go
waodim: I agree it is somewhat nonsensical - I removed it.

https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/3ffed9da3144dff5bb80952713570e86ec0f4229

pkg/injection/namespace/mapper/dynakubes.go
waodim: very nice finding, I agree.

done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/3ffed9da3144dff5bb80952713570e86ec0f4229

pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -20,32 +22,62 @@ type DynakubeMapper struct {
...
+		)
+	}
+
+	if dm.dk.MetadataEnrichment().IsEnabled() {
+		log.Info("namespaces monitored",
+			"selector", "MetadataEnrichment",
+			"count (at most 10 are displayed)", len(dm.matchedMENamespaces),
+			"namespaces", dm.matchedMENamespaces,
+		)
+	}
```
0sewa0: pls remove, as they are no longer needed, you have it in the `pkg/injection/namespace/mapper/conditions.go` line 31 😅

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/5484/commits/4e78c9582874320c2998b4d5629775275e1ebf66

PR #5485, opened 2025-10-02
Implement webhook integration test base

avorima:
## Description

Implement basic webhook integration tests for EdgeConnect and DynaKube conversion.
The tests are meant to be similar to how a user would interact with these resources, e.g. with kubectl. They load files from disk, apply the using the envtest client and then compare the result.
Only resource create happy path is covered, but the code can easily be extended as needed.

In order to ensure that the integration tests use the same setup as the live binary I moved the webhook setup functions into the respective validation packages. The test code is basically duplicated apart from the webhook code and could be moved to a separate package (together with the webhook setup) at a later point.

Add default cases to conversion type switches as a fly-by.

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-13997)

## How can this be tested?

Run go tests:

```
make go/test
```

### DIFF
--- a/pkg/api/validation/dynakube/webhook_integration_test.go
+++ b/pkg/api/validation/dynakube/webhook_integration_test.go
@@ -1,0 +1,84 @@
+package validation_test
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	validation "github.com/Dynatrace/dynatrace-operator/pkg/api/validation/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/integrationtests"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/yaml"
+)
+
+func TestWebhook(t *testing.T) {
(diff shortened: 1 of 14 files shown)


andriisoldatenko:
LGTM, works on my machine :) 
```
{"level":"info","ts":"2025-10-03T08:34:57.889+0200","msg":"Stopping and waiting for HTTP servers"}
{"level":"info","ts":"2025-10-03T08:34:57.889+0200","msg":"Wait completed, proceeding to shutdown the manager"}
--- PASS: TestWebhook (6.39s)
    --- PASS: TestWebhook/v1beta3 (2.07s)
    --- PASS: TestWebhook/v1beta4 (0.02s)
    --- PASS: TestWebhook/v1beta5 (0.03s)
PASS
ok      command-line-arguments  6.970s
```

pkg/api/validation/dynakube/webhook_integration_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/yaml"
+)
+
+func TestWebhook(t *testing.T) {
+	clt := integrationtests.SetupWebhookTestEnvironment(t,
+		envtest.WebhookInstallOptions{
+			// TODO(avorima): Load this from a file using Paths
+			ValidatingWebhooks: []*admissionv1.ValidatingWebhookConfiguration{
```
andriisoldatenko, line 27: the only thing i'm worried, that we don't use validationWebhook from file, and we need to make sure we update tests here. 
can we refactor it? or let's create a story and link story to the comment.

avorima: already created it last time you asked about it :) https://github.com/Dynatrace/dynatrace-operator/pull/5380#discussion_r2352207443

andriisoldatenko (approved):

PR #5487, opened 2025-10-03
Update envtest to latest version

gkrenn:
## Description

This PR updates Envtest to always use the latest version. 

In addition it closes 2 bugs in our integration test and adds an automatic cleanup if higher versions are installed. 

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-13500)

## How can this be tested?

- `make go/test`
- Look at GitHub actions, they should be green

### DIFF
--- a/hack/make/prerequisites.mk
+++ b/hack/make/prerequisites.mk
@@ -17,10 +17,7 @@ helmunittest_version=v1.0.2
 # renovate depName=github.com/CycloneDX/cyclonedx-gomod/cmd/cyclonedx-gomod
 cyclonedx_gomod_version=v1.9.0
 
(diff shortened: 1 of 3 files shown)


hack/make/prerequisites.mk
```diff
@@ -60,10 +57,13 @@ prerequisites/envtest:
 
 prerequisites/setup-envtest: prerequisites/envtest
 	@echo "Setting up envtest binaries for Kubernetes version $(ENVTEST_K8S_VERSION)..."
-	setup-envtest use $(ENVTEST_K8S_VERSION) --bin-dir $(GOBIN) -p path || { \
+	@setup-envtest use $(ENVTEST_K8S_VERSION) --bin-dir $(GOBIN) -p path || { \
 		echo "Error: Failed to set up envtest binaries for version $(ENVTEST_K8S_VERSION)."; \
 		exit 1; \
 	}
+	@echo
```
andriisoldatenko, line 64:
```suggestion
	@echo "remove all versions older than $(ENVTEST_K8S_VERSION) from disk"
```

PR #5490, opened 2025-10-03
Add cleanup make target

gkrenn:
## Description

This PR introduces a cleanup command that removes all Dynatrace components from the cluster, regardless of the deployment method.

Additionally, any missing labels on our components have been added.

[Jira](https://dt-rnd.atlassian.net/browse/DAQ-11951)

## How can this be tested?

- `make deploy`
  - Missing labels are there
- `make cleanup`
  - All dt related stuff is removed from the cluster, but files on the node

> Please note that only proper labeled components will be cleaned up. Components rolled out with older releases may not be covered. E.g. CRD, pod disruption budget, ...


0sewa0 (approved):
tested it with an upgrade (from `main`) and observed no issues, labels were present, clean up afterwards also worked

hack/make/deploy/deploy.mk
```diff
@@ -25,10 +25,25 @@ deploy: manifests/crd/helm
...
-undeploy:
+undeploy: cleanup/cr
+	helm uninstall dynatrace-operator \
+			--namespace dynatrace
+
+## Remove all Dynatrace Operator resources from the cluster
+cleanup: cleanup/cr
+	kubectl delete namespace dynatrace --ignore-not-found
+	kubectl delete \
+		clusterrole,clusterrolebinding,mutatingwebhookconfigurations,validatingwebhookconfigurations,crd,priorityclass,poddisruptionbudget \
```
andriisoldatenko: you can do: `kubectl delete all -l app.kubernetes.io/name=dynatrace-operator --ignore-not-found` so we don'nt need to maintain this line
cc @0sewa0 

tested in my cluster:
```
k delete all -l app.kubernetes.io/name=dynatrace-operator
pod "dynatrace-oneagent-csi-driver-4v59m" deleted
pod "dynatrace-oneagent-csi-driver-gxxdt" deleted
pod "dynatrace-oneagent-csi-driver-mnhcd" deleted
pod "dynatrace-operator-54785bf79d-xwcp8" deleted
pod "dynatrace-webhook-6b55698cdf-b9ptz" deleted
pod "dynatrace-webhook-6b55698cdf-qsscd" deleted
service "dynakube-activegate" deleted
service "dynakube-extensions-controller" deleted
service "dynatrace-webhook" deleted
daemonset.apps "dynatrace-oneagent-csi-driver" deleted
deployment.apps "dynatrace-operator" deleted
deployment.apps "dynatrace-webhook" deleted
```

hack/make/deploy/deploy.mk
```diff
@@ -25,10 +25,25 @@ deploy: manifests/crd/helm
...
+		-l app.kubernetes.io/name=dynatrace-operator --ignore-not-found
+
+	@echo -n "Removing Dynatrace Operator secrets and configmaps from all namespaces "
+	@for ns in $$(kubectl get ns -o jsonpath="{.items[*].metadata.name}"); do \
+		kubectl delete secret,cm -l app.kubernetes.io/name=dynatrace-operator --ignore-not-found -n $$ns > /dev/null 2>&1 || true; \
+		printf '.'; \
+	done
+	@echo " done"
+
+cleanup/cr:
```
andriisoldatenko: could you please share with me reasoning behind naming `cr` ? is it custom resource?

hack/make/deploy/deploy.mk
avorima: `all` doesn't actually include all kinds, afaik it's mostly common namespaced types. on my cluster it doesn't even include PDBs. we need the specific kinds here

andriisoldatenko: @avorima  yes you are right, totally missed this part, [here is another way to achieve it without specifying everything](https://github.com/kubernetes/kubectl/issues/151#issuecomment-402003022), but I'm fine to keep as is, only we need to make sure we maintain this list.

andriisoldatenko:
```
kubectl api-resources --verbs=list --namespaced -o name | xargs -n 1 kubectl get --show-kind --ignore-not-found -l app.kubernetes.io/name=dynatrace-operator
```
something like it? @avorima wdyt?

andriisoldatenko: it shows me more now:
```
NAME                                        TYPE                DATA   AGE
secret/dynakube-activegate-tls-secret       Opaque              3      4d21h
secret/dynakube-extensions-controller-tls   kubernetes.io/tls   2      4d21h
NAME                                               SECRETS   AGE
serviceaccount/dynatrace-activegate                0         4d21h
serviceaccount/dynatrace-dynakube-oneagent         0         4d21h
serviceaccount/dynatrace-edgeconnect               0         4d21h
serviceaccount/dynatrace-extensions-controller     0         4d21h
serviceaccount/dynatrace-kubernetes-monitoring     0         4d21h
serviceaccount/dynatrace-logmonitoring             0         4d21h
serviceaccount/dynatrace-node-config-collector     0         4d21h
serviceaccount/dynatrace-oneagent-csi-driver       0         4d21h
serviceaccount/dynatrace-opentelemetry-collector   0         4d21h
serviceaccount/dynatrace-operator                  0         4d21h
serviceaccount/dynatrace-webhook                   0         4d21h
NAME                                                                      ROLE                                     AGE
rolebinding.rbac.authorization.k8s.io/dynatrace-oneagent-csi-driver       Role/dynatrace-oneagent-csi-driver       4d21h
rolebinding.rbac.authorization.k8s.io/dynatrace-operator                  Role/dynatrace-operator                  4d21h
rolebinding.rbac.authorization.k8s.io/dynatrace-operator-supportability   Role/dynatrace-operator-supportability   4d21h
rolebinding.rbac.authorization.k8s.io/dynatrace-webhook                   Role/dynatrace-webhook                   4d21h
NAME                                                               CREATED AT
role.rbac.authorization.k8s.io/dynatrace-oneagent-csi-driver       2025-10-01T15:27:07Z
role.rbac.authorization.k8s.io/dynatrace-operator                  2025-10-01T15:27:07Z
role.rbac.authorization.k8s.io/dynatrace-operator-supportability   2025-10-01T15:27:07Z
role.rbac.authorization.k8s.io/dynatrace-webhook                   2025-10-01T15:27:07Z
```

waodim: tbh I think this is nice. I personally am for it. 👍

avorima: without `--namespaced` the list also includes cluster-wide resources

andriisoldatenko: yes, we can do it as well, since we filter by label. wdyt @gkrenn ?

aorcholski:
```
kind: CSIDriver
  name: csi.oneagent.dynatrace.com
```
still exists after `make cleanup` is used

hack/make/deploy/deploy.mk
gkrenn: Yes, will write it out to avoid confusion

hack/make/deploy/deploy.mk
gkrenn: Sounds like a great idea to create the list on the fly, will update 👍

avorima: since the list includes *all* resources, please document to run this *after* undeploy to avoid deleting stuff like pods or endpoints

gkrenn: Is fixed with the generic delete approach

hack/make/deploy/deploy.mk
```diff
@@ -25,10 +25,27 @@ deploy: manifests/crd/helm
...
+undeploy: cleanup/custom-resources
+	helm uninstall dynatrace-operator \
+			--namespace dynatrace
+
+## Remove all Dynatrace Operator resources from the cluster
+cleanup: cleanup/custom-resources
+	kubectl delete namespace dynatrace --ignore-not-found
+	@echo "Removing Dynatrace Operator cluster-scoped resources"
+	@kubectl api-resources --verbs=list -o name --namespaced=false | \
+		xargs -n 1 -I {} sh -c \
```
aorcholski:
```
[]$ make cleanup
...
Removing Dynatrace Operator cluster-scoped resources
xargs: warning: options --max-args and --replace/-I/-i are mutually exclusive, ignoring previous --max-args value
...
```

hack/make/deploy/deploy.mk
```diff
@@ -25,10 +25,27 @@ deploy: manifests/crd/helm
...
+		"kubectl delete {} --ignore-not-found -l app.kubernetes.io/name=dynatrace-operator 2>&1 | \
+		grep -v 'No resources found'" || true
```
0sewa0: may the future generation/us forgive us for this 😭

hack/make/deploy/deploy.mk
gkrenn: Well, does not happen on macos. Guess it is only on GNU xargs. But does not matter, I can just remove the `-n 1`

gkrenn: Well, does not happen on macos. Guess it is only on GNU xargs. But does not matter, I can just remove the `-n 1`

hack/make/deploy/deploy.mk
```diff
@@ -25,10 +25,27 @@ deploy: manifests/crd/helm
 			--set debug=$(DEBUG)
 
 ## Undeploy the current operator installation
-undeploy:
+undeploy: cleanup/custom-resources
+	helm uninstall dynatrace-operator \
+			--namespace dynatrace
+
+## Remove all Dynatrace Operator resources from the cluster
+cleanup: cleanup/custom-resources
```
avorima: the dependency is easy to break e.g. when dynakubes were already uninstalled but the rest is still there. cleanup should work regardless

hack/make/deploy/deploy.mk
```diff
@@ -25,10 +25,27 @@ deploy: manifests/crd/helm
...
+	kubectl delete namespace dynatrace --ignore-not-found
+	@echo "Removing Dynatrace Operator cluster-scoped resources"
+	@kubectl api-resources --verbs=list -o name --namespaced=false | \
```
avorima: `--namespaced=false` only includes cluster-wide resources, is this intended?

hack/make/deploy/deploy.mk
gkrenn: @avorima Can you elaborate more what you mean? - In the current implementation, I call this command after I deleted the namespace - So there shouldn't be any of our pods left. 

The whole `cleanup` command should be an alternative to `undeploy` in case you are not using helm, so it should work without `undeploy`.

FYI: In the final implementation I only list clusterwide resources to safe time, as anyway the ns got deleted.

hack/make/deploy/deploy.mk
gkrenn: Yes, as the namespace was already deleted in the line before

hack/make/deploy/deploy.mk
gkrenn: Does not really matter as long as a CRD is still there. the delete command will still work if no CRs are deployed.

gkrenn: Will add a check however, to make it more resistant to failures

gkrenn: It's idempotent now 👍

hack/make/deploy/deploy.mk
avorima: i didn't know it's meant be used interchangeably

andriisoldatenko (commented):
Can we make `cleanup-dynatrace-operator.sh` or something (single file) and invoke it inside 1 make command?

Why? Imagine if we want to share the script with a **customer** or with the **support team**. They need to clone the repo and have `make` installed locally. But can we instead give them a link to a single file, and they can run it?

Wdyt? @0sewa0 @gkrenn @avorima

gkrenn:
Moved it into a script 👍

andriisoldatenko:
tested independently and it works:
```
./hack/cluster/cleanup-dynatrace-objects.sh
```

PR #5491, opened 2025-10-06
WIP: Create secret containing DataIngest token for OTLP exporter configuration

bacherfl:
## Description

This PR adds an OTLP reconciler which is responsible for creating the secret containing the DataIngest token required for the OTLP exporter auto configuration. 

Fixes https://dt-rnd.atlassian.net/browse/DAQ-13778

## How can this be tested?

- Unit tests

### DIFF
--- a/pkg/controllers/dynakube/controller.go
+++ b/pkg/controllers/dynakube/controller.go
@@ -23,6 +23,7 @@ import (
 	logmondaemonset "github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/logmonitoring/daemonset"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/oneagent"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/otelc"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/otlp"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/proxy"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/token"
 	"github.com/Dynatrace/dynatrace-operator/pkg/injection/namespace/mapper"
@@ -86,6 +87,7 @@ func NewDynaKubeController(kubeClient client.Client, apiReader client.Reader, co
 		logMonitoringReconcilerBuilder:      logmonitoring.NewReconciler,
 		proxyReconcilerBuilder:              proxy.NewReconciler,
 		kspmReconcilerBuilder:               kspm.NewReconciler,
+		otlpReconcilerBuilder:               otlp.NewReconciler,
 	}
 }
 
@@ -123,6 +125,7 @@ type Controller struct {
 	logMonitoringReconcilerBuilder      logmonitoring.ReconcilerBuilder
 	proxyReconcilerBuilder              proxy.ReconcilerBuilder
 	kspmReconcilerBuilder               kspm.ReconcilerBuilder
+	otlpReconcilerBuilder               otlp.ReconcilerBuilder
 
 	tokens            token.Tokens
 	operatorNamespace string
@@ -425,6 +428,25 @@ func (controller *Controller) reconcileComponents(ctx context.Context, dynatrace
 	err = kspmReconciler.Reconcile(ctx)
 	if err != nil {
 		log.Info("could not reconcile kspm")
+
+		componentErrors = append(componentErrors, err)
+	}
+
+	err = controller.otlpReconcilerBuilder(controller.client,
+		controller.apiReader,
+		dynatraceClient,
+		dk).
+		Reconcile(ctx)
+	if err != nil {
+		if errors.Is(err, oaconnectioninfo.NoOneAgentCommunicationHostsError) {
+			// missing communication hosts is not an error per se, just make sure next the reconciliation is happening ASAP
+			// this situation will clear itself after AG has been started
+			controller.setRequeueAfterIfNewIsShorter(fastUpdateInterval)
+
+			return goerrors.Join(componentErrors...)
+		}
+
+		log.Info("could not reconcile OTLP")
 
 		componentErrors = append(componentErrors, err)
 	}
--- a/pkg/controllers/dynakube/otlp/reconciler.go
+++ b/pkg/controllers/dynakube/otlp/reconciler.go
@@ -1,0 +1,99 @@
+package otlp
+
+import (
+	"context"
+	goerrors "errors"
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers"
+	"github.com/Dynatrace/dynatrace-operator/pkg/injection/namespace/mapper"
+	"github.com/Dynatrace/dynatrace-operator/pkg/otlp/exporterconfig"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/conditions"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Reconciler struct {
+	client          client.Client
+	apiReader       client.Reader
+	dynatraceClient dynatrace.Client
+	dk              *dynakube.DynaKube
+}
+
+type ReconcilerBuilder func(
+	client client.Client,
+	apiReader client.Reader,
+	dynatraceClient dynatrace.Client,
+	dk *dynakube.DynaKube,
+) controllers.Reconciler
+
+//nolint:revive
+func NewReconciler(
+	client client.Client,
+	apiReader client.Reader,
+	dynatraceClient dynatrace.Client,
+	dk *dynakube.DynaKube,
+) controllers.Reconciler {
+	return &Reconciler{
+		client:          client,
+		apiReader:       apiReader,
+		dynatraceClient: dynatraceClient,
+		dk:              dk,
+	}
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context) error {
+	var setupErrors []error
+
+	if r.dk.Spec.OTLPExporterConfiguration == nil {
+		defer r.cleanup(ctx)
+	} else {
+		dkMapper := r.createDynakubeMapper(ctx)
+		if err := dkMapper.MapFromDynakube(); err != nil {
+			log.Info("update of a map of namespaces failed")
+
+			setupErrors = append(setupErrors, err)
+		}
+
+		err := r.generateSecret(ctx)
+		if err != nil {
+			setupErrors = append(setupErrors, err)
+		}
+	}
+
+	if len(setupErrors) > 0 {
+		return goerrors.Join(setupErrors...)
+	}
+
+	log.Info("OTLP exporter configuration reconciled")
+
+	return nil
+}
+
+func (r *Reconciler) cleanup(ctx context.Context) {
+	namespaces, err := mapper.GetNamespacesForDynakube(ctx, r.apiReader, r.dk.Name)
+	if err != nil {
+		log.Error(err, "failed to list namespaces for dynakube", "dkName", r.dk.Name)
+	}
+
+	err = exporterconfig.Cleanup(ctx, r.client, r.apiReader, namespaces, r.dk)
+}
+
+func (r *Reconciler) createDynakubeMapper(ctx context.Context) *mapper.DynakubeMapper {
+	operatorNamespace := r.dk.GetNamespace()
+	dkMapper := mapper.NewDynakubeMapper(ctx, r.client, r.apiReader, operatorNamespace, r.dk)
+
+	return &dkMapper
+}
+
+func (r *Reconciler) generateSecret(ctx context.Context) error {
+	err := exporterconfig.NewSecretGenerator(r.client, r.apiReader, r.dynatraceClient).GenerateForDynakube(ctx, r.dk)
+	if err != nil {
+		if conditions.IsKubeAPIError(err) {
+			conditions.SetKubeAPIError(r.dk.Conditions(), otlpExporterConfigurationConditionType, err)
+		}
+
+		return err
+	}
+
+	return nil
+}
--- a/pkg/controllers/dynakube/otlp/reconciler_test.go
+++ b/pkg/controllers/dynakube/otlp/reconciler_test.go
@@ -1,0 +1,120 @@
+package otlp
+
+import (
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube/otlpexporterconfiguration"
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/scheme/fake"
+	dtclient "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
+	"github.com/Dynatrace/dynatrace-operator/pkg/consts"
+	dtwebhook "github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/mutator"
+	dtclientmock "github.com/Dynatrace/dynatrace-operator/test/mocks/pkg/clients/dynatrace"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
+)
+
+const (
+	testDataIngestToken = "test-ingest-token"
+
+	testDynakube   = "test-dynakube"
+	testDynakube2  = "test-dynakube2"
+	testNamespace  = "test-namespace"
+	testNamespace2 = "test-namespace2"
+
+	testNamespaceSelectorLabel = "namespaceSelector"
+
+	testNamespaceDynatrace = "dynatrace"
+)
+
+func TestReconciler_Reconcile(t *testing.T) {
+	t.Run("reconcile OTLP exporter configuration", func(t *testing.T) {
+		dk := &dynakube.DynaKube{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      testDynakube,
+				Namespace: testNamespaceDynatrace,
+			},
+			Spec: dynakube.DynaKubeSpec{
+				OTLPExporterConfiguration: &otlpexporterconfiguration.Spec{},
+			},
+		}
+
+		clt := fake.NewClientWithIndex(
+			clientInjectedNamespace(testNamespace, testDynakube),
+			clientNotInjectedNamespace(testNamespace2, testDynakube2),
+			clientSecret(testDynakube, testNamespaceDynatrace, map[string][]byte{
+				dtclient.DataIngestToken: []byte(testDataIngestToken),
+			}),
+			dk,
+		)
+		dtClient := dtclientmock.NewClient(t)
+
+		rec := NewReconciler(clt, clt, dtClient, dk)
+
+		err := rec.Reconcile(t.Context())
+		require.NoError(t, err)
+
+		assertSecretFound(t, clt, consts.OTLPExporterSecretName, testNamespace)
+		assertSecretNotFound(t, clt, consts.OTLPExporterSecretName, testNamespace2)
+	})
+}
+
+func assertSecretFound(t *testing.T, clt client.Client, secretName string, secretNamespace string) {
+	var secret corev1.Secret
+	err := clt.Get(t.Context(), client.ObjectKey{Name: secretName, Namespace: secretNamespace}, &secret)
+	require.NoError(t, err, "%s.%s secret not found, error: %s", secretName, secretNamespace, err)
+}
+
+func assertSecretNotFound(t *testing.T, clt client.Client, secretName string, secretNamespace string) {
+	var secret corev1.Secret
+	err := clt.Get(t.Context(), client.ObjectKey{Name: secretName, Namespace: secretNamespace}, &secret)
+	require.Error(t, err, "%s.%s secret found, error: %s ", secretName, secretNamespace, err)
+	assert.True(t, k8serrors.IsNotFound(err), "%s.%s secret, unexpected error: %s", secretName, secretNamespace, err)
+}
+
+func clientInjectedNamespace(namespaceName string, dynakubeName string) *corev1.Namespace {
+	return &corev1.Namespace{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "corev1",
+			Kind:       "Namespace",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespaceName,
+			Labels: map[string]string{
+				dtwebhook.InjectionInstanceLabel: dynakubeName,
+			},
+		},
+	}
+}
+
+func clientNotInjectedNamespace(namespaceName string, dynakubeName string) *corev1.Namespace {
+	return &corev1.Namespace{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "corev1",
+			Kind:       "Namespace",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespaceName,
+			Labels: map[string]string{
+				testNamespaceSelectorLabel: dynakubeName,
+			},
+		},
+	}
+}
+
+func clientSecret(secretName string, namespaceName string, data map[string][]byte) *corev1.Secret {
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "core/v1",
+			Kind:       "Secret",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: namespaceName,
+		},
+		Data: data,
+	}
+}
--- a/pkg/otlp/exporterconfig/conditions.go
+++ b/pkg/otlp/exporterconfig/conditions.go
@@ -1,0 +1,3 @@
+package exporterconfig
+
+const ConfigConditionType = "OTLPExporterConfig"
--- a/pkg/otlp/exporterconfig/secret_generator.go
+++ b/pkg/otlp/exporterconfig/secret_generator.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/timeprovider"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -34,7 +35,7 @@ func NewSecretGenerator(client client.Client, apiReader client.Reader, dtClient 
 	}
 }
 
-// GenerateForDynakube creates/updates the init secret for EVERY namespace for the given dynakube.
+// GenerateForDynakube creates/updates the OTLP exporter config secret for EVERY namespace for the given dynakube.
 // Used by the dynakube controller during reconcile.
 func (s *SecretGenerator) GenerateForDynakube(ctx context.Context, dk *dynakube.DynaKube) error {
 	log.Info("reconciling namespace bootstrapper init secret for", "dynakube", dk.Name)
@@ -66,6 +67,17 @@ func (s *SecretGenerator) GenerateForDynakube(ctx context.Context, dk *dynakube.
 	return nil
 }
 
+func Cleanup(ctx context.Context, client client.Client, apiReader client.Reader, namespaces []corev1.Namespace, dk *dynakube.DynaKube) error {
+	err := cleanupConfig(ctx, client, apiReader, namespaces, dk)
+	if err != nil {
+		log.Error(err, "failed to cleanup bootstrapper config secrets")
+
+		return errors.WithStack(err)
+	}
+
+	return nil
+}
+
 func (s *SecretGenerator) generateConfig(ctx context.Context, dk *dynakube.DynaKube) (map[string][]byte, error) {
 	data := map[string][]byte{}
 
@@ -80,7 +92,7 @@ func (s *SecretGenerator) generateConfig(ctx context.Context, dk *dynakube.DynaK
 		return nil, errors.WithMessage(err, "failed to query tokens")
 	}
 
-	data["apiToken"] = tokens.Data[dtclient.APIToken]
+	data[dtclient.DataIngestToken] = tokens.Data[dtclient.DataIngestToken]
 
 	return data, nil
 }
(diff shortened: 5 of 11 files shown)


pkg/controllers/dynakube/otlp/reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+		dynatraceClient: dynatraceClient,
+		dk:              dk,
+	}
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context) error {
+	var setupErrors []error
+
+	if !r.dk.OTLPExporterConfiguration().IsEnabled() {
+		defer r.cleanup(ctx)
```
avorima: this defer makes the code less clear. please call the function normally

pkg/controllers/dynakube/otlp/reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+	log.Info("OTLP exporter configuration reconciled")
+
+	return nil
+}
+
+func (r *Reconciler) cleanup(ctx context.Context) {
+	if meta.FindStatusCondition(*r.dk.Conditions(), otlpExporterConfigurationConditionType) == nil {
+		return
+	}
+	defer meta.RemoveStatusCondition(r.dk.Conditions(), otlpExporterConfigurationConditionType)
```
avorima: this defer is redundant unless you return early, but even if it did, unconditionally removing the condition even though cleanup failed is not good. the function should return an error and bubble it up to retry cleanup.

pkg/controllers/dynakube/otlp/reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers"
+	"github.com/Dynatrace/dynatrace-operator/pkg/injection/namespace/mapper"
+	"github.com/Dynatrace/dynatrace-operator/pkg/otlp/exporterconfig"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/conditions"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Reconciler struct {
+	client          client.Client
```
0sewa0: could call it `kubeClient` to not clash with the package name (even though it doesn't matter here)

edit: I see you used that name somewhere else :P (you can consider `kubeClient` code as the "newer" code 😅 )

pkg/controllers/dynakube/otlp/reconciler_test.go
```diff
@@ -0,0 +1,257 @@
...
+func (f failingReader) List(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error { //nolint:revive
+	if f.fail {
+		return errors.New("err")
+	}
+	return f.Reader.List(ctx, list, opts...)
+}
+
+func TestReconciler_Reconcile(t *testing.T) {
+	t.Run("reconcile OTLP exporter configuration", func(t *testing.T) {
+		dk := &dynakube.DynaKube{
```
avorima: the tests all use very similar dynakubes. a constructor would reduce duplication

pkg/controllers/dynakube/otlp/reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+
+	if !r.dk.OTLPExporterConfiguration().IsEnabled() {
+		defer r.cleanup(ctx)
+	} else {
+		dkMapper := r.createDynakubeMapper(ctx)
+		if err := dkMapper.MapFromDynakube(); err != nil {
+			log.Info("update of a map of namespaces failed")
+
+			setupErrors = append(setupErrors, err)
+		}
```
0sewa0: this is not a cheap operation, I would avoid doing this multiple times

- If it was adding a different label and such, so we would have true "separation of concerns" then the "cost" of the operation could be argued for, but as of right now, it does the same as the other call

I would do this in the `injection` reconciler

currently we have:
```go
if !r.dk.OneAgent().IsAppInjectionNeeded() && !r.dk.MetadataEnrichment().IsEnabled() {
	defer r.cleanup(ctx)
} else {
	dkMapper := r.createDynakubeMapper(ctx)

	if err := dkMapper.MapFromDynakube(); err != nil {
		log.Info("update of a map of namespaces failed")

	}

	err := r.generateInitSecret(ctx)
	if err != nil {
	}
}
```
this could be turned into something like (my names are bad, so only examples)
```go
if !r.dk.OneAgent().IsAppInjectionNeeded() && !r.dk.MetadataEnrichment().IsEnabled() && !r.dk.OTLPExporterConfiguration().IsEnabled(){
	defer r.unmap(ctx) // renamed from cleanup, and would only care about the mapping
} else {
	dkMapper := r.createDynakubeMapper(ctx)

	if err := dkMapper.MapFromDynakube(); err != nil {
		log.Info("update of a map of namespaces failed")

	}
}

if r.dk.OneAgent().IsAppInjectionNeeded() || r.dk.MetadataEnrichment().IsEnabled() {
	err := r.generateInitSecret(ctx)
	if err != nil {
	}
} else { 
	r.cleanupInitSecret(...)
}

if r.dk.OTLPExporterConfiguration().IsEnabled() {
	err := r.generateOtelpSecret(ctx)
	if err != nil {
	}
} else { 
	r.cleanupOtelpSecret(...)
}
```

pkg/controllers/dynakube/otlp/reconciler_test.go
```diff
@@ -0,0 +1,257 @@
...
+
+	t.Run("no exporter config triggers cleanup", func(t *testing.T) {
+		dk := &dynakube.DynaKube{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      testDynakube,
+				Namespace: testNamespaceDynatrace,
+			},
+			Status: dynakube.DynaKubeStatus{
+				Conditions: []metav1.Condition{
+					{},
```
avorima: what's the purpose of this empty condition?

pkg/otlp/exporterconfig/conditions.go
```diff
@@ -0,0 +1,3 @@
+package exporterconfig
+
+const ConfigConditionType = "OTLPExporterConfigSecret"
```
0sewa0:
```suggestion
const ConditionType = "OTLPExporterConfigSecret"
```

`exporterconfig.ConfigConditionType` is very stuttery

pkg/otlp/exporterconfig/secret_generator.go
```diff
@@ -0,0 +1,174 @@
...
+
+	data[dtclient.DataIngestToken] = tokens.Data[dtclient.DataIngestToken]
+
+	return data, nil
+}
+
+func (s *SecretGenerator) createSecretForNSlist( //nolint:revive // argument-limit
+	ctx context.Context,
+	secretName string,
+	conditionType string,
```
0sewa0: isn't the condtionType for you static? (ie. it always the same)

pkg/controllers/dynakube/otlp/reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ReconcilerBuilder func(
+	client client.Client,
+	apiReader client.Reader,
+	dynatraceClient dynatrace.Client,
+	dk *dynakube.DynaKube,
+) controllers.Reconciler
+
+//nolint:revive
```
avorima: removing this doesn't cause any failures locally. what is this for?

pkg/webhook/mutation/pod/secrets/replicate.go
```diff
@@ -0,0 +1,36 @@
...
+func EnsureReplicated(mutationRequest *dtwebhook.MutationRequest, kubeClient client.Client, apiReader client.Reader, sourceSecretName, targetSecretName string, logger logd.Logger) error { //nolint:revive
+	var initSecret corev1.Secret
+
+	secretObjectKey := client.ObjectKey{Name: targetSecretName, Namespace: mutationRequest.Namespace.Name}
+
+	err := apiReader.Get(mutationRequest.Context, secretObjectKey, &initSecret)
+	if k8serrors.IsNotFound(err) {
+		logger.Info(targetSecretName+" is not available, trying to replicate", "pod", mutationRequest.PodName())
+
+		return bootstrapperconfig.Replicate(
```
0sewa0: got confused here for a sec 😅 

I didn't realize how "generic" the `bootstrapperconfig.Replicate` func is, so I like the idea of reusing it, but couldn't we find a better place for it (and not the `bootstrapperconfig` package 😅 )

pkg/controllers/dynakube/controller.go
```diff
@@ -324,7 +327,7 @@ func (controller *Controller) setupTokensAndClient(ctx context.Context, dk *dyna
 	return dynatraceClient, nil
 }
 
-func (controller *Controller) reconcileComponents(ctx context.Context, dynatraceClient dtclient.Client, istioClient *istio.Client, dk *dynakube.DynaKube) error {
+func (controller *Controller) reconcileComponents(ctx context.Context, dynatraceClient dtclient.Client, istioClient *istio.Client, dk *dynakube.DynaKube) error { //nolint:revive
```
avorima: same as above, doesn't change the result if it's removed

pkg/webhook/mutation/pod/secrets/replicate.go
0sewa0: also, the general signature could be improved, because the `dynakube` is only used for it's namespace

so we could have a generic "replicate" func that has a signature like `Replicate(ctx, source, target)`
and then your `EnsureReplicated` basically "unpacks" the `mutationRequest *dtwebhook.MutationRequest` and calls `Replicate` accordingly.

wdyt?

pkg/otlp/exporterconfig/secret_generator.go
```diff
@@ -0,0 +1,174 @@
...
+// Used by the dynakube controller during reconcile.
+func (s *SecretGenerator) GenerateForDynakube(ctx context.Context, dk *dynakube.DynaKube) error {
+	log.Info("reconciling namespace OTLP exporter secret for", "dynakube", dk.Name)
+
+	data, err := s.generateConfig(ctx, dk)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	nsList, err := mapper.GetNamespacesForDynakube(ctx, s.apiReader, dk.Name)
```
avorima: nsList should be "namespaces" instead. this includes the function name

pkg/otlp/exporterconfig/secret_generator.go
```diff
@@ -0,0 +1,174 @@
...
+		return err
+	}
+
+	log.Info("done updating OTLP exporter config secrets")
+	conditions.SetSecretCreatedOrUpdated(dk.Conditions(), conditionType, consts.OTLPExporterSecretName)
+
+	return nil
+}
+
+func (s *SecretGenerator) createSourceForWebhook(ctx context.Context, dk *dynakube.DynaKube, secretName, conditionType string, data map[string][]byte) error {
```
avorima: the secretName can be generated in the function from the DynaKube and the conditionType is always the same. this can also be applied to function that revive complains about

pkg/otlp/exporterconfig/conditions.go
avorima: AFAICS, this condition isn't used outside of this package, so i'm wondering if we really need to add yet another condition to the DynaKube

pkg/otlp/exporterconfig/secret_generator_test.go
```diff
@@ -0,0 +1,353 @@
...
+func TestCleanup(t *testing.T) {
+	dk := &dynakube.DynaKube{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testDynakube,
+			Namespace: testNamespaceDynatrace,
+		},
+		Spec: dynakube.DynaKubeSpec{},
+		Status: dynakube.DynaKubeStatus{
+			Conditions: []metav1.Condition{
+				{Type: ConfigConditionType},
```
avorima: you're not asserting that this condition is gone

pkg/webhook/mutation/pod/mutator/otlp/exporter/injectors.go
```diff
@@ -33,8 +33,9 @@ func (ti *traceInjector) Inject(c *corev1.Container, apiURL string, override boo
 		return false
 	}
 
-	addEnvVarLiteralValue(c, OTLPTraceEndpointEnv, fmt.Sprintf("%s/%s", apiURL, "traces"))
+	addEnvVarLiteralValue(c, OTLPTraceEndpointEnv, fmt.Sprintf("%s/%s", apiURL, "v1/traces"))
```
avorima: only one of the `%s` is actually variable. the other can be part of the format string

pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator_test.go
```diff
@@ -197,13 +210,32 @@ func TestMutator_Mutate(t *testing.T) {
...
 			Value: "http/protobuf",
 		})
+
+		// verify headers env vars added with Authorization header referencing DT_API_TOKEN
+		assert.Contains(t, containerEnvVars, corev1.EnvVar{Name: OTLPTraceHeadersEnv, Value: OTLPAuthorizationHeader})
+		assert.Contains(t, containerEnvVars, corev1.EnvVar{Name: OTLPMetricsHeadersEnv, Value: OTLPAuthorizationHeader})
+		assert.Contains(t, containerEnvVars, corev1.EnvVar{Name: OTLPLogsHeadersEnv, Value: OTLPAuthorizationHeader})
+
+		// verify DT_API_TOKEN secret ref env var
+		var dtTokenVar *corev1.EnvVar
```
avorima: if you move this out into a helper function you can remove the nolinit directive

pkg/webhook/mutation/pod/mutator/otlp/exporter/env.go
```diff
@@ -20,4 +20,8 @@ const (
 	OTLPLogsCertificateEnv     = "OTEL_EXPORTER_OTLP_LOGS_CERTIFICATE"
 	OTLPMetricsCertificateEnv  = "OTEL_EXPORTER_OTLP_METRICS_CERTIFICATE"
 	OTLPExporterCertificateEnv = "OTLP_EXPORTER_OTLP_CERTIFICATE"
+
+	OTLPAuthorizationHeader = "Authorization=Api-Token $(DT_API_TOKEN)"
+
+	DynatraceAPIToken = "DT_API_TOKEN"
```
avorima: shouldn't this be `DynatraceAPITokenEnv`?

pkg/webhook/mutation/pod/annotations/annotations.go
```diff
@@ -0,0 +1,21 @@
+package annotations
+
+import "github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/mutator"
+
+func SetDynatraceInjectedAnnotation(mutationRequest *mutator.MutationRequest) {
```
avorima: what's the purpose of moving these out here?

pkg/webhook/mutation/pod/handler/otlp/handle.go
```diff
@@ -48,3 +74,23 @@ func (h *Handler) Handle(mutationRequest *dtwebhook.MutationRequest) error {
 
 	return nil
 }
+
+func (h *Handler) isInputSecretPresent(mutationRequest *dtwebhook.MutationRequest, sourceSecretName, targetSecretName string) bool {
+	err := secrets.EnsureReplicated(mutationRequest, h.kubeClient, h.apiReader, sourceSecretName, targetSecretName, log)
+	if k8serrors.IsNotFound(err) {
+
+		annotations.SetNotInjectedAnnotations(mutationRequest, NoOTLPExporterConfigSecretReason)
```
avorima: this uses the generic mutator reason annotation, but other handlers use their domain-specific one. the AnnotationOTLPReason constant isn't even referenced by anything

pkg/webhook/mutation/pod/handler/otlp/handle_test.go
```diff
@@ -112,18 +128,108 @@ func TestHandler_Handle(t *testing.T) {
...
 		err := h.Handle(request)
 		assert.Error(t, err)
 	})
+	t.Run("skip injection and annotate when input secret missing and not replicable", func(t *testing.T) {
+		mockEnvVarMutator := webhookmock.NewMutator(t)
+		mockResourceAttributeMutator := webhookmock.NewMutator(t)
+
+		// enable env var mutator so that secret presence is checked
+		mockEnvVarMutator.On("IsEnabled", mock.Anything).Return(true)
+		mockEnvVarMutator.AssertNotCalled(t, "Mutate", mock.Anything)
```
avorima: AFAIK, the mock will just fail the test if it encounters a call that isn't anticipated, i.e. by using `.On(...)`. so these negative assertions aren't needed

pkg/otlp/exporterconfig/conditions.go
avorima: cc @0sewa0

pkg/webhook/mutation/pod/secrets/replicate.go
bacherfl: sounds good - looking at the current structure, the `pkg/controllers/util/kubeobjects/secrets` package might be a reasonable place to move this func to - wdyt?

pkg/controllers/dynakube/otlp/reconciler_test.go
bacherfl: this was most likely the result of an overly eager auto completion in the IDE ;) will remove that

pkg/controllers/dynakube/otlp/reconciler.go
bacherfl: was added due to an earlier approach where the otlp secret reconciler was called from the dynakube controller, but is not needed anymore

pkg/webhook/mutation/pod/annotations/annotations.go
bacherfl: The OTLP and injection handler both use these, so it made sense to extract this out of the injection related package. However this needs to be adapted, as the OTLP handler mistakenly also sets the generic annotations, as you correctly found in https://github.com/Dynatrace/dynatrace-operator/pull/5491/#discussion_r2452399599 - I will adapt this to pass the annotations as parameters, then it should make more sense

pkg/webhook/mutation/pod/handler/otlp/handle.go
bacherfl: see https://github.com/Dynatrace/dynatrace-operator/pull/5491/#discussion_r2454384419

pkg/webhook/mutation/pod/handler/otlp/handle_test.go
bacherfl: yes, the purpose of these is essentially to make it more explicit, but you are right that they are not needed, as the test will fail anyway, if the calls have not been registered with `On()` - I will remove them again

pkg/controllers/dynakube/otlp/reconciler.go
bacherfl: reducing the `MapFromDynakube()` calls is a valid argument here - In this case I would move the logic from the OTLP reconciler into the injection reconciler. Since it depends from the results of the mapper we cannot hide it behind a `otlp.Reconcile()` anymore

bacherfl: I think while we're at it, we can also fetch the namespaces once at the beginning of the injection reconciler, as those are otherwise fetched again multiple times across the different `SecretGenerators` and during the cleanup calls

bacherfl: another related question, since this came up for the cleanup logic in the (now likely obsolete) OTLP reconciler: https://github.com/Dynatrace/dynatrace-operator/pull/5491/#discussion_r2452043429

is there a reason why the original `cleanup()` function just logged any errors, but did not pass them back to the reconciler?

pkg/controllers/dynakube/otlp/reconciler.go
bacherfl: I agree - but since i kept this function to the cleanup logic of the existing injection reconciler, I think we first need to clarify if there maybe was a reason for just logging the error instead of passing it back - see https://github.com/Dynatrace/dynatrace-operator/pull/5491#discussion_r2454481913

pkg/controllers/dynakube/otlp/reconciler.go
0sewa0: > is there a reason why ...

yes, we do not want to "fail" due to some cleanup logic, this you can see in many places, that if some cleanup fails, we log the error and move on

not ideal, but with the amount of different usecases covered with a single CR, keeping track "what cleanup is error worthy" just gets too complicated. There are tickets to rethink our "reconcile flow" to hopefully make things better.

pkg/webhook/mutation/pod/secrets/replicate.go
0sewa0: could even be put as part of our `generic` query implementation, however I do not see any real usecase of replicating other things (maybe configmaps)

pkg/controllers/dynakube/otlp/reconciler.go
avorima: ack. ok @bacherfl, then just make the flow linear (remove defer) in the cleanup and we think about returning errors at a later point

pkg/controllers/dynakube/controller_test.go
```diff
@@ -392,7 +392,7 @@ func TestReconcileComponents(t *testing.T) {
 
 		require.Error(t, err)
 		// goerrors.Join concats errors with \n
-		assert.Len(t, strings.Split(err.Error(), "\n"), 7) // ActiveGate, Extension, OtelC, OneAgent LogMonitoring, and Injection reconcilers
+		assert.Len(t, strings.Split(err.Error(), "\n"), 7) // ActiveGate, Extension, OtelC, OneAgent LogMonitoring, Injection and OTLP reconcilers
```
avorima: should be KSPM, not OTLP. how about we make each reconciler return a unique error string and check that this aggregated error contains all of the expected strings instead of having this comment?

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -204,3 +239,30 @@ func (r *Reconciler) createDynakubeMapper(ctx context.Context) *mapper.DynakubeM
 
 	return &dkMapper
 }
+
+func (r *Reconciler) cleanupInitSecret(ctx context.Context, namespaces []corev1.Namespace) {
+	if meta.FindStatusCondition(*r.dk.Conditions(), codeModulesInjectionConditionType) == nil &&
+		meta.FindStatusCondition(*r.dk.Conditions(), metaDataEnrichmentConditionType) == nil {
+		return
+	}
+	defer meta.RemoveStatusCondition(r.dk.Conditions(), codeModulesInjectionConditionType)
```
avorima: defer not needed

pkg/controllers/dynakube/injection/reconciler_test.go
```diff
@@ -468,6 +493,47 @@ func TestCleanupOneAgentInjection(t *testing.T) {
 	})
 }
 
+func TestCleanupOTLPInjection(t *testing.T) {
+	ctx := context.Background()
```
avorima: use t.Context

pkg/util/kubeobjects/secret/replicate.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Replicate will only create the secret once, doesn't mean for keeping the secret up to date
+func Replicate(ctx context.Context, query QueryObject, sourceSecretName, targetSecretName, sourceNs, targetNs string) error { //nolint:revive
```
avorima: i once spent hours on a bug that turned out to be mixed up name and namespace args. my preventative measure at the time was  to use `client.ObjectKey`. i propose to do the same for this API. it's what the query needs anyways

pkg/webhook/mutation/pod/annotations/annotations.go
```diff
@@ -0,0 +1,21 @@
+package annotations
+
+import "github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/mutator"
+
+func SetDynatraceInjectedAnnotation(mutationRequest *mutator.MutationRequest, injectedAnnotation, reasonAnnotation string) {
```
avorima: package is already called annotations, so these functions don't need to repeat it. also, the "Dynatrace" can be removed

pkg/webhook/mutation/pod/handler/otlp/config.go
```diff
@@ -7,3 +7,15 @@ import (
 var (
 	log = logd.Get().WithName("pod-mutation-otlp")
 )
+
+const (
+	OTLPAnnotationPrefix = "otlp-exporter-configuration"
+	// AnnotationOTLPInjectionEnabled controls whether the automatic injection of OTLP env vars and resource attributes should happen for a pod
+	AnnotationOTLPInjectionEnabled = OTLPAnnotationPrefix + ".dynatrace.com/inject"
```
avorima: this isn't used. there is another constant with the same name in pkg/webhook/mutation/pod/mutator/config.go which is used though, along with another set of all of these constants. please only use one set

pkg/controllers/dynakube/controller_test.go
bacherfl: fixed the comment in [b31611b](https://github.com/Dynatrace/dynatrace-operator/pull/5491/commits/b31611be1f33f2e39461c799263d37a94c5b91e5). Adapting the error strings should be done in a seperate PR though, as this is out of scope, and the PR is already quite large

pkg/controllers/dynakube/injection/reconciler.go
bacherfl: adapted with [b31611b](https://github.com/Dynatrace/dynatrace-operator/pull/5491/commits/b31611be1f33f2e39461c799263d37a94c5b91e5)

pkg/controllers/dynakube/injection/reconciler_test.go
bacherfl: adapted with [b31611b](https://github.com/Dynatrace/dynatrace-operator/pull/5491/commits/b31611be1f33f2e39461c799263d37a94c5b91e5)

pkg/util/kubeobjects/secret/replicate.go
bacherfl: good idea, i added that with [b31611b](https://github.com/Dynatrace/dynatrace-operator/pull/5491/commits/b31611be1f33f2e39461c799263d37a94c5b91e5)

pkg/webhook/mutation/pod/annotations/annotations.go
bacherfl: adapted with [b31611b](https://github.com/Dynatrace/dynatrace-operator/pull/5491/commits/b31611be1f33f2e39461c799263d37a94c5b91e5)

pkg/webhook/mutation/pod/handler/otlp/config.go
bacherfl: good catch, seems like that slipped through in my earlier commit. fixed now with [b31611b](https://github.com/Dynatrace/dynatrace-operator/pull/5491/commits/b31611be1f33f2e39461c799263d37a94c5b91e5)

pkg/controllers/dynakube/injection/reconciler_test.go
avorima: that's weird i could've sworn i saw `t.Context` added in the diff overview, but `context.Background` still here

pkg/util/kubeobjects/secret/replicate_test.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+const testSourceSecretName = "source-secret"
+
+var (
+	testLog = logd.Get().WithName("secret-test")
+)
+
+func TestReplicate(t *testing.T) {
+	ctx := context.Background()
```
avorima: `t.Context` please

pkg/controllers/dynakube/injection/reconciler_test.go
bacherfl: ah correct, I accidentally did that for the very similarly named `CleanupOneAgentInjection` test, which should be fine as well. Did the same for the `CleanupOTLP` now. If you'd like, I can do the same for the other tests in this file as well, as there are quite a lot of other occurrences of `context.Background()` - given the size of this PR i'm a bit hesitant to do too much refactoring though

avorima: no, it's alright. we actually have a ticket to use it everywhere, but i want to new code to already use it before then

pkg/webhook/mutation/pod/handler/injection/handle.go
```diff
@@ -77,13 +77,22 @@ func (h *Handler) Handle(mutationRequest *dtwebhook.MutationRequest) error {
 		}
 
 		if !mutated {
-			setNotInjectedAnnotations(mutationRequest, NoMutationNeededReason)
+			annotations.SetNotInjected(
```
0sewa0, line 86: found a small hiccup in this file, while testing

the code
```go
	if !h.isInputSecretPresent(mutationRequest, bootstrapperconfig.GetSourceConfigSecretName(mutationRequest.DynaKube.Name), consts.BootstrapperInitSecretName) {
		return nil
	}
```

will always run, even if only otelp is configured

We should have a check similar to the one in `otepl/handle.go`, but for the oa/meta injection
```go
func (h *Handler) Handle(mutationRequest *dtwebhook.MutationRequest) error {
	if !mutationRequest.DynaKube.OneAgent().IsAppInjectionNeeded() && !mutationRequest.DynaKube.MetadataEnrichment().IsEnabled(){ ... }
```

pkg/webhook/mutation/pod/handler/otlp/handle.go
```diff
@@ -1,30 +1,57 @@
...
 	if h.envVarMutator.IsEnabled(mutationRequest.BaseRequest) {
+		if !h.isInputSecretPresent(
+			mutationRequest,
+			exporterconfig.GetSourceConfigSecretName(mutationRequest.DynaKube.Name),
+			consts.OTLPExporterSecretName,
+		) {
+			log.Debug("required input secret not present, skipping OTLP injection", "podName", mutationRequest.PodName(), "namespace", mutationRequest.Namespace.Name)
+
+			return nil
+		}
```
0sewa0, line 53: something is wrong here
during my test I managed to have the injected Container error
```
Error: secret "dynatrace-otlp-exporter-config" not found
```

How to reproduce:
- Have a `token` secret, that has no `dataIngestToken`
- So the Operator will correctly error, as it can't create the `<dk>-otlp-exporter-config`
- But the webhook still injects

0sewa0 (changes requested):

pkg/webhook/mutation/pod/handler/otlp/handle.go
bacherfl: there seems to have been an error in the original secret replication logic where any errors other than `NotFound` were not returned to the handler. In this case the issue was then that the webhook `ClusterRole` did not have the correct permission for the otlp secret, and the related error, which in this case was not a `NotFound`, was not considered and the webhook proceeded to be called. Should be fixed now with [af8475a](https://github.com/Dynatrace/dynatrace-operator/pull/5491/commits/af8475ae91b161b7c6c74ad2ba21e8eb638a394c)

pkg/webhook/mutation/pod/handler/injection/handle.go
bacherfl: added with [3141841](https://github.com/Dynatrace/dynatrace-operator/pull/5491/commits/3141841e5261b0cd69f02854eed2bf4aa3428658)

pkg/util/kubeobjects/secret/replicate_test.go
bacherfl: adapted with [a2d8d11](https://github.com/Dynatrace/dynatrace-operator/pull/5491/commits/a2d8d1103dec84bd8378b4b2f4e13b8985ba6680)

PR #5500, opened 2025-10-07
Remove the `autoUpdate` field

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/DAQ-11402)

## Description

`autoUpdate` field removed from oa.CFS / oa.CNFS and oa.HI specs.

`conversion.internal.dynatrace.com/auto-update` annotation stores the value when converting a CR from an older version to the latest version.

ci.yaml: setup-envtest requires golang >=1.25.0.

## How can this be tested?

1) unittests
2) apply a dynakube of version v1beta3 / v1beta4 / v1beta5 with (oa.CFS / oa.CNFS / oa.HI).`autoUpdate` set to true/false
check if the annotation exists and has correct value
```
kubectl -n dynatrace get -o yaml dynakube.dynatrace.com/dynakube
```
3) apply a dynakube of version v1beta6 with `conversion.dynatrace.com/auto-update: "false"` annotation and oa.CFS / oa.CNFS / oa.HI
check if the `autoUpdate` field is set
```
kubectl -n dynatrace get -o yaml dynakube.v1beta3/4/5.dynatrace.com/dynakube
```

### DIFF
--- a/pkg/controllers/dynakube/otelc/endpoint/reconciler.go
+++ b/pkg/controllers/dynakube/otelc/endpoint/reconciler.go
@@ -122,5 +122,6 @@ func BuildOTLPEndpoint(dk dynakube.DynaKube) (string, error) {
 
 		dtEndpoint = fmt.Sprintf("https://%s/e/%s/api/v2/otlp", serviceFQDN, tenantUUID)
 	}
+
 	return dtEndpoint, nil
 }
--- a/pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
+++ b/pkg/webhook/mutation/pod/mutator/otlp/exporter/mutator.go
@@ -2,8 +2,8 @@ package exporter
 
 import (
 	"fmt"
-	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/otelc/endpoint"
 
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/otelc/endpoint"
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubeobjects/env"
 	maputils "github.com/Dynatrace/dynatrace-operator/pkg/util/map"


pkg/api/conversion/conversion.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+const (
+	Prefix = "conversion.dynatrace.com/"
+
+	AutoUpdateKey = Prefix + "auto-update"
+)
+
+type RemovedFields struct {
+	annotations *map[string]string
```
avorima: don't need to use a pointer to map to modify it

pkg/api/conversion/conversion.go
```diff
@@ -0,0 +1,50 @@
+package conversion
+
+import (
+	"strconv"
+
+	"k8s.io/utils/ptr"
+)
+
+const (
+	Prefix = "conversion.dynatrace.com/"
```
avorima: the name should contain something like `internal` or `private` to let clearly show that the annotation is not meant for general use

pkg/api/conversion/conversion.go
```diff
@@ -0,0 +1,50 @@
...
+
+type RemovedFields struct {
+	annotations *map[string]string
+}
+
+func NewRemovedFields(annotations *map[string]string) *RemovedFields {
+	return &RemovedFields{annotations: annotations}
+}
+
+func (rf *RemovedFields) IsAutoUpdate() *bool {
```
avorima: GetAutoUpdate would match better with the setter

pkg/api/conversion/conversion.go
0sewa0: why is it a `*map`? 

maps are always passed by reference

pkg/api/latest/dynakube/oneagent/props.go
```diff
@@ -33,6 +34,8 @@ func NewOneAgent(spec *Spec, status *Status, codeModulesStatus *CodeModulesStatu
 		featureOneAgentPrivileged:        featureOneAgentPrivileged,
 		featureOneAgentSkipLivenessProbe: featureOneAgentSkipLivenessProbe,
 		featureBootstrapperInjection:     featureBootstrapperInjection,
+
+		removedFeatureAutoUpdate: removedFeatureAutoUpdate,
```
0sewa0: the latest version does not care about the value of this field

`autoUpdate` is "enabled" if no version/image is set

pkg/api/v1beta3/dynakube/convert_from.go
avorima, line 1: the annotation should be removed when converting down

pkg/controllers/dynakube/oneagent/oneagent_reconciler_test.go
```diff
@@ -316,12 +317,14 @@ func TestReconcile_InstancesSet(t *testing.T) {
...
 	t.Run("Status.OneAgent.Instances set, if autoUpdate is false", func(t *testing.T) {
 		dk := base.DeepCopy()
-		autoUpdate := false
+
 		reconciler.dk = dk
 		reconciler.connectionInfoReconciler = createConnectionInfoReconcilerMock(t)
 		reconciler.versionReconciler = createVersionReconcilerMock(t)
 		reconciler.tokens = createTokens()
-		dk.Spec.OneAgent.ClassicFullStack.AutoUpdate = &autoUpdate //nolint:staticcheck
+		dk.Annotations = map[string]string{
```
chrismuellner: should this test use the annotation?

.github/workflows/ci.yaml
```diff
@@ -136,7 +136,7 @@ jobs:
       - name: Setup Golang
         uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
         with:
-          go-version-file: "${{ github.workspace }}/go.mod"
+          go-version: "1.25.2"
```
0sewa0: ?

if this is needed, then pls create a tech-debt story for revisting and fixing this properly after the PR, so we don't forget about it (comment the ticket number here so it is obvious)

aorcholski: tech-debt ticket DAQ-15287

0sewa0:
```suggestion
          go-version: "1.25.2" # to be fixed by DAQ-15287
```

I meant it like this 😅🙈

andriisoldatenko: PR https://github.com/Dynatrace/dynatrace-operator/pull/5531 for it.

pkg/api/conversion/conversion.go
```diff
@@ -0,0 +1,55 @@
...
+
+	AutoUpdateKey = Prefix + "auto-update"
+)
+
+type Field[T any] struct {
+	data map[string]string
+	name string
+}
+
+func (f Field[T]) Key() string {
```
avorima: this key method is unused outside of this package, which means it can be removed.

an alternative would be to remove the AutoUpdateKey constant (or make it private) in favor of calls like this `dk.RemovedFields().AutoUpdate.Key()`. this would ensure that these keys are always accessed through the removed fields at the cost of verbosity.

pkg/api/v1beta3/dynakube/convert_from.go
```diff
@@ -28,6 +29,8 @@ func (dst *DynaKube) ConvertFrom(srcRaw conversion.Hub) error {
 	dst.fromActiveGateSpec(src)
 	dst.fromTemplatesSpec(src)
 
+	delete(dst.Annotations, dkconversion.AutoUpdateKey)
```
avorima: this deletion could happen in the field getter. need to make sure that we don't call Get twice though

pkg/api/conversion/conversion.go
aorcholski: `Key()` function removed
`AutoUpdateKey` is used by unittests in other modules.

pkg/api/v1beta3/dynakube/convert_from.go
aorcholski: removing of "internal" annotations from older CR (dst) is unrelated to Field.Set/Get. I think I moved it to better place. Now we see why/how they were created.
[437ecb5](https://github.com/Dynatrace/dynatrace-operator/pull/5500/commits/437ecb581316a5178fc00bb852c3a30a6cf39ea1)

pkg/api/v1beta3/dynakube/convert_from.go
```diff
@@ -37,6 +38,7 @@ func (dst *DynaKube) fromBase(src *dynakubelatest.DynaKube) {
 	}
 
 	dst.ObjectMeta = *src.ObjectMeta.DeepCopy() // DeepCopy mainly relevant for testing
+	delete(dst.Annotations, dkconversion.AutoUpdateKey)
```
0sewa0: you should remove this AFTER, `dst.fromOneAgentSpec(src)` runs, doesn't this cause problems?

IMO, what we could do, is introduce a `dkconversion.CleanupAnnotations(dst.annotations)`, which would remove all the these annotations.

And you could just call it via a defer in the beginning:
```go 
func (dst *DynaKube) ConvertFrom(srcRaw conversion.Hub) error {
	defer dkconversion.CleanupAnnotations(dst.annotations)

	src := srcRaw.(*dynakubelatest.DynaKube)
    // ...
```

aorcholski: The `dst` is old version of dynakube. Annotations from the old CR  are not considered as part of any conversion logic I mean from latest to v1beta3/4/5.

IMHO `oldCR.field = oldCR.Annotation[key]` instead of `oldCR.field = latestCR.Annotation[key]` is a BUG.

I will add `CleanupAnnotations` function but called directly.

0sewa0: ?

In this case (`ConvertFrom`) `dst` is the `v1beta3` DynaKube, where you NEED to consider the annotation (and remove them after conversion)

In the `ConvertTo` where the `dst` is the `latest` DynaKube, where you need to set the annotations

aorcholski: We don't need v1beta3."internal" annotations as a result of conversion to v1beta3 and we don't need them (`v1beta3=DeepCopy(latest)`)  to perform conversion of the autoUpdate field.

we do perform the conversion in the following way :
```
v1beta3.autoUpdate =                                                 = latestAnnotation["auto-update"]
latestAnnotation["auto-update"] =                                    = v1beta3.autoUpdate
```
it would be strange this way :
```
v1beta3.autoUpdate =              v1beta3.Annotation["auto-update"] = latestAnnotation["auto-update"]
latestAnnotation["auto-update"] = v1beta3.Annotation["auto-update"] = v1beta3.autoUpdate
```

0sewa0: ah, because the annotation is in the `src`, so you just have to remove it from the `dst` after copying the annotations.

got it, thx :D 

🧠 💨

PR #6102, opened 2026-01-27
Add opt-out feature flag enable-attributes-dt.kubernetes

jokuniew:
## Description

old ticket: https://dt-rnd.atlassian.net/browse/DAQ-1441
new ticket: https://dt-rnd.atlassian.net/browse/ICP-2018
This PR introduces new Feature Flag `feature.dynatrace.com/enable-attributes-dt.kubernetes` with default value `true`

Old properties (dt.kubernetes cluster.id/workload.kind/workload.name) are removed (when ff set to `false`) from:
- bootstrapper's `--attribute` args
- `OTEL_RESOURCE_ATTRIBUTES` (with otlpExporterConfiguration in dynakube) 

If feature flag is enabled, old properties are added

## How can this be tested?
`make go/test` (unittest & integration test)

Tested manually by:
```bash
make deploy
k create -f dynakube.yaml (with otlpExporterConfiguration, no feature flag set)
k create ns sample-app && k run nginx --image=nginx -n sample-app
k get pod -n sample-app nginx -o yaml |grep dt.kubernetes # bootstrapper --attribute & otel's OTEL_RESOURCE_ATTRIBUTES should be listed

# add feature flag to dynakube feature.dynatrace.com/enable-attributes-dt.kubernetes: "false"
k delete pod -n sample-app nginx
k run nginx --image=nginx -n sample-app
k get pod -n sample-app nginx -o yaml |grep dt.kubernetes # no output expected

```

## E2E
`make test/e2e/applicationmonitoring/metadataenrichment`
> [!WARNING]
> ~During e2e tests I've found out that indeed the `dt.kubernetes` values were removed from bootstrapper's `--attribute` list but still writes to `enrichment/dt_metadata.json` file [here](https://github.com/Dynatrace/dynatrace-bootstrapper/blob/main/pkg/configure/enrichment/metadata/content.go#L74-L76)~

~It was fixed with https://github.com/Dynatrace/dynatrace-bootstrapper/pull/157 This is dependency for this PR.~

~Commit to be reverted once dependencies PR is merged https://github.com/Dynatrace/dynatrace-operator/pull/6102/commits/9dc71241a2cb0b9eda1066ce1f445bd379a683fc~

> [!NOTE]
>  Dependency was resolved and delivered in `dynatrace-bootstrapper` `1.3.0` version. Changelog [here](https://github.com/Dynatrace/dynatrace-bootstrapper/releases/tag/v1.3.0)
>
> Used latest `dynatrace-boostrapper` in `go.mod` in this PR


pkg/webhook/mutation/pod/mutator/metadata/mutator.go
```diff
@@ -68,7 +68,9 @@ func (mut *Mutator) Mutate(request *dtwebhook.MutationRequest) error {
 		WorkloadName: workloadInfo.Name,
 	}
 
-	SetDeprecatedAttributes(&attrs)
+	if request.DynaKube.FF().EnableAttributesDtKubernetes() {
+		SetDeprecatedAttributes(&attrs)
```
avorima: can we make this private as a fly-by? i don't see it being used anywhere else

pkg/webhook/mutation/pod/mutator/metadata/mutator_test.go
```diff
@@ -310,104 +310,133 @@ func TestMutate(t *testing.T) {
...
+				err := mut.Mutate(&request)
+				require.NoError(t, err)
+				require.NotEqual(t, *expectedPod, *request.Pod)
+				require.NotEmpty(t, request.Pod.OwnerReferences)
+
+				assert.Contains(t, request.InstallContainer.Args, buildArgument("k8s.workload.kind", request.Pod.OwnerReferences[0].Kind))
+				assert.Contains(t, request.InstallContainer.Args, buildArgument("k8s.workload.name", request.Pod.OwnerReferences[0].Name))
+				assert.Contains(t, request.InstallContainer.Args, buildArgument(nsMetaAnnotationKey, nsMetaAnnotationValue))
+
+				if tc.withDeprecatedAnnotations {
```
avorima, line 421: the other test also asserts that annotations are not present

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator.go
```diff
@@ -117,13 +116,24 @@ func (m *Mutator) addResourceAttributes(request *dtwebhook.BaseRequest, c *corev
 		"k8s.node.name":                "$(K8S_NODE_NAME)",
 	}
 
+	if request.DynaKube.FF().EnableAttributesDtKubernetes() {
+		kubernetesMetaDataAttributes.Merge(Attributes{
```
avorima: i don't know why merge was used here initially, but a simple map assignment is less magic

pkg/webhook/mutation/pod/webhook_integration_test.go
```diff
@@ -348,7 +286,87 @@ func TestWebhook(t *testing.T) {
 	})
 }
 
-func TestOTLPWebhook(t *testing.T) {
+func PropagationTest(t *testing.T, clt client.Client, withDeprecatedAnnotations bool) {
```
avorima: t.Helper

test/features/applicationmonitoring/metadata_enrichment.go
```diff
@@ -138,6 +146,31 @@ func MetadataEnrichment(t *testing.T) features.Feature {
 	return builder.Feature()
 }
 
+func MetadataEnrichmentDeprecatedAttributes(t *testing.T) features.Feature {
+	builder := features.New("metadata-enrichment")
```
avorima: use a different name

test/features/applicationmonitoring/metadata_enrichment.go
```diff
@@ -138,6 +146,31 @@ func MetadataEnrichment(t *testing.T) features.Feature {
...
+	secretConfig := tenant.GetSingleTenantSecret(t)
+
+	// Register actual test
```
avorima: this and the install comment can be removed

test/helpers/components/metadataenrichment/assertion.go
```diff
@@ -20,8 +20,12 @@ const (
 )
 
 type Metadata struct {
-	WorkloadKind string `json:"dt.kubernetes.workload.kind,omitempty"`
-	WorkloadName string `json:"dt.kubernetes.workload.name,omitempty"`
+	K8sWorkloadKind string `json:"k8s.workload.kind,omitempty"`
```
avorima: We'll get rid of the deprecated fields at one point so then the distinction isn't relevant anymore. Let's just keep using WorkloadKind/Name

pkg/webhook/mutation/pod/mutator/metadata/mutator.go
jokuniew: made it private here https://github.com/Dynatrace/dynatrace-operator/pull/6102/commits/804a61ee222d2040ffd6bb50422dae1d93c57480

pkg/webhook/mutation/pod/mutator/metadata/mutator_test.go
jokuniew: you mean to to `else` statement? Added here https://github.com/Dynatrace/dynatrace-operator/pull/6102/changes/e99870889f633a6a01b58abcedc9b0789808b776

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator.go
jokuniew: just map assignments here https://github.com/Dynatrace/dynatrace-operator/pull/6102/commits/a2252b0facbb3b20f99395e79600401a096799c7

pkg/webhook/mutation/pod/webhook_integration_test.go
jokuniew: done here https://github.com/Dynatrace/dynatrace-operator/pull/6102/changes/44d1497e9f2f88a7c24c0508b2a2a9ab393a03fa

test/features/applicationmonitoring/metadata_enrichment.go
jokuniew: extended name to `metadata-enrichment-deprecated-attributes`
here https://github.com/Dynatrace/dynatrace-operator/pull/6102/commits/94e6f90990af7221a161e6559b142645fd434961

Still, one target ` make test/e2e/applicationmonitoring/metadataenrichment` will launch those two test suites. I don't see the need to run them separately

test/features/applicationmonitoring/metadata_enrichment.go
jokuniew: removed here https://github.com/Dynatrace/dynatrace-operator/commit/94e6f90990af7221a161e6559b142645fd434961

test/helpers/components/metadataenrichment/assertion.go
jokuniew: removed k8s suffix from those fields here https://github.com/Dynatrace/dynatrace-operator/pull/6102/commits/a7086469c735e470340ef6fdbb41f2c83172226c

avorima (dismissed):

jokuniew:
Additional changes will be introduced

jokuniew:
Extra changes added - add flag to `bootstrap`, so dynatrace-bootstrapoer's method can be triggered with boolean parameter - whether to add deprecated values or not

test/e2e/features/applicationmonitoring/metadata_enrichment.go
```diff
@@ -135,6 +145,29 @@ func MetadataEnrichment(t *testing.T) features.Feature {
 	return builder.Feature()
 }
 
+func MetadataEnrichmentDeprecatedAttributes(t *testing.T) features.Feature {
```
waodim:
```suggestion
func MetadataEnrichmentWithoutDeprecatedAttributes(t *testing.T) features.Feature {
```

waodim (dismissed):
lgtm

```
kubectl get pod -n bh-day-ns php-deployment-b2-65848c584c-pr2hg  -o yaml |grep dt.kubernetes 
      value: dt.entity.kubernetes_cluster=KUBERNETES_CLUSTER-79FB555A2F2EB7FE,k8s.namespace.name=bh-day-ns,dt.kubernetes.workload.name=php-deployment-b2,k8s.node.name=$(K8S_NODE_NAME),k8s.cluster.uid=1b69074c-40da-4d09-91f2-3dbb6dcbfef2,dt.kubernetes.cluster.id=1b69074c-40da-4d09-91f2-3dbb6dcbfef2,k8s.workload.kind=deployment,k8s.workload.name=php-deployment-b2,k8s.container.name=php,k8s.pod.uid=$(K8S_PODUID),k8s.cluster.name=dynakube,k8s.pod.name=$(K8S_PODNAME),dt.kubernetes.workload.kind=deployment
albian.krasniqi@DT-H0J42G95FW /Users/albian.krasniqi/Desktop/tmp 
% kubectl get pod -n bh-day-ns php-deployment-b2-65848c584c-pr2hg  -o yaml |grep dt.kubernetes 
Error from server (NotFound): pods "php-deployment-b2-65848c584c-pr2hg" not found
albian.krasniqi@DT-H0J42G95FW /Users/albian.krasniqi/Desktop/tmp 
% kubectl get pod -n bh-day-ns php-deployment-b2-65848c584c-tpnkq  -o yaml |grep dt.kubernetes 
albian.krasniqi@DT-H0J42G95FW /Users/albian.krasniqi/Desktop/tmp 
```

maybe let's wait for another approval before merging to make sure nothing was overseen

jokuniew:
@waodim  I've set FF to false + deployed pod `nginx`. 
Then I removed FF (to test default `true` behavior) and deployed nginx2 

That's the comparison of attributes passed to boostrapper (initContainer)

image

so, there should be no `dt.kubernetes` attributes there. There are still attr like `dt.entity.kubernetes_cluster`, but to `dt.kubernetes`



and the actual differences in container should be visible in `/var/lib/dynatrace/enrichment/dt_metadata.json`

image

cmd/bootstrapper/cmd.go
```diff
@@ -69,6 +71,10 @@ func AddFlags(cmd *cobra.Command) {
 
 	cmd.PersistentFlags().Lookup(MetadataEnrichmentFlag).NoOptDefVal = "true"
 
+	cmd.PersistentFlags().BoolVar(&enableAttributesDTKubernetes, EnableAttributesDTKubernetesFlag, true, "(Optional) Should the deprecated attributes dt.kubernetes be added to the metadata enrichment.")
+
+	cmd.PersistentFlags().Lookup(EnableAttributesDTKubernetesFlag).NoOptDefVal = "true"
```
avorima: i don't know what the initial initial with this was, but this doesn't do anything. not for this flag neither for the others

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator.go
```diff
@@ -117,16 +116,21 @@ func (m *Mutator) addResourceAttributes(request *dtwebhook.BaseRequest, c *corev
 		"k8s.node.name":                "$(K8S_NODE_NAME)",
 	}
 
+	if request.DynaKube.FF().EnableAttributesDTKubernetes() {
+		kubernetesMetaDataAttributes["dt.kubernetes.cluster.id"] = request.DynaKube.Status.KubeSystemUUID
```
avorima: could also be a constant

jokuniew: Used constant https://github.com/Dynatrace/dynatrace-operator/pull/6102/commits/7ee56b23749c980b381737b1aa79c939446412cf

cmd/bootstrapper/cmd.go
jokuniew: I can agree for EnableAttributesDTKubernetesFlag since it has default value set to `true` anyway
https://github.com/Dynatrace/dynatrace-operator/pull/6102/commits/9804811faabec82318db2d7bf9305bd52df0ca37

But I think that for `--metadata-enrichment` and `--supress-error` this NoOptDefVal option is needed.
Default value of those flags is `false`, but if user pass it without option like `--metadata-enrichment` it should be set to `true`.
And we use it that way in bootstrapper (without explicit value)
image

pkg/webhook/mutation/pod/handler/injection/deprecated.go
```diff
@@ -3,13 +3,13 @@ package injection
 import podattr "github.com/Dynatrace/dynatrace-bootstrapper/cmd/k8sinit/configure/attributes/pod"
 
 const (
-	deprecatedClusterIDKey = "dt.kubernetes.cluster.id"
+	DeprecatedClusterIDKey = "dt.kubernetes.cluster.id"
```
avorima, line 6: it's a bit weird to have this here and the other two keys in metadata/deprecated.go

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator.go
avorima: please look for further occurrences of the string `dt.kubernetes.cluster.id`

cmd/bootstrapper/cmd.go
avorima: then i'm wondering why the default value is false if we want true

pkg/webhook/mutation/pod/handler/injection/deprecated.go
jokuniew: I was trying to test the hypothesis that dt.kubernetes.cluster.id is not part of metadata-enrichment, but apparently setting `metadataEnrichment enable:false` has no effect >:(

I might need to ask @0sewa0 why it was done in two places https://github.com/Dynatrace/dynatrace-operator/pull/5601

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator.go
jokuniew: right, sorry for that https://github.com/Dynatrace/dynatrace-operator/pull/6102/commits/55f0d0f0b5eea89a9b9c726cdec5a9cf907d8c97

cmd/bootstrapper/cmd.go
jokuniew: In terms of `--suppress-error` and especially `--metadata-enrichment` sounds to me like tech debt ticket to simplify this (we always set it - change default value to true), do you agree? I can create it

avorima: yes, please

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator.go
avorima: i still see one result
```
pkg/webhook/mutation/pod/webhook_integration_test.go
547:					assert.Equal(t, dk.Status.KubeSystemUUID, gotResourceAttributes["dt.kubernetes.cluster.id"])
```

cmd/bootstrapper/cmd.go
```diff
@@ -22,9 +22,10 @@ const (
 	SuppressErrorsFlag = k8sinit.SuppressErrorsFlag
 	TechnologiesFlag   = move.TechnologyFlag
 
-	TargetVersionFlag      = "version"
-	FlavorFlag             = "flavor"
-	MetadataEnrichmentFlag = "metadata-enrichment"
+	TargetVersionFlag                = "version"
+	FlavorFlag                       = "flavor"
+	MetadataEnrichmentFlag           = "metadata-enrichment"
+	EnableAttributesDTKubernetesFlag = "enable-attributes-dt-kubernetes"
```
waodim, line 28: just to be on the safe side is it on purpose that the cli flag is called `enable-attributes-dt-kubernetes` and the feature flag `enable-attributes-dt.kubernetes`?

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator.go
jokuniew: sorry x2! 🙈
https://github.com/Dynatrace/dynatrace-operator/pull/6102/commits/2e38cc8d141e903a1f8d2d98b3b0815f6e40656a

pkg/webhook/mutation/pod/handler/injection/deprecated.go
waodim: Maybe the split happened because this one is cluster-level metadata that also goes with oA injection and the other two are workload-level metadata that are bound to metadata enrichment? This would maybe explain why `metadataEnrichment enable:false` has no effect

cmd/bootstrapper/cmd.go
jokuniew: I've started writing ticket for this case and I think it's really how it should have been right now @avorima 
it works just like `--debug` flag - if not set -> false, if is there -> true. So it does not have to be `--debug=true`

I was not 100% correct with statement "we always set it"), if user do FF `feature.dynatrace.com/injection-failure-policy=silent we do not set the arg
https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/webhook/mutation/pod/handler/injection/init.go#L30-L32

That arguments why false is the default value

avorima: ok, so `NoOptDefVal = "true"` does in fact do nothing

jokuniew: My bad, `NoOptDefVal` is really redundant line. 
I've blindly assumed that we need to specify explicitly default value + NoOptDefVal to have it, but boolean flags behaves like this natively. Sorry for confusion! https://github.com/Dynatrace/dynatrace-operator/pull/6102/commits/85730e81ae78f601641dc5e80838464f74ed2df5

cmd/bootstrapper/cmd.go
jokuniew: yes, i've avoided the dot `.` character in cli param name on purpose, and the feature flag is just copied from ticket requirements

I did a bit deeper investigation if it would be the problem if i just copypase FF value to param name
and tl;dr - in cobra - no

but if by any chance we decide to use viper on top https://github.com/spf13/viper?tab=readme-ov-file#setting-overrides
the nested fields are defined with dot, so it would be interpreted as
```
enable-attributes-dt:
    kubernetes
```
So if i don't need dot, I'd like to avoid it here imo

pkg/webhook/mutation/pod/handler/injection/deprecated.go
0sewa0: TLDR: It should be ok to have all the deprecated attributes setting in one place

the weirdness comes from the fact that `pkg/webhook/mutation/pod/handler/injection` is the general handler for both `metadataEnrichment` and `oneAgent` injection.

This used to kinda make sense, as the `cluster.id` was ALWAYS provided as an attribute.
While the rest of the `dt.kubernetes.` attributes were only used in case of `metadataEnrichment`. (also this was before this double `handler` madness)

IMO, now that `metadataEnrichment` is always on when we inject (either because of OA/OTLP injection, or standalone metadataEnricment) it should be OK to have it in one place.

> `metadataEnrichment enable:false` has no effect

yes, this is ugly/confusing, but there is no better way, other than just removing the `.enabled` field, as it servers no real purpose. (`enable: false` == default, no standalone metadataEnrichment, `enable: true`, turns on standalone metadataEnrichment, but you should use a `namespaceSelector`)
This "standalone metadataEnrichment" is mainly a usecase for classic and standalone logmon

jokuniew: Created a subpr to talk about it https://github.com/Dynatrace/dynatrace-operator/pull/6494
it looks nice since we have all deprecated fields in 1 place.
But i'm a bit concerned about what @0sewa0  wrote, that `cluster.id` was always provided and others were provided conditionally -> it looks like with this change we mix logic, but on the other hand this single attribute that was meant to disappear finally 🤷‍♀️

avorima (dismissed):

avorima (approved):

cmd/bootstrapper/cmd.go
jokuniew: the "why" for the record cause we are doing the same with bootstrapper 
https://github.com/spf13/pflag/blob/master/bool.go#L62

PR #6281, opened 2026-03-09
Decrease Dynakube & Edgeconnect default re-queue interval to 15m

jdanieck:
## Description

Centralize env-var duration parsing with pkg/util/envvars and set `Dynakube` & `Edgeconnect` default requeue interval to 15m.

Tracked in ICP-1085.

## How can this be tested?

```sh
 go test ./pkg/util/envvars -run '^TestGetDurationMinutes$' -v
=== RUN   TestGetDurationMinutes
=== RUN   TestGetDurationMinutes/env_var_is_set_to_valid_minutes_and_properly_parsed
=== RUN   TestGetDurationMinutes/env_var_is_set_to_zero_and_returns_default
=== RUN   TestGetDurationMinutes/env_var_is_set_to_large_value_and_properly_parsed
=== RUN   TestGetDurationMinutes/env_var_is_set_to_invalid_value_and_returns_default
=== RUN   TestGetDurationMinutes/env_var_is_set_to_negative_value_and_returns_default
=== RUN   TestGetDurationMinutes/env_var_is_set_to_float_value_and_returns_default
=== RUN   TestGetDurationMinutes/env_var_is_NOT_set_and_returns_default
=== RUN   TestGetDurationMinutes/env_var_is_empty_string_and_returns_default
=== RUN   TestGetDurationMinutes/different_default_values_are_respected
=== RUN   TestGetDurationMinutes/env_var_is_set_above_maximum_and_is_capped
=== RUN   TestGetDurationMinutes/env_var_is_set_to_maximum_and_properly_parsed
--- PASS: TestGetDurationMinutes (0.00s)
    --- PASS: TestGetDurationMinutes/env_var_is_set_to_valid_minutes_and_properly_parsed (0.00s)
    --- PASS: TestGetDurationMinutes/env_var_is_set_to_zero_and_returns_default (0.00s)
    --- PASS: TestGetDurationMinutes/env_var_is_set_to_large_value_and_properly_parsed (0.00s)
    --- PASS: TestGetDurationMinutes/env_var_is_set_to_invalid_value_and_returns_default (0.00s)
    --- PASS: TestGetDurationMinutes/env_var_is_set_to_negative_value_and_returns_default (0.00s)
    --- PASS: TestGetDurationMinutes/env_var_is_set_to_float_value_and_returns_default (0.00s)
    --- PASS: TestGetDurationMinutes/env_var_is_NOT_set_and_returns_default (0.00s)
    --- PASS: TestGetDurationMinutes/env_var_is_empty_string_and_returns_default (0.00s)
    --- PASS: TestGetDurationMinutes/different_default_values_are_respected (0.00s)
    --- PASS: TestGetDurationMinutes/env_var_is_set_above_maximum_and_is_capped (0.00s)
    --- PASS: TestGetDurationMinutes/env_var_is_set_to_maximum_and_properly_parsed (0.00s)
PASS

```


pkg/controllers/edgeconnect/controller.go
```diff
@@ -41,8 +42,9 @@ import (
 )
 
 const (
-	fastUpdateInterval    = 1 * time.Minute
-	defaultUpdateInterval = 30 * time.Minute
+	fastRequeueInterval    = 1 * time.Minute
+	defaultRequeueInterval = 15 * time.Minute
+	requeueEnvVar          = "DT_EDGECONNECT_REQUEUE_INTERVAL_MIN"
```
andriisoldatenko, line 47: It can be confusing for someone, min - sometimes means minimum - i prefer to rename to MINUTES or use seconds.
```
EC_REQUEUE_INTERVAL_SECONDS
```
also you can simplify name, cause it's internal parameter not for customer. And Ticket doesn't propose to use it, i would not introduce it.

pkg/controllers/dynakube/controller.go
```diff
@@ -50,8 +51,9 @@ import (
 )
 
 const (
-	fastUpdateInterval    = 1 * time.Minute
-	defaultUpdateInterval = 30 * time.Minute
+	fastRequeueInterval    = 1 * time.Minute
+	defaultRequeueInterval = 15 * time.Minute
+	requeueEnvVar          = "DT_DYNAKUBE_REQUEUE_INTERVAL_MIN"
```
0sewa0, line 56: related to @avorima's question

This should become obsolete after https://dt-rnd.atlassian.net/browse/ICP-1115 

I don't know if its worth introducing this envvar parsing.
- If we want to still add this, I would not document it officially, as it will be (hopefully) only used in a single version.

I guess you were inspired by what is in `pkg/controllers/dynakube/oneagent/oneagent_reconciler.go`, which is very old logic, and should possibly be removed when the fate of the `nodes-controller` is finally decided. I remember discussing this, and saying leave that `oneagent_reconciler.go` logic as is.

Also, @andriisoldatenko had a good comment: These envs are not set able via the helm chart, so not very user friendly in general. (the `oneagent_reconciler.go` logic is older than the helm chart :) )

jdanieck: @avorima @andriisoldatenko  @0sewa0  thanks for the feedback!

You're right — the env var wasn't in the original scope. Here's my thinking:

The ticket asked to change hardcoded 30m → 15m. I added the env var to make it configurable without recompilation for scale/perf testing and per-deployment fine-tuning by devs/SREs if 15m proves problematic. It can be manually added to the Helm chart when needed without being officially exposed.

I noticed this pattern was already used in the oneagent controller, so I extracted and applied the same approach to other resources - which reinforced my thinking it was the right way to go.

I hear the concerns:
- Becomes obsolete after ICP-1115
- Adds complexity for a short-lived feature

That said, I'd tend not to assume ICP-1115 will land soon - things get deprioritized, people get sick, etc. Since we're already touching this code, having it configurable now doesn't hurt and gives us flexibility if this code sticks around longer than expected.

I think it's worth keeping, but happy to simplify if you prefer.

jdanieck: As discussed offline, we decided to go with simple hardcoded 15m.

PR #6293, opened 2026-03-11
Default OneAgent seccomp profile to RuntimeDefault

averagenative:
## Summary
- When no custom `secCompProfile` is configured, the OneAgent container now defaults to `seccompProfile.type: RuntimeDefault` instead of leaving it unset (`nil`)
- This aligns the OneAgent container with all other operator components (ActiveGate, KSPM, OTEL, EdgeConnect, init containers) which already default to `RuntimeDefault`
- Custom `Localhost` profiles via `secCompProfile` continue to work as before

## Motivation
Clusters with security policies (e.g. Gatekeeper / Azure Policy) that require an explicit seccomp profile reject OneAgent pods because no profile is set. Setting `secCompProfile` in the CR doesn't help either, as it always maps to `type: Localhost` — there was no way to get `type: RuntimeDefault`.

## Test plan
- [x] Existing daemonset tests updated and passing
- [ ] Deploy OneAgent without `secCompProfile` set — verify container gets `seccompProfile.type: RuntimeDefault`
- [ ] Deploy OneAgent with custom `secCompProfile` — verify `Localhost` type still works
- [ ] Deploy OneAgent in privileged mode — verify no seccomp profile is set


pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -432,34 +432,42 @@ func (b *builder) securityContext() *corev1.SecurityContext {
...
+		return &corev1.SeccompProfile{
+			Type:             corev1.SeccompProfileTypeLocalhost,
+			LocalhostProfile: &secCompName,
+		}
+	}
+
+	return &corev1.SeccompProfile{
+		Type: corev1.SeccompProfileTypeRuntimeDefault,
+	}
+}
```
0sewa0, line 452: there is a simpler way to do this:
```suggestion
func seccompProfile(dk *dynakube.DynaKube) *corev1.SeccompProfile {
	if dk != nil && dk.OneAgent().GetSecCompProfile() != "" {
		return &corev1.SeccompProfile{
			Type:             corev1.SeccompProfileTypeLocalhost,
			LocalhostProfile: ptr.To(dk.OneAgent().GetSecCompProfile()),
		}
	}

	return &corev1.SeccompProfile{
		Type: corev1.SeccompProfileTypeRuntimeDefault,
	}
}
```

0sewa0 (changes requested):
still trying to verify if this is a safe change, 
- ie.: is `RuntimeDefault` supported on all supported OneAgent version on all supported runtimes.

I tested it with the internal `1.336` Agent, on GCP, in both CloudNative and Classic mode, and seems to work.

@averagenative do you have any info about this?

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
averagenative: @0sewa0  RuntimeDefault should be okay with most of the modern clusters.. my biggest concern would likely be around Openshift / ROSA / GKE clusters which seem to have more instances of Gatekeeper running. However, a lot of the default policies seem to prefer this profile. 

RE: oa permissions, I think it's fairly safe as It doesn't apply to privileged mode at all. In unprivileged mode, the 14 explicit capabilities (not seccomp) are what grant OneAgent its instrumentation powers. RuntimeDefault doesn't block ptrace, process_vm_*, or other instrumentation-related syscalls. Full testing would probably be a good idea to confirm though.. not sure what scenarios are currently tested

averagenative (commented):
The changes you made look good

0sewa0:
I will close the PR as it is not the safest to set `RuntimeDefault` in all cases, without a way to undo that behavior. (depending on your runtime it may or may not work).

And we don't want to keep pilling on feature-flags and such, so we want to solve this properly:
- We have a ticket https://dt-rnd.atlassian.net/browse/ICP-2432 to make this configurable in a better way in general
  - For none internals, the gist:
     - Move away from the `secCompProfile` field (bad name in general)
     - Add a new `securityContext.seccompProfile` field, that is a 1-to-1 of what you can set on the Pod, so it can be localhost, or RuntimeDefault or anything that suites your fancy.

PR #6327, opened 2026-03-18
Use listType=set marker for KSPM mappedHostPaths

avorima:
## Description

I stumbled upon the GetUniqueMappedHostPaths function and thought that this was going at it the wrong way.
Providing duplicate paths should not be valid.

This only affects new resources, so we need to keep the deduplication code around until all invalid resources have been corrected.

ICP-2454

## How can this be tested?

* Deploy DynaKube with different values for mapped host paths


0sewa0:
what is the state here?

avorima:
> what is the state here?

No prio. I'll close this. We can pick it up if we want to

PR #6427, opened 2026-04-08
Add cache middleware for the dtclient

0sewa0:
## Description

https://dt-rnd.atlassian.net/browse/ICP-1115 (kinda incorrect, will need to update it)

Adds a little cache middleware to the dtclient.

It only caches GET requests that are not binary downloads
The TTL of each cache entry is the `spec.dynatraceApiRequestThreshold`
The key for each cache entry is the URL + body + tokens hashed:
- So if anything relevant changes in the DynaKube, this key should change

There is also a simple goroutine that keeps this cache clean.
- This is necessary due to the precious cache entry key
   - example: The user changes tokens, the old cache entry for this will never be hit again, so it wont be "auto" fixed
- Also necessary because users can turn off features, so they will stop making calls to certain APIs

~Rips out any and all previous dtclient throttling (or just misused throttling) from the code:~
- Moved to https://github.com/Dynatrace/dynatrace-operator/pull/6460

## How can this be tested?

~Just try it :D~ 

Now the part where it is actively used is https://github.com/Dynatrace/dynatrace-operator/pull/6460 (where nothing is in the way)

### DIFF
--- a/cmd/operator/cmd.go
+++ b/cmd/operator/cmd.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/scheme"
+	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/middleware"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/certificates"
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/installconfig"
@@ -59,7 +60,10 @@ func runInPod(kubeCfg *rest.Config) error {
 		return err
 	}
 
-	return errors.WithStack(operatorManager.Start(ctrl.SetupSignalHandler()))
+	signalHandler := ctrl.SetupSignalHandler()
+	go middleware.RunPeriodicCacheCleanup(signalHandler, k8senv.GetDTClientCacheCleanInterval(log))
+
+	return errors.WithStack(operatorManager.Start(signalHandler))
 }
 
 func runLocally(ctx context.Context, kubeCfg *rest.Config) error {
--- a/pkg/clients/dynatrace/middleware/cache.go
+++ b/pkg/clients/dynatrace/middleware/cache.go
@@ -1,0 +1,149 @@
+package middleware
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"maps"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
+)
+
+var (
+	log = logd.Get().WithName("dynatraceapi-cache")
+
+	cache = newCache()
+)
+
+// cacheEntry holds all relevant info for a cached response.
+// The only not obvious part is that the `body` has to be stored separate from the actual `response`.
+// The reason for this is that the `Body` part of the `http.Response` is an `io.ReadCloser`.
+// This means the you may only read the body once, which is not good if we would like to reuse the `http.Response`
+// To combat this we read the `Body` while creating the `cacheEntry` and store it inside this new entry.
+// We always "recreate"(put it into a new Reader) the `Body` in the affected `http.Response` or when we returned the cached `http.Response`.
+type cacheEntry struct {
+	response     *http.Response
+	body         []byte
+	creationTime time.Time
+	ttl          time.Duration
+}
+
+func (ce *cacheEntry) isOutdated() bool {
+	return time.Since(ce.creationTime) > ce.ttl
+}
+
+type responseCache struct {
+	entries map[string]*cacheEntry
+	mu      sync.Mutex
+}
+
+func (rc *responseCache) get(key string) *http.Response {
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
+
+	entry, ok := rc.entries[key]
+	if !ok {
+		return nil
+	}
+
+	if entry.isOutdated() {
+		log.Debug("outdated entry found, removing", "url", entry.response.Request.URL)
+
+		delete(rc.entries, key)
+
+		return nil
+	}
+
+	// Reconstruct a fresh response so every caller gets a full, open body and
+	// an independent header map (shallow copy shares the underlying map).
+	resp := *entry.response
+	resp.Header = entry.response.Header.Clone()
+	resp.Body = io.NopCloser(bytes.NewReader(entry.body))
+
+	return &resp
+}
+
+func (rc *responseCache) remove(key string) {
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
+
+	delete(rc.entries, key)
+}
+
+func (rc *responseCache) set(key string, response *http.Response, ttl time.Duration) error {
+	if response == nil {
+		return nil
+	}
+
+	body, err := io.ReadAll(response.Body)
+	if err != nil {
+		response.Body.Close()
+
+		return err
+	}
+
+	response.Body.Close()
+
+	// Restore body for the current caller so they can still read it.
+	response.Body = io.NopCloser(bytes.NewReader(body))
+
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
+
+	rc.entries[key] = &cacheEntry{
+		response:     response,
+		body:         body,
+		creationTime: time.Now(),
+		ttl:          ttl,
+	}
+
+	return nil
+}
+
+func (rc *responseCache) cleanup() {
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
+
+	maps.DeleteFunc(rc.entries, func(key string, value *cacheEntry) bool {
+		isOutdated := value.isOutdated()
+		if isOutdated {
+			log.Debug("outdated entry found, removing", "url", value.response.Request.URL)
+		}
+
+		return isOutdated
+	})
+}
+
+func newCache() *responseCache {
+	return &responseCache{
+		entries: make(map[string]*cacheEntry),
+		mu:      sync.Mutex{},
+	}
+}
+
+// RunPeriodicCacheCleanup will remove all expired entries from the dtclient cache once every period.
+// This is necessary because if the user changes tokens, disables features etc., some entries will never be hit again,
+// so we can't clean them up during normal reconciles.
+// This only makes sense to use in the Operator Pod.
+// - The bootstrap command only runs once, and can only download a ~700MB archive. (can be less or more, depending on the technologies present in the archive)
+// - The csidriver (mainly `provisioner` container) does not make "in-memory cacheable" requests, it only downloads +700MB archives, that we "cache" on the node.
+// - The webhook (and any other not listed command) does not use the dtclient.
+func RunPeriodicCacheCleanup(ctx context.Context, period time.Duration) {
+	ticker := time.NewTicker(period)
+	defer ticker.Stop()
+
+	log.Debug("periodic cache cleanup routine setup", "period", period)
+
+	for {
+		select {
+		case <-ticker.C:
+			log.Debug("periodic cache cleanup started")
+			cache.cleanup()
+		case <-ctx.Done():
+			return
+		}
+	}
+}
--- a/pkg/clients/dynatrace/middleware/cache_test.go
+++ b/pkg/clients/dynatrace/middleware/cache_test.go
@@ -1,0 +1,180 @@
+package middleware
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"io"
+	"net/http"
+	"net/url"
+	"testing"
+	"testing/synctest"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// freshCache returns a new responseCache and replaces the package-level one,
+// preventing state from leaking between tests.
+func freshCache(t *testing.T) *responseCache {
+	t.Helper()
+	c := newCache()
+	cache = c
+
+	return c
+}
+
+func fakeResponse(body string) *http.Response {
+	return &http.Response{
+		StatusCode: http.StatusOK,
+		Header:     make(http.Header),
+		Body:       io.NopCloser(bytes.NewBufferString(body)),
+		Request:    &http.Request{URL: &url.URL{Host: "example.com"}},
+	}
+}
+
+func TestResponseCache_SetAndGet(t *testing.T) {
+	t.Run("miss on empty cache", func(t *testing.T) {
+		c := freshCache(t)
+		assert.Nil(t, c.get("anykey"))
+	})
+
+	t.Run("hit after store within TTL", func(t *testing.T) {
+		c := freshCache(t)
+		require.NoError(t, c.set("key", fakeResponse("hello"), time.Minute))
+
+		resp := c.get("key")
+		require.NotNil(t, resp)
+		body, _ := io.ReadAll(resp.Body)
+		require.NoError(t, resp.Body.Close())
+		assert.Equal(t, "hello", string(body))
+	})
+	t.Run("expired entry is evicted on get and is a miss", func(t *testing.T) {
+		c := freshCache(t)
+
+		synctest.Test(t, func(t *testing.T) {
+			require.NoError(t, c.set("key", fakeResponse("data"), time.Minute))
+			time.Sleep(time.Minute + time.Second)
+
+			entry := c.get("key") // triggers eviction
+			assert.Nil(t, entry, "expired entry must be treated as a miss")
+		})
+
+		assert.NotContains(t, c.entries, "key")
+	})
+
+	t.Run("nil response is not stored", func(t *testing.T) {
+		c := freshCache(t)
+		require.NoError(t, c.set("key", nil, time.Minute))
+		assert.Nil(t, c.get("key"))
+	})
+
+	t.Run("body can be read multiple times from cached response", func(t *testing.T) {
+		c := freshCache(t)
+		require.NoError(t, c.set("key", fakeResponse("body-content"), time.Minute))
+
+		resp1 := c.get("key")
+		require.NotNil(t, resp1)
+		body1, _ := io.ReadAll(resp1.Body)
+		require.NoError(t, resp1.Body.Close())
+
+		resp2 := c.get("key")
+		require.NotNil(t, resp2)
+		body2, _ := io.ReadAll(resp2.Body)
+		require.NoError(t, resp2.Body.Close())
+
+		assert.Equal(t, "body-content", string(body1))
+		assert.Equal(t, "body-content", string(body2))
+	})
+
+	t.Run("caller's original response body is restored after store", func(t *testing.T) {
+		c := freshCache(t)
+		resp := fakeResponse("original")
+		require.NoError(t, c.set("key", resp, time.Minute))
+
+		body, _ := io.ReadAll(resp.Body)
+		assert.Equal(t, "original", string(body))
+	})
+
+	t.Run("returns error and does not cache when reading body fails", func(t *testing.T) {
+		c := freshCache(t)
+		readErr := errors.New("read error")
+		body := &ioErrReader{err: readErr}
+		resp := fakeResponse("")
+		resp.Body = body
+
+		err := c.set("key", resp, time.Minute)
+
+		require.ErrorIs(t, err, readErr)
+		assert.Nil(t, c.get("key"))
+		assert.True(t, body.closed, "body must be closed when reading fails")
+	})
+}
+
+func TestResponseCache_Cleanup(t *testing.T) {
+	t.Run("removes expired entries, keeps fresh ones", func(t *testing.T) {
+		c := freshCache(t)
+
+		synctest.Test(t, func(t *testing.T) {
+			require.NoError(t, c.set("expired", fakeResponse("old"), time.Minute))
+			require.NoError(t, c.set("fresh", fakeResponse("new"), 3*time.Minute))
+			time.Sleep(2 * time.Minute)
+
+			c.cleanup()
+		})
+
+		assert.NotContains(t, c.entries, "expired")
+		assert.Contains(t, c.entries, "fresh")
+	})
+
+	t.Run("does nothing on empty cache", func(t *testing.T) {
+		c := freshCache(t)
+		assert.NotPanics(t, cache.cleanup)
+		assert.Empty(t, c.entries)
+	})
+}
+
+func TestRunPeriodicCacheCleanup(t *testing.T) {
+	t.Run("evicts expired entries after a tick", func(t *testing.T) {
+		c := freshCache(t)
+
+		synctest.Test(t, func(t *testing.T) {
+			require.NoError(t, c.set("key", fakeResponse("data"), time.Minute))
+
+			go RunPeriodicCacheCleanup(t.Context(), 2*time.Minute)
+
+			time.Sleep(4 * time.Minute)
+
+			synctest.Wait()
+		})
+
+		assert.Empty(t, c.entries)
+	})
+
+	t.Run("goroutine exits when context is canceled", func(t *testing.T) {
+		freshCache(t)
+		ctx, cancel := context.WithCancel(t.Context())
+		assert.NotPanics(t, func() {
+			go RunPeriodicCacheCleanup(ctx, time.Hour)
+			cancel()
+		})
+	})
+}
+
+// ioErrReader is a ReadCloser that always returns the configured error on Read
+// and tracks whether Close was called.
+type ioErrReader struct {
+	err    error
+	closed bool
+}
+
+func (r *ioErrReader) Read(_ []byte) (int, error) {
+	return 0, r.err
+}
+
+func (r *ioErrReader) Close() error {
+	r.closed = true
+
+	return nil
+}
--- a/pkg/clients/dynatrace/middleware/roundtripper.go
+++ b/pkg/clients/dynatrace/middleware/roundtripper.go
@@ -1,0 +1,78 @@
+package middleware
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/core"
+)
+
+type roundTripperFunc func(*http.Request) (*http.Response, error)
+
+func (rt roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	if rt == nil {
+		panic("empty roundTripperFunc used")
+	}
+
+	return rt(r)
+}
+
+func NewCacheRoundTripper(next http.RoundTripper, ttl time.Duration) http.RoundTripper {
+	return roundTripperFunc(func(r *http.Request) (*http.Response, error) {
+		if r.Method != http.MethodGet ||
+			r.Header.Get("Accept") == "application/octet-stream" {
+			return next.RoundTrip(r)
+		}
+
+		cacheKey := buildCacheKey(r)
+
+		if ttl == 0 {
+			log.Debug("cache ttl is 0, no cache will be used")
+
+			// if the caching was turned off intermittently, those entries should be removed
+			cache.remove(cacheKey)
+
+			return next.RoundTrip(r)
+		}
+
+		cachedResponse := cache.get(cacheKey)
+		if cachedResponse != nil {
+			cachedResponse.Header.Set(core.CacheHitHeader, "true")
+			cachedResponse.Request = r
+
+			return cachedResponse, nil
+		}
+
+		// send the actual request
+		resp, err := next.RoundTrip(r)
+		if err == nil && core.IsSuccessResponse(resp) {
+			// err is ignored, as cache.set can only error due to failing to read the body, which will cause an error down the line anyway
+			// adding extra logs will just repeat the same, adding no value
+			_ = cache.set(cacheKey, resp, ttl)
+
+			return resp, nil
+		}
+
+		return resp, err
+	})
+}
+
+func buildCacheKey(r *http.Request) string {
+	h := sha256.New()
+
+	fmt.Fprint(h, r.URL.String())
+	fmt.Fprint(h, r.Header.Get("Authorization"))
+
+	if r.Body != nil {
+		body, _ := io.ReadAll(r.Body)
+		r.Body = io.NopCloser(bytes.NewReader(body))
+		h.Write(body)
+	}
+
+	return hex.EncodeToString(h.Sum(nil))
+}
--- a/pkg/clients/dynatrace/middleware/roundtripper.go
+++ b/pkg/clients/dynatrace/middleware/roundtripper.go
@@ -31,9 +31,7 @@ func NewCacheRoundTripper(next http.RoundTripper, ttl time.Duration) http.RoundT
 
 		cacheKey := buildCacheKey(r)
 
-		if ttl == 0 {
-			log.Debug("cache ttl is 0, no cache will be used")
-
+		if ttl == 0 || r.Header.Get(core.CacheSkipHeader) != "" {
 			// if the caching was turned off intermittently, those entries should be removed
 			cache.remove(cacheKey)
 
--- a/pkg/clients/dynatrace/middleware/roundtripper_test.go
+++ b/pkg/clients/dynatrace/middleware/roundtripper_test.go
@@ -78,6 +78,31 @@ func TestNewCacheRoundTripper(t *testing.T) {
 		assert.Equal(t, 2, *calls, "POST must always reach backend")
 	})
 
+	t.Run("CacheSkipHeader bypasses cache and evicts existing entry", func(t *testing.T) {
+		rt, calls := makeCachedRT(t, time.Minute,
+			fakeResponse("r1"),
+			fakeResponse("r2"),
+			fakeResponse("r3"),
+		)
+
+		// Populate cache
+		_, err := rt.RoundTrip(newRequest(t, http.MethodGet, endpoint))
+		require.NoError(t, err)
+		assert.Equal(t, 1, *calls)
+
+		// Call with skip header — evicts cached entry and calls backend
+		req2 := newRequest(t, http.MethodGet, endpoint)
+		req2.Header.Set(core.CacheSkipHeader, "true")
+		_, err = rt.RoundTrip(req2)
+		require.NoError(t, err)
+		assert.Equal(t, 2, *calls, "skip header must bypass the cache")
+
+		// Next call without skip — cache was evicted, backend is called again
+		_, err = rt.RoundTrip(newRequest(t, http.MethodGet, endpoint))
+		require.NoError(t, err)
+		assert.Equal(t, 3, *calls, "evicted entry must not be served from cache")
+	})
+
 	t.Run("zero TTL disables caching", func(t *testing.T) {
 		rt, calls := makeCachedRT(t, 0,
 			fakeResponse("r1"),
--- a/pkg/clients/dynatrace/middleware/roundtripper_test.go
+++ b/pkg/clients/dynatrace/middleware/roundtripper_test.go
@@ -1,0 +1,392 @@
+package middleware
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"net/http"
+	"testing"
+	"testing/synctest"
+	"time"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/core"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewCacheRoundTripper(t *testing.T) {
+	const endpoint = "http://api.example.com/v1/resource"
+
+	newRequest := func(t *testing.T, method, rawURL string) *http.Request {
+		t.Helper()
+		r, err := http.NewRequest(method, rawURL, nil)
+		require.NoError(t, err)
+
+		return r
+	}
+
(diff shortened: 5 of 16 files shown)


cmd/operator/cmd.go
```diff
@@ -80,6 +86,9 @@ func runLocally(ctx context.Context, kubeCfg *rest.Config) error {
 	if err != nil {
 		return err
 	}
+	// TODO: make configurable
+	// Added this here as a formality, doesn't really make much sense to this while debugging
```
StefanHauth: Could this even disturb debugging? If the program is halted too long and an unexpected cleanup is triggered meanwhile?

pkg/clients/dynatrace/middleware/cache.go
```diff
@@ -0,0 +1,124 @@
+package middleware
```
StefanHauth, line 1: So far, this package only deals with cache stuff. Wouldn't that justify calling it "cache"?

pkg/clients/dynatrace/middleware/roundtripper.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (rt RoundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	if rt == nil {
+		return http.DefaultTransport.RoundTrip(r)
+	}
+
+	return rt(r)
+}
+
+func CacheRoundTripper(next http.RoundTripper, ttl time.Duration) http.RoundTripper {
```
StefanHauth:
```suggestion
func NewCacheRoundTripper(next http.RoundTripper, ttl time.Duration) http.RoundTripper {
```

pkg/clients/dynatrace/middleware/roundtripper.go
```diff
@@ -0,0 +1,63 @@
...
+		if cachedResponse != nil {
+			log.Info("client using cached response", "endpoint", r.URL)
+
+			return cachedResponse, nil
+		}
+
+		// send the actual request
+		resp, err := next.RoundTrip(r)
+		if err == nil {
+			cache.store(cacheKey, resp, ttl)
```
jokuniew: question: if get request returns non 2XX return code, should it be cached?
Now if I mock server to return 5xx on first call and 2xx on 2nd call - i get cached 5xx response instead of 2xx one on 2nd call (i assume that we have exp backoff retry or just another reconcile)

0sewa0: good point, should use the error handling we have already for the API somehow, to be consistent

pkg/clients/dynatrace/middleware/cache.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	// Reconstruct a fresh body reader so every caller gets a full, open body.
+	resp := *entry.response
+	resp.Body = io.NopCloser(bytes.NewReader(entry.body))
+
+	return &resp
+}
+
+func (rc *responseCache) store(key string, response *http.Response, ttl time.Duration) {
```
andriisoldatenko:
```suggestion
func (rc *responseCache) set(key string, response *http.Response, ttl time.Duration) {
```

andriisoldatenko: just by convention it's usually get/set

cmd/operator/cmd.go
0sewa0: removed it everywhere where it makes no sense + added a doc string to the func explaining when to use it

pkg/clients/dynatrace/middleware/roundtripper.go
0sewa0: done

pkg/clients/dynatrace/middleware/roundtripper.go
0sewa0: should be fixed now

pkg/clients/dynatrace/middleware/cache.go
0sewa0: done

pkg/clients/dynatrace/middleware/cache.go
0sewa0: I like my middleware folder, but I wouldn't be against putting it in `pkg/clients/dynatrace/middleware/cache/...`

WDYT?

avorima: This cache doesn't need to be exposed it's only used for the middleware. Also "cache" is a very overloaded package name

pkg/clients/dynatrace/middleware/cache.go
```diff
@@ -0,0 +1,137 @@
...
+	cache = newCache()
+)
+
+// cacheEntry holds all relevant info for a cached response.
+// The only not obvious part is that the `body` has to be stored separate from the actual `response`.
+// The reason for this is that the `Body` part of the `http.Response` is an `io.ReadCloser`.
+// This means the you may only read the body once, which is not good if we would like to reuse the `http.Response`
+// To combat this we read the `Body` while creating the `cacheEntry` and store it inside this new entry.
+// We always "recreate"(put it into a new Reader) the `Body` in the affected `http.Response` or when we returned the cached `http.Response`.
+type cacheEntry struct {
```
avorima, line 27: have you considered using an existing LRU cache implementation?

0sewa0: you mean like https://github.com/hashicorp/golang-lru ?

after writing it, yes (wanted to know how complex it is)
but the one I linked is very generic, I doubt it works with `http.Response` due to the hack I had to do.
and I didn't find any http caching impl.

but to be honest, didn't look that deeply into it, I found it simple enough without having to rely on an external dependency 🤷

avorima: I'm a defender of the "a little boilerplate is better than a little dependency" go proverb, but this is not just boilerplate. What the cache is doing is not really complex and translates well to using a generic cache like the one you linked. I would also not expect anyone outside our team to object to adding this dependency since it's well established.
While I agree that the current picture is very simple we know full well how things can devolve if an opportunity arises. Given that this is an active area of development I think it's a valid concern.
We can always go back and implement a custom cache if the need arises.

pkg/clients/dynatrace/middleware/roundtripper.go
```diff
@@ -0,0 +1,70 @@
...
+			ttl == 0 ||
+			r.Header.Get("Accept") == "application/octet-stream" {
+			return next.RoundTrip(r)
+		}
+
+		cacheKey := buildCacheKey(r)
+
+		cachedResponse := cache.get(cacheKey)
+		if cachedResponse != nil {
+			log.Info("client using cached response", "endpoint", r.URL)
```
avorima: this cache hit/miss and store logs duplicate a lot of the core client logs. the client log is just missing the information whether the response was cached or not, which could be done by injecting a header into the response. WDYT?

pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -93,3 +96,14 @@ func GetNodeName() string {
...
 }
+
+func GetDTClientCacheCleanInterval() time.Duration {
+	rawDuration := os.Getenv(DTClientCacheCleanInterval)
+
+	parsedDuration, err := time.ParseDuration(rawDuration)
+	if err != nil {
+		return defaultDTClientCacheCleanInterval
+	}
+
```
aorcholski, line 124: we should check allowed range (1ns Go scheduler may be a little bit busy, 2400h = disabled)

pkg/clients/dynatrace/middleware/cache.go
0sewa0: did a quick check for science, checked the hashicorp and the k8s util one, neither can handle http.Response 
https://github.com/Dynatrace/dynatrace-operator/commit/d16d6b15de53bbba6d65311d418ad436bcb6fccc

pkg/clients/dynatrace/middleware/cache.go
0sewa0: I will leave it where it is

pkg/clients/dynatrace/middleware/roundtripper.go
0sewa0: done

pkg/util/kubernetes/fields/k8senv/env.go
0sewa0: done

pkg/clients/dynatrace/middleware/cache.go
avorima: the http response body handling is not related to caching. you can still store both fields in a cache entry. i'm just talking about the container

0sewa0: but then you have to do the same or more boilerplate but probably more confusing, because you would have 2 caches, one for the responses and one for the bodies, and then our boilerplate would combine it.

You would still need a mutex, a custom cleanup logic (the hashycorp one wants the ttl at cache init, which we do not know, as that is defined in the dynakube)

pkg/clients/dynatrace/middleware/roundtripper.go
avorima: there's still the error handling logs which are duplicated

pkg/clients/dynatrace/middleware/cache.go
avorima: there's no boilerplate when using the LRU cache, you're just left with converting the inputs and outputs

```go
type cacheEntry struct {
    resp *http.Response
    body []byte
}

type reponseCache struct {
    *expirable.LRU[string, cacheEntry]
}
```

but you're right, having a shared cache doesn't work well with the dynamic TTL from the DynaKube. I guess that's the actual make or buy argument here

pkg/clients/dynatrace/middleware/cache.go
```diff
@@ -0,0 +1,139 @@
...
+	ttl          time.Duration
+}
+
+func (ce *cacheEntry) isOutdated() bool {
+	return time.Since(ce.creationTime) > ce.ttl
+}
+
+type responseCache struct {
+	entries map[string]*cacheEntry
+	mu      *sync.Mutex
```
avorima: use mutex values unless the containing struct is supposed to be copyable

pkg/clients/dynatrace/middleware/cache.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (rc *responseCache) set(key string, response *http.Response, ttl time.Duration) {
+	if response == nil {
+		return
+	}
+
+	body, err := io.ReadAll(response.Body)
+	if err != nil {
+		return
```
avorima: this error needs to fail the round trip otherwise the client may get an OK response with an empty body

pkg/clients/dynatrace/middleware/cache.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	response.Body.Close()
```
avorima, line 88: close should also happen if reading fails

pkg/clients/dynatrace/middleware/cache.go
```diff
@@ -0,0 +1,139 @@
...
+
+// RunPeriodicCacheCleanup creates a small goroutine that will remove all expired entries from the dtclient cache.
+// This is necessary because if the user changes tokens, disables features etc., some entries will never be hit again,
+// so we can't clean them up during normal reconciles.
+// This only makes sense to use in the Operator Pod.
+// - The bootstrap command only runs once, and can only download a ~700MB archive. (can be less or more, depending on the technologies present in the archive)
+// - The csidriver (mainly `provisioner` container) does not make "in-memory cacheable" requests, it only downloads +700MB archives, that we "cache" on the node.
+// - The webhook (and any other not listed command) does not use the dtclient.
+func RunPeriodicCacheCleanup(ctx context.Context, period time.Duration) {
+	go func() {
```
avorima: it's better to let the caller control concurrency. the unit test shows that you can't really be sure the goroutine has exited otherwise (without introspecting the runtime)

pkg/clients/dynatrace/middleware/cache_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// freshCache returns a new responseCache and replaces the package-level one,
+// preventing state from leaking between tests.
+func freshCache(t *testing.T) *responseCache {
+	t.Helper()
+	c := newCache()
+	cache = c
+	t.Cleanup(func() { cache = newCache() })
```
avorima: you're resetting the cache twice. wouldn't the cleanup be enough after each test?

pkg/clients/dynatrace/middleware/cache_test.go
```diff
@@ -0,0 +1,157 @@
...
+		require.NotNil(t, resp)
+		body, _ := io.ReadAll(resp.Body)
+		assert.Equal(t, "hello", string(body))
+	})
+	t.Run("expired entry is evicted on get and is a miss", func(t *testing.T) {
+		c := freshCache(t)
+
+		synctest.Test(t, func(t *testing.T) {
+			c.set("key", fakeResponse("data"), time.Minute)
+			time.Sleep(5 * time.Minute)
```
avorima: please use something shortly after the expiry. this test would still pass if someone accidentally added some time to the isOutdated

pkg/clients/dynatrace/middleware/cache_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+			entry := c.get("key") // triggers eviction
+			assert.Nil(t, entry, "expired entry must be treated as a miss")
+		})
+
+		c.mu.Lock()
```
avorima: locking is only needed in parallel testing

pkg/clients/dynatrace/middleware/cache_test.go
```diff
@@ -0,0 +1,157 @@
...
+		assert.Empty(t, c.entries)
+	})
+}
+
+func TestRunPeriodicCacheCleanup(t *testing.T) {
+	t.Run("evicts expired entries after a tick", func(t *testing.T) {
+		c := freshCache(t)
+
+		synctest.Test(t, func(t *testing.T) {
+			c.set("key", fakeResponse("data"), time.Nanosecond)
```
avorima: let's use realistic time values. we don't have to pay for them in the synctest bubble

pkg/clients/dynatrace/middleware/cache_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+			RunPeriodicCacheCleanup(t.Context(), time.Millisecond)
+
+			time.Sleep(time.Hour)
```
avorima: this also is a bit overblown. we really only want to check that it runs once, maybe twice

pkg/clients/dynatrace/middleware/cache_test.go
```diff
@@ -0,0 +1,157 @@
...
+		c.mu.Lock()
+		remaining := len(c.entries)
+		c.mu.Unlock()
+
+		assert.Equal(t, 0, remaining)
+	})
+
+	t.Run("goroutine exits when context is canceled", func(t *testing.T) {
+		freshCache(t)
+		ctx, cancel := context.WithCancel(context.Background())
```
avorima: contexts can depend on t.Context()

pkg/clients/dynatrace/middleware/cache_test.go
```diff
@@ -0,0 +1,157 @@
...
+		_, freshExists := c.entries["fresh"]
+		c.mu.Unlock()
+
+		assert.False(t, expiredExists)
+		assert.True(t, freshExists)
+	})
+
+	t.Run("does nothing on empty cache", func(t *testing.T) {
+		c := freshCache(t)
+		c.cleanup() // must not panic
```
avorima: there's `assert.NotPanics`

pkg/clients/dynatrace/middleware/cache_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+			entry := c.get("key") // triggers eviction
+			assert.Nil(t, entry, "expired entry must be treated as a miss")
+		})
+
+		c.mu.Lock()
+		_, exists := c.entries["key"]
+		c.mu.Unlock()
+
+		assert.False(t, exists, "expired entry must be deleted from the map")
```
avorima: all of these could be `assert.Contains/NotContains`

pkg/clients/dynatrace/middleware/roundtripper.go
```diff
@@ -0,0 +1,65 @@
...
+	"time"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/core"
+)
+
+type RoundTripperFunc func(*http.Request) (*http.Response, error)
+
+func (rt RoundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	if rt == nil {
+		return http.DefaultTransport.RoundTrip(r)
```
avorima: this isn't covered and also feels like it deserves a panic

pkg/clients/dynatrace/middleware/roundtripper.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	return rt(r)
+}
+
+func NewCacheRoundTripper(next http.RoundTripper, ttl time.Duration) http.RoundTripper {
+	return RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
+		if r.Method != http.MethodGet ||
+			ttl == 0 ||
```
avorima: the ttl is constant so you can short-circuit this by returning next directly instead of the caching roundtripper

pkg/clients/dynatrace/middleware/roundtripper_test.go
```diff
@@ -0,0 +1,309 @@
...
+	"time"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/core"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func newRequest(t *testing.T, method, rawURL string) *http.Request {
+	t.Helper()
+	u, err := url.Parse(rawURL)
```
avorima: this seems like a roundabout way of calling `http.NewRequest(method, rawURL, nil)`

pkg/clients/dynatrace/middleware/roundtripper_test.go
```diff
@@ -0,0 +1,309 @@
...
+		resp2, err := rt.RoundTrip(newRequest(t, http.MethodGet, endpoint))
+		require.NoError(t, err)
+		assert.Equal(t, http.StatusOK, resp2.StatusCode)
+
+		assert.Equal(t, 2, calls, "error response must not be cached; second call must reach backend")
+	})
+}
+
+func TestBuildCacheKey(t *testing.T) {
+	makeReq := func(t *testing.T, rawURL, auth string, body []byte) *http.Request {
```
avorima, line 335: this looks even more like http.NewRequest

pkg/clients/dynatrace/client_v2.go
```diff
@@ -166,6 +168,18 @@ func WithUserAgentSuffix(suffix string) OptionV2 {
 	}
 }
 
+func WithCacheTTL(ttl time.Duration) OptionV2 {
+	return func(c *ConfigV2) error {
+		if ttl == 0 {
+			log.Debug("cache ttl is 0, no cache will be used")
```
avorima: if this really should be logged it should be in roundtripper.go

pkg/controllers/dynakube/token/token.go
```diff
@@ -12,8 +12,8 @@ import (
 )
 
 type Token struct {
-	Type     string
-	Value    string
+	Type     string `json:"type"`
```
avorima: i don't think we're planning on marshaling this anywhere. also, omitting a json tag does not omit it from the output

pkg/controllers/nodes/nodes_controller.go
```diff
@@ -178,6 +179,12 @@ func (controller *Controller) sendMarkedForTermination(ctx context.Context, dk *
 		return err
 	}
 
+	// NOTE: The dtclient in-memory cache is intentionally disabled here.
+	// `GetEntityIDForIP` fetches ALL host entities for the entire tenant, not just those belonging to this Kubernetes cluster.
+	// Which means a single call can pull in data from every cluster and environment reporting to the same tenant.
+	// Mark-for-termination events are rare, caching this possibly large dataset would waste memory with no meaningful benefit.
+	dk.Spec.DynatraceAPIRequestThreshold = ptr.To(uint16(0))
```
avorima, line 186: this should really be handled by the builder, e.g. a DisableCache option, but it can wait until we're done with the refactoring

pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -93,3 +100,22 @@ func GetNodeName() string {
 func GetCSIDataDir() string {
 	return os.Getenv(CSIDataDir)
 }
+
+func GetDTClientCacheCleanInterval(log logd.Logger) time.Duration {
+	rawDuration := os.Getenv(DTClientCacheCleanInterval)
```
avorima, line 105: an empty env var should just use the default value without logging an error

test/e2e/features/logmonitoring/logmonitoring.go
```diff
@@ -170,7 +171,7 @@ func triggerDaemonSetReconcile(dk dynakube.DynaKube) features.Func {
 
 		require.NoError(t, resources.Get(ctx, dk.Name, dk.Namespace, &dk))
 		// Force reconciliation by simulating the passage of time
-		dk.Status.DynatraceAPI.LastTokenScopeRequest.Time = dk.Status.DynatraceAPI.LastTokenScopeRequest.Add(-2 * dk.APIRequestThreshold())
+		dk.Spec.DynatraceAPIRequestThreshold = ptr.To(uint16(1))
```
avorima: this is permanent unlike the status update. is this even still needed now that the conditions are gone?

avorima (commented):
it would've been better to split up the cache from the removal of the timeprovider

test/e2e/features/logmonitoring/logmonitoring.go
0sewa0: AFAIK, this test wants to force a reconcile, so it needs to change something, changing that value is safe

0sewa0:
@avorima 
> it would've been better to split up the cache from the removal of the timeprovider

I only removed it where:
- it was used to throttle the dtclient (hard to test a cache, if there is another throttling mechanism blocking me)
- where it is not used for anything, it is "just there"

There are a few places where it is still used (like checking cert expirations and such), I didn't remove it here, as it is not related to my change and is not a quick safe fix.

pkg/clients/dynatrace/middleware/roundtripper.go
0sewa0: I will also make `RoundTripperFunc` private, it has no reason being public

pkg/clients/dynatrace/middleware/cache.go
0sewa0: feels a bit overkill for this, because nothing needs to wait for this to exit, only in tests.

There is no special thing we do if the context gets canceled, there is no graceful shutdown for RunPeriodicCacheCleanup, as everything is in memory.

avorima: it's just moving the `go` outside of the function, but the effect is that the caller can know exactly when the function exited. it's good API design

test/e2e/features/logmonitoring/logmonitoring.go
avorima: it's unfortunate that we lose the "ignore cache for one reconcile" functionality with this change. it's a nice feature to have

0sewa0: Related to this, I am using the `DynatraceAPIRequestThreshold` for the default `requeueAfter` as well.

I thought this was a nice idea, to no have that many fields in the DynaKube that are influencing how often we make requests.

But by doing this, I made the cache annoying to turn off. -> you go into an infinite reconcile loop

So, I want to introduce a `requeueAfter` field in the spec so the user can configure that, and leave `DynatraceAPIRequestThreshold` to be just the TTL (which we could rename in the next apiVersion bump)
- in a separate ticket
- In this PR, I would only undo my "using the `DynatraceAPIRequestThreshold` for the default `requeueAfter` as well." change

This way you can at least safely turn off the cache, without totally messing up the requeueAfter

WDYT?

0sewa0: this also made me think of adding some logic to support the "turn the cache on/off" behavior, dunno if we need something like this 🤔 
```diff
diff --git a/pkg/clients/dynatrace/middleware/cache.go b/pkg/clients/dynatrace/middleware/cache.go
index 9c350fc26..4187daa32 100644
--- a/pkg/clients/dynatrace/middleware/cache.go
+++ b/pkg/clients/dynatrace/middleware/cache.go
@@ -96,6 +96,13 @@ func (rc *responseCache) set(key string, response *http.Response, ttl time.Durat
 	return nil
 }
 
+func (rc *responseCache) remove(key string) {
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
+
+	delete(rc.entries, key)
+}
+
 func (rc *responseCache) cleanup() {
 	rc.mu.Lock()
 	defer rc.mu.Unlock()
diff --git a/pkg/clients/dynatrace/middleware/roundtripper.go b/pkg/clients/dynatrace/middleware/roundtripper.go
index 9f211e73d..b2a2537d7 100644
--- a/pkg/clients/dynatrace/middleware/roundtripper.go
+++ b/pkg/clients/dynatrace/middleware/roundtripper.go
@@ -23,12 +23,6 @@ func (rt roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 }
 
 func NewCacheRoundTripper(next http.RoundTripper, ttl time.Duration) http.RoundTripper {
-	if ttl == 0 {
-		log.Debug("cache ttl is 0, no cache will be used")
-
-		return next
-	}
-
 	return roundTripperFunc(func(r *http.Request) (*http.Response, error) {
 		if r.Method != http.MethodGet ||
 			r.Header.Get("Accept") == "application/octet-stream" {
@@ -37,6 +31,15 @@ func NewCacheRoundTripper(next http.RoundTripper, ttl time.Duration) http.RoundT
 
 		cacheKey := buildCacheKey(r)
 
+		if ttl == 0 {
+			log.Debug("cache ttl is 0, no cache will be used")
+
+			// if the caching was turned off intermittently, those entries should be removed
+			cache.remove(cacheKey)
+
+			return next.RoundTrip(r)
+		}
+
 		cachedResponse := cache.get(cacheKey)
 		if cachedResponse != nil {
 			cachedResponse.Header.Set(core.CacheHitHeader, "true")

```

jdanieck:
> @avorima
> 
> > it would've been better to split up the cache from the removal of the timeprovider
> 
> I only removed it where:
> 
> * it was used to throttle the dtclient (hard to test a cache, if there is another throttling mechanism blocking me)
> * where it is not used for anything, it is "just there"
> 
> There are a few places where it is still used (like checking cert expirations and such), I didn't remove it here, as it is not related to my change and is not a quick safe fix.

I attempted to review this, but had to stop after about an hour due to the amount of unrelated changes, which makes the review unnecessarily difficult. The caching middleware is well isolated and should be introduced in a separate PR, together with its unit tests and any design discussion. The removals and cleanup can then follow in a subsequent PR.

We need to be more intentional about review-ability when creating PRs. While splitting changes requires more effort from the author, it significantly reduces review time and improves overall team efficiency.

test/e2e/features/logmonitoring/logmonitoring.go
avorima: +1 for both ideas. clearing the cache when TTL is 0 also makes sense, but it's not that critical given that it will get evicted over time

avorima: sorry, didn't read this thoroughly enough. i'm somewhat opposed to adding this as a spec field. most of our issues with the current state of the CRD is because we just piled onto it. the DynaKube is supposed to specify how the dynatrace resources are deployed and requeueing has nothing to do with that

0sewa0: We can discuss it how we make the default requeue configurable, I didn't create a ticket for it yet, so all good, I will write the ticket accordingly.

pkg/clients/dynatrace/middleware/roundtripper.go
```diff
@@ -0,0 +1,73 @@
...
+		if cachedResponse != nil {
+			cachedResponse.Header.Set(core.CacheHitHeader, "true")
+
+			return cachedResponse, nil
+		}
+
+		// send the actual request
+		resp, err := next.RoundTrip(r)
+		if err == nil && core.IsSuccessResponse(resp) {
+			return resp, cache.set(cacheKey, resp, ttl)
```
StefanHauth: If setting the cache fails, a succesfull response is discarded, because the error of cache.set is returned and the caller would by convention ignores the response.
```suggestion
            err := cache.set(cacheKey, resp, ttl)
            if err != nil {
               log.Error(err, "could not cache response")
             }
			return resp, nil
```

pkg/clients/dynatrace/middleware/cache.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+// cacheEntry holds all relevant info for a cached response.
+// The only not obvious part is that the `body` has to be stored separate from the actual `response`.
+// The reason for this is that the `Body` part of the `http.Response` is an `io.ReadCloser`.
+// This means the you may only read the body once, which is not good if we would like to reuse the `http.Response`
+// To combat this we read the `Body` while creating the `cacheEntry` and store it inside this new entry.
+// We always "recreate"(put it into a new Reader) the `Body` in the affected `http.Response` or when we returned the cached `http.Response`.
+type cacheEntry struct {
+	response     *http.Response
```
StefanHauth, line 28: Do we need to store a reference to the whole response (which also contains a reference to the original request)? Wouldn't it be more efficient and cleaner to only store what we really want to cache?

pkg/clients/dynatrace/middleware/roundtripper.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func buildCacheKey(r *http.Request) string {
+	h := sha256.New()
+
+	fmt.Fprint(h, r.URL.String())
+	fmt.Fprint(h, r.Header.Get("Authorization"))
+
+	if r.Body != nil {
+		body, _ := io.ReadAll(r.Body)
```
StefanHauth, line 72: Is it ok to swallow the error? Doesn't that error indicate that there's something wrong with the request and that it can't be sent anyways (if the request body is unreadable)?

StefanHauth (dismissed):

pkg/clients/dynatrace/middleware/roundtripper.go
avorima: the request body is always a byte slice, so this boils down to `copy(body, r.Body)`

pkg/clients/dynatrace/middleware/cache.go
avorima: good point, the request should be injected into the cached response in the roundtripper.

i'm a bit hesitant to pick and select field to copy from the response, because it's easy to forget that we're doing it which can lead to expecting fields to be set when they're not. OTOH, having a clean response feels like the correct thing to do. the body is the bulk of the cache data, so storing the whole response struct doesn't really impact efficiency that much

pkg/clients/dynatrace/middleware/roundtripper.go
0sewa0: I will just remove the error check, because right now, the only thing that can fail in `set` is the reading of the body, if the cache can't read the body, the original caller will also fail to do so, so the outcome is kinda the same.

I wanted to introduce a Debug log instead, but the current client package will log an error anyway, so it will be just noise.

If / when we add some logic to `cache.set` that can fail independently, then this can be revised.

pkg/clients/dynatrace/middleware/cache.go
0sewa0: for now I added the `the request should be injected` logic, we can optimize what we store in the cache later if necessary.
I agree that the majority of the info is the Body, so storing the rest has no noticable performance penalties, but greatly simplifies our life.

pkg/controllers/nodes/nodes_controller.go
jdanieck, line 186: Nit: using TTL = 0 to effectively disable caching feels a bit indirect. We could consider an explicit .EnableCache(...) at the builder level that wires in the caching middleware which would make the intent clearer.

pkg/clients/dynatrace/middleware/roundtripper.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (rt roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	if rt == nil {
+		panic("empty roundTripperFunc used")
+	}
+
+	return rt(r)
+}
+
+func NewCacheRoundTripper(next http.RoundTripper, ttl time.Duration) http.RoundTripper {
```
jdanieck, line 25: Nit: assuming a client‑level cache is sufficient, we could create the cache inside NewCacheRoundTripper. Since the cache is inherently tied to the round tripper (and thus the client), this would remove the need for a package‑level cache variable, make ownership and intent clearer, and likely simplify unit tests.

jdanieck (dismissed):
LGTM — dropped a couple of design nits (non‑blocking, possibly just taste).

pkg/clients/dynatrace/middleware/roundtripper.go
avorima: we need the cache to persist past a single reconcile, so this is the same as not caching at all. we don't really do many of the same requests in a single reconcile

pkg/controllers/nodes/nodes_controller.go
avorima: already mentioned here. https://github.com/Dynatrace/dynatrace-operator/pull/6427#discussion_r3064907226. i will create a ticket to track it, since we're already piling up so much on the last refactoring ticket
ICP-4002

avorima:
please merge latest changes from main. this branch causes a NPE atm

pkg/clients/dynatrace/middleware/roundtripper.go
StefanHauth: But will we think about this then? IMHO better do it right from start.

avorima:
another panic after merge due to type assertion of http.Transport in `AsV2`. i don't know if you want to fix this, since it will be gone after https://github.com/Dynatrace/dynatrace-operator/pull/6482 is merged

pkg/clients/dynatrace/middleware/roundtripper.go
jdanieck: I assumed dt client is created at controller creation (like k8s client), thus persists across reconciles; isn't it the case ? if so, why not ? : )

avorima: the client is constructed from a dynakube, which could be different on every reconcile. you're free to challenge this, but let's unblock the PR and take this discussion elsewhere

avorima (dismissed):
Deployed operator and deleted some conditions to force a cached lookup


| Timestamp | Duration | Path | Cached |
| --- | --- | --- | --- |
| 2026-04-22T07:55:53.437Z | 34.642461ms | /api/v2/settings/objects                                        | false |
| 2026-04-22T07:55:53.503Z | 64.710235ms | /api/v1/deployment/installer/gateway/connectioninfo             | false |
| 2026-04-22T07:55:53.604Z | 40.782248ms | /api/v1/deployment/installer/gateway/unix/latest/metainfo       | false |
| 2026-04-22T07:56:23.086Z | 33.314889ms | /api/v2/settings/objects                                        | false |
| 2026-04-22T07:56:23.181Z | 46.792372ms | /api/v1/deployment/installer/agent/connectioninfo               | false |
| 2026-04-22T07:56:23.436Z | 52.535818ms | /api/v1/deployment/installer/agent/unix/paas/latest/metainfo    | false |
| 2026-04-22T07:56:23.491Z | 41.775806ms | /api/v2/settings/effectiveValues                                | false |
| 2026-04-22T07:56:23.816Z | 63.720689ms | /api/v1/deployment/installer/agent/processmoduleconfig          | false |
| 2026-04-22T07:56:24.275Z | 42.26845ms  | /api/v1/deployment/installer/agent/unix/default/latest/metainfo | false |
| 2026-04-22T08:03:27.434Z | 253.989µs   | /api/v1/deployment/installer/gateway/connectioninfo             | true  |
| 2026-04-22T08:03:27.710Z | 87.389µs    | /api/v1/deployment/installer/agent/connectioninfo               | true  |
| 2026-04-22T08:03:27.916Z | 414.164µs   | /api/v1/deployment/installer/agent/processmoduleconfig          | true  |
| 2026-04-22T08:18:30.441Z | 35.093141ms | /api/v2/settings/objects                                        | false |
| 2026-04-22T08:18:30.475Z | 33.619736ms | /api/v1/deployment/installer/gateway/connectioninfo             | false |
| 2026-04-22T08:18:30.826Z | 48.112191ms | /api/v2/settings/objects                                        | false |
| 2026-04-22T08:18:30.875Z | 32.836709ms | /api/v1/deployment/installer/agent/connectioninfo               | false |
| 2026-04-22T08:18:30.962Z | 33.976847ms | /api/v1/deployment/installer/agent/unix/paas/latest/metainfo    | false |
| 2026-04-22T08:18:31.023Z | 50.354879ms | /api/v2/settings/effectiveValues                                | false |
| 2026-04-22T08:18:31.177Z | 32.400213ms | /api/v1/deployment/installer/agent/processmoduleconfig          | false |
| 2026-04-22T08:18:31.546Z | 33.070441ms | /api/v1/deployment/installer/agent/unix/default/latest/metainfo | false |

avorima (approved):

jdanieck (approved):

PR #6440, opened 2026-04-09
Replace hardcoded image versions in E2E tests

s-quatres:
## Description
[ICP-3298](https://dt-rnd.atlassian.net/browse/ICP-3298) This change is to replace hardcoded E2E test image versions with latest or recent stable tag. This way we test with a more recent version of components.

## How can this be tested?
This can be tested by running the relevant E2E tests from this branch and ensuring they get the newest image (should now be logged too).

[ICP-3298]: https://dt-rnd.atlassian.net/browse/ICP-3298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/e2e/helpers/components/dynakube/options.go
```diff
@@ -292,33 +298,44 @@ func WithExtensionsDatabases(databases ...extensions.DatabaseSpec) Option {
...
-// If the image differs from the default value, the custom pull secret is set on the DynaKube.
+// setImageRefFromEnvs populates the image.Ref from an environment variable, falling back to the latest image from the registry.
+// If the image repo differs from the default repo, the custom pull secret is set on the DynaKube.
 // Returns true, if the pull secret was set.
-func setImageRefFromEnvs(dk *dynakube.DynaKube, image *image.Ref, envVar, defaultValue string) bool {
-	value := os.Getenv(envVar)
-	if value == "" {
-		value = defaultValue
-	}
+func setImageRefFromEnvs(t *testing.T, dk *dynakube.DynaKube, imageRef *image.Ref, envVar, defaultRepo string) bool {
```
waodim:
```suggestion
func setImageRefFromEnv(t *testing.T, dk *dynakube.DynaKube, imageRef *image.Ref, envVar, defaultRepo string) bool {
```

test/e2e/helpers/components/dynakube/options.go
```diff
@@ -292,33 +298,44 @@ func WithExtensionsDatabases(databases ...extensions.DatabaseSpec) Option {
...
+	t.Helper()
+
+	envImage := os.Getenv(envVar)
+	if envImage != "" {
+		repo, tag, _ := strings.Cut(envImage, ":")
```
waodim: Nit: But I think we still should check the found bool var. WDYT?

andriisoldatenko: I think in this case we can just check empty string.

test/e2e/helpers/components/dynakube/options.go
andriisoldatenko: `setImageRefFromEnvsOrDefault`

test/e2e/helpers/registry/registry.go
```diff
@@ -22,40 +27,59 @@ const (
...
+			t.Logf("using image from env %s: %s", envVar, val)
+
+			return val
+		}
+	}
+
+	if uri, ok := latestImageURIs[repoURI]; ok {
+		t.Logf("using cached resolved newest image: %s", uri)
+
+		return uri
```
andriisoldatenko, line 56: ~Let's print to logs, iamge we are using, so we can see image we install in logs, (helps for troubleshooting)~

andriisoldatenko: something similar https://github.com/Dynatrace/dynatrace-operator/pull/6505 but for components we are rollout (AG, EC, CM etc) so we can do it here.

andriisoldatenko: nvm you already have t.Logf

test/e2e/helpers/registry/registry.go
```diff
@@ -64,7 +88,26 @@ func getLatestImageURI(t *testing.T, repoURI string) string {
 	repo, err := name.NewRepository(repoURI)
 	require.NoError(t, err)
 
-	tags, err := remote.List(repo)
+	var tags []string
+	for attempt := range 3 {
```
andriisoldatenko: you can use what we use for our production code https://github.com/dynatrace/dynatrace-operator/blob/48e72deafd2fe698c20d795d13459e996e718971/pkg/controllers/dynakube/activegate/service.go#L37-L47 instead of custom retry.

test/e2e/helpers/registry/registry.go
```diff
@@ -64,7 +88,26 @@ func getLatestImageURI(t *testing.T, repoURI string) string {
...
+		tags, err = remote.List(repo)
+		if err == nil {
+			break
+		}
+
+		var transportErr *transport.Error
+		if errors.As(err, &transportErr) && transportErr.StatusCode == http.StatusTooManyRequests {
+			wait := time.Duration(5*(attempt+1)) * time.Second
```
mihaitanasedt: After the last attempt the code sleeps for 15 seconds and then exits the loop wasting the sleep time. We shouldn't sleep after the last attempt, as it will never try again

test/e2e/helpers/registry/registry.go
```diff
@@ -82,7 +125,6 @@ func getLatestImageURI(t *testing.T, repoURI string) string {
 
 		return semver.Compare(semverA, semverB)
 	})
-	require.NoError(t, err)
```
mihaitanasedt, line 85: We can require the filteredTags to no be empty. So it will not panic if something was wrong and no tag is found. ex:
```go
require.NotEmpty(t, filteredTags)
```

test/e2e/helpers/registry/registry.go
```diff
@@ -22,40 +27,59 @@ const (
 	cmPublicECR = "public.ecr.aws/dynatrace/dynatrace-codemodules"
 )
 
-var (
-	latestActiveGateURI string
-	latestOneAgentURI   string
-	latestCodeModuleURI string
-)
+// repoEnvVars maps ECR repository URIs to env var overrides for GetLatestImageURI.
+var repoEnvVars = map[string]string{
```
mihaitanasedt: Should we add the ECRs and envs to the `consts.go` like other images?

test/e2e/helpers/components/dynakube/options.go
s-quatres: @andriisoldatenko - Albian mentioned above "...FromEnv" instead "...FromEnvs", so FromEnv looks right, and "OrDefault" implies a fixed default, but we get the newest image if no env var set. "OrLatest" would imply "latest", so how about setImageRefFromEnvOrNewest(...) that shows it is env var or newest image?

andriisoldatenko: `setImageRefFromEnvOrLatest` is ok

test/e2e/helpers/components/dynakube/options.go
```diff
@@ -292,33 +298,44 @@ func WithExtensionsDatabases(databases ...extensions.DatabaseSpec) Option {
...
 // Returns true, if the pull secret was set.
-func setImageRefFromEnvs(dk *dynakube.DynaKube, image *image.Ref, envVar, defaultValue string) bool {
-	value := os.Getenv(envVar)
-	if value == "" {
-		value = defaultValue
-	}
+func setImageRefFromEnvs(t *testing.T, dk *dynakube.DynaKube, imageRef *image.Ref, envVar, defaultRepo string) bool {
+	t.Helper()
+
+	envImage := os.Getenv(envVar)
```
mihaitanasedt: Can we do the env lookups in `GetLatestImageURI`? This way there will be only one point where the image is read from env or fetched. And in this method we can focus on the `CustomPullSecret` and parsing the image ref

mihaitanasedt (changes requested):

test/e2e/features/consts/consts.go
```diff
@@ -8,16 +8,6 @@ const (
...
-	LogMonitoringImageEnvVar = "E2E_LOGMON_IMAGE"
-	DefaultKSPMRepo          = "public.ecr.aws/dynatrace/dynatrace-k8s-node-config-collector"
-	KSPMImageEnvVar          = "E2E_KSPM_IMAGE"
-	DefaultOtelCollectorRepo = "public.ecr.aws/dynatrace/dynatrace-otel-collector"
-	OtelCollectorImageEnvVar = "E2E_OTELC_IMAGE"
-	DefaultDBExecutorRepo    = "public.ecr.aws/dynatrace/dynatrace-database-datasource-executor"
-	DBExecutorImageEnvVar    = "E2E_DB_EXECUTOR_IMAGE"
-	DefaultCodeModulesImage  = "ghcr.io/dynatrace/dynatrace-bootstrapper:snapshot"
-	CodeModulesImageEnvVar   = "E2E_CODEMODULES_IMAGE"
+	DefaultCodeModulesImage = "ghcr.io/dynatrace/dynatrace-bootstrapper:snapshot"
```
mihaitanasedt: Lets move these last 2 vars to the `node_image_pull.go`, the only place were they are used. To not have any leftovers in this file

test/e2e/helpers/components/dynakube/options.go
s-quatres: resolved with Andrii's comment below.

s-quatres: changed to setImageRefFromEnvOrLatest

test/e2e/helpers/registry/registry.go
s-quatres: done

mihaitanasedt (approved):

PR #6461, opened 2026-04-14
Adapt csv files to include missing files for 1.9.0

waodim:
## Description

follow up of https://github.com/Dynatrace/dynatrace-operator/pull/6457

This worked well but since `update-csv-bundles` failed some things were missing like:

```
schema.yaml
Chart.yaml
changelog
```

Also I renamed the bundle dockerfile since we are using concourse for this release we need to point to the correct file (FYI @avorima) 

https://bitbucket.lab.dynatrace.org/projects/CPN/repos/dynatrace-operator-release-pipeline/browse/tasks/61_release-community-operator/task.sh#84


config/helm/chart/default/Chart.yaml
```diff
@@ -18,8 +18,8 @@ description: The Dynatrace Operator Helm chart for Kubernetes and OpenShift
 icon: https://assets.dynatrace.com/global/resources/Signet_Logo_RGB_CP_512x512px.png
 home: https://www.dynatrace.com/
 type: application
-version: 0.0.0-snapshot
-appVersion: 0.0.0-snapshot
+version: 1.9.0
```
andriisoldatenko, line 21: why do we want to update this file?

config/helm/schema.yaml
```diff
@@ -15,84 +15,87 @@
 x-google-marketplace:
   schemaVersion: v2
   applicationApiVersion: v1beta1
-  publishedVersion: 0.6.0
+  publishedVersion: 1.9.0
```
andriisoldatenko, line 18: hm.... this file will be updated each release? why not have folder for it?

waodim: Yes it is updated each release

Sounds like a meaningful improvement, feel free to create a story for it

config/helm/chart/default/Chart.yaml
waodim: actually i am not sure, it might be that this is some left over, probably worth looking into it

avorima: the charts from the release branches always have the version set it was just not done this time due to an issue with the workflow (again)

andriisoldatenko (approved):

PR #6463, opened 2026-04-14
move apiToken paasToken dataIngestToken consts to pkg/controllers/dynakube/token

jokuniew:
## Description

As part of ICP-2063
old dtclient was imported in bunch of places just to import the token consts.

Two places have conflict with version  so variables and imports for client were renamed 
- `pkg/injection/namespace/bootstrapperconfig/endpoints.go`
- `pkg/controllers/dynakube/injection/reconciler_test.go`

## How can this be tested?
`make go/test`

### DIFF
--- a/pkg/consts/token.go
+++ b/pkg/consts/token.go
@@ -1,1 +1,0 @@
-package consts
--- a/pkg/consts/token.go
+++ b/pkg/consts/token.go
@@ -1,7 +1,1 @@
 package consts
-
-const (
-	PaasToken       = "paasToken"
-	APIToken        = "apiToken"
-	DataIngestToken = "dataIngestToken"
-)
--- a/pkg/consts/token.go
+++ b/pkg/consts/token.go
@@ -1,0 +1,7 @@
+package consts
+
+const (
+	PaasToken       = "paasToken"
+	APIToken        = "apiToken"
+	DataIngestToken = "dataIngestToken"
+)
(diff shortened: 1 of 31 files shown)


avorima:
These should be moved into the tokens package where the main logic around the tokens exist

pkg/consts/token.go
avorima, line 1: can be removed

jokuniew:
@avorima moved to `pkg/controllers/dynakube/token`

pkg/controllers/dynakube/token/tokens.go
```diff
@@ -11,6 +11,12 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/dynatraceapi"
 )
 
+const (
+	PaasToken       = "paasToken"
```
avorima: now that these are in the token package the "token" part of the name is redundant. instead, let's add the "Key" suffix to make it clear that this is not a token, i.e. the type that is exported by this package, but rather just used for lookup

also, nit: PaaS

pkg/consts/token.go
jokuniew: removed https://github.com/Dynatrace/dynatrace-operator/pull/6463/commits/bfc19809efc42a7fb16d006154c9bad34e5efaa0

pkg/controllers/dynakube/token/tokens.go
jokuniew: done here https://github.com/Dynatrace/dynatrace-operator/pull/6463/changes/c885b652000f6730998aa711ce37d0c9109526c9

avorima (approved):

PR #6464, opened 2026-04-14
Make WithPath additive

avorima:
## Description

Extend the WithPath request method to be additive instead of always overwriting the path. It also uses url.JoinPath which obsoletes the use of string formatting to build a path.
Since this interface change makes the mocks expect a slice, I decided to only keep this behavior change in the WithPath method to keep the change focused on the packages that actually make use of this feature.

I overlooked this initially, because all the paths I was working with were static. Now that we're finishing up the refactoring we can revisit the interfaces and improve on them.

Having `WithPath(string)` be additive would be enough, but I changed it to varargs for ease of use.

ICP-2063

## How can this be tested?

Run activegate, edgeconnect and bootstrapper e2e tests

### DIFF
--- a/pkg/clients/dynatrace/core/client.go
+++ b/pkg/clients/dynatrace/core/client.go
@@ -28,8 +28,8 @@ type APIClient interface {
 
 // APIRequest provides a fluent interface for building and executing HTTP requests
 type APIRequest interface {
-	// WithPath sets the path for the request
-	WithPath(path string) APIRequest
+	// WithPath sets the path for the request. Path parts will be joined, ignoring leading or trailing slashes.
+	WithPath(path ...string) APIRequest
 	// WithQueryParams adds multiple query parameters to the request, overwriting existing keys if they exist
 	WithQueryParams(params map[string]string) APIRequest
 	// WithRawQueryParams adds multiple query parameters to the request
@@ -128,9 +128,9 @@ func (c *Client) DELETE(ctx context.Context, path string) APIRequest {
 	return c.newRequest(ctx).withMethod(http.MethodDelete).WithPath(path)
 }
 
-// WithPath sets the path for the request
-func (r *Request) WithPath(path string) APIRequest {
-	r.path = path
+// WithPath sets the path for the request. Path parts will be joined, ignoring leading or trailing slashes
+func (r *Request) WithPath(path ...string) APIRequest {
+	r.path = (&url.URL{Path: r.path}).JoinPath(path...).Path
 
 	return r
 }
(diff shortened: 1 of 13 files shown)


pkg/clients/dynatrace/core/client.go
```diff
@@ -28,8 +28,8 @@ type APIClient interface {
 
 // APIRequest provides a fluent interface for building and executing HTTP requests
 type APIRequest interface {
-	// WithPath sets the path for the request
-	WithPath(path string) APIRequest
+	// WithPath sets the path for the request. Path parts will be joined, ignoring leading or trailing slashes.
+	WithPath(path ...string) APIRequest
```
andriisoldatenko, line 32:
```suggestion
	WithJoinPath(path ...string) APIRequest
```
I prefer to explicitly set name to what it does, to not confuse. It's known fact that `join` works well with single path.

avorima: usually "With" implies an addition. for the body setters it's not really feasible, but everything else i would expect this to be the case. i didn't challenge it earlier because we were only using static paths. Just making `WithPath("foo")` additive would fix my actual issue, but having the varargs is less verbose on the caller side

andriisoldatenko: okey.

pkg/clients/dynatrace/oneagent/version.go
```diff
@@ -27,10 +31,11 @@ type GetParams struct {
 // Get gets the agent package for the given OS, installer type, flavor, arch and version.
 func (c *Client) Get(ctx context.Context, args GetParams, writer io.Writer) error {
 	if len(args.OS) == 0 || len(args.InstallerType) == 0 {
-		return errors.New("os or installerType is empty")
+		return errEmptyOSOrInstallerType
```
waodim: Would there be some kind of benefit if we split this one up to have a dedicated error message for both cases?

andriisoldatenko (dismissed):

pkg/clients/dynatrace/oneagent/version.go
avorima: yes. at some point i think we should merge all deployment/v1 API clients and the AG client already has an emptyOS error which could be reused for this

waodim (dismissed):

StefanHauth (approved):
Ran `make 
test/e2e/activegate 
test/e2e/edgeconnect 
test/e2e/applicationmonitoring/bootstrapper-csi test/e2e/applicationmonitoring/bootstrapper-no-csi` succesfully.

PR #6472, opened 2026-04-15
feat: logging: add context logger helpers

svfoxat:
## Description

https://dt-rnd.atlassian.net/browse/ICP-2916

- adds three helpers to `pkg/logd/context.go` to thread the controller-runtime–seeded logger through `context.Context`:
  - `NewFromContext(ctx, name, keysAndValues...)` — call at package entry points; appends a name and stores the enriched logger back into ctx
  - `FromContext(ctx)` — call in downstream helpers within the same package; falls back to `logd.Get()` when no logger is in ctx
  - `IntoContext(ctx, log)` — explicitly seed or replace a logger in ctx
- updates `doc/coding-style-guide.md` with the new pattern: `NewFromContext` at package boundaries, `FromContext` inside, `var log` globals only in `cmd/`

### Stack
- **PR 1 (this):** https://github.com/Dynatrace/dynatrace-operator/pull/6472
- PR 2: https://github.com/Dynatrace/dynatrace-operator/pull/6473
- PR 3: https://github.com/Dynatrace/dynatrace-operator/pull/6474

## How can this be tested?
`make go/test`

### DIFF
--- a/doc/coding-style-guide.md
+++ b/doc/coding-style-guide.md
@@ -495,34 +495,15 @@ Use `"github.com/Dynatrace/dynatrace-operator/pkg/logd"` (alias `logd`) for all 
   ctx = logd.IntoContext(ctx, myLogger)
   ```
 
-- **Utility functions that emit logs should accept `log logd.Logger` as a parameter** rather than closing over a package-level `log`.
-  This makes the function independently testable, its output attributable to the right component, and avoids hidden coupling to the package global.
+- **Utility functions that emit logs should accept `ctx context.Context` as a parameter** rather than using a package-level `log`.
 
   ```go
   // ✓
-  func printKubernetesVersion(log logd.Logger, kubeConfig *rest.Config) {
-      log.Info("kubernetes version", "version", ...)
-  }
-
-  // ✗ hidden dependency on package-level log
-  func printKubernetesVersion(kubeConfig *rest.Config) {
+  func printKubernetesVersion(ctx context.Context, kubeConfig *rest.Config) {
+      log := logd.FromContext(ctx)
       log.Info("kubernetes version", "version", ...)
   }
   ```
-
-- Include **enough context fields** in log messages to be useful during troubleshooting.
-  Reviewers frequently ask for additional fields (e.g. Go version, resource name, namespace). When in doubt, add more rather than less.
-
-### Exceptions
-
-`cmd/` packages and functions that have no `context.Context` parameter (e.g. background singletons, startup helpers) may keep a package-level logger declared with `logd.Get().WithName(...)`. Document the exception with a comment:
-
-```go
-// log is kept as a package-level logger because GetMinorVersion is a background singleton
-// cache with no context.Context parameter. This is an intentional exception to the
-// context-logger pattern.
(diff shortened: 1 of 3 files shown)


doc/coding-style-guide.md
```diff
@@ -463,15 +463,39 @@ import (
...
+  }
+  ```
+
+- **Use `logd.IntoContext(ctx, log)` when explicitly seeding a logger** into a context (e.g. in tests or when constructing a sub-context manually):
+
+  ```go
+  ctx = logd.IntoContext(ctx, myLogger)
+  ```
+
+- **Utility functions that emit logs should accept `log logd.Logger` as a parameter** rather than closing over a package-level `log`.
```
avorima: They should accept a context

doc/coding-style-guide.md
```diff
@@ -489,6 +513,17 @@ import (
...
 
+### Exceptions
+
+`cmd/` packages and functions that have no `context.Context` parameter (e.g. background singletons, startup helpers) may keep a package-level logger declared with `logd.Get().WithName(...)`. Document the exception with a comment:
+
+```go
+// log is kept as a package-level logger because GetMinorVersion is a background singleton
+// cache with no context.Context parameter. This is an intentional exception to the
+// context-logger pattern.
+var log = logd.Get().WithName("k8sversion")
```
avorima: I had actually planned to plumb the context through to all of these "singleton" loggers. There really isn't any reason to use a package variable anymore

pkg/logd/context.go
```diff
@@ -0,0 +1,34 @@
...
+	return logr.NewContext(ctx, log.Logger), log
+}
+
+func FromContext(ctx context.Context) Logger {
+	if ctx == nil {
+		return Get()
+	}
+
+	logger, err := logr.FromContext(ctx)
+	if logger.GetSink() == nil || err != nil {
```
avorima, line 21: the discard logger has a nil sink, so this function would prevent the use of it. if you check [controller-runtime's implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.23.3/pkg/log/log.go#L91) they don't have this behavior

avorima:
<strike>Hold on, this doesn't do want we want yet. We're still using our dumb logger instead of the enriched logger from the controller-runtime</strike>

Nvm, the propagation works

doc/coding-style-guide.md
avorima: Looking at the few exceptions most of them are just because there is no package logger in the commands, which could easily be fixed. I guess a good rule of thumb whether a package should have a logger or not is whether it's an entrypoint, i.e. main

pkg/logd/context_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"testing"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
+	"github.com/go-logr/logr"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFromContext(t *testing.T) {
```
mihaitanasedt, line 11: We can also test the case when ctx is nil

avorima (approved):

PR #6479, opened 2026-04-15
Enforce spelling of OS

avorima:
## Description

Should always be OS (Operating System), not Os.


waodim (approved):
nice catch

andriisoldatenko (approved):
good catch!

PR #6483, opened 2026-04-16
Detect platform tokens

avorima:
## Description

Add utility functions to detect whether users provided a [platform token](https://docs.dynatrace.com/docs/manage/identity-access-management/access-tokens-and-oauth-clients/platform-tokens).

ICP-3723

## How can this be tested?

Run unit tests

### DIFF
--- a/pkg/controllers/dynakube/token/reader.go
+++ b/pkg/controllers/dynakube/token/reader.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/dttoken"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -32,7 +33,7 @@ func (reader Reader) HasPlatformToken(ctx context.Context) (bool, error) {
 
 	token := tokens.APIToken()
 	if token != nil {
-		return token.Platform, nil
+		return dttoken.IsPlatform(token.Value), nil
 	}
 
 	return false, nil
--- a/pkg/controllers/dynakube/token/reader.go
+++ b/pkg/controllers/dynakube/token/reader.go
@@ -22,6 +22,22 @@ func NewReader(apiReader client.Reader, dk *dynakube.DynaKube) Reader {
 	}
 }
 
+// HasPlatformToken inspects the token secret and checks if the apiToken is a platform token.
+// Returns an error if reading the secret fails.
+func (reader Reader) HasPlatformToken(ctx context.Context) (bool, error) {
+	tokens, err := reader.readTokens(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	token := tokens.APIToken()
+	if token != nil {
+		return token.Platform, nil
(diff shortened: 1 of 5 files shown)


pkg/controllers/dynakube/token/reader.go
```diff
@@ -22,6 +22,22 @@ func NewReader(apiReader client.Reader, dk *dynakube.DynaKube) Reader {
...
+// Returns an error if reading the secret fails.
+func (reader Reader) HasPlatformToken(ctx context.Context) (bool, error) {
+	tokens, err := reader.readTokens(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	token := tokens.APIToken()
+	if token != nil {
+		return token.Platform, nil
```
StefanHauth:
```suggestion
		return IsPlatformToken(token.Value), nil
```

pkg/util/dttoken/token.go
```diff
@@ -47,3 +52,7 @@ func generateRandom(size int) (string, error) {
 
 	return base32.StdEncoding.EncodeToString(b)[:size], nil
 }
+
+func IsPlatform(token string) bool {
```
StefanHauth, line 58: Is it an option to make this a method of Token? Or is the intended usage later easier with the string param?
```suggestion
func (t *Token) IsPlatform() bool {
```

pkg/util/dttoken/token.go
```diff
@@ -47,3 +52,7 @@ func generateRandom(size int) (string, error) {
...
+	return strings.HasPrefix(token, PlatformPrefix)
```
StefanHauth, line 59:
```suggestion
	return strings.HasPrefix(t.value, PlatformPrefix)
```

pkg/controllers/dynakube/token/token.go
```diff
@@ -8,13 +8,15 @@ import (
...
 	dtclient "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/dttoken"
 	"github.com/pkg/errors"
 )
 
 type Token struct {
 	Type     string
 	Value    string
 	Features []Feature
+	Platform bool
```
StefanHauth: Why introduce a bool flag if this information is anyways part of `value`? What's the benefit of separately storing this information?
Also as "Value" is exported, there is a possibility of wrong use so that Value is changed but the Platform flag is not, so the state is then inconsistent.

StefanHauth (changes requested):

pkg/controllers/dynakube/token/token.go
avorima: it's a shortcut so consumers of tokens don't have to call the function again

pkg/util/dttoken/token.go
avorima: yes, for example the dtclient cannot consume the token package due to an import cycle

pkg/controllers/dynakube/token/token.go
StefanHauth: Feels like over-optimization to me.

pkg/util/dttoken/token.go
StefanHauth: Ok, fair enough. Then it makes sense

StefanHauth (approved):

PR #6486, opened 2026-04-17
add k8s 1.35 kind cluster and bump old versions

andriisoldatenko:
## Description

Introduce k8s 1.35 for kind cluster and e2e tests

## How can this be tested?

run kind e2e tests locally.

```
make test/e2e/kind
```


avorima (approved):

hack/kind/setup.sh
```diff
@@ -16,10 +16,11 @@ redcross="❌  "
 # TODO: we can use renovate to update these image versions
```
mihaitanasedt, line 16: Maybe we should create a ticket for this ?

andriisoldatenko: We have ticket already

PR #6489, opened 2026-04-17
Skip node reconcile when platform token is provided

avorima:
## Description

Read the token secret and skip the reconciler if it contains a platform token.
Log a message every 15 minutes that the node reconcile is disabled.

ICP-3201

## How can this be tested?

Some APIs are still under development, which makes testing a bit tricky. Since we just want to verify that the node reconciler is skipped, it's a bit easier though. Essentially, we want to reduce the number of API calls that the operator makes to make it easier to spot our log message.

* Create a DynaKube with only CNFS/hostmon OneAgent and platform token secret
  * Set images to skip tenant registry lookup
* Patch `status.oneAgent.instances` manually to include the cluster nodes. Run this command to get the patch data: `kubectl get nodes -o json | jq '.items | map({(.metadata.name): {podName: "oneagent-pod"}}) | add'`
* Check the logs for the message to appear. You can trigger a node reconcile like this: `kubectl annotate node --all test=test`


pkg/controllers/dynakube/token/tokens_test.go
```diff
@@ -605,3 +607,10 @@ func TestCheckForDataIngestToken(t *testing.T) {
 		assert.False(t, CheckForDataIngestToken(tokens))
 	})
 }
+
+func TestDisableLookupForPlatformToken(t *testing.T) {
+	tokens := Tokens{APIKey: &Token{Value: dttoken.PlatformPrefix + "test", Features: []Feature{{}}}}
```
StefanHauth: To properly test the new path in `VerifyScopes` the feature slice shouldn't be empty.

pkg/controllers/nodes/nodes_controller.go
```diff
@@ -277,3 +288,16 @@ func (controller *Controller) pruneCache(ctx context.Context, nodeCache *cache.C
 
 	return nil
 }
+
+var lastSkipLogTimestamp time.Time
```
StefanHauth: Nit: Don't we have vars and consts usually at the top of the file?

pkg/controllers/nodes/nodes_controller.go
```diff
@@ -277,3 +288,16 @@ func (controller *Controller) pruneCache(ctx context.Context, nodeCache *cache.C
...
+
+const logSkipInterval = 15 * time.Minute
```
StefanHauth: This const has a "configurational" character, so I'd prefer to have it at least on top and not buried here.

Also ticket requires "log every hour"

pkg/controllers/nodes/nodes_controller.go
```diff
@@ -277,3 +288,16 @@ func (controller *Controller) pruneCache(ctx context.Context, nodeCache *cache.C
...
+
+// NOT THREAD-SAFE!!!
+func logEveryInterval(message string) {
```
StefanHauth: Such a "log throttler" could be useful also somewhere else, worth moving to the logd package?

pkg/controllers/nodes/nodes_controller_test.go
```diff
@@ -386,3 +389,36 @@ func createDefaultFakeClient() client.Client {
 			},
 		})
 }
+
+func Test_logEveryInterval(t *testing.T) {
+	lastSkipLogTimestamp = time.Time{}
```
StefanHauth: Skip initiatlization to test "lastSkipLogTimestamp.IsZero()" path.

pkg/controllers/nodes/nodes_controller_test.go
StefanHauth, line 331: Add a unit test for new functionality:

```
 t.Run("Skip reconcile when platform token is detected", func(t *testing.T) {                                                                                                                                   
      fakeClient := fake.NewClient(                                                                                                                                                                              
          &corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "node1"}},
          &dynakube.DynaKube{ObjectMeta: metav1.ObjectMeta{Name: "oneagent1", Namespace: testNamespace}},                                                                                                        
          &corev1.Secret{                                                                                                                                                                                        
              ObjectMeta: metav1.ObjectMeta{Name: "oneagent1", Namespace: testNamespace},                                                                                                                        
              Data: map[string][]byte{                                                                                                                                                                           
                  token.APIKey: []byte(dttoken.PlatformPrefix + ".sometoken"),                                                                                                                                   
              },                                                                                                                                                                                                 
          },
      )                                                                                                                                                                                                          
      ctrl := createDefaultReconciler(t, fakeClient, nil)
      result, err := ctrl.Reconcile(t.Context(), createReconcileRequest("node1"))                                                                                                                                
      require.NoError(t, err)                                                                                                                                                                                    
      assert.Equal(t, reconcile.Result{}, result)                                                                                                                                                                
  })
```

pkg/controllers/nodes/nodes_controller.go
```diff
@@ -277,3 +288,16 @@ func (controller *Controller) pruneCache(ctx context.Context, nodeCache *cache.C
 
 	return nil
 }
+
+var lastSkipLogTimestamp time.Time
+
+const logSkipInterval = 15 * time.Minute
+
+// NOT THREAD-SAFE!!!
```
StefanHauth: Why not make it thread safe with a mutex or atomic?

pkg/controllers/nodes/nodes_controller.go
```diff
@@ -69,14 +69,25 @@ func NewControllerFromClient(clt client.Client) *Controller {
...
 
+	if skip, err := token.NewReader(controller.apiReader, dk).HasPlatformToken(ctx); err != nil {
+		return reconcile.Result{}, err
+	} else if skip {
+		logEveryInterval("node controller disabled due to detected platform token in secret")
+
+		return reconcile.Result{}, nil
+	}
+
+	// Reset log throttling for the edge case that the user switches back to 2nd gen token and then to platform token again.
```
StefanHauth:
```suggestion
	// Reset log throttling so the next switch from 2nd gen to a platform token logs immediately.
```

0sewa0 (dismissed):

StefanHauth (changes requested):

pkg/controllers/dynakube/token/tokens_test.go
avorima: it's not empty

pkg/controllers/nodes/nodes_controller.go
avorima: yes, there is a linter/formatter that enforces this but i think this grouping makes sense for one off functions like this because it keeps the focus

pkg/controllers/nodes/nodes_controller.go
avorima: I changed it to 15 minutes since that's our default API request interval

pkg/controllers/nodes/nodes_controller.go
avorima: it might be, but it's not a big deal to move it when we actually need it

pkg/controllers/nodes/nodes_controller_test.go
avorima: that is the zero path though

pkg/controllers/nodes/nodes_controller.go
avorima: this is just for our future selves to not forget if we ever actually make the controllers concurrent. i'd rather have this be a focused effort than sprinkling in mutexes and atomics where they're not needed

pkg/controllers/dynakube/token/tokens_test.go
StefanHauth: Right, but IMHO thats a bit hidden in the thousend brackets, how about: 
```
	tokens := Tokens{APIKey: &Token{
			Value: dttoken.PlatformPrefix + "test", 
			Features: []Feature{
				{
					Name:           "foobar",
				},
			},
		},
	}
```

pkg/controllers/nodes/nodes_controller.go
StefanHauth: True, but... I'm not sure if someone who will need this functionality will look for it here. I'm a bit concerned, that this will result in re-implementation.
Even more so, as there already is a package (`logd`) where it perfectly fits.

avorima: One of my favorite go proverbs is "a little copying is better than a little dependency". I actually already implemented another log throttling mechanism in the k8sentity reconciler, but it didn't fit this use case so I wrote the 5 LOC to cover this case here.
Adding this to logd means sharing global state, which I already regret introducing for the version lookup cache, but in that case it actually had a big impact.

There are actually a few open questions in this topic that we should look into.  ICP-4176

avorima: I thought about it again and decided to remove the throttling altogether. The other implementation will be removed in #6507.

I added the AC because I was looking at another ticket, but actually nobody requested us to implement this and the added complexity isn't worth it.

waodim (dismissed):
`{"level":"info","ts":"2026-04-23T08:44:28.279Z","logger":"nodes","msg":"node controller disabled due to detected platform token in secret","node":"gke-alkr-dev-default-pool-7dc64aeb-cm5p"}`


But please make sure all conversations are resolved with @StefanHauth before merging

aorcholski (dismissed):
LGTM test PASS

StefanHauth (dismissed):

StefanHauth (approved):

PR #6482, opened 2026-04-16
Remove old v1 client and AsV2

jokuniew:
## Description

ICP-2063

One acceptance criteria of ICP-2063 was to introduce the `NewClientFromDynakube` to replace `BuilderV2`
This will be done **in separate PR**.

### Highlights 
- removed WithTimeout option as it was used only in tests

## How can this be tested?
`make go/test`, e2e to be run

PR #6491, opened 2026-04-20
move dynatraceapi package functionality to client/core

jokuniew:
## Description

ICP-3998
As part of #6482 

- moved `IsUnreachable` to `core` package
- replaced `ServerError` with `HTTPError` in `concatErrors`. This replacement is as close as possible to previous functionality, no refactoring.

## How can this be tested?
`make go/test` e2e in progress


avorima (approved):
tbh, this should actually just use errors.Join

PR #6494, opened 2026-04-20
move "dt.kubernetes.cluster.id" deprecated attribute from injection t…

jokuniew:
## Description

As part of #6102 and discussion https://github.com/Dynatrace/dynatrace-operator/pull/6102#discussion_r3085054345
Created a sub-pr to find space to discuss if move of single deprecated label from `injection` to `metadata` is enough (or should be done at all).

The move looks satisfactory from perspective of having all deprecated fields in single place.

Metadata handler uses `ClusterInfo` now, so the open question is:
In context of recent `oa injection? -> enable metadataenrichment always` feature, should other fields be moved as well? 

## How can this be tested?
`make go/test`
`make test/e2e/applicationmonitoring/metadataenrichment`


StefanHauth (approved):
LGTM, for running unit and E2E tests I trust our CI, and moving all to one place seams fairly reasonable to me.

PR #6504, opened 2026-04-22
Validate install-path at injection time

0sewa0:
## Description

[ICP-4033](https://dt-rnd.atlassian.net/browse/ICP-4033)

Basically the same as https://github.com/Dynatrace/dynatrace-bootstrapper/pull/168, but at injection time.

Not validating the installPath properly can cause some weird volumeMounts as we just pass it to their. I was surprised the k8s even lets you do that.

Obvious questions:
- Why both here and in the bootstrapper? -> So that no security researcher can complain that it is exploitable in the boostrapper alone.
- Why not reuse the logic from the bootstrapper? -> Sharing logic between the 2 repos is a pain, that I would like to avoid.

## How can this be tested?

1. Default OA injection works as usual
2. Play around with `oneagent.dynatrace.com/install-path: "..."`, and try to break it.

[ICP-4033]: https://dt-rnd.atlassian.net/browse/ICP-4033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
+++ b/pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
@@ -20,11 +20,11 @@ const (
 	EphemeralVolumeType = "ephemeral"
 )
 
-type InvalidInstallPathErr struct {
+type invalidInstallPathError struct {
 	InstallPath string
 }
 
-func (err InvalidInstallPathErr) Error() string {
+func (err invalidInstallPathError) Error() string {
 	return fmt.Sprintf("the installPath (%s) must be clean, absolute and without whitespace and separators like ,:", err.InstallPath)
 }
 
@@ -82,7 +82,7 @@ func validateInstallPath(installPath string) error {
 		strings.ContainsAny(installPath, "\x00,:") ||
 		filepath.Clean(installPath) != installPath {
 		return dtwebhook.MutatorError{
-			Err:      InvalidInstallPathErr{installPath},
+			Err:      invalidInstallPathError{installPath},
 			Annotate: setNotInjectedAnnotationFunc(InvalidInstallPathReason),
 		}
 	}
--- a/pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
+++ b/pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
@@ -2,6 +2,7 @@ package oneagent
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 	"unicode"
@@ -78,6 +79,7 @@ func (mut *Mutator) IsInjected(request *dtwebhook.BaseRequest) bool {
 
 func validateInstallPath(installPath string) error {
 	if !filepath.IsAbs(installPath) ||
+		installPath == string(os.PathSeparator) ||
 		strings.ContainsFunc(installPath, unicode.IsSpace) ||
 		strings.ContainsAny(installPath, "\x00,:") ||
 		filepath.Clean(installPath) != installPath {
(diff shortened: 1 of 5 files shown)


waodim (dismissed):
LGTM
```
Annotations:      oneagent.dynatrace.com/injected: false                                                                                                                                                                                                │
│                   oneagent.dynatrace.com/install-path: /some/bad:path                                                                                                                                                                                   │
│                   oneagent.dynatrace.com/reason: InvalidInstallPath 
```

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
```diff
@@ -15,6 +20,14 @@ const (
 	EphemeralVolumeType = "ephemeral"
 )
 
+type InvalidInstallPathErr struct {
```
avorima: nit: the convention is that variables have the prefix `err` and types have the suffix `Error`

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
```diff
@@ -63,9 +76,27 @@ func (mut *Mutator) IsInjected(request *dtwebhook.BaseRequest) bool {
 	return maputils.GetFieldBool(request.Pod.Annotations, AnnotationInjected, false)
 }
 
+func validateInstallPath(installPath string) error {
+	if !filepath.IsAbs(installPath) ||
+		strings.ContainsFunc(installPath, unicode.IsSpace) ||
+		strings.ContainsAny(installPath, "\x00,:") ||
+		filepath.Clean(installPath) != installPath {
+		return dtwebhook.MutatorError{
+			Err:      InvalidInstallPathErr{installPath},
```
jokuniew: we don't to any Errors.As over this error. Can it be unexported?

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
```diff
@@ -15,6 +20,14 @@ const (
 	EphemeralVolumeType = "ephemeral"
 )
 
+type InvalidInstallPathErr struct {
+	InstallPath string
+}
+
+func (err InvalidInstallPathErr) Error() string {
```
jokuniew:
```suggestion
func (err *InvalidInstallPathErr) Error() string {
```
[why new custom errors should use pointer receivers](https://www.reddit.com/r/golang/comments/16eoxsv/comment/jzx628d/)
(we don't have errors.As on this type, but let's keep it clear from the beginning)

jokuniew (commented):
~nits~ nitpicks

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
0sewa0: the problem here is that we do have such things for the `MutatorError`, so I would conform to that.

we can look at that separately to fix our custom error handling

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
```diff
@@ -63,9 +76,27 @@ func (mut *Mutator) IsInjected(request *dtwebhook.BaseRequest) bool {
 	return maputils.GetFieldBool(request.Pod.Annotations, AnnotationInjected, false)
 }
 
+func validateInstallPath(installPath string) error {
+	if !filepath.IsAbs(installPath) ||
+		strings.ContainsFunc(installPath, unicode.IsSpace) ||
+		strings.ContainsAny(installPath, "\x00,:") ||
+		filepath.Clean(installPath) != installPath {
```
Incredible-O, line 85: `installPath = "/"` would pass these checks 🤠

Incredible-O (approved):

avorima (approved):

PR #6512, opened 2026-04-23
New Client from Dynakube

jokuniew:
## Description

ICP-2063
Follow-up to #6482 

### Highlights
- created `NewClientFromDynakube` method to fully replace Builder pattern for dtClient
  - used just `string` values for token to avoid circular dependency
- replaced `dynatraceClientBuilder` with `dtClientFactory` in controllers so we can still inject dtClient in tests
- added `String()` to `token.go` since i hit nil ptr deref if tokens did not contain paasToken

## How can this be tested?

`make go/test`
`make test/e2e/standard`
~🚧  faced issue in `TestStandard_cloudnative_codemodules_image`~ nevermind, multi-tenant.yaml misconfig


cmd/troubleshoot/dynakube.go
```diff
@@ -158,10 +155,7 @@ func checkAPIURLForLatestAgentVersion(ctx context.Context, baseLog logd.Logger,
 
 	logInfof(log, "checking if can pull latest agent version")
 
-	dtClient, err := dynatraceclient.NewBuilder(apiReader).
-		SetDynakube(*dk).
-		SetTokens(dynatraceAPISecretTokens).
-		Build(ctx)
+	dtClient, err := dynatrace.NewClientFromDynakube(ctx, apiReader, *dk, dynatraceAPISecretTokens.APIToken().String(), dynatraceAPISecretTokens.PaasToken().String(), "")
```
andriisoldatenko: why userAgentSuffix is empty string?

jokuniew: it was not set before here. What's the expected value here? `troubleshoot`?

andriisoldatenko: yes i think so

jokuniew: set there https://github.com/Dynatrace/dynatrace-operator/pull/6512/commits/496b0c62c88d43cceb80ebbded2d6732e4cd4233

pkg/clients/dynatrace/dynakube.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ClientFactory func(ctx context.Context, apiReader client.Reader, dk dynakube.DynaKube, apiToken, paasToken, userAgentSuffix string) (*Client, error)
```
avorima: we should generally use pointers for Kubernetes objects like the DynaKube

jokuniew: 👍 
Changed signature of NewClientFromDynakube + ClientFactory

and changed also the `installAgent`, `getInstaller`,`getJobInstaller`,  `getTargetDir`, `setupFileSystem`, `createLatestVersionSymlink` in `install.go` as fly-by
https://github.com/Dynatrace/dynatrace-operator/pull/6512/commits/f9f75c8b59a62c9159e9c41ac7ca93dc1a409db4

avorima (dismissed):

avorima (approved):

PR #6513, opened 2026-04-23
Remove HTTP client from config

avorima:
## Description

Remove the HTTP client from the client Config struct along with the option to set it.
The option was only used together with a httptest.Server, but for this use case setting the base URL works just as well.
Also remove the Timeout field from the Config struct. The corresponding option was already removed previously.

ICP-2063

## How can this be tested?

Unit tests


waodim (approved):
LGTM

jokuniew (approved):

PR #6516, opened 2026-04-24
Make dtClient cache opt-in per response

0sewa0:
## Description

follow up https://dt-rnd.atlassian.net/browse/ICP-1115
possible fix for https://dt-rnd.atlassian.net/browse/ICP-4242

Having the dtClient cache enabled for basically all GET responses turned out to be dangerous. We have many responses that shouldn't be cached in certain conditions. Mainly when they are "empty".  It depends per response what is "empty".

The plan is to turn on the cache per response. (will be done in separate PRs as we remove the other throttling mechanism)
- If the `struct` that is passed to the `Execute` func implements the new `Cacheable` interface -> will be cached
  - `Cacheable` only demands one method, `IsEmpty() bool` 

An example, Settings related responses:
- `IsEmpty` if `totalCount == 0`
- We shouldn't cache such responses, because we tend to instantly create a Setting after an empty response, and check the just created Setting for some more info added by the tenant, or in the next reconcile. 

Only enabled it for a single response, just to show an example: [The token scope checking](https://github.com/Dynatrace/dynatrace-operator/commit/99b864cc17728017f1611b5dc0908c94c65893e3)

Moves a few things around to avoid import cycles
- all the cache related Header keys are in the `middleware` package
- moved the `middleware` package underneath the `core` package

## How can this be tested?

1. Have 1 missing required scope in your token 
   -  (you can't check it with a "positive" example, because the current throttling method would just not make the requests)
3. Deploy the Operator + DK
4. See that we will only make 1 actual token-scope request to the API, all other attempts will be cached
5. Change the token to a correct one
6. We make another actual requests to the API, because it is a "different" request, all later attempts will be cached


pkg/clients/dynatrace/core/middleware/roundtripper.go
```diff
@@ -24,14 +35,13 @@ func (rt roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 
 func NewCacheRoundTripper(next http.RoundTripper, ttl time.Duration) http.RoundTripper {
 	return roundTripperFunc(func(r *http.Request) (*http.Response, error) {
-		if r.Method != http.MethodGet ||
-			r.Header.Get("Accept") == "application/octet-stream" {
+		if r.Header.Get("Accept") == "application/octet-stream" {
```
avorima: is this still needed if we cache per request anyways?

pkg/clients/dynatrace/core/middleware/roundtripper.go
```diff
@@ -40,15 +50,17 @@ func NewCacheRoundTripper(next http.RoundTripper, ttl time.Duration) http.RoundT
...
+			cachedResponse.Header.Set(CacheHitHeader, "true")
+			cachedResponse.Header.Set(CacheKeyHeader, cacheKey)
 			cachedResponse.Request = r
 
 			return cachedResponse, nil
 		}
 
 		// send the actual request
 		resp, err := next.RoundTrip(r)
-		if err == nil && core.IsSuccessResponse(resp) {
```
avorima, line 51: are we caching errors just to circumvent an import cycle?

avorima (commented):
i think we should just move the middleware into the core package (later)

pkg/clients/dynatrace/core/middleware/roundtripper.go
0sewa0: basically 😅 

> i think we should just move the middleware into the core package (later)

☝️

avorima (approved):

PR #6517, opened 2026-04-24
Remove token check status based throttling

0sewa0:
## Description

follow up for https://dt-rnd.atlassian.net/browse/ICP-1115 (dtClient cache)

The first throttling mechanism to go.
Token scope requests are cached in the dtClient, so no need to track it separately in the DynaKube + controller code

> [!WARNING]
>  We noticed an issue with the cache implementation, that it is too broad, and we should have it opt in per response instead which should be fixed in https://github.com/Dynatrace/dynatrace-operator/pull/6516
> This PR is independent of that, as the token-scope response is the example used in that PR for the new approach. 

## How can this be tested?

See that the controller makes the token-scope check API calls, but the client uses the cache to provide the response.

Do a `grep` for `"cached": true`, the core-client's debug logs should show if a request was cached or not


test/e2e/features/logmonitoring/logmonitoring.go
```diff
@@ -170,7 +171,7 @@ func triggerDaemonSetReconcile(dk dynakube.DynaKube) features.Func {
 
 		require.NoError(t, resources.Get(ctx, dk.Name, dk.Namespace, &dk))
 		// Force reconciliation by simulating the passage of time
-		dk.Status.DynatraceAPI.LastTokenScopeRequest.Time = dk.Status.DynatraceAPI.LastTokenScopeRequest.Add(-2 * dk.APIRequestThreshold())
+		dk.Spec.DynatraceAPIRequestThreshold = ptr.To(uint16(1))
```
avorima: this has no effect on the cache entry TTL

StefanHauth:
Deployed CNSF DK, token lookup is cached:

```
{"level":"debug","ts":"2026-04-28T07:06:09.077Z","logger":"dtclient-core","msg":"API request","method":"POST","host":"exq67461.dev.dynatracelabs.com","path":"/api/v2/apiTokens/lookup","query":"","status_code":200,"duration":"286.303µs","cached":true,"request_body":"{\"token\":\"dt0c01.2F53GEMK2YACIHS3HLETSVRM.***\"}","response_body":"{\"id\":\"dt0c01.2F53GEMK2YACIHS3HLETSVRM\",\"name\":\"Dynatrace Operator\",\"enabled\":true,\"personalAccessToken\":false,\"owner\":\"stefan.hauth@dynatrace.com\",\"creationDate\":\"2022-08-23T08:37:28.007Z\",\"lastUsedDate\":\"2026-04-28T07:00:27.514Z\",\"lastUsedIpAddress\":\"34.57.18.5\",\"modifiedDate\":\"2023-07-04T19:04:21.887Z\",\"scopes\":[\"InstallerDownload\",\"DataExport\",\"entities.read\",\"activeGateTokenManagement.create\",\"settings.read\",\"settings.write\"]}"}
```

StefanHauth (dismissed):

test/e2e/features/logmonitoring/logmonitoring.go
```diff
@@ -170,7 +171,7 @@ func triggerDaemonSetReconcile(dk dynakube.DynaKube) features.Func {
 
 		require.NoError(t, resources.Get(ctx, dk.Name, dk.Namespace, &dk))
 		// Force reconciliation by simulating the passage of time
-		dk.Status.DynatraceAPI.LastTokenScopeRequest.Time = dk.Status.DynatraceAPI.LastTokenScopeRequest.Add(-2 * dk.APIRequestThreshold())
+		dk.Spec.DynatraceAPIRequestThreshold = ptr.To(uint16(0))
```
avorima, line 174: i think this can actually be removed. we don't want to test that the token scopes got refreshed

0sewa0: you do:
```go
	if secretConfig.APITokenNoSettings == "" {
		t.Skip("skipping test. no token with missing settings scopes provided")
	}
 	...
	builder.Assess("update token secret", tenant.CreateTenantSecret(secretConfig.APIToken, secretConfig.DataIngestToken, testDynakube.Name, testDynakube.Namespace))
	builder.Assess("trigger reconcile", triggerDaemonSetReconcile(testDynakube))
	builder.Assess("log agent restarted", k8sdaemonset.WaitForDaemonset(testDynakube.LogMonitoring().GetDaemonSetName(), testDynakube.Namespace))
```

that is the actual test, but you could put anything in that field. because the cache is smart enough to notice the change in tokens.

The whole point there is to cause a reconcile (so the Operator actually gets the new tokens from the secret)

avorima: ok we'll address this in ICP-4002

0sewa0:
(had to rebase because api type name changed 🙄 )

avorima (approved):
<strike>we really need to </strike>

don't remember what i was trying to say here. thanks github for not deleting this

PR #6519, opened 2026-04-27
display each deprecated feature flag as a separate warning message

svfoxat:
## Description

https://dt-rnd.atlassian.net/browse/ICP-2313

If multiple deprecated feature-flags are detected, each one is listed in one sentence instead of one sentence per ff

#### Before:
```
Warning: Feature flag feature.dynatrace.com/oneagent-ignore-proxy is deprecated.Feature flag feature.dynatrace.com/activegate-updates is deprecated.
```

#### After:
```
Warning: Using deprecated feature flags: feature.dynatrace.com/oneagent-ignore-proxy, feature.dynatrace.com/activegate-updates`

```

## How can this be tested?

`make go/test`

`make build && make deploy` and apply a DynaKube with deprecated FFs

### DIFF
--- a/pkg/api/validation/dynakube/featureflag.go
+++ b/pkg/api/validation/dynakube/featureflag.go
@@ -3,7 +3,6 @@ package validation
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/exp"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
@@ -13,21 +12,21 @@ const (
 	warningFeatureFlagDeprecated = `Feature flag %s is deprecated.`
 )
 
-var deprecatedFeatureFlags = []string{
+var deprecatedFeatureFlagKeys = []string{
 	exp.OAProxyIgnoredKey,   //nolint:staticcheck
 	exp.AGUpdatesKey,        //nolint:staticcheck
 	exp.AGDisableUpdatesKey, //nolint:staticcheck
 	exp.OAMaxUnavailableKey, //nolint:staticcheck
 }
 
-func deprecatedFeatureFlag(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
-	var results strings.Builder
+func deprecatedFeatureFlags(_ context.Context, _ *Validator, dk *dynakube.DynaKube) []string {
+	var results []string
 
-	for _, flag := range deprecatedFeatureFlags {
+	for _, flag := range deprecatedFeatureFlagKeys {
 		if dk.Annotations != nil && dk.Annotations[flag] != "" {
-			fmt.Fprintf(&results, warningFeatureFlagDeprecated, flag)
+			results = append(results, fmt.Sprintf(warningFeatureFlagDeprecated, flag))
 		}
 	}
 
-	return results.String()
+	return results
 }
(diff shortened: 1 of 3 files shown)


pkg/api/validation/dynakube/featureflag.go
```diff
@@ -13,21 +12,21 @@ const (
 	warningFeatureFlagDeprecated = `Feature flag %s is deprecated.`
 )
 
-var deprecatedFeatureFlags = []string{
+var deprecatedFeatureFlagKeys = []string{
```
mihaitanasedt: The flags are already the annotation keys, I think the naming was already correct

pkg/api/validation/dynakube/validation.go
```diff
@@ -189,3 +192,13 @@ func getDynakube(obj runtime.Object) (dk *dynakube.DynaKube, err error) {
 
 	return
 }
+
+func (v *Validator) runMultiValidators(ctx context.Context, validators []multiValidatorFunc, dk *dynakube.DynaKube) []string {
+	results := []string{} //nolint:prealloc // each multiValidator returns a variable-length []string, so the total size is unknown upfront
```
mihaitanasedt: nit, because we already use `//nolint:prealloc` we can also do it without allocation:
```go
var results []string //nolint:prealloc
```

mihaitanasedt (changes requested):

pkg/api/validation/dynakube/featureflag.go
```diff
@@ -13,21 +12,21 @@ const (
...
 	exp.OAProxyIgnoredKey,   //nolint:staticcheck
 	exp.AGUpdatesKey,        //nolint:staticcheck
 	exp.AGDisableUpdatesKey, //nolint:staticcheck
 	exp.OAMaxUnavailableKey, //nolint:staticcheck
 }
 
-func deprecatedFeatureFlag(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
-	var results strings.Builder
+func deprecatedFeatureFlags(_ context.Context, _ *Validator, dk *dynakube.DynaKube) []string {
```
avorima: Changing the message to `Using deprecated feature flags: foo, bar, baz` would work within our established framework

svfoxat: The warning message is now saying: 
```
Warning: Using deprecated feature flags: feature.dynatrace.com/oneagent-ignore-proxy, feature.dynatrace.com/activegate-updates`
```

Should we keep the FF prefix (`feature.dynatrace.com/`) or should we show it without?

pkg/api/validation/dynakube/featureflag.go
```diff
@@ -21,13 +21,17 @@ var deprecatedFeatureFlags = []string{
...
 	for _, flag := range deprecatedFeatureFlags {
 		if dk.Annotations != nil && dk.Annotations[flag] != "" {
-			fmt.Fprintf(&results, warningFeatureFlagDeprecated, flag)
+			results = append(results, flag)
 		}
 	}
 
-	return results.String()
+	if len(results) > 0 {
+		return fmt.Sprintf(warningFeatureFlagDeprecated, strings.Join(results, ", "))
```
mihaitanasedt: Would suggest to use string concat `+`, it is safer and faster. Interesting [article](https://hackernoon.com/fmtsprintf-looks-simple-but-will-burn-a-hole-in-your-pocket) about the performance.

pkg/api/validation/dynakube/featureflag.go
avorima: I think we should keep it. There's only a handful of entries in the list and it allows to use the value as-is

avorima (approved):

mihaitanasedt (approved):

PR #6521, opened 2026-04-27
Remove dynatrace.com/updated-via-operator annotation

avorima:
## Description

The operator sets the `dynatrace.com/updated-via-operator` on namespaces to prevent the webhook from doing duplicate work when it marks a namespace as monitored.
The webhook looks for this annotation and deletes it via admission patch response.
This round trip can be replaced by inspecting the user info from the admission request. No additional handling on the operator side needed and in the webhook it's also clearer why the request is skipped.

Also refactored UnmapFromDynaKube to not fail early in case of an error. This aligns with the rest of our injection cleanup.

ICP-4253

## How can this be tested?

* Deploy operator
* Create a sample app: `kubectl create deploy nginx --image nginx`
* Create DynaKube

```yaml
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
spec:
  oneAgent:
    applicationMonitoring:
      namespaceSelector:
        matchLabels:
          kubernetes.io/metadata.name: default
```
* Check that bootstrapper secret is created
* Delete DynaKube
* Check that bootstrapper secret is deleted
* Webhook logs should contain 2 entries of "ignoring change from operator"

### DIFF
--- a/pkg/injection/namespace/mapper/dynakubes.go
+++ b/pkg/injection/namespace/mapper/dynakubes.go
@@ -2,6 +2,7 @@ package mapper
 
 import (
 	"context"
+	goerrors "errors"
 	"slices"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
@@ -51,8 +52,10 @@ func (dm *DynakubeMapper) MapFromDynakube() error {
 		return err
 	}
 
-	if err := dm.updateNamespaces(modifiedNs); err != nil {
-		return err
+	for _, ns := range modifiedNs {
+		if err := dm.client.Update(dm.ctx, ns); err != nil {
+			return err
+		}
 	}
 
 	oaActive := dm.dk.OneAgent().IsAppInjectionNeeded()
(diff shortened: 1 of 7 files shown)


pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -80,31 +83,16 @@ func (dm *DynakubeMapper) MatchingNamespaces() ([]*corev1.Namespace, error) {
...
 
-		err = dm.secrets.DeleteForNamespace(dm.ctx, consts.OTLPExporterSecretName, ns.Name)
-		if err != nil {
-			return err
-		}
-
-		err = dm.secrets.DeleteForNamespace(dm.ctx, consts.OTLPExporterCertsSecretName, ns.Name)
-		if err != nil {
+		if err := goerrors.Join(
+			errors.WithMessagef(dm.client.Update(dm.ctx, &ns), "failed to remove label %s from namespace %s", dtwebhook.InjectionInstanceLabel, ns.Name),
```
waodim: As much as I like this approach, I want to understand this. What happens if this fails for example? Does it fail immediately or will the secret deletions still happen since we collect the errors in goerrors.Join? This would mean that this step fails (removing the label) but the secrets still get removed. I don't think this is a state we want to be in.

So if I am correct, I guess moving the update call outside of the collection would prevent this. WDYT?

avorima: yes, everything gets executed. i'll move the update out

waodim (dismissed):
tested it - logs appear, everything works as intended

pkg/webhook/mutation/namespace/webhook.go
```diff
@@ -39,25 +41,24 @@ type webhook struct {
 //     but from the dynakube's side (during dynakube reconcile) and we don't want to repeat ourselves. So we just remove the annotation.
```
StefanHauth: this comment is outdated

pkg/webhook/mutation/namespace/webhook.go
```diff
@@ -39,25 +41,24 @@ type webhook struct {
...
 func (wh *webhook) Handle(ctx context.Context, request admission.Request) admission.Response {
 	if wh.namespace == request.Namespace {
-		return admission.Patched("")
+		return admission.Allowed("")
 	}
 
-	log.Info("namespace request", "namespace", request.Name, "operation", request.Operation)
+	logger := log.WithValues("namespace", request.Name, "operator", request.Operation)
```
StefanHauth:
```suggestion
	logger := log.WithValues("namespace", request.Name, "operation", request.Operation)
```

StefanHauth (changes requested):
Some nitpicks

0sewa0 (approved):
less code == best code ❤️

waodim (approved):

andriisoldatenko (approved):

StefanHauth (approved):

PR #6525, opened 2026-04-29
Adapt to fixed eTag handling on server side.

StefanHauth:
## Description

Server-side eTag handling is now done according to RFC spec. We need to adapt the handling on our end.

[JIRA](https://dt-rnd.atlassian.net/browse/ICP-1849)
[JIRA](https://dt-rnd.atlassian.net/browse/ICP-4081)

## How can this be tested?

```
export DT_API_URL='https://<tenant>.dev.dynatracelabs.com/api'
export DT_API_TOKEN='dt0c01.....'
export DT_CLUSTER_ID="${1:-}"

go test -v -run TestGetProcessGroupingConfig_Integration ./pkg/clients/dynatrace/oneagent/
```

### DIFF
--- a/pkg/clients/dynatrace/oneagent/processgroupingconfig_test.go
+++ b/pkg/clients/dynatrace/oneagent/processgroupingconfig_test.go
@@ -14,11 +14,10 @@ import (
 )
 
 const (
-	testCBORData        = "\x81\x01" // minimal CBOR: array of one integer
-	testETag            = `"abc123"`
-	testResponseETag    = "def456"
-	testResponseETagRaw = `"` + testResponseETag + `"`
-	testClusterID       = "my-cluster"
+	testCBORData     = "\x81\x01" // minimal CBOR: array of one integer
+	testETag         = `"abc123"`
+	testResponseETag = `"def456"`
+	testClusterID    = "my-cluster"
 )
 
 // setupMockedProcessGroupingClient builds a client backed by a mock core.Client.
@@ -70,7 +69,7 @@ func setupMockedProcessGroupingClient(
 func TestGetProcessGroupingConfig(t *testing.T) {
 	t.Run("success_200_with_etag", func(t *testing.T) {
 		var buf bytes.Buffer
-		respHeaders := http.Header{"Etag": []string{testResponseETagRaw}}
+		respHeaders := http.Header{"Etag": []string{testResponseETag}}
 
 		client := setupMockedProcessGroupingClient(t,
 			map[string]string{},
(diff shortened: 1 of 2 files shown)


pkg/clients/dynatrace/oneagent/processgroupingconfig_test.go
```diff
@@ -14,11 +14,10 @@ import (
 )
 
 const (
-	testCBORData        = "\x81\x01" // minimal CBOR: array of one integer
-	testETag            = `"abc123"`
-	testResponseETag    = "def456"
-	testResponseETagRaw = `"` + testResponseETag + `"`
-	testClusterID       = "my-cluster"
+	testCBORData     = "\x81\x01" // minimal CBOR: array of one integer
+	testETag         = `"abc123"`
```
andriisoldatenko, line 18: just for the record: when i send manually etag to dummy string server returns bad request:
```
08:34:33.447719 [0-0] > If-None-Match: "test"
08:34:33.447719 [0-0] >
08:34:33.448272 [0-0] * Request completely sent off
08:34:33.699544 [0-0] { [5 bytes data]
08:34:33.699771 [0-0] * TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
08:34:33.699854 [0-0] { [57 bytes data]
08:34:33.700059 [0-0] * TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
08:34:33.700110 [0-0] { [57 bytes data]
08:34:33.700512 [0-0] < HTTP/1.1 400 Bad Request
```
for me it only works with numeric numbers using quotes "0"

May be it makes sense to addopt to real behaviour.

pkg/clients/dynatrace/oneagent/processgroupingconfig_test.go
```diff
@@ -70,7 +69,7 @@ func setupMockedProcessGroupingClient(
 func TestGetProcessGroupingConfig(t *testing.T) {
```
andriisoldatenko, line 69: can we add unit-test with client error?
Example:
```
curl -vv -X 'GET' \
  'https://*.dev.dynatracelabs.com/api/v1/deployment/installer/agent/processgroupingconfig?ETag=0' \
  -H 'accept: application/cbor' \
  -H 'Authorization: Api-Token *.*' -H 'If-None-Match: "test"' --output d.cbor
```

return bad request 400.

pkg/clients/dynatrace/oneagent/processgroupingconfig_test.go
StefanHauth: The [RFC](https://www.rfc-editor.org/rfc/rfc9110.html#name-if-none-match) explicitly specifies alphanumerics, so I don't see a reason why we should limit ourselves to numerics. I could change it to digits, but this test doesn't care about the string content anyway.

andriisoldatenko: I know, point is implementation of tenant is different. but it's okey

pkg/clients/dynatrace/oneagent/processgroupingconfig_test.go
StefanHauth: [1626256](https://github.com/Dynatrace/dynatrace-operator/pull/6525/commits/16262564bebe342d6ea6d524b6d0fc9fd63b3618)

andriisoldatenko (approved):

PR #6527, opened 2026-04-29
Stop using Conditions for optional scope handling

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/ICP-1044)

## Description

Removes `ApiTokenSettingsRead` and `ApiTokenSettingsWrite` conditions. Status fields are used instead of them. 
```
  apiToken:
    availableOptionalScopes:
      settingsRead: true
      settingsWrite: true  
  phase: Deploying
```
New `ApiTokenOptionalScopes` condition is used for informing the user about missing optional scopes. 
```
   conditions:
   - lastTransitionTime: "2026-04-30T10:53:18Z"
     message: settings.read,settings.write optional scope(s) not available, some features
       may not work
     reason: ScopeMissing
     status: "False"
     type: ApiTokenOptionalScopes 
```

## How can this be tested?

unittests

make test/e2e/logmonitoring/optionalscopes
make test/e2e/kspm/optionalscopes


pkg/clients/dynatrace/token/client.go
```diff
@@ -24,9 +21,9 @@ const (
 )
 
 var (
-	OptionalScopes = map[string]string{
-		ScopeSettingsRead:  ConditionTypeAPITokenSettingsRead,
-		ScopeSettingsWrite: ConditionTypeAPITokenSettingsWrite,
+	OptionalScopes = []string{
```
avorima, line 24: We should take the chance and move all of these constants into the controller token package

pkg/api/latest/dynakube/dynakube_status.go
```diff
@@ -63,6 +63,13 @@ type DynaKubeStatus struct { //nolint:revive
 	// +listType=map
 	// +listMapKey=type
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	Tenant OptionalScopes `json:"tenant,omitempty"`
+}
+
+type OptionalScopes struct {
+	APITokenSettingsReadAvailable  bool `json:"apiTokenSettingsReadAvailable,omitempty"`
```
avorima: These should be bool pointers to correctly show `false` when the scopes are missing

pkg/api/latest/dynakube/dynakube_status.go
```diff
@@ -63,6 +63,13 @@ type DynaKubeStatus struct { //nolint:revive
 	// +listType=map
 	// +listMapKey=type
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	Tenant OptionalScopes `json:"tenant,omitempty"`
```
avorima: This name is already reserved by #6537. The scopes don't have anything to do with what's on the tenant, so maybe something like `Token TokenStatus`. This would also allow to drop the `apiToken` prefix from the field names

pkg/util/tenant/optionalscopes/optionalscopes.go
```diff
@@ -0,0 +1,35 @@
+package optionalscopes
```
avorima: Prefer the singular `optionalscope`. It works well with `optionalscope.IsAvailable(scopes, myscope)`

pkg/util/tenant/optionalscopes/optionalscopes.go
```diff
@@ -0,0 +1,35 @@
...
+	case token.ScopeSettingsRead:
+		return tokenStatus.APITokenSettingsReadAvailable
+	case token.ScopeSettingsWrite:
+		return tokenStatus.APITokenSettingsWriteAvailable
+	}
+
+	return false
+}
+
+func Missing(tokenStatus *dynakube.OptionalScopes, scope string) {
```
avorima: SetMissing/SetAvailable are clearer in their intent

pkg/controllers/dynakube/controller.go
```diff
@@ -54,6 +56,8 @@ const (
 	defaultRequeueInterval = 15 * time.Minute
 
 	controllerName = "dynakube-controller"
+
+	conditionTypeAPITokenOptionalScopes = "ApiTokenOptionalScopes"
```
avorima, line 62: This can be part of the Token condition. Less is more

pkg/controllers/dynakube/controller.go
```diff
@@ -495,23 +499,32 @@ func (controller *Controller) verifyTokenScopes(ctx context.Context, dtClient to
...
-			k8sconditions.SetOptionalScopeMissing(&dkStatus.Conditions, conditionType, scope+" optional scope not available, some features may not work")
+			optionalscopes.Missing(dk.OptionalScopes(), scope)
+			unavailableScopes = append(unavailableScopes, scope)
 		}
 	}
+
+	if len(unavailableScopes) > 0 {
+		k8sconditions.SetOptionalScopeMissing(&dk.Status.Conditions, conditionTypeAPITokenOptionalScopes, strings.Join(unavailableScopes, ",")+" optional scope(s) not available, some features may not work")
+	} else {
+		_ = meta.RemoveStatusCondition(&dk.Status.Conditions, conditionTypeAPITokenOptionalScopes)
```
avorima, line 528: we should also clean up the old conditions. no need to keep constants around. we did something similar for previous condition rename cleanups

pkg/util/tenant/optionalscopes/optionalscopes.go
avorima: Actually, maybe just pass in the whole dynakube instead of the scopes: `optionalscope.IsAvailable(dk, myscope)`

pkg/api/latest/dynakube/dynakube_status.go
waodim: Some ideas:

- `APITokenScopeStatus`
- `TokenScopes`

pkg/clients/dynatrace/token/client.go
aorcholski: let's do it separately

pkg/controllers/dynakube/controller.go
aorcholski: done. Old conditions are removed in `updateOptionalScopesConditions()`

pkg/controllers/dynakube/controller.go
aorcholski: It would be a mix of 2 messages with  1 more OptionalScopes related.

avorima (approved):

PR #6528, opened 2026-04-29
Decouple update-csv-bundles workflow from release workflow

waodim:
## Description

[ICP-3777](https://dt-rnd.atlassian.net/browse/ICP-3777)

Previously the `update-csv-bundles` workflow was triggered by pushing a tag -> in other words triggering the whole release. The workflow was then dependent by all prior steps, which made it close to impossible to re run the step if something failed. And in reality the step has nothing to do with the release workflow.

With this it now listens to:

- the release being published
- workflow dispatch with tag as input parameter (for re runs should it fail)

## How can this be tested?

[ICP-3777]: https://dt-rnd.atlassian.net/browse/ICP-3777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/update-csv-bundles.yaml
+++ b/.github/workflows/update-csv-bundles.yaml
@@ -49,6 +49,7 @@ jobs:
           echo "dry-run=${DRY_RUN}" >> "$GITHUB_OUTPUT"
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          ref: ${{ steps.vars.outputs.tag }}
           persist-credentials: false
       - name: Fetch image digest
         id: digest
--- a/.github/workflows/update-csv-bundles.yaml
+++ b/.github/workflows/update-csv-bundles.yaml
@@ -50,6 +50,10 @@ jobs:
           DIGEST=$(docker buildx imagetools inspect \
             "public.ecr.aws/dynatrace/dynatrace-operator:${TAG}" \
             --format '{{.Manifest.Digest}}')
+          if [ -z "$DIGEST" ]; then
+            echo "Failed to fetch digest for ${TAG} — image may not exist on public.ecr.aws"
+            exit 1
+          fi
           echo "digest=${DIGEST}" >> "$GITHUB_OUTPUT"
       - name: Generate release notes
         env:
--- a/.github/workflows/update-csv-bundles.yaml
+++ b/.github/workflows/update-csv-bundles.yaml
@@ -13,6 +13,7 @@ on:
 jobs:
   update-csv-bundles:
     name: Update Operatorhub Bundles
+    environment: Release
     runs-on: ubuntu-24.04
     if: ${{ github.event_name == 'workflow_dispatch' || !github.event.release.prerelease }}
     permissions:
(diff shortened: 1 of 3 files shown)


.github/workflows/update-csv-bundles.yaml
```diff
@@ -0,0 +1,56 @@
+name: Update CSV Bundles
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag (e.g. v1.2.3)"
```
jokuniew, line 9: PR description says `workflow dispatch with digest as input parameter (for re runs should it fail)`
were here it is tag.

So if tag has changed (for example vendor problems etc.) we can re-tag and re-trigger this and it'll gather ingest on its own (as it is in code now)?

question is: should it work as it is now and fix description, or should we start with digest first?

waodim: sorry I mistyped there, it should be tag and not digest

waodim: I updated the description

.github/workflows/update-csv-bundles.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ steps.vars.outputs.tag }}
+          persist-credentials: false
+      - name: Fetch image digest
+        id: digest
+        env:
+          TAG: ${{ steps.vars.outputs.tag }}
+        run: |
+          DIGEST=$(docker buildx imagetools inspect \
```
jokuniew, line 59: the only improvement worth doing here i see is to check if DIGEST is empty, cause it silences the fail of command
if by some reason we pass non-existing tag, or image does not exist on public.ecr.aws - we wouldn't see the problem early enough
Image

jokuniew: Update: i'm not sure at this point if `run:` propagates the error `set -e`. If yes, the comment is invalid. Need to check it

waodim: good point https://github.com/Dynatrace/dynatrace-operator/pull/6528/commits/9969b37055e50b84ee3909389e930aeac4b9d5a5

waodim: did not see the updated comment, keep me updated, if it is not needed I will revert the change

jokuniew: yup, the `-e` flag is passed always
https://docs.github.com/en/actions/how-tos/create-and-publish-actions/set-exit-codes#jobsjob_idstepsshell

Image
change can be reverted. sorry for confusion!

waodim: no problem better be safe than sorry https://github.com/Dynatrace/dynatrace-operator/pull/6528/commits/778121357d738bd6f383704a16081f0fd7a076fa

jokuniew (dismissed):

.github/workflows/update-csv-bundles.yaml
```diff
@@ -0,0 +1,65 @@
...
+        description: "Release tag (e.g. v1.2.3)"
+        required: true
+        type: string
+
+jobs:
+  update-csv-bundles:
+    name: Update Operatorhub Bundles
+    environment: Release
+    runs-on: ubuntu-24.04
+    if: ${{ github.event_name == 'workflow_dispatch' || !github.event.release.prerelease }}
```
avorima: What about a dry-run mode for pre-releases? I think it might be helpful so see the bundle before it gets published

waodim: I like the idea
I added a dry-run mode to the action. When triggered by a prerelease event it activates automatically, and for workflow_dispatch there's an optional checkbox. Instead of creating a PR it uploads the generated bundles as a workflow artifact (csv-bundles-<version>) so they can be inspected before the real release.

WDYT? https://github.com/Dynatrace/dynatrace-operator/pull/6528/commits/ffd7748a08503c5ffc80c644bd4dbc4b2b2d3b64

avorima (dismissed):

andriisoldatenko (dismissed):

waodim:
Have a look at my fork where I tested the workflow dispatch: https://github.com/waodim/dynatrace-operator/actions/runs/25157308674

You can inspect the artifact with the bundles.

andriisoldatenko (dismissed):

waodim:
Reverted to https://github.com/Dynatrace/dynatrace-operator/pull/6528/commits/c08cd776d76610cc49061d72419f8da3e6b95dbc again since this will be needed when a new release is there. Currently this won't "work" because 1.9.0 does not have the changes included and so the checkout does not work but as soon as we have a new tag with those changes in the workflow included it will work. Discussed with @avorima

avorima (approved):

PR #6533, opened 2026-05-01
fix: return error from installAgentFromImage when extraction fails

knQzx:
## description

`installAgentFromImage` logs the extraction error but returns nil. so `InstallAgent` never sees the error and skips the `os.RemoveAll(targetDir)` cleanup

for single-layer codemodule images this mostly hides itself - the symlink step fails afterwards on the missing dir and the error still surfaces, just with a misleading log

multi-layer is where it bites. layer 1 renames into targetDir successfully then layer 2 fails. targetDir has partial files. error swallowed. symlink created over the partial dir. `InstallAgent` returns `(true, nil)`. next reconcile sees `isAlreadyPresent=true` and skips install forever. silent broken install no auto recovery

## root cause

regression from a803e7787 ("Make the image download function use the go-containerregistry library") - the refactor dropped the `return errors.WithStack(err)` and changed the log message to "via proxy"

## fix

put the return back

## how can this be tested?

- existing `pkg/injection/codemodule/installer/image/...` tests pass
- manually: trigger an extraction failure (unreachable registry, bad auth) - the error now propagates instead of falling through to the symlink step


avorima (dismissed):
Hi, thank you for your contribution.

This looks okay to merge. Omitting the error at this point somewhat obfuscated that multi-layer images are not supported. I've created a follow-up task on our end to improve reporting in that case.

knQzx:
Pushed a build fix - after main was merged in, imageCacheDir became a field on imagePullInfo rather than a local, so the failure-log line stopped compiling. It now references the CacheDir constant, which is the exact value passed into imagePullInfo, so the log stays accurate. CI should go green again. @avorima the approval was dismissed on the update - could you take another look when you get a chance?

andriisoldatenko (approved):

0sewa0 (approved):

PR #6534, opened 2026-05-01
fix: remove EdgeConnect finalizer only after tenant cleanup succeeds

knQzx:
## description

`reconcileEdgeConnectDeletion` clears the finalizer and persists it BEFORE running tenant cleanup (`buildEdgeConnectClient`, `getEdgeConnectByName`, `deleteConnectionSetting`, `DeleteEdgeConnect`). if any step fails the k8s object has already lost its finalizer and gets garbage collected. tenant resource is orphaned forever

## root cause

been there since `8fa3f094d` ("Add edgeconnect provisioner mode") - no test for the failure path, no comment, no rationale

## fix

- run tenant cleanup first, only `controllerutil.RemoveFinalizer` after success
- pull tenant cleanup into a `cleanupTenantEdgeConnect` helper
- switch from `Finalizers = nil` to `controllerutil.RemoveFinalizer` so third-party finalizers don't get wiped

## how can this be tested?

existing edgeconnect tests pass

two new regression tests in `controller_test.go`:
- `finalizer is kept when DeleteEdgeConnect fails`
- `finalizer is kept when buildEdgeConnectClient fails`

both fail on main (object gets GC'd, finalizer gone) and pass on this branch

## note for operators

if the OAuth secret or `caCertsRef` configmap is deleted before the EdgeConnect CR, `buildEdgeConnectClient` fails and deletion gets retried indefinitely. recovery: recreate the secret/configmap or `kubectl patch` the finalizer off

### DIFF
--- a/pkg/controllers/edgeconnect/controller.go
+++ b/pkg/controllers/edgeconnect/controller.go
@@ -151,13 +151,6 @@ func (controller *Controller) reconcileEdgeConnectDeletion(ctx context.Context, 
 		return err
 	}
 
-	ec.Finalizers = nil
-	if err := controller.client.Update(ctx, ec); err != nil {
-		log.Debug("updating the EdgeConnect object failed, couldn't remove the finalizers")
-
-		return errors.WithStack(err)
-	}
-
 	edgeConnectClient, err := controller.buildEdgeConnectClient(ctx, ec)
 	if err != nil {
 		log.Debug("building EdgeConnect client failed")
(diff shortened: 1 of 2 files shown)


avorima:
Hi, thank your for your work on this.

The early finalizer removal is actually a [conscious trade off](https://github.com/Dynatrace/dynatrace-operator/pull/2298#discussion_r1395583586) for the case that you mentioned in your "note for operators" section. We prioritize not blocking deletion for any reason, so that would actually be a regression, albeit for an edge-case.

I'll make a note for when we discuss cleanup in the context of the operator again, but I don't think we can merge this.

knQzx:
got it, thanks!

PR #3543, opened 2024-07-26
Adds v1alpha2 EdgeConnect CRD

aorcholski:
[K8S-10077](https://dt-rnd.atlassian.net/browse/K8S-10077)

## Description

Adds v1alpha2 ~v1beta1~ EdgeConnect CRD.

## How can this be tested?

1) Install operator from the main branch.

2) Install EdgeConnect of version v1alpha1 with `hostRestrictions`:
```
apiVersion: dynatrace.com/v1alpha1
spec:
  hostRestrictions: "a.au,b.au" 
```

3) Update operator.

4) Check hostRestrictions:
```
apiVersion: dynatrace.com/v1alpha2
spec:
  hostRestrictions:
  - a.au
  - b.au
```
5) Check operator logs - no errors expected

PR #6535, opened 2026-05-01
fix: validate EdgeConnect CRD CA bundle in webhook cert controller

knQzx:
## description

the cert controller updates the CA bundle on both the DynaKube CRD and the EdgeConnect CRD in `Reconcile`, but `getCRD` only fetches DynaKube and `isUpToDate` only validates that one. if the EC CRD's CA bundle is stale while DynaKube's is fine, `isUpToDate` returns true, the reconciler short-circuits with `RequeueAfter: 3h`, and EC's conversion webhook stays broken until secret rotation or operator restart

## root cause

PR #3543 (Aug 2024) added the EC CRD update calls but didn't extend the validation. existing `updateCRDConfiguration(ctx, k8scrd.EdgeConnectName, ...)` already requires EC CRD to exist so this PR doesn't introduce any new "EC must exist" assumption

## fix

- rename `getCRD` -> `getCRDs`, return both DynaKube and EdgeConnect CRDs
- `isUpToDate` validates the bundle on each returned CRD

## how can this be tested?

existing `pkg/controllers/certificates/...` tests pass

new regression test `edgeconnect_crd_ca_bundle_is_fixed_even_if_dynakube_crd_is_already_valid` seeds both CRDs, wipes only the EC CRD's CA bundle, asserts the reconcile fixes it. fails on main passes on this branch

### DIFF
--- a/pkg/controllers/certificates/webhook_cert_controller.go
+++ b/pkg/controllers/certificates/webhook_cert_controller.go
@@ -140,7 +140,7 @@ func (controller *WebhookCertificateController) Reconcile(ctx context.Context, r
 
 	mutatingWebhookConfiguration, validatingWebhookConfiguration := controller.getWebhooksConfigurations(ctx)
 
-	crd, err := controller.getCRD(ctx)
+	crds, err := controller.getCRDs(ctx)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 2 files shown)


avorima (dismissed):
Thank you for digging into this.

While your case is certainly possible, it requires a certain kind of setup which users usually can remediate themselves. So it didn't get reported yet.

knQzx:
rebased on main - govulncheck was failing on `golang.org/x/net@v0.52.0` (GO-2026-5026, GO-2026-4918), which main has long since moved past, the branch was just three months stale

one thing the rebase did not catch on its own: `prepareController` gained a `testing.TB` parameter on main, so the new test case was still calling the old signature and the package would not build. fixed that too

checked that the branch adds nothing of its own - govulncheck reports the exact same set of findings on this branch as on clean main, and `golangci-lint` (v2.12.2, the version pinned in CI) is clean on the touched package

PR #4903, opened 2025-05-21
Fix missing telemetryIngest converter from beta4 to latest

albertogdd:
## Description

Fix https://dt-rnd.atlassian.net/browse/DAQ-8775

`spec.telemetryIngest` converter from beta4 to latest was missing.

## How can this be tested?

- Unit tests :heavy_check_mark: 
- Deploy DK with `telemetryIngest` field on main and on this branch, check that on main the field is gone once deployed, and on this branch it is converted nicely.

```yaml
apiVersion: dynatrace.com/v1beta4
kind: DynaKube
metadata:
  name: dynatrace-otel
  namespace: dynatrace
spec:
  apiUrl: https://pap94091.dev.dynatracelabs.com/api
  telemetryIngest:
    protocols:
    - "otlp"
    - "zipkin"
    - "jaeger"
    - "statsd"
    tlsRefName: ""
    serviceName: dynatrace-otel-telemetry-ingest
  tokens: dynakube
```

PR #6536, opened 2026-05-01
fix: preserve TelemetryIngest spec in v1beta5 ConvertFrom

knQzx:
## description

`Spec.TelemetryIngest` is copied by `ConvertTo` (v1beta5 -> latest) via `toTelemetryIngestSpec`, but the matching `ConvertFrom` (latest -> v1beta5) has no `fromTelemetryIngestSpec` and silently drops it. the field exists on v1beta5, so anyone GETting a DynaKube as v1beta5 sees `spec.telemetryIngest = null`, and a following PUT clobbers the latest storage version (v1beta6) with null, dropping the field from etcd

## root cause

PR #4903 added the converter only in the convert_to direction. the reverse was overlooked and got inherited when v1beta5 came in

## fix

- add `fromTelemetryIngestSpec` to v1beta5 `convert_from.go`, mirroring `toTelemetryIngestSpec`, and wire it into the `ConvertFrom` dispatcher
- dropped the v1beta4 part from the original PR as discussed, since that version is no longer served

this was the only asymmetric field, the other `to*`/`from*` helpers are all symmetric (checked the full diff of helper invocations)

## how can this be tested?

- existing `pkg/api/v1beta5/dynakube/...` tests pass
- new v1beta5 round-trip unit test (`ConvertTo` then `ConvertFrom`, asserts every subfield is preserved) plus a `migrate telemetryIngest` test for the latest -> v1beta5 direction
- extended the webhook integration test (`pkg/api/validation/dynakube`) to also exercise the backward path: served versions are now read back in their own version (hub -> spoke / `ConvertFrom`), unserved versions stay excluded (v1beta4 only asserts a NoMatchError), and `telemetryIngest` was added to the testdata so the dropped field is actually covered
- all of the above fail on the old code and pass with the fix

### DIFF
--- a/pkg/api/v1beta5/dynakube/convert_from.go
+++ b/pkg/api/v1beta5/dynakube/convert_from.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta5/dynakube/kspm"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta5/dynakube/logmonitoring"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta5/dynakube/oneagent"
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1beta5/dynakube/telemetryingest"
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
 )
 
@@ -29,6 +30,7 @@ func (dst *DynaKube) ConvertFrom(srcRaw conversion.Hub) error {
 	dst.fromOneAgentSpec(src)
 	dst.fromActiveGateSpec(src)
 	dst.fromTemplatesSpec(src)
+	dst.fromTelemetryIngestSpec(src)
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


avorima (changes requested):
Generally agree with the changes, but there's some additional work to be done.

The v1beta4 convert_from code is effectively unreachable since it's no longer served and we have some pending tasks to strip most code from deprecated API versions in general to reduce the maintenance burden. So to keep it focused, please remove that part.

Given that we missed the v1beta5 I think we need to add an integration test to catch these things earlier next time. In the validation webhook tests we only exercise the path forward. Please adapt it to also do the path backwards. Of course excluding unserved versions.

ICP-5447

knQzx:
done - removed the v1beta4 part, so it's v1beta5 only now

extended the webhook integration test to also exercise the backward path (reading a served version back, i.e. hub -> spoke / ConvertFrom) for served versions only, and added telemetryIngest to the testdata so the dropped field is actually covered - fails on the old code, passes with the fix

avorima (approved):
LGTM

PR #6537, opened 2026-05-04
Store tenantPhase in DynaKube status

waodim:
## Description

With this we are now storing the tenant phase in the DK status by querying it via client. A new field is added to the DK status and it is populated with the response (essentially just an int).

[ICP-3199](https://dt-rnd.atlassian.net/browse/ICP-3199)

## How can this be tested?

- unit tests
- if you have a gen 3 tenant: use the endpoint, or check the operator logs

[ICP-3199]: https://dt-rnd.atlassian.net/browse/ICP-3199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/api/latest/dynakube/tenant/status.go
```diff
@@ -0,0 +1,7 @@
+package tenant
+
+// +kubebuilder:object:generate=true
+
+type Status struct {
```
avorima: We already have dynakube_status.go. Let's put the struct there to keep things simple. I don't predict this package will grow much beyond adding a few other fields to this struct

pkg/clients/dynatrace/client.go
```diff
@@ -96,6 +98,7 @@ func NewClient(options ...Option) (*Client, error) {
 		OneAgent:   oneagent.NewClient(apiClient, config.HostGroup, config.NetworkZone),
 		Version:    version.NewClient(apiClient),
 		Token:      token.NewClient(apiClient),
+		Platform:   platform.NewClient(apiClient),
```
avorima: This doesn't quite work. The API URL is remapped to 2nd gen, but for this API we always need 3rd gen

pkg/api/latest/dynakube/tenant/status.go
waodim: This was actually my first approach but then I thought maybe best practice is to have a file for every "prop". Thanks, I agree with your opinion.

https://github.com/Dynatrace/dynatrace-operator/pull/6537/commits/1e8d518c291b2597450b139b8c04b518385d0ed2

pkg/clients/dynatrace/client.go
waodim: Good point, I tried to rework the mapping and always store api and platform clients. And just pass the platform client. https://github.com/Dynatrace/dynatrace-operator/pull/6537/commits/1e8d518c291b2597450b139b8c04b518385d0ed2

pkg/clients/dynatrace/client.go
avorima: platform client should not use paasToken

pkg/clients/dynatrace/client.go
```diff
@@ -314,6 +318,25 @@ func proxyWrapper(proxyConfig httpproxy.Config) func(req *http.Request) (*url.UR
...
+		apiCopy := *base
+		mapThirdGenAPIURL(&apiCopy)
+		apiURL = &apiCopy
+	} else {
+		if path.Base(base.Path) != "api" {
+			apiURL = base.JoinPath("api")
+		} else {
+			apiURL = base
+		}
+		platformURL = apiURL
```
avorima: the platform URL should never have the `/api` suffix

pkg/clients/dynatrace/client.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/6537/commits/d3e9ba5daf7589bee9ed7f1d32d9a6118c652994

pkg/clients/dynatrace/client.go
waodim: Thanks updated the logic https://github.com/Dynatrace/dynatrace-operator/pull/6537/commits/d3e9ba5daf7589bee9ed7f1d32d9a6118c652994

pkg/clients/dynatrace/client.go
```diff
@@ -328,11 +327,13 @@ func selectClientURLs(base *url.URL) (apiURL, platformURL *url.URL) {
 	} else {
 		if path.Base(base.Path) != "api" {
 			apiURL = base.JoinPath("api")
+			platformURL = base
 		} else {
 			apiURL = base
+			withoutAPI := *base
+			withoutAPI.Path = strings.TrimSuffix(base.Path, "/api")
```
avorima: This needs to handle the edge-case of `/api/`. An alternative to nested `strings.TrimSuffix` is `path.Dir(path.Clean(path))`

pkg/clients/dynatrace/client_test.go
```diff
@@ -437,16 +437,16 @@ func TestSelectClientURLs(t *testing.T) {
...
+			expectedPlatform: "https://tenant.live.dynatrace.com",
 		},
 		{
-			name:             "2nd-gen already with /api: unchanged",
+			name:             "2nd-gen already with /api: api unchanged, platform strips /api",
 			input:            "https://tenant.live.dynatrace.com/api",
 			expectedAPI:      "https://tenant.live.dynatrace.com/api",
-			expectedPlatform: "https://tenant.live.dynatrace.com/api",
+			expectedPlatform: "https://tenant.live.dynatrace.com",
 		},
```
avorima, line 456: The code already covers this, but a unit test would be good to document the intent: For ActiveGate URLs we only need to strip the `/api` suffix

```
{
	name:             "activegate URL",
	input:            "https://activegate.some.domain:9999/e/tenant123/api",
	expectedAPI:      "https://activegate.some.domain:9999/e/tenant123/api",
	expectedPlatform: "https://activegate.some.domain:9999/e/tenant123",
},                                                                                                                                                                          
```

pkg/clients/dynatrace/client.go
```diff
@@ -311,6 +317,28 @@ func proxyWrapper(proxyConfig httpproxy.Config) func(req *http.Request) (*url.UR
...
+		platformURL = &original
+		apiCopy := *base
+		mapThirdGenAPIURL(&apiCopy)
+		apiURL = &apiCopy
+	} else {
+		if path.Base(base.Path) != "api" {
+			apiURL = base.JoinPath("api")
+			platformURL = base
+		} else {
+			apiURL = base
```
mihaitanasedt: We should use a copy in this case, otherwise later changes on apiUrl will overwrite the base. Same for `platformURL`

pkg/clients/dynatrace/client.go
```diff
@@ -311,6 +317,28 @@ func proxyWrapper(proxyConfig httpproxy.Config) func(req *http.Request) (*url.UR
 	return func(req *http.Request) (*url.URL, error) { return proxyFunc(req.URL) }
 }
 
+func selectClientURLs(base *url.URL) (apiURL, platformURL *url.URL) {
```
mihaitanasedt, line 320: The code nesting can be simplified if we return early on gen3. (some comments would also help to improve readability)
example

```go
func selectClientURLs(base *url.URL) (apiURL, platformURL *url.URL) {
    platform := *base
    // If gen3
    if strings.Contains(base.Hostname(), ".apps.") {
        api := *base
        mapThirdGenAPIURL(&api)
        return &api, &platform
    }

    // If gen2 with no /api path
    if path.Base(base.Path) != "api" {
        return base.JoinPath("api"), &platform
    }

    api := *base
    // If it has /api path strip it for platform
    platform.Path = strings.TrimSuffix(base.Path, "/api")
    return &api, &platform
}
```

pkg/clients/dynatrace/client.go
waodim: Thanks for the tip

https://github.com/Dynatrace/dynatrace-operator/pull/6537/commits/f7aeeccb33172900b5a5344fa1c8de9ef862ba0c

pkg/clients/dynatrace/client_test.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/6537/commits/f7aeeccb33172900b5a5344fa1c8de9ef862ba0c

mihaitanasedt (changes requested):

pkg/clients/dynatrace/client.go
waodim: good point, thanks: https://github.com/Dynatrace/dynatrace-operator/pull/6537/commits/b6dead09ed245bb137746a430bd7a80fd341d906

pkg/clients/dynatrace/client.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/6537/commits/b6dead09ed245bb137746a430bd7a80fd341d906

mihaitanasedt (dismissed):

pkg/clients/dynatrace/client.go
```diff
@@ -311,6 +317,28 @@ func proxyWrapper(proxyConfig httpproxy.Config) func(req *http.Request) (*url.UR
...
+	if strings.Contains(base.Hostname(), ".apps.") {
+		api := *base
+		mapThirdGenAPIURL(&api)
+
+		return &api, &platform
+	}
+
+	// If gen2 with no /api path: append /api for the API client
+	if path.Base(base.Path) != "api" {
+		return base.JoinPath("api"), &platform
```
avorima, line 332: sorry, i mixed up some requirements. to map an AG 2nd gen to 3rd gen, we just have to strip the /api, but for any other URL we need to add the "apps" part in the domain. I'm thinking that maybe we're being too smart about this.

waodim:
The approach on how to do the mapping of URLs is being discussed. Up until a decision is reached do not merge this PR.

waodim:
This PR will be closed since the requirements changed. We now no longer will determine the tenant phase and therefor don't need to add it to the DK status. cc @avorima

avorima:
we will keep using 2nd gen endpoints for the foreseeable time, so the mapping from 3rd to 2nd gen will remain. the platform client should be the only one that uses the `.apps.` URL so the mapping back from 2nd to 3rd can be kept in the constructor there

0sewa0:
closing, because:

> This PR will be closed since the requirements changed. We now no longer will determine the tenant phase and therefor don't need to add it to the DK status.

PR #6538, opened 2026-05-04
Improve openapi generation CI and config

mihaitanasedt:
## Description

This PR improves the CI and configuration for generating OpenAPI clients. The new setup supports generating clients from multiple sources, with a more flexible global and per-schema configuration that allows targeting specific APIs.

The sync-openapi-schema.yaml workflow was simplified by removing the remote repo cloning step. `openapi-generator-cli` takes an input the remote source and credentials per-schema and generates the clients. The models for the clients are autogenerated with a new script. Additionally the PR description in the workflow was adapted to the new changes.

A `README` was added to guide contributors through adding new clients.

[Jira ICP-1905](https://dt-rnd.atlassian.net/browse/ICP-1905)

## How can this be tested?
Trigger `sync-openapi-schema.yaml‎` workflow from this branch with a test schema in `sync-config.yaml`

Test [workflow run](https://github.com/Dynatrace/dynatrace-operator/actions/runs/25497550499).
Test auto-generated PR https://github.com/Dynatrace/dynatrace-operator/pull/6558  using the schemas
```yaml
schemas:
 - name: test
   specUrlEnvVar: "TENANT_API_V2_SPEC_URL"
   authEnvVar: "TENANT_API_V2_TOKEN"
   generate:
     packageName: test
     globalProperties:
       apis:
        - "ActiveGates"
 ```
  
 
(The test coverage step is failing on the auto-generated PR because the `.testcoverage` exclusion will be merged with this PR)

### CODE REVIEW

File: hack/make/oapi.mk [hunk 1]
```diff
@@ -2,45 +2,66 @@ OAPI_CONFIG_DIR := api/oapi
...
-	| sed -E 's%https://github.com/%%' \
-	| sed -E 's%git@github.com:%%' \
-	| sed -E 's%\.git$$%%')
-OAPI_GIT_USER_ID ?= $(firstword $(subst /, ,$(OAPI_GIT_REMOTE)))
-OAPI_GIT_REPO_ID ?= $(lastword $(subst /, ,$(OAPI_GIT_REMOTE)))
 
 ## Generate Go SDKs from OpenAPI specs
-oapi/generate: prerequisites/openapi-generator-cli
-	@yq -o=json '.schemas | map(select(.generate))' "$(OAPI_SYNC_CONFIG)" \
+oapi/generate: prerequisites/openapi-generator-cli prerequisites/yq
```
gkrenn: Please move the make command to a separate bash script or python maybe if it makes it more readable / easier to understand.

Keeping all the pipes and double dollar signs makes it hard to support in the future.

hack/make/bin/oapi-models-for-api.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+# Usage: oapi-models-for-api.sh <spec-file> <api>
+# Prints all model names (direct + transitive) used by operations with the given API tag
+set -euo pipefail
+
+SPEC="${1:?Usage: $0 <spec-file> <api>}"
+API="${2:?Usage: $0 <spec-file> <api>}"
+
+# BFS to find all models of an api
+jq -r --arg tag "$API" '
```
gkrenn: Would advise you to not implement it in jq language which we don't have in the repo.
Python would be easier to understand as we already have it in our repo and do not introduce another language.

hack/make/oapi.mk
gkrenn: Maybe with python we could even skip introducing another dependency? (yq)

gkrenn (changes requested):

hack/make/bin/oapi-models-for-api.sh
mihaitanasedt: Yes, this was also an idea, will try to implement it in python 👍

hack/make/oapi.mk
mihaitanasedt: I reimplemented it in python, could you have a second look?

hack/scripts/oapi-generate.py
```diff
@@ -0,0 +1,166 @@
...
+
+    default_version = gen_cfg.get("generatorVersion", "").lstrip("v")
+    default_output_dir = gen_cfg.get("outputDir", "")
+    default_additional_props = gen_cfg.get("additionalProperties", "")
+    default_global_props = gen_cfg.get("globalProperties", "")
+
+    with open(SYNC_CONFIG) as f:
+        sync_cfg = yaml.safe_load(f)
+
+    for schema in sync_cfg.get("schemas") or []:
```
gkrenn: Would move the main logic into own generate scheme function

hack/scripts/oapi-generate.py
```diff
@@ -0,0 +1,166 @@
...
+        name = schema["name"]
+        pkg = (schema.get("generate") or {}).get("packageName") or name
```
gkrenn: Would `gen = schema.get("generate") or {}` pull into own variable and use it there

hack/scripts/oapi-generate.py
```diff
@@ -0,0 +1,166 @@
...
+        version = ((schema.get("generate") or {}).get("generatorVersion") or default_version).lstrip("v")
+        schema_additional_props = (schema.get("generate") or {}).get("additionalProperties") or ""
+        additional_props = ",".join(p for p in [default_additional_props, schema_additional_props] if p)
+        output_dir = f"{default_output_dir}/{pkg}"
+
+        spec_url_var = schema.get("specUrlEnvVar", "")
+        spec_url = os.environ.get(spec_url_var, "") if spec_url_var else ""
```
gkrenn: Can be simplified: `spec_url = os.environ.get(schema.get("specUrlEnvVar", ""))`

gkrenn: Same for the other two-line env-var resolves

hack/scripts/oapi-generate.py
```diff
@@ -0,0 +1,166 @@
...
+            shutil.rmtree(output_dir, ignore_errors=True)
+            os.makedirs(output_dir, exist_ok=True)
+            shutil.copy(IGNORE_FILE, f"{output_dir}/.openapi-generator-ignore")
+
+            cmd = [
+                GENERATOR_CLI, "generate",
+                "-i", tmp_spec, "-g", "go", "-o", output_dir,
+                "--skip-validate-spec", "--minimal-update",
+            ]
+            if pkg:
```
gkrenn: the default is schema.name, can this ever be false?

hack/scripts/oapi-generate.py
```diff
@@ -0,0 +1,166 @@
...
+                cmd += ["--package-name", pkg]
+            if additional_props:
+                cmd += [f"--additional-properties={additional_props}"]
+            if global_props:
+                cmd += [f"--global-property={global_props}"]
+
+            env = {**os.environ, "OPENAPI_GENERATOR_VERSION": version}
+            if subprocess.run(cmd, env=env).returncode != 0:
+                print(f"ERROR: generation failed for {name}", file=sys.stderr)
```
gkrenn: Just write the string in `sys.exit("ERROR: ...")`

hack/scripts/oapi-generate.py
```diff
@@ -0,0 +1,166 @@
...
+            print(f"WARNING: no specUrlEnvVar set for {name}, skipping.")
+            continue
+
+        auth_var = schema.get("authEnvVar", "")
+        auth_token = os.environ.get(auth_var, "") if auth_var else ""
+
+        print(f"Downloading spec for {name}...")
+        try:
+            spec_data = download_spec(spec_url, auth_token)
+        except RuntimeError as e:
```
gkrenn: If you catch the url error here you don't have to wrap it in a runtime error:

```
    except urllib.error.URLError as e:
        sys.exit(f"ERROR: failed to download spec from {spec_url}: {e}")
```

hack/scripts/oapi-generate.py
```diff
@@ -0,0 +1,166 @@
...
+
+def main():
+    with open(GENERATOR_CONFIG) as f:
+        gen_cfg = yaml.safe_load(f)
+
+    default_version = gen_cfg.get("generatorVersion", "").lstrip("v")
+    default_output_dir = gen_cfg.get("outputDir", "")
+    default_additional_props = gen_cfg.get("additionalProperties", "")
+    default_global_props = gen_cfg.get("globalProperties", "")
+
```
gkrenn:
```suggestion
    defaults = {
        "version": gen_cfg.get("generatorVersion", ""),
        "output_dir": gen_cfg.get("outputDir", ""),
        "additional": gen_cfg.get("additionalProperties", ""),
        "global": gen_cfg.get("globalProperties", ""),
    }
```

hack/scripts/oapi-generate.py
```diff
@@ -0,0 +1,166 @@
+import sys
```
gkrenn, line 1: I know we didn't test python files before, but would be a nice start here. Especially as this logic is not trivial to understand.

gkrenn (changes requested):
Sorry for so many ask, mainly minor ones

hack/scripts/oapi-generate.py
andriisoldatenko: yes we can add doctests, then we don't need to have test files or so.

hack/scripts/oapi-generate.py
andriisoldatenko: specUrlEnvVar -> `OPENAPI_SPEC_URL`

hack/scripts/oapi-generate.py
mihaitanasedt: Done https://github.com/Dynatrace/dynatrace-operator/pull/6538/changes/17d5803dac6ec63bd61d14b9f361aaa71b808987

hack/scripts/oapi-generate.py
mihaitanasedt: Done https://github.com/Dynatrace/dynatrace-operator/pull/6538/changes/17d5803dac6ec63bd61d14b9f361aaa71b808987

hack/scripts/oapi-generate.py
mihaitanasedt: You are right, fixed: https://github.com/Dynatrace/dynatrace-operator/pull/6538/changes/17d5803dac6ec63bd61d14b9f361aaa71b808987

hack/scripts/oapi-generate.py
mihaitanasedt: Done https://github.com/Dynatrace/dynatrace-operator/pull/6538/changes/17d5803dac6ec63bd61d14b9f361aaa71b808987

hack/scripts/oapi-generate.py
mihaitanasedt: Done https://github.com/Dynatrace/dynatrace-operator/pull/6538/changes/17d5803dac6ec63bd61d14b9f361aaa71b808987

hack/scripts/oapi-generate.py
mihaitanasedt: Done https://github.com/Dynatrace/dynatrace-operator/pull/6538/changes/17d5803dac6ec63bd61d14b9f361aaa71b808987

hack/scripts/oapi-generate.py
mihaitanasedt: Added doctests https://github.com/Dynatrace/dynatrace-operator/pull/6538/changes/17d5803dac6ec63bd61d14b9f361aaa71b808987. 
Looks good?

hack/scripts/oapi-generate.py
mihaitanasedt: @andriisoldatenko  `specUrlEnvVar ` is the key name in the yaml, it holds as value the name of the env variable where the spec is stored so its not an env. Or you mean something else?

mihaitanasedt: Done https://github.com/Dynatrace/dynatrace-operator/pull/6538/changes/17d5803dac6ec63bd61d14b9f361aaa71b808987

.github/workflows/sync-openapi-schema.yaml
```diff
@@ -7,196 +7,121 @@ permissions: {}
...
         with:
           persist-credentials: 'false'
-      - name: Parse config into matrix
+      - name: Parse config
         id: config
         run: |
           set -euo pipefail
           schemas=$(yq -o=json -I=0 '.schemas // []' api/oapi/sync-config.yaml)
           count=$(echo "${schemas}" | jq 'length')
+          output_dir=$(yq -r '.outputDir' api/oapi/generator-config.yaml)
```
gkrenn, line 29: You remove yq from the prerequisites, does this still work?

hack/scripts/oapi-generate.py
gkrenn: Will close as done, it's one line now and `specUrlEnvVar` is just the field name, that's alright. Feel free to reopen @andriisoldatenko

.github/workflows/sync-openapi-schema.yaml
mihaitanasedt: Yes it does, because it runs on another runner, `ubuntu-latest` has it by default. Our own `operator-e2e` doesn't so I moved the logic to the `ubuntu-latest` to remove the dependency

gkrenn: Awesome 👍

.github/workflows/sync-openapi-schema.yaml
```diff
@@ -7,196 +7,121 @@ permissions: {}
...
-            EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
-            echo "pr_body<<${EOF}" >> "$GITHUB_OUTPUT"
-            echo "${pr_body}" >> "$GITHUB_OUTPUT"
-            echo "${EOF}" >> "$GITHUB_OUTPUT"
-          fi
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          echo "pr_body<<${EOF}" >> "$GITHUB_OUTPUT"
+          echo "${pr_body}" >> "$GITHUB_OUTPUT"
+          echo "${EOF}" >> "$GITHUB_OUTPUT"
         env:
```
gkrenn: This is quite overengineered. Consider using a static string for reproducability.

```suggestion
          echo "pr_body<<EOF" >> "$GITHUB_OUTPUT"
          echo "${pr_body}" >> "$GITHUB_OUTPUT"
          echo "EOF" >> "$GITHUB_OUTPUT"
        env:
```

.github/workflows/sync-openapi-schema.yaml
```diff
@@ -7,196 +7,121 @@ permissions: {}
...
+          pr_body="# Description"$'\n'"This PR syncs OpenAPI schemas and (re-)generates Go clients."$'\n\n'
+          pr_body+="| | |"$'\n'"|---|---|"$'\n'
+          pr_body+="| **Schemas** | \`${schemas_summary}\` |"$'\n'
+          pr_body+="| **APIs generated** | \`${apis_summary}\` |"$'\n'
+          pr_body+="| **Packages changed** | \`${changed_packages}\` |"$'\n'
+          pr_body+="| **Workflow run** | [#${WORKFLOW_RUN_NUMBER}](${WORKFLOW_RUN_URL}) |"$'\n'
 
-            pr_body+=$'\n'"| | |"$'\n'"|---|---|"$'\n'
-            pr_body+="| **Workflow run** | [#${WORKFLOW_RUN_NUMBER}](${WORKFLOW_RUN_URL}) |"$'\n'
+          echo "changed=true" >> "$GITHUB_OUTPUT"
```
gkrenn:
```suggestion
          pr_body=$(cat <<EOF
          # Description
          This PR syncs OpenAPI schemas and (re-)generates Go clients.

          | | |
          |---|---|
          | **Schemas** | \`${schemas_summary}\` |
          | **APIs generated** | \`${apis_summary}\` |
          | **Packages changed** | \`${changed_packages}\` |
          | **Workflow run** | [#${WORKFLOW_RUN_NUMBER}](${WORKFLOW_RUN_URL}) |
          EOF
          )

          echo "changed=true" >> "$GITHUB_OUTPUT"
```

gkrenn (changes requested):
Python file looks good, tests are passing. Still want to test your whole pipeline tomorrow.

Some more changes in the workflow I would do.

.github/workflows/sync-openapi-schema.yaml
mihaitanasedt: `EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)`  is a common approach to avoid any collisions ever. I would keep it, better be safe.

.github/workflows/sync-openapi-schema.yaml
mihaitanasedt: I don't think it will work, it will keep the whitespaces and and the table will be off + `EOF` terminator has to be at column 0 for it to be recognized. 

Another approach is: 
```bash
pr_body=$(printf '%s\n' \
            "# Description" \
            "This PR syncs OpenAPI schemas and (re-)generates Go clients." \
            "" \
            "| | |" \
            "|---|---|" \
            "| **Schemas** | \`${schemas_summary}\` |" \
            "| **APIs generated** | \`${apis_summary}\` |" \
            "| **Packages changed** | \`${changed_packages}\` |" \
            "| **Workflow run** | [#${WORKFLOW_RUN_NUMBER}](${WORKFLOW_RUN_URL}) |" \
          )
```

If you think its more readable I can change it 👍

gkrenn: Are you sure? - In yaml the number of spaces you're using to indent a block will be automatically guessed from its first line. I am pretty confident that the whitespaces wouldn't be there. - But I didn't try it out.

Otherwise your solution looks fine as well

.github/workflows/sync-openapi-schema.yaml
```diff
@@ -7,196 +7,121 @@ permissions: {}
...
-    needs: [prepare, clone-source]
+    needs: prepare
     if: needs.prepare.outputs.has_schemas == 'true'
-    runs-on: ubuntu-latest
+    runs-on: operator-e2e
+    environment: oapi
     permissions:
       contents: write
       pull-requests: write
+    # env vars for specUrlEnvVar / authEnvVar referenced in sync-config.yaml must be added here as secrets
```
gkrenn, line 50: As we have the bitbucket secrets configured for this repo - Please add the env vars here already, even though they are not used yet.

gkrenn: Or remove them from the stored GitHub secrets

gkrenn (changes requested):

.github/workflows/sync-openapi-schema.yaml
mihaitanasedt: Done [6841003](https://github.com/Dynatrace/dynatrace-operator/pull/6538/commits/6841003900d80041883e2ef5cd8dd6798a8558f0). 
My mistake, had a smiliar but different problem before. Your approach works! 👍

.github/workflows/sync-openapi-schema.yaml
mihaitanasedt: I would not keep in code reference that are unused or unlikely to be used. Let's leave in the GitHub secrets only the API tokens that might be reused in the future, so they won't be generated every time. Wdyt?

mihaitanasedt:
@gkrenn I retriggered the workflow with the latest changes. 

[workflow run](https://github.com/Dynatrace/dynatrace-operator/actions/runs/25784519745/job/75734642510)

PR https://github.com/Dynatrace/dynatrace-operator/pull/6673

.github/workflows/sync-openapi-schema.yaml
```diff
@@ -7,196 +7,127 @@ permissions: {}
...
-            pr_body+="| **Workflow run** | [#${WORKFLOW_RUN_NUMBER}](${WORKFLOW_RUN_URL}) |"$'\n'
+          echo "changed=true" >> "$GITHUB_OUTPUT"
+          echo "schemas_summary=${schemas_summary}" >> "$GITHUB_OUTPUT"
 
-            EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
-            echo "pr_body<<${EOF}" >> "$GITHUB_OUTPUT"
-            echo "${pr_body}" >> "$GITHUB_OUTPUT"
-            echo "${EOF}" >> "$GITHUB_OUTPUT"
-          fi
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
```
avorima: this looks very bizarre, could you explain?

hack/scripts/oapi-generate.py
gkrenn: Fine for me, makes the code easier to understand 👍

.github/workflows/sync-openapi-schema.yaml
gkrenn: https://github.com/Dynatrace/dynatrace-operator/pull/6538#discussion_r3219228107

It is used as a random string generator. I would still replace it with a static EOF marker, where we know it won't appear in the file.

.github/workflows/sync-openapi-schema.yaml
gkrenn: I am not happy with having secrets which we don't use in our github repo. Either use them or drop them 😄

.github/workflows/sync-openapi-schema.yaml
```diff
@@ -7,196 +7,127 @@ permissions: {}
...
-              continue
-            fi
+      - name: Generate Go SDKs
+        run: make oapi/generate
 
-            old_version=$(git show HEAD:"${target_file}" 2>/dev/null | jq -r '.info.version // "unknown"' 2>/dev/null || echo "N/A")
-            cp "${source_file}" "${target_file}"
+      - name: Check for changes
+        id: sync
+        run: |
```
gkrenn, line 92: Nitpick: Set run as the last property of the job object. i.e. name, id, env should be defined before run.

gkrenn (commented):
Added a small nitpick and we should decide what we do with the secrets, otherwise looks fine for me

.github/workflows/sync-openapi-schema.yaml
avorima: For reference: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#multiline-strings

We would need to introduce EOF in the config ourselves for this randomness to be necessary. I would go for the simpler variant. Doesn't even need the intermediate variable, you can use the `{ ... } > GITHUB_OUTPUT` syntax

.github/workflows/sync-openapi-schema.yaml
mihaitanasedt: Done [61f7f96](https://github.com/Dynatrace/dynatrace-operator/pull/6538/commits/61f7f96f66264814a4468a4ad680443123848e9a)

.github/workflows/sync-openapi-schema.yaml
mihaitanasedt: Like this? [61f7f96](https://github.com/Dynatrace/dynatrace-operator/pull/6538/commits/61f7f96f66264814a4468a4ad680443123848e9a)

.github/workflows/sync-openapi-schema.yaml
mihaitanasedt: I would say either remove them or only keep the Bitbucket access token in the secrets and environment variables. And when a new API needs to be generated from Bitbucket, just add the URL env and reuse the token. We should know that if a token already exists, we should reuse it instead of creating a new one. Wdyt?

.github/workflows/sync-openapi-schema.yaml
```diff
@@ -7,183 +7,119 @@ permissions: {}
...
-        run: make oapi/generate
+          schemas_summary=$(echo "${SCHEMAS_JSON}" | jq -r '[.[].name] | join(", ")')
+          apis_summary=$(echo "${SCHEMAS_JSON}" | jq -r '[.[] | .name + ": " + (.generate.globalProperties.apis // [] | if length == 0 then "all" else join(", ") end)] | join(", ")')
 
-      - name: Tidy Go modules
-        if: steps.sync.outputs.changed == 'true'
-        run: go mod tidy
+          changed_packages=$(git status --porcelain -u | sed 's/^...//' | grep "^${OUTPUT_DIR}/" | sed "s|^${OUTPUT_DIR}/||" | cut -d'/' -f1 | sort -u | tr '\n' ' ' || true)
+
+          pr_body=$(cat <<EOF
```
avorima: This multi-line construction can be simplified

```
foo=somedata
<<EOF
foo
EOF
```

vs

```
<<EOF
somedata
EOF
```

.github/workflows/sync-openapi-schema.yaml
```diff
@@ -7,183 +7,119 @@ permissions: {}
...
+          |---|---|
+          | **Schemas** | \`${schemas_summary}\` |
+          | **APIs generated** | \`${apis_summary}\` |
+          | **Packages changed** | \`${changed_packages}\` |
+          | **Workflow run** | [#${WORKFLOW_RUN_NUMBER}](${WORKFLOW_RUN_URL}) |
+          EOF
+          )
+
+          {
+            echo "changed=true"
```
avorima: this `{ }` construct is only needed for the multi-line variant. the rest can keep using normal `echo >> `

mihaitanasedt: Done [e3394a6](https://github.com/Dynatrace/dynatrace-operator/pull/6538/commits/e3394a6cd13b4d54c2d165fb8bbb15013faddfd6)

.github/workflows/sync-openapi-schema.yaml
mihaitanasedt: Done [e3394a6](https://github.com/Dynatrace/dynatrace-operator/pull/6538/commits/e3394a6cd13b4d54c2d165fb8bbb15013faddfd6)

.github/workflows/sync-openapi-schema.yaml
gkrenn: The problem is in one month we won't remember that this specific secret exists. I would be fine with leaving the secrets in and just mention them as a comment.

gkrenn (dismissed):

.github/workflows/sync-openapi-schema.yaml
mihaitanasedt: Done [dfb8d5e](https://github.com/Dynatrace/dynatrace-operator/pull/6538/commits/dfb8d5eb45cb15d8803ea71f906be4cfc03d157b)

avorima (approved):

PR #6539, opened 2026-05-04
Propagate resource attributes via `-p` args for standalone log module

jokuniew:
## Description

ICP-3792

## How can this be tested?
`make go/test`

manually
deployed dk with 
```yaml
  logMonitoring: {}
  resourceAttributes:
    fookey: foovalue
```
verified logmon yaml that `-p fookey=foovalue` was added

### DIFF
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/args.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/args.go
@@ -2,6 +2,8 @@ package daemonset
 
 import (
 	"fmt"
+	"maps"
+	"slices"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
 )
@@ -24,8 +26,9 @@ func getInitArgs(dk dynakube.DynaKube) []string {
 		baseArgs = append(baseArgs, fmt.Sprintf("-p k8s.cluster.name=$(%s)", clusterNameEnv), fmt.Sprintf("-p dt.entity.kubernetes_cluster=$(%s)", entityEnv))
 	}
 
-	for key, value := range dk.GetResourceAttributes() {
-		baseArgs = append(baseArgs, fmt.Sprintf("-p %s=%s", key, value))
+	attrs := dk.GetResourceAttributes()
+	for _, key := range slices.Sorted(maps.Keys(attrs)) {
+		baseArgs = append(baseArgs, fmt.Sprintf("-p %s=%s", key, attrs[key]))
 	}
 
 	return append(baseArgs, dk.LogMonitoring().Template().Args...)
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/args.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/args.go
@@ -26,12 +26,10 @@ func getInitArgs(dk dynakube.DynaKube) []string {
 		baseArgs = append(baseArgs, fmt.Sprintf("-p k8s.cluster.name=$(%s)", clusterNameEnv), fmt.Sprintf("-p dt.entity.kubernetes_cluster=$(%s)", entityEnv))
 	}
 
-	baseArgs = append(baseArgs, dk.LogMonitoring().Template().Args...)
-
 	attrs := dk.GetResourceAttributes()
 	for _, key := range slices.Sorted(maps.Keys(attrs)) {
 		baseArgs = append(baseArgs, fmt.Sprintf("-p %s=%s", key, attrs[key]))
 	}
 
-	return baseArgs
+	return append(baseArgs, dk.LogMonitoring().Template().Args...)
 }
--- a/pkg/controllers/dynakube/logmonitoring/daemonset/args.go
+++ b/pkg/controllers/dynakube/logmonitoring/daemonset/args.go
@@ -26,10 +26,12 @@ func getInitArgs(dk dynakube.DynaKube) []string {
 		baseArgs = append(baseArgs, fmt.Sprintf("-p k8s.cluster.name=$(%s)", clusterNameEnv), fmt.Sprintf("-p dt.entity.kubernetes_cluster=$(%s)", entityEnv))
 	}
 
+	baseArgs = append(baseArgs, dk.LogMonitoring().Template().Args...)
+
 	attrs := dk.GetResourceAttributes()
 	for _, key := range slices.Sorted(maps.Keys(attrs)) {
 		baseArgs = append(baseArgs, fmt.Sprintf("-p %s=%s", key, attrs[key]))
 	}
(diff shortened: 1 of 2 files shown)


pkg/controllers/dynakube/logmonitoring/daemonset/args_test.go
```diff
@@ -61,4 +65,41 @@ func TestGetInitArgs(t *testing.T) {
 			assert.NotEmpty(t, arg)
 		}
 	})
+
+	t.Run("propagate spec.resourceAttributes as -p args", func(t *testing.T) {
```
mihaitanasedt: If we sort the args based on keys lets also test the order

mihaitanasedt (changes requested):

pkg/controllers/dynakube/logmonitoring/daemonset/args_test.go
jokuniew: gathered last len(number of resourceattributes entries) and compared against expectedAttrs
https://github.com/Dynatrace/dynatrace-operator/pull/6539/changes/009b7cd706ba756c516a77f0b4462b9241647da9

mihaitanasedt (dismissed):

aorcholski (commented):
what if predefined name is used?

dynakube:
```
  resourceAttributes:
    k8s.node.name: overwritten
    fookey: foovalue
```

```
   - args:
     - -p k8s.cluster.uid=$(K8S_CLUSTER_UID)
     - -p k8s.node.name=$(K8S_NODE_NAME) <--
     - -c k8s_fullpodname $(K8S_POD_NAME)
    ...
     - -p fookey=foovalue
     - -p k8s.node.name=overwritten <--
```

jokuniew:
> what if predefined name is used?
> 
> dynakube:
> 
> ```
>   resourceAttributes:
>     k8s.node.name: overwritten
>     fookey: foovalue
> ```
> 
> ```
>    - args:
>      - -p k8s.cluster.uid=$(K8S_CLUSTER_UID)
>      - -p k8s.node.name=$(K8S_NODE_NAME) <--
>      - -c k8s_fullpodname $(K8S_POD_NAME)
>     ...
>      - -p fookey=foovalue
>      - -p k8s.node.name=overwritten <--
> ```

@aorcholski  in yaml it is exactly like in your snippet
image
pod is fine with it 

according to acceptance criteria 
_Existing cluster metadata -p args are not affected._
i think i should not deduplicate/override values.

@waodim can you confirm?

waodim:
> > what if predefined name is used?
> > dynakube:
> > ```
> >   resourceAttributes:
> >     k8s.node.name: overwritten
> >     fookey: foovalue
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> >    - args:
> >      - -p k8s.cluster.uid=$(K8S_CLUSTER_UID)
> >      - -p k8s.node.name=$(K8S_NODE_NAME) <--
> >      - -c k8s_fullpodname $(K8S_POD_NAME)
> >     ...
> >      - -p fookey=foovalue
> >      - -p k8s.node.name=overwritten <--
> > ```
> 
> @aorcholski in yaml it is exactly like in your snippet image pod is fine with it
> 
> according to acceptance criteria _Existing cluster metadata -p args are not affected._ i think i should not deduplicate/override values.
> 
> @waodim can you confirm?

Yes, your approach is the correct one. Pre-defined values should not be overwritten/touched. What was first there, stays. What we maybe could do is maybe add some kind of validation that warns, in the operator when a key in `spec.resourceAttributes` shadows a pre-defined metdata key. I am not sure, however, if there is really a lot of benefit doing this and I question the use case of setting keys that are already pre-defined.

aorcholski (dismissed):

pkg/controllers/dynakube/logmonitoring/daemonset/args.go
```diff
@@ -24,5 +26,10 @@ func getInitArgs(dk dynakube.DynaKube) []string {
 		baseArgs = append(baseArgs, fmt.Sprintf("-p k8s.cluster.name=$(%s)", clusterNameEnv), fmt.Sprintf("-p dt.entity.kubernetes_cluster=$(%s)", entityEnv))
 	}
 
+	attrs := dk.GetResourceAttributes()
+	for _, key := range slices.Sorted(maps.Keys(attrs)) {
+		baseArgs = append(baseArgs, fmt.Sprintf("-p %s=%s", key, attrs[key]))
+	}
+
 	return append(baseArgs, dk.LogMonitoring().Template().Args...)
```
waodim: Can we switch those? I think it would make more sense to at first process the args set in the logMon template before setting the resourceAttributes. WDYT?

pkg/controllers/dynakube/logmonitoring/daemonset/args_test.go
```diff
@@ -61,4 +65,52 @@ func TestGetInitArgs(t *testing.T) {
...
```
waodim: Please also introduce a test case where we set resourceAttributes as well as user-defined args, that are then called via `dk.LogMonitoring().Template().Args`

pkg/controllers/dynakube/logmonitoring/daemonset/args.go
jokuniew: flipped the order here https://github.com/Dynatrace/dynatrace-operator/pull/6539/commits/42cb29c7a8794f2e0065e87e01e69b1f772854d0

pkg/controllers/dynakube/logmonitoring/daemonset/args_test.go
jokuniew: added test here https://github.com/Dynatrace/dynatrace-operator/pull/6539/commits/42cb29c7a8794f2e0065e87e01e69b1f772854d0

and refactor into table driven tests here cause they were pretty repetitive https://github.com/Dynatrace/dynatrace-operator/pull/6539/changes/78c19b8053aa022834578442160fa7b03b988cad

pkg/controllers/dynakube/logmonitoring/daemonset/args.go
waodim: I am dumb. I checked requirements and local args should win over the resourceAttr in general. I guess we need to revert this. Sorry for being confusing 😅

jokuniew: no worries - reverted the logic and adjusted tests https://github.com/Dynatrace/dynatrace-operator/pull/6539/commits/2ae126cf6e6ff39434dff9455c22f1a3519bd227

waodim (approved):
LGTM

PR #6541, opened 2026-05-04
feat: writing pgc to bootstrapperconfig

svfoxat:
## Description
https://dt-rnd.atlassian.net/browse/ICP-1936

- fetching the processgroupconfig from the configured tenant and write it into `dynatrace-bootstrapper-config` secret. 
- checking the size of the response to not write more than 980KiB into the secret (1MiB is the max secret size + some buffer for already existing data in the secret)

## How can this be tested?

- `make go/test`
- deploy to dev cluster, apply dynakube(s), check for the `declarative.cbor` key in the `dynatrace-bootstrapper-config` secret.


pkg/injection/namespace/bootstrapperconfig/pgc.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"bytes"
+	"context"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8sconditions"
+)
+
+const (
```
andriisoldatenko, line 20: const (
    _ = 1 << (10 * iota)
    KiB
)
wdyt to add such a sugar?

`800 * KiB`

svfoxat: something like [c4b4b09e38afc3dda1a1c40b0f12d036204d83e7](https://github.com/Dynatrace/dynatrace-operator/pull/6541/commits/c4b4b09e38afc3dda1a1c40b0f12d036204d83e7)? 

Intent is a bit clearer with that.

pkg/injection/namespace/bootstrapperconfig/pgc.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return nil
+}
+
+func (s *SecretGenerator) preparePGC(ctx context.Context, dk *dynakube.DynaKube) ([]byte, error) {
+	log := logd.FromContext(ctx)
+
+	var buf bytes.Buffer
+
+	_, err := s.dtClient.GetProcessGroupingConfig(ctx, dk.Status.KubernetesClusterMEID, "", &buf)
```
svfoxat: Should we store the ETag of the successful response as annotation on the secret?

pkg/injection/namespace/bootstrapperconfig/pgc.go
```diff
@@ -0,0 +1,60 @@
...
+
+const (
+	_ = 1 << (10 * iota) //nolint:mnd
+	KiB
+)
+
+const (
+	declarativeInputFileName = "declarative.cbor"
+	declarativeWarnSizeBytes = 800 * KiB
+	declarativeMaxSizeBytes  = 980 * KiB
```
avorima: I think 20KiB is way too low of a buffer. I get ~14KiB on my cluster and i don't even do anything crazy. Let's go for 100 to be safe

pkg/injection/namespace/bootstrapperconfig/pgc_test.go
```diff
@@ -0,0 +1,133 @@
...
+			Return("", nil)
+
+		sg := NewSecretGenerator(clt, clt, mockDTClient)
+		result, err := sg.preparePGC(t.Context(), dk)
+
+		require.NoError(t, err)
+		assert.Equal(t, payload, result)
+	})
+
+	t.Run("success - response between 800 KiB and 980 KiB triggers warning", func(t *testing.T) {
```
avorima: this exercises the code path, but doesn't actually verify it. i think we can just use this as the base case, since we always expect it to work

pkg/injection/namespace/bootstrapperconfig/pgc.go
avorima: We should change the API a bit here. We know we need the raw data and an etag, so they can be grouped. This way the caller doesn't need to handle the buffer. Makes test setup simpler as well

```go
type ProcessGroupingConfig struct {
    Etag string
    Data []byte
}

func (c *ClientImpl) GetProcessGroupinConfig(ctx context.Context, kubernetesClusterID, etag string) (ProcessGroupingConfig, error)
```

waodim: Hmm I know this is probably a very unlikely scenario because I guess some reconcilers run that populate `KubernetesClusterMEID` such that it is not empty here, but should we still check it? Return early i empty with a debug log or so? WDYT?

pkg/injection/namespace/bootstrapperconfig/pgc.go
```diff
@@ -0,0 +1,60 @@
...
+	_, err := s.dtClient.GetProcessGroupingConfig(ctx, dk.Status.KubernetesClusterMEID, "", &buf)
+	if err != nil {
+		k8sconditions.SetDynatraceAPIError(dk.Conditions(), ConfigConditionType, err)
+
+		return nil, err
+	}
+
+	size := buf.Len()
+	if size > declarativeMaxSizeBytes {
+		log.Error(nil, "DPG API response exceeds maximum size, skipping declarative.cbor", "size", size, "maxSize", declarativeMaxSizeBytes)
```
waodim: Should we pass a custom error here?

pkg/injection/namespace/bootstrapperconfig/pgc_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/scheme/fake"
+	oneagentclientmock "github.com/Dynatrace/dynatrace-operator/test/mocks/pkg/clients/dynatrace/oneagent"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Test_SecretGenerator_prepareDeclarativeConfig(t *testing.T) {
```
mihaitanasedt:
```suggestion
func Test_SecretGenerator_preparePGC(t *testing.T) {
```

pkg/injection/namespace/bootstrapperconfig/pgc_test.go
```diff
@@ -0,0 +1,133 @@
...
+			Return("", nil)
+
+		sg := NewSecretGenerator(clt, clt, mockDTClient)
+		result, err := sg.preparePGC(t.Context(), dk)
+
+		require.NoError(t, err)
+		assert.Nil(t, result)
+	})
+
+	t.Run("API error is propagated and condition is set", func(t *testing.T) {
```
mihaitanasedt, line 105: It doesn't test if the condition is set

mihaitanasedt (dismissed):

pkg/injection/namespace/bootstrapperconfig/pgc.go
avorima: using nil is valid if you just want to print a message at error level https://pkg.go.dev/github.com/go-logr/logr#Logger.Error

pkg/injection/namespace/bootstrapperconfig/pgc.go
svfoxat: I added a early return and a log statement, just in case.

pkg/injection/namespace/bootstrapperconfig/pgc.go
svfoxat: Good idea, I lowered the thresholds.

pkg/injection/namespace/bootstrapperconfig/pgc_test.go
svfoxat: Good catch, I added the check for the condition in the test.

pkg/clients/dynatrace/oneagent/processgroupingconfig.go
```diff
@@ -15,19 +15,24 @@ const (
...
 //     If the server responds with 304 Not Modified, the underlying HTTP error is returned.
 //     Use core.HasStatusCode(err, http.StatusNotModified) to check for this case.
-//   - writer: the io.Writer to stream the CBOR response body into.
 //
 // Returns:
-//   - The ETag value from the response header on success (HTTP 200), or the original etag on 304.
-//   - An error if the request failed. On HTTP 304, the error satisfies core.HasStatusCode(err, http.StatusNotModified).
-func (c *ClientImpl) GetProcessGroupingConfig(ctx context.Context, kubernetesClusterID string, etag string, writer io.Writer) (string, error) {
+//   - On HTTP 200: *ProcessGroupConfig with ETag from response header and CBOR data.
+//   - On HTTP 304: *ProcessGroupConfig with the original ETag and nil Data, plus a non-nil error satisfying core.HasStatusCode(err, http.StatusNotModified).
```
avorima: We should do another simplification, because we're checking the 304 status twice. The actual info we want is whether the etag we get out is equal to the one we put in.

```go
headers, err := req.ExecuteWriter(&buf)
if err != nil && !core.IsNotModified(err) {
    return nil, err
}

// body is empty on 304, so this is fine
return &PCG{Etag: header.Etag, Data: buf.Bytes()}, nil
```

```go
pcg, err := GetProcessGroupingConfig(ctx, "meID", "old-etag")
if err != nil { ... }
if pcg.Etag == "old-etag" {
    // no change, use old-etag
}
// change
```

pkg/clients/dynatrace/oneagent/processgroupingconfig_integration_test.go
avorima, line 1: I'm thinking that we can drop this test. To check if the integration works we can look at the secret

pkg/controllers/dynakube/injection/reconciler_test.go
```diff
@@ -126,6 +129,7 @@ func TestReconciler(t *testing.T) {
 		versionClient := versionclientmock.NewClient(t)
 		versionClient.EXPECT().GetLatestAgentVersion(anyCtx, mock.AnythingOfType("string"), mock.AnythingOfType("string")).Return("", nil)
 		oneAgentClient.EXPECT().GetProcessModuleConfig(anyCtx).Return(&oneagentclient.ProcessModuleConfig{}, nil).Once()
+		oneAgentClient.EXPECT().GetProcessGroupingConfig(anyCtx, mock.Anything, "").Return(&oneagentclient.ProcessGroupConfig{}, nil).Once()
```
avorima: Since we're mocking everything, we should assert simple inputs when we can. A constant would make this easier

```suggestion
		oneAgentClient.EXPECT().GetProcessGroupingConfig(anyCtx, "KUBERNETES_CLUSTER-test", "").Return(&oneagentclient.ProcessGroupConfig{}, nil).Once()
```

pkg/injection/namespace/bootstrapperconfig/pgc.go
```diff
@@ -0,0 +1,94 @@
...
+		annotations[annotationPGCETag] = pgc.ETag
+	}
+
+	return nil
+}
+
+func (s *SecretGenerator) preparePGC(ctx context.Context, dk *dynakube.DynaKube) (*oneagent.ProcessGroupConfig, error) {
+	log := logd.FromContext(ctx)
+
+	if dk.Status.KubernetesClusterMEID == "" {
```
avorima, line 45: the client treats this value is optional, so we should align the behavior. make it required on the client side so that we don't accidentally pass in an empty ID

pkg/injection/namespace/bootstrapperconfig/pgc.go
```diff
@@ -0,0 +1,94 @@
...
+		log.Info("kubernetesClusterMEID not available, skipping processgroupingconfig")
+
+		return nil, nil //nolint:nilnil
```
avorima: preparePGC can set the necessary values, so this nolint directive can be removed

pkg/injection/namespace/bootstrapperconfig/pgc.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	return pgc, nil
+}
+
+func (s *SecretGenerator) readCachedPGC(ctx context.Context, dk *dynakube.DynaKube) *oneagent.ProcessGroupConfig {
+	var secret corev1.Secret
+
+	key := types.NamespacedName{Name: GetSourceConfigSecretName(dk.Name), Namespace: dk.Namespace}
+	_ = s.apiReader.Get(ctx, key, &secret)
```
avorima: a bit paranoid, but if this fails for any reason we use empty values

avorima:
The endpoint is on its way to become GA, but there might be a gap to our targeted release date. The bootstrapper should ignore a 404 error with a log, i.e. endpoint not available, without causing a reconcile failure. This can be handled in the client, same as for metadata enrichment settings for example

andriisoldatenko:
> without causing a reconcile failure. This can be handled in the client, same as for metadata enrichment settings for example

agree

svfoxat:
> The endpoint is on its way to become GA, but there might be a gap to our targeted release date. The bootstrapper should ignore a 404 error with a log, i.e. endpoint not available, without causing a reconcile failure. This can be handled in the client, same as for metadata enrichment settings for example

addressed in [d83fe66](https://github.com/Dynatrace/dynatrace-operator/pull/6541/commits/d83fe667035a5c647cd80722956a85edd8f7993c)

pkg/clients/dynatrace/oneagent/processgroupingconfig.go
```diff
@@ -31,8 +32,11 @@ type ProcessGroupConfig struct {
 // Returns:
 //   - On HTTP 200: *ProcessGroupConfig with ETag from response header and CBOR data, nil error.
 //   - On HTTP 304: *ProcessGroupConfig with the original ETag and nil Data, nil error.
+//   - On HTTP 404: *ProcessGroupConfig (empty), nil error. Endpoint not available.
 //   - On other errors: non-nil error.
 func (c *ClientImpl) GetProcessGroupingConfig(ctx context.Context, kubernetesClusterID string, etag string) (*ProcessGroupConfig, error) {
+	log := logd.FromContext(ctx)
```
avorima: Should use NewFromContext

avorima (approved):

pkg/injection/namespace/bootstrapperconfig/pgc.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (s *SecretGenerator) addPGC(ctx context.Context, dk *dynakube.DynaKube, data map[string][]byte, annotations map[string]string) error {
+	pgc, err := s.preparePGC(ctx, dk)
+	if err != nil {
+		return err
+	}
+
+	if pgc != nil && len(pgc.Data) != 0 {
+		data[declarativeInputFileName] = pgc.Data
+		annotations[annotationPGCETag] = pgc.ETag
```
0sewa0, line 36: does it make sense to use a mix of annotation/data?
why can't be the ETag in the data aswell?
- if we worry that it will be mounted as a file, would that bother anyone?

IMO it would simplify the code if we didn't do this mix of annotations/data.

svfoxat: > does it make sense to use a mix of annotation/data?

I'd say yes, annotations are the place to hold metadata about the data [docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#:~:text=Build%2C%20release%2C%20or%20image%20information%20like%20timestamps%2C%20release%20IDs%2C%20git%20branch%2C%20PR%20numbers%2C%20image%20hashes%2C%20and%20registry%20address.)

> if we worry that it will be mounted as a file, would that bother anyone?

I guess no, because the bootstraper won't copy this additional file then, if you want to store the etag in a field like `data["pgc-etag"]`

0sewa0: > annotations are the place to hold metadata about the data

sounds reasonable, it is maybe a bit more "complex" but also more correct, so lets leave it as is 👍

PR #6542, opened 2026-05-04
Simplify isAllowedSuffixAPIServer

avorima:
## Description

Refactor of `isAllowedSuffixAPIServer` in `pkg/api/validation/edgeconnect/api_server.go`. No behavior change for any realistic hostname.

**Motivation.** The previous implementation split both the configured `apiServer` and each candidate suffix on `.`, then compared field counts to assert that a tenant label was present. Because every entry in `allowedSuffix` begins with a literal `.`, this is equivalent to "input is longer than the suffix" — but the indirection (two `strings.FieldsFunc` calls, a `break`-then-fall-through, and a now-unused `log.Info` branch) obscured the intent.

**Change.** Replace the field-split logic with `strings.HasSuffix(host, suffix) && len(host) > len(suffix)`. The function drops from 25 lines to 9, the unused `ctx` parameter becomes `_`, and the validator-registry signature is preserved. Behavior is identical for well-formed hostnames; the only divergence is on non-realistic inputs containing consecutive dots (e.g. `..apps.dynatrace.com`), where the new version is strictly no stricter than the old.

No linked Jira/GitHub issue — drive-by simplification spotted while reviewing the EdgeConnect validation package.

## How can this be tested?

No environment setup required; this is a pure-Go change covered by existing unit tests.

- `make go/lint` — clean
- `go test ./pkg/api/validation/edgeconnect/...` — passes

### DIFF
--- a/pkg/api/validation/edgeconnect/api_server.go
+++ b/pkg/api/validation/edgeconnect/api_server.go
@@ -36,26 +36,10 @@ var (
 	}
 )
 
-func isAllowedSuffixAPIServer(ctx context.Context, _ *Validator, ec *edgeconnect.EdgeConnect) string {
-	log := logd.FromContext(ctx)
-
+func isAllowedSuffixAPIServer(_ context.Context, _ *Validator, ec *edgeconnect.EdgeConnect) string {
 	for _, suffix := range allowedSuffix {
-		if strings.HasSuffix(ec.Spec.APIServer, suffix) {
-			hostnameWithDomains := strings.FieldsFunc(suffix,
-				func(r rune) bool { return r == '.' },
-			)
-
-			hostnameWithTenant := strings.FieldsFunc(ec.Spec.APIServer,
-				func(r rune) bool { return r == '.' },
-			)
-
-			if len(hostnameWithTenant) > len(hostnameWithDomains) {
-				return ""
-			}
-
-			log.Info("apiServer is not a valid hostname", "apiServer", ec.Spec.APIServer)
-
-			break
+		if strings.HasSuffix(ec.Spec.APIServer, suffix) && len(ec.Spec.APIServer) > len(suffix) {
+			return ""
 		}
 	}
 
--- a/pkg/api/validation/edgeconnect/api_server.go
+++ b/pkg/api/validation/edgeconnect/api_server.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha2/edgeconnect"
-	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


mihaitanasedt:
Since we touch this file, can you also remove the log from `checkAPIServerProtocolNotSet` in the same file? seems redundant to me

avorima:
> Since we touch this file, can you also remove the log from `checkAPIServerProtocolNotSet` in the same file? seems redundant to me

We don't want users to set the protocol, since we always use https. Removing the validation means we need to handle it in code, which could lead to wrong expectations from users

mihaitanasedt:
> > Since we touch this file, can you also remove the log from `checkAPIServerProtocolNotSet` in the same file? seems redundant to me
> 
> We don't want users to set the protocol, since we always use https. Removing the validation means we need to handle it in code, which could lead to wrong expectations from users

I mean to remove this logging line, not the whole validation

https://github.com/Dynatrace/dynatrace-operator/blob/a17e28504b6cd34102ee8f7aec7a6c818a1415bc/pkg/api/validation/edgeconnect/api_server.go#L54

aorcholski (commented):
```
..dev.apps.dynatracelabs.com
...dev.apps.dynatracelabs.com
```
are rejected by the current code and approved by the PR 😄

aorcholski (dismissed):

avorima:
> ```
> ..dev.apps.dynatracelabs.com
> ...dev.apps.dynatracelabs.com
> ```
> 
> are rejected by the current code and approved by the PR 😄

I actually prepared a second options which is 

```go
tenant := strings.TrimSuffix(ec.Spec.APIServer, suffix)
if tenant != ec.Spec.APIServer && tenant != "" && !strings.Contains(tenant, ".") {
```

This would be stricter than the current validation, since it also disallows `foo.bar.apps.dynatrace.com`. I didn't know whether that was a use-case so I went with the simpler option

mihaitanasedt:
What about a using a regex?

avorima:
> What about a using a regex?

For simple suffix comparison this is enough

andriisoldatenko (approved):

pkg/api/validation/edgeconnect/api_server.go
```diff
@@ -36,39 +35,19 @@ var (
...
-
-			hostnameWithTenant := strings.FieldsFunc(ec.Spec.APIServer,
-				func(r rune) bool { return r == '.' },
-			)
-
-			if len(hostnameWithTenant) > len(hostnameWithDomains) {
-				return ""
-			}
-
-			log.Info("apiServer is not a valid hostname", "apiServer", ec.Spec.APIServer)
```
andriisoldatenko, line 56: i think keeping invalid option in logs is not bad idea, for troubleshooting, but we already have it inside EC spec as part of support archive.

avorima: it should be immediately obvious why a URL got rejected from just looking at the spec

PR #6544, opened 2026-05-05
Propagate resource attributes to `deployment.properties` for ActiveGate

gkrenn:
## Description

Propagates the new resource attributes from the dynakube field to the activegate.

It does this by creating a new config map and mounts it as volume to the activegate.

[ICP-3804 Propagate resource attributes to `deployment.properties` for ActiveGate - Jira](https://dt-rnd.atlassian.net/browse/ICP-3804)

## How can this be tested?

- Apply regular dk with ag: nothing new should happen.
- Apply regular dk with ag and resource attributes set:
  - `/var/lib/dynatrace/secrets/config/deployment.properties` should be mounted on the ag pod
  - `kubectl -n dynatrace logs dynakube-activegate-0 | grep deployment.prop -A 5` should indicate that the file was processed by the ag.

### DIFF
--- a/pkg/controllers/dynakube/activegate/consts/consts.go
+++ b/pkg/controllers/dynakube/activegate/consts/consts.go
@@ -20,6 +20,10 @@ const (
 	AuthTokenSecretVolumeName = "ag-authtoken-secret"
 	AuthTokenMountPoint       = connectioninfo.TokenBasePath + "/auth-token"
 
+	DeploymentPropertiesVolumeName = "deployment-properties"
+	DeploymentPropertiesFileName   = "deployment.properties"
+	DeploymentPropertiesBasePath   = "/var/lib/dynatrace/secrets/config"
+
 	EnvDTCapabilities    = "DT_CAPABILITIES"
 	EnvDTIDSeedNamespace = "DT_ID_SEED_NAMESPACE"
 	EnvDTIDSeedClusterID = "DT_ID_SEED_K8S_CLUSTER_ID"
--- a/pkg/controllers/dynakube/activegate/deploymentproperties/deploymentproperties.go
+++ b/pkg/controllers/dynakube/activegate/deploymentproperties/deploymentproperties.go
@@ -1,0 +1,26 @@
+package deploymentproperties
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+func BuildContent(attrs map[string]string) string {
+	keys := make([]string, 0, len(attrs))
+	for k := range attrs {
+		keys = append(keys, k)
+	}
+
+	sort.Strings(keys)
+
+	var sb strings.Builder
+
+	sb.WriteString("[resource_attributes]\n")
+
+	for _, k := range keys {
+		fmt.Fprintf(&sb, "%s = %s\n", k, attrs[k])
+	}
+
+	return sb.String()
+}
--- a/pkg/controllers/dynakube/activegate/deploymentproperties/deploymentproperties.go
+++ b/pkg/controllers/dynakube/activegate/deploymentproperties/deploymentproperties.go
@@ -2,7 +2,7 @@ package deploymentproperties
 
 import (
 	"fmt"
-	"sort"
+	"slices"
 	"strings"
 )
 
@@ -12,7 +12,7 @@ func BuildContent(attrs map[string]string) string {
 		keys = append(keys, k)
 	}
 
-	sort.Strings(keys)
+	slices.Sort(keys)
 
 	var sb strings.Builder
 
--- a/pkg/controllers/dynakube/activegate/deploymentproperties/deploymentproperties.go
+++ b/pkg/controllers/dynakube/activegate/deploymentproperties/deploymentproperties.go
@@ -2,16 +2,17 @@ package deploymentproperties
 
 import (
 	"fmt"
+	"maps"
 	"slices"
 	"strings"
 )
 
 func BuildContent(attrs map[string]string) string {
-	keys := make([]string, 0, len(attrs))
-	for k := range attrs {
-		keys = append(keys, k)
+	if len(attrs) == 0 {
+		return ""
 	}
 
+	keys := slices.Collect(maps.Keys(attrs))
 	slices.Sort(keys)
 
 	var sb strings.Builder
--- a/pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/deploymentproperties.go
+++ b/pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/deploymentproperties.go
@@ -26,7 +26,7 @@ type DeploymentPropertiesModifier struct {
 }
 
 func (mod DeploymentPropertiesModifier) Enabled() bool {
-	return true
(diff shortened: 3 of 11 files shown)


pkg/controllers/dynakube/activegate/consts/consts.go
```diff
@@ -20,6 +20,10 @@ const (
 	AuthTokenSecretVolumeName = "ag-authtoken-secret"
 	AuthTokenMountPoint       = connectioninfo.TokenBasePath + "/auth-token"
 
+	DeploymentPropertiesVolumeName = "deployment-properties"
+	DeploymentPropertiesFileName   = "deployment.properties"
+	DeploymentPropertiesBasePath   = "/var/lib/dynatrace/secrets/config"
```
andriisoldatenko, line 25: we use this location a lot everywhere, would be nice to have reusable constant. wdyt? may be unrealted to your PR

pkg/controllers/dynakube/activegate/deploymentproperties/deploymentproperties.go
```diff
@@ -0,0 +1,26 @@
+package deploymentproperties
+
+import (
+	"fmt"
+	"slices"
+	"strings"
+)
+
+func BuildContent(attrs map[string]string) string {
+	keys := make([]string, 0, len(attrs))
```
andriisoldatenko: maps.Keys()?

andriisoldatenko:
```
keys := slices.Collect(maps.Keys(m))
```

pkg/controllers/dynakube/activegate/internal/statefulset/reconciler.go
```diff
@@ -101,12 +102,14 @@ func (r *Reconciler) calculateActiveGateConfigurationHash(ctx context.Context, d
 		return "", err
 	}
 
-	if len(customPropertyData) < 1 && len(authTokenData) < 1 {
+	resourceAttributesData := deploymentproperties.BuildContent(dk.Spec.ResourceAttributes)
+
+	if len(customPropertyData) < 1 && len(authTokenData) < 1 && len(resourceAttributesData) < 1 {
```
andriisoldatenko:
```suggestion
	if len(customPropertyData) == 0 && len(authTokenData) == 0 && len(resourceAttributesData) == 0 {
```
or
```
if allEmpty(customPropertyData, authTokenData, resourceAttributesData) {
    return "", nil
}

func allEmpty[T any](slices ...[]T) bool {
    for _, s := range slices {
        if len(s) > 0 {
            return false
        }
    }
    return true
}
```

jokuniew: `resourceAttributesData` will always have at least `"[resource_attributes]\n"` text in there, so this if will always be false so we'll never return early

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/deploymentproperties.go
```diff
@@ -0,0 +1,74 @@
+package modifiers
+
+import (
+	"path"
```
jokuniew: > Avoid using the `path` package for operations on filesystem paths. Use "path/filepath" package.

https://github.com/Dynatrace/dynatrace-operator/blob/main/doc/coding-style-guide.md?plain=1#L50

pkg/controllers/dynakube/activegate/consts/consts.go
gkrenn: Will do it in a follow up PR 👍

pkg/controllers/dynakube/activegate/deploymentproperties/deploymentproperties.go
gkrenn: Done

pkg/controllers/dynakube/activegate/internal/statefulset/reconciler.go
gkrenn: Done

pkg/controllers/dynakube/activegate/deploymentproperties/deploymentproperties.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"fmt"
+	"maps"
+	"slices"
+	"strings"
+)
+
+func BuildContent(attrs map[string]string) string {
+	if len(attrs) == 0 {
```
andriisoldatenko, line 11: do we really need this check? if so why 
```
t.Run("empty map produces empty section", func(t *testing.T) {
		content := BuildContent(map[string]string{})
		assert.Equal(t, "[resource_attributes]\n", content)
	})
```
works?

```
package main

import "fmt"

func main() {
	m := map[string]string{}
	if len(m) == 0 {
		fmt.Println("Hello, 世界")
	}

}
// Hello, 世界
```

andriisoldatenko: ah okey: 
[CI / Run unit tests (pull_request)](https://github.com/Dynatrace/dynatrace-operator/actions/runs/25427068191/job/74583278904?pr=6544)
CI / Run unit tests (pull_request)Failing after 4m

it explains it :)

pkg/controllers/dynakube/activegate/internal/statefulset/reconciler.go
gkrenn: Removed the if as the hash annotation, even if it is empty, is always added to the ag stateful set. In addition we always have an auth token, so the if was pretty useless.

gkrenn: Removed the if as the hash annotation, even if it is empty, is always added to the ag stateful set. In addition we always have an auth token, so the if was pretty useless.

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/deploymentproperties.go
gkrenn: Done

pkg/controllers/dynakube/activegate/deploymentproperties/deploymentproperties.go
gkrenn: I was distracted by another task, now the PR is ready for review again.

jokuniew (approved):

waodim (approved):
```
2026-05-07 08:12:46 UTC INFO	[agctl] Setting properties based on: /var/lib/dynatrace/secrets//config/deployment.properties
2026-05-07 08:12:46 UTC INFO	[agctl] Saving properties to: /var/lib/dynatrace/gateway/config/config.properties
2026-05-07 08:12:46 UTC INFO	[agctl] [resource_attributes]
2026-05-07 08:12:46 UTC INFO	[agctl] a = b
2026-05-07 08:12:46 UTC INFO	[agctl] custom.only = spec-value
2026-05-07 08:12:46 UTC INFO	[agctl] test = name
```

LGTM

PR #6550, opened 2026-05-06
Skip scope checks when platform token is used

avorima:
## Description

The API to lookup token scopes will not be available when using a platform token. This change adds HTTP status 403 handling to the reconcilers that make use of the optional scopes.
Scope check is skipped when platform token is set. Any related conditions or status fields are cleared.

Some fly-by changes:
* Fix a bug while rendering the logmonitoring config. This actually affects main as well, but it seems easier to trigger with this branch.
* Reduce duplication in HTTP errors. In most cases the body and HTTP header should have the same status code, so mentioning it twice is redundant.

ICP-3719

### API changes

* Add `status.apiToken.platform` boolean field. It's set to either `false` or `true` depending if the user provided a platform token.

## How can this be tested?

> [!IMPORTANT]
> The tenant registry does not and will not support platform tokens so don't use it (mainly affects OA and AG)

Generate a platform token, e.g. [here](https://myaccount-dev.dynatracelabs.com/platformTokens)
* `fleet-management:oneagents:download`
* `fleet-management:activegate.connection-info:read`
* `fleet-management:oneagent.connection-info:read`
* `fleet-management:activegate.tokens:create`
* `settings:objects:read`
* `settings:objects:write`

Deploy a DynaKube with KSPM and Logmonitoring

```yaml
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
  annotations:
    feature.dynatrace.com/k8s-app-enabled: "true"
spec:
  activeGate:
    image: public.ecr.aws/dynatrace/dynatrace-activegate:1.335.20.20260403-154514
    capabilities:
      - routing
      - kubernetes-monitoring
  logMonitoring: {}
  kspm: {}
  templates:
    logMonitoring:
      imageRef:
        repository: public.ecr.aws/dynatrace/dynatrace-logmodule
        tag: 1.335.58.20260423-145051
```

Update the token secret from classic to platform and back to see the status change.


mihaitanasedt:
What do you think about making `AvailableOptionalScopes` a pointer? To make it nil when platform token is used and remove it from the resource status. Otherwise its just there empty 
Screenshot 2026-05-06 at 16 03 16

https://github.com/Dynatrace/dynatrace-operator/blob/0c82673ce1b43bb8cb96703bd94dbe7e9760affa/pkg/api/latest/dynakube/dynakube_status.go#L71

Besides that everything looks good, did some manual testing and it works 👍

cmd/troubleshoot/dynakube.go
```diff
@@ -127,6 +128,12 @@ func checkDynatraceAPITokenScopes(ctx context.Context, baseLog logd.Logger, apiR
 		return errors.Wrapf(err, "invalid '%s:%s' secret", dk.Namespace, dk.Tokens())
 	}
 
+	if dttoken.IsPlatform(tokens.APIToken().Value) {
+		logInfof(log, "skipping token scope lookup due to platform token")
```
waodim, line 132: Why not make this check at the very beginning?

avorima: building the client and verifying token values should still happen to catch a general misconfiguration. although i think the verify values call should happen while reading the tokens from secret tbh

avorima:
> What do you think about making `AvailableOptionalScopes` a pointer? To make it nil when platform token is used and remove it from the resource status. Otherwise its just there empty Screenshot 2026-05-06 at 16 03 16
> 
> https://github.com/Dynatrace/dynatrace-operator/blob/0c82673ce1b43bb8cb96703bd94dbe7e9760affa/pkg/api/latest/dynakube/dynakube_status.go#L71
> 
> Besides that everything looks good, did some manual testing and it works 👍

We have a ticket to tackle rendering empty values as a whole ICP-973

mihaitanasedt (approved):

andriisoldatenko (approved):

PR #6553, opened 2026-05-06
use-public-registry FF + dk.spec.PublicRegistryOverride

jokuniew:
## Description
ICP-3021
create
- new Feature Flag `use-public-registry`
- DynaKube spec field `PublicRegistryOverride`
- `images` subclient that calls `/fleetManagement/components/containerImages`
- implement `LatestImageInfo` in updater for every component oa/ag/cm - write new imageURI to k8s object `image:`
- Dynakube's validation error when user specify `PublicRegistryOverride` without the Feature flag `use-public-registry`

removed
- `Maybe()` calls in updater's tests 

## How can this be tested?
`make go/test`

### manual e2e test for oa/ag
-  get pull secret & deploy as secret on kubernetes
- specify `customPullSecret` in Dynakube
- specify `use-public-registry` and eventually `publicRegistryOverride` in dk
- deploy and check if oa/ag are deployed with right image

### testing codemodules
- `ENABLE_CSI=false make deploy`
- deploy pull secret in target app namespace
- deploy dynakube with `use-public-registry` and `node-image-pull` set to `true` 
- deploy app that uses pull secret capable to download image from public registry
- inspect deployment if uses desired image in init container


pkg/api/exp/flag.go
```diff
@@ -33,6 +34,10 @@ func (ff *FeatureFlags) UseEECLegacyMounts() bool {
 	return ff.getBoolWithDefault(UseEECLegacyMountsKey, true)
 }
 
+func (ff *FeatureFlags) IsPublicRegistry() bool {
+	return ff.getBoolWithDefault(UsePublicRegistryKey, false)
```
avorima, line 39: The default value here depends on whether a platform token is used. I don't see this being implemented, will this be done in a follow-up?

pkg/clients/dynatrace/images/client_test.go
```diff
@@ -0,0 +1,132 @@
...
+		client := setupClient(t, nil, map[string]string{}, expectedImageURI)
+		_, err := client.ComponentLatestImageInfo(t.Context(), "aasddasd", "")
+		require.Error(t, err)
+	})
+
+	t.Run("api error 404", func(t *testing.T) {
+		client := setupClient(t, &core.HTTPError{StatusCode: 404, Message: "nope"}, map[string]string{}, expectedImageURI)
+		_, err := client.ComponentLatestImageInfo(t.Context(), ActiveGate, "")
+		require.True(t, core.IsNotFound(err))
+		assert.EqualError(t, err, "get latest activegate image: nope")
```
avorima: tests should not depend on the message to assert what kind of error we're dealing with

pkg/clients/dynatrace/image/client.go
avorima, line 1: Package name should use singular: `image.Client`, `image.OneAgent`, `image.Info`

pkg/clients/dynatrace/images/client.go
```diff
@@ -0,0 +1,108 @@
...
+	Tag      string
+	Registry string
+}
+
+const (
+	containerImagesPath = "/v2/fleetManagement/components/containerImages"
+)
+
+type Client interface {
+	ComponentLatestImageInfo(ctx context.Context, component ComponentType, registry string) (*ImageInfo, error)
```
avorima: client methods usually include verbs like "get"

pkg/controllers/dynakube/version/updater.go
```diff
@@ -101,6 +117,23 @@ func determineSource(updater StatusUpdater) status.VersionSource {
...
+	log := logd.FromContext(ctx)
+
+	log.Info("updating image version info",
+		"image", imageInfo,
+		"oldImageID", target.ImageID)
+
+	target.Version = imageInfo.Tag
+	target.ImageID = imageInfo.URI
+
+	log.Info("updated image version info",
```
avorima, line 130: this information can be provided in one log message

pkg/api/exp/flag.go
jokuniew: as far as I know it will be done in follow-up [ICP-3643](https://dt-rnd.atlassian.net/browse/ICP-3643) here

pkg/clients/dynatrace/image/client.go
jokuniew: `images` -> `image` https://github.com/Dynatrace/dynatrace-operator/pull/6553/changes/a84d2b03ae7d6ec4ab1df7b0de228afae080fec9

`image.ImageInfo` -> `image.Info` https://github.com/Dynatrace/dynatrace-operator/pull/6553/changes/dae8fca060c12bfdd9e09faba625b09687a0ee36

pkg/clients/dynatrace/images/client_test.go
jokuniew: EqualError -> Error https://github.com/Dynatrace/dynatrace-operator/pull/6553/changes/f7b87383507b5686406577df7ac3e43d326f4391

pkg/clients/dynatrace/images/client.go
jokuniew: [ComponentLatestImageInfo -> GetComponentLatestInfo](https://github.com/Dynatrace/dynatrace-operator/pull/6553/changes/e9589dbeb7e909c4142eb4a86e96c2900a394d86)

pkg/controllers/dynakube/version/updater.go
jokuniew: saved previous value and log once 
done for other log messages in this file as well
https://github.com/Dynatrace/dynatrace-operator/pull/6553/changes/f171d74617388ad197b40b66f351645f6f4e2c8a

pkg/api/exp/flag.go
andriisoldatenko: @jokuniew please add TODO, link ticket, and add comment. it will be easier to fix in next PR.

jokuniew: todo added https://github.com/Dynatrace/dynatrace-operator/pull/6553/commits/f85e92d5c715123a5913d0b4ccfa723883e7838a

pkg/api/validation/dynakube/public_registry.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/exp"
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+)
+
+const (
+	errorPublicRegistryOverrideWithoutPublicRegistry = `The publicRegistryOverride field is set, but the feature flag "%s" is not enabled. Either enable the feature flag or remove the publicRegistryOverride field.`
```
0sewa0, line 12: nitpick: similar TODO applies as in `pkg/api/exp/flag.go` 

> TODO: ICP-3643 - Implement public registry selection based on gen3 platformToken; if user defines FF use-public-registry, ignore it and display warning when platformToken is used.

pkg/clients/dynatrace/image/client.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+type componentResponse struct {
+	Type     ComponentType `json:"type"`
+	ImageURI string        `json:"imageUri"`
+}
+
+type containerImagesResponse struct {
+	Components []componentResponse `json:"components"`
+}
```
0sewa0, line 42: nitpick: could already use the cache

only a nitpick because this can be handled, when we move away from the version-reconcilers status timestamp

```suggestion
type containerImagesResponse struct {
	Components []componentResponse `json:"components"`
}

func (cir *containerImagesResponse) IsEmpty() bool {
	return len(cir.Components) == 0
}
```

pkg/controllers/dynakube/version/updater.go
```diff
@@ -78,20 +80,34 @@ func (r *reconciler) run(ctx context.Context, updater StatusUpdater) error {
...
+	if err != nil {
+		log.Info("could not get public image", "updater", updater.Name(), "error", err)
+
+		return err
+	}
+
+	isDowngrade, err := updater.CheckForDowngrade(ctx, publicImage.Tag)
+	if err != nil || isDowngrade {
+		return err
+	}
```
0sewa0, line 97: AFAIK, only host agent are not allowed to downgrade, but this will enforce it for EVERYTHING

0sewa0 (changes requested):

pkg/controllers/dynakube/version/updater.go
jokuniew: i need a bit more context here. Why enforcing for everything is not correct? 
CheckForDowngrade for activegate and codemodules always `return false, nil`
so this `if` section will be skipped

0sewa0: AH, me dumb 🙈 

you are correct 👍

pkg/api/validation/dynakube/public_registry.go
jokuniew: added the same comment for sake of context
https://github.com/Dynatrace/dynatrace-operator/pull/6553/commits/e866aec2ee659781e6589382a84edcdb99b4c810
but wouldn't it in practive mean `if platform token -> ignore this check at all`?

pkg/clients/dynatrace/image/client.go
jokuniew: added IsEmpty
https://github.com/Dynatrace/dynatrace-operator/pull/6553/changes/318a58a0564d606a1c7c64a9b2c3f666f3f43656

pkg/api/validation/dynakube/public_registry.go
0sewa0: yes, but shouldn't the validation error also mention the platform token?
Like `The publicRegistryOverride field is set, but the feature flag "%s" is not enabled nor is a platformToken used`

jokuniew: true
added more details to todo
https://github.com/Dynatrace/dynatrace-operator/pull/6553/commits/d74c9c57b4e82204710e6f88e704c8cc383a5d36

pkg/clients/dynatrace/image/client.go
```diff
@@ -0,0 +1,112 @@
...
+	if registry != "" {
+		params["registry"] = registry
+	}
+
+	err := c.apiClient.GET(ctx, containerImagesPath).WithQueryParams(params).Execute(&resp)
+	if err != nil {
+		return nil, fmt.Errorf("get latest %s image: %w", component, err)
+	}
+
+	if len(resp.Components) == 0 {
```
0sewa0:
```suggestion
	if resp.IsEmpty() {
```

avorima (approved):
We should get to removing the version reconciler timestamp. It's bad UX to have to wait 15 minutes for another image lookup to happen after changing the registry override

0sewa0 (approved):

PR #6554, opened 2026-05-06
Changes permissions of the appmounts directory

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/ICP-4570)

## Description

Changes permissions of the `appmounts` directory and `upperdir` paths  from 0777 to 0755.
Permissions of existing upperdir paths  are not changed after operator update.

## How can this be tested?
Use CSI Server shell to check permissions of `/data/*` files and directories.

1) New installation.

make cleanup/node-fs
Deploy operator from the branch
Deploy CNFS DK
Deploy sample app
Remove everything

Expected permissions:
- appmounts - 0755
- appmounts/volumeID - 0755
- appmounts/volumeID/var - 0755

2) Update

make cleanup/node-fs
Deploy operator from the `main` branch
Deploy CNFS DK
Deploy sample app
Update operator to the branch version
Install new app (or reinstall old app)

Expected permissions (old app):
- appmounts - 0775
- appmounts/volumeID - 0777
- appmounts/volumeID/var - 0777

Expected permissions (new app):
- appmounts - 0775
- appmounts/volumeID - 0755
- appmounts/volumeID/var - 0755

### DIFF
--- a/pkg/controllers/csi/metadata/correctness.go
+++ b/pkg/controllers/csi/metadata/correctness.go
@@ -64,7 +64,7 @@ func (checker *CorrectnessChecker) migrateAppMounts(ctx context.Context) {
 		}
 	}
 
-	err = os.MkdirAll(checker.path.AppMountsBaseDir(), os.ModePerm)
+	err = os.MkdirAll(checker.path.AppMountsBaseDir(), 0755)
 	if err != nil {
 		log.Error(err, "failed to create app mounts base directory")
 	}
--- a/pkg/controllers/csi/metadata/correctness.go
+++ b/pkg/controllers/csi/metadata/correctness.go
@@ -64,10 +64,12 @@ func (checker *CorrectnessChecker) migrateAppMounts(ctx context.Context) {
 		}
 	}
 
-	err = os.MkdirAll(checker.path.AppMountsBaseDir(), 0755)
+	err = os.MkdirAll(checker.path.AppMountsBaseDir(), dtcsi.AppmountsDirPermissions)
 	if err != nil {
 		log.Error(err, "failed to create app mounts base directory")
 	}
+
+	_ = os.Chmod(checker.path.AppMountsBaseDir(), dtcsi.AppmountsDirPermissions)
 
 	for _, appMount := range oldAppMounts {
 		oldPath := filepath.Dir(appMount.Path)
--- a/pkg/controllers/csi/metadata/correctness.go
+++ b/pkg/controllers/csi/metadata/correctness.go
@@ -67,9 +67,12 @@ func (checker *CorrectnessChecker) migrateAppMounts(ctx context.Context) {
 	err = os.MkdirAll(checker.path.AppMountsBaseDir(), dtcsi.AppmountsDirPermissions)
 	if err != nil {
 		log.Error(err, "failed to create app mounts base directory")
+	} else {
+		err = os.Chmod(checker.path.AppMountsBaseDir(), dtcsi.AppmountsDirPermissions)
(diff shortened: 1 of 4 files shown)


pkg/controllers/csi/metadata/correctness.go
```diff
@@ -64,7 +64,7 @@ func (checker *CorrectnessChecker) migrateAppMounts(ctx context.Context) {
 		}
 	}
 
-	err = os.MkdirAll(checker.path.AppMountsBaseDir(), os.ModePerm)
+	err = os.MkdirAll(checker.path.AppMountsBaseDir(), 0755)
```
waodim: Can we introduce a constant for it?

pkg/controllers/csi/server/server.go
```diff
@@ -95,6 +95,10 @@ func (srv *Server) Start(ctx context.Context) error {
 		hostvolumes.Mode: hostvolumes.NewPublisher(srv.mounter, srv.path),
 	}
 
+	if err := os.MkdirAll(srv.path.AppMountsBaseDir(), 0755); err != nil {
+		return errors.WithMessagef(err, "failed to create '%s' directory", srv.path.AppMountsBaseDir())
+	}
+
```
waodim, line 101: ~Would it be possible to run chmod here? https://pkg.go.dev/os#Chmod Or are there some risks?~

~Such that no reinstall is needed?~

edit: sorry this probably does not make sense, why would we need to change it if we set the mode anyways

----

edit-v2: Ok, I remember why I asked this myself.

What if the AppMountsBaseDir already exists from an older version?

From Go docs: `If path is already a directory, MkdirAll does nothing and returns nil.`

So to fix this case maybe `os.Chmod` would make sense?

aorcholski: chmod added

0sewa0 (dismissed):

pkg/controllers/csi/metadata/correctness.go
```diff
@@ -64,11 +64,13 @@ func (checker *CorrectnessChecker) migrateAppMounts(ctx context.Context) {
...
 	}
 
-	err = os.MkdirAll(checker.path.AppMountsBaseDir(), os.ModePerm)
+	err = os.MkdirAll(checker.path.AppMountsBaseDir(), dtcsi.AppmountsDirPermissions)
 	if err != nil {
 		log.Error(err, "failed to create app mounts base directory")
 	}
 
+	_ = os.Chmod(checker.path.AppMountsBaseDir(), dtcsi.AppmountsDirPermissions)
+
```
avorima: Should call this in the else case to prevent unnecessary calls. Also the error should be logged
```suggestion
} else {
	_ = os.Chmod(checker.path.AppMountsBaseDir(), dtcsi.AppmountsDirPermissions)
}
```

aorcholski: done

avorima (approved):

pkg/controllers/csi/metadata/correctness.go
0sewa0: wasn't the idea that we want to always have this new more restricted permissions?

> // If path is already a directory, MkdirAll does nothing
// and returns nil.

So doing it in `else` will do nothing? 

Also, this is in the `correctness.go` which is for "migration" related logic, and only runs once on startup.

pkg/controllers/csi/metadata/correctness.go
aorcholski: done

PR #6557, opened 2026-05-07
Create consts for the DTComponentsSecretsRootDir used in many components

gkrenn:
## Description

Follow up PR for https://dt-rnd.atlassian.net/browse/ICP-3804. 

Moves the widely used `/var/lib/dynatrace/secrets` path into one place.

## How can this be tested?

- run e2e, everything should work as before

### DIFF
--- a/pkg/controllers/dynakube/activegate/consts/consts.go
+++ b/pkg/controllers/dynakube/activegate/consts/consts.go
@@ -2,7 +2,7 @@ package consts
 
 import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api"
-	sharedconsts "github.com/Dynatrace/dynatrace-operator/pkg/consts"
+	"github.com/Dynatrace/dynatrace-operator/pkg/consts"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/connectioninfo"
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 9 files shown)


pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/certs.go
```diff
@@ -62,7 +62,7 @@ func (mod CertificatesModifier) getVolumeMounts() []corev1.VolumeMount {
 		{
 			ReadOnly:  true,
 			Name:      jettyCerts,
-			MountPath: filepath.Join(secretsRootDir, "tls"),
+			MountPath: filepath.Join(sharedconsts.DTComponentsSecretsRootDir, "/tls"),
```
avorima: the `/` is not needed

pkg/controllers/dynakube/activegate/consts/consts.go
```diff
@@ -2,6 +2,7 @@ package consts
 
 import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api"
+	sharedconsts "github.com/Dynatrace/dynatrace-operator/pkg/consts"
```
waodim:
```suggestion
	operatorconsts "github.com/Dynatrace/dynatrace-operator/pkg/consts"
```

since we use this alias also for different things (I think for TLS secrets or so)

WDYT?

avorima: i would argue that the alias isn't even needed

gkrenn: I changed it where it was needed

pkg/controllers/dynakube/activegate/internal/statefulset/builder/modifiers/certs.go
gkrenn: done

StefanHauth (approved):
LGTM

PR #6561, opened 2026-05-08
Refactor metadata attributes handling in injection webhook

StefanHauth:
## Description

This PR is a refactoring of the metadata attribute handling in the injection webhook. It's goal is to have all collection of attributes from various sources and the combination of the back into a single list in one implementation that can then be used by the webhook mutators. The rational is, that precedence handling is implemented in one place instead of having it spread all over the place. It shall be easier to understand and be the same for all components that use the attributes.
The same goes for setting metadata annotations to application pods. 

This refactoring is the pre-requisite for [JIRA](https://dt-rnd.atlassian.net/browse/ICP-3793).

## How can this be tested?

* unit tests
* E2E tests

### Hands On
* deploy a DK with metadata enrichment, OA injection or OTLP exporter auto-config
* create a namespace and add some `metadata.dynatrace.com/...` annotations
* deploy some app pods with `metadata.dynatrace.com/...` annotations
* check the command line of the init container or `OTLP_RESOURCE_ATTRIBUTES`, it shall show the following keys

```
k8s.pod.name
k8s.pod.uid
k8s.node.name
k8s.namespace.name

k8s.cluster.uid
k8s.cluster.name
dt.entity.kubernetes_cluster

k8s.workload.kind
k8s.workload.name

If deprecated attributes are enabled (feature.dynatrace.com/enable-attributes-dt.kubernetes: true, its default)
dt.kubernetes.workload.kind
dt.kubernetes.workload.name
dt.kubernetes.cluster.id
```
The attributes set on your namespace and pod should also show up, pod overrides namespace.

The pod should have all annotations from the namespace as well (pod again wins).

Bonus:
Play with enrichment rules:
```
  metadataEnrichment:
    rules:
    - source: label-rule2
      type: LABEL
    - source: label-rule1
      target: dt.cost.product
      type: LABEL
    - source: rule2
      type: ANNOTATION
    - source: rule1
      target: dt.security_context
      type: ANNOTATION
 ```

Here are manifests for a namespace and an nginx deployment with some labels and annotations as inspiration:

[nginx-glibc-complete.yaml](https://github.com/user-attachments/files/27537155/nginx-glibc-complete.yaml)
[namespace.yaml](https://github.com/user-attachments/files/27537156/namespace.yaml)

A good test is also to deploy the setup with current main and with the changed code and compare the resulting pod manifests. They should be the same wrt. annotations, bootstrapper arguments and OTLP resource attributes.

### CODE REVIEW

File: pkg/webhook/mutation/pod/handler/injection/handle.go [hunk 1]
```diff
@@ -135,7 +135,9 @@ func (h *Handler) handlePodMutation(mutationRequest *dtwebhook.MutationRequest)
 		mutated = true
 	}
 
-	if h.metaMutator.IsEnabled(mutationRequest.Context, mutationRequest.BaseRequest) {
+	// metaMutator is automatically enabled if oaMutator is enabled, but it can also be enabled standalone
+	if h.metaMutator.IsEnabled(mutationRequest.Context, mutationRequest.BaseRequest) ||
+		h.oaMutator.IsEnabled(mutationRequest.Context, mutationRequest.BaseRequest) {
 		err := h.metaMutator.Mutate(mutationRequest)
 		if err != nil {
```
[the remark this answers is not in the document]
StefanHauth: This is intenionally. Metadata enrichment shall be automatically enabled when OA injection is enabled.

pkg/webhook/mutation/pod/attributes/annotations.go
```diff
@@ -0,0 +1,59 @@
...
+func (attrs *PodAttributes) readPodAnnotationAttributes(pod corev1.Pod) {
+	// pod annotations take precedence over namespace annotations
+	for key, value := range pod.Annotations {
+		if after, ok := strings.CutPrefix(key, metadataenrichment.Prefix); ok {
+			attrs.podAnnotations[after] = value
+		}
+	}
+}
+
+func (attrs *PodAttributes) readFromEnrichmentRules(namespace corev1.Namespace, dk dynakube.DynaKube) {
```
0sewa0: naming nitpick:

This is more a `executeEnrichmentRules` / `userEnrichmentRules` / `evalEnrichmentRules`.
Because it "gets attributes based on the enrichment rules"

pkg/webhook/mutation/pod/attributes/containerinfo.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+)
+
+type ContainerInfos struct {
```
0sewa0:
```suggestion
type ContainerInfo struct {
```

pkg/webhook/mutation/pod/attributes/containerinfo.go
```diff
@@ -0,0 +1,57 @@
...
+	infos.Repository, tag, found = strings.Cut(repo, ":")
+	if found {
+		infos.Tag = tag
+	}
+
+	return infos
+}
+
+// Converts the whole object to a single JSON string used by the bootstrapper
+func (attrs *ContainerInfos) ToJSON() (string, error) {
```
0sewa0:
```suggestion
func (c *ContainerInfos) ToJSON() (string, error) {
```

pkg/webhook/mutation/pod/attributes/container.go
```diff
@@ -0,0 +1,22 @@
+package attributes
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+type ContainerAttributes struct {
```
0sewa0: I am a "many smaller files" fan, but IMO this could be in the same file as `pkg/webhook/mutation/pod/attributes/containerinfo.go`

pkg/webhook/mutation/pod/attributes/convert.go
```diff
@@ -0,0 +1,14 @@
+package attributes
```
0sewa0: why separate file?

pkg/webhook/mutation/pod/attributes/attributes.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// make sure we use the same precedence as in combine()
+	maps.Copy(combined, attrs.workloadInfo)
+	maps.Copy(combined, attrs.namespaceAnnotations)
+	maps.Copy(combined, attrs.rulesPropagate)
+
+	return combined
+}
+
+func (attrs *PodAttributes) combineForJSONAnnotation() map[string]string {
```
0sewa0: could be `ToJSON() ([]byte, error)`

so you would simplify the `setPodMetadataJSONAnnotation` method

pkg/webhook/mutation/pod/attributes/podannotations.go
```diff
@@ -0,0 +1,61 @@
...
+func (attrs *PodAttributes) setWorkloadAnnotations(pod *corev1.Pod) {
+	if pod.Annotations == nil {
+		pod.Annotations = make(map[string]string)
+	}
+
+	pod.Annotations[metadataenrichment.Prefix+K8sWorkloadNameAttr] = attrs.workloadInfo[K8sWorkloadNameAttr]
+	pod.Annotations[metadataenrichment.Prefix+K8sWorkloadKindAttr] = attrs.workloadInfo[K8sWorkloadKindAttr]
+}
+
+func setPodAnnotationIfNotExists(pod *corev1.Pod, key, value string) {
```
0sewa0: could be a `k8spod` helper, `k8spod.SetAnnotationIfNotExists`

pkg/webhook/mutation/pod/attributes/pod.go
```diff
@@ -0,0 +1,30 @@
...
+	attrs.podInfo[K8sPodUIDAttr] = createEnvVarRef(K8sPodUIDEnv)
+	attrs.podInfo[K8sNodeNameAttr] = createEnvVarRef(K8sNodeNameEnv)
+	attrs.podInfo[K8sNamespaceNameAttr] = request.Pod.Namespace
+
+	attrs.clusterInfo[K8sClusterUIDAttr] = request.DynaKube.Status.KubeSystemUUID
+	attrs.clusterInfo[K8sClusterNameAttr] = request.DynaKube.Status.KubernetesClusterName
+	attrs.clusterInfo[K8sDTClusterEntityAttr] = request.DynaKube.Status.KubernetesClusterMEID
+}
+
+func createEnvVarRef(envName string) string {
```
0sewa0: could be a `k8senv` helper: `func NewRef(envName string) string`

0sewa0 (dismissed):
I like to have multiple files over few gigantic ones, but here I feel it is very messy 😅 

everything is spread around, making it hard to review / understand.

I would combine files based on more general things, like `read` / `apply` and types

So one possible structure could be:
- `../attributes/pod.go`: `PodAttributes` type + exported methods + constructor
- `../attributes/container.go`: `ContainerInfo` / `ContainerAttributes` type + exported methods + constructor
- `../attributes/deprecated.go`: (I like a separate deprecated thing)
- `../attributes/read.go`: Anything that has to do with the reading attributes
- `../attributes/apply.go`: Everything that has to do with the applying attributes

The more "go way" would be:
- `../attributes/pod.go`:  (EVERYTHING) `PodAttributes`
- `../attributes/container.go`: (EVERYTHING) `ContainerInfo` / `ContainerAttributes`
- `../attributes/deprecated.go`: (I like a separate deprecated thing)

EDIT: Tried to make it clear that I meant separate files and not packages

(note: maybe naming also could be optimized so we don't stutter, but that is more of a nitpick, due to the possible confusion it may cause)

avorima:
> I like to have multiple files over few gigantic ones, but here I feel it is very messy 😅
> 
> everything is spread around, making it hard to review / understand.
> 
> I would combine files based on more general things, like `read` / `apply` and types
> 
> So one possible structure could be:
> 
>     * `../attributes/pod`: `PodAttributes` type + exported methods + constructor
> 
>     * `../attributes/container`: `ContainerInfo` / `ContainerAttributes` type + exported methods + constructor
> 
>     * `../attributes/deprecated`: (I like a separate deprecated thing)
> 
>     * `../attributes/read`: Anything that has to do with the reading attributes
> 
>     * `../attributes/apply`: Everything that has to do with the applying attributes
> 
> 
> The more "go way" would be:
> 
>     * `../attributes/pod`:  (EVERYTHING) `PodAttributes`
> 
>     * `../attributes/container`: (EVERYTHING) `ContainerInfo` / `ContainerAttributes`
> 
>     * `../attributes/deprecated`: (I like a separate deprecated thing)
> 
> 
> (note: maybe naming also could be optimized so we don't stutter, but that is more of a nitpick, due to the possible confusion it may cause)

`pod` and `container` are bad package names since they're also very nice variable names and we already moved away from them while refactoring the kubernetes utils. I think the package makes sense. There are pod attributes (`attributes.Pod`) and container attributes (`attributes.Container`), each with a small API to aggregate and collect information. I agree that some files could be merged, but splitting, especially to the degree of your first variant, will just increase API surface, complicate naming and will break as soon as we need some shared code between them.

pkg/webhook/mutation/pod/attributes/annotations.go
StefanHauth: I went for `applyEnrichmentRules` to better express that the `attrs` object is altered.

pkg/webhook/mutation/pod/attributes/attributes.go
StefanHauth: I want to keep precedence handling separate from other things to have it stick out and clearly documented. And I want it to be clear that those three funcs somwhat belong together and have a common theme (btw, I'm a bit unhappy that we need three funcs here, would prefer to do it in one).

0sewa0:
@avorima I never wanted separate packages 🙈 
just fever separate files, (should have written `../attributes/pod.go`, but I missed it, I edited my original comment now)

StefanHauth:
> @avorima I never wanted separate packages 🙈 just fever separate files, (should have written `../attributes/pod.go`, but I missed it, I edited my original comment now)

Thanks for clarifying, that confused me a bit as well :-)

pkg/webhook/mutation/pod/attributes/annotations.go
0sewa0: Nice, didn't think of that, but I also like it the best 👍

pkg/webhook/mutation/pod/attributes/attributes.go
0sewa0: that is ok, but splitting the creation of the `map` for the JSON annotation (already ultra specific logic), and actually creation of the JSON `[]byte` feels odd to me.

Doing the `json.Marshal(annotations)`in this func would be good, and do the `setPodAnnotationIfNotExists` in a different func

And if we do that, then `setPodMetadataJSONAnnotation` would be kinda unnecessary, and could just have it in `ApplyAnnotationsToPod`, like

```go
func (attrs *PodAttributes) ApplyAnnotationsToPod(pod *corev1.Pod) error {
	annotations := attrs.combineForMetadataAnnotations()

	for key, value := range annotations {
		setPodAnnotationIfNotExists(pod, metadataenrichment.Prefix+key, value)
	}

	// set workload annotations no matter what
	attrs.setWorkloadAnnotations(pod)
	
	jsonBytes, err := attrs.ToJSON()
    if err != nil {
		return err
	}

     // attrs.ToJSON() could also just return `string` already
	setPodAnnotationIfNotExists(pod, metadataenrichment.Prefix, string(jsonBytes))

	return attrs.setPodMetadataJSONAnnotation(pod)
}
```

This `ApplyAnnotationsToPod` is still simple and easy to read. You separated the precedence handling into `ToJSON` and the applying-logic to the `ApplyAnnotationsToPod`

sidenote: I am not married to `ToJSON` as a name, it can still be called `combineForJSONAnnotation` if we want to be verbose

pkg/webhook/mutation/pod/attributes/annotations.go
StefanHauth: [d142b79](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/d142b79f6714b6953d9459b3e1007c8d09ca06cb)

pkg/webhook/mutation/pod/attributes/containerinfo.go
StefanHauth: [d142b79](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/d142b79f6714b6953d9459b3e1007c8d09ca06cb)

pkg/webhook/mutation/pod/attributes/containerinfo.go
StefanHauth: [d142b79](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/d142b79f6714b6953d9459b3e1007c8d09ca06cb)

pkg/webhook/mutation/pod/attributes/container.go
StefanHauth: [d142b79](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/d142b79f6714b6953d9459b3e1007c8d09ca06cb)

pkg/webhook/mutation/pod/attributes/convert.go
StefanHauth: [d142b79](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/d142b79f6714b6953d9459b3e1007c8d09ca06cb)

pkg/webhook/mutation/pod/attributes/podannotations.go
StefanHauth: [d142b79](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/d142b79f6714b6953d9459b3e1007c8d09ca06cb)

pkg/webhook/mutation/pod/attributes/pod.go
StefanHauth: [d142b79](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/d142b79f6714b6953d9459b3e1007c8d09ca06cb)

pkg/webhook/mutation/pod/attributes/attributes.go
StefanHauth: I get your point, and I agree. The only thing I don't like is to name that combining function simply `ToJSON` because that doesn't convey properly what that function is mainly about (the immediate JSON conversion is more the cherry on top).

I've reworked it in [6c00c9c](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/6c00c9cfa0daa0c05eecbb371bc8ff0fa259508c)

StefanHauth:
Reworked the package structure following @0sewa0 's suggestion. Instead of `apply.go` I chose `annotations.go` because annotations are the only thing touched by the funcs in that file.

pkg/util/kubernetes/objects/k8spod/pod.go
```diff
@@ -29,3 +29,13 @@ func GetName(pod corev1.Pod) string {
 
 	return pod.GenerateName
 }
+
+func SetPodAnnotationIfNotExists(pod *corev1.Pod, key, value string) {
```
0sewa0:
```suggestion
func SetAnnotationIfNotExists(pod *corev1.Pod, key, value string) {
```
to avoid stuttering

pkg/webhook/mutation/pod/attributes/read.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return nil
+}
+
+func (attrs *PodAttributes) readPodAttributes(request mutator.BaseRequest) {
+	attrs.podEnvVars = append(attrs.podEnvVars, []corev1.EnvVar{
+		{Name: K8sPodNameEnv, ValueFrom: k8senv.NewSourceForField("metadata.name")},
+		{Name: K8sPodUIDEnv, ValueFrom: k8senv.NewSourceForField("metadata.uid")},
+		{Name: K8sNodeNameEnv, ValueFrom: k8senv.NewSourceForField("spec.nodeName")},
+	}...)
```
0sewa0: nitpick: the "pack it into a list just to instantly unpack it to skip on type declaration"  feels like a ugly hack to me 😅 
```suggestion
	attrs.podEnvVars = append(attrs.podEnvVars,
		corev1.EnvVar{Name: K8sPodNameEnv, ValueFrom: k8senv.NewSourceForField("metadata.name")},
		corev1.EnvVar{Name: K8sPodUIDEnv, ValueFrom: k8senv.NewSourceForField("metadata.uid")},
		corev1.EnvVar{Name: K8sNodeNameEnv, ValueFrom: k8senv.NewSourceForField("spec.nodeName")},
	)
```

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator.go
```diff
@@ -54,22 +56,25 @@ func (m *Mutator) Reinvoke(ctx context.Context, request *dtwebhook.ReinvocationR
 
 func (m *Mutator) mutate(ctx context.Context, request *dtwebhook.BaseRequest) (bool, error) {
 	log := logd.FromContext(ctx)
-	mutated := false
 
-	log.Debug("injecting OTLP resource Attributes")
+	log.Debug("injecting OTLP resource PodAttributes")
```
0sewa0: was this intentional? you also inject container attributes :D

0sewa0 (commented):
Code LGTM, just some small comments

will now try it live

EDIT: Done, works as expected 👍

pkg/webhook/mutation/pod/attributes/container.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+)
+
+type ContainerAttributes struct {
```
avorima: Are we set on these type names? `attributes.Container` and `attributes.Pod` are less verbose

StefanHauth: [030e678](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/030e67882bcb7b0f3e7d645807806fc39074930a)

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator.go
StefanHauth: [030e678](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/030e67882bcb7b0f3e7d645807806fc39074930a)

pkg/webhook/mutation/pod/attributes/read.go
StefanHauth: [030e678](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/030e67882bcb7b0f3e7d645807806fc39074930a)

pkg/util/kubernetes/objects/k8spod/pod.go
StefanHauth: [030e678](https://github.com/Dynatrace/dynatrace-operator/pull/6561/commits/030e67882bcb7b0f3e7d645807806fc39074930a)

0sewa0 (approved):

PR #6562, opened 2026-05-08
Filter out invalid symlinks that attempt to escape the extraction directory

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/ICP-4550)

## Description

Adds dedicated IsSafe* functions for symlink, hardlink and the target file.  

The original IsSafeToLink function concatenates _target_ (full path including filename part of the symlink) with the _link_ using Join so the result contains also filename part of the symlink in the middle increasing the depth of the path by one.

Nothing should be saved/loaded out of the _/data_ directory
```
/data/opt/symlink -> ../../outside.txt  
```
Without the fix:
```
/data/opt/symlink/../../outside.txt 
/data/opt/../outside.txt 
/data/outside.txt
```
the (correct) evaluated path returned by _EvalSymLink()_ would be
```
/outside.txt  <-- should be blocked
```

## How can this be tested?

unittests

### DIFF
--- a/pkg/injection/codemodule/installer/zip/gzip.go
+++ b/pkg/injection/codemodule/installer/zip/gzip.go
@@ -55,20 +55,10 @@ func extractFilesFromGzip(ctx context.Context, targetDir string, reader *tar.Rea
 			return errors.WithStack(err)
 		}
 
-		target := filepath.Join(targetDir, header.Name)
+		target := evaluateTargetPath(targetDir, header.Name)
 
-		// Check for ZipSlip: https://snyk.io/research/zip-slip-vulnerability
-		if !strings.HasPrefix(target, targetDir) {
-			return errors.Errorf("illegal file path: %s", target)
-		}
-
-		rel, err := filepath.Rel(targetDir, target)
-		if err != nil {
+		if err := isTargetSafeToCreate(targetDir, header.Name, target); err != nil {
 			return err
-		}
-
-		if strings.HasPrefix(rel, ".."+string(os.PathSeparator)) {
-			return errors.Errorf("%q is outside of %q", header.Name, targetDir)
 		}
 
 		err = extract(ctx, targetDir, reader, header, target)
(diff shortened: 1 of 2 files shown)


pkg/injection/codemodule/installer/zip/gzip.go
StefanHauth, line 186: The ticket suggests this fix as well, did you omit this intentionally?
```suggestion
	if fi, err := os.Lstat(target); err == nil && fi.Mode()&os.ModeSymlink != 0 {
		return errors.Errorf("refusing to write through symlink: %s", target)
	}

	mode := header.FileInfo().Mode()
```

StefanHauth, line 128: Should this also be hardened? Is there a chance of escape as well?
Here's what Klaus had to say:
```
isSafeToLink is called identically for both TypeSymlink (extractSymlink:125) and TypeLink (extractLink:111), but
  symlinks and hard links have different resolution semantics:

  - Symlinks: resolved relative to the symlink's containing directory → filepath.Dir(target) is now correct.
  - Hard links in tar: header.Linkname is an archive-root-relative path, not a directory-relative path. The actual link
  creation at line 112 uses filepath.Join(targetDir, header.Linkname), but isSafeToLink now validates against
  filepath.Join(filepath.Dir(target), header.Linkname) — these are different paths.

  Before this fix, both callers were equally wrong. Now they are asymmetrically wrong — symlinks are correctly validated
   while hard links can be under-blocked or over-blocked depending on the archive layout. The correct check for hard
  links would be:

  finalPath := filepath.Join(targetDir, hardlink)

  This is likely low-impact in practice (agent archives probably don't use hard links), but the function's contract and
  documentation should be corrected, or the function should be split into isSafeSymlink / isSafeHardlink.
```

StefanHauth: This function could benefit of a a dedicated unit tests, not just thru full-blown archive extraction.

StefanHauth (changes requested):

pkg/injection/codemodule/installer/zip/gzip.go
```diff
@@ -55,29 +55,45 @@
 			return errors.WithStack(err)
 		}
 
-		target := filepath.Join(targetDir, header.Name)
+		target := evaluateTargetPath(targetDir, header.Name)
```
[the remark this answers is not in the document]
aorcholski: target path is verified in the following line. The logic has been moved to isTargetSafeToCreate function. It allows to create simpler unittests.

pkg/injection/codemodule/installer/zip/gzip.go
aorcholski: done

aorcholski: fixed

aorcholski: I'm not sure if the image can't use a symlink to create other files (symlink to a directory)

pkg/injection/codemodule/installer/zip/gzip.go
```diff
@@ -119,10 +135,26 @@ func extractLink(ctx context.Context, targetDir, target string, header *tar.Head
 	return nil
 }
 
+// isHardlinkSafeToLink checks that the provided relative hardlink is NOT pointing outside of the `targetDir`
+func isHardlinkSafeToLink(link, targetDir string) bool {
```
waodim, line 139: Would it make sense to refactor this a bit?

`isSymlinkSafeToLink` takes 3 args and this one 2. The only real difference is the base dir that is used to resolve the link name. We could unify this to a function:

`isLinkTargetSafe(linkname , base, targetDir string) bool { ... }`

and call it with `base = targetDir` for hard links and `base = filepath.Dir(target)` for symlinks.

WDYT? Does that make sense?

aorcholski: part of the logic would be "moved" out of the isSafe function. The base directory is the core part of the logic so I would like to keep it inside the functions. It's just a few lines more.

avorima (approved):
LGTM. did some manual testing with the archives and the link is no longer extracted

waodim (approved):

StefanHauth (approved):

PR #6563, opened 2026-05-08
increase prconcurrentlimit=20 for renovate

andriisoldatenko:
## Description

Context:

Renovate didn't upgrade go version so we do -> [https://github.com/Dynatrace/dynatrace-operator/pull/6560](https://github.com/Dynatrace/dynatrace-operator/pull/6560)

To avoid this problem we should increase renovate prconcurrentlimit = 20 **(default is 10)**

Full list of pending PRs https://developer.mend.io/github/Dynatrace/dynatrace-operator

Example that renovate will update go:

image

## How can this be tested?

n/a

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -35,6 +35,7 @@
   automergeType: "pr",
   automerge: true,
   branchNameStrict: true,
+  prconcurrentlimit: 20,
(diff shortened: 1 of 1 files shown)


StefanHauth (dismissed):
Sounds reasonable, and we can always go back if we run into troubles.

StefanHauth (approved):

PR #6602, opened 2026-05-11
Add data ingest token hash to OTelC manifest to trigger restarts on data ingest token rotation

StefanHauth:
## Description

OTelC pod needs to be restarted, if data ingest token is rotated.

[JIRA](https://dt-rnd.atlassian.net/browse/ICP-4734)

## How can this be tested?

* deploy DK with telemetry ingest
* change data ingest token
* wait for next recon loop
* check that OTelC pod restarts and token hash has been updated

### DIFF
--- a/pkg/controllers/dynakube/otelc/statefulset/reconciler.go
+++ b/pkg/controllers/dynakube/otelc/statefulset/reconciler.go
@@ -31,6 +31,7 @@ const (
 	serviceAccountName                                  = "dynatrace" + consts.OTELCollectorNameSuffix
 	annotationTelemetryIngestSecretHash                 = api.InternalFlagPrefix + "telemetry-ingest-secret-hash"
 	annotationTelemetryIngestConfigurationConfigMapHash = api.InternalFlagPrefix + "telemetry-ingest-config-hash"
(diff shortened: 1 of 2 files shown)


pkg/controllers/dynakube/otelc/statefulset/reconciler.go
```diff
@@ -213,6 +221,17 @@ func (r *Reconciler) calculateConfigMapHash(ctx context.Context, configMapName s
...
 
+func (r *Reconciler) calculateDataIngestTokenHash(ctx context.Context, dk *dynakube.DynaKube) (string, error) {
+	tokenReader := token.NewReader(r.apiReader, dk)
+
+	tokens, err := tokenReader.ReadTokens(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	return hasher.GenerateHash(tokens.DataIngestToken().Value)
```
0sewa0:
```suggestion
	return hasher.GenerateSecureHash(tokens.DataIngestToken().Value)
```

It is a token, so we should try to be "as secure as possible"

0sewa0 (changes requested):

0sewa0 (approved):

PR #6678, opened 2026-05-13
Handle new settings field

avorima:
## Description

Handle the new enrichment rule schema in the settings client. It needs to handle two different types of environment: Managed and SaaS. Managed will keep using the `builtin:kubernetes.generic.metadata.enrichment` and SaaS will eventually move towards `builtin:ingest.enrichment.config`.
We have no practical way of knowing which environment the operator runs in and what state the environment has, so we need to infer the schema according to the API responses.

There are 2 cases when the new schema should be used
* Old schema was disabled
* New schema is opted in

Add a new error case: if the new schema is opted in, but it's not present on the tenant the injection reconciliation fails and the pods are not injected. This should never happen.

This PR adds the constants for the downstream enrichment rule handling, but that will be implemented in another change.

ICP-3071
ICP-4353

## How can this be tested?

The tenant setup is a bit involved. Please refer to [this comment](https://dt-rnd.atlassian.net/browse/ICP-1164?focusedCommentId=4602926) on the epic.

To verify the API calls are simply enable metadata enrichment

```yaml
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
spec:
  metadataEnrichment:
    enabled: true
    namespaceSelector:
      matchLabels:
        kubernetes.io/metadata.name: default
```

Procedure:
* Modify tenant
* Deploy DK
* Check logs for API calls
* Check status

Scenarios to test:
* No schema enabled
* Schema enabled, but flag disabled
* Schema enabled, flag enabled, no new rules
* Schema enabled, flag enabled, with rules
* Only new schema enabled

### DIFF
--- a/pkg/api/latest/dynakube/metadataenrichment/spec.go
+++ b/pkg/api/latest/dynakube/metadataenrichment/spec.go
@@ -10,8 +10,8 @@ const (
 	LabelRule      RuleType = "LABEL"
 	AnnotationRule RuleType = "ANNOTATION"
 
-	K8sLabelRule      RuleType = "K8S_NAMESPACE_LABEL"
-	K8sAnnotationRule RuleType = "K8S_NAMESPACE_ANNOTATION"
+	K8sNamespaceLabelRule      RuleType = "K8S_NAMESPACE_LABEL"
+	K8sNamespaceAnnotationRule RuleType = "K8S_NAMESPACE_ANNOTATION"
 	// TODO: implement support for this type.
 	CustomRule RuleType = "CUSTOM"
 
@@ -48,8 +48,8 @@ func IsSupportedType(ruleType RuleType) bool {
 	switch ruleType {
 	case LabelRule,
 		AnnotationRule,
-		K8sLabelRule,
-		K8sAnnotationRule:
+		K8sNamespaceLabelRule,
+		K8sNamespaceAnnotationRule:
 		return true
 	}
 
--- a/pkg/api/latest/dynakube/metadataenrichment/spec.go
+++ b/pkg/api/latest/dynakube/metadataenrichment/spec.go
@@ -7,11 +7,16 @@ import (
 type RuleType string
 
 const (
-	LabelRule          RuleType = "LABEL"
-	AnnotationRule     RuleType = "ANNOTATION"
-	Annotation         string   = "metadata.dynatrace.com"
-	Prefix                      = Annotation + "/"
-	namespaceKeyPrefix string   = "k8s.namespace."
+	LabelRule      RuleType = "LABEL"
+	AnnotationRule RuleType = "ANNOTATION"
+
+	K8sLabelRule      RuleType = "K8S_NAMESPACE_LABEL"
+	K8sAnnotationRule RuleType = "K8S_NAMESPACE_ANNOTATION"
+	CustomRule        RuleType = "CUSTOM"
+
+	Annotation         = "metadata.dynatrace.com"
+	Prefix             = Annotation + "/"
+	namespaceKeyPrefix = "k8s.namespace."
 )
 
 type MetadataEnrichment struct {
--- a/pkg/api/latest/dynakube/metadataenrichment/spec.go
+++ b/pkg/api/latest/dynakube/metadataenrichment/spec.go
@@ -12,7 +12,8 @@ const (
 
 	K8sLabelRule      RuleType = "K8S_NAMESPACE_LABEL"
 	K8sAnnotationRule RuleType = "K8S_NAMESPACE_ANNOTATION"
-	CustomRule        RuleType = "CUSTOM"
+	// TODO: implement support for this type.
+	CustomRule RuleType = "CUSTOM"
 
 	Annotation         = "metadata.dynatrace.com"
 	Prefix             = Annotation + "/"
@@ -40,4 +41,17 @@ type Rule struct {
 	Type   RuleType `json:"type,omitempty"`
 	Source string   `json:"source,omitempty"`
 	Target string   `json:"target,omitempty"`
+}
+
+// IsSupportedType returns true if a rule's type should be used for further processing.
+func IsSupportedType(ruleType RuleType) bool {
+	switch ruleType {
+	case LabelRule,
+		AnnotationRule,
+		K8sLabelRule,
+		K8sAnnotationRule:
+		return true
+	}
+
+	return false
 }
--- a/pkg/api/latest/dynakube/metadataenrichment/spec.go
+++ b/pkg/api/latest/dynakube/metadataenrichment/spec.go
@@ -49,7 +49,8 @@ func IsSupportedType(ruleType RuleType) bool {
 	case LabelRule,
 		AnnotationRule,
 		K8sNamespaceLabelRule,
-		K8sNamespaceAnnotationRule:
+		K8sNamespaceAnnotationRule,
+		CustomRule:
 		return true
 	}
 
--- a/pkg/clients/dynatrace/settings/enchrichment.go
+++ b/pkg/clients/dynatrace/settings/enchrichment.go
@@ -10,15 +10,15 @@ import (
 )
 
 const (
-	metadataEnrichmentSettingsSchemaID = "builtin:kubernetes.generic.metadata.enrichment"
-	scopeQueryParam                    = "scope"
-	globalScope                        = "environment"
-	effectiveValuesPath                = "/v2/settings/effectiveValues"
+	legacyMetadataEnrichmentSchemaID = "builtin:kubernetes.generic.metadata.enrichment"
+	metadataEnrichmentSchemaID       = "builtin:ingest.enrichment.config"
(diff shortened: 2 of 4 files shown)


pkg/clients/dynatrace/settings/enchrichment.go
```diff
@@ -27,9 +27,27 @@ type ruleItem struct {
...
+
+	// These fields are embedded into the value instead of part of a rules list, when using the builtin:ingest.enrichment.config schema.
+	// Group them in a struct for easy emptiness check.
+	ingestEnrichmentConfig
+}
+
+type ingestEnrichmentConfig struct {
+	Type        string `json:"type"`
+	Target      string `json:"target"`
+	ValueSource string `json:"valueSource"`
```
0sewa0: I missed something in the requirements of https://dt-rnd.atlassian.net/browse/ICP-4353

there are 2 other fields, `condition` and `valueExtraction`. (not 100% about the naming accuracy)

we should only parse them so that we can throw rules away that have them set. as these fields are not meant to be supported by the operator just yet.

more context in [slack](https://dynatrace.slack.com/archives/C0AD8TXGG1J/p1780310277012049?thread_ts=1780309328.523699&cid=C0AD8TXGG1J)

(we can fix this in a separate PR)

avorima: That would technically make the `isEmpty` check inaccurate. I'd post-pone this until we actually use these fields. Our models are made to fit anyways

avorima: Oh, the thread says that rules should be ignored if these optional fields are are set. that's unfortunate

0sewa0: > make the isEmpty check inaccurate

IMO, not reaaally, because they are "not empty", they just "have unsupported fields" 🤷

pkg/api/latest/dynakube/metadataenrichment/spec.go
```diff
@@ -36,3 +42,16 @@ type Rule struct {
...
 }
+
+// IsSupportedType returns true if a rule's type should be used for further processing.
+func IsSupportedType(ruleType RuleType) bool {
+	switch ruleType {
+	case LabelRule,
+		AnnotationRule,
+		K8sLabelRule,
+		K8sAnnotationRule:
+		return true
```
StefanHauth, line 54: `CustomRule` is missing here. They should be supported as well (if they do not have condition and/or valueExtraction set).

StefanHauth (changes requested):

pkg/api/latest/dynakube/metadataenrichment/spec.go
```diff
@@ -7,11 +7,17 @@ import (
...
 const (
-	LabelRule          RuleType = "LABEL"
-	AnnotationRule     RuleType = "ANNOTATION"
-	Annotation         string   = "metadata.dynatrace.com"
-	Prefix                      = Annotation + "/"
-	namespaceKeyPrefix string   = "k8s.namespace."
+	LabelRule      RuleType = "LABEL"
+	AnnotationRule RuleType = "ANNOTATION"
+
+	K8sLabelRule      RuleType = "K8S_NAMESPACE_LABEL"
```
andriisoldatenko:
```suggestion
	K8sNSlabelRule      RuleType = "K8S_NAMESPACE_LABEL"
```
maybe? because it confusing together with `LabelRule`

pkg/api/latest/dynakube/metadataenrichment/spec.go
avorima: they're not supported yet

pkg/api/latest/dynakube/metadataenrichment/spec.go
avorima: true, there might be other k8s related rules in the future

pkg/clients/dynatrace/settings/enchrichment.go
```diff
@@ -44,32 +57,92 @@ func (c *ClientImpl) GetRules(ctx context.Context, kubeSystemUUID, entityID stri
...
-		WithQueryParams(map[string]string{
-			validateOnlyQueryParam: "true",
-			schemaIDsQueryParam:    metadataEnrichmentSettingsSchemaID,
-			scopeQueryParam:        scope,
-		}).
-		Execute(&resp)
-	if err != nil {
-		if core.IsNotFound(err) {
-			log.Info("enrichment settings not available on cluster, skipping getting the enrichment rules", "schemaID", metadataEnrichmentSettingsSchemaID)
+	if err := c.apiClient.GET(ctx, effectiveValuesPath).WithQueryParams(params).Execute(&resp); err != nil {
```
StefanHauth, line 71: What happens if the legacy schema exists, but carries no rules because the rules are all already defined in the new schema (so basically an empty response)? In that case this if would fall thru to line 84, right? And it would not even try the new schema.
Is there a contract, that in case of the new schema, at least one "dummy" rule holding the "use-new-schema" flag is present?

StefanHauth (changes requested):

pkg/clients/dynatrace/settings/enchrichment.go
avorima: There is no such fallback. I would say this falls under the same category as users that forget to migrate their rules to the new schema

pkg/clients/dynatrace/settings/enchrichment.go
```diff
@@ -44,32 +57,92 @@ func (c *ClientImpl) GetRules(ctx context.Context, kubeSystemUUID, entityID stri
...
+	// Retry the request with the new schema. For managed this will always fail, but we have no practical way of knowing which environment we're running in.
+	params[schemaIDsQueryParam] = metadataEnrichmentSchemaID
+	// Clear the input so that we don't keep stale data
+	resp = getRulesResponse{}
+
+	if err := c.apiClient.GET(ctx, effectiveValuesPath).WithQueryParams(params).Execute(&resp); err != nil {
+		if !useNewSchema && core.IsNotFound(err) {
+			// Keep the established behavior of not failing when the legacy schema is not available
+			// This covers the managed use-case.
+			log.Info("enrichment settings not available on cluster, skipping getting the enrichment rules", "schemaID", legacyMetadataEnrichmentSchemaID)
```
StefanHauth:
```suggestion
			log.Info("enrichment settings not available on cluster, skipping getting the enrichment rules", "schemaID", metadataEnrichmentSchemaID)
```

StefanHauth (changes requested):

pkg/clients/dynatrace/settings/enchrichment.go
StefanHauth: But what if users migrate rules to the new scheme leaving the legacy schema empty? Then the legacy schema would exist, but it has no rules with the flag to indicate to use the new schema, the new schema would be ignored although it properly holds the rules. Same goes for users who never used rules and start using them with the new schema.

pkg/clients/dynatrace/settings/enchrichment.go
avorima: no, that's actually the previous log message when the schema was not found

pkg/clients/dynatrace/settings/enchrichment.go
avorima: The old rules are not cleared. They will keep existing as long as the schema exists. To have rules in the new schema and no rules in the legacy schema you would have to either:
* manually delete the legacy rules
* only create rules in the new schema, but in classic environments this requires explicitly enabling feature flags

pkg/clients/dynatrace/settings/enchrichment.go
```diff
@@ -44,32 +57,92 @@ func (c *ClientImpl) GetRules(ctx context.Context, kubeSystemUUID, entityID stri
...
-			validateOnlyQueryParam: "true",
-			schemaIDsQueryParam:    metadataEnrichmentSettingsSchemaID,
-			scopeQueryParam:        scope,
-		}).
-		Execute(&resp)
-	if err != nil {
-		if core.IsNotFound(err) {
-			log.Info("enrichment settings not available on cluster, skipping getting the enrichment rules", "schemaID", metadataEnrichmentSettingsSchemaID)
+	if err := c.apiClient.GET(ctx, effectiveValuesPath).WithQueryParams(params).Execute(&resp); err != nil {
+		if !core.IsNotFound(err) {
```
StefanHauth, line 72:
```suggestion
		if core.IsNotFound(err) {
		   log....
		} else {
		   return ...
		}
```
This would avoid the "is not not found" double negotiation that always wrecks my brain.

pkg/clients/dynatrace/settings/enchrichment.go
StefanHauth: was clarified internally.

StefanHauth (changes requested):

StefanHauth (approved):

PR #6679, opened 2026-05-14
Introduce e2e test for use-public-registry + publicRegistryOverride features

jokuniew:
## Description
ICP-1105

E2E tests that:
deploy `use-public-registry` for each component individually and test with/without `publicRegistryOverride`

## How can this be tested?
To run tests locally, remember to provide `devregistry` secret before run, then:
```bash
# run sequentially all components
export E2E_PUBLIC_REGISTRY_OVERRIDE=some.registry
make test/e2e/usepublicregistry

# or individually
export E2E_PUBLIC_REGISTRY_OVERRIDE=some.registry
make test/e2e/usepublicregistry/oneagent 
make test/e2e/usepublicregistry/activegate
make test/e2e/usepublicregistry/codemodules
```

env variable set by export is defined in github actions variables


.github/actions/run-e2e/action.yaml
```diff
@@ -126,6 +126,7 @@ runs:
         TENANT2_NAME: ${{ inputs.tenant2-name }}
         TENANT2_APITOKEN: ${{ inputs.tenant2-apitoken }}
         TENANT2_DATAINGESTTOKEN: ${{ inputs.tenant2-dataingesttoken }}
+        E2E_PUBLIC_REGISTRY_OVERRIDE: ${{ vars.E2E_PUBLIC_REGISTRY_OVERRIDE }}
```
andriisoldatenko, line 129: I would not expose it here, since we don't use it inside workflow, plus our dev tenant has hardcoded default, and i'm not sure if we can add second registry there.

So just remote it here, and later we can reconsider.

test/e2e/features/usepublicregistry/deploy.go
```diff
@@ -0,0 +1,256 @@
...
+	builder.Assess("devregistry pull secret exists", requireDevRegistrySecret())
+
+	secretConfig := tenant.GetSingleTenantSecret(t)
+
+	options := []dynakubeComponents.Option{
+		dynakubeComponents.WithName(dkName),
+		dynakubeComponents.WithAPIURL(secretConfig.APIURL),
+		dynakubeComponents.WithCloudNativeSpec(cloudnative.DefaultCloudNativeSpec()),
+		dynakubeComponents.WithUsePublicRegistryFF(),
+		dynakubeComponents.WithCustomPullSecret(consts.DevRegistryPullSecretName),
```
andriisoldatenko, line 106: may be add a comment, that our dev tenant use dev ECR registry, so we need to add custompull secret.

I'm just not sure, may be in future, dev tenants will return public ecr as well /shrug.

Incredible-O:
Nit: The commit message (i.e. the title of this PR) could be a bit more descriptive 😄

test/e2e/features/usepublicregistry/deploy.go
jokuniew: comment added https://github.com/Dynatrace/dynatrace-operator/pull/6679/commits/c694d1a1470c99a045a24d0ccc69e3bf5ccd2ccf

.github/actions/run-e2e/action.yaml
jokuniew: as discussed offline, 
- we keep this env var 
- regroupped targets over publicRegistryOverride, so most of the cases local dev will run over use-public-registry and less oftenly `-with-public-registry-override` so the requirement to provide env will happen in one suite

test/e2e/features/usepublicregistry/deploy.go
```diff
@@ -0,0 +1,259 @@
...
+
+const (
+	publicRegistryOverrideEnvVar = "E2E_PUBLIC_REGISTRY_OVERRIDE"
+)
+
+func publicRegistryOverride(t *testing.T) string {
+	t.Helper()
+	val := os.Getenv(publicRegistryOverrideEnvVar)
+	if val == "" {
+		t.Fatalf("%s must be set", publicRegistryOverrideEnvVar)
```
avorima: this should skip, not fail

andriisoldatenko: yeah we can do it,

jokuniew: changed to skipf
https://github.com/Dynatrace/dynatrace-operator/pull/6679/changes/d64d02227233803ab906d50b9bc03585023f130c
image

jokuniew:
#6684 got merged, will simplify codemodules test case
update: removed `node-image-pull` from codemodules dk

hack/make/tests/e2e.mk
```diff
@@ -171,6 +171,46 @@ test/e2e/applicationmonitoring/bootstrapper-no-csi:
 test/e2e/publicregistry:
 	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/standard -run "public_registry_images" $(SKIPCLEANUP)
 
+## Runs use-public-registry e2e scenarios with feature flag enabled, without publicRegistryOverride
+## Includes OneAgent / ActiveGate / CodeModules (OA + AG on standard, CodeModules on nocsi)
+test/e2e/usepublicregistry:
```
avorima, line 180: I would expect this to run all public registry tests, not just the ones without override. Same for all "sub" targets

hack/make/tests/e2e.mk
```diff
@@ -171,6 +171,46 @@ test/e2e/applicationmonitoring/bootstrapper-no-csi:
...
+	RC=0; \
+	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/standard -run "use_public_registry_without_override" $(SKIPCLEANUP) || RC=1; \
+	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/nocsi -run "use_public_registry_without_override" $(SKIPCLEANUP) || RC=1; \
+	exit $$RC
+
+## Runs use-public-registry e2e scenarios with publicRegistryOverride enabled
+## Includes OneAgent / ActiveGate / CodeModules (OA + AG on standard, CodeModules on nocsi)
+test/e2e/usepublicregistry-with-public-registry-override:
```
avorima: The name could be shortened. Don't need to mention public registry twice

hack/make/tests/e2e.mk
jokuniew: that was the initial implementation 
but after discussion with @andriisoldatenko  there was feedback to group them over `without` `with` override
https://github.com/Dynatrace/dynatrace-operator/pull/6679#discussion_r3258481280
so dev will use mostly `without` override - so no need for env value 
and in special cases the `with override`

jokuniew: that was the initial implementation 
but after discussion with @andriisoldatenko  there was feedback to group them over `without` `with` override
https://github.com/Dynatrace/dynatrace-operator/pull/6679#discussion_r3258481280
so dev will use mostly `without` override - so no need for env value 
and in special cases the `with override`

andriisoldatenko: yeah, but before it was t.Fatal, as it now t.Skip i'm fine to revert as it was before.

hack/make/tests/e2e.mk
andriisoldatenko: usepublicregistry-w-override ?

hack/make/tests/e2e.mk
jokuniew: reverted the commit https://github.com/Dynatrace/dynatrace-operator/pull/6679/changes/bc11ff5e3ea4b2d2ded60bc6856d5f50ceddc259

so there is one target `test/e2e/usepublicregistry`
and subtargets per components

hack/make/tests/e2e.mk
jokuniew: no need for rename, grouped back targets to one `test/e2e/usepublicregistry`

avorima (dismissed):
Please document the E2E_PUBLIC_REGISTRY_OVERRIDE variable in CONTRIBUTING.md

andriisoldatenko (dismissed):

jokuniew:
> Please document the E2E_PUBLIC_REGISTRY_OVERRIDE variable in CONTRIBUTING.md

done

avorima (approved):

PR #6680, opened 2026-05-15
feat: add configurable liveness and readiness probes to OtelCollector

kyledharrington:
## Description

The DynaKube CR's `spec.templates.otelCollector` exposes fields for replicas, resources, tolerations, topology spread constraints, etc., but does not allow users to configure liveness or readiness probes for the OpenTelemetry Collector StatefulSet. As a result the collector container has no health checks — Kubernetes can't detect a hung or unhealthy collector, won't restart it, and won't gate traffic readiness on its health.

This PR adds optional `livenessProbe` and `readinessProbe` fields (both `*corev1.Probe`) to `OpenTelemetryCollectorSpec`, mirroring Kubernetes' own `PodSpec` convention so users can configure any probe type (HTTP/TCP/exec) with full control. When nil, no probe is applied — existing deployments are unaffected.

### Summary of changes
- Added `LivenessProbe` and `ReadinessProbe` to `OpenTelemetryCollectorSpec` in `pkg/api/v1beta5/dynakube/opentelemetry.go` and `pkg/api/latest/dynakube/opentelemetry.go` (v1beta6).
- Wired the fields through the v1beta5↔latest conversion functions.
- Threaded the probes into the container builder at `pkg/controllers/dynakube/otelc/statefulset/container.go`.
- Regenerated `zz_generated.deepcopy.go` and the CRD manifest YAMLs via `make manifests`.
- Added a unit test (`TestProbes`) covering both the default (nil) case and the custom-probe case.
- Extended the existing v1beta5 conversion round-trip tests to cover the new fields.

v1beta4 is intentionally left unchanged; conversion drops the new fields when stepping down.

No GitHub issue or ticket is linked; this addresses a gap encountered when deploying the operator against an environment that requires health-gated rollouts.

## How can this be tested?

A cluster with an existing DynaKube that has the OtelCollector active (via `spec.extensions.prometheus` or `spec.telemetryIngest`) is sufficient.

1. Build and deploy this branch's operator image to the cluster (e.g. `make deploy`).
2. Update the DynaKube CR with probes under `spec.templates.otelCollector`, for example:
\`\`\`yaml
spec:
  templates:
    otelCollector:
      livenessProbe:
        httpGet:
          path: /
          port: 13133
        initialDelaySeconds: 10
        periodSeconds: 30
      readinessProbe:
        httpGet:
          path: /
          port: 13133
        initialDelaySeconds: 5
        periodSeconds: 10
\`\`\`
3. Verify the probes are applied to the collector container:
\`\`\`sh
STS=\$(kubectl get sts -n dynatrace -o name | grep -i otel | head -1)
kubectl get \$STS -n dynatrace -o jsonpath='{.spec.template.spec.containers[0].livenessProbe}{"\n"}{.spec.template.spec.containers[0].readinessProbe}'
\`\`\`
Both probes should print as JSON matching the CR. When the fields are unset on the CR, the container should have no probes (preserving previous behavior).

Note: the snippet above uses the otel-collector's default `health_check` extension port (13133) for illustration. The probe handler is fully user-controlled, so an `exec` or `tcpSocket` probe works equally well if you don't run the collector's HTTP health extension.


StefanHauth:
This approach seems over-engineered to me. What benefit do we have from user-configurable liveness probes? The collector configuration (as of now) is under complete operator control and cannot be altered by users (at least in a meaningful way).
IMHO it would be better to tailor the probes to what is supported by the collector and (if necessary) adopt the configuration of the health extensions.

StefanHauth:
This has been implemented in a simpler fashion in #6787

PR #6681, opened 2026-05-15
Add image digest field to imageRef spec

waodim:
## Description

[ICP-4254](https://dt-rnd.atlassian.net/browse/ICP-4254)

With this PR `digest` field is added to the `imageRef` spec. Updated docs and added a new test file to test behavior.

## How can this be tested?

- unit tests
- deploy operator, test out the digest field for the components.

[ICP-4254]: https://dt-rnd.atlassian.net/browse/ICP-4254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/api/shared/image/types.go
+++ b/pkg/api/shared/image/types.go
@@ -12,8 +12,10 @@ type Ref struct {
 	// Indicates a tag of the image to use
 	Tag string `json:"tag,omitempty"`
 
-	// Digest pins the image to a specific content-addressable hash (e.g. sha256:...).
-	// When set, the tag is ignored when rendering the image reference.
+	// Digest pins the image to a specific content-addressable hash in the OCI
+	// `<algorithm>:<hex>` form (e.g. `sha256:abc...`). When set, the tag is ignored
+	// when rendering the image reference.
+	// +kubebuilder:validation:Pattern:="^[a-z0-9]+:([a-f0-9]+|[A-F0-9]+)$"
 	Digest string `json:"digest,omitempty"`
 
 	// Image pull policy to use
--- a/pkg/api/shared/image/types.go
+++ b/pkg/api/shared/image/types.go
@@ -45,12 +45,6 @@ func (ref Ref) String() string {
 	return ref.Repository + ":" + ref.Tag
 }
 
-// IsZero returns true if the image ref is empty.
-// Prefer this name over IsEmpty for compatibility with JSON omitzero.
-func (ref *Ref) IsZero() bool {
-	return ref == nil || *ref == Ref{}
-}
-
 // HasImage returns true when the ref points to a resolvable image — i.e. a repository
 // plus at least one of a tag or a digest.
 func (ref *Ref) HasImage() bool {
--- a/pkg/api/shared/image/types.go
+++ b/pkg/api/shared/image/types.go
@@ -12,11 +12,17 @@ type Ref struct {
 	// Indicates a tag of the image to use
 	Tag string `json:"tag,omitempty"`
 
+	// Digest pins the image to a specific content-addressable hash (e.g. sha256:...).
+	// When set, the tag is ignored when rendering the image reference.
+	Digest string `json:"digest,omitempty"`
+
 	// Image pull policy to use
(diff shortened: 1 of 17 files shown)


pkg/api/validation/dynakube/databases.go
```diff
@@ -30,7 +30,7 @@ func missingDatabaseExecutorImage(ctx context.Context, _ *Validator, dk *dynakub
 		return ""
 	}
 
-	if dk.Spec.Templates.SQLExtensionExecutor.ImageRef.Repository == "" || dk.Spec.Templates.SQLExtensionExecutor.ImageRef.Tag == "" {
+	if !dk.Spec.Templates.SQLExtensionExecutor.ImageRef.HasImage() {
```
0sewa0, line 33: 😍

0sewa0 (dismissed):

pkg/api/shared/image/types.go
```diff
@@ -39,6 +51,16 @@ func (ref *Ref) IsZero() bool {
 	return ref == nil || *ref == Ref{}
 }
 
+// HasImage returns true when the ref points to a resolvable image — i.e. a repository
+// plus at least one of a tag or a digest.
+func (ref *Ref) HasImage() bool {
```
avorima, line 52: I think this should actually replace IsZero. It stopped being correct after adding the pull policy. omitzero will still work OOTB with the annotation

waodim: good point. Replaces IsZero: https://github.com/Dynatrace/dynatrace-operator/pull/6681/commits/4dc1ef9ca06bab59c85c93404a3881fee16ea324

pkg/api/validation/dynakube/telemetryservice.go
```diff
@@ -163,7 +163,7 @@ func missingOtelCollectorImage(_ context.Context, _ *Validator, dk *dynakube.Dyn
 }
 
 func ignoredOtelCollectorTemplate(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
-	if !dk.TelemetryIngest().IsEnabled() && !dk.Spec.Templates.OpenTelemetryCollector.ImageRef.IsZero() {
+	if !dk.TelemetryIngest().IsEnabled() && dk.Spec.Templates.OpenTelemetryCollector.ImageRef.HasImage() {
```
avorima: Sorry, I forgot about this. In this case we actually want to check if any field is set to keep parity with the logmonitoring template warning

waodim: https://github.com/Dynatrace/dynatrace-operator/pull/6681/commits/47c3763e4f2b7f9fce607d2067a2b2dfcd37af91

pkg/api/shared/image/types.go
```diff
@@ -12,11 +12,17 @@ type Ref struct {
 	// Indicates a tag of the image to use
 	Tag string `json:"tag,omitempty"`
 
+	// Digest pins the image to a specific content-addressable hash (e.g. sha256:...).
```
avorima: Right now the "sha256:" prefix is a hidden requirement. Either it's optional, i.e. it's appended when resolving the image string from the ref, or we add validation so that users always specify it

waodim: WDYT about adding a validation via kubebuilder?

something along the lines like: `// +kubebuilder:validation:Pattern:="^[a-z0-9]+:[a-fA-F0-9]+$"`

avorima: Should probably be `^[a-z0-9]+:([a-f0-9]+|[A-F0-9]+)$`. Mixed-case hex encoding is likely invalid

waodim: https://github.com/Dynatrace/dynatrace-operator/pull/6681/commits/ce72b9cad44c486a67978e6ec15f8e68cab01eb4

avorima (approved):

PR #6684, opened 2026-05-15
Make node-image-pull only affect the CSI Job

0sewa0:
## Description

https://dt-rnd.atlassian.net/browse/ICP-4400

Now the `node-image-pull` FF only affects how the CSI driver pulls images.

If the image for the codemodule is know (because `codeModulesImage` is set or public-registy FF is used / phase-3) then the "self-extracting" init container can be injected, either because there is no CSI driver or the pod level annotation `oneagent.dynatrace.com/volume-type: "ephemeral"` is used.

## How can this be tested?

Scenarios:
- codeModulesImage/public-registry + No-CSI => self-extracting" init container can be injected
- codeModulesImage/public-registry + CSI + pod level annotation  => self-extracting" init container can be injected
- codeModulesImage/public-registry + CSI + node-image-pull FF => Jobs are used + csi volume gets injected
- codeModulesImage/public-registry + CSI + node-image-pull FF + pod level annotation  => self-extracting" init container can be injected


0sewa0:
~Note: I want to add a validation warning for `no-CSI + node-image-pull FF`, because that has no affect anymore~
- Edit: Done in https://github.com/Dynatrace/dynatrace-operator/pull/6684/commits/e68f63925b819f8b445ffad89a79c6a5b524e547

pkg/api/exp/oneagent.go
```diff
@@ -57,6 +57,12 @@ func (ff *FeatureFlags) IsNodeImagePull() bool {
 	return ff.getBoolWithDefault(OANodeImagePullKey, false)
 }
 
+func (ff *FeatureFlags) IsNodeImagePullSet() bool {
```
andriisoldatenko: what wrong with function on line 56? why we need new one?

0sewa0: this test case:
```
title:           "CSI not available, node-image-pull explicitly disabled => warning",
```

0sewa0: to be fare, explicitly setting the value to `false` never did anything, 
but if you set it to `false` now then you will skip the warning, even though IMO the warning still applies

0sewa0: (I want people to remove the FFs that have no affect)

andriisoldatenko: oh sorry., i misread it, we want to check if FF flag annotation is exist, I would rename function.

```
IsNodeImagePullAnnottaionIsPresent
```

```
IsNodeImagePullAnnottaionIsSet
```

0sewa0: we are already in the `FeatureFlags` struct, sayin `Annotation` feels like stuttering to me

andriisoldatenko: true, :morning: ☕

andriisoldatenko:
BTW may be we can extend tests after you PR https://github.com/Dynatrace/dynatrace-operator/pull/6679, or if @jokuniew first , we can follow up.

pkg/api/validation/dynakube/featureflag.go
```diff
@@ -34,3 +35,11 @@ func deprecatedFeatureFlag(_ context.Context, _ *Validator, dk *dynakube.DynaKub
 
 	return ""
 }
+
+func isNodeImagePullWithoutCSIDisabled(_ context.Context, v *Validator, dk *dynakube.DynaKube) string {
+	if !dk.OneAgent().IsCSIAvailable() && dk.FF().IsNodeImagePullSet() {
```
andriisoldatenko: FYI: my point of confusion regarding naming is comming from here, as you can see it's very similar to:
`dk.FF().IsNodeImagePull()` and 
`dk.FF().IsNodeImagePullSet()`.

0sewa0: I am working on a nicer solution

andriisoldatenko: but it's okey :) i think one can read impl.

pkg/api/validation/dynakube/featureflag.go
```diff
@@ -34,3 +39,11 @@ func deprecatedFeatureFlag(_ context.Context, _ *Validator, dk *dynakube.DynaKub
 
 	return ""
 }
+
+func isNodeImagePullWithoutCSI(_ context.Context, v *Validator, dk *dynakube.DynaKube) string {
+	if !dk.OneAgent().IsCSIAvailable() && dk.FF().IsSet(exp.OANodeImagePullKey) {
```
waodim, line 44: Since this uses `IsSet` it would also warn if the FF is set to `false`, no? Is this intentional? Am I missing something?

0sewa0: yes, that is intentional

0sewa0: (setting it to false does nothing)

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -18,10 +18,6 @@ import (
 const (
 	errorConflictingOneagentMode = `The DynaKube specification attempts to use multiple OneAgent modes simultaneously, which is not supported.`
 
-	errorImageFieldSetWithoutCSIFlag = `The DynaKube specification attempts to enable ApplicationMonitoring/CloudNativeFullstack mode and retrieve the respective codeModules image, but the CSI driver and/or node image pull is not enabled.`
```
waodim, line 21: Isn't this somewhat risky? Or in other words should we also remove the `(oa.IsCSIAvailable() || oa.featureBootstrapperInjection)` check in `GetCustomCodeModulesImage()` since, without it, AppMon + no CSI + FF unset + spec image set now passes validation but wouldn't the function in this case return "" and no the `CodeModulesImage`? Or am I wrong?

0sewa0: good catch, forgot to remove it there 👍

pkg/webhook/mutation/pod/mutator/metadata/mutator_test.go
```diff
@@ -144,39 +133,36 @@ func TestIsEnabled(t *testing.T) {
 			withoutCSI: true,
 		},
 		{
-			title:   "OA + FF enabled => enabled",
+			title:   "OA => disabled",
```
0sewa0, line 136: FYI: I changed this because in the rebase I noticed that the `withCSI` / `withoutCSI` values don't reflect the titles

waodim (approved):
Tested all scenarios, LGTM


example:
```
Warning: The `feature.dynatrace.com/node-image-pull` annotation is set, but the CSI driver is not available on this cluster. This feature flag only affects the behavior of the CSI driver, so it will have no effect. Other previous `node-image-pull` related behavior has been defaulted.
```

avorima (approved):

PR #6693, opened 2026-05-18
Propagate resourceAttributes on injection

StefanHauth:
## Description

Propagate 
* `.spec.resourceAttributes`
* `.spec.oneagent.*.additionalResourceAttributes`
* `.spec.otlpExporterConfiguration.additionalResourceAttributes`
to resp. components.

[JIRA #1](https://dt-rnd.atlassian.net/browse/ICP-3793)
[JIRA #2](https://dt-rnd.atlassian.net/browse/ICP-2453)
[JIRA #3](https://dt-rnd.atlassian.net/browse/ICP-2455)

### Metadata only
* deploy a DK with metadata enrichment but without OneAgent or OTLP exporter configuration, set `.spec.resourceAttributes`
* deploy sample app, make sure it it gets metadata injection
* the set resource attributes shall be part of the bootstrapper arguments in the injected init container

### OneAgent
* same as "metadata only" but with a OneAgent injection (e.g. app mon) configured
* set `.spec.resourceAttributes` + `.spec.oneagent.*.additionalResourceAttributes` 
* the set resource attributes shall be part of the bootstrapper arguments in the injected init container, `additionalResourceAttributes` take precedence

### OTLP
* same as "metadata only" but with OTLP exporter auto-config
* set `.spec.resourceAttributes` + `.spec.otlpExporterConfiguration.additionalResourceAttributes` 
* the set resource attributes shall be part of env var `OTEL_RESOURCE_ATTRIBUTES` on injected pods, `additionalResourceAttributes` take precedence


pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator.go
```diff
@@ -99,7 +102,9 @@ func (m *Mutator) addResourceAttributes(podAttrs *attributes.Pod, c *corev1.Cont
 		})
 	}
 
-	podAttrs.SetCustomAttributes(existingResourceAttrs)
+	// container's existing OTEL_RESOURCE_ATTRIBUTES wins on collisions with Dynakube-derived attributes
+	merged := sharedra.Merge(dkOTLPAttrs, existingResourceAttrs)
+	podAttrs.SetCustomAttributes(merged)
```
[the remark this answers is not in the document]
StefanHauth: The comment exactly states what's happening, the existing RAs shall have precedence over RAs from the DK, and they do.

pkg/webhook/mutation/pod/attributes/pod.go
```diff
@@ -18,6 +18,7 @@ type Pod struct {
 	rulesPropagate       map[string]string
 	namespaceAnnotations map[string]string
 	podAnnotations       map[string]string
+	dynakube             map[string]string
```
avorima: Can we add some comments here that document where we expect the data for each field to come from? I already forgot what custom was and had to go hunting for where it's used

pkg/webhook/mutation/pod/attributes/pod.go
```diff
@@ -107,6 +114,7 @@ func (attrs *Pod) combineForMetadataAnnotations() map[string]string {
 
 	// make sure we use the same precedence as in combine()
```
avorima: there is no combine() function

pkg/webhook/mutation/pod/attributes/pod.go
```diff
@@ -117,6 +125,7 @@ func (attrs *Pod) combineForJSONAnnotation() (string, error) {
 	combined := make(map[string]string)
 
 	// make sure we use the same precedence as in combineAll()
```
avorima: if they're meant to be the same, could we use one function to do it?

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator_test.go
```diff
@@ -484,3 +485,154 @@ func Test_Mutator_Reinvoke(t *testing.T) {
...
+		globalValue   = "global-value"
+		otlpKey       = "otlp-key"
+		otlpValue     = "otlp-value"
+		collisionKey  = "collision-key"
+		globalCollVal = "global-collision-value"
+		otlpCollVal   = "otlp-collision-value"
+		containerVal  = "container-value"
+	)
+
+	_ = appsv1.AddToScheme(scheme.Scheme)
```
avorima: both of these schemes are already registered. also, tests shouldn't modify this global variable since it has side-effects on other tests

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator_test.go
```diff
@@ -484,3 +485,154 @@ func Test_Mutator_Reinvoke(t *testing.T) {
...
+	_ = corev1.AddToScheme(scheme.Scheme)
+
+	baseNamespace := corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNamespace}}
+
+	newPod := func(containerEnv ...corev1.EnvVar) *corev1.Pod {
+		return &corev1.Pod{
+			TypeMeta:   metav1.TypeMeta{Kind: "Pod", APIVersion: "v1"},
```
avorima: Type meta isn't needed unless it's used to build a patch or ownerrefs

pkg/webhook/mutation/pod/mutator/metadata/mutator_test.go
```diff
@@ -427,6 +427,185 @@ func buildArgument(attr string, value string) string {
 	return fmt.Sprintf("--%s=%s=%s", podattr.Flag, attr, strings.ToLower(value))
 }
 
+func podAttrArg(key, value string) string {
+	return fmt.Sprintf("--%s=%s=%s", podattr.Flag, key, value)
+}
```
0sewa0: why we have a new one that is different from `buildArgument`? (by not calling `ToLower`)

also, this seems to be a returning pattern probably would make sense to have a public func for it somewhere

Image

pkg/controllers/dynakube/logmonitoring/daemonset/reconciler.go
```diff
@@ -46,7 +46,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, dk *dynakube.DynaKube) error
```
waodim, line 45: I noticed that when both `metadataEnrichment` and `otlpExporterConfiguration` are enabled on the same pod, OTLP `additionalResourceAttributes` lose to OA on collision for the `metadata.dynatrace.com `JSON annotation (the OA-side metadata mutator writes the annotation first via `SetAnnotationIfNotExists`, so the OTLP mutator's write no-ops). But AFAIK we agreed the opposite "set both, OTLP wins on collision." Isn't this misleading? Should we align it, or keep as-is and document the caveat?

Example:

```
spec:
  apiUrl: https://....dev.dynatracelabs.com/api

  # Global attrs
  resourceAttributes:
    dt.shared.global: "from-global"

  # Explicitly enabled so the metadata mutator definitely runs
  metadataEnrichment:
    enabled: true

  oneAgent:
    cloudNativeFullStack:
      additionalResourceAttributes:
        dt.shared.collision: "from-oa"    # collides with OTLP below
        dt.oa.only: "oa-value"            # OA-only

  otlpExporterConfiguration:
    signals:
      metrics: {}
      traces: {}
      logs: {}
    additionalResourceAttributes:
      dt.shared.collision: "from-otlp"    # this should win
      dt.otlp.only: "otlp-value"          # OTLP-only, should appear in annotation
```


Actual result:

```
│ Annotations:      dynakube.dynatrace.com/injected: true                                                                                                                                                                                                    │
│                   metadata-enrichment.dynatrace.com/injected: true                                                                                                                                                                                         │
│                   metadata.dynatrace.com: {"dt.oa.only":"oa-value","dt.shared.collision":"from-oa","dt.shared.global":"from-global"}                                                                                                                       │
│                   metadata.dynatrace.com/dt.oa.only: oa-value                                                                                                                                                                                              │
│                   metadata.dynatrace.com/dt.otlp.only: otlp-value                                                                                                                                                                                          │
│                   metadata.dynatrace.com/dt.shared.collision: from-oa                                                                                                                                                                                      │
│                   metadata.dynatrace.com/dt.shared.global: from-global                                                                                                                                                                                     │
│                   metadata.dynatrace.com/k8s.workload.kind: deployment                                                                                                                                                                                     │
│                   metadata.dynatrace.com/k8s.workload.name: nginx                                                                                                                                                                                          │
│                   oneagent.dynatrace.com/inject: true                                                                                                                                                                                                      │
│                   oneagent.dynatrace.com/injected: true                                                                                                                                                                                                    │
│                   otlp-exporter-configuration.dynatrace.com/injected: true
```


TLDR:

```
With:
  spec.resourceAttributes:                 { dt.shared.global: from-global }
  spec.oneAgent.cloudNativeFullStack.additionalResourceAttributes:
    dt.shared.collision: from-oa
    dt.oa.only: oa-value
  spec.otlpExporterConfiguration.additionalResourceAttributes:
    dt.shared.collision: from-otlp   # should win
    dt.otlp.only: otlp-value

Got:
  metadata.dynatrace.com = {
    "dt.oa.only":"oa-value",
    "dt.shared.collision":"from-oa",       # OA wins; should be "from-otlp"
    "dt.shared.global":"from-global"
                                            # dt.otlp.only missing entirely
  }
```


Could be that I am entirely wrong here, but that's just something I really want to clarify before we merge

StefanHauth: This is contributed to the previous refactoring. All annotations are now set on a "set-if-not-present" basis, and as OA is first, it now wins. Given, that the decision on what should win was on taken based on what's implemented, we could just document it as-is.

waodim: fine by me. What's your take? @0sewa0

0sewa0: TLDR: basically what stefan says :D 

the `metadata.dynatrace.com` is a shared thing, so if there is a collision then we can consider it as user error / tech limitation.

Nobody has a preference about "which should win", because it doesn't really make things better whichever way we go. So we will just let the impl decide. Just make sure we document the correct thing :D

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator_test.go
StefanHauth: [fb8dcdd](https://github.com/Dynatrace/dynatrace-operator/pull/6693/commits/fb8dcdd3949aca72da87872abe129202d40c3f74)

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/mutator_test.go
StefanHauth: [fb8dcdd](https://github.com/Dynatrace/dynatrace-operator/pull/6693/commits/fb8dcdd3949aca72da87872abe129202d40c3f74)

pkg/webhook/mutation/pod/attributes/pod.go
StefanHauth: [fb8dcdd](https://github.com/Dynatrace/dynatrace-operator/pull/6693/commits/fb8dcdd3949aca72da87872abe129202d40c3f74)

pkg/webhook/mutation/pod/attributes/pod.go
StefanHauth: [fb8dcdd](https://github.com/Dynatrace/dynatrace-operator/pull/6693/commits/fb8dcdd3949aca72da87872abe129202d40c3f74)

pkg/webhook/mutation/pod/attributes/pod.go
StefanHauth: I had a similar thought, I don't like this code duplication either. Tried a refactoring in [753321f](https://github.com/Dynatrace/dynatrace-operator/pull/6693/commits/753321f57c8874354ed33ac3598846c2e3224d6d)

pkg/webhook/mutation/pod/mutator/metadata/mutator_test.go
StefanHauth: [fb8dcdd](https://github.com/Dynatrace/dynatrace-operator/pull/6693/commits/fb8dcdd3949aca72da87872abe129202d40c3f74)

pkg/webhook/mutation/pod/mutator/metadata/mutator_test.go
```diff
@@ -423,8 +421,179 @@ func TestMutate(t *testing.T) {
...
+	}
+
+	cases := []testCase{
+		{
+			name: "OA enabled, no resource attributes set",
+			dkSpec: dynakube.DynaKubeSpec{
+				OneAgent: oneagent.Spec{
+					ApplicationMonitoring: &oneagent.ApplicationMonitoringSpec{},
+				},
+				MetadataEnrichment: metadataenrichment.Spec{Enabled: ptr.To(true)},
```
avorima: use new instead of ptr.To

pkg/api/validation/dynakube/resource_attributes_test.go
```diff
@@ -314,3 +314,130 @@ func TestResourceAttributesSyntaxValidation(t *testing.T) {
...
+
+	t.Run("no oneAgent additionalResourceAttributes — sanitization validators do not fire", func(t *testing.T) {
+		dk := &dynakube.DynaKube{
+			ObjectMeta: defaultDynakubeObjectMeta,
+			Spec: dynakube.DynaKubeSpec{
+				APIURL:             testAPIURL,
+				ResourceAttributes: map[string]string{"k8s.pod.name": "val"},
+			},
+		}
+		warnings, err := assertAllowed(t, dk)
```
jokuniew:
```suggestion
		assertAllowedWithoutWarnings(t, dk)
```

pkg/webhook/mutation/pod/attributes/combine_test.go
```diff
@@ -0,0 +1,330 @@
...
+func TestCombine_ContainerWinsOverClusterInfo(t *testing.T) {
+	attrs := newTestPodAttributes()
+	attrs.clusterInfo[K8sContainerNameAttr] = "from-cluster"
+	container := Container{ContainerName: "from-container"}
+
+	result := attrs.combineAll(container)
+
+	assert.Equal(t, "from-container", result[K8sContainerNameAttr])
+}
+
```
jokuniew, line 56: the dynakube field in `newTestPodAttributes` is not initialized and not tested here 
```
		dynakube:             make(map[string]string),
```
plus
```suggestion
func TestCombine_DynakubeWinsOverContainer(t *testing.T) {
	attrs := newTestPodAttributes()
	container := Container{ContainerName: "from-container"}
	attrs.dynakube[K8sContainerNameAttr] = "from-dynakube"

	result := attrs.combineAll(container)

	assert.Equal(t, "from-dynakube", result[K8sContainerNameAttr])
}
```
plus inlcude `dynakube` in `TestCombine_UniqueKeysFromAllSourcesMerged`

pkg/webhook/mutation/pod/mutator/metadata/mutator_test.go
```diff
@@ -408,8 +406,179 @@ func TestMutate(t *testing.T) {
...
+					ApplicationMonitoring: &oneagent.ApplicationMonitoringSpec{
+						AdditionalResourceAttributes: map[string]string{collisionKey: addCollVal},
+					},
+				},
+				MetadataEnrichment: metadataenrichment.Spec{Enabled: new(true)},
+				ResourceAttributes: map[string]string{collisionKey: globalCollVal},
+			},
+			wantArgs:    []string{attributes.ToArg(collisionKey, addCollVal)},
+			notWantArgs: []string{attributes.ToArg(collisionKey, globalCollVal)},
+		},
```
jokuniew, line 496: there might be rare case when user sets .spec.resourceAttributes that get sanitized into valid OA/OTEL key
i've run the test 100 times, 18 times the global value was picked (test fail)
It would be better in `GetResourceAttributes` to sanitize attributes first and then merge them
or reject this configuration during validation
```suggestion
		},
				{
			name: "OA enabled, global got renamed, key collision - random pick",
			dkSpec: dynakube.DynaKubeSpec{
				OneAgent: oneagent.Spec{
					ApplicationMonitoring: &oneagent.ApplicationMonitoringSpec{
						AdditionalResourceAttributes: map[string]string{"collision_key": addCollVal},
					},
				},
				MetadataEnrichment: metadataenrichment.Spec{Enabled: new(true)},
				ResourceAttributes: map[string]string{"collision/key": globalCollVal},
			},
			wantArgs:    []string{attributes.ToArg("collision_key", addCollVal)},
			notWantArgs: []string{attributes.ToArg("collision_key", globalCollVal)},
		},
```

pkg/api/validation/dynakube/resource_attributes_test.go
```diff
@@ -314,3 +314,130 @@ func TestResourceAttributesSyntaxValidation(t *testing.T) {
...
+		}
+		assertAllowedWithoutWarnings(t, dk)
+	})
+
+	t.Run("global key with slash produces warning", func(t *testing.T) {
+		dk := &dynakube.DynaKube{
+			ObjectMeta: defaultDynakubeObjectMeta,
+			Spec: dynakube.DynaKubeSpec{
+				APIURL:             testAPIURL,
+				ResourceAttributes: map[string]string{"foo/bar": "val"},
```
jokuniew, line 335: when i apply dynakube with
```yaml
  resourceAttributes:
    "foo/bar": "v1"
```
i get no validation warning about rename to foo_bar

jokuniew, line 335: when i apply dynakube with
```yaml
  resourceAttributes:
    "foo/bar": "v1"
```
i get no validation warning in terminal

there is warning in `runValidators` but they are not getting displayed.

aorcholski: It seems that multi-line messages are not printed. I removed all \n and the message appeared on the console.

per key line : \n -> ,
between key lines and summary line: \n -> TAB
```
Warning: spec.resourceAttributes contains invalid keys that will be automatically renamed:,   * key "foo/bar" will be renamed to "foo_bar" TAB Consider updating these keys in your DynaKube spec to avoid confusion.
```

pkg/api/shared/resourceattributes/sanitize.go
```diff
@@ -0,0 +1,18 @@
...
+
+var (
+	invalidKeyCharsRe      = regexp.MustCompile(`[^a-zA-Z0-9\-_.]`)
+	invalidKeyBoundariesRe = regexp.MustCompile(`(?:^[^a-zA-Z0-9]+)|(?:[^a-zA-Z0-9]+$)`)
+)
+
+// SanitizeKey returns a sanitized copy of key that is valid as a Kubernetes annotation name segment.
+// Characters outside [a-zA-Z0-9\-_.] are replaced with '_', and leading/trailing non-alphanumeric
+// characters are stripped. Returns an empty string if nothing valid remains.
+func SanitizeKey(key string) string {
```
0sewa0, line 13: where was this behavior discussed?

not that I am against it, just wanting to make sure that this is somewhere written down (that is not github) that we will sanitize the keys. (I checked the epic / tickets, it is mentioned nowhere)

sidenote: probably should have been a separate PR just for the sanitize / validation part

waodim: Should we create a little story for it? To have it documented

or does a comment suffice?

0sewa0: comment is enough for me, or description update

StefanHauth: The sanitization is kind of a must-have. Otherwise it would be possible to break pods (and hence customer app deployments) with the new resource attribute feature.

0sewa0: I totally understand why we do it, but we just have to document this for everyone to understand the limitations. 👍

pkg/api/validation/dynakube/resource_attributes_test.go
jokuniew: still not applied

pkg/webhook/mutation/pod/mutator/metadata/mutator_test.go
jokuniew: since[ the additional validation was added](https://github.com/Dynatrace/dynatrace-operator/pull/6693/changes#diff-387e6a75c5133671302f81ae2f50e6c91272a6059a5789cbc55e43726e3f809eR418-R444) i don't think that this test  `"OA enabled, global got renamed, key collision - random pick` should be added to source code.
it was only for review purposes to outline the problem

pkg/api/validation/dynakube/resource_attributes_test.go
StefanHauth: It's fixed in 50e0326ad7fdd86711026598cec184bfee144716

jokuniew: ah, im blind, sorry!

0sewa0 (approved):

jokuniew (approved):

PR #6723, opened 2026-05-20
fix e2e - read vars in workflow calls and pass value to actions

jokuniew:
## Description

in #6679 there was `vars` added to composite action
based on [bugs on github ](https://github.com/orgs/community/discussions/49689#discussioncomment-5266453)and error message in e2e tests - the composite action can't read the `vars`
they have to be called in workflows and passed as input parameters

## How can this be tested?
run e2e tests on workflow from this branch
tests are starting - previously they fail immediately

### DIFF
--- a/.github/actions/run-e2e/action.yaml
+++ b/.github/actions/run-e2e/action.yaml
@@ -50,6 +50,9 @@ inputs:
   github-token:
     description: The GitHub token
     required: true
+  e2e-public-registry-override:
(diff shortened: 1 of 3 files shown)


jokuniew:
env is getting passed to action - test is starting without env error, marking as ready for review 
image
image

avorima (approved):

Incredible-O (approved):

PR #6724, opened 2026-05-20
Automatically enable public registry and warn if FF is set

gkrenn:
## Description

This PR automatically enables the public registry feature as soon as a platform token is used. 
In addition, if the FF + platform token is used, a warning will be displayed.

[ICP-3643 Use new endpoint to update OA/CM/AG and ignore FF with warning when gen3 (platformToken) detected - Jira](https://dt-rnd.atlassian.net/browse/ICP-3643)

## How can this be tested?

Ensure you have a gen 3 tenant.

- Create platform token
- deploy operator without CSI
- deploy cloudnativefs
- check oneagent/ag images
- create application and check init container/code module

### DIFF
--- a/pkg/api/exp/flag.go
+++ b/pkg/api/exp/flag.go
@@ -22,11 +22,7 @@ type FeatureFlags struct {
 	hasPlatformToken bool
 }
 
-func NewFlags(annotations map[string]string) *FeatureFlags {
-	return &FeatureFlags{annotations: annotations}
-}
-
-func NewFlagsWithPlatformToken(annotations map[string]string, hasPlatformToken bool) *FeatureFlags {
+func NewFlags(annotations map[string]string, hasPlatformToken bool) *FeatureFlags {
 	return &FeatureFlags{annotations: annotations, hasPlatformToken: hasPlatformToken}
 }
 
--- a/pkg/api/exp/flag.go
+++ b/pkg/api/exp/flag.go
@@ -18,11 +18,16 @@ const (
 )
 
 type FeatureFlags struct {
-	annotations map[string]string
+	annotations      map[string]string
+	hasPlatformToken bool
 }
 
 func NewFlags(annotations map[string]string) *FeatureFlags {
 	return &FeatureFlags{annotations: annotations}
+}
+
+func NewFlagsWithPlatformToken(annotations map[string]string, hasPlatformToken bool) *FeatureFlags {
+	return &FeatureFlags{annotations: annotations, hasPlatformToken: hasPlatformToken}
 }
 
 // IsSet checks if the annotation(feature-flag) is present, does not check the value in any way.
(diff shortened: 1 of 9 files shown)


pkg/api/exp/flag.go
```diff
@@ -18,13 +18,18 @@ const (
...
 
 type FeatureFlags struct {
-	annotations map[string]string
+	annotations      map[string]string
+	hasPlatformToken bool
 }
 
 func NewFlags(annotations map[string]string) *FeatureFlags {
 	return &FeatureFlags{annotations: annotations}
 }
```
0sewa0: could we mark this as `deprecated`, because it is only used in prev apiVersions

avorima: I'd rather have one function that just takes a bool in v1beta6

0sewa0: yeah, you could just have `false` for all other apiVersions for that bool

avorima (dismissed):
Tested upgrade from classic token -> platform. Pods get restarted and workload gets injected using the public registry image

pkg/api/validation/dynakube/public_registry_test.go
```diff
@@ -47,4 +48,46 @@ func TestPublicRegistryOverrideWithoutPublicRegistry(t *testing.T) {
...
+	t.Run("FF set with platform token returns warning", func(t *testing.T) {
+		dk := newDynakube(map[string]string{exp.UsePublicRegistryKey: "true"})
+		warnings, _ := assertAllowed(t, dk, platformTokenSecret())
+		assert.Contains(t, warnings, fmt.Sprintf(warningPublicRegistryFlagIgnoredForPlatformToken, exp.UsePublicRegistryKey))
+	})
+
+	t.Run("FF set with regular token returns no warning", func(t *testing.T) {
+		dk := newDynakube(map[string]string{exp.UsePublicRegistryKey: "true"})
+		warnings, _ := assertAllowed(t, dk, regularTokenSecret())
+		assert.Empty(t, warnings)
```
jokuniew:
```suggestion
		assertAllowedWithoutWarnings(t, dk, regularTokenSecret())
```
in both test scenarios

jokuniew (dismissed):
nitpick with test assertion

deployed operator without CSI driver with platformtoken + FF + pull secrets for dt and apps
flag is ignored, warning displayed 
oa/ag/cm pulled public registry images

pkg/api/exp/flag.go
gkrenn: Create one combined `NewFlags` constructor

pkg/api/validation/dynakube/public_registry_test.go
gkrenn: Done

0sewa0 (approved):

PR #6738, opened 2026-05-20
Add check for stale network-zone AG Service IPs

0sewa0:
## Description
[ICP-3504](https://dt-rnd.atlassian.net/browse/ICP-3504)                                                                                                                                                                                                                                 

Adds a check for identifying "stale IPs" in a networkzone setup and waiting for it to resolve before starting/restarting the OneAgent.

## How can this be tested?                                                                                                                                                                                                                                                                                     
1. [Create a restricted networkzone](https://docs.dynatrace.com/docs/ingest-from/setup-on-k8s/guides/networking-security-compliance/network-configurations/network-zones#kubernetes-cluster-with-restricted-egress)
2. Create a DynaKube with a local ActiveGate that has the `routing` capability + networkzone from step 1 + a CloudNative enabled
3. Wait until everything gets read
4. Delete the DynaKube, and recreate it
  - This will make it that the AG will not be deregistered fast enough and the IP changes because of new Service being created -> stale IP
5. Everything gets ready and can connect

Example condition:
<img width="738" height="111" alt="image" src="https://github.com/user-attachments/assets/adf9c340-f8e9-4883-b81e-25110a1debf5" />

[ICP-3504]: https://dt-rnd.atlassian.net/browse/ICP-3504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controllers/dynakube/connectioninfo/oneagent/restricted_networkzone.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+// restrictedNetworkZoneEndpointCount is the number of OneAgent endpoints the Dynatrace
+// cluster returns when the tenant lives in a restricted network-zone served only by a
+// single local ActiveGate: one IP-based endpoint plus the AG Service's DNS entry point.
+//
+// The ActiveGate's Service is a ClusterIP service, with a default(unset) ipFamilyPolicy which is SingleStack.
+// If we used DualStack, then it would be possible to have 2 IP-based endpoints (ipv4/ipv6).
+// This logic/const expects that we are always using a SingleStack Service.
+const restrictedNetworkZoneEndpointCount = 2
```
0sewa0: Open question: Should I remove this expectation?

I don't see why this would change, but implementing this in a way to support DualStack is not that complicated.

avorima: A more general approach would go over the endpoints to see if the service IP(s) are present. Could be gated by checking if a network zone is set, unless we expect the default zone to also be restricted

0sewa0: yes but if there are valid entries in the endpoints, then the OA can start, and will reconnect to the local AG once it registers and the OA decides to change endpoints

~I don't want to block the rollout of the OA basically always when networkzones are used~

EDIT: I will just try it, because it trivializes the implementation soo much, will ask product if we are fine with this.

0sewa0: Good news, it is much simpler now: https://github.com/Dynatrace/dynatrace-operator/pull/6738/commits/fba115dc61363fa05518a539455a152134fc57f5

and product agreed that my cursed check for restricted networkzone is not needed

pkg/controllers/dynakube/connectioninfo/oneagent/reconciler.go
```diff
@@ -135,11 +150,6 @@ func (r *reconciler) reconcileConnectionInfo(ctx context.Context) error {
 	return nil
 }
 
-func (r *reconciler) setDynakubeStatus(connectionInfo oneagent.ConnectionInfo) {
-	r.dk.Status.OneAgent.ConnectionInfo.TenantUUID = connectionInfo.TenantUUID
-	r.dk.Status.OneAgent.ConnectionInfo.Endpoints = connectionInfo.Endpoints
-}
-
```
0sewa0, line 142: will undo this, it is "correct", but IMO it is more informative to show the endpoints in the status, even if they are stale

pkg/controllers/dynakube/activegate/capability/capability.go
```diff
@@ -89,6 +89,12 @@ func BuildServiceName(dynakubeName string) string {
 	return dynakubeName + "-" + consts.MultiActiveGateName
```
0sewa0, line 89: changes in this file are leftovers from a previous impl.

they are just renames, but IMO an improvement.

can undo if people disagree 😅

avorima (approved):
I couldn't verify the condition appears, but code LGTM

PR #6757, opened 2026-05-25
Introduce Kubernetes Monitoring (kubemon) operand 

jdanieck:
## Description

Splits kubernetes-monitoring out of the all-in ActiveGate into its own operand, so it can be sized independently from routing/ingest.

This PR delivers a minimal working operand — enough to provision a functioning kubemon StatefulSet end-to-end (behind a gate), with the remaining runtime features to follow. 

### What's in this PR

- `spec.kubernetesMonitoring` & `status.kubernetesMonitoring` on DynaKube v1beta6 (presence enables the operand) + ; CRD regenerated
- kubmon reconciler with connection-info and StatefulSet sub-reconcilers, wired into the DynaKube controller
- Unit tests (mocked) and envtest integration tests for both sub-reconcilers
- Temporary gate keeping the operand disabled by default (`KUBEMON_OPERAND_ENABLED` / `operator.kubemonOperandEnabled`)

### Shared-code touchpoints (review carefully)

- Extracted the StatefulSet rollout check into k8sstatefulset.IsRolloutComplete; ActiveGate phase detection now uses it too — logic is equivalent, but it's a shared path now.
- Demoted the generic query's "update not needed" log from Info to Debug — affects all reconcilers, not just kubemon.

### Design decisions worth a look

- Single condition — the reconciler owns one `KubernetesMonitoringAvailable` condition instead of one per resource, for a simpler status surface.

- K8S cached client throughout — reads and writes go through the controller-manager cache, in production and in envtests alike. Tests use `require.Eventually` loops to absorb cache lag rather than reading through to the API server.

- DT client caching instead of status caching — the AG reconciler caches the token hash in status to skip API fetches; kubemon fetches fresh each reconcile as the cache is now built into the Dynatrace client instead.

- New reconcilers rather than reusing the AG ones — the AG reconcilers set per-resource conditions and write AG-specific status fields, which don't fit kubemon's single-condition model. Rather than refactor them and risk breaking the legacy AG path, kubemon gets its own focused reconcilers.

### Merging rationale

Merged to main incrementally rather than as one large PR — this makes review easier, avoids merge conflicts and lets us verify early that nothing in the existing functionality breaks. The gate keeps the operand disabled by default, so this WIP never reaches users until it's complete.

ICP-2300

## How can this be tested?

1. Enable the gate:
```sh
make build-local
make deploy-local
  -n dynatrace \
  --reuse-values \
  --set operator.kubemonEnableOperand=true
```
2. Apply a DynaKube that keeps the ActiveGate but drops its kubernetes-monitoring capability, and adds the new kubernetesMonitoring section:
```yaml
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
spec:
  apiUrl: https://ENVIRONMENTID.live.dynatrace.com/api
  activeGate:
    capabilities:
      - routing        # kubernetes-monitoring intentionally omitted
  kubernetesMonitoring:
    image: "public.ecr.aws/dynatrace/dynatrace-activegate:1.337.36.20260526-135434"   # required until image discovery lands
    replicas: 1
```
3. Verify the operand comes up:
```sh
# owned resources exist
kubectl -n dynatrace get statefulset dynakube-kubemon
kubectl -n dynatrace get configmap dynakube-kubemon-connection-info
kubectl -n dynatrace get secret dynakube-kubemon-tenant-secret

# condition reached True
kubectl -n dynatrace get dynakube dynakube \
  -o jsonpath='{.status.conditions[?(@.type=="KubernetesMonitoringAvailable")]}'

# connection info populated in status
kubectl -n dynatrace get dynakube dynakube \
  -o jsonpath='{.status.kubernetesMonitoring.connectionInfoStatus}'
```
4. The condition should reach status: `True`, reason `Available`.
5. Remove `spec.kubernetesMonitoring` and confirm the StatefulSet, ConfigMap, and Secret are deleted and the condition is cleared.


pkg/api/latest/dynakube/kubemon/status.go
```diff
@@ -0,0 +1,15 @@
+package kubemon
+
+import (
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/shared/communication"
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/status"
+)
+
+// +kubebuilder:object:generate=true
+
+type Status struct {
```
jdanieck, line 10: @0sewa0 @andriisoldatenko @avorima could you please explain why we have connectionInfoStatus & version in the AG status? Do we need it for something else than caching the value ? Since we have a caching DT client now I am wondering if I can drop it in AG Kubemon.

andriisoldatenko: it's handy when we describe DK to see what's wrong, DT cache is not included in support-archive.

pkg/util/kubernetes/objects/internal/query/query.go
```diff
@@ -103,7 +103,7 @@ func (c Generic[T, L]) CreateOrUpdate(ctx context.Context, newObject T) (bool, e
 	}
 
 	if c.IsEqual(currentObject, newObject) {
-		c.log(ctx, newObject).Info("update not needed, no changes detected")
+		c.log(ctx, newObject).Debug("update not needed, no changes detected")
```
jdanieck, line 106: demoted to Debug as it is cluttering logs at Info; please let me know if it was Debug for some specific reason (which I don't see)

pkg/controllers/dynakube/kubemon/reconciler.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// Reconciler orchestrates the kubemon operand. Sub-reconciler fields are interfaces so they
+// can be mocked in tests.
+type Reconciler struct {
+	connectionInfoReconciler connectionInfoReconciler
+	statefulsetReconciler    statefulsetReconciler
+}
+
+func NewReconciler(kubeClient client.Client) *Reconciler {
```
jdanieck, line 57: I am building the reconciler so it is safe to use with cached k8s client from the beginning. The convention in code is to pass both clients `NewReconciler(clt client.Client, apiReader client.Reader)` - since I don't want to use direct API client, I deviated on purpose to make intent explicit. @andriisoldatenko @avorima @0sewa0 please let me know early if there are any objections.

pkg/api/latest/dynakube/kubemon/status.go
jdanieck: Ok, so it is used for debug, good enough for me to keep it

pkg/controllers/dynakube/kubemon/reconciler.go
0sewa0: there is some research ticket in the backlog about this, so maybe mention it there or something

pkg/api/latest/dynakube/kubemon/status.go
0sewa0: it is used for "caching" and sharing it between reconcilers

if we cached it on the dtclient it would be nicer IMO, but I agree with @andriisoldatenko that we would lose informational value.

However, I doubt that users (other than us) actually look at the `status`, especially the list of endpoints.
- So if we just logged the responses from the API, we wouldn't lose much.
   - Logs are actually better, because the `status` only shows me the moment in time when the support-archive ran, and not the 5min prior when the `status` may have shown me something totally different that could indicate the actual issue. 

On a related note: We should consider what we want to use the `status` for, because we overused it, now we want to move away from it. We should find some balance.

pkg/controllers/dynakube/kubemon/reconciler.go
avorima: using the cached client is fine as long as the reconciler only manages resources in the dynatrace namespace. we can revisit this limitation in the context of ICP-930

jdanieck:
I am still testing manually, but the PR is in good enough shape to start reviewing.

test/integrationtests/environment.go
```diff
@@ -70,6 +71,66 @@ func SetupTestEnvironment(tb testing.TB, opts ...TestEnvOpt) client.Client {
...
+		tb.Fatal(err)
+	}
+
+	if setup != nil {
+		if err := setup(mgr); err != nil {
+			tb.Fatal(err)
+		}
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
avorima: t.Context would have the same effect

pkg/api/latest/dynakube/kubemon/status.go
```diff
@@ -0,0 +1,19 @@
...
+	status.VersionStatus `json:",inline"`
+
+	// Information about KubernetesMonitoring's connections.
+	ConnectionInfo communication.ConnectionInfo `json:"connectionInfoStatus,omitempty"`
```
avorima:
```suggestion
	ConnectionInfo communication.ConnectionInfo `json:"connectionInfo,omitempty"`
```

pkg/api/latest/dynakube/kubemon/spec.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Registration configures automatic cluster registration in Dynatrace.
+// Presence (even as {}) enables registration — follows CRD presence-based enablement.
+type Registration struct {
+	// ClusterName is the display name used during registration. Defaults to the DynaKube name.
+	// +kubebuilder:validation:Optional
+	ClusterName string `json:"clusterName,omitempty"`
+
+	// Enable the Kubernetes app (cluster details, workload views) in Dynatrace.
```
avorima, line 109: We deprecated the feature flag in #6822. This field should also be marked as deprecated

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_integration_test.go
```diff
@@ -0,0 +1,258 @@
...
+	t.Run("re-enable", func(t *testing.T) { runReEnablePhase(t, deps) })
+}
+
+func runProvisionPhase(t *testing.T, deps lifecycleDeps) {
+	t.Helper()
+
+	dtClient := agclientmock.NewClient(t)
+	dtClient.EXPECT().GetConnectionInfo(anyContext).Return(deps.baselineConnectionInfo, nil)
+
+	require.Eventually(t, func() bool {
```
avorima: the eventually blocks should only cover the parts that are actually eventually consistent. reconcile should run exactly once

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_integration_test.go
```diff
@@ -0,0 +1,258 @@
...
+func runStabilizePhase(t *testing.T, deps lifecycleDeps) {
+	t.Helper()
+
+	dtClient := agclientmock.NewClient(t)
+	dtClient.EXPECT().GetConnectionInfo(anyContext).Return(deps.rotatedConnectionInfo, nil)
+
+	// Cached reads can lag writes, so this no-op assertion compares uncached APIReader
+	// resourceVersions across two reconciles with identical input. If either object is
+	// rewritten, the API server bumps its resourceVersion and this condition will fail.
+	require.Eventually(t, func() bool {
```
avorima: this should be a [Consistently](https://pkg.go.dev/github.com/onsi/gomega#Consistently) when using gomega. i would not expect the return value to ever be false

pkg/api/latest/dynakube/kubemon/props.go
```diff
@@ -0,0 +1,61 @@
...
+
+	OperandNameSuffix = "-kubemon"
+
+	ServiceAccountName = "dynatrace-activegate"
+)
+
+// KubeMon wraps Spec and Status for ergonomic access via dk.KubernetesMonitoring().
+type KubeMon struct {
+	*Spec
+	*Status
```
jokuniew, line 16: that's a nit-picky comment in comparison to scope of this PR 
but, what's the purpose of combined Spec+Status props when 
1. whole repo handles spec and status separately
2. status is not used in this props.go file at all?
In general I've seen strong separation for spec and status handling.

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_integration_test.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func assertManagedLabels(t *testing.T, labels map[string]string, dk *dynakube.DynaKube) {
+	t.Helper()
+
+	assert.Equal(t, k8slabel.ActiveGateComponentLabel, labels[k8slabel.AppComponentLabel])
+	assert.Equal(t, dk.Name, labels[k8slabel.AppCreatedByLabel])
+}
+
+func assertOwnedBy(t *testing.T, refs []metav1.OwnerReference, dk *dynakube.DynaKube) {
```
avorima: could use `metav1.IsControlledBy` instead

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_test.go
```diff
@@ -0,0 +1,307 @@
...
+	t.Run("returns error when getting connection info fails", func(t *testing.T) {
+		dk := newTestDynaKube(true)
+		fakeClient := fake.NewClient(dk)
+		r := connectioninfo.NewReconciler(fakeClient)
+
+		dtClient := agclientmock.NewClient(t)
+		dtClient.EXPECT().GetConnectionInfo(anyCtx).Return(agclient.ConnectionInfo{}, errors.New("dt api error")).Once()
+
+		err := r.Reconcile(t.Context(), dtClient, dk)
+		require.Error(t, err)
```
avorima: use ErrorIs to ensure the returned error is the one we expect

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_test.go
```diff
@@ -0,0 +1,307 @@
...
+				dk := newTestDynaKube(true)
+				fakeClient := fake.NewClient(dk)
+				r := connectioninfo.NewReconciler(fakeClient)
+
+				info := testConnectionInfo()
+				mutate(&info)
+				dtClient := newDTClientMock(t, info)
+
+				err := r.Reconcile(t.Context(), dtClient, dk)
+				require.Error(t, err)
```
avorima: this could also be improved by having an error variable for "not ready". take care to not use pkg/errors for these variables

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_test.go
```diff
@@ -0,0 +1,307 @@
...
+		err := r.Reconcile(t.Context(), dtClient, dk)
+		require.Error(t, err)
+
+		assert.Empty(t, dk.Status.KubernetesMonitoring.ConnectionInfo)
+
+		assertResources(t, fakeClient, dk, false, false)
+	})
+
+	t.Run("returns transient error when connection info is incomplete, creates no resources", func(t *testing.T) {
+		tests := map[string]func(*agclient.ConnectionInfo){
```
avorima: this map messes with subtest detection. also, tests should have deterministic order. randomness should be introduced with `-shuffle`.
for brevity, struct fields do not need to specified in the initialization if all fields are set

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_test.go
```diff
@@ -0,0 +1,307 @@
...
+	})
+}
+
+// TestReconcileWriteFailures covers create-path failures per resource. A ConfigMap failure aborts
+// before any write; a Secret failure leaves the ConfigMap behind — both must leave status empty.
+func TestReconcileWriteFailures(t *testing.T) {
+	// In both cases status must stay empty and the failing write is never persisted. They differ
+	// only in whether the ConfigMap survives: a ConfigMap failure aborts before either object is
+	// written, while a Secret failure happens after the ConfigMap was already created.
+	tests := map[string]struct {
```
avorima: use list of structs (repeated)

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -0,0 +1,237 @@
...
+	desiredStatefulSet, err := r.buildDesiredStatefulSet(ctx, dk)
+	if err != nil {
+		return err
+	}
+
+	if _, err = r.sts.WithOwner(dk).CreateOrUpdate(ctx, desiredStatefulSet); err != nil {
+		return err
+	}
+
+	currentStatefulSet, err := r.sts.Get(ctx, types.NamespacedName{Name: desiredStatefulSet.Name, Namespace: desiredStatefulSet.Namespace})
```
avorima: prefer client.ObjectKey over types.NamespacedName

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	return err
+}
+
+// buildEnvs prepends the mandatory AG runtime env vars (and optional DT_GROUP) to any user-supplied vars.
+func buildEnvs(dk *dynakube.DynaKube) []corev1.EnvVar {
+	connInfoCM := dk.KubernetesMonitoring().GetConnectionInfoConfigMapName()
+
+	required := make([]corev1.EnvVar, 0, requiredEnvsCapacity+len(dk.KubernetesMonitoring().Env))
```
avorima: let's keep this constant local to the function

pkg/api/latest/dynakube/kubemon/props.go
0sewa0: > what's the purpose of combined Spec+Status props when

Other packages do it the same ([example](https://github.com/Dynatrace/dynatrace-operator/blob/1a7b9b5f02b4bc2512eead106b451d00a79dc1c0/pkg/api/latest/dynakube/oneagent/spec.go#L11-L23))

The idea behind this structure is simple, to group all related fields in one.
So you can just get everything  OA related by saying `dk.OneAgent()...` instead of having to find things in `dk.Spec` and `dk.Status` separately.

I am open to rethinking this if we don't like it.

config/helm/chart/default/values.yaml
```diff
@@ -68,6 +68,7 @@ operator:
   startupProbe: {} # configurable: periodSeconds, timeoutSeconds, failureThreshold
   requeueAfter: "" # requeue period for the controller, defaults to 15m
   clientCacheCleanupInterval: "" # defined in the Golang time.Duration format, like "30m" == 30 minutes. Defaults to 1h
+  kubemonOperandEnabled: false # temporary gate; set to true to enable kubemon in dev/test. Remove when kubemon is complete.
```
jdanieck: Any thoughts on this gate? My goal is to merge to main with kubemon disabled by default. Any better ideas how to do it ? Maybe we have some pattern in codebase already for this ?

pkg/controllers/dynakube/kubemon/reconciler.go
```diff
@@ -0,0 +1,148 @@
...
+	Reconcile(ctx context.Context, agClient agclient.Client, dk *dynakube.DynaKube) error
+}
+
+type statefulsetReconciler interface {
+	Reconcile(ctx context.Context, dk *dynakube.DynaKube) error
+}
+
+// Reconciler orchestrates the kubemon operand. Sub-reconciler fields are interfaces so they
+// can be mocked in tests.
+type Reconciler struct {
```
jdanieck, line 52: I noticed during testing that I missed authTokenReconciler - working on it. Currently Kubemon AG is running but it can't properly authorize to the Dynatrace tenant (which is not reflected by any probe unfortunetly, so from k8s perspective it looks OK).

pkg/controllers/dynakube/kubemon/statefulset/reconciler_integration_test.go
```diff
@@ -0,0 +1,267 @@
...
+		desired = *sts.Spec.Replicas
+	}
+
+	sts.Status.ObservedGeneration = sts.Generation
+	sts.Status.Replicas = desired
+	sts.Status.ReadyReplicas = desired
+	require.NoError(t, clt.Status().Update(ctx, sts))
+}
+
+func requireIntegrationStatefulSet(t *testing.T, clt client.Client, dk *dynakube.DynaKube) *appsv1.StatefulSet {
```
avorima: AI went a bit overboard here let's remove this

pkg/controllers/dynakube/kubemon/reconciler.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// Reconcile is the operand entry point called by the parent DynaKube controller.
+// Sub-reconcilers mutate dk.Status.KubernetesMonitoring.* and dk.Status.Conditions in-memory;
+// the parent controller persists status changes via deferred Status().Update().
+func (r *Reconciler) Reconcile(ctx context.Context, dk *dynakube.DynaKube, agClient agclient.Client, _ token.Tokens) error {
+	ctx, log := logd.NewFromContext(ctx, "dynakube-kubemon")
+
+	// Temporary gate, to be removed once kubemon is complete
+	if !k8senv.IsKubemonOperandEnabled() {
```
0sewa0, line 71: cool idea 👍

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
```diff
@@ -0,0 +1,324 @@
...
+		assert.Equal(t, corev1.DNSNone, sts.Spec.Template.Spec.DNSPolicy)
+	})
+
+	t.Run("priority class name", func(t *testing.T) {
+		assert.Equal(t, "high-priority", sts.Spec.Template.Spec.PriorityClassName)
+	})
+
+	t.Run("termination grace period", func(t *testing.T) {
+		require.NotNil(t, sts.Spec.Template.Spec.TerminationGracePeriodSeconds)
+		assert.Equal(t, grace, *sts.Spec.Template.Spec.TerminationGracePeriodSeconds)
```
avorima: assert.Equal can actually compare pointer values, so the nil check isn't needed

pkg/api/latest/dynakube/kubemon/spec.go
```diff
@@ -0,0 +1,121 @@
...
+	PriorityClassName string `json:"priorityClassName,omitempty"`
+
+	// Configures the terminationGracePeriodSeconds parameter of the KubernetesMonitoring pod.
+	// +kubebuilder:validation:Optional
+	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds,omitempty"`
+
+	// Use an emptyDir volume instead of the PersistentVolumeClaim defined in volumeClaimTemplate.
+	// Cached state is lost on pod restart; only enable for transient environments.
+	// +kubebuilder:validation:Optional
+	UseEphemeralVolume bool `json:"useEphemeralVolume,omitempty"`
```
0sewa0: I don't think we need this for kubemon, only for telemetry ingest

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
```diff
@@ -0,0 +1,324 @@
...
+	err := statefulset.NewReconciler(fakeClient).Reconcile(t.Context(), dk)
+	require.Error(t, err)
+	require.NotErrorIs(t, err, k8sstatefulset.ErrRolloutInProgress)
+	require.NotErrorIs(t, err, statefulset.ErrImageRequired)
+	assert.False(t, writeAttempted)
+}
+
+// TestReconcileBuildsStatefulSet covers the shape of the produced StatefulSet. The fake client has
+// no StatefulSet controller, so reconcile always reports rollout in progress.
+func TestReconcileBuildsStatefulSet(t *testing.T) {
```
avorima, line 107: subtests should be self-contained. either create a new statefulset every time or merge the cases

pkg/api/latest/dynakube/kubemon/spec.go
```diff
@@ -0,0 +1,121 @@
...
+	// +kubebuilder:validation:Optional
+	Env []corev1.EnvVar `json:"env,omitempty"`
+
+	// Adds TopologySpreadConstraints for the KubernetesMonitoring pods.
+	// +kubebuilder:validation:Optional
+	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+
+	// Defines storage device.
+	// +kubebuilder:validation:Optional
+	VolumeClaimTemplate *corev1.PersistentVolumeClaimSpec `json:"volumeClaimTemplate,omitempty"`
```
0sewa0: same as previous
> I don't think we need this for kubemon, only for telemetry ingest

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
```diff
@@ -0,0 +1,324 @@
...
+					return errors.New("kube api error")
+				}
+
+				return c.Create(ctx, obj, opts...)
+			},
+		}, dk, newTestTenantSecret(dk))
+
+		err := statefulset.NewReconciler(fakeClient).Reconcile(t.Context(), dk)
+		require.Error(t, err)
+		require.NotErrorIs(t, err, k8sstatefulset.ErrRolloutInProgress)
```
avorima: rather than asserting what this isn't, this should check what it is

0sewa0 (commented):
(still looking at the reconcilers)

config/helm/chart/default/values.yaml
0sewa0: [I like it](https://github.com/Dynatrace/dynatrace-operator/pull/6757#discussion_r3551278600) 😄

pkg/controllers/dynakube/kubemon/reconciler.go
```diff
@@ -0,0 +1,148 @@
+// Package kubemon reconciles the dedicated Kubernetes Monitoring operand.
+//
+// Non-obvious behavior:
+//   - A single condition (KubernetesMonitoringAvailable) is owned by this orchestrator.
+//   - kubemon uses a single Kubernetes client dependency; caller chooses the concrete
+//     implementation (cached or direct)
+//   - Condition mapping is based on sub-reconciler errors:
+//   - nil => True/Available
+//   - persistent sentinels => False/Error
+//   - all other errors => Unknown/Reconciling
```
avorima: i must've overlooked this in the design doc, because this is very non-standard

pkg/controllers/dynakube/kubemon/reconciler.go
```diff
@@ -0,0 +1,148 @@
...
+func (r *Reconciler) setConditionByError(dk *dynakube.DynaKube, err error) {
+	condition := metav1.Condition{
+		Type: kubemonapi.KubeMonAvailableConditionType,
+	}
+
+	switch {
+	case isPersistent(err):
+		condition.Status = metav1.ConditionFalse
+		condition.Reason = reasonError
+		condition.Message = err.Error()
```
avorima: use Unwrap to prevent accidentally putting stacktraces here

pkg/controllers/dynakube/kubemon/reconciler.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// Temporary gate, to be removed once kubemon is complete
+	if !k8senv.IsKubemonOperandEnabled() {
+		log.Debug("kubemon gate not enabled, skipping")
+
+		return nil
+	}
+
+	// Fast-path guard: never created, nothing to converge.
+	if !dk.KubernetesMonitoring().IsEnabled() && !hasCondition(dk) {
```
avorima: conditions are signals, not hard facts. reconciliation should always run

pkg/controllers/dynakube/phase.go
```diff
@@ -93,14 +103,8 @@ func (controller *Controller) determineStatefulSetPhase(ctx context.Context, dk
...
-	scheduledReplicas := int32(0)
-	if statefulSet.Spec.Replicas != nil {
-		scheduledReplicas = *statefulSet.Spec.Replicas
-	}
-
-	if statefulSet.Generation != statefulSet.Status.ObservedGeneration ||
-		scheduledReplicas != statefulSet.Status.ReadyReplicas {
-		log.Info("statefulset is still deploying", "statefulset", statefulsetName)
+	if !k8sstatefulset.IsRolloutComplete(statefulSet) {
+		log.Info("statefulset is still deploying", "dynakube", dk.Name, "statefulset", statefulsetName)
```
avorima:
```suggestion
		log.Info("statefulset is still deploying", "statefulset", statefulsetName)
```

pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -23,6 +23,7 @@ const (
 	AppVersion                  = "APP_VERSION"
 
 	DTExtractCodeModulesImageLinksEnvVar = "DT_EXTRACT_CODEMODULES_IMAGE_LINKS"
+	KubemonOperandEnabled                = "KUBEMON_OPERAND_ENABLED"
```
avorima: nit: flags should have the form "do stuff" instead of "stuff done" -> enable-foo instead of foo-enabled

pkg/api/latest/dynakube/kubemon/props.go
jokuniew: ahh right..  i see it now in oa/ag/kspm 
If no method will be created for status, this struct can be slim down like logmon/extensions
closing thread cause im blind 🙈

pkg/util/kubernetes/objects/k8sstatefulset/builder.go
```diff
@@ -124,6 +126,48 @@ func SetRollingUpdateStrategyType() builder.Option[*appsv1.StatefulSet] {
 	}
 }
 
+func SetVolumes(volumes []corev1.Volume) builder.Option[*appsv1.StatefulSet] {
```
avorima: if there's already an alias defined the functions here should also use it

pkg/util/kubernetes/objects/k8sstatefulset/rollout_test.go
avorima, line 1: don't need an extra file for this. move to statefulset_test.go

pkg/controllers/dynakube/controller.go
```diff
@@ -395,6 +405,20 @@ func (controller *Controller) reconcileComponents(ctx context.Context, dtClient
...
 	}
 
+	log.Debug("start reconciling KubernetesMonitoring")
+
+	if err := controller.kubemonReconciler.Reconcile(ctx, dk, dtClient.ActiveGate, controller.tokens); err != nil {
+		if errors.Is(err, k8sstatefulset.ErrRolloutInProgress) {
+			controller.setRequeueAfterIfNewIsShorter(fastRequeueInterval)
+
+			return goerrors.Join(componentErrors...)
+		}
```
mihaitanasedt, line 418: Why returning early? It will block the reconciliation of other components

pkg/controllers/dynakube/kubemon/reconciler.go
mihaitanasedt, line 71: Wouldn't this block cleanup and keep orphans in the cluster? Let's say we have it enabled via env, and then we disable,  it will never cleanup the resources

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
0sewa0: feels a bit over-optimizing, considering we have a conditional env

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -0,0 +1,237 @@
...
+		opts = append(opts, k8sstatefulset.SetDNSPolicy(km.DNSPolicy))
+	}
+
+	if km.PriorityClassName != "" {
+		opts = append(opts, k8sstatefulset.SetPriorityClassName(km.PriorityClassName))
+	}
+
+	if km.TerminationGracePeriodSeconds != nil {
+		opts = append(opts, k8sstatefulset.SetTerminationGracePeriodSeconds(km.TerminationGracePeriodSeconds))
+	}
```
0sewa0: couldn't the `nil` / empty values be handled within the options?

pkg/controllers/dynakube/kubemon/reconciler.go
```diff
@@ -0,0 +1,148 @@
...
+	log.Debug("reconciled kubernetes monitoring")
+
+	return nil
+}
+
+func hasCondition(dk *dynakube.DynaKube) bool {
+	return meta.FindStatusCondition(*dk.Conditions(), kubemonapi.KubeMonAvailableConditionType) != nil
+}
+
+func (r *Reconciler) setAvailableCondition(dk *dynakube.DynaKube) {
```
0sewa0: bit confusing name as it also removes the condition when things got disabled 😅

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler.go
```diff
@@ -0,0 +1,109 @@
...
+	if err := r.createOrUpdateConfigMap(ctx, dk, info); err != nil {
+		return err
+	}
+
+	if err := r.createOrUpdateSecret(ctx, dk, info.TenantToken); err != nil {
+		return err
+	}
+
+	dk.Status.KubernetesMonitoring.ConnectionInfo.TenantUUID = info.TenantUUID
+	dk.Status.KubernetesMonitoring.ConnectionInfo.Endpoints = info.Endpoints
```
mihaitanasedt, line 57: Why about the tenant token hash?

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -0,0 +1,237 @@
...
+		ctx,
+		r.kubeClient,
+		types.NamespacedName{Name: dk.KubernetesMonitoring().GetStatefulSetName(), Namespace: dk.Namespace},
+		dk.KubernetesMonitoring().Replicas,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	tokenHash, err := r.getTenantTokenHash(ctx, dk)
```
mihaitanasedt, line 162: The other AG SS reconciler reads it from status. Why is here different?

mihaitanasedt (dismissed):
(just some observations, will continue reviewing)

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
jdanieck: AG cached the tenant token hash in status to avoid re-reading the Secret on every reconcile, back when there was no [client-side caching](https://github.com/Dynatrace/dynatrace-operator/pull/6757#discussion_r3490858941). kubemon skips that: both clients are cached now, so live reads are cheap — simpler, and no derived field to keep in sync.

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler.go
jdanieck: Since we don't use it for [caching anymore](https://github.com/Dynatrace/dynatrace-operator/pull/6757#discussion_r3552746317) I didn't populate as I don't see [any other reason to keep it](https://github.com/Dynatrace/dynatrace-operator/pull/6757#discussion_r3490763245) - if there is, please tell me, I can populate, just need to understand the usecase.

pkg/controllers/dynakube/controller.go
jdanieck: you are right, it was a mistake - fixed in https://github.com/Dynatrace/dynatrace-operator/pull/6757/changes/3a07764e0f3faf3c36b750f1918dc2e4dab85638

pkg/controllers/dynakube/kubemon/reconciler.go
jdanieck: Yes, this is intentional. The flag semantics are straightforward: `on` = kubemon reconciler active, `off` = reconciler disabled.

It's a temporary developer-facing escape hatch for the period when kubemon is merged to main but not yet fully functional. Once the operand is complete, the flag will be removed.

This is a common pattern in [trunk-based development](https://www.atlassian.com/continuous-delivery/continuous-integration/trunk-based-development) — merge early, gate behind a flag, clean up when ready.

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler.go
mihaitanasedt: I would say to stay consistent with the other active gate builder. Since they both have a very similar task, building an AG SS, they should have similar logic if they use the same inputs.

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
mihaitanasedt: Let's continue in other comment thread 
https://github.com/Dynatrace/dynatrace-operator/pull/6757#discussion_r3551767210

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler.go
jdanieck: Consistency with a pattern whose motivation doesn't apply here isn't a good enough reason to copy it. New code is the right place to establish the better pattern; we can align the old one separately.

0sewa0: TLDR: pls don't add tenant token hash to the status 🙏 

The tenant-token hash in the status only had 1 purpose:
- To "cache" this value between the reconciler**s**

Otherwise, it is just noise in the status.

Jozef is using only the `client.Client` (so the one that caches stuff by default), so when he does something like https://github.com/Dynatrace/dynatrace-operator/pull/6757/changes#diff-756e31856c97afbaa0c484e6ee462407e573603d4cb17e2793775e54084f1823R225, it is fine as the Secret is cached by the client and not our status.
- lets hope our superstition about preferring to use the (not caching) `client.Reader` is soon gone 🤞 

Sidenote:
The current AG reconcilers / logic is FAAAAR from optimal, they were planned to be refactored several times over the years. So the less inspirations you take from them, the better. 😬

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
jdanieck: you are right, I'll refactor it; this preallocation was done before i have the conditional env and linter forced me to do it

```sh
make go/lint
/Users/jozef.daniecki/GolandProjects/dynatrace-operator/bin/golangci-lint run --fix
pkg/controllers/dynakube/kubemon/statefulset/reconciler.go:73:2: Consider preallocating required with capacity 3 + len(dk.KubernetesMonitoring().Env) (prealloc)
        required := []corev1.EnvVar{
        ^
1 issues:
* prealloc: 1
make: *** [go/golangci] Error 1
```

jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/changes/7cf7a3069315147aabe3ffc712dc79ffe545f010

jdanieck: the constant was removed in https://github.com/Dynatrace/dynatrace-operator/commit/7cf7a3069315147aabe3ffc712dc79ffe545f010 because of https://github.com/Dynatrace/dynatrace-operator/pull/6757/files/cafefe3b9eaa18b363f8786f0114c4867a0cb6d9#r3551618901

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_integration_test.go
jdanieck: The pattern was generalized from the statefulset reconciler, where the reconciler reads state it previously wrote via the cached client - so one reconcile isn't always sufficient. Since this is the first time we use a cached client in integration tests, I wanted to establish a uniform pattern to avoid coupling each test to assumptions about its reconciler's cache behaviour. 
 
I don't have a strong opinion here, but it would be good to reach agreement and document it in the coding style guide to avoid the same conversation in the future.

pkg/controllers/dynakube/kubemon/reconciler.go
0sewa0: > reconciliation should always run

is that realistic considering the size of the DynaKube?

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler.go
mihaitanasedt: Agree, best is to refactor the old bad pattern.

Was concerned about increasing complexity 😬

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_integration_test.go
jdanieck: We had an offline discussion with following conclusion:

- use _Eventually_ only when strictly needed (see examples bellow)
- keep caching client in integration tests and eventually refactor all integration tests to use it 

Examples:
image

pkg/controllers/dynakube/kubemon/reconciler.go
avorima: yes. that's the sub reconcilers are for

jdanieck: I followed the AG convention. Need to rely on your judgment whether this is premature or justified optimization.

pkg/controllers/dynakube/kubemon/reconciler.go
jdanieck: The design doc is not written in stone, we can update it if needed. I tried to reflect what the [API conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) advise. 

What standard mean to you ? Use 'False' instead of 'Unknown' for transient state? 

From design:
 - True/Available — all resources healthy                                                                                                                                                                                                                           
  - Unknown/Reconciling — provisioning or pods not yet ready                                                                                                                                                                                                         
  - False/Error — persistent failure 

Relevant quotes:

> Condition type names should describe the current observed state of the
  resource, rather than describing the current state transitions. This
  typically means that the name should be an adjective ("Ready", "OutOfDisk")
  or a past-tense verb ("Succeeded", "Failed") rather than a present-tense verb
  ("Deploying"). Intermediate states may be indicated by setting the `status` of
  the condition to `Unknown`

> Controllers should apply their conditions to a resource the first time they
  visit the resource, even if the `status` is Unknown. This allows other
  components in the system to know that the condition exists and the controller
  is making progress on reconciling that resource.

> Not all controllers will observe the previous advice about reporting
     "Unknown" or "False" values. For known conditions, the absence of a
     condition `status` should be interpreted the same as `Unknown`, and
     typically indicates that reconciliation has not yet finished (or that the
     resource state may not yet be observable).

pkg/controllers/dynakube/kubemon/reconciler.go
jdanieck: removed in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/3a76740bc72a0e4b4006f2d7eff98c73495a9de5

pkg/controllers/dynakube/kubemon/reconciler.go
jdanieck: renamed to `reconcileAvailableCondition` in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/8af95255339b4bdb35ec01d8ab505a2df3a07583

pkg/controllers/dynakube/kubemon/reconciler.go
avorima: "Intermediate states may be indicated by setting the status of the condition to Unknown" is somewhat at odds with "ConditionUnknown means kubernetes can't decide if a resource is in the condition or not" ([link](https://pkg.go.dev/k8s.io/api/core/v1#ConditionStatus)), depending on how you define intermediate states. To give an example why this seems non-standard to me: a node condition unknown is a bigger red flag than a node condition false. The same holds for most other controllers.

0sewa0: Is there any realistic scenario where we will be in the `Reconciling` state? (`Unknown` as a state is really worrying as Mario mentioned, so we should avoid it)

From my understanding, this part of the code doesn't actually check if the Pods of the Statefulset is ready or not. That is done "outside", in the DynaKube controller.
- So this will not cause a `Reconciling` state

And if any other sub-reconcilers fail here, they should be an error.
- Connection info, for the AG, should always return/setup something, otherwise it is an error.
- Statefulset creation is similar, so statefulset was successfully created or not

In the future when there will be more sub-reconcilers here, it might make sense to have this "waiting" state, but not really now.

pkg/api/latest/dynakube/kubemon/status.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/0245acd7a1b9b8b57f412f4eb063711284ad38df

pkg/controllers/dynakube/kubemon/statefulset/reconciler_integration_test.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/10679ce4dde11aa67ff2d8dff4c1ebe792fcffd8

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/3c386c6d4610908ff45638e2a5e888434c6fecde

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/5f7a4ae9c7f12a69cb095b6540aff553ae81cce1

pkg/util/kubernetes/objects/k8sstatefulset/builder.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/5f7a4ae9c7f12a69cb095b6540aff553ae81cce1

pkg/util/kubernetes/objects/k8sstatefulset/rollout_test.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/629bf0ab3172acc8af61a3afc8286c96599ee5a5

pkg/util/kubernetes/objects/k8sstatefulset/builder.go
avorima: we now have a mix. would be good to be consistent in the package. please make that fly-by change

jdanieck: Fair point, it's a simple change — but touching existing code tends to snowball and I'd rather keep this PR focused.

pkg/util/kubernetes/fields/k8senv/env.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/2f7e5cd4389f5834086fbdf9993814ecfb63586f

avorima: update the helm value as well to match naming

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_integration_test.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/2a596134d0e333b1a6d50c3463b4ac27b387ac87

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_integration_test.go
jdanieck: I use [Eventually](https://pkg.go.dev/github.com/stretchr/testify/assert#Eventually) from testify and I don't see `Consistently` like equivalent in it; pulling in gomega just for this check feels like an overkill, hence I used simple `for` instead in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/2a596134d0e333b1a6d50c3463b4ac27b387ac87

pkg/util/kubernetes/fields/k8senv/env.go
jdanieck: updated in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/70ddbc3c0e38678908535c0077ca0bfbfccd8141

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_integration_test.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/bcb8601ffcfb10b0e79aa7185174be7abefd0e5b

test/integrationtests/environment.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/8dc03906ebcef3beab30fa4e6a65b9d706133eee

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_test.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/af67ff7ade8f5c53d60569e394b24ed928f20e9d

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_test.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/af67ff7ade8f5c53d60569e394b24ed928f20e9d

pkg/controllers/dynakube/kubemon/reconciler.go
jdanieck: @avorima @0sewa0 after we resolve all open comments, would you be ok merging this PR only with connectionInfo & statefulset reconcilers ? I think it'll make review easier if authTokenReconciler go in separate PR. No strong opinion though, what you think ?

avorima: the PR is big enough as it is. i think it's fine to not have a fully working solution at the start

pkg/api/latest/dynakube/kubemon/props.go
```diff
@@ -0,0 +1,61 @@
...
+func (km *Spec) GetServiceAccountName() string {
+	return ServiceAccountName
+}
+
+func (km *KubeMon) GetStatefulSetName() string {
+	return km.name + OperandNameSuffix
+}
+
+func (km *KubeMon) GetConnectionInfoConfigMapName() string {
+	return km.name + "-kubemon-connection-info"
```
aorcholski:
```suggestion
	return km.name + OperandNameSuffix +"-connection-info"
```
I think it should be always consistent with OperandName

pkg/api/latest/dynakube/kubemon/props.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (km *KubeMon) GetTenantSecretName() string {
+	return km.name + "-kubemon-tenant-secret"
```
aorcholski:
```suggestion
	return km.name + OperandNameSuffix +"-tenant-secret"
```
I think it should be always consistent with OperandName

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -0,0 +1,219 @@
...
+func buildVolumeMounts(dk *dynakube.DynaKube) []corev1.VolumeMount {
+	km := dk.KubernetesMonitoring()
+	mounts := []corev1.VolumeMount{{
+		Name:      connectioninfo.TenantSecretVolumeName,
+		ReadOnly:  true,
+		MountPath: connectioninfo.TenantTokenMountPoint,
+		SubPath:   connectioninfo.TenantTokenKey,
+	}}
+
+	if km.UseEphemeralVolume || km.VolumeClaimTemplate != nil {
```
aorcholski: see https://github.com/Dynatrace/dynatrace-operator/pull/6757/changes#r3551300610
EmptyDir should be mounted to `/var/tmp/dynatrace/gateway`

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -0,0 +1,219 @@
...
+		Name: connectioninfo.TenantSecretVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName:  km.GetTenantSecretName(),
+				DefaultMode: new(int32(0o640)),
+			},
+		},
+	}}
+
+	if km.UseEphemeralVolume {
```
aorcholski: see https://github.com/Dynatrace/dynatrace-operator/pull/6757/changes#r3551300610
EmptyDir should be mounted to `/var/tmp/dynatrace/gateway`

pkg/controllers/dynakube/kubemon/reconciler.go
jdanieck: @avorima thanks for the link, let's go with `False` instead of `Unknown`. 

@0sewa0 yes, we check [rollout](https://github.com/Dynatrace/dynatrace-operator/pull/6757/files#diff-756e31856c97afbaa0c484e6ee462407e573603d4cb17e2793775e54084f1823R58) in statefulset reconciler. We have only one condition named `KubernetesMonitoringAvailable` which represent operand as a whole, hence I felt it should be `True` only when the operand is fully deployed. 

New (simplified) code in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/a023c2c469b979d09d03b0743579435859cdfb66

pkg/controllers/dynakube/kubemon/reconciler.go
jdanieck: Unwrap returns nil on non-wrapped errors (like our sentinels), hence I used `pkgerrors.Cause` instead; done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/a023c2c469b979d09d03b0743579435859cdfb66

aorcholski (commented):
rbac.kubernetes-monitoring disabled + AG.capablity.kubernetes-monitoring -> rejected
```
Error from server (Forbidden): error when creating "../dynakube/dynakube-cn3.yaml": admission webhook "v1beta6.dynakube.webhook.dynatrace.com" denied the request: 
 1. KubernetesMonitoring has been disabled during Operator install. The necessary resources for KubernetesMonitoring to work are not present on the cluster. Redeploy the Operator via Helm with all the necessary resources enabled.
```

FYI however no validation for spec.kubernetesMonitoring

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
jdanieck: done in [dbcd208](https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/dbcd20817d65080d84a22a05d79e3845fd97ec5a)

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
jdanieck: done in [dbcd208](https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/dbcd20817d65080d84a22a05d79e3845fd97ec5a)

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
jdanieck: done in [dbcd208](https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/dbcd20817d65080d84a22a05d79e3845fd97ec5a)

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_test.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/734ca12a48efb05e3a19d61752aa76c9ac26544f

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler_test.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/commit/734ca12a48efb05e3a19d61752aa76c9ac26544f

pkg/controllers/dynakube/kubemon/reconciler.go
```diff
@@ -0,0 +1,122 @@
...
+	// Temporary gate, to be removed once kubemon is complete
+	if !k8senv.IsKubemonOperandEnabled() {
+		log.Debug("kubemon gate not enabled, skipping")
+
+		return nil
+	}
+
+	log.Debug("reconciling kubernetes monitoring")
+
+	defer r.reconcileCondition(dk, &err)
```
avorima: capturing the error in a closure circumvents the need of a pointer to interface

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler.go
```diff
@@ -0,0 +1,112 @@
...
+	cm, err := k8sconfigmap.Build(dk,
+		dk.KubernetesMonitoring().GetConnectionInfoConfigMapName(),
+		data,
+		k8sconfigmap.SetLabels(coreLabels.BuildLabels()),
+	)
+	if err != nil {
+		return pkgerrors.WithStack(err)
+	}
+
+	_, err = r.configMaps.CreateOrUpdate(ctx, cm)
```
avorima, line 95: all resources should be owned by the dynakube

pkg/api/latest/dynakube/kubemon/spec.go
jdanieck: Does it even make sense to introduce a deprecated field at all?

avorima: We may have to amend the design doc here. That's a valid point and it probably does not make sense to add a deprecated field

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/2952cd1f4172de0cd7f3b6873270e56555bcbe36

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
jdanieck: done in done in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/2952cd1f4172de0cd7f3b6873270e56555bcbe36

pkg/api/latest/dynakube/kubemon/spec.go
jdanieck: removed in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/2952cd1f4172de0cd7f3b6873270e56555bcbe36

pkg/api/latest/dynakube/kubemon/spec.go
jdanieck: removed in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/2952cd1f4172de0cd7f3b6873270e56555bcbe36

pkg/api/latest/dynakube/kubemon/spec.go
jdanieck: Let's keep it as is for now and have this conversation as part of ICP-7297.

pkg/controllers/dynakube/kubemon/reconciler.go
jdanieck: refactored in https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/24c62bc34d3a52bad6eef7171041bf48743ffc52

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler.go
jdanieck: they should be already, what did I miss ?

I build the objects with [builder](https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/util/kubernetes/objects/k8sconfigmap/builder.go#L20) helper which set owner ref internally; also it is asserted in [integration tests](https://github.com/Dynatrace/dynatrace-operator/pull/6757/files#diff-b2e76683eb7ab4325ef9108cb076e51ee676980b46b9b628a3651df14d446db4R112)
 
```
❯ k get cm dev-dk-kubemon-connection-info -o json | jq .metadata.ownerReferences
[
  {
    "apiVersion": "dynatrace.com/v1beta6",
    "blockOwnerDeletion": true,
    "controller": true,
    "kind": "DynaKube",
    "name": "dev-dk",
    "uid": "ccdd9444-355d-472c-9407-9fb5d9bf0b7a"
  }
]
```

jdanieck:
@0sewa0  @aorcholski  @avorima  @jokuniew  @mihaitanasedt thank you for all great feedback ! All comments are either resolved or answered. PR ready for second round of review. Ideally, I'd like to merge it till end of the week.

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler.go
avorima: the mixed use of WithOwner is confusing

jdanieck: I'll align; what is the prefered way ?

avorima:
PR looks good to merge. One thing I found is that the apparmor annotation mapping is missing (https://github.com/Dynatrace/dynatrace-operator/pull/6431)

aorcholski (dismissed):
LGTM
securityContexts are missing but I assume will be implemented in the following PRs

jdanieck:
The [TestReconcileLifecycle](https://github.com/Dynatrace/dynatrace-operator/actions/runs/29504240911/job/87640800363?pr=6757#step:4:329) seems to be flaky, won't be merging until I reproduce and get the fix in.

avorima (approved):
tried reproducing it yesterday, but i never got this caching issue

jdanieck:
> tried reproducing it yesterday, but i never got this caching issue

@avorima @0sewa0 I also tried hard to reproduce [this failure](https://github.com/Dynatrace/dynatrace-operator/actions/runs/29504240911/job/87640800363?pr=6757#step:4:329) locally yesterday, but without success. The logs make the cause clear, though, so I applied a fix based on them in [a0a0fb6](https://github.com/Dynatrace/dynatrace-operator/pull/6757/commits/a0a0fb6f152134f732431a1b42e43a8cc8ce0669).

That said, these tests are clearly error-prone to author. I took a shot at simplifying that in #6972. We still have to reason about cache behaviour to get the tests right — but maybe that's a good thing: it forces us to understand how the code behaves in production and to confirm whether that behaviour is intentional.

PR #6758, opened 2026-05-26
Add NodeSelector to OtelC in DynaKube

waodim:
## Description

https://dt-rnd.atlassian.net/browse/ICP-1087

Thid PR adds a `nodeSelector` field to `.spec.templates.otelCollector`, mirroring the option already available on other components (ActiveGate, OneAgent, KSPM, etc.). The reconciler now propagates it to the OTel collector StatefulSet's pod spec (sts.spec.template.spec.nodeSelector).

## How can this be tested?

- manual deployment and apply a DK with `spec.templates.otelCollector.nodeSelector` and confirm the resulting STS's pod spec has the nodeSelector set. (nice-to-have: test it on GKE-autopilot with `nodeSelector` + `tolerations` -> this combination should actually pin Otel pod to the targeted node pool)
- Unit tests


pkg/api/v1beta4/dynakube/extensions.go
```diff
@@ -79,6 +79,10 @@ type OpenTelemetryCollectorSpec struct {
 	// +kubebuilder:validation:Optional
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 
+	// Node selector to control the selection of nodes for the OtelCollector pods
+	// +kubebuilder:validation:Optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
0sewa0: why put it into old versions?

should only be in latest

waodim: misconception on my side. Actually happy to not add it to old versions 😅 

will update

waodim: https://github.com/Dynatrace/dynatrace-operator/pull/6758/commits/6d33f7511c6a293b1f1893f58f2e3274509794ec

0sewa0 (approved):

PR #6762, opened 2026-05-26
Validate env refs in resource attributes

Incredible-O:
## Description
Our pod mutating webhook builds `OTEL_RESOURCE_ATTRIBUTES` from values that include namespace annotations. In `sanitizeValue`, env refs like `$(SOME_VAR)` are passed through unencoded (and expanded by the kubelet), as long as they start with `$(` and end with `)`.
Sadly, this also includes refs like `$(DT_API_TOKEN)`. Improve the validation by introducing an allowlist for legit ref names.

Addresses [DTSEC-25207](https://dt-rnd.atlassian.net/browse/DTSEC-25207).

## How can this be tested?
1. Set `TENANT` in the script.
2. Run the script before and after applying the change.
3. Compare the before/after values of `OTEL_RESOURCE_ATTRIBUTES`.
<details>
<summary>Repro script</summary>

```sh
#!/usr/bin/env bash

set -euo pipefail

NAMESPACE=env-ref-repro
TENANT=<some-tenant>

kubectl apply -f - <<EOF
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
spec:
  apiUrl: https://${TENANT}.dev.dynatracelabs.com/api
  otlpExporterConfiguration:
    signals:
      traces: {}
      metrics: {}
      logs: {}
  metadataEnrichment:
    enabled: true
EOF

echo "Waiting for dynakube-otlp-exporter-config secret..."
for _ in {1..60}; do
  kubectl get secret dynakube-otlp-exporter-config -n dynatrace >/dev/null 2>&1 && break
  sleep 2
done
kubectl get secret dynakube-otlp-exporter-config -n dynatrace >/dev/null

kubectl create namespace "$NAMESPACE"
kubectl label namespace "$NAMESPACE" dynakube.internal.dynatrace.com/instance=dynakube
kubectl annotate namespace "$NAMESPACE" 'metadata.dynatrace.com/exfil=$(DT_API_TOKEN)'

kubectl apply -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: sample-app
  namespace: $NAMESPACE
spec:
  restartPolicy: Never
  containers:
  - name: app
    image: busybox:1.36
    command: ["sleep", "3600"]
EOF

kubectl wait --for=condition=Ready pod/sample-app -n "$NAMESPACE" --timeout=60s
kubectl exec sample-app -n "$NAMESPACE" -- env | grep 'OTEL_RESOURCE'
```
</details>

[DTSEC-25207]: https://dt-rnd.atlassian.net/browse/DTSEC-25207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/webhook/mutation/pod/attributes/keys.go
+++ b/pkg/webhook/mutation/pod/attributes/keys.go
@@ -18,3 +18,9 @@ const (
 	K8sWorkloadKindAttr = "k8s.workload.kind"
 	K8sWorkloadNameAttr = "k8s.workload.name"
 )
+
+var SafeEnvRefs = []string{
+	K8sPodNameEnv,
+	K8sPodUIDEnv,
+	K8sNodeNameEnv,
+}
--- a/pkg/webhook/mutation/pod/attributes/keys.go
+++ b/pkg/webhook/mutation/pod/attributes/keys.go
@@ -1,5 +1,7 @@
 package attributes
 
+import "github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8senv"
+
 const (
 	K8sContainerNameAttr = "k8s.container.name"
 	K8sNodeNameEnv       = "K8S_NODE_NAME"
@@ -20,7 +22,7 @@ const (
 )
 
 var SafeEnvRefs = []string{
-	K8sPodNameEnv,
-	K8sPodUIDEnv,
-	K8sNodeNameEnv,
+	k8senv.NewRef(K8sPodNameEnv),
+	k8senv.NewRef(K8sPodUIDEnv),
+	k8senv.NewRef(K8sNodeNameEnv),
 }
--- a/pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/attributes.go
+++ b/pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/attributes.go
@@ -40,9 +40,10 @@ func isSafeEnvRef(value string) bool {
 	return found && slices.Contains(attributes.SafeEnvRefs, before)
 }
 
+// sanitizeValue percent-encodes value for use in OTEL_RESOURCE_ATTRIBUTES, except for the
+// small set of operator-injected env refs that the kubelet must expand at pod startup.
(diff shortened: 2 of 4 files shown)


pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/attributes.go
```diff
@@ -27,10 +29,21 @@ func NewAttributesFromEnv(envs []corev1.EnvVar, name string) (map[string]string,
...
+		return false
+	}
+
+	before, found := strings.CutSuffix(after, ")")
+
+	return found && slices.Contains(attributes.SafeEnvRefs, before)
+}
+
 func sanitizeValue(value string) string {
 	// apply percent encoding to prevent errors when passing attribute values with special characters to the OTEL SDKs
```
avorima: this comment looks a bit out of place above this check

pkg/webhook/mutation/pod/attributes/keys.go
```diff
@@ -18,3 +18,9 @@ const (
 	K8sWorkloadKindAttr = "k8s.workload.kind"
 	K8sWorkloadNameAttr = "k8s.workload.name"
 )
+
+var SafeEnvRefs = []string{
+	K8sPodNameEnv,
```
avorima: given that we have a very limited subset here, we could just use the actual values. makes comparison a bit simpler
```suggestion
	k8senv.NewRef(K8sPodNameEnv),
```

pkg/webhook/mutation/pod/mutator/otlp/resourceattributes/attributes.go
Incredible-O: Addressed in 27ab1919d

pkg/webhook/mutation/pod/attributes/keys.go
Incredible-O: Simplified in c4c378f3b

avorima (approved):

PR #6473, opened 2026-04-15
feat: logging: migrate to context loggers

svfoxat:
## Description

https://dt-rnd.atlassian.net/browse/ICP-2916

- migrates public utility APIs to accept `context.Context` instead of an explicit `logd.Logger` parameter
- removes `var log` package-level globals from migrated packages; callers pass `ctx` instead of a logger

> Assisted-by: Claude:claude-sonnet-4.6

### Stack
- PR 1: https://github.com/Dynatrace/dynatrace-operator/pull/6472
- **PR 2 (this):** https://github.com/Dynatrace/dynatrace-operator/pull/6473
- ~~PR 3: https://github.com/Dynatrace/dynatrace-operator/pull/6474~~ (merged into *PR2*) 

## How can this be tested?
`make go/test`

PR #6766, opened 2026-05-27
chore: reduce amount of small files

svfoxat:
## Description

this is a followup to: https://github.com/Dynatrace/dynatrace-operator/pull/6473
https://dt-rnd.atlassian.net/browse/ICP-1043

- added a section in coding guidelines (feedback wanted!)
- moved `const` and `var` and some `interface` declarations from their own (small) files to the files where they are used.

## How can this be tested?

`make go/lint`
`make go/test`


0sewa0:
what is the state here?

pkg/api/latest/dynakube/conversion.go
```diff
@@ -1,7 +0,0 @@
-// Copyright Dynatrace LLC
```
andriisoldatenko, line 1: i would keep this file, cc @avorima wdyt?

pkg/api/v1alpha2/edgeconnect/conversion.go
```diff
@@ -1,7 +0,0 @@
-// Copyright Dynatrace LLC
```
andriisoldatenko, line 1: same here.

pkg/api/v1alpha2/edgeconnect/edgeconnect_types.go
```diff
@@ -134,6 +136,10 @@ type EdgeConnectList struct { //nolint:revive
 }
 
 const (
-	KubernetesDefaultDNS     = "kubernetes.default.svc.cluster.local"
-	kubernetesHostnameSuffix = "kubernetes-automation"
+	KubernetesDefaultDNS       = "kubernetes.default.svc.cluster.local"
+	kubernetesHostnameSuffix   = "kubernetes-automation"
+	DefaultMinRequestThreshold = 15 * time.Minute
```
andriisoldatenko: is it related/or used in reconciler? can we move it there?

pkg/api/latest/dynakube/conversion.go
svfoxat: I wasn't really sure about the merges in `/pkg/api`, functionally it should make no difference. I'm happy to revert them to before if we want to keep the separation there.

pkg/api/v1alpha2/edgeconnect/edgeconnect_types.go
svfoxat: moved the `DefaultMinRequestThreshold` to the file where it is used 👍

pkg/api/latest/dynakube/otlpexporterconfiguration_props.go
```diff
@@ -1,10 +0,0 @@
-// Copyright Dynatrace LLC
-// SPDX-License-Identifier: Apache-2.0
-
-package dynakube
-
-import "github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube/otlp"
-
-func (dk *DynaKube) OTLPExporterConfiguration() *otlp.ExporterConfiguration {
```
0sewa0, line 8: We could move all "sub section" into one file. There are files for OneAgent/Logmon/.../
- otlp is only different in 1 way, that it is simple :D 

like "oprands_props.go"

wdyt?

pkg/api/latest/dynakube/dynakube_props.go
```diff
@@ -192,3 +193,10 @@ func (dk *DynaKube) GetResourceAttributes() map[string]string {
 func (dk *DynaKube) PublicRegistryOverride() string {
 	return dk.Spec.PublicRegistryOverride
 }
+
+// Hub tags this version as the 'source' of the conversion for controller runtime.
+func (*DynaKube) Hub() {}
```
0sewa0: you put this into the `_types.go` for EdgeConnect

which is IMO more correct.


related to https://github.com/Dynatrace/dynatrace-operator/pull/6766/changes#r3674201968 (cc @andriisoldatenko)

I think `conversion.go` would make some sense, if we didn't always convert to the latest version.
Then it would be an easier way to figure out which version is the `Hub` just by looking at the folder structure.
- Even this is kinda trying really hard to give a purpose for it being a separate file 😅
- If this was our actual intention, then calling it `hub.go` would be more fitting 😬

andriisoldatenko: ok with me to put to `*_types.go` but please don't put to `*_props.go`

pkg/api/latest/dynakube/conversion.go
avorima: we should only keep small files if they are managed by code generation tools. i think in this case we can remove the files. we have a ticket in the backlog to revisit improving tooling, e.g. conversion-gen, integration

andriisoldatenko: >we should only keep small files if they are managed by code generation tools. 

why? can you please argument you idea?

pkg/api/latest/dynakube/dynakube_props.go
andriisoldatenko: @avorima i think if we move Hub to *_types.go then we can remve conversion.go, but sometimes i still prefer to keep small files if them helps with readability.

pkg/api/v1alpha2/edgeconnect/edgeconnect_types.go
```diff
@@ -137,3 +137,6 @@ const (
 	KubernetesDefaultDNS     = "kubernetes.default.svc.cluster.local"
 	kubernetesHostnameSuffix = "kubernetes-automation"
 )
+
+// Hub tags this version as the 'source' of the conversion for controller runtime.
+func (*EdgeConnect) Hub() {}
```
andriisoldatenko: i think it should be on the top?
wdyt @avorima ?

pkg/api/latest/dynakube/conversion.go
avorima: because the files will just be recreated if we merge them somewhere else. but we don't really use the kubebuilder or upstream codegen tooling, so it's fine

andriisoldatenko: u didn't get the point, my idea is sometimes keeping small files helps with readability. In this case i'm fine to remove it.

pkg/api/latest/dynakube/otlpexporterconfiguration_props.go
svfoxat: We can do that, `operands_props.go` is imo scoped enough to not fall under general `props` or `utils`.

> otlp is only different in 1 way, that it is simple :D

I'd move `OtelCollectorStatefulsetName` and two boolean checks (`IsAGCertificateNeeded`, *IsCACertificateNeeded*) to `dynakube_props.go`  cause they don't really follow the "operand" pattern like the others. 
wdyt 6135268

pkg/api/latest/dynakube/dynakube_props.go
```diff
@@ -195,3 +197,27 @@ func (dk *DynaKube) GetResourceAttributes() map[string]string {
 func (dk *DynaKube) PublicRegistryOverride() string {
 	return dk.Spec.PublicRegistryOverride
 }
+
+func (dk *DynaKube) OTLPExporterConfiguration() *otlp.ExporterConfiguration {
+	return otlp.NewExporterConfiguration(dk.Spec.OTLPExporterConfiguration, dk.GetResourceAttributes())
+}
```
0sewa0: move to `pkg/api/latest/dynakube/operands_props.go`

pkg/controllers/dynakube/activegate/internal/customproperties/reconciler.go
```diff
@@ -29,6 +29,8 @@ const (
 
 	clientInternalSection = "[http.client.internal]"
 	noProxyFieldName      = "proxy-non-proxy-hosts"
+
+	customPropertiesConditionType string = "CustomPropertiesSecret"
```
0sewa0, line 33: would this work? so we don't stutter 
```suggestion
	conditionType string = "CustomPropertiesSecret"
```

pkg/controllers/dynakube/activegate/internal/statefulset/reconciler.go
```diff
@@ -24,6 +24,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	ActiveGateStatefulSetConditionType string = "ActiveGateStatefulSet"
```
0sewa0, line 28:
```suggestion
	ConditionType string = "ActiveGateStatefulSet"
```

(ignore it if it causes to many changes, as this is public)

pkg/controllers/dynakube/connectioninfo/activegate/reconciler.go
```diff
@@ -24,6 +24,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const activeGateConnectionInfoConditionType string = "ActiveGateConnectionInfo"
```
0sewa0, line 27:
```suggestion
const conditionType string = "ActiveGateConnectionInfo"
```

pkg/controllers/dynakube/extension/eec/reconciler.go
```diff
@@ -18,6 +18,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const extensionControllerStatefulSetConditionType string = "ExtensionControllerStatefulSet"
```
0sewa0, line 21:
```suggestion
const conditionType string = "ExtensionControllerStatefulSet"
```

last one I will comment, feel free to ignore these, I thought there will be only a few, but we are super inconsistent 🙈

pkg/api/latest/dynakube/dynakube_props.go
svfoxat: d2d05e66e8e68f797548a129119720c2cf113f4a

0sewa0 (dismissed):

svfoxat:
seems like stackoverflow is now rejecting our requests (maybe flagged as bot traffic?)

```
40 links checked.
  [✖] https://stackoverflow.com/questions/73433300/what-is-the-difference-between-errors-and-sentinel-errors → Status: 403
ERROR: 1 dead link found!
```
https://github.com/Dynatrace/dynatrace-operator/actions/runs/30792847626/job/91627214792?pr=6766#step:4:259

Should we just add `stackoverflow.com` to the `ignorePatterns` in `hack/markdown-link-check.json`
https://github.com/Dynatrace/dynatrace-operator/pull/6766/commits/4639572241d661ea1dfa491c47eb3ed3192112f3

0sewa0 (approved):

PR #6773, opened 2026-05-28
Remove loadsim command from support archive

waodim:
## Description

[ICP-3633](https://dt-rnd.atlassian.net/browse/ICP-3633)

With this PR we remove the unused, undocumented `loadsim` command from the support archive. If there was ever a use case for it, it is not known today, as far as I can tell it is not used anywhere, not documented anywhere, so we are safe to just remove it.

## How can this be tested?

- run the support archive command
- ***Please check our docs repo and check if the loadsim command is mentioned somewhere (I also checked and did not find anything but a second pair of eyes would help)***

[ICP-3633]: https://dt-rnd.atlassian.net/browse/ICP-3633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aorcholski (approved):
LGTM
don't see `loadsim` in the doc repo.

PR #6774, opened 2026-05-28
Add validation for NetworkZone and HostGroup input strings

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/ICP-5479)

## Description

A dynakube CR that has an invalid NetworkZone or Host Group value is rejected by validation webhook.

## How can this be tested?

unittests

try to apply a dynakube with invalid networkZone or hostGroup value

### DIFF
--- a/pkg/api/validation/dynakube/networkzone.go
+++ b/pkg/api/validation/dynakube/networkzone.go
@@ -1,0 +1,33 @@
+package validation
+
+import (
+	"context"
+	"strings"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+)
+
+const (
+	errorInvalidNetworkZone = "The DynaKube's specification has an invalid Network Zone value set. Make sure to remove all whitespace characters (newline, tab, carriage return, null) from the Network Zone value in your custom resource."
+)
+
+func invalidNetworkZone(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
+	if dk.Spec.NetworkZone != "" {
+		sanitizedNetworkZone := strings.Map(removeWhiteSpaceCharacters, dk.Spec.NetworkZone)
+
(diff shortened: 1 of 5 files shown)


pkg/api/validation/dynakube/networkzone.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+)
+
+const (
+	errorInvalidNetworkZone = "The DynaKube's specification has an invalid Network Zone value set. Make sure to remove all whitespace characters (newline, tab, carriage return, null) from the Network Zone value in your custom resource."
+)
+
+func invalidNetworkZone(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
+	if dk.Spec.NetworkZone != "" {
+		sanitizedNetworkZone := strings.Map(removeWhiteSpaceCharacters, dk.Spec.NetworkZone)
```
avorima: would be faster to use strings.ContainsFunc. we don't really need the mapped string

pkg/api/validation/dynakube/networkzone.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+	"strings"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+)
+
+const (
+	errorInvalidNetworkZone = "The DynaKube's specification has an invalid Network Zone value set. Make sure to remove all whitespace characters (newline, tab, carriage return, null) from the Network Zone value in your custom resource."
```
avorima: "all" isn't really accurate since we allow spaces for example. maybe "invalid" or "unsupported" or rephrase to: Make sure to remove newline, .... characters from the Network Zone value

avorima (approved):

PR #6775, opened 2026-05-28
Remove platform token requirement for 3rd gen URL

0sewa0:
## Description

https://dt-rnd.atlassian.net/browse/ICP-5483

3rd gen URLs support non-platform tokens, as it is handled in the client.
Therefore this validation is an artificial limitation, that becomes problematic if the user changes their URL / token without thinking about this. (we validate the DK -> can deny it, we do NOT validate the token secret -> user is free to "break it")

## How can this be tested?

Use your non-platform token
Add use the 3rd gen URL of your tenant
All should work


aorcholski (approved):
LGTM
classic token, dev.apps URL -> works

waodim (approved):

PR #6784, opened 2026-05-29
Fix pull secret not updating on api token change

mihaitanasedt:
## Description
When the API token is updated (rotation, migration between token types, etc.), the pull secret is not regenerated until the 15-minute TTL expires. This happens because the pull secret reconciler is gated behind the same TTL used to rate-limit Dynatrace API calls, even though it makes no Dynatrace API calls. As a result, the pull secret holds stale credentials and images are pulled using the outdated token for up to 15 minutes.

This PR fixes the issue and adapts the tests.

> [!IMPORTANT]
Additionally after discussions it was decided to not create the secret if a platform token is used and to remove the existing secret if there is a migration to platform token. Old tests were adapted and new ones were introduced

[ICP-5547 Jira](https://dt-rnd.atlassian.net/browse/ICP-5547)
## How can this be tested?

`make go/test`

### DIFF
--- a/pkg/controllers/dynakube/dtpullsecret/reconciler.go
+++ b/pkg/controllers/dynakube/dtpullsecret/reconciler.go
@@ -6,7 +6,6 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/token"
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
-	"github.com/Dynatrace/dynatrace-operator/pkg/util/dttoken"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8sconditions"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8slabel"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/objects/k8ssecret"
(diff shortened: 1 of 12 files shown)


0sewa0:
related to the discussion in the meeting:
- when you no longer create the pull-secret when platformToken is used, don't forget to update the `ImagePullSecretReferences` method to also respect this

pkg/controllers/dynakube/dtpullsecret/reconciler.go
```diff
@@ -22,21 +22,19 @@ const (
...
-		secrets:      k8ssecret.Query(clt, apiReader),
+		secrets: k8ssecret.Query(clt, apiReader),
 	}
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context, dk *dynakube.DynaKube, tokens token.Tokens) error {
 	ctx, log := logd.NewFromContext(ctx, "dynakube-pullsecret")
 
-	if !dk.OneAgent().IsDaemonsetRequired() && !dk.ActiveGate().IsEnabled() {
+	if dttoken.IsPlatform(tokens.APIToken().String()) || (!dk.OneAgent().IsDaemonsetRequired() && !dk.ActiveGate().IsEnabled()) {
```
0sewa0: I am noticing an inconsistency in the code base

`dttoken.IsPlatform` used 8 times in the code
- 5 with `tokens.APIToken().Value`
- 1 with `tokens.APIToken().String` (this is this line)
- 1 with `tokens.APIToken().Value` + nil check
- 1 with raw string

wouldn't it make sense to introduce a method for the `token.Tokens`, like:
```go
func (tokens Tokens) HasPlatformToken() bool {
	return tokens.APIToken() != nil && dttoken.IsPlatform(tokens.APIToken().Value)
}
``` 

cc @avorima

mihaitanasedt: good point 👍

mihaitanasedt: Done https://github.com/Dynatrace/dynatrace-operator/pull/6784/changes/9585e8afac7fcbb2930c4688641e4f062b541631 

I used `tokens.APIToken().String()` instead of `.Value` because it already does the `nil` check

0sewa0 (approved):

PR #6787, opened 2026-06-01
Add liveness and readiness probes to OpenTelemetry collector deployment

StefanHauth:
## Description

The OpenTelemetry pods up to now do not have liveness and readiness probes configured although the resp. health_check extensions is already configured and used in the collector.

Add a hardcoded default configuration for those probes.

[JIRA](https://dt-rnd.atlassian.net/browse/ICP-5165)

## How can this be tested?

1. Deploy a DynaKube with `telemetryIngest` enabled.
2. Observe the OpenTelemetry collector pod(s)
    * check their prob configs
    * produce some load on the collector
    * observe that the collector doesn't restart unexpectedly

### DIFF
--- a/pkg/controllers/dynakube/otelc/statefulset/container.go
+++ b/pkg/controllers/dynakube/otelc/statefulset/container.go
@@ -5,7 +5,9 @@ import (
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/consts"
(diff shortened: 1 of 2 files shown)


avorima (dismissed):

waodim (commented):
I'd approve this just one question:

TimeoutSeconds/FailureThreshold/SuccessThreshold are not set in these probes but in all other probes throughout the codebase they are set. Not that I worry much, just wanted to ask, do we want/need consistency here, or do we not care?

waodim (approved):

PR #6788, opened 2026-06-01
Add configuration option for init container resources to DynaKube

StefanHauth:
## Description

Resources for the init containers in case of metadata-only shall be configurable in the DynaKube.
* The configuration option `.spec.metadataEnrichment.initRersources` has been added to the CRD
* The options are used only in case of metadata-only

[JIRA](https://dt-rnd.atlassian.net/browse/ICP-5282)

## How can this be tested?

* deploy metadata-only DK with resources set, deploy sample app and check that the init resources are applied
* deploy a metadata + OA DK with resources set for both, deploy sample app and check that OA init resources are applied
* deploy DK with the features but without resource settings, check that default resource values are set on the init container in the sample app


avorima (dismissed):
Tested listed scenarios. Resources are set accordingly

pkg/webhook/mutation/pod/handler/injection/handle.go
```diff
@@ -146,6 +146,10 @@ func (h *Handler) handlePodMutation(mutationRequest *dtwebhook.MutationRequest)
 		mutated = true
 	}
 
+	if !oaEnabled {
+		mutationRequest.InstallContainer.Resources = metadataEnrichmentInitResources(mutationRequest.DynaKube)
+	}
+
```
0sewa0: this flow feels weird to me, 

we are checking the `oaEnabled` 2 times, with no clear benefit
you could just do
```go
if oaEnabled {
...
} else {
	mutationRequest.InstallContainer.Resources = metadataEnrichmentInitResources(mutationRequest.DynaKube)
}
```
(the weird edgecase where globally the features are turned on, but on the pod level they are disabled is also not handled by the `!oaEnabled` check)


also you could easily enforce the precedence of the settings by:
- running the `metaMutator` first
- setting `mutationRequest.InstallContainer.Resources = metadataEnrichmentInitResources(mutationRequest.DynaKube)` inside the `metaMutator`
- just overrule it in the oaMutator`


sidenote: the weird edgecase I mentioned does nothing, because `mutated` will still be false, so no init-container will be added, this is just a matter of (IMO) confusing flow

0sewa0 (approved):
tested it, works as expected
- standalone metadata-enrichment respects the new field
- OA field wins, even if empty
- pod level annotations are also respected

PR #6792, opened 2026-06-02
Adopt warning when use_public_registry FF used or (platform tokens gen 3) and version field is set

andriisoldatenko:
## Description

Resolves https://dt-rnd.atlassian.net/browse/ICP-5467

In gen 3 when platform token is used and where public registry is enforced and defaulted, and if user still have version field set inside dynakube spec, they should see warning about version field is ignored and deprecated. 
Also in gen2/normal token but when public registry FF is enabled we should do same.

This PRs add extra warning and action item:

```
version field is deprecated and ignored. Please remove the version field from the DynaKube specification.`
```

## How can this be tested?

unit tests

to understand better problem and solution focus on new/existing unit tests.

### DIFF
--- a/pkg/api/validation/dynakube/oneagent.go
+++ b/pkg/api/validation/dynakube/oneagent.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/exp"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/token"
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/dtversion"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8senv"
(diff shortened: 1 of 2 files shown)


0sewa0 (dismissed):

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -278,10 +281,20 @@ func forbiddenHostIDSourceArgument(_ context.Context, _ *Validator, dk *dynakube
...
 	oa := dk.OneAgent()
 
 	if oa.GetCustomVersion() != "" {
+		if dk.FF().IsPublicRegistry() {
+			return warningDeprecatedVersionIgnored
+		}
+
+		// For new DynaKubes (status not yet set), check the token secret directly.
+		hasPlatformToken, err := token.NewReader(dv.apiReader, dk).HasPlatformToken(ctx)
+		if err == nil && hasPlatformToken {
```
jokuniew, line 294: nit: check for err != nil and log error message if checking token has failed (example scenario - secret was deleted/ never created)

andriisoldatenko: Based on other places where we have similar code, we don't handle this error. We just assume if error we return "" no warning, in my case we assume there is not platformToken and move on.

We can reconsider it in follow-ups, so we can fix all at-once 
https://github.com/dynatrace/dynatrace-operator/blob/e8d281c9c6bc51d34498e3587e55450182833c0d/pkg/api/validation/dynakube/public_registry.go#L25

https://github.com/dynatrace/dynatrace-operator/blob/e8d281c9c6bc51d34498e3587e55450182833c0d/pkg/api/validation/dynakube/public_registry.go#L56

and https://github.com/dynatrace/dynatrace-operator/blob/e8d281c9c6bc51d34498e3587e55450182833c0d/pkg/api/validation/dynakube/public_registry.go#L37-L40

0sewa0 (approved):

avorima (approved):

PR #6794, opened 2026-06-03
add idea to comment results back to PR simialr to kind workflow

andriisoldatenko:


0sewa0:
what is the state here?

PR #6803, opened 2026-06-05
fix: stop hard links escaping the codemodule extraction dir

knQzx:
## problem

hard link extraction in the codemodule gzip unpacker checks the wrong base path

`extractLink` validates with `isSafeToLink(header.Linkname, targetDir, target)` but `os.Link` resolves the source as `filepath.Join(targetDir, header.Linkname)`
the check uses `target` (the deep `targetDir/header.Name`) while os.Link uses `targetDir`
tar hard link linknames are relative to the archive root so os.Link is right and the check is wrong - the extra depth in `target` eats leading `../`
so a layer entry with a deep name and a matching `../` linkname passes the check but links outside the extraction dir

## impact

extraction runs in the privileged csi provisioner and the unpacked dir is renamed into the codemodule dir mounted into pods
a malicious or compromised codemodule image can hard link to a file outside the extraction root on the same csi volume - cross pod disclosure
bounded by exdev so not arbitrary host files
the symlink branch is fine - only the hard link base was wrong

## fix

validate the linkname against `targetDir` - the same base os.Link uses

## test

`extract gzip with malicious hard link blocks escaping link` fails on main and passes here
plus a positive case so legit in archive hard links still work

### DIFF
--- a/pkg/injection/codemodule/installer/zip/gzip.go
+++ b/pkg/injection/codemodule/installer/zip/gzip.go
@@ -108,7 +108,7 @@ func extract(ctx context.Context, targetDir string, reader *tar.Reader, header *
 func extractLink(ctx context.Context, targetDir, target string, header *tar.Header) error {
 	log := logd.FromContext(ctx)
 
-	if isSafeToLink(header.Linkname, targetDir, target) && isSafeToLink(header.Name, targetDir, target) {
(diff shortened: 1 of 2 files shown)


0sewa0:
Hi, I am closing this as https://github.com/Dynatrace/dynatrace-operator/pull/6874 makes this unnecessary

we don't want to improve that logic, we wish to remove it.

PR #6811, opened 2026-06-09
feat: process group config for hostAgent

svfoxat:
## Description

https://dt-rnd.atlassian.net/browse/ICP-5529

Mounts `declarative.cbor` for processgroupconfig into OneAgent Daemonset in `HostMonitoring` and `CloudNativeFullStack` modes.

Reuses already implemented fetching, storing and caching of the processgroupconfig in the `bootstrapperconfig-secret` (https://github.com/Dynatrace/dynatrace-operator/pull/6541) 

When PGC changes on the tenant a rollout of the OneAgent Daemonset is triggered on the next reconcile.

## How can this be tested?

`make go/test`
`make go/lint`

 `make test/e2e/pgc`


pkg/injection/namespace/bootstrapperconfig/bootstrapperconfig.go
```diff
@@ -242,6 +244,10 @@ func NeedsDownloadConfig(dk *dynakube.DynaKube) bool {
 	return dk.OneAgent().IsAppInjectionNeeded() && !dk.OneAgent().IsCSIAvailable() && dk.OneAgent().GetCodeModulesImage() == ""
 }
 
+func NeedsPGC(dk *dynakube.DynaKube) bool {
+	return dk.OneAgent().IsAppInjectionNeeded() || dk.OneAgent().IsHostMonitoringMode()
```
StefanHauth, line 248: Isn't that just "not ClassicFullStack"?

```suggestion
	return !dk.OneAgent().IsClassiFullStackMode()
```

hack/make/tests/e2e.mk
```diff
@@ -171,13 +171,23 @@ test/e2e/applicationmonitoring/bootstrapper-csi:
...
 
-## Runs PGC bootstrapper with CloudNativeFullStack (without CSI) e2e test only
+## Runs PGC bootstrapper and host agent tests for CloudNativeFullStack (without CSI) e2e test only
 test/e2e/bootstrapper/pgc-fullstack-no-csi:
-	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "pgc_with_fullstack" $(SKIPCLEANUP)
+	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "pgc_with_fullstack|host_agent_pgc_cloudnative" $(SKIPCLEANUP)
+
+## Runs PGC bootstrapper and host agent tests for CloudNativeFullStack (both CSI and no-CSI)
+test/e2e/bootstrapper/pgc-fullstack:
+	$(MAKE) test/e2e/bootstrapper/pgc-fullstack-csi && \
```
avorima: if this is meant to exit early, you can use separate lines instead of `&&`. make will exit when one command fails

pkg/controllers/dynakube/oneagent/daemonset/volumes.go
```diff
@@ -240,3 +249,28 @@ func getRootVolume() corev1.Volume {
...
+				Optional: new(true),
+			},
+		},
+	}
+}
+
+func getPGCSecretFileMount() corev1.VolumeMount {
+	return corev1.VolumeMount{
+		Name:      pgcSecretVolumeName,
+		MountPath: csiStorageVolumeMount + "/opt/agent/conf/" + bootstrapperconfig.DeclarativeInputFileName,
```
avorima, line 272: the mount path should be the directory

hack/make/tests/e2e.mk
```diff
@@ -171,13 +171,23 @@ test/e2e/applicationmonitoring/bootstrapper-csi:
 test/e2e/applicationmonitoring/bootstrapper-no-csi:
 	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "node_image_pull_with_no_csi" $(SKIPCLEANUP)
 
-## Runs PGC bootstrapper with CloudNativeFullStack (with CSI) e2e test only
+## Runs PGC bootstrapper and host agent tests for CloudNativeFullStack (with CSI) e2e test only
 test/e2e/bootstrapper/pgc-fullstack-csi:
-	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/standard -run "pgc_with_fullstack" $(SKIPCLEANUP)
+	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/standard -run "pgc_with_fullstack|host_agent_pgc_cloudnative" $(SKIPCLEANUP)
```
avorima: When looking at the top-level `test/e2e/pgc` it's a bit weird that the cloudnative PGC tests are implicitly run by the bootstrapper tests. Moreover `bootstrapper` doesn't quite fit into the structure of the other e2e tests. They usually mention the feature, e.g. appmon, cloudnative, and then a subcategory.
Given that pcg_with_fullstack uses CNFS as well, I'd rather place both of these in the cloudnative tests

pkg/injection/namespace/bootstrapperconfig/bootstrapperconfig.go
svfoxat: You're right, it is.

test/e2e/features/cloudnative/pgc_hostagent.go
```diff
@@ -0,0 +1,29 @@
...
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func HostAgentPGC(t *testing.T) features.Feature {
+	builder := features.New("host-agent-pgc")
+	secretConfig := tenant.GetSingleTenantSecret(t)
+
+	testDynakube := *componentDynakube.New(
+		componentDynakube.WithAPIURL(secretConfig.APIURL),
+		componentDynakube.WithCloudNativeSpec(DefaultCloudNativeSpec()),
```
avorima, line 22: why doesn't this create an AG?

pkg/injection/namespace/bootstrapperconfig/bootstrapperconfig.go
svfoxat: Seems like this breaks some tests (other than the build error). 

`!dk.OneAgent().IsClassicFullStackMode()` would also return true when only `MetadataEnrichement` is activated, like in the `MetadataOnly` test.
If this is wanted then I can adapt but I only know about the OA modes mentioned in the PR description @StefanHauth

test/e2e/features/cloudnative/pgc_hostagent.go
svfoxat: No specific reason, was an oversight.

test/e2e/results/TestStandard_host_agent_pgc_cloudnative_operator.log
```diff
@@ -0,0 +1,10 @@
+{"level":"info","ts":"2026-06-11T12:40:45.945Z","logger":"install-config","msg":"envvar content read and set","envvar":"modules.json","value":"{\n  \"csiDriver\": true,\n  \"activeGate\": true,\n  \"oneAgent\": true,\n  \"extensions\": true,\n  \"logMonitoring\": true,\n  \"edgeConnect\": true,\n  \"supportability\": true,\n  \"kubernetesMonitoring\": true,\n  \"kspm\": true\n}\n"}
```
avorima: Please add a .gitignore entry for these log files

pkg/injection/namespace/bootstrapperconfig/bootstrapperconfig.go
andriisoldatenko: better to keep it as is.

test/e2e/results/TestStandard_host_agent_pgc_cloudnative_operator.log
andriisoldatenko: yes +1

pkg/controllers/dynakube/oneagent/daemonset/volumes.go
svfoxat: This seems to break OneAgent, while testing these changes it started to complain that the Filesystem is read-only and the pod started crashlooping.

test/e2e/results/TestStandard_host_agent_pgc_cloudnative_operator.log
svfoxat: Done 👍

pkg/controllers/dynakube/oneagent/oneagent_reconciler.go
```diff
@@ -337,6 +347,21 @@ func (r *Reconciler) buildDesiredDaemonSet(ctx context.Context, dk *dynakube.Dyn
 	return ds, nil
 }
 
+func (r *Reconciler) pgcConfigHash(ctx context.Context, dk *dynakube.DynaKube) (string, error) {
```
waodim, line 352: Would it make sense to add a unit test for this specific function?

test/e2e/helpers/pgc/pgc.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+const FilePath = "/mnt/volume_storage_mount/opt/agent/conf/" + bootstrapperconfig.DeclarativeInputFileName
```
waodim, line 25: There exists a constant `csiStorageVolumeMount = "/mnt/volume_storage_mount"`

Don't know whether this is likely, but what if the value changes? Should we export this constant and use it here? @0sewa0 @andriisoldatenko WDYT?

pkg/controllers/dynakube/oneagent/oneagent_reconciler.go
svfoxat: Good idea, added 3 testcases.

test/e2e/helpers/pgc/pgc.go
svfoxat: Not sure if it is worth exposing an implementation detail for just one e2e test. If it should change, the test would fail anyway.

andriisoldatenko: agree let's keep it as is.

andriisoldatenko (approved):

PR #6813, opened 2026-06-11
Improve cleanup-node-fs.sh: selectable cleanup modes, deep classic cleanup, per-node reporting

averagenative:
## Description

`hack/cluster/cleanup-node-fs.sh` previously always removed **both** classicFullStack host paths and CSI driver data, and judged success solely by the DaemonSet ready count. This PR makes the cleanup scope selectable and the results verifiable.

### Selectable cleanup modes

Interactive menu when run on a TTY without flags, or `--mode classic|csi|all` for automation (`-y` skips all prompts). Positional `[NAMESPACE]`, `SKIP_RUNNING_PODS_WARNING` and `MAX_WAIT_SECONDS` keep working as before.

- **classic** — host-installed OneAgent: runs `uninstall.sh` (if present), removes `/opt/dynatrace`, `/var/lib/dynatrace`, `/var/log/dynatrace`
- **csi** — CSI driver data dir plus `/var/opt/dynatrace` (default `storageHostPath` of non-CSI readonly installs); can be scoped to one tenant/dynakube via `--csi-subpath`
- **all** — both (previous behavior)

The groups are disjoint (OneAgent only writes the typical host paths when CSI is *not* in use), so **classic-only cleanup is safe alongside a working cloudNativeFullStack/hostMonitoring install** — a common support scenario where residue of an old classicFullStack deployment interferes with a working CSI-based one.

`--extra-dir` covers custom `storageHostPath` values, `--kubelet-path` covers non-standard distros (k0s/microk8s/RKE2-style).

### Deep classic cleanup (also when uninstall.sh is missing/failed)

- scrubs `oneagent` entries from `/etc/ld.so.preload` (prevents `ld.so: object ... cannot be preloaded` spam on every process), preserving unrelated entries
- stops/disables and removes `oneagent*.service` systemd units, then `daemon-reload`
- kills lingering host processes running from dynatrace paths (SIGTERM, grace, SIGKILL); requires `hostPID`. Containerized CSI-based agents report the same exe paths, so only processes sharing the host root filesystem (`/proc/<pid>/root` vs `/proc/1/root`) are killed — a live CSI-based install is never touched

### Coverage verification

The DaemonSet silently skips nodes whose taints it doesn't tolerate, so `ready == desired` is not proof every node was cleaned. The script now compares cleanup pods against all Linux nodes and lists uncovered nodes with their taints.

### Evidence gathering

- per-path results (`REMOVED` incl. size / `MISSING` / `FAILED` / `LEFTOVER` post-rm verification) reported through the init container termination message and shown per node
- pre-removal `du`/`ls` inventory in the init container logs
- all init logs + pod manifests saved per node to a local report directory (`--report-dir`) before the DaemonSet is deleted

### Bug fix

On failure the script kept the DaemonSet "for investigation" but then deleted the namespace it had created — destroying the DaemonSet anyway. The namespace is now kept whenever the DaemonSet is kept.

## How can this be tested?

- `shellcheck` clean (outer script and the generated init script)
- Generated DaemonSet manifest validated; init script executed against a fake node root **inside the pinned `ubi9-micro` image** — this caught that `grep` does not exist in ubi-micro, so the preload scrub uses shell pattern matching instead
- Full script flow exercised end-to-end against a stubbed `kubectl` (mode selection, coverage warning, per-node report, log collection, failure path)
- Not yet exercised on a real cluster node with a live classic OneAgent (hostPID process-kill path)

### DIFF
--- a/hack/cluster/cleanup-node-fs.sh
+++ b/hack/cluster/cleanup-node-fs.sh
@@ -1,21 +1,229 @@
 #!/bin/bash
 
 # DESCRIPTION:
-#   This script deploys a DaemonSet that runs on all Linux nodes (amd64/arm64) in the cluster
-#   to clean up Dynatrace OneAgent installations and CSI driver artifacts. It uses an init
-#   container to perform the cleanup, ensuring each node is cleaned exactly once without restarts.
+#   This script deploys a DaemonSet that runs on all Linux nodes (amd64/arm64/ppc64le/s390x)
+#   in the cluster to clean up Dynatrace OneAgent installations and/or CSI driver artifacts.
+#   It uses an init container to perform the cleanup, ensuring each node is cleaned exactly
+#   once without restarts.
+#
+#   What gets cleaned depends on the selected mode:
+#     classic - artifacts of a classicFullStack (host-installed) OneAgent:
+#                 /opt/dynatrace, /var/lib/dynatrace, /var/log/dynatrace
+#               and runs /opt/dynatrace/oneagent/agent/uninstall.sh first, if present.
+#               Additionally removes system-level residue (even when uninstall.sh is
+#               missing or failed): oneagent entries in /etc/ld.so.preload, oneagent
+#               systemd units, and still-running host processes started from dynatrace
+#               paths. Containerized agents (CSI-based installs) are never touched -
+#               only processes running directly on the host are killed.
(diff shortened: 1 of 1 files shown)


Kortecoder-DT:
Dont forget to check  /etc/containers/oci/hooks.d for signs of oneagenthelper / hook json file references when doing the manual clean up on CRIO based systems, looking at openshift based deployments.

PR #6815, opened 2026-06-11
Disable OTELC probes when EEC and prometheus is enabled

avorima:
## Description

When deploying the OTEL statefulset with the Prometheus extension the configuration is missing the health check port which causes the probes to fail.
The probes work when we deploy it using telemetryIngest.

Fixes test failures introduced in #6787

ICP-6010

## How can this be tested?

* Deploy prometheus extension and check that probes are not present
* Deploy telemetryIngest and check the pod comes up

### DIFF
--- a/pkg/controllers/dynakube/otelc/statefulset/container.go
+++ b/pkg/controllers/dynakube/otelc/statefulset/container.go
@@ -17,7 +17,7 @@ const (
 )
 
 func getContainer(dk *dynakube.DynaKube, replicas int32) corev1.Container {
-	return corev1.Container{
+	container := corev1.Container{
 		Name:            containerName,
 		Image:           dk.Spec.Templates.OpenTelemetryCollector.ImageRef.String(),
 		ImagePullPolicy: dk.Spec.Templates.OpenTelemetryCollector.ImageRef.GetPullPolicy(),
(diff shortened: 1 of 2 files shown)


pkg/controllers/dynakube/otelc/statefulset/container.go
```diff
@@ -26,9 +26,16 @@ func getContainer(dk *dynakube.DynaKube, replicas int32) corev1.Container {
 		Resources:       dk.Spec.Templates.OpenTelemetryCollector.Resources,
 		Args:            buildArgs(dk),
 		VolumeMounts:    buildContainerVolumeMounts(dk),
-		LivenessProbe:   buildLivenessProbe(),
-		ReadinessProbe:  buildReadinessProbe(),
 	}
+
+	// It seems like EEC doesn't send the health probe config to OTELC, so disable the probes for it.
+	// The EEC + prometheus is not GA and may be removed in a future release, so it's an accepted caveat.
+	if !dk.Extensions().IsPrometheusEnabled() {
```
StefanHauth: To be on the safe side, only rely on having "health_check_extension" when we have the OTelC config under our control.
```suggestion
	if dk.TelemetryIngest().IsEnabled() {
```

StefanHauth (changes requested):

pkg/controllers/dynakube/otelc/statefulset/container.go
```diff
@@ -26,9 +26,16 @@ func getContainer(dk *dynakube.DynaKube, replicas int32) corev1.Container {
 		Resources:       dk.Spec.Templates.OpenTelemetryCollector.Resources,
 		Args:            buildArgs(dk),
 		VolumeMounts:    buildContainerVolumeMounts(dk),
-		LivenessProbe:   buildLivenessProbe(),
-		ReadinessProbe:  buildReadinessProbe(),
 	}
+
+	// It seems like EEC doesn't send the health probe config to OTELC, so disable the probes for it.
```
andriisoldatenko:
```suggestion
	// It EEC doesn't send the health probe config to OTELC, so disable the probes for it.
```

andriisoldatenko (approved):

StefanHauth (approved):

PR #6822, opened 2026-06-15
ICP-5582: Deprecate k8s-app-enabled feature flag

gkrenn:
## Description

Deprecates the `k8s-app-enabled` feature flag. The `builtin:app-transition.kubernetes` schema is no longer available on phase 3 tenants, so the flag has run its course. Users who still have it set will receive a validation warning on apply; the underlying behavior is unchanged.

[ICP-5582](https://dt-rnd.atlassian.net/browse/ICP-5582)

## How can this be tested?

1. Deploy a DynaKube with the annotation set:
   ```yaml
   metadata:
     annotations:
       feature.dynatrace.com/k8s-app-enabled: "true"
   ```
2. Apply it and observe the deprecation warning in `kubectl apply` output:
   ```
   Warning: Using deprecated feature flags: feature.dynatrace.com/k8s-app-enabled
   ```
3. Verify the DynaKube reaches `Running` state.
4. Remove the annotation and re-apply — no warning should appear and the DynaKube should remain `Running`.

[ICP-5582]: https://dt-rnd.atlassian.net/browse/ICP-5582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/api/exp/activegate.go
+++ b/pkg/api/exp/activegate.go
@@ -12,8 +12,8 @@ const (
 	AGAutomaticK8sAPIMonitoringKey            = FFPrefix + "automatic-kubernetes-api-monitoring"
 	AGAutomaticK8sAPIMonitoringClusterNameKey = FFPrefix + "automatic-kubernetes-api-monitoring-cluster-name"
 	// Deprecated: The builtin:app-transition.kubernetes schema is no longer available on phase 3 tenants.
-	AGK8sAppEnabledKey = FFPrefix + "k8s-app-enabled"
-	AGAutomaticTLSCertificateKey              = FFPrefix + "automatic-tls-certificate"
+	AGK8sAppEnabledKey           = FFPrefix + "k8s-app-enabled"
+	AGAutomaticTLSCertificateKey = FFPrefix + "automatic-tls-certificate"
 )
 
 // IsActiveGateUpdatesDisabled is a feature flag to disable ActiveGate updates.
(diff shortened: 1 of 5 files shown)


pkg/api/exp/activegate.go
```diff
@@ -11,8 +11,9 @@ const (
 	AGAppArmorKey                             = FFPrefix + "activegate-apparmor"
 	AGAutomaticK8sAPIMonitoringKey            = FFPrefix + "automatic-kubernetes-api-monitoring"
 	AGAutomaticK8sAPIMonitoringClusterNameKey = FFPrefix + "automatic-kubernetes-api-monitoring-cluster-name"
-	AGK8sAppEnabledKey                        = FFPrefix + "k8s-app-enabled"
-	AGAutomaticTLSCertificateKey              = FFPrefix + "automatic-tls-certificate"
+	// Deprecated: The builtin:app-transition.kubernetes schema is no longer available on phase 3 tenants.
+	AGK8sAppEnabledKey           = FFPrefix + "k8s-app-enabled"
+	AGAutomaticTLSCertificateKey = FFPrefix + "automatic-tls-certificate"
```
andriisoldatenko, line 16: can we align with another similar FF? meaning can we add an action item for customer?

also we will print warning also for phase 2 tenants, because we don't check platoform token... or am i miss something?

gkrenn: This should bring a warning for any customer, as even for most older tenants the setting has been auto created by now, according to the jira ticket.

andriisoldatenko (dismissed):

pkg/api/exp/activegate.go
gkrenn: I improved the deprecation message.

gkrenn: Please have a look again @andriisoldatenko

andriisoldatenko (approved):

PR #6824, opened 2026-06-15
Call create-sbom reusable workflow

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/ICP-3903)

## Description

The create-sbom step is replaced by a job that calls reusable workflow.

### DIFF
--- a/.github/actions/copy-image-to-artifact/action.yaml
+++ b/.github/actions/copy-image-to-artifact/action.yaml
@@ -2,7 +2,7 @@ name: Copy an image to artifact
 description: Copy an image to artifact
 inputs:
   image:
-    description: Base name of the image tag
+    description: Full image reference to pull (e.g. registry/repo:tag@sha256:...)
     required: true
   artifact-name:
     description: Artifact name to upload the image to
@@ -12,7 +12,7 @@ inputs:
     required: false
     default: ''
   registry-token:
-    description: token for registry authentication
+    description: Password/token for registry auth; for public.ecr.aws provide the role ARN to assume
     required: false
   registry-username:
     description: username for registry authentication
--- a/.github/actions/copy-image-to-artifact/action.yaml
+++ b/.github/actions/copy-image-to-artifact/action.yaml
@@ -51,9 +51,7 @@ runs:
         docker save -o /tmp/${ARTIFACT_NAME}.tar ${IMAGE}
     - name: Upload image artifact
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
-      env:
-        ARTIFACT_NAME: ${{ inputs.artifact-name }}
       with:
-        name: ${ARTIFACT_NAME}
-        path: /tmp/${ARTIFACT_NAME}.tar
+        name: ${{ inputs.artifact-name }}
+        path: /tmp/${{ inputs.artifact-name }}.tar
         retention-days: 1
--- a/.github/actions/copy-image-to-artifact/action.yaml
+++ b/.github/actions/copy-image-to-artifact/action.yaml
@@ -1,0 +1,57 @@
+name: Copy an image to artifact
+description: Copy an image to artifact
+inputs:
+  image:
+    description: Base name of the image tag
+    required: true
+  artifact-name:
+    description: Artifact name to upload the image to
+    required: true
+    type: string
+    default: ''
+  registry:
+    description: registry to authenticate against before scanning (optional)
+    required: false
+    type: string
+    default: ''
+  registry-token:
+    description: token for registry authentication
+    required: false
+  registry-username:
+    description: username for registry authentication
+    required: false
+runs:
+  using: "composite"
+  env:
+    IMAGE: ${{ inputs.image }}
+    ARTIFACT_NAME: ${{ inputs.artifact-name }}
+  steps:
+    - name: Login to Registry
+      if: ${{ inputs.registry != '' && inputs.registry != 'public.ecr.aws' }}
+      uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4.2.0
+      with:
+        registry: ${{ inputs.registry }}
+        username: ${{ inputs.registry-username }}
+        password: ${{ inputs.registry-token }}
+    - name: Configure aws credentials
+      if: ${{ inputs.registry == 'public.ecr.aws' }}
+      uses: aws-actions/configure-aws-credentials@99214aa6889fcddfa57764031d71add364327e59 # v6.1.3
+      with:
+        role-to-assume: ${{ inputs.registry-token }}
+        aws-region: us-east-1
+    - name: Login to Amazon ECR
+      if: ${{ inputs.registry == 'public.ecr.aws' }}
+      uses: aws-actions/amazon-ecr-login@fa648b43de3d4d023bcb3f89ed6940096949c419 # v2.1.5
+      with:
+        registry-type: public
+    - name: Save docker image
+      shell: bash
+      run: |
+        docker pull ${IMAGE}
+        docker save -o /tmp/${ARTIFACT_NAME}.tar ${IMAGE}
+    - name: Upload image artifact
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      with:
+        name: ${ARTIFACT_NAME}
+        path: /tmp/${ARTIFACT_NAME}.tar
+        retention-days: 1
--- a/.github/actions/copy-image-to-artifact/action.yaml
+++ b/.github/actions/copy-image-to-artifact/action.yaml
@@ -7,12 +7,9 @@ inputs:
   artifact-name:
     description: Artifact name to upload the image to
     required: true
-    type: string
-    default: ''
   registry:
     description: registry to authenticate against before scanning (optional)
     required: false
-    type: string
     default: ''
   registry-token:
     description: token for registry authentication
--- a/.github/actions/copy-image-to-artifact/action.yaml
+++ b/.github/actions/copy-image-to-artifact/action.yaml
@@ -22,9 +22,6 @@ inputs:
     required: false
 runs:
   using: "composite"
-  env:
-    IMAGE: ${{ inputs.image }}
-    ARTIFACT_NAME: ${{ inputs.artifact-name }}
   steps:
     - name: Login to Registry
       if: ${{ inputs.registry != '' && inputs.registry != 'public.ecr.aws' }}
@@ -46,11 +43,16 @@ runs:
         registry-type: public
     - name: Save docker image
       shell: bash
+      env:
+        ARTIFACT_NAME: ${{ inputs.artifact-name }}
+        IMAGE: ${{ inputs.image }}
       run: |
         docker pull ${IMAGE}
         docker save -o /tmp/${ARTIFACT_NAME}.tar ${IMAGE}
     - name: Upload image artifact
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      env:
+        ARTIFACT_NAME: ${{ inputs.artifact-name }}
       with:
         name: ${ARTIFACT_NAME}
         path: /tmp/${ARTIFACT_NAME}.tar
--- a/.github/actions/copy-image-to-artifact/action.yaml
+++ b/.github/actions/copy-image-to-artifact/action.yaml
@@ -44,8 +44,8 @@ runs:
         ARTIFACT_NAME: ${{ inputs.artifact-name }}
         IMAGE: ${{ inputs.image }}
       run: |
-        docker pull ${IMAGE}
-        docker save -o /tmp/${ARTIFACT_NAME}.tar ${IMAGE}
+        docker pull "${IMAGE}"
+        docker save -o "/tmp/${ARTIFACT_NAME}.tar" "${IMAGE}"
     - name: Upload image artifact
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
       with:
--- a/.github/workflows/create-sbom.yaml
+++ b/.github/workflows/create-sbom.yaml
@@ -10,7 +10,8 @@ on:
         description: name of the uploaded image artifact to scan locally
         required: true
         type: string
-permissions: {}
+permissions:
+  actions: write
 jobs:
   create-sbom:
     runs-on: ubuntu-24.04
--- a/.github/workflows/create-sbom.yaml
+++ b/.github/workflows/create-sbom.yaml
@@ -2,54 +2,28 @@ name: Create SBOM
 on:
   workflow_call:
     inputs:
-      image:
-        description: full image reference to scan
-        required: true
-        type: string
-      artifact-name:
+      sbom-artifact-name:
         description: name for the uploaded SBOM artifact
         required: true
         type: string
-      registry:
-        description: registry to authenticate against before scanning (optional)
-        required: false
+      image-artifact-name:
+        description: name of the uploaded image artifact to scan locally
+        required: true
         type: string
-        default: ''
-    secrets:
-      registry-token:
-        description: token for registry authentication
-        required: false
-      registry-username:
-        description: username for registry authentication
-        required: false
-permissions:
-  id-token: write
+permissions: {}
 jobs:
   create-sbom:
     runs-on: ubuntu-24.04
     steps:
-      - name: Login to Registry
-        if: ${{ inputs.registry != '' && inputs.registry != 'public.ecr.aws' }}
-        uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4.2.0
+      - name: Download image artifact
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
         with:
-          registry: ${{ inputs.registry }}
-          username: ${{ secrets.registry-username }}
-          password: ${{ secrets.registry-token }}
-      - name: Configure aws credentials
-        if: ${{ inputs.registry == 'public.ecr.aws' }}
-        uses: aws-actions/configure-aws-credentials@99214aa6889fcddfa57764031d71add364327e59 # v6.1.3
-        with:
-          role-to-assume: ${{ secrets.registry-token }}
-          aws-region: us-east-1
-      - name: Login to Amazon ECR
-        if: ${{ inputs.registry == 'public.ecr.aws' }}
-        uses: aws-actions/amazon-ecr-login@fa648b43de3d4d023bcb3f89ed6940096949c419 # v2.1.5
-        with:
-          registry-type: public
+          name: ${{ inputs.image-artifact-name }}
+          path: image
       - name: Create SBOM
         uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
         with:
-          image-ref: ${{ inputs.image }}
+          input: image/${{ inputs.image-artifact-name }}.tar
           format: 'cyclonedx'
           output: 'result.json'
           skip-dirs: '/usr/share/dynatrace-operator/third_party_licenses'
@@ -57,6 +31,6 @@ jobs:
       - name: Upload SBOM artifact
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: ${{ inputs.artifact-name }}
+          name: ${{ inputs.sbom-artifact-name }}
           path: result.json
           retention-days: 1
--- a/.github/workflows/create-sbom.yaml
+++ b/.github/workflows/create-sbom.yaml
@@ -22,7 +22,8 @@ on:
       registry-username:
         description: username for registry authentication
         required: false
-permissions: {}
+permissions:
+  id-token: write
 jobs:
   create-sbom:
     runs-on: ubuntu-24.04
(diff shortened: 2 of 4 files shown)


.github/workflows/fips-release.yaml
```diff
@@ -79,9 +79,12 @@ jobs:
     runs-on: ubuntu-24.04
     permissions:
       id-token: write
-      attestations: write
-      artifact-metadata: write
     environment: Release
+    outputs:
+      dockerhub-image: ${{ steps.image-ref.outputs.dockerhub-image }}
```
mihaitanasedt: I don't thing this will work. When a matrix is run each job is running separatly and only the `GITHUB_OUTPUT` of the last job is passed down to the next ones. This means only one output will be populated and the others will be empty. Did you somehow test it ?

https://github.com/orgs/community/discussions/17245

.github/workflows/fips-release.yaml
```diff
@@ -79,9 +79,12 @@ jobs:
     runs-on: ubuntu-24.04
     permissions:
       id-token: write
-      attestations: write
-      artifact-metadata: write
```
mihaitanasedt: This permission  `artifact-metadata` was removed and its missing from the new jobs `upload-sbom-attest-amazon-ecr`,  `upload-sbom-attest-dockerhub`.  I guess the step `attest-artifact` might need it.

.github/workflows/release.yaml
```diff
@@ -84,10 +84,85 @@ jobs:
...
+      image: ${{ matrix.url }}/${{ vars[matrix.repository] }}:${{ needs.prepare.outputs.version }}@${{ needs.build-push.outputs.digest }}
+      artifact-name: release-sbom-${{ matrix.registry }}
+      registry: ${{ matrix.url }}
+    secrets:
+      registry-token: ${{ secrets.GCR_JSON_KEY }}
+      registry-username: ${{ secrets.GCR_USERNAME }}
+
+  create-sbom-dockerhub:
+    needs: [prepare, build-push]
+    strategy:
```
gkrenn, line 222: Why do you use matrix everywhere if you only have one entry?

.github/workflows/fips-release.yaml
```diff
@@ -119,30 +122,133 @@ jobs:
...
-          skip-dirs: '/usr/share/dynatrace-operator/third_party_licenses'
-          skip-files: '/usr/local/bin/dynatrace-operator'
-      - name: Upload sbom to ${{matrix.registry}}
+      - name: Set image reference output for ${{ matrix.registry }}
+        id: image-ref
+        env:
+          IMAGE_REF: ${{ matrix.url }}/${{ vars[matrix.repository] }}:${{ needs.prepare.outputs.version }}-fips@${{ steps.image-index.outputs.digest }}
+        run: echo "${{ matrix.registry }}-image=${IMAGE_REF}" >> $GITHUB_OUTPUT
+
+  create-sbom-dockerhub:
```
mihaitanasedt: All the create sbom jobs are structurally identical,  can we merge them in a matrix?

.github/workflows/release.yaml
```diff
@@ -84,10 +84,85 @@ jobs:
     outputs:
       digest: ${{steps.build-image.outputs.digest}}
 
+  create-sbom-gcr:
```
mihaitanasedt: same heree, can we merge the jobs in a matrix ?

.github/workflows/release.yaml
```diff
@@ -84,10 +84,85 @@ jobs:
...
+            url: public.ecr.aws
+            repository: ECR_REPOSITORY
+    permissions: {}
+    uses: ./.github/workflows/create-sbom.yaml
+    with:
+      image: ${{ matrix.url }}/${{ vars[matrix.repository] }}:${{ needs.prepare.outputs.version }}@${{ needs.build-push.outputs.digest }}
+      artifact-name: release-sbom-${{ matrix.registry }}
+      registry: ${{ matrix.url }}
+    secrets:
+        registry-token: ${{ secrets.ECR_IMAGEPUSH_ROLE }}
```
mihaitanasedt: Wrong indentation

gkrenn (changes requested):

mihaitanasedt (changes requested):

.github/workflows/release.yaml
aorcholski: Copy&Paste&Cut of the `signing` job. During refactoring I wanted to preserve the image value. It seems more readable to me to have `registry`,`url`,`repository` values defined in one place.

gkrenn:
```yaml
create-sbom-dockerhub:
  needs: [prepare, build-push]
  permissions: {}
  uses: ./.github/workflows/create-sbom.yaml
  with:
    image: docker.io/${{ vars.DOCKERHUB_REPOSITORY }}:${{ needs.prepare.outputs.version }}@${{ needs.build-push.outputs.digest }}
    artifact-name: release-sbom-dockerhub
    registry: docker.io
  secrets:
    registry-token: ${{ secrets.DOCKERHUB_PASSWORD }}
    registry-username: ${{ secrets.DOCKERHUB_USERNAME }}
```

Idk, this looks better in my opinion.

.github/workflows/release.yaml
gkrenn: I guess the intention is that the workflow does not see all secrets at once and you can't programmatically access secrets in a matrix. You would need to provide them all at once. 

Which would not be a problem in my opinion as the trivy-action anyway has seen all of them after all 3 create-sbom-* jobs ran.

@aorcholski What do you think? - I matrix with all secrets provided would look nicer and does not increase the attack surface.

mihaitanasedt: AFAIK matrix jobs run in isolation and dont know about each other, so its the same as create separate ones

.github/workflows/fips-release.yaml
aorcholski: unfortunately we can't because
```
   secrets:
      registry-token: <secret name>
```
secrets can't be referenced via index (secret's name can't be resolved during runtime). One need to provide specific secrets that are passed to a reusable workflow.

.github/workflows/release.yaml
aorcholski: https://github.com/Dynatrace/dynatrace-operator/pull/6824#discussion_r3420963639

.github/workflows/fips-release.yaml
mihaitanasedt: This should work 

https://github.com/orgs/community/discussions/26302#discussioncomment-3251298
```go
create-sbom:
  needs: [prepare, build-push]
  permissions: {}
  strategy:
    matrix:
      include:
        - registry: gcr
          url: gcr.io
          repository: GCR_REPOSITORY
          token_secret: GCR_JSON_KEY
          username_secret: GCR_USERNAME
        - registry: dockerhub
          url: docker.io
          repository: DOCKERHUB_REPOSITORY
          token_secret: DOCKERHUB_PASSWORD
          username_secret: DOCKERHUB_USERNAME
        - registry: amazon-ecr
          url: public.ecr.aws
          repository: ECR_REPOSITORY
          token_secret: ECR_IMAGEPUSH_ROLE
        - registry: rhcc
          url: quay.io
          repository: RHCC_REPOSITORY
          token_secret: RHCC_PASSWORD
          username_secret: RHCC_USERNAME
  uses: ./.github/workflows/create-sbom.yaml
  with:
    image: ${{ matrix.url }}/${{ vars[matrix.repository] }}:${{ needs.prepare.outputs.version }}@${{ needs.build-push.outputs.digest }}
    artifact-name: release-sbom-${{ matrix.registry }}
    registry: ${{ matrix.url }}
  secrets:
    registry-token: ${{ secrets[matrix.token_secret] }}
    registry-username: ${{ secrets[matrix.username_secret] }}

```

gkrenn (dismissed):
From my perspective code looks fine. 

Indentation could be improved as Mihai mentioned.

.github/workflows/create-sbom.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Login to Registry
+        if: ${{ inputs.registry != '' && inputs.registry != 'public.ecr.aws' }}
+        uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4.2.0
+        with:
+          registry: ${{ inputs.registry }}
+          username: ${{ secrets.registry-username }}
+          password: ${{ secrets.registry-token }}
+      - name: Configure aws credentials
+        if: ${{ inputs.registry == 'public.ecr.aws' }}
+        uses: aws-actions/configure-aws-credentials@99214aa6889fcddfa57764031d71add364327e59 # v6.1.3
```
Incredible-O: For authentication in AWS to work (via OIDC), the `id-token` permission is required: https://github.com/aws-actions/configure-aws-credentials/blob/93823cfbdc340541ab49e4596d1da6f0536e05d0/README.md#quick-start-oidc-recommended

You could try pulling ECR images anonymously (as long as there are no rate limits?), in which case we wouldn't need to authenticate at all.

Incredible-O (changes requested):

.github/workflows/fips-release.yaml
mihaitanasedt: After explanation from Adam will resolve the comment. Using indexes, loads the whole secrets context in the runner leaking the all secrets to the jobs. Current implementation is the way to go 👍  

Couldn't find any docs about this, but the `actions/runner` code confirms it https://github.com/actions/runner/blob/16c52e389da99f4d21d142c8c87f34a1e0767823/src/Runner.Worker/ExecutionContext.cs#L1006

.github/workflows/create-sbom.yaml
aorcholski: `id-token` permission added. We shouldn't depend on no-rate-limits

.github/workflows/fips-release.yaml
aorcholski: artifact-metadata permission added to upload-sbom-attest-amazon-ecr job

.github/workflows/release.yaml
aorcholski: fixed

.github/workflows/fips-release.yaml
aorcholski: It works. Each iteration creates one of "variables" declared on `outputs` list. I tested the "idea" using my fork repo.

mihaitanasedt (dismissed):
LGTM 🚀

.github/workflows/create-sbom.yaml
```diff
@@ -0,0 +1,62 @@
...
+        default: ''
+    secrets:
+      registry-token:
+        description: token for registry authentication
+        required: false
+      registry-username:
+        description: username for registry authentication
+        required: false
+permissions:
+  id-token: write
```
Incredible-O: Maybe we _can_ drop all permissions if we pass prebuilt OCI artifacts (exported in `release.yaml`) to `create-sbom.yaml`. That way, this workflow never sees any secrets, even if trivy is compromised (again).

Adam is already working on that, we'll see if this approach is feasible. 🤞🏿

Incredible-O (changes requested):

.github/actions/copy-image-to-artifact/action.yaml
```diff
@@ -0,0 +1,57 @@
...
+    default: ''
+  registry-token:
+    description: token for registry authentication
+    required: false
+  registry-username:
+    description: username for registry authentication
+    required: false
+runs:
+  using: "composite"
+  env:
```
Incredible-O: As far as I know, composite actions don't support top-level `env` blocks (only in `steps`); see also https://github.com/orgs/community/discussions/51280#discussioncomment-8726096.
... but maybe I've missed something and GitHub has now added this feature? 🤔

.github/workflows/release.yaml
```diff
@@ -95,10 +95,172 @@ jobs:
...
+            repository: DOCKERHUB_REPOSITORY
+          - registry: amazon-ecr
+            url: public.ecr.aws
+            repository: ECR_REPOSITORY
+          - registry: rhcc
+            url: quay.io
+            repository: RHCC_REPOSITORY
+    uses: ./.github/workflows/create-sbom.yaml
+    with:
+      image: ${{ matrix.url }}/${{ vars[matrix.repository] }}-fips:${{ needs.prepare.outputs.version }}@${{ needs.build-push.outputs.fips-digest }}
```
Incredible-O: No longer required
```suggestion
```

.github/actions/copy-image-to-artifact/action.yaml
```diff
@@ -0,0 +1,57 @@
...
+        password: ${{ inputs.registry-token }}
+    - name: Configure aws credentials
+      if: ${{ inputs.registry == 'public.ecr.aws' }}
+      uses: aws-actions/configure-aws-credentials@99214aa6889fcddfa57764031d71add364327e59 # v6.1.3
+      with:
+        role-to-assume: ${{ inputs.registry-token }}
+        aws-region: us-east-1
+    - name: Login to Amazon ECR
+      if: ${{ inputs.registry == 'public.ecr.aws' }}
+      uses: aws-actions/amazon-ecr-login@fa648b43de3d4d023bcb3f89ed6940096949c419 # v2.1.5
```
Incredible-O, line 38: I'm not sure this will work 🤔 
AFAIK, the `id-token` permission is required for OIDC to work: https://github.com/aws-actions/configure-aws-credentials/blob/93823cfbdc340541ab49e4596d1da6f0536e05d0/README.md#quick-start-oidc-recommended

Incredible-O (commented):
We'll need to test this PR thoroughly 😄

.github/actions/copy-image-to-artifact/action.yaml
aorcholski: envs moved to steps

.github/workflows/release.yaml
aorcholski: removed

.github/actions/copy-image-to-artifact/action.yaml
aorcholski: id-token permission added on the caller side

.github/actions/copy-image-to-artifact/action.yaml
```diff
@@ -0,0 +1,59 @@
...
+        IMAGE: ${{ inputs.image }}
+      run: |
+        docker pull ${IMAGE}
+        docker save -o /tmp/${ARTIFACT_NAME}.tar ${IMAGE}
+    - name: Upload image artifact
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      env:
+        ARTIFACT_NAME: ${{ inputs.artifact-name }}
+      with:
+        name: ${ARTIFACT_NAME}
```
Incredible-O: I see what you're trying to do and why 😄 Unfortunately, this won't work. Environment variables defined in `env` blocks can't be used as `with` inputs (and won't be expanded).
... actually, we don't need to use intermediate environment variables here:

```suggestion
        name: ${{ inputs.artifact-name }}
```

.github/actions/copy-image-to-artifact/action.yaml
```diff
@@ -0,0 +1,59 @@
...
+        path: /tmp/${ARTIFACT_NAME}.tar
```
Incredible-O:
```suggestion
        path: /tmp/${{ inputs.artifact-name }}.tar
```

.github/actions/copy-image-to-artifact/action.yaml
aorcholski: envs removed

.github/actions/copy-image-to-artifact/action.yaml
aorcholski: envs removed

Incredible-O (dismissed):
LGTM 🚀 

Let's wait for @Dynatrace/kubernetes-operator-release to approve this PR before merging it 👍🏿

andriisoldatenko:
@aorcholski release 1.10.2 is out, let's finish this PR and close the story.

.github/workflows/release.yaml
```diff
@@ -182,20 +322,15 @@ jobs:
...
-      - name: Create sbom for ${{matrix.registry}} fips
-        id: sbom-fips
-        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
-        with:
-          image-ref: ${{ matrix.url }}/${{ vars[matrix.repository] }}:${{ needs.prepare.outputs.version }}-fips@${{ needs.build-push.outputs.fips-digest }}
-          format: 'cyclonedx'
-          output: 'fips-result.json'
-          skip-dirs: '/usr/share/dynatrace-operator/third_party_licenses'
-          skip-files: '/usr/local/bin/dynatrace-operator'
+      - name: Download SBOM artifact
```
waodim:
```suggestion
      - name: Download SBOM artifact (fips)
```

.github/workflows/release.yaml
```diff
@@ -182,20 +322,15 @@ jobs:
...
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
+        with:
+          name: release-fips-sbom-${{ matrix.registry }}
       - name: Upload sbom to ${{matrix.registry}} fips
         uses: ./.github/actions/upload-sbom
         with:
           image: ${{ matrix.url }}/${{ vars[matrix.repository] }}:${{ needs.prepare.outputs.version }}-fips@${{ needs.build-push.outputs.fips-digest }}
-          sbom: 'fips-result.json'
+          sbom: 'result.json'
```
waodim, line 333: We upload both fips and no fips to the same file, is this wrong?

Should we create a distinction?

Like `sbom: fips-sbom/result.json` / `sbom: sbom/result.json` or so or keep it as it was with `fips-result`?

aorcholski: `create-sbom` workflow is called twice for 2 different images so 2 different artifacts contains a `result.json` file.
- artifact CLASSIC is downloaded and used 
- artifact FIPS is downloaded and used

so there is no conflict between them.

.github/actions/copy-image-to-artifact/action.yaml
```diff
@@ -0,0 +1,54 @@
...
+      uses: aws-actions/amazon-ecr-login@fa648b43de3d4d023bcb3f89ed6940096949c419 # v2.1.5
+      with:
+        registry-type: public
+    - name: Save docker image
+      shell: bash
+      env:
+        ARTIFACT_NAME: ${{ inputs.artifact-name }}
+        IMAGE: ${{ inputs.image }}
+      run: |
+        docker pull "${IMAGE}"
```
waodim, line 47: qq: why use docker here and not podman?

waodim (commented):
I don't have anything further to add tbh. IMO it LGTM but maybe someone else can confirm

.github/workflows/nightly.yaml
```diff
@@ -87,11 +87,79 @@ jobs:
...
+    permissions:
+      actions: write
+    uses: ./.github/workflows/create-sbom.yaml
+    with:
+      sbom-artifact-name:  nightly-sbom
+      image-artifact-name: nightly-image
+
+  create-sbom-fips:
+    if: github.repository_owner == 'Dynatrace'
+    needs: [copy-image-to-artifact]
```
Incredible-O: Probably a typo, but we should be careful here: I think this _could_ cause race conditions. Currently, `create-sbom-fips` depends on `copy-image-to-artifact`, which produces the `nightly-image` artifact (see line 109) below

https://github.com/Dynatrace/dynatrace-operator/blob/d3d7c18b5af79828dc231dae1fa2ff85203bf4d9/.github/workflows/nightly.yaml#L105-L112

`copy-sbom-fips` should depend on `copy-image-to-artifact-fips` instead. Otherwise, if we're unlucky, the `create-sbom-fips` jobs although the artifact we actually need (`nightly-image-fips`) hasn't yet been uploaded

```suggestion
    needs: [copy-image-to-artifact-fips]
```

.github/workflows/nightly.yaml
```diff
@@ -139,23 +202,18 @@ jobs:
       - name: Sign FIPS image
         uses: ./.github/actions/sign-image
         with:
-          image: ${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_NAME }}:${{ needs.prepare.outputs.version }}-fips@${{ needs.build-push-fips.outputs.fips-digest }}
+          image: ${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_NAME }}:${{ needs.prepare.outputs.version }}-fips@${{ needs.build-push.outputs.fips-digest }}
```
Incredible-O: As far as I can tell, `build-push` only returns `digest` (and not `fips-digest`)

https://github.com/Dynatrace/dynatrace-operator/blob/d3d7c18b5af79828dc231dae1fa2ff85203bf4d9/.github/workflows/nightly.yaml#L58-L59

`build-push-fips` does return `fips-digest` though:

https://github.com/Dynatrace/dynatrace-operator/blob/d3d7c18b5af79828dc231dae1fa2ff85203bf4d9/.github/workflows/nightly.yaml#L87-L88

... , so this is probably what you want? 😄 


```suggestion
          image: ${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_NAME }}:${{ needs.prepare.outputs.version }}-fips@${{ needs.build-push-fips.outputs.fips-digest }}
```

.github/workflows/nightly.yaml
aorcholski: good :eyes:

.github/workflows/nightly.yaml
aorcholski: fixed

waodim (approved):
As said I can not add much more here. Will approve as it LGTM

Incredible-O (approved):
🚢

PR #6828, opened 2026-06-16
fix: gate pgc volume attach

svfoxat:
## Description

followup to https://dt-rnd.atlassian.net/browse/ICP-5529 because it broke e2e tests.

In some conditions kubelet fails to mount the pgc volume mount when the `declarative.cbor` key is not available in the `bootstrapper-config`, e.g. when there is no AG with `kubernetes-monitoring` like in the following e2e tests.
  - `test/e2e/features/hostmonitoring/without_csi.go`
  - `test/e2e/features/resourceattributes/logmon_only.go`

This PR changes the behaviour to a more defensive, "only attach the volume when the key is available in the secret" approach instead of relying only on the optional flag.

## How can this be tested?
     make go/test                                                                                                           
     make go/lint                                                                                                             
     make test/e2e/standard


pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
```diff
@@ -90,30 +90,33 @@ type builder struct {
 	hostInjectSpec *oneagent.HostInjectSpec
 	clusterID      string
 	deploymentType string
+	pgcReady       bool
```
gkrenn:
```suggestion
	processGroupConfigReady       bool
```

It's not an obvious abbreviation. Let's keep it readible.

pkg/controllers/dynakube/oneagent/oneagent_reconciler.go
```diff
@@ -311,6 +311,19 @@ func (r *Reconciler) getOneagentPods(ctx context.Context, dk *dynakube.DynaKube,
 }
 
 func (r *Reconciler) buildDesiredDaemonSet(ctx context.Context, dk *dynakube.DynaKube) (*appsv1.DaemonSet, error) {
+	var configHash string
+
+	if bootstrapperconfig.NeedsPGC(dk) {
```
andriisoldatenko: I prefer to move this code 316:323 lines directly to:
```
func (b *builder) BuildDaemonSet(ctx context.Context) (*appsv1.DaemonSet, error) {
	ctx, _ = logd.NewFromContext(ctx, "oneagent-daemonset")
	dk := b.dk

	podSpec, err := b.podSpec(ctx)
	if err != nil {
		return nil, err
	}

	versionLabelValue := dk.OneAgent().GetVersion()

	appLabels := k8slabel.NewAppLabels(k8slabel.OneAgentComponentLabel, dk.Name,
		b.deploymentType, versionLabelValue)
	labels := maputils.MergeMap(
		appLabels.BuildLabels(),
		b.hostInjectSpec.Labels,
	)

	templateAnnotations := map[string]string{
		appArmorAnnotation:                appArmorUnconfined,
		webhook.AnnotationDynatraceInject: "false",
		annotationTenantTokenHash:         dk.Status.OneAgent.ConnectionInfo.TenantTokenHash,
		annotationEnableDaemonSetEviction: "false",
	}
```
you have dk, you read secret when you build DS, and if key is there you add annotation if not, do nothing


it's generic code, and you have check so you are protected, it won't be added to `ds, err = daemonset.NewClassicFullStack(dk, r.clusterID).BuildDaemonSet(ctx)`


then you don't need to add pgcReady to top level interface `daemonset.NewCloudNativeFullStack(dk, r.clusterID, pgcReady).BuildDaemonSet(ctx)` here, we can do it inside

andriisoldatenko (changes requested):

pkg/controllers/dynakube/oneagent/daemonset/daemonset.go
svfoxat: I'll remove this property in favor of @andriisoldatenko's suggestion.

pkg/controllers/dynakube/oneagent/daemonset/volumes.go
```diff
@@ -37,7 +37,7 @@ func prepareVolumeMounts(dk *dynakube.DynaKube) []corev1.VolumeMount {
 		volumeMounts = append(volumeMounts, getHTTPProxyMount())
 	}
 
-	if bootstrapperconfig.NeedsPGC(dk) {
+	if bootstrapperconfig.NeedsPGC(dk) && pgcReady {
```
gkrenn: Just a nitpick, but all the appreviations here could also be improved

gkrenn:
Just ran it in cloudNativeFullStack => works for me

andriisoldatenko (approved):

gkrenn (approved):

svfoxat:
e2e green: https://github.com/Dynatrace/dynatrace-operator/actions/runs/27674851034

PR #6845, opened 2026-06-18
Prevent mounting invalid user-provided volumes

avorima:
## Description

Fail Pod injection when a volume with matching name, but conflicting configuration is already present.
Previously, we only looked at the name which could lead to a hostPath volume being mounted instead of an emptyDir.
This change adds checks to the volume type as well as some fields, e.g. the secret name in the case of a secret volume.

Add two new annotations related to failed injection:
* `volume.dynatrace.com/injected`
* `volume.dynatrace.com/reason`

ICP-6281

## How can this be tested?

* Run tests
* Deploy operator and check that invalid volumes are not mounted and annotations are present

### DIFF
--- a/pkg/util/kubernetes/fields/k8svolume/volumes.go
+++ b/pkg/util/kubernetes/fields/k8svolume/volumes.go
@@ -1,12 +1,11 @@
 package k8svolume
 
 import (
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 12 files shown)


jokuniew (approved):
deployed pod with 
```yaml
  volumes:
    - name: dynatrace-config
      hostPath:
        path: /tmp
```
got annotation 

Image

PR #6853, opened 2026-06-19
Add fine grained auth option and logic to helm chart, use it for logmonitoring clusterrole.

s-quatres:
## Description
[ICP-4010](https://dt-rnd.atlassian.net/browse/ICP-4010) We already merged a change that adds nodes/pods to the existing ClusterRole for logmonitoring. However, having both nodes/proxy and nodes/pods is not minimal in terms of security, so this change uses nodes/pods from k8s versions >= 1.36 or if the new helm value (in this PR) `kubeletFineGrainedAuthz` is set, otherwise nodes/proxy. This `kubeletFineGrainedAuthz` setting will be needed for GKE-Autopilot (no nodes/proxy allowed), GKE AP detection is not reliable e.g. current logic for AllowLists also installs them on standard GKE clusters.

## How can this be tested?
The new tests in the PR can be run from the branch, or helm can be invoked with/without setting. One could also test a pre and post 1.36 k8s. cluster.

[ICP-4010]: https://dt-rnd.atlassian.net/browse/ICP-4010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/helm/chart/default/values.yaml
```diff
@@ -273,6 +273,7 @@ csidriver:
         memory: 30Mi
 
 rbac:
+  kubeletFineGrainedAuthz: ""
```
avorima, line 276: should use `false` to document the intended type

config/helm/chart/default/templates/_platform.tpl
```diff
@@ -34,6 +34,23 @@ Auto-detect whether or not we need allowlisting for logagent and csi-driver
...
 
+{{/*
+Determine if RBAC should use KubeletFineGrainedAuthz (nodes/pods) instead of nodes/proxy.
+Auto-enabled on k8s >= 1.36 (KubeletFineGrainedAuthz GA). GKE Autopilot on older k8s also
+requires nodes/pods (nodes/proxy is forbidden) but cannot be reliably detected from Helm
+capabilities, so it must be enabled there via the override below.
+Override with .Values.rbac.kubeletFineGrainedAuthz: true/false
+*/}}
+{{- define "dynatrace-operator.kubeletFineGrainedAuthz" -}}
+    {{- if kindIs "bool" .Values.rbac.kubeletFineGrainedAuthz -}}
```
avorima, line 45: nit: reflection is a bit overkill here. any empty value ("", nil, false) will go into the else branch

avorima: discussed offline: we need to be able to out out on >1.36

config/helm/chart/default/templates/Common/logmonitoring/clusterrole-logmonitoring.yaml
```diff
@@ -22,8 +22,11 @@ rules:
 - apiGroups:
     - ""
   resources:
-    - nodes/proxy
+{{- if eq (include "dynatrace-operator.kubeletFineGrainedAuthz" .) "true" }}
```
avorima: This `eq` isn't actually needed, since the function returns nothing otherwise. See https://github.com/Dynatrace/dynatrace-operator/pull/6858

s-quatres: Changed as requested.

avorima (approved):

PR #6782, opened 2026-05-29
feat: configurable requeueAfter for dynakube controller

svfoxat:
## Description

  https://dt-rnd.atlassian.net/browse/ICP-4102

  ### Changes

  - `DT_DEFAULT_REQUEUE_AFTER` env variable now controls the dynakube controller requeue interval
    - Falls back to existing default of `15m` when not set
    - Accepted range: `1m` – `1h` (feedback welcome)
  - Added `operator.requeueAfter` to Helm chart `values.yaml` (empty by default)
  
  - additionally changed `GetDTClientCacheCleanInterval` to accept `context.Context` instead of `log logd.Logger` (consistent with new `GetDefaultRequeueAfter`)

  ---

  ## How to Test

  ### Automated

  ```sh
  make go/test
  make go/lint
  make test/helm
  ``` 

### Dev Cluster

  **1. Deploy**

  ```sh
  make build && make deploy
```

  2. Apply an example DynaKube, then verify env var is absent by default
```sh
  kubectl get deployment dynatrace-operator -n dynatrace \
    -o jsonpath='{.spec.template.spec.containers[0].env}' \
    | jq '.[] | select(.name=="DT_DEFAULT_REQUEUE_AFTER")'
```
 **Expected: no output**

  3. Set a custom requeue interval
```sh
  helm upgrade dynatrace-operator ./config/helm/chart/default \
    --reuse-values \
    --set operator.requeueAfter="30m" \
    -n dynatrace
```

  4. Verify env var is present
```sh
  kubectl get deployment dynatrace-operator -n dynatrace \
    -o jsonpath='{.spec.template.spec.containers[0].env}' \
    | jq '.[] | select(.name=="DT_DEFAULT_REQUEUE_AFTER")'
```
 Expected: `{ "name": "DT_DEFAULT_REQUEUE_AFTER", "value": "30m" }`

PR #6861, opened 2026-06-23
feat: configurable cert-manager requeue and expiry intervals

svfoxat:
## Description

https://dt-rnd.atlassian.net/browse/ICP-1197

Extends the webhook certificate controller with configurable cert durations and requeue interval, similar to what was done for the dynakube controller in #6782.

- Added 4 env variables / helm values to control webhook cert behavior:
  - **DT_WEBHOOK_CERTS_REQUEUE_AFTER / webhook.certs.requeueAfter** — requeue interval when certs are up to date (default 3h, range 5m–11h)
  - **DT_WEBHOOK_CERTS_RENEWAL_THRESHOLD / webhook.certs.renewalThreshold** — how early before expiry certs get renewed (default 12h, range 12h–720h)
  - **DT_WEBHOOK_CERTS_SERVER_DURATION / webhook.certs.serverDuration** — server cert validity (default 168h, range 24h–8760h)
  - **DT_WEBHOOK_CERTS_ROOT_DURATION / webhook.certs.rootDuration** — CA root cert validity (default 8760h, range 168h–87600h)
  - All fall back to their defaults when unset or out of range
- Enforces ordering constraints at startup: `rootDuration > serverDuration > renewalThreshold > requeueAfter`. `renewalThreshold` has a minimum of 12h (≥ 2× the cert watcher's 6h poll interval) to guarantee the watcher loads renewed certs before the old ones expire.

> Note: Changing cert duration values does not immediately re-issue existing valid certificates — the new duration takes effect on the next renewal cycle. To force immediate re-issuance, delete the webhook cert secret:
> `kubectl delete secret dynatrace-webhook-certs -n dynatrace`
> The cert controller will regenerate it on the next reconcile with the new durations.

## How can this be tested?

`make go/lint`
`make go/test`
`make test/helm`

**In-cluster — verify defaults (no env vars set):**
```
kubectl get deployment dynatrace-operator -n dynatrace \
  -o jsonpath='{.spec.template.spec.containers[0].env}' \
  | jq '[.[] | select(.name | startswith("DT_WEBHOOK_CERTS"))]'
```
Expected: empty array

**Set a custom server cert duration:**
```
helm upgrade dynatrace-operator ./config/helm/chart/default \
  --reuse-values \
  --set webhook.certs.serverDuration="48h" \
  -n dynatrace

kubectl get deployment dynatrace-operator -n dynatrace \
  -o jsonpath='{.spec.template.spec.containers[0].env}' \
  | jq '.[] | select(.name=="DT_WEBHOOK_CERTS_SERVER_DURATION")'
```
Expected: `{ "name": "DT_WEBHOOK_CERTS_SERVER_DURATION", "value": "48h" }`

**Verify issued cert duration:** (see note above)
```
kubectl get secret dynatrace-webhook-certs -n dynatrace \
  -o jsonpath='{.data.tls\.crt}' \
  | base64 -d \
  | openssl x509 -noout -dates
```
Expected: notAfter should reflect the configured serverDuration from the current time.

> Note: If debug logs are enabled, more information will be available in the logs for these variable
```
{"level":"debug","logger":"k8senv","msg":"no custom env set, using default","env":"DT_WEBHOOK_CERTS_SERVER_DURATION","default":604800}

{"level":"debug","logger":"webhook-cert-controller","msg":"webhook cert controller config","requeueAfter":3600,"renewalThreshold":43200,"serverDuration":172800,"rootDuration":31536000}

{"level":"info","logger":"k8senv","msg":"duration not in allowed range, using default","env":"DT_WEBHOOK_CERTS_SERVER_DURATION","value":60,"min":86400,...}

{"level":"info","ts":"2026-06-25T13:23:14.394Z","logger":"k8senv","msg":"using custom duration","env":"DT_WEBHOOK_CERTS_SERVER_DURATION","value":176400}      
```


pkg/controllers/certificates/webhook_cert_controller.go
```diff
@@ -49,23 +53,63 @@ func InitReconcile(ctx context.Context, clt client.Client, namespace string) err
...
+			serverDuration, renewalThreshold, k8senv.WebhookCertsServerDurationEnvVar, renewalThreshold)
+	}
+
+	if rootDuration <= serverDuration {
+		return nil, fmt.Errorf("root cert duration (%s) must exceed server cert duration (%s); set %s to a value greater than %s",
+			rootDuration, serverDuration, k8senv.WebhookCertsRootDurationEnvVar, serverDuration)
+	}
+
+	renewalWindow := serverDuration - renewalThreshold
+	if requeueAfter >= renewalWindow {
```
avorima: **Secondary: this guardrail checks the wrong quantity.** `renewalWindow := serverDuration - renewalThreshold` is the *fresh* period, not the window in which renewal can happen. Renewal triggers only in the last `renewalThreshold` before expiry, so to guarantee a reconcile lands there the real constraint is `requeueAfter < renewalThreshold`. As written, `serverDuration=24h, renewalThreshold=1h, requeueAfter=12h` all pass (`12h < 23h`) yet the 1h renewal window is easily skipped by the 12h requeue cadence — the cert expires before it's renewed. Suggest guarding against `renewalThreshold` (ideally with a safety factor, e.g. `requeueAfter <= renewalThreshold/2`), and adding a unit test in `webhook_cert_controller_test.go` covering this scenario (short `serverDuration`/`renewalThreshold` with a large `requeueAfter`) so the guardrail rejects it.

avorima (commented):
**The webhook's cert watcher polls on a hardcoded 6h cadence that this PR's configurable durations can outrun**

This PR makes the controller-side cert durations configurable, but the *consumer* — the cert watcher running inside the webhook pod (`cmd/webhook/certificates/watcher.go`) — is untouched and still hardcoded:

```go
certificateRenewalInterval = 6 * time.Hour   // watcher.go:23

<-time.After(certificateRenewalInterval)      // :56  poll interval
...
ValidateCertificateExpiration(ctx, secret.Data[ServerCert], certificateRenewalInterval, time.Now())  // :89  staleness threshold
```

The watcher is what actually loads the renewed cert from the secret onto the disk the webhook TLS server serves from. The same `6h` constant is used both as the poll interval and as the staleness threshold.

The old system was safe because of an implicit ordering between hardcoded values:

```
webhook poll interval (6h) < renewalThreshold (12h) < serverDuration (168h)
```

The controller renewed the secret ≥12h before expiry, and the webhook — polling at most every 6h — always loaded the new cert with margin. This PR lets `renewalThreshold` drop to **1h** (min) and `serverDuration` to **24h** (min), with no guardrail aware of the watcher's 6h. Two failure modes result:

1. **Renewed cert not loaded before expiry (timely-refresh failure).** With `renewalThreshold = 1h`, the controller only writes the renewed cert into the secret when the old one is within 1h of expiry. The watcher may not poll for up to 6h, so the fresh cert sits in the secret while the pod keeps serving the old on-disk cert — which expires first. The webhook then serves an expired TLS cert and the API server rejects mutating/validating calls (injection + DynaKube validation fail) until the next 6h tick.

2. **Watcher rejects certs the controller considers current.** With `renewalThreshold < 6h`, there's a window where the controller reports the cert "up to date" (e.g. 5h left) while `ValidateCertificateExpiration` at :89 treats it as outdated (<6h remaining) and `updateCertificatesFromSecret` returns an error. Steady-state that's just noisy logging (files were already written at :83-87), but at startup `WaitForCertificates` (:117-132) only breaks on a nil error, so it burns the full 5-minute wait before starting the watcher.

The controller-side guardrails in `newWebhookCertificateController` give a false sense that these durations are validated end-to-end — they aren't; the serving side is unaware of them.

**Suggested fix:** enforce a floor of `renewalThreshold ≥ 2 × certificateRenewalInterval` (i.e. ≥ 12h given the watcher's fixed 6h poll) so the watcher's poll cadence can always catch a renewal before the old cert expires. This keeps the watcher untouched while closing the gap the new configurability opens.

config/helm/chart/default/values.yaml
```diff
@@ -130,6 +130,11 @@ webhook:
   mutatingWebhook:
     failurePolicy: Ignore
     timeoutSeconds: 10
+  certs:
+    rootDuration: "" # CA root certificate validity duration, defaults to 8760h (365 days), min: 168h (7 days), max: 87600h (10 years)
```
avorima: mention the enforced ordering "must be greater than ..." here and in schema.yaml. want to be extra clear here

pkg/controllers/certificates/webhook_cert_controller.go
```diff
@@ -49,23 +56,67 @@ func InitReconcile(ctx context.Context, clt client.Client, namespace string) err
...
+		Complete(controller)
 }
 
-func newWebhookCertificateController(clt client.Client, apiReader client.Reader) *WebhookCertificateController {
-	return &WebhookCertificateController{
-		client:    clt,
-		apiReader: apiReader,
+func newWebhookCertificateController(clt client.Client, apiReader client.Reader) (*WebhookCertificateController, error) {
+	ctx := context.Background()
+	_, log := logd.NewFromContext(ctx, "webhook-cert-controller")
```
avorima: is this omitting of the webhook-cert-controller logger name from k8senv calls intentional?

svfoxat: Ah, no, my error. `webhook-cert-controller.k8senv` is the wanted loggername.

avorima (dismissed):

avorima (approved):

PR #6866, opened 2026-06-23
fix skip changes when new release branch and base_sha: 0000000000000000000000000000000000000000

andriisoldatenko:
## Description

Resolves https://github.com/Dynatrace/dynatrace-operator/actions/runs/28031216027/job/83021282950

Fix with copilot but validated: 
> Make base_sha conditional so it is only passed when it is valid for the event type, and avoid using merge_group.head_sha as a base commit. For merge queues, use the merge group’s base SHA if available, otherwise let the action determine the comparison automatically.

> On a push event, that resolves to github.event.before. When the push creates a new branch, rewrites history, or otherwise references a commit GitHub can’t resolve through the API, changed-files fails while trying to fetch that commit object.

## How can this be tested?

n/a

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -66,7 +66,7 @@ jobs:
           files: |
             .github/workflows/**
             .github/actions/**
-          base_sha: ${{ github.event.pull_request.base.sha || github.event.merge_group.head_sha || github.event.before }}
+          base_sha: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.event.merge_group.head_sha || github.event_name == 'push' && github.event.before != '0000000000000000000000000000000000000000' && github.event.before || '' }}
       - name: Detect Go file changes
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -66,43 +66,43 @@ jobs:
           files: |
             .github/workflows/**
             .github/actions/**
-          base_sha: ${{ github.event.pull_request.base.sha || github.event.merge_group.head_sha || github.event.before }}
+          base_sha: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.event.merge_group.head_sha || github.event_name == 'push' && github.event.before != '0000000000000000000000000000000000000000' && github.event.before || '' }}
```
avorima, line 69: github runner may understand this, but please add some parenthesis to group things a bit

andriisoldatenko:
closing AI generated PR, and will file a bug to fix properly.

PR #6874, opened 2026-06-24
Deprecate link extraction from codemodules image

avorima:
## Description

The implementation of our codemodules image extraction has a few features that are not actually used, namely unpacking hardlinks and symlinks.
The relevant files that are needed for injection are regular files.

Add a global config option DT_EXTRACT_CODEMODULES_IMAGE_LINKS that controls whether unpacking hardlinks and symlinks from images archives is enabled.
The feature will be disabled by default. Planned to be removed at a later point.

ICP-6190

## How can this be tested?

Make sure to clear your node filesystem first to ensure that you get an image download.

* Deploy DynaKube with codemodules, tenant registry/public registry/url download
* Check that applications get injected and data appears on the tenant.

### DIFF
--- a/pkg/injection/codemodule/installer/zip/zip.go
+++ b/pkg/injection/codemodule/installer/zip/zip.go
@@ -35,7 +35,7 @@ import (
 // Note: O_NOFOLLOW only guards the final path component, not intermediate
 // directory components. Fully confining resolution would require openat2 with
 // RESOLVE_NO_SYMLINKS (or filepath-securejoin).
-const hardenedOpenFileFlags = os.O_CREATE | os.O_RDWR | os.O_TRUNC | unix.O_NOFOLLOW
+const hardenedOpenFileFlags = os.O_CREATE | os.O_WRONLY | os.O_TRUNC | unix.O_NOFOLLOW
 
 func (extractor OneAgentExtractor) ExtractZip(ctx context.Context, sourceFile *os.File, targetDir string) error {
 	ctx, log := logd.NewFromContext(ctx, "oneagent-zip")
--- a/pkg/injection/codemodule/installer/zip/zip.go
+++ b/pkg/injection/codemodule/installer/zip/zip.go
@@ -12,7 +12,30 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
 	"github.com/klauspost/compress/zip"
 	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
 )
+
+// hardenedOpenFileFlags is used whenever a regular file is created during code
+// module extraction.
+//
+// Attack vector: the CSI driver extracts code module images as root onto a
+// shared hostPath (the CSI data dir) that is reused across DynaKubes. A
+// malicious image layer can contain symlink entries that, once created, form a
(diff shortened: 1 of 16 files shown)


pkg/injection/codemodule/installer/zip/zip.go
```diff
@@ -12,8 +12,31 @@ import (
...
+// unix.O_NOFOLLOW makes os.OpenFile refuse (ELOOP) to follow a symlink at the
+// final path component, so a pre-existing or concurrently-planted symlink at the
+// destination cannot redirect the written bytes. It is applied unconditionally -
+// on top of skipping symlink/hardlink archive entries by default - so the
+// protection still holds when link extraction is explicitly opted in.
+//
+// Note: O_NOFOLLOW only guards the final path component, not intermediate
+// directory components. Fully confining resolution would require openat2 with
+// RESOLVE_NO_SYMLINKS (or filepath-securejoin).
+const hardenedOpenFileFlags = os.O_CREATE | os.O_RDWR | os.O_TRUNC | unix.O_NOFOLLOW
```
waodim: Should we use `os.O_WRONLY` instead of `os.O_RDWR`? Only thing we do is copy, if I did not miss something

pkg/webhook/mutation/pod/handler/injection/init.go
```diff
@@ -46,10 +47,12 @@ func (h *Handler) createInitContainerBase(pod *corev1.Pod, dk dynakube.DynaKube)
...
 		SecurityContext: securityContextForInitContainer(pod, dk, h.isOpenShift),
 		Resources:       defaultInitContainerResources(),
-		Args:            []string{bootstrapper.Use},
+		Args:            append([]string{bootstrapper.Use}, arg.ConvertArgsToStrings(args)...),
 	}
 
-	initContainer.Args = append(initContainer.Args, arg.ConvertArgsToStrings(args)...)
+	if k8senv.GetDTExtractCodeModulesImageLinks(ctx) {
+		initContainer.Env = []corev1.EnvVar{{Name: k8senv.DTExtractCodeModulesImageLinksEnvVar, Value: "true"}}
+	}
```
0sewa0: this is only needed for the ZIP download, I would put it [here](https://github.com/Dynatrace/dynatrace-operator/blob/b6ecaa4635ed70d7fd18e49128933e26a5f660da/pkg/webhook/mutation/pod/mutator/oneagent/init.go#L71-L80) instead, to follow the current pattern.

0sewa0 (commented):
it works, just a small comment

A question in general:

According [your latest comment](https://dt-rnd.atlassian.net/browse/ICP-6190?focusedCommentId=4662330) we wanted to straight out remove it. Why decided to only deprecate instead?
If we go with deprecation, pls don't forget to create a followup to remove it,

avorima:
> it works, just a small comment
> 
> A question in general:
> 
> According [your latest comment](https://dt-rnd.atlassian.net/browse/ICP-6190?focusedCommentId=4662330) we wanted to straight out remove it. Why decided to only deprecate instead? If we go with deprecation, pls don't forget to create a followup to remove it,

It was decided to be safer than just outright removing it. We'll give it two releases 
ICP-6839

0sewa0 (approved):

PR #6891, opened 2026-06-25
Sanitize OneAgent and Logmon args

avorima:
## Description

Add validation for no-proxy feature flag to prevent invalid characters from being added in the command-line arguments.
Sanitize all OneAgent and logmon command-line arguments by removing invalid characters.

ICP-6481

## How can this be tested?

* Try using invalid characters for oneagent no-proxy, oneagent args, logmond args -> blocked
* Disable webhook and provide invalid characters in the args of oneagent and logmon -> sanitized

### DIFF
--- a/pkg/api/validation/dynakube/featureflag_test.go
+++ b/pkg/api/validation/dynakube/featureflag_test.go
@@ -206,3 +206,23 @@ func TestIsNodeImagePullWithoutCSI(t *testing.T) {
 		})
 	}
 }
+
+func TestInvalidNoProxy(t *testing.T) {
+	invalidChars := []rune{
+		'\n',
+		'\t',
+		'\r',
+		'\x00',
+	}
+
+	for _, c := range invalidChars {
+		t.Run(fmt.Sprintf("reject %c", c), func(t *testing.T) {
+			dk := &dynakube.DynaKube{
+				ObjectMeta: metav1.ObjectMeta{Name: "dynakube", Annotations: map[string]string{exp.NoProxyKey: "foo" + string(c) + "bar"}},
+				Spec:       dynakube.DynaKubeSpec{APIURL: testAPIURL},
+			}
+
+			assertDenied(t, []string{errorInvalidNoProxy}, dk)
+		})
+	}
+}
--- a/pkg/api/validation/dynakube/featureflag_test.go
+++ b/pkg/api/validation/dynakube/featureflag_test.go
@@ -208,21 +208,10 @@ func TestIsNodeImagePullWithoutCSI(t *testing.T) {
 }
 
 func TestInvalidNoProxy(t *testing.T) {
-	invalidChars := []rune{
-		'\n',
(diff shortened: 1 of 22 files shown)


pkg/util/sanitize/sanitize.go
```diff
@@ -0,0 +1,42 @@
+// Package sanitize provides functions to sanitize user input.
+package sanitize
+
+import (
+	"strings"
+	"sync"
+)
+
+var invalidCommandLineChars = [...]rune{
```
andriisoldatenko, line 9: @0sewa0 one more runes usage in house :)

pkg/controllers/dynakube/logmonitoring/daemonset/args.go
```diff
@@ -28,8 +29,8 @@ func getInitArgs(dk dynakube.DynaKube) []string {
 
 	attrs := dk.GetResourceAttributes()
 	for _, key := range slices.Sorted(maps.Keys(attrs)) {
-		baseArgs = append(baseArgs, fmt.Sprintf("-p %s=%s", key, attrs[key]))
+		baseArgs = append(baseArgs, fmt.Sprintf("-p %s=%s", sanitize.CommandLineArg(key), sanitize.CommandLineArg(attrs[key])))
 	}
 
-	return append(baseArgs, dk.LogMonitoring().Template().Args...)
+	return append(baseArgs, sanitize.CommandLineArgs(dk.LogMonitoring().Template().Args)...)
```
waodim, line 35: Silently stripping chars without logging means a user with a templating error (for example a trailing newline in a value or so) has no way to know their arg was modified. Should we add a `log.Warn("sanitized invalid characters from arg")`? Would probably help with debugging. Do you think this is worth it?

I am just asking because no-proxy gets loudly blocked and logmon/oneagent args are changed silently.

pkg/api/validation/dynakube/featureflag_test.go
```diff
@@ -206,3 +206,12 @@ func TestIsNodeImagePullWithoutCSI(t *testing.T) {
 		})
 	}
 }
+
+func TestInvalidNoProxy(t *testing.T) {
+	dk := &dynakube.DynaKube{
+		ObjectMeta: metav1.ObjectMeta{Name: "dynakube", Annotations: map[string]string{exp.NoProxyKey: "foo\nbar"}},
```
waodim: probably would be worth it to table test the other possibilities as well, wdyt?

pkg/controllers/dynakube/logmonitoring/daemonset/args.go
avorima: I'll add validation for args so that it matches the sanitization in the controllers, but we should keep both for defense in depth. Adding logging for these edge cases complicates things, so I'd prefer to just document the behavior. A real user should never get sanitized

waodim (approved):

PR #6896, opened 2026-06-26
Add usage of JSON metadata annoation to OTelC config.

StefanHauth:
## Description

The OpenTelemetry collector spawned for telemetry ingest shall utilize attributes provided in the JSON block in `metadata.dynatrace.com`.

[JIRA](https://dt-rnd.atlassian.net/browse/ICP-6439)

## How can this be tested?

* Deploy a DK with `.spec.telemetryIngest` and `.spec.metadataEnrichment` (or full CNFS) enabled
* Create a namespace that is annotated with `metadata.dynatrace.com/foo: bar`
* Create a workload in that namespace that emits traces, metrics and/or logs to Dynatrace via the telemetry ingest mechanism
* Check in tenant if the ingested data has the `foo: bar` attribute.

Here are some sample pods to generate data:
[telemetrygen-pods.yaml](https://github.com/user-attachments/files/29372787/telemetrygen-pods.yaml)

### DIFF
--- a/pkg/otelcgen/processors.go
+++ b/pkg/otelcgen/processors.go
@@ -63,6 +63,11 @@ func (c *Config) buildProcessors() map[component.ID]component.Config {
 						"key_regex": "metadata.dynatrace.com/(.*)",
 						"tag_name":  "$$1",
 					},
+					{
+						"from":     "pod",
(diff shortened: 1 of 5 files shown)


avorima (approved):
Annotation appears in the distributed tracing app

PR #6908, opened 2026-06-30
fix: Trigger CI on new branch push

waodim:
## Description

When a new branch is pushed for the first time (e.g. `release-1.10` or a dev branch), `github.event.before` is `0000000000000000000000000000000000000000` (zero SHA). This caused `tj-actions/changed-files` to fail, which meant the `detect-changes` job never completed and all downstream CI jobs were blocked.

The fix detects a new branch push early and skips the `changed-files` steps in that case (since there's nothing to diff). A new `is_new_branch` output is propagated to all downstream job conditions so CI runs as expected.

[ICP-6541](https://dt-rnd.atlassian.net/browse/ICP-6541)

## How can this be tested?

Push a new branch with no additional commits on top:

```sh
git push origin HEAD:<name>
```
The CI workflow should trigger, the Detect Changes job should succeed, and downstream jobs (tests, linting, helm, etc.) should run instead of being skipped.

[ICP-6541]: https://dt-rnd.atlassian.net/browse/ICP-6541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,6 +33,7 @@ jobs:
       prerequisites_changed: ${{ steps.prerequisites.outputs.any_changed == 'true' && 'true' || '' }}
       ci_changed: ${{ steps.ci.outputs.any_changed == 'true' && 'true' || '' }}
       is_protected_branch: ${{ steps.is-protected-branch.outputs.is_protected_branch == 'true' && 'true' || '' }}
+      is_new_branch: ${{ steps.is-new-branch.outputs.is_new_branch == 'true' && 'true' || '' }}
       is_snyk_authorized: ${{ steps.is-snyk-authorized.outputs.is_authorized == 'true' && 'true' || '' }}
       is_docker_authorized: ${{ steps.is-docker-authorized.outputs.is_authorized == 'true' && 'true' || '' }}
       markdown_changed: ${{ steps.markdown.outputs.any_changed == 'true' && 'true' || '' }}
@@ -43,6 +44,14 @@ jobs:
           fetch-depth: ${{ github.event_name == 'merge_group' && '0' || '2' }}
           ref: ${{ github.event.pull_request.head.sha || github.event.merge_group.head_sha || github.sha }}
           persist-credentials: false
+      - name: Check if new branch push
+        id: is-new-branch
+        run: |
+          if [[ "${{ github.event.before }}" == "0000000000000000000000000000000000000000" ]]; then
+            echo "is_new_branch=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "is_new_branch=false" >> "$GITHUB_OUTPUT"
+          fi
       - name: Check if it is a protected branch
         id: is-protected-branch
         run: |
@@ -60,6 +69,7 @@ jobs:
         run: |
           echo "is_authorized=${{ secrets.QUAY_USERNAME != '' && secrets.QUAY_PASSWORD != '' }}" >> $GITHUB_OUTPUT
       - name: Detect CI file changes
+        if: steps.is-new-branch.outputs.is_new_branch != 'true'
         uses: tj-actions/changed-files@9426d40962ed5378910ee2e21d5f8c6fcbf2dd96 # v47.0.6
         id: ci
         with:
@@ -68,6 +78,7 @@ jobs:
             .github/actions/**
           base_sha: ${{ github.event.pull_request.base.sha || github.event.merge_group.head_sha || github.event.before }}
       - name: Detect Go file changes
+        if: steps.is-new-branch.outputs.is_new_branch != 'true'
         uses: tj-actions/changed-files@9426d40962ed5378910ee2e21d5f8c6fcbf2dd96 # v47.0.6
         id: go-files
         with:
@@ -76,6 +87,7 @@ jobs:
             go.mod
           base_sha: ${{ github.event.pull_request.base.sha || github.event.merge_group.head_sha || github.event.before }}
       - name: Detect Dockerfile changes
+        if: steps.is-new-branch.outputs.is_new_branch != 'true'
         uses: tj-actions/changed-files@9426d40962ed5378910ee2e21d5f8c6fcbf2dd96 # v47.0.6
         id: dockerfile
         with:
@@ -83,6 +95,7 @@ jobs:
             Dockerfile
           base_sha: ${{ github.event.pull_request.base.sha || github.event.merge_group.head_sha || github.event.before }}
       - name: Detect Helm config changes
+        if: steps.is-new-branch.outputs.is_new_branch != 'true'
         uses: tj-actions/changed-files@9426d40962ed5378910ee2e21d5f8c6fcbf2dd96 # v47.0.6
         id: helm
         with:
@@ -90,6 +103,7 @@ jobs:
             config/**
           base_sha: ${{ github.event.pull_request.base.sha || github.event.merge_group.head_sha || github.event.before }}
       - name: Detect prerequisites.mk changes
+        if: steps.is-new-branch.outputs.is_new_branch != 'true'
         uses: tj-actions/changed-files@9426d40962ed5378910ee2e21d5f8c6fcbf2dd96 # v47.0.6
         id: prerequisites
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -33,6 +33,7 @@ jobs:
       prerequisites_changed: ${{ steps.prerequisites.outputs.any_changed == 'true' && 'true' || '' }}
       ci_changed: ${{ steps.ci.outputs.any_changed == 'true' && 'true' || '' }}
       is_protected_branch: ${{ steps.is-protected-branch.outputs.is_protected_branch == 'true' && 'true' || '' }}
+      is_new_branch: ${{ steps.is-new-branch.outputs.is_new_branch == 'true' && 'true' || '' }}
```
andriisoldatenko: just our of curiosity, is it possible to calculate it directly here, since is job level and should be avaialable here:
>"${{ github.event.before }}" == "0000000000000000000000000000000000000000"

waodim: Should be possible actually: https://github.com/Dynatrace/dynatrace-operator/pull/6908/commits/37f0034d45fc3e8c0067c9e1d0c2ab3a020e4e25

.github/workflows/ci.yaml
andriisoldatenko, line 23: calude suggested to define it here:
```
env:
  IS_FIRST_PUSH: ${{ github.event.before == '0000000000000000000000000000000000000000' }}
```

and reuse it

.github/workflows/ci.yaml
```diff
@@ -60,6 +61,7 @@ jobs:
         run: |
           echo "is_authorized=${{ secrets.QUAY_USERNAME != '' && secrets.QUAY_PASSWORD != '' }}" >> $GITHUB_OUTPUT
       - name: Detect CI file changes
+        if: github.event.before != '0000000000000000000000000000000000000000'
```
andriisoldatenko:
```suggestion
        if: env.IS_FIRST_PUSH
```

.github/workflows/ci.yaml
waodim: Actually at first I liked the idea but I would tend towards keeping the inline checks as is though, the env var approach has a couple of issues:

the step guards currently mean "run when NOT a new branch" (!= '000...000'). Replacing them with `if: env.IS_FIRST_PUSH` would flip that, running the steps only on new branches.

`if: env.IS_FIRST_PUSH` would always run the step regardless of value because `IS_FIRST_PUSH` would be the string "false" (not a boolean), and non-empty strings are true in github actions `if: expressions` AFAIK.

To fix both we would need `if: env.IS_FIRST_PUSH != 'true'`, which is IMO less clear than the literal zero SHA check. WDYT?

.github/workflows/ci.yaml
```diff
@@ -33,6 +33,7 @@ jobs:
       prerequisites_changed: ${{ steps.prerequisites.outputs.any_changed == 'true' && 'true' || '' }}
       ci_changed: ${{ steps.ci.outputs.any_changed == 'true' && 'true' || '' }}
       is_protected_branch: ${{ steps.is-protected-branch.outputs.is_protected_branch == 'true' && 'true' || '' }}
+      is_new_branch: ${{ github.event.before == '0000000000000000000000000000000000000000' && 'true' || '' }}
```
Incredible-O: Question: When is `is_new_branch != is_protected_branch`? Would it make sense to reuse `is_protected_branch` instead?

.github/workflows/ci.yaml
Incredible-O: Have you considered using `github.event.created` instead? https://docs.github.com/en/rest/using-the-rest-api/github-event-types?apiVersion=2026-03-10#createevent


```suggestion
        if: ${{ !github.event.created }}
```

waodim: `github.event.created `is true when a PR is first opened, not just when a branch is newly pushed. So `!github.event.created` would skip all change detection on newly opened PRs, no?

The zero SHA check only triggers on push events for new branches

.github/workflows/ci.yaml
waodim: Good point, but `is_protected_branch` is a superset: it is true for every push to `main, release-*`, and `merge_group` and not just the first one (so no "new branch" without prior commits). Swapping it in would skip change detection on all subsequent pushes to those branches too, which would break CI for the common case. Or am I wrong there?

andriisoldatenko (dismissed):

.github/workflows/ci.yaml
Incredible-O: > `github.event.created `is true when a PR is first opened, not just when a branch is newly pushed. 

`created` is a field of the `push` event meaning "this push created the branch". Usually, it isn't part of the `pull_request` payload at all. I confirmed it on a test PR: https://github.com/Dynatrace/dynatrace-operator-staging/pull/8.
Related workflow run: https://github.com/Dynatrace/dynatrace-operator-staging/actions/runs/28442081634/job/84282244354?pr=8
`github.event.created` was empty in that run.

> So `!github.event.created` would skip all change detection on newly opened PRs, no?

In the same PR run, `!github.event.created` was `true` (because the field was empty 😄). As a result, change detection wouldn't be skipped. 

> The zero SHA check only triggers on push events for new branches

So does `github.event.created`: I verified this by pushing a new branch, `release-2.0`, which triggered this workflow run: https://github.com/Dynatrace/dynatrace-operator-staging/actions/runs/28437585676. When new branches are pushed, `github.event.created` evaluates to true

Also, note that comparing `github.event.before` and '000000...' can cause unintended side effects: When pull requests are created, `github.event.before` is null (it simply doesn't exist): https://docs.github.com/en/rest/using-the-rest-api/github-event-types?apiVersion=2026-03-10#pullrequestevent
When GitHub Actions compares null values against a string, it casts both sides to numbers: https://docs.github.com/en/rest/using-the-rest-api/github-event-types?apiVersion=2026-03-10#pullrequestevent
In other words, `github.event.before` and '00000...' would both become `0`. The comparison would then basically be `0 != 0`, which is... false 🤠

.github/workflows/ci.yaml
waodim: Going with `if: ${{ !github.event.created }}` instead 

for more info look at @Incredible-O 's comment

.github/workflows/ci.yaml
waodim: You're right on all counts, I was wrong about `github.event.created` behaving differently on PR events, don't know where I got that from. Applied `github.event.created`.

Thanks!

https://github.com/Dynatrace/dynatrace-operator/pull/6908/commits/841523bb947e7699937c2b66190bdcc040a4069a

andriisoldatenko (dismissed):

.github/workflows/ci.yaml
Incredible-O: > `is_protected_branch` is a superset: it is true for every push to `main`, `release-*`, and `merge_group`, and not just the first one (so no "new branch" without prior commits).

I agree 😄  that's exactly why `is_new_branch` is probably redundant It can only be true on a push that creates a branch, and this workflow's `push` only fires on `main`/`release-*`: https://github.com/Dynatrace/dynatrace-operator/blob/841523bb94/.github/workflows/ci.yaml#L4-L7
Such branches always set `is_protected_branch=true`: https://github.com/Dynatrace/dynatrace-operator/blob/841523bb94/.github/workflows/ci.yaml#L47.
So, whenever `is_new_branch` is true, `is_protected_branch` already is.

> Swapping it in would skip change detection on all subsequent pushes to those branches too, which would break CI for the common case.
As far as I can see, `is_new_branch` doesn't control whether detection steps should be run (`github.event.created` does, right?). It seems like `is_new_branch` is basically yet another OR term in several if statements (where `is_protected_branch` is already true)

Just something to consider, not a blocker IMO. Maybe I'm the one who's wrong 😂

waodim: Removed is_new_branch entirely https://github.com/Dynatrace/dynatrace-operator/pull/6908/commits/a33fd68a6d9acd9fd7ac8794be40b525b3ee4aba

Incredible-O (approved):

PR #6916, opened 2026-06-30
Fix API URL handling for 3rd gen

avorima:
## Description

We allow 3rd gen URLs, but most if not all components will keep using the 2nd gen form for a while.
Add a new package dtapiurl that handles the conversion and call it from the APIURL DynaKube helper. Validation and conversion will keep using the raw value, but everything else should use the helper to ensure the 2nd gen form is used throughout.
Keep dtapiurl outside the API to reduce the moving target when we introduce the next version.

Changes made by claude.

ICP-6861

## How can this be tested?

Deploy DynaKube with 3rd gen URL
* OTEL + metadata enrichment -> should work
* AG + OTEL + metadata enrichment -> should work
* Any other metrics ingest -> should work

### DIFF
--- a/pkg/api/latest/dynakube/activegate_props.go
+++ b/pkg/api/latest/dynakube/activegate_props.go
@@ -5,7 +5,10 @@ import (
 )
 
 func (dk *DynaKube) ActiveGate() *activegate.ActiveGate {
-	dk.Spec.ActiveGate.SetAPIURL(dk.APIURL())
+	// The stored API URL is only used to derive the tenant image registry host, which is only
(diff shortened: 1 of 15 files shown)


0sewa0:
It works but the description is a bit confusing 😅 

> Tenant registry -> should not work (should not also not be possible due to validation webhook)

what validation?
I could easily get a `image: hvs0985d.dev.apps.dynatracelabs.com/linux/activegate:1.343.5-raw` with the correct setup 😅 

> when we introduce v1beta6

v1beta7

> Validation will keep using the raw value,

and Conversion (+some is-empty checks using `len`)

0sewa0 (approved):

PR #6917, opened 2026-06-30
Run e2e tests with phase3 tenant

mihaitanasedt:
## Description
This PR extends the E2E testing pipelines with an additional phase 3 tenant + platform token scenario. For the nighlty workflow there is a new job for it. And for the ondemand, the existing platfom token scenarios were splitted between the default phase2 tenant and the new phase3 tenant. Phase 3 +  classic api token will not be separately tested, the existing migration test from platform to classic api token confirms that the classic api token is compatible.

[Jira ICP-6670](https://dt-rnd.atlassian.net/browse/ICP-6670)

## How can this be tested?
Run phase 3 E2E tests.
`make test/e2e/<testcase>/phase3`


.github/actions/run-e2e/action.yaml
```diff
@@ -53,12 +53,39 @@ inputs:
   tenant2-platform-token:
     description: The platform token of Tenant 2
     required: true
+  tenant-phase3-name:
```
andriisoldatenko:
```suggestion
  tenant3-phase3-name:
```

.github/actions/run-e2e/action.yaml
```diff
@@ -140,8 +167,17 @@ runs:
         TENANT2_APITOKEN: ${{ inputs.tenant2-apitoken }}
         TENANT2_DATAINGESTTOKEN: ${{ inputs.tenant2-dataingesttoken }}
         TENANT2_PLATFORM_TOKEN: ${{ inputs.tenant2-platform-token }}
+        TENANT_PHASE3_NAME: ${{ inputs.tenant-phase3-name }}
```
andriisoldatenko:
```suggestion
        TENANT3_PHASE3_NAME: ${{ inputs.tenant-phase3-name }}
```

.github/scripts/prepare-e2e-secrets.sh
```diff
@@ -17,6 +17,16 @@ platformToken: $TENANT1_PLATFORM_TOKEN
 platformTokenNoSettings: $TENANT1_PLATFORM_TOKEN_NOSETTINGS
 EOF
 
+
+  cat << EOF > phase3-tenant.yaml
+tenantUid: $TENANT_PHASE3_NAME
```
andriisoldatenko:
```suggestion
tenantUid: $TENANT3_PHASE3_NAME
```

.github/actions/run-e2e/action.yaml
mihaitanasedt: I don't want to mix it with tenant1,tenant2 that are phase 2 tenants. With tenant3-phase3 it seems like the also have tenant1,2-phase3, which we don't

andriisoldatenko: for me it's just consistency, we use 3 tenants: tenant1, tenant2 and tenant3 :) 
/shrug

mihaitanasedt: [873db92](https://github.com/Dynatrace/dynatrace-operator/pull/6917/commits/873db92249b66c982781031bda5c92f96d3d4377)

.github/actions/run-e2e/action.yaml
mihaitanasedt: [873db92](https://github.com/Dynatrace/dynatrace-operator/pull/6917/commits/873db92249b66c982781031bda5c92f96d3d4377)

.github/scripts/prepare-e2e-secrets.sh
mihaitanasedt: [873db92](https://github.com/Dynatrace/dynatrace-operator/pull/6917/commits/873db92249b66c982781031bda5c92f96d3d4377)

andriisoldatenko (dismissed):

test/e2e/helpers/tenant/secrets.go
```diff
@@ -24,6 +24,8 @@ var (
 	defaultSingleTenant      = filepath.Join(project.TestDataDir(), "secrets/single-tenant.yaml")
 	defaultMultiTenant       = filepath.Join(project.TestDataDir(), "secrets/multi-tenant.yaml")
 	defaultEdgeConnectTenant = filepath.Join(project.TestDataDir(), "secrets/edgeconnect-tenant.yaml")
+	phase3Tenant             = filepath.Join(project.TestDataDir(), "secrets/phase3-tenant.yaml")
+	edgeConnectPhase3Tenant  = filepath.Join(project.TestDataDir(), "secrets/edgeconnect-phase3-tenant.yaml")
```
avorima, line 29: don't need separate config for edgeconnect it's already using new APIs

avorima (commented):
Duplicating the secrets doesn't seem necessary. Both will use the same sets of credentials

test/e2e/helpers/tenant/secrets.go
mihaitanasedt: We can't reuse the credentials from other phase2 tenant. We either run them with new credentials from phase3 tenant or disable them if it brings no value.  

Or am I missing something?

avorima: both use classic and platform token and they never run at the same time, so there is only ever one single-tenant secret in use. the only distinction is what the pipeline writes to the files but that doesn't need to be reflected in the actual tests

mihaitanasedt: I did it for convenience to run locally, no need to change files to run all tenants. Wdyt?

avorima: sorry, i mixed a few things up. we can have separate files for separate tenants, but i think we should consolidate the format at some point

test/e2e/helpers/tenant/secrets.go
```diff
@@ -115,6 +123,10 @@ func GetSingleTenantSecret(t *testing.T) Secret {
 }
 
 func GetMultiTenantSecret(t *testing.T) []Secret {
+	if UsePhase3Tenant() {
+		t.Skip("multi-tenant secrets are not supported with phase3 tenant")
```
avorima: these tests should work on phase 3

mihaitanasedt: Yes, but then we have to add and maintain a second phase3 tenant. Those multi tenants test do not cover a specific tenant functionality, more from operator side.

Do you think we should add an additional phase3 tenant for these?

avorima: yes, we'll need the same setup

mihaitanasedt: [22ec408](https://github.com/Dynatrace/dynatrace-operator/pull/6917/commits/22ec408131edc58e5bcb08ed5afed45660dedf1a)

I can't trigger a test run yet, I need someone with gh secrets permissions, to add the new ones.

avorima (approved):

PR #6919, opened 2026-07-01
Fix update csv bundles workflow for RC tags

waodim:
## Description

[ICP-6931](https://dt-rnd.atlassian.net/browse/ICP-6931)

When a pre-release tag (e.g. `v1.10.0-rc.1`) is published, the `Update CSV Bundles` workflow fails immediately with:

```
A branch or tag with the name 'release-1.10.0-rc' could not be found
```

In `.github/actions/update-csv-bundles/action.yml`, `MAJOR_MINOR` is derived using `${VERSION%.*}`, which strips from the **last dot** to the end of the string. For a normal release like `1.10.0` this correctly yields `1.10`. For an RC version like `1.10.0-rc.1`, the last dot-segment is `.1`, so the result is `1.10.0-rc` instead of `1.10`

## How can this be tested?

Trigger `Update CSV Bundles` via `workflow_dispatch` with an RC tag (e.g. `v1.10.0-rc.1`) and verify the workflow reaches the bundle generation step without failing on the `Checkout release branch` step.

[ICP-6931]: https://dt-rnd.atlassian.net/browse/ICP-6931?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/update-csv-bundles/action.yml
+++ b/.github/actions/update-csv-bundles/action.yml
@@ -40,7 +40,8 @@ runs:
       env:
         VERSION: ${{ inputs.version }}
       run: |
-        MAJOR_MINOR="${VERSION%.*}"
+        BASE_VERSION="${VERSION%%-*}"
+        MAJOR_MINOR="${BASE_VERSION%.*}"
         echo "major_minor=${MAJOR_MINOR}" >> "$GITHUB_OUTPUT"
     - name: Checkout release branch
       uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
--- a/.github/workflows/update-csv-bundles.yaml
+++ b/.github/workflows/update-csv-bundles.yaml
@@ -49,7 +49,6 @@ jobs:
           echo "dry-run=${DRY_RUN}" >> "$GITHUB_OUTPUT"
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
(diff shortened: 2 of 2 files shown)


andriisoldatenko (dismissed):

Incredible-O (dismissed):

avorima (dismissed):

waodim:
The `action.yml` fix alone isn't enough. The workflow explicitly checks out the release tag into the workspace before calling `uses: ./.github/actions/update-csv-bundles`. How I am currently understanding it: GitHub loads local actions from `$GITHUB_WORKSPACE` at step execution time, so the runner reads `action.yml` from the tag and not from the branch the workflow is running on. The fix on the branch is never seen.

AFAIK removing `ref: <tag>` from the initial checkout keeps the workflow branchs content in the workspace, so the fixed `action.yml` is actually picked up.

This also needs to be cherry-picked onto `release-1.10` because for `release: [published]` events the workspace is populated from the tags commit, which lives on `release-1.10`. Future RC tags need to point to a commit that already includes this change.

waodim:
Seems to work now. Will need to cherry pick to release-1.10 as well to fetch the correct commit when creating a new RC. https://github.com/Dynatrace/dynatrace-operator/actions/runs/28574599494

Incredible-O (approved):

andriisoldatenko (approved):

PR #6920, opened 2026-07-02
IsPublicRegistry feature flag supported by CSI driver provisioner installer selection function

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/ICP-6944)

## Description

The CSI driver provisioner's getInstaller function uses a public registry image when IsPublicRegistry is enabled (via feature flag or platform token).

The node-image-pull FF only allowed when codeModulesImage is set or public registry is enabled.

The case
```
	case dk.OneAgent().GetCodeModulesImage() != "":
```
covers `dk.OneAgent().GetCustomCodeModulesImage() != ""` and ` dk.FF().IsPublicRegistry` conditions. In both cases status.ImageID is set.

## How can this be tested?

unittests

standard e2e tests
```
go test -v -count 1 -tags e2e -timeout 200m ./test/e2e/scenarios/standard -run "use_public_registry_codemodules_with_csi" 
 ```

### DIFF
--- a/pkg/api/validation/dynakube/oneagent.go
+++ b/pkg/api/validation/dynakube/oneagent.go
@@ -48,7 +48,7 @@ Use a nodeSelector to avoid this conflict. Conflicting DynaKubes: %s`
 
 	warningDeprecatedVersionIgnored = `version field is deprecated and ignored. Please remove the version field from the DynaKube specification.`
 
-	errorImagePullRequiresCodeModulesImage = `The DynaKube specification enables node image pull, but the code modules image is not set.`
+	errorImagePullRequiresCodeModulesImage = `The DynaKube specification enables node image pull, but neither a code modules image is set nor a public registry is used.`
 )
 
 func conflictingOneAgentConfiguration(ctx context.Context, _ *Validator, dk *dynakube.DynaKube) string {
--- a/pkg/api/validation/dynakube/oneagent.go
+++ b/pkg/api/validation/dynakube/oneagent.go
@@ -47,6 +47,8 @@ Use a nodeSelector to avoid this conflict. Conflicting DynaKubes: %s`
 	warningDeprecatedVersion = `version field is deprecated. Please use "%s" field instead to set a version.`
 
 	warningDeprecatedVersionIgnored = `version field is deprecated and ignored. Please remove the version field from the DynaKube specification.`
+
+	errorImagePullRequiresCodeModulesImage = `The DynaKube specification enables node image pull, but the code modules image is not set.`
 )
 
 func conflictingOneAgentConfiguration(ctx context.Context, _ *Validator, dk *dynakube.DynaKube) string {
@@ -330,4 +332,14 @@ func findDuplicates[S ~[]E, E comparable](s S) []E {
 	}
 
 	return duplicates
+}
+
+func missingCodeModulesImage(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
+	if dk.OneAgent().IsAppInjectionNeeded() && dk.FF().IsNodeImagePull() {
+		if dk.OneAgent().GetCustomCodeModulesImage() == "" && !dk.FF().IsPublicRegistry() {
+			return errorImagePullRequiresCodeModulesImage
+		}
+	}
+
+	return ""
 }
--- a/pkg/api/validation/dynakube/oneagent_test.go
+++ b/pkg/api/validation/dynakube/oneagent_test.go
@@ -1175,3 +1175,101 @@ func TestInvalidOneAgentArguments(t *testing.T) {
 		dk.Spec.OneAgent.CloudNativeFullStack.Args = []string{value}
 	}, errorInvalidOneAgentArgument)
 }
+
+func TestMissingCodeModulesImage(t *testing.T) {
+	baseDK := &dynakube.DynaKube{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "dynakube",
+			Namespace: testNamespace,
+		},
+		Spec: dynakube.DynaKubeSpec{
+			APIURL: testAPIURL,
+		},
+	}
+
+	testcases := []struct {
+		name        string
+		oaSpec      oneagent.Spec
+		annotations map[string]string
+		isValid     bool
+	}{
+		{
+			"node-image-pull + application monitoring",
+			oneagent.Spec{ApplicationMonitoring: &oneagent.ApplicationMonitoringSpec{}},
+			map[string]string{
+				exp.OANodeImagePullKey: "true",
+			},
+			false,
+		},
+		{
+			"node-image-pull + application monitoring + codemodules image",
+			oneagent.Spec{ApplicationMonitoring: &oneagent.ApplicationMonitoringSpec{
+				AppInjectionSpec: oneagent.AppInjectionSpec{
+					CodeModulesImage: "test-image",
+				},
+			}},
+			map[string]string{
+				exp.OANodeImagePullKey: "true",
+			},
+			true,
+		},
+		{
+			"node-image-pull + application monitoring + public registry",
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: pkg/controllers/csi/provisioner/install.go [hunk 1]
```diff
@@ -70,6 +70,26 @@ func (provisioner *OneAgentProvisioner) getInstaller(ctx context.Context, dk *dy
 			return nil, err
 		}
 
+		return imageInstaller, nil
+	case dk.FF().IsPublicRegistry():
+		imageURI := dk.OneAgent().GetCodeModulesImage()
+
+		if imageURI == "" {
+			imageURI = "public.ecr.aws/dynatrace/dynatrace-codemodules:" + dk.OneAgent().GetCodeModulesVersion()
```
andriisoldatenko: this won't work, please don't hardcode registry

andriisoldatenko: in this case you should use new endpoint, cc @jokuniew

jokuniew: you mean @andriisoldatenko to use fleetManagement/components/containerImages codemodules imageURI, right?

andriisoldatenko: yes i think the solution is not like we did in another places.

test/e2e/features/usepublicregistry/deploy.go
```diff
@@ -426,6 +518,29 @@ func statusSourceIs(dk dynakube.DynaKube, component image.ComponentType, expecte
 	}
 }
 
+func getStatusImageID(dk dynakube.DynaKube, component image.ComponentType, imageID *string) features.Func {
```
andriisoldatenko: i understand intention of this helper, but we don't reuse it AFAIK, why not simplify it?

pkg/controllers/csi/provisioner/controller_test.go
```diff
@@ -170,6 +172,26 @@ func TestReconcile(t *testing.T) {
 		assert.True(t, areFsDirsCreated(t, prov, dk))
 	})
 
+	t.Run("dynakube with public registry => image installer used, no error", func(t *testing.T) {
+		imageURI := "public.ecr.aws/dynatrace/codemodules:1.0.0"
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
avorima: this test should use the client mock

test/e2e/features/usepublicregistry/deploy.go
```diff
@@ -230,6 +242,86 @@ func codeModulesFeature(t *testing.T, featureName, dkName, sampleNamespaceName,
...
+			Namespace: namespace,
+		}).ForEachPod(checkProvisionerLog(ctx, t, envConfig, imageID))
+
+		assert.NoError(t, err)
+
+		return ctx
+	}
+}
+
+func checkProvisionerLog(ctx context.Context, t *testing.T, envConfig *envconf.Config, imageID *string) k8sdaemonset.PodConsumer {
```
avorima: imageID is not mutated. use a value instead of a pointer to make this obvious

test/e2e/features/usepublicregistry/deploy.go
```diff
@@ -230,6 +242,86 @@ func codeModulesFeature(t *testing.T, featureName, dkName, sampleNamespaceName,
...
+	return func(pod corev1.Pod) {
+		provisionerLog := logs.ReadLog(ctx, t, envConfig, pod.Namespace, pod.Name, "provisioner")
+
+		// expected message: `{"level":"info",...,"msg":"pullOciImage",...,"ref.String":"<registry>:<version>@sha256:<sha256>"}`
```
avorima, line 312: we need to find a better way to do this. grepping logs is very fragiile

pkg/controllers/csi/provisioner/install.go
```diff
@@ -70,6 +71,34 @@ func (provisioner *OneAgentProvisioner) getInstaller(ctx context.Context, dk *dy
...
 		}
 
+		return imageInstaller, nil
+	case dk.FF().IsPublicRegistry():
+		dtClient, err := buildDtc(provisioner, ctx, dk)
+		if err != nil {
+			return nil, err
+		}
+
+		imageInfo, err := dtClient.Images.GetComponentLatestInfo(ctx, imageclient.CodeModules, dk.PublicRegistryOverride())
```
avorima: the codemodules are already resolved by the version updater, so the provisioner can just use the status

test/e2e/features/usepublicregistry/deploy.go
aorcholski: I agree and I thought about files/directories but they do not confirm the true source of the image. I don't know better solution

pkg/controllers/csi/provisioner/install.go
aorcholski: it was my first approach in the first commit but dtClient is suggested as better solution considering changes in status fields in the near future

pkg/api/validation/dynakube/oneagent_test.go
```diff
@@ -1175,3 +1175,101 @@ func TestInvalidOneAgentArguments(t *testing.T) {
...
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			dk := baseDK.DeepCopy()
+			dk.Annotations = tc.annotations
+			dk.Spec.OneAgent = tc.oaSpec
+
+			if tc.isValid {
+				warnings, err := assertAllowed(t, dk)
```
avorima: use assertAllowedWithoutWarnings

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -48,6 +48,8 @@ Use a nodeSelector to avoid this conflict. Conflicting DynaKubes: %s`
 	warningDeprecatedVersion = `version field is deprecated. Please use "%s" field instead to set a version.`
 
 	warningDeprecatedVersionIgnored = `version field is deprecated and ignored. Please remove the version field from the DynaKube specification.`
+
+	errorImagePullRequiresCodeModulesImage = `The DynaKube specification enables node image pull, but the code modules image is not set.`
```
avorima: error message should mention public registry feature as well

test/e2e/features/usepublicregistry/deploy.go
```diff
@@ -230,6 +242,86 @@ func codeModulesFeature(t *testing.T, featureName, dkName, sampleNamespaceName,
...
+	builder.Assess("get CodeModules imageID", getStatusCodeModulesImageID(testDynakube, &imageID))
+
+	builder.Assess("verify if CSI provisioner uses public-registry image", isProvisionerUsingPublicRegistry(testDynakube.Namespace, &imageID))
+
+	builder.Teardown(sampleApp.Uninstall())
+
+	return builder.Feature()
+}
+
+func isProvisionerUsingPublicRegistry(namespace string, imageID *string) features.Func {
```
avorima, line 293: this also doesn't need to be a pointer

avorima (commented):
it would be good if the test could be invoked with make `test/e2e/usepublicregistry/codemodules`

test/e2e/features/usepublicregistry/deploy.go
aorcholski: It has to be a pointer. Otherwise `expected` is empty. We need a value set when the test is executed.
```
Error:      Not equal: 
      	        expected: ""
      	        actual  : "<registry>/dynatrace/dynatrace-codemodules:1.344.0....."
```

avorima: <strike>only getStatusCodeModulesImageID needsa pointer since it modifies the value</strike>

nvm you're right the closure captures over the empty string

avorima (dismissed):

avorima (approved):

PR #6926, opened 2026-07-03
Do not create/attach tenant pull-secret when public-registry FF is used

0sewa0:
## Description

[ICP-7075](https://dt-rnd.atlassian.net/browse/ICP-7075)

When you use the `use-public-registry` FF then the tenant pull-secret is useless, and will cause unnecessary restarts when moving from using the FF to using platform token.

> [!NOTE]
> Possible edgecase: You can technically set the `image` field of the OA/AG to point at the tenant registry, in which case you need this pull-secret. The only (cursed) scenario I can think of is classicfullstack + something else that can use the public-registries

## How can this be tested?

1. Use `feature.dynatrace.com/use-public-registry: true` deploy oa / ag
2. Check that the pull-secret is not created
3. Check that the pull-secret is not attached to the pods

[ICP-7075]: https://dt-rnd.atlassian.net/browse/ICP-7075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/api/latest/dynakube/dynakube_props.go
+++ b/pkg/api/latest/dynakube/dynakube_props.go
@@ -132,7 +132,7 @@ func (dk *DynaKube) CustomPullSecretReferences() []corev1.LocalObjectReference {
 func (dk *DynaKube) ImagePullSecretReferences() []corev1.LocalObjectReference {
 	var refs []corev1.LocalObjectReference
 
-	if !ptr.Deref(dk.Status.APIToken.Platform, false) {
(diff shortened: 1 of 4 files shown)


andriisoldatenko (approved):

andriisoldatenko:
@0sewa0 nice catch!

andriisoldatenko:
@0sewa0 may be we can tweak existing e2e test to check secret is not there?

avorima (approved):

PR #6928, opened 2026-07-06
Always read token secret for validation

avorima:
## Description

Read the token secret before running validators and set the value of `status.apiToken.platform` accordingly for the individual admission request.
Do this once per entrypoint, so that the validators themselves don't do duplicate work.

This is needed to prevent users switching from a platform token to a classic token from keeping features that are exclusive to platform tokens.

ICP-7129

## How can this be tested?

Deploy DynaKubes with platform token and switch back to classic token while enabling features that require a platform token


mihaitanasedt (approved):
looks much better, LGTM!

jokuniew (approved):
💯

PR #6935, opened 2026-07-08
Remove ineffective ImageDigest field

avorima:
## Description

The image installer properties have an ImageDigest field that's no longer used.
It was introduced in #1954. At that time the field was populated by extracting the digest from the image URI. This functionality got removed in #2654, but the field remained.
The result is that the image cache directory has been the top-level cache directory ever since then (`filepath.Join` ignores empty elements).

## How can this be tested?

`make test/e2e/cloudnative/codemodules`


aorcholski (approved):
LGTM tests PASS

PR #6942, opened 2026-07-09
feat: add omitzero to struct fields

VLazarevic:
## Description

https://dt-rnd.atlassian.net/browse/ICP-973

Several struct-typed fields in the DynaKube CRD were annotated with `omitempty`.
That tag has no effect on non-pointer struct fields, so those fields were always
serialized, even when they held nothing but zero values. This PR switches them to
`omitzero` (Go 1.24+), which drops a field when its value is the zero value.

Changes:
- Added `omitzero` to the struct-typed fields in `pkg/api/latest/dynakube`
  (spec and status), plus the affected fields in the sub-packages
  (`activegate`, `kspm`, `extensions`).
- Added a unit test that marshals a DynaKube and asserts the empty struct fields
  are dropped, and that populated ones are still rendered.
- Updated `doc/coding-style-guide.md` with a short section on when to use
  `omitempty` vs `omitzero`.
- Regenerated the CRDs.

Note on the API version: the ticket mentions `v1beta5`, but since it was written
`pkg/api/latest` (served as `v1beta6`) has become the storage version, so the
change was made there instead. `v1beta5` is untouched.

### Finding: conversion re-introduces the empty fields

While verifying with ArgoCD I noticed the empty fields are still present if the
DynaKube is applied with an older `apiVersion` and goes through the conversion
webhook. The conversion materializes empty maps, slices and pointers, so the
values arriving in `v1beta6` are not zero any more and `omitzero` correctly does
not drop them. The clearest example is `nodeAffinity: {}` on a
`*corev1.NodeAffinity` field, a nil pointer would always be omitted by plain
`omitempty`, so the pointer must have been set to a non-nil empty value during
conversion.

So the change works for resources created natively as `v1beta6`, but not for
resources converted up from an older version. Fixing that would mean touching the
conversion code itself, which felt out of scope here, happy to follow up in a
separate ticket if that's wanted.

## How can this be tested?

Any Kubernetes cluster works, no Dynatrace tenant needed, the DynaKube ends up in
`phase: Error` because of the dummy token, but that doesn't affect what's being
tested here (the serialized output).

1. Deploy the operator from this branch.
2. Apply a minimal DynaKube where only `apiUrl` is set, using the current
   apiVersion:

```yaml
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
spec:
  apiUrl: https://abc12345.live.dynatrace.com/api
```

3. Read it back with `kubectl get dynakube dynakube -n dynatrace -o yaml`.

The spec should contain only `apiUrl`, and the status only the fields that carry
actual data. Before the change, a number of empty struct fields were rendered as
`{}`.

To see the conversion behaviour described above, apply the same manifest with
`apiVersion: dynatrace.com/v1beta5` instead, the empty fields come back.

<details>
<summary>Before</summary>

```yaml
spec:
  activeGate:
    resources: {}
  apiUrl: https://abc12345.live.dynatrace.com/api
  metadataEnrichment:
    namespaceSelector: {}
  oneAgent: {}
  templates:
    extensionExecutionController:
      imageRef: {}
      resources: {}
      imageRef: {}
      nodeAffinity: {}
      resources: {}
    otelCollector:
      imageRef:
        repository: public.ecr.aws/dynatrace/dynatrace-otel-collector
        tag: latest
      resources: {}
    sqlExtensionExecutor:
      imageRef: {}
status:
  activeGate:
    connectionInfoStatus: {}
  codeModules: {}
  conditions:
  - lastTransitionTime: "2026-07-16T11:26:11Z"
    message: Token verification failed
    reason: TokenError
    status: "False"
    type: Tokens
  dynatraceApi: {}
  kspm: {}
  metadataEnrichment: {}
  oneAgent:
    connectionInfoStatus: {}
  phase: Error
```

</details>

<details>
<summary>After</summary>

```yaml
spec:
  apiUrl: https://abc12345.live.dynatrace.com/api
status:
  apiToken:
    platform: false
  conditions:
  - lastTransitionTime: "2026-07-20T08:11:41Z"
    message: Token verification failed
    reason: TokenError
    status: "False"
    type: Tokens
  phase: Error
  updatedTimestamp: "2026-07-20T08:11:41Z"
```

</details>


andriisoldatenko (changes requested):
please fix unit tests

pkg/api/latest/dynakube/dynakube_omitzero_test.go
```diff
@@ -0,0 +1,153 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TestOmitzero verifies that the struct-typed fields tagged with `omitzero`
+// (see dynakube_types.go and dynakube_status.go) are dropped from the marshaled
+// JSON when they hold a zero value, and are still rendered when they carry a
+// value. `omitempty` has no effect on non-pointer struct fields, which is why
+// these fields need `omitzero` (Go 1.24+).
+func TestOmitzero(t *testing.T) {
```
avorima: i don't think we should unit test omitzero. instead, something like the conversion integration tests with YAML golden files would be a better approach

svfoxat:
I think you may have missed some fields there:
(didn't see it was put back into draft when i started the comment)

| File | Field |
|---|---|
| [`oneagent/spec.go:121`](https://github.com/Dynatrace/dynatrace-operator/blob/e5951e85cc0420cb4d66e0023903c2b518d6be61/pkg/api/latest/dynakube/oneagent/spec.go#L121) | `OneAgentResources` |
| [`oneagent/spec.go:188`](https://github.com/Dynatrace/dynatrace-operator/blob/e5951e85cc0420cb4d66e0023903c2b518d6be61/pkg/api/latest/dynakube/oneagent/spec.go#L188) | `NamespaceSelector` |
| [`kubemon/spec.go:41`](https://github.com/Dynatrace/dynatrace-operator/blob/e5951e85cc0420cb4d66e0023903c2b518d6be61/pkg/api/latest/dynakube/kubemon/spec.go#L41) | `Resources` |
| [`logmonitoring/spec.go:44`](https://github.com/Dynatrace/dynatrace-operator/blob/e5951e85cc0420cb4d66e0023903c2b518d6be61/pkg/api/latest/dynakube/logmonitoring/spec.go#L44) | `ImageRef` |
| [`logmonitoring/spec.go:60`](https://github.com/Dynatrace/dynatrace-operator/blob/e5951e85cc0420cb4d66e0023903c2b518d6be61/pkg/api/latest/dynakube/logmonitoring/spec.go#L60) | `Resources` |
| [`metadataenrichment/spec.go:38`](https://github.com/Dynatrace/dynatrace-operator/blob/e5951e85cc0420cb4d66e0023903c2b518d6be61/pkg/api/latest/dynakube/metadataenrichment/spec.go#L38) | `NamespaceSelector` |
| [`otlp/spec.go:31`](https://github.com/Dynatrace/dynatrace-operator/blob/e5951e85cc0420cb4d66e0023903c2b518d6be61/pkg/api/latest/dynakube/otlp/spec.go#L31) | `Signals` |
| [`otlp/spec.go:40`](https://github.com/Dynatrace/dynatrace-operator/blob/e5951e85cc0420cb4d66e0023903c2b518d6be61/pkg/api/latest/dynakube/otlp/spec.go#L40) | `NamespaceSelector` |

pkg/api/latest/dynakube/activegate/spec.go
```diff
@@ -161,7 +161,7 @@ type CapabilityProperties struct {
 	// Define resources requests and limits for single ActiveGate pods
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Resource Requirements",order=34,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced","urn:alm:descriptor:com.tectonic.ui:resourceRequirements"}
-	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+	Resources corev1.ResourceRequirements `json:"resources,omitempty,omitzero"`
```
mihaitanasedt: Value typed structs are not affected by omitempty. They need just omitezero, we should remove it from there. Also in other places

pkg/api/latest/dynakube/dynakube_omitzero_test.go
VLazarevic: You’re right, thanks. I am going to remove the unit test. 


The existing conversion/webhook integration tests with the YAML golden files already cover this, and they’re what caught the changed output in the first place (where the tests failed yesterday), so I’ve updated those fixtures instead. Let me know if you’d still like a dedicated case added on top, otherwise I think we’re covered :)

avorima: The current golden files are designed to test conversion so all fields are set. We want empty fields for omitzero. They also use the unstructured client, so the omitzero is not actually used. Create a separate set of golden files and use the go structs to get proper coverage

pkg/api/latest/dynakube/dynakube_types.go
```diff
@@ -58,9 +58,9 @@ const (
 type DynaKube struct {
 	metav1.TypeMeta `json:",inline"`
 
-	Status            DynaKubeStatus `json:"status,omitempty"`
+	Status            DynaKubeStatus `json:"status,omitempty,omitzero"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
```
mihaitanasedt: I don't think omitempty works here

pkg/api/latest/dynakube/dynakube_types.go
```diff
@@ -183,13 +183,13 @@ type TemplatesSpec struct {
...
+	KSPMNodeConfigurationCollector kspm.NodeConfigurationCollectorSpec `json:"kspmNodeConfigurationCollector,omitempty,omitzero"`
 	// +kubebuilder:validation:Optional
-	OpenTelemetryCollector OpenTelemetryCollectorSpec `json:"otelCollector,omitempty"`
+	OpenTelemetryCollector OpenTelemetryCollectorSpec `json:"otelCollector,omitempty,omitzero"`
 	// +kubebuilder:validation:Optional
-	SQLExtensionExecutor extensions.DatabaseExecutorSpec `json:"sqlExtensionExecutor,omitempty"`
+	SQLExtensionExecutor extensions.DatabaseExecutorSpec `json:"sqlExtensionExecutor,omitempty,omitzero"`
 	// +kubebuilder:validation:Optional
-	ExtensionExecutionController extensions.ExecutionControllerSpec `json:"extensionExecutionController,omitempty"`
+	ExtensionExecutionController extensions.ExecutionControllerSpec `json:"extensionExecutionController,omitempty,omitzero"`
```
mihaitanasedt: The fields of the `extensions.ExecutionControllerSpec` were not assessed

pkg/api/latest/dynakube/dynakube_omitzero_test.go
mihaitanasedt, line 1: I don't think we need unit tests for this

VLazarevic: I know, talked with Mario to do golden file tests

mihaitanasedt: I see now,  I agree with his approach

pkg/api/validation/dynakube/testdata/latest-default.yaml
```diff
@@ -266,17 +265,7 @@ spec:
...
   tokens: token
   trustedCAs: trusted-ca
 status:
-  activeGate:
-    connectionInfoStatus: {}
-  apiToken:
-    availableOptionalScopes: {}
-  codeModules: {}
-  kspm: {}
   metadataEnrichment: {}
```
mihaitanasedt: This should also not be here. It is because of the conversion hook.
```go
dst.Status.MetadataEnrichment.Rules = make([]EnrichmentRule, 0)
	for _, rule := range src.Status.MetadataEnrichment.Rules {
		dst.Status.MetadataEnrichment.Rules = append(dst.Status.MetadataEnrichment.Rules,
			EnrichmentRule{
				Type:   EnrichmentRuleType(rule.Type),
				Source: rule.Source,
				Target: rule.Target,
			})
	}
``` 
Should we skip this step if the `len(src.Status.MetadataEnrichment.Rules) == 0`? @andriisoldatenko

VLazarevic: Good catch 🥇

pkg/api/validation/dynakube/serialization_integration_test.go
```diff
@@ -0,0 +1,203 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+// updateSerializationGolden regenerates the serialization golden files instead
+// of asserting against them. Run:
+//
+//	go test ./pkg/api/validation/dynakube/ -run TestSerialization -update
+var updateSerializationGolden = flag.Bool("update", false, "update the serialization golden files in testdata/")
+
+var latestGVK = schema.GroupVersionKind{Group: "dynatrace.com", Version: "v1beta6", Kind: "DynaKube"}
```
avorima: this becomes outdated. use `latest.GroupVersion.WithKind("DynaKube")`

VLazarevic: [e082355](https://github.com/Dynatrace/dynatrace-operator/pull/6942/commits/e082355288a175e515992cc6514244736ca3d3af)

pkg/api/latest/dynakube/dynakube_types.go
VLazarevic: [6e02c6a](https://github.com/Dynatrace/dynatrace-operator/pull/6942/commits/6e02c6a6f15e943667ff7665f7c7dda951a7bb53)

pkg/api/latest/dynakube/activegate/spec.go
VLazarevic: [6e02c6a](https://github.com/Dynatrace/dynatrace-operator/pull/6942/commits/6e02c6a6f15e943667ff7665f7c7dda951a7bb53)

pkg/api/latest/dynakube/dynakube_types.go
VLazarevic: [0b85b3d](https://github.com/Dynatrace/dynatrace-operator/pull/6942/commits/0b85b3d6a53d9f7bc49ec49a87310b77eb535067)

pkg/api/latest/dynakube/dynakube_types.go
```diff
@@ -57,10 +57,11 @@ const (
...
 	metav1.TypeMeta `json:",inline"`
-
-	Status            DynaKubeStatus `json:"status,omitempty"`
+	// +kubebuilder:validation:Optional
+	Status            DynaKubeStatus `json:"status,omitzero"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Spec              DynaKubeSpec `json:"spec,omitempty"`
+	// +kubebuilder:validation:Optional
+	Spec DynaKubeSpec `json:"spec,omitzero"`
 }
```
avorima, line 75: Let's align this with kubebuilder. The order was changed previously because we used the fieldalignment linter. Here's what I get from kubebuilder version v4.15.0

```go
type Frigate struct {
	metav1.TypeMeta `json:",inline"`

	// metadata is a standard object metadata
	// +optional
	metav1.ObjectMeta `json:"metadata,omitzero"`

	// spec defines the desired state of Frigate
	// +required
	Spec FrigateSpec `json:"spec"`

	// status defines the observed state of Frigate
	// +optional
	Status FrigateStatus `json:"status,omitzero"`
}
```

VLazarevic: [1573637](https://github.com/Dynatrace/dynatrace-operator/pull/6942/commits/1573637ecd9937e1c4cc0556f1f569d7feb77188)

pkg/api/latest/dynakube/dynakube_types.go
VLazarevic: [1573637](https://github.com/Dynatrace/dynatrace-operator/pull/6942/commits/1573637ecd9937e1c4cc0556f1f569d7feb77188)

avorima (dismissed):

andriisoldatenko:
@VLazarevic so we don't forget: please remove disabling of linter https://github.com/Dynatrace/dynatrace-operator/blob/main/.golangci.yml#L163-L165

VLazarevic:
> @VLazarevic so we don't forget: please remove disabling of linter https://github.com/Dynatrace/dynatrace-operator/blob/main/.golangci.yml#L163-L165

[9921e48](https://github.com/Dynatrace/dynatrace-operator/pull/6942/commits/9921e48ccc75d8f7cefffc7fbb3568aafc4f9776)

pkg/api/validation/dynakube/testdata/latest-default.yaml
andriisoldatenko: @mihaitanasedt i not quite sure what is you proposal:

if `len(src.Status.MetadataEnrichment.Rules) == 0`, loop will be skipped in go. we dont need to other checks.

andriisoldatenko: fixed https://github.com/Dynatrace/dynatrace-operator/pull/6942/changes/8eb3ddc57505999e5c28a06f2f603d9f5ef8adad

pkg/webhook/mutation/pod/attributes/container.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 type Container struct {
-	ContainerName string `json:"k8s.container.name,omitempty"`
+	ContainerName string `json:"k8s.container.name,omitzero"`
```
avorima: omitempty should work fine for string

pkg/api/v1beta4/dynakube/convert_from.go
```diff
@@ -218,14 +218,17 @@ func (dst *DynaKube) fromStatus(src *dynakubelatest.DynaKube) {
...
-	dst.Status.MetadataEnrichment.Rules = make([]EnrichmentRule, 0)
-	for _, rule := range src.Status.MetadataEnrichment.Rules {
-		dst.Status.MetadataEnrichment.Rules = append(dst.Status.MetadataEnrichment.Rules,
-			EnrichmentRule{
-				Type:   EnrichmentRuleType(rule.Type),
-				Source: rule.Source,
-				Target: rule.Target,
-			})
+	if len(src.Status.MetadataEnrichment.Rules) > 0 {
+		dst.Status.MetadataEnrichment.Rules = make([]EnrichmentRule, 0)
```
avorima: if you're not pre-allocating then this is just a more verbose `[]EnrichmentRule{}`

andriisoldatenko: nah, it's `src` and `dst` or what do you mean?
best you can see how it works check integration test, if Old.status.MetadataEnrichment.Rules are not in the status we add them.

pkg/api/validation/dynakube/testdata/latest-default.yaml
andriisoldatenko: cc @avorima based on this comment.

pkg/api/v1beta4/dynakube/convert_from.go
andriisoldatenko: https://github.com/Dynatrace/dynatrace-operator/pull/6942#discussion_r3638146121

pkg/webhook/mutation/pod/attributes/container.go
andriisoldatenko: true, tested, @VLazarevic please revert this change.

pkg/api/validation/dynakube/testdata/latest-default.yaml
```diff
@@ -266,17 +265,6 @@ spec:
...
   tokens: token
   trustedCAs: trusted-ca
-status:
-  activeGate:
-    connectionInfoStatus: {}
-  apiToken:
-    availableOptionalScopes: {}
-  codeModules: {}
-  kspm: {}
-  metadataEnrichment: {}
```
andriisoldatenko, line 280: @avorima if we pre-allocate rules, golang will add `metadataEnrichment: {}` during serialization, if we keep it nil, we can ommit it. That's the idea.

pkg/webhook/mutation/pod/attributes/container.go
VLazarevic: [f9fcf40](https://github.com/Dynatrace/dynatrace-operator/pull/6942/commits/f9fcf40f389dd8779cfa6c4c1c7a010d566abc64)

my mistake, thx and good catch 🥇

pkg/api/validation/dynakube/testdata/latest-default.yaml
avorima: the logic was changed to check if there are rules before copying, so preallocating is fine

pkg/api/v1beta4/dynakube/convert_from.go
avorima: `make([]EnrichmentRule, 0)` is the same as `[]EnrichmentRule{}` but more verbose. i'm saying that if we don't use the prellocation feature from make then we should use the shorter form

andriisoldatenko: done in 7feaff4e4

andriisoldatenko (approved):

avorima (approved):

PR #6947, opened 2026-07-13
Add resources for OA DS init container

0sewa0:
## Description

https://dt-rnd.atlassian.net/browse/ICP-7367

- Add a new `oneAgentInitResources` field in the DK, 
   - yes it is a mouthful, but it is called like this so we don't conflict with `initResources` that is for the injected init-container. (the conflict is only present incase of `cloudNativeFullstack`)
   - it only affects the resources of the little bootstrapper init-container (confusingly called `dynatrace-operator` 🙂) we add to the OA DaemonSets
   - I made it configurable just to be safe.
- Also made sure there is a default for the resources of the little bootstrapper init-container
  - should be static, so you don't need to scale it "according to env size" (if you have more `additionalAttributes` it should consume "more", but probably in measurable in bytes and not in Mi 🙂)
  - the defaults are based on my testing
     - `cpu: 20m`: Works with `10m`, but kinda slow, so a bit more goes a long way
     - `memory: 20Mi`: Also works with `10Mi` (dies at `5Mi`), but for safety, bumped it to 20.

(+) Moved all "init-related" logic to a separate files, seemed to fit the general structure of the package more.

### Open questions
- ~Should we set `limits` by default?~ -> Default `limits` were removed

## How can this be tested?

Check that the OA DaemonSet's `dynatrace-operator` initContainer has `resources` set. 
- (+) play around with `spec.oneAgent.<>.oneAgentInitResources` to see if it works


pkg/controllers/dynakube/oneagent/daemonset/init_test.go
```diff
@@ -0,0 +1,251 @@
...
+	spec := dsBuilder.initContainerSpec()
+	require.NoError(t, os.Unsetenv(k8senv.DTOperatorImageEnvName))
+
+	assert.Equal(t, testOperatorImageName, spec.Image)
+	assert.Equal(t, initContainerName, spec.Name)
+	assert.Empty(t, spec.ImagePullPolicy)
+	assert.NotEmpty(t, spec.Env)
+	assert.NotEmpty(t, spec.Args)
+	assert.NotEmpty(t, spec.VolumeMounts)
+	assert.NotNil(t, spec.SecurityContext)
```
svfoxat, line 45:
```Go
assert.Equal(t, k8sresource.NewResourceList("20m", "20Mi"), spec.Resources.Requests)
assert.Equal(t, k8sresource.NewResourceList("20m", "20Mi"), spec.Resources.Limits)
```

or even define the `20m` and `20Mi` as `const` in `init.go` and export?

/edit: I just saw the `func TestInitContainerResources(t *testing.T)` now 😆

waodim: Would it also make sense to nil assert `spec.Resources` here?

pkg/controllers/dynakube/oneagent/daemonset/init.go
```diff
@@ -0,0 +1,107 @@
...
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "spec.nodeName",
+				},
+			},
+		},
+	}
+}
+
+func (b *builder) initContainerResources() corev1.ResourceRequirements {
+	if b.hostInjectSpec.OneAgentInitResources == nil {
```
mihaitanasedt, line 46: Add nil check, it panics
```suggestion
	if b.hostInjectSpec == nil || b.hostInjectSpec.OneAgentInitResources == nil {
```

pkg/controllers/dynakube/oneagent/daemonset/init.go
```diff
@@ -0,0 +1,107 @@
...
+		return corev1.ResourceRequirements{
+			Requests: k8sresource.NewResourceList("20m", "20Mi"),
```
avorima, line 48: we only set CPU requests for the main container, but here it's also memory requests and limits which can be risky

pkg/controllers/dynakube/oneagent/daemonset/init.go
```diff
@@ -0,0 +1,107 @@
...
+		"k8s.cluster.name=" + b.dk.Status.KubernetesClusterName,
+		"k8s.cluster.uid=" + b.dk.Status.KubeSystemUUID,
+		"k8s.node.name=$(DT_K8S_NODE_NAME)",
+	}
+
+	if b.dk.Status.KubernetesClusterMEID != "" {
+		attributes = append(attributes, "dt.entity.kubernetes_cluster="+b.dk.Status.KubernetesClusterMEID)
+	}
+
+	resourceAttrs := b.dk.OneAgent().GetResourceAttributes()
```
avorima, line 66: sanitize these command-line args (pkg/util/sanitize)

pkg/controllers/dynakube/oneagent/daemonset/init.go
0sewa0: other tests fail because they are testing an impossible scenario... (`hostInjectSpec` will never ever be nil in real life)

will try to fix the actual tests (already found 1 which tests nothing :) )

pkg/controllers/dynakube/oneagent/daemonset/init.go
0sewa0: nice catch, totally unrelated, will do a "fly by"

pkg/controllers/dynakube/oneagent/daemonset/init.go
0sewa0: I have the "open question" about this for a reason.

I added them, because this is our component (the main container of the OA is not) so we have a good idea what it does. 

My actual main reason for not adding `limits` would be that IMO `limits` should be for long running memory-leaky/cpu-hungry components only (which is more of a defense mechanism), not for an init-container that should do its job in less than a second 😅 

I will remove the `limits`. We have to document this anyway, I will document possible `limits` there instead.

pkg/controllers/dynakube/oneagent/daemonset/init.go
mihaitanasedt: even better 👍

mihaitanasedt:
Wdyt about adding reconciler tests to `oneagent_reconciler_test.go`?

pkg/controllers/dynakube/oneagent/daemonset/init_test.go
0sewa0: done

pkg/controllers/dynakube/oneagent/daemonset/init.go
0sewa0: done

pkg/controllers/dynakube/oneagent/daemonset/init.go
0sewa0: done

pkg/controllers/dynakube/oneagent/daemonset/init.go
0sewa0: done

0sewa0:
> Wdyt about adding reconciler tests to oneagent_reconciler_test.go?

@mihaitanasedt you mean add stuff to `TestHasSpecChanged` or similar?

So, IMO:
1. that whole test case shouldn't be in `oneagent_reconciler_test.go`, but in the `daemonset` package.
    -  I know that technically the hash annotation is set in the `oneagent_reconciler.go`, but that makes no sense, and should be done in the `daemonset` package anyway
2. that whole test case should be cleaned up as it is filled with duplicate entries

All in all, I see no benefit, it would be adding tests for the sake of it, and not actually testing something new.

(the whole oneagent/daemonset packages could use a refactor, as they are ancient)

svfoxat (approved):

mihaitanasedt:
> > Wdyt about adding reconciler tests to oneagent_reconciler_test.go?
> 
> @mihaitanasedt you mean add stuff to `TestHasSpecChanged` or similar?
> 
> So, IMO:
> 
> 1. that whole test case shouldn't be in `oneagent_reconciler_test.go`, but in the `daemonset` package.
>    
>    * I know that technically the hash annotation is set in the `oneagent_reconciler.go`, but that makes no sense, and should be done in the `daemonset` package anyway
> 2. that whole test case should be cleaned up as it is filled with duplicate entries
> 
> All in all, I see no benefit, it would be adding tests for the sake of it, and not actually testing something new.
> 
> (the whole oneagent/daemonset packages could use a refactor, as they are ancient)

I mean to add a reconciler test where you reconcile a DK with the init resources field set, then fetch the daemon init container to check if the values match. The new unit tests, test only if the builder works, not if actually the values will be taken from DK and passed down to init container.

pkg/controllers/dynakube/oneagent/daemonset/daemonset_test.go
```diff
@@ -650,7 +606,8 @@ func TestPodSpecProbes(t *testing.T) {
 	})
 	t.Run("nil probes when dynakube oneagent status has no healthcheck", func(t *testing.T) {
 		builder := builder{
-			dk: &dynakube.DynaKube{},
+			dk:             &dynakube.DynaKube{},
+			hostInjectSpec: new(oneagent.HostInjectSpec{}),
```
avorima: nit: this is just `&oneagent.HostInjectSpec{}` with 4 additional characters

0sewa0:
> I mean to add a reconciler test where you reconcile a DK with the init resources field set, then fetch the daemon init container to check if the values match. The new unit tests, test only if the builder works, not if actually the values will be taken from DK and passed down to init container.

The builder works based on the DynaKube, and the `hostInjectSpec` that comes from the DynaKube.

based on this logic I should also add a test to the `pkg/controllers/dynakube/controller_test.go` to make sure it is passed correctly 😅 

it is the builder's responsibility to handle the new field correctly. I see no value in adding an extra check for something that has been already checked.

avorima (approved):

mihaitanasedt:
> > I mean to add a reconciler test where you reconcile a DK with the init resources field set, then fetch the daemon init container to check if the values match. The new unit tests, test only if the builder works, not if actually the values will be taken from DK and passed down to init container.
> 
> The builder works based on the DynaKube, and the `hostInjectSpec` that comes from the DynaKube.
> 
> based on this logic I should also add a test to the `pkg/controllers/dynakube/controller_test.go` to make sure it is passed correctly 😅
> 
> it is the builder's responsibility to handle the new field correctly. I see no value in adding an extra check what has been already checked.

They operate at different testing levels. The unit tests verify that values are computed correctly in isolation, while the reconciler tests verify that those values (not specifically the resources, but in general) are properly passed to the init container (kind of "e2e", without actually running).   

Idk, we have to align what kind of tests we write in which scenario :D 

Let's finish this PR first, we can discuss later

mihaitanasedt (approved):

PR #6958, opened 2026-07-14
Remove `oneAgentInitResources` reuse `oneAgentResources` instead

0sewa0:
## Description

[dt-rnd.atlassian.net/browse/ICP-7367](https://dt-rnd.atlassian.net/browse/ICP-7367)
Basically a simplification of https://github.com/Dynatrace/dynatrace-operator/pull/6947

Instead of a dedicated field for the init-container `resources`, will just reuse the `oneAgentResources` for the OA DaemonSet's init-container.

The logic being, that separate `resources` field only would make sense if the init-container would use more resources.
This is because  how the "effective-requests" are calculated for a pod when scheduling:
```
effective_request = max(
    max(each init container's request),
    sum(all app container requests)
)
```

This is also consistent with how we do it for the standalone logmonitoring daemonset

## How can this be tested?

Check that the OA DaemonSet's init-container resources respect the `oneAgentResources`.


pkg/controllers/dynakube/oneagent/daemonset/init.go
```diff
@@ -25,7 +24,7 @@ func (b *builder) initContainerSpec() corev1.Container {
 		Args:            b.initContainerArguments(),
 		VolumeMounts:    b.initContainerVolumeMounts(),
 		SecurityContext: b.initContainerSecurityContext(),
-		Resources:       b.initContainerResources(),
+		Resources:       b.hostInjectSpec.OneAgentResources,
```
avorima: this should use `b.resources()` same as the main container

aorcholski: init-resources are {} for `cloudNativeFullStack:{}`

0sewa0: same issue as https://github.com/Dynatrace/dynatrace-operator/pull/6958#discussion_r3579051384

0sewa0: fixed

avorima (approved):

aorcholski (approved):

PR #6970, opened 2026-07-17
Do not use `timeprovider.Provider` in the version reconciler

0sewa0:
## Description

https://dt-rnd.atlassian.net/browse/ICP-7580

This fixes the issue that autoUpdate didn't work.

## How can this be tested?

So, as this would be a long running test, here is the hacky way to check fast:

1. Create DynaKube with OA/AG/CM and set the `spec.dynatraceApiRequestThreshold: 1`
2. Check the logs that the `updating version status` doesn't only happen on the first reconcile


pkg/controllers/dynakube/version/reconciler_test.go
```diff
@@ -191,20 +182,17 @@ func TestNeedsUpdate(t *testing.T) {
 		updatedDynakube := dk.DeepCopy()
 		setOneAgentCustomImageStatus(updatedDynakube, oldImage)
 		updatedDynakube.Spec.OneAgent.ClassicFullStack.Image = newImage
-		updatedDynakube.Status.OneAgent.LastProbeTimestamp = timeProvider.Now()
-		r := Reconciler{
-			timeProvider: timeProvider,
-		}
+		now := metav1.Now()
+		updatedDynakube.Status.OneAgent.LastProbeTimestamp = &now
```
avorima: nit: use the more concise `new` form

pkg/controllers/dynakube/version/reconciler_test.go
```diff
@@ -72,48 +71,49 @@ func TestReconcile(t *testing.T) {
...
+			err = versionReconciler.ReconcileCodeModules(ctx, dk, nil, versionClient)
+			require.NoError(t, err)
+			assert.Equal(t, previousProbe, *dkStatus.CodeModules.LastProbeTimestamp)
+
+			// change if probe old enough
+			time.Sleep(15*time.Minute + 1*time.Second)
+
+			err = versionReconciler.ReconcileCodeModules(ctx, dk, nil, versionClient)
+			require.NoError(t, err)
+			assert.NotEqual(t, previousProbe, *dkStatus.CodeModules.LastProbeTimestamp)
```
avorima: we can actually assert ordering here. the library can compare `time.Time` values using Less/Greater

pkg/controllers/dynakube/version/reconciler_test.go
```diff
@@ -72,48 +71,49 @@ func TestReconcile(t *testing.T) {
...
+
+			assertStatusBasedOnTenantRegistry(t, dk.ActiveGate().GetDefaultImage(latestActiveGateVersion), latestActiveGateVersion, dkStatus.ActiveGate.VersionStatus)
+			assertStatusBasedOnTenantRegistry(t, dk.OneAgent().GetDefaultImage(latestOneAgentVersion), latestOneAgentVersion, dkStatus.OneAgent.VersionStatus)
+			assert.Equal(t, latestAgentVersion, dkStatus.CodeModules.Version)
+
+			// no change if probe not old enough
+			previousProbe := *dkStatus.CodeModules.LastProbeTimestamp
+			err = versionReconciler.ReconcileCodeModules(ctx, dk, nil, versionClient)
+			require.NoError(t, err)
+			assert.Equal(t, previousProbe, *dkStatus.CodeModules.LastProbeTimestamp)
```
avorima: for completeness, let's check all components

0sewa0:
@avorima && @andriisoldatenko 

I reworked it so that we do not use or set that timestamp and just rely on the dtclient cache.
https://github.com/Dynatrace/dynatrace-operator/pull/6970/commits/6ae220c4e00a0e37dc0ada23e64700b52bba6145
https://github.com/Dynatrace/dynatrace-operator/pull/6970/commits/45c8e74ac42a8f1b78dc1b6457b11c6a1b85c185
https://github.com/Dynatrace/dynatrace-operator/pull/6970/commits/e86cbba3c14c3a64e135472b94e4a5617fc387c9

To be fair, the version-reconciler is still a bit overcomplicated now, with all this "did the source change" or "did the custom field change" checks, that has 0 affect on the dtclient cache. Should I remove those checks now?

andriisoldatenko:
> Should I remove those checks now?

assuming we are note releasing it: gor for it, granular simplification of version reconciler is good strategy.

avorima:
@0sewa0 let's not mix things up. now we have a timestamp that doesn't get updated anymore. let's keep it focused to fixing the bug and then do the version reconciler refactoring and status changes in one go
also we shouldn't forget about the edgeconnect version reconciler + status

0sewa0:
@andriisoldatenko @avorima 

I removed all checks, still didn't drop the status fully, only the timestamp in it. (I just don't set it)
I added a small "integration" tests to check the the removal of the checks is safe and the cache works as intended.

I agree that bigger refactor (pruning the status and such) should be done later.
- IMO we won't be able to fully move to the `registry.ResolveImage` for the codemodules, due to the CSI driver
   - reason: If you wish to "resolve" the correct codemodule version/image PER CSI driver, then on a big cluster(1000node) you would do a small DDoS on the tenant 😅

pkg/controllers/dynakube/version/reconciler_autoupdate_test.go
```diff
@@ -0,0 +1,361 @@
+// Copyright 2025 Dynatrace LLC
```
andriisoldatenko, line 1:
```suggestion
reconciler_autoupdate_test.go
```
since it's a "fancy" unit tests

avorima (approved):

PR #6971, opened 2026-07-17
Do not use `timeprovider.Freeze()` in live code.

0sewa0:
## Description

https://dt-rnd.atlassian.net/browse/ICP-7580

The less code change version of https://github.com/Dynatrace/dynatrace-operator/pull/6970, this also fixes the autoUpdate bug.

Note: There are no adjustments in the unit-tests, as I can't really test if `Freeze` was called or not. (could have turned it into a testing func with `t.Testing` as a param, but that would have went against the "minimal change" idea.)

## How can this be tested?

So, as this would be a long running test, here is the hacky way to check fast:

1. Create DynaKube with OA/AG/CM and set the `spec.dynatraceApiRequestThreshold: 1`
2. Check the logs that the `updating version status` doesn't only happen on the first reconcile


avorima (approved):
I have registry issues in dev, but I can confirm that the operator logs the update line more frequently with this change

svfoxat (approved):

PR #6972, opened 2026-07-17
add cache-sync helpers for cached-client integration tests

jdanieck:
## Goal

Simplify authoring integration tests that use a cached client (initially introduced in https://github.com/Dynatrace/dynatrace-operator/pull/6757).

## What
- New `test/integrationtests/cache.go` with `WaitForCachedMatch` / `WaitForCachedGone` helpers.
- Refactor the `kubemon/statefulset` and `kubemon/connectioninfo` integration tests to use them.
- New guide in `doc/coding-style-guide.md` for integration testing with a cached client.

## Why
Reconcilers run on the manager's cached client, which is eventually consistent — a reconcile can read a stale object right after a write. Tests that assumed one reconcile was enough [flaked](https://github.com/Dynatrace/dynatrace-operator/actions/runs/29504240911/job/87640800363?pr=6757) (e.g. re-enable after delete read a stale StatefulSet and skipped recreation), which shows this is easy to get wrong without a shared pattern. Each phase now waits for the cache to reflect its setup write before reconciling, then asserts via the direct `apiReader`.

## Testing
`make go/test`

### DIFF
--- a/.github/workflows/zz-temp-stress.yaml
+++ b/.github/workflows/zz-temp-stress.yaml
@@ -1,0 +1,59 @@
+# TEMPORARY — stress-tests the kubemon cached-client integration tests to flush out
+# informer-cache timing races. Revert before merging this PR.
+name: TEMP stress integration tests
+
+on:
+  push:
+    branches: [integration-test-cache]
+  workflow_dispatch:
+    inputs:
+      parallel:
+        description: concurrent test processes per shard (induces CPU contention)
+        default: "4"
+      count:
+        description: iterations per process
+        default: "50"
+
+permissions:
+  contents: read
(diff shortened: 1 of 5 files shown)


jdanieck:
@0sewa0 @avorima I overcomplicated this. I wanted the integration tests to cover eventual consistency, but that belongs in e2e — not here. Without a real controller manager, simulating external events and cache sync means hand-gating every phase, and getting that right is too complicated for what it buys: low gain, high pain. It never caught a real bug, and a stale read just requeues and self-corrects anyway. And as I painfully learned, it's hard to get right — I made it flaky twice. **Switching these tests to a direct client.**

PR #6975, opened 2026-07-17
Fix OA DaemonSet secret mount permissions. (#6974)

0sewa0:
## Description
[dt-rnd.atlassian.net/browse/ICP-7532](https://dt-rnd.atlassian.net/browse/ICP-7532)

cherry pick of #6974

## How can this be tested?

same as #6974

### DIFF
--- a/pkg/controllers/dynakube/oneagent/daemonset/volumes.go
+++ b/pkg/controllers/dynakube/oneagent/daemonset/volumes.go
@@ -11,6 +11,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	// This should result in file permissions of -r--r--r--
+	// This is necessary because the OneAgent (in classicFullstack) may replicate these files into instrumented user pods
+	// where all processes need to be able to read these.
+	globalReadFilePerm = 0o444
+)
+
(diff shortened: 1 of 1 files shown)


avorima (approved):

pkg/controllers/dynakube/oneagent/daemonset/volumes.go
```diff
@@ -11,6 +11,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	// This should result in file permissions of -r--r--r--
+	// This is necessary because the OneAgent (in classicFullstack) may replicate these files into instrumented user pods
+	// where all processes need to be able to read these.
+	globalReadFilePerm = 0o444
```
andriisoldatenko, line 18: i know it's cheery-pick: but IMO better to say:
```
ReadFileAllPerm = 0o444
```
or add a comment:
// makes a file readable to everyone

andriisoldatenko (approved):

pkg/controllers/dynakube/oneagent/daemonset/volumes.go
0sewa0: there is already a comment with `// This should result in file permissions of -r--r--r--`

which is the clearest I can make it 😅

PR #6977, opened 2026-07-17
kubemon: ag auth reconciler

jdanieck:
## Description

Continuation of #6757. Registration in the UI is planned as a separate PR — for now it's done manually via the Settings API (step 3 below).

### What
Enables the kubemon operand to authenticate to the tenant and report as a Kubernetes cluster. Previously the operand rolled out but couldn't connect; this PR provisions its ActiveGate auth token and the runtime identity it needs to show up in the UI.

### Behavior
- **Auth token**: the operator provisions an ActiveGate auth token for the operand, keeps it rotated ahead of expiry, and removes it when kubernetesMonitoring is turned off. Token rotation automatically restarts the operand.
- **Cluster identity**: the operand is now seeded with the cluster/namespace identity and deployment metadata required to register as a monitored Kubernetes cluster.
- **Gating**: the operand only starts reporting once the cluster's `kube-system` UUID is known; everything stays behind the temporary kubemon-operand gate. A Helm install/upgrade warning now reminds users the operand is under development and not production-ready while this gate is enabled.

### How can this be tested?

1. Enable the gate and apply the DynaKube from #6757 (same steps).

2. Verify the auth-token secret is created and the operand rolls out:

```sh
kubectl -n dynatrace get secret dynakube-kubemon-authtoken-secret
kubectl -n dynatrace get statefulset dynakube-kubemon
kubectl -n dynatrace get pod -l app.kubernetes.io/name=dynakube-kubemon
```

3. Register the cluster in the Dynatrace manually via API (auto-registration lands in a follow-up). Token needs the `settings.write` scope:

```sh
CLUSTER_ID=$(kubectl get ns kube-system -o jsonpath='{.metadata.uid}')
curl -sS -X POST "$DT_ENV_URL/api/v2/settings/objects" \
  -H "Authorization: Api-Token $DT_API_TOKEN" -H "Content-Type: application/json" \
  -d '[{"schemaId":"builtin:cloud.kubernetes","schemaVersion":"3.0.0","value":{
    "enabled":true,"label":"my-kubemon-cluster","clusterIdEnabled":true,
    "clusterId":"'"$CLUSTER_ID"'","certificateCheckEnabled":true,"hostnameVerificationEnabled":true}}]'
```

4. Wait until cluster is visible under **Kubernetes** in Dynatrace UI.

5. Remove `spec.kubernetesMonitoring` and confirm the auth-token Secret (along with the StatefulSet, ConfigMap, and tenant Secret from #6757) is deleted and the condition is cleared.

### Expected state

The monitored Kubernetes cluster is visible in the Dynatrace UI:

image


config/helm/chart/default/templates/NOTES.txt
```diff
@@ -19,3 +19,6 @@ WARNING: The `webhook.highAvailability` field is deprecated and will be removed
 {{- if .Values.enableInsecurePprofEndpoint }}
 WARNING: Golang profiling endpoint is enabled. Make sure to restrict external access to the operator{{ if .Values.csidriver.enabled }}, webhook and CSI driver{{ else }} and webhook{{ end }}.
 {{- end }}
+{{- if .Values.operator.kubemonEnableOperand }}
+WARNING: The kubemon operand is enabled. It is under development and not ready for production use.
```
andriisoldatenko, line 23: just q: is it confusing to customer to read "kubemon operand "?

pkg/api/latest/dynakube/kubemon/props.go
andriisoldatenko:
```suggestion
	NameSuffix = "-kubemon"
```
why not just?

it's inside kubemon package, so kubemon.NameSuffix is easy to read.

Also in many other places we use kubemon, and not kubemonOperand.

pkg/controllers/dynakube/deploymentmetadata/config.go
```diff
@@ -20,5 +20,6 @@ const (
 
 	OneAgentMetadataKey   = "oneagent"
 	ActiveGateMetadataKey = "activegate"
+	KubemonMetadataKey    = "kubemon"
```
andriisoldatenko, line 23: for example: here we don't say `KubemonOperandMetadataKey`, so we need to be consistent.

pkg/controllers/dynakube/deploymentmetadata/reconciler.go
```diff
@@ -55,6 +57,19 @@ func (r *Reconciler) addActiveGateDeploymentMetadata(dk *dynakube.DynaKube, conf
 	configMapData[ActiveGateMetadataKey] = NewDeploymentMetadata(r.clusterID, ActiveGateMetadataKey).AsString()
 }
 
+func (r *Reconciler) addKubemonDeploymentMetadata(dk *dynakube.DynaKube, configMapData map[string]string) {
+	// Temporary gate, to be removed once kubemon is complete
+	if !k8senv.IsKubemonOperandEnabled() {
```
andriisoldatenko, line 62: here we do.

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
```diff
@@ -0,0 +1,118 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// SecretKey is the key used inside the auth token Secret, matching the AG convention.
+	SecretKey = "auth-token"
+
+	// DefaultRotationInterval mirrors the AG value: rotate at 29 days against a 60-day token
+	// validity to avoid expiry warnings in the Dynatrace UI.
+	DefaultRotationInterval = time.Hour * 24 * 29
```
andriisoldatenko:
```suggestion
    day = 24 * time.Hour
	DefaultRotationInterval = 29 * day
```
in go we usually do: number * time (it's easy to read), similar to 5 seconds in human language

also it's wired, why 24 * 29 hours? what does it mean? 29 days?

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
```diff
@@ -0,0 +1,118 @@
...
+
+// WithRotationInterval overrides the default rotation interval, e.g. to use a short-lived
+// interval in tests instead of waiting out the production value.
+func WithRotationInterval(interval time.Duration) Option {
+	return func(r *Reconciler) {
+		r.rotationInterval = interval
+	}
+}
+
+func NewReconciler(kubeClient client.Client, opts ...Option) *Reconciler {
```
andriisoldatenko: for now we only support one `WithRotationInterval` so why not just simplify stuff and pass it here, you can test it and override.

config/helm/chart/default/templates/NOTES.txt
jdanieck: I guess depends on who you ask :) we use operand term internally often and this flag is never ment to go to production - it's a temporary gate to have the code merged to main and disabled by default, we will remove it completely when operand is finished (before we ship v1.11)

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
jdanieck: yes, 29 days - code borrowed from AG; I assumed it is enough to explain in the comment, will refactor as you suggest

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
jdanieck: the functional option pattern is ubiquitous in the codebase, so just sticked with it; ultimately it's a matter of taste, I can go with the parameter if you prefer it this way

pkg/controllers/dynakube/deploymentmetadata/reconciler.go
jdanieck: I'll keep  KubemonOperand for the temporary gate only (helm, env, this function) - this code eventually will be deleted and never reach production; for everything else will use `kubemon`

jokuniew (dismissed):
:+1: one thing found in automated review - it's pointing to previous PR but it affects this code as well, that `errors.Is(err, kubemonconnectioninfo.ErrConnectionInfoNotReady)` should also be in if check there:
https://github.com/Dynatrace/dynatrace-operator/pull/6757/changes#diff-e932d14a6695f41e4645835e28c517caee0ca26f3cfcc897514b2afad3ef151bR411
So it's considered as transient error 
Apply as you wish - in this or next PRs

pkg/api/latest/dynakube/kubemon/props.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/commit/0a35b79107e44b6024cad1ff3f629e9ce9aaafc0

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6977/commits/bd85063d88bc45ad3327d1a7af0a4e9acccd27e3

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6977/commits/c2f8fdffa765e807b16d5ce8d816104e85160e0d

pkg/controllers/dynakube/deploymentmetadata/config.go
jdanieck: ack, we'll stick to kubemon here

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
```diff
@@ -126,23 +135,39 @@ func TestReconcileBuildsStatefulSet(t *testing.T) {
...
-	// env vars: capabilities first, then connection info injected envs, then custom
-	require.Len(t, container.Env, 4)
+	// env vars: capabilities, seed envs, deployment metadata, connection info, then custom
+	require.Len(t, container.Env, 7)
 	assert.Equal(t, agconsts.EnvDTCapabilities, container.Env[0].Name)
 	assert.Equal(t, activegate.KubeMonCapability.ArgumentName, container.Env[0].Value)
-	assert.Equal(t, connectioninfo.EnvDTTenant, container.Env[1].Name)
-	assert.Equal(t, connectioninfo.EnvDTServer, container.Env[2].Name)
-	assert.Equal(t, "CUSTOM", container.Env[3].Name)
+	assert.Equal(t, agconsts.EnvDTIDSeedNamespace, container.Env[1].Name)
```
waodim: Nitpick: consider `k8senv.Find` instead of fixed indices so failures name the missing var rather than showing an index mismatch. Not blocking since this might not be worth as this would add some lines of code.

jdanieck:
> 👍 one thing found in automated review - it's pointing to previous PR but it affects this code as well, that `errors.Is(err, kubemonconnectioninfo.ErrConnectionInfoNotReady)` should also be in if check there: https://github.com/Dynatrace/dynatrace-operator/pull/6757/changes#diff-e932d14a6695f41e4645835e28c517caee0ca26f3cfcc897514b2afad3ef151bR411 So it's considered as transient error Apply as you wish - in this or next PRs

fixed in https://github.com/Dynatrace/dynatrace-operator/pull/6977/commits/bea0f638439e66cd04c53ac550db489aa42996bf

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6977/commits/5881d71fe872a53889a5966fccbefc143a286a8f

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
```diff
@@ -0,0 +1,106 @@
...
+
+type Reconciler struct {
+	secrets          k8ssecret.QueryObject
+	rotationInterval time.Duration
+}
+
+// NewReconciler creates a Reconciler. rotationInterval controls how long an auth token secret is
+// kept before being rotated; pass DefaultRotationInterval in production, or a short-lived value
+// in tests instead of waiting out the production value.
+func NewReconciler(kubeClient client.Client, rotationInterval time.Duration) *Reconciler {
```
0sewa0: > or a short-lived value in tests instead of waiting out the production value.

Just use `"testing/synctest"` for messing with time

or does it not work with `envtest`?

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
waodim, line 177: Due to linting failing because of exceeded statements per function, should we revert this again? 😅

jdanieck: No, it was a good change. I'll fix the issue linter is reporting :) I am just surprised because locally my linter passed.

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -26,10 +29,15 @@ import (
...
 	ContainerName             = "kubemon"
 	AnnotationTenantTokenHash = api.InternalFlagPrefix + "kubemon-tenant-token-hash"
+	AnnotationAuthTokenHash   = api.InternalFlagPrefix + "kubemon-authtoken-hash"
 	StorageVolumeName         = "kubemon-storage"
+	AuthTokenVolumeName       = "kubemon-authtoken-secret"
 )
 
-var ErrImageRequired = errors.New("kubernetes monitoring image is required")
+var (
+	ErrImageRequired        = errors.New("kubernetes monitoring image is required")
```
0sewa0, line 39: just for clarity (for others aswell), this will be removed in https://dt-rnd.atlassian.net/browse/ICP-7295

jdanieck: yes, it is temporary as we don't have image discovery implemented yet

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
jdanieck: fixed in https://github.com/Dynatrace/dynatrace-operator/pull/6977/commits/6b4c46ea43ddbb49d44e9a575263af25f7a60fed

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
jdanieck: Doesn't work here — synctest only fakes the clock for goroutines inside its bubble, but in this test CreationTimestamp is stamped by envtest's real kube-apiserver process (used to detect rotation in isOutdated, comparing it against time.Now()), on its own real clock. There's no fake time to advance on that side, so the bubble can't help.

I already use it in the unit test though, where the fake client doesn't stamp CreationTimestamp itself, so we set it manually with metav1.Now() inside the bubble — same fake clock on both sides.

pkg/controllers/dynakube/kubemon/connectioninfo/reconciler.go
```diff
@@ -35,6 +36,8 @@ func NewReconciler(kubeClient client.Client) *Reconciler {
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context, agClient agclient.Client, dk *dynakube.DynaKube) error {
+	ctx, _ = logd.NewFromContext(ctx, "kubemon-connectioninfo")
```
0sewa0:
```suggestion
	ctx, _ = logd.NewFromContext(ctx, "connectioninfo")
```

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
```diff
@@ -0,0 +1,106 @@
...
+	return &Reconciler{
+		secrets:          k8ssecret.Query(kubeClient, kubeClient),
+		rotationInterval: rotationInterval,
+	}
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context, agClient agclient.Client, dk *dynakube.DynaKube) error {
+	ctx, log := logd.NewFromContext(ctx, "kubemon-authtoken")
```
0sewa0:
```suggestion
	ctx, log := logd.NewFromContext(ctx, "authtoken")
```

pkg/controllers/dynakube/kubemon/reconciler.go
0sewa0:
```suggestion
	ctx, log := logd.NewFromContext(ctx, "kubemon")
```

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -44,6 +52,8 @@ func NewReconciler(kubeClient client.Client) *Reconciler {
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context, dk *dynakube.DynaKube) error {
+	ctx, _ = logd.NewFromContext(ctx, "kubemon-statefulset")
```
0sewa0:
```suggestion
	ctx, _ = logd.NewFromContext(ctx, "statefulset")
```

0sewa0 (commented):
just making the logs pretty

otherwise LGTM

0sewa0 (dismissed):

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
andriisoldatenko: may be add TODO remove it with ticket number

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -66,14 +77,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, dk *dynakube.DynaKube) error
 }
 
 // buildEnvs prepends the mandatory AG runtime env vars (and optional DT_GROUP) to any user-supplied vars.
-func buildEnvs(dk *dynakube.DynaKube) []corev1.EnvVar {
+func buildEnvs(dk *dynakube.DynaKube) ([]corev1.EnvVar, error) {
```
andriisoldatenko: IMO it's just builder, and you can check error before calling it,

if dk.Status.KubeSystemUUID == "" {
		return nil, ErrMissingKubeSystemUID
}

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -212,3 +275,17 @@ func (r *Reconciler) getTenantTokenHash(ctx context.Context, dk *dynakube.DynaKu
...
 	return hash, nil
 }
+
+func (r *Reconciler) getAuthTokenHash(ctx context.Context, dk *dynakube.DynaKube) (string, error) {
+	var secret corev1.Secret
+	if err := r.kubeClient.Get(ctx, client.ObjectKey{Name: dk.KubernetesMonitoring().GetAuthTokenSecretName(), Namespace: dk.Namespace}, &secret); err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	hash, err := hasher.GenerateHash(string(secret.Data[kubemonauthtoken.SecretKey]))
```
andriisoldatenko: theoretically if `secret.Data` doesn't have a `kubemonauthtoken.SecretKey` it returns you ""

and you will add hash of empty string,

```
package main

import "fmt"

func main() {
	data := make(map[string]string)
	result := string(data["SecretKey"])
	fmt.Println(result)
}

```

So imo we should check that secret key has value and then generate hash of empty string.

```
// You can edit this code!
// Click here and start typing.
package main

import (
	"encoding/json"
	"fmt"
)

func main() {
	data := make(map[string]string)
	result := string(data["SecretKey"])
	fmt.Println(json.Marshal(result))
}
// [34 34] <nil>
```
and json doesn't safe us here

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
andriisoldatenko: make sure you run `make prerequisites` and linter / go version is the same. Sometimes it cause it

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
```diff
@@ -0,0 +1,106 @@
...
+			return errors.WithStack(err)
+		}
+
+		return r.createOrUpdateSecret(ctx, agClient, dk)
+	}
+
+	return nil
+}
+
+func (r *Reconciler) createOrUpdateSecret(ctx context.Context, agClient agclient.Client, dk *dynakube.DynaKube) error {
```
avorima: nit: the secret is only ever created, so the update part is not accurate

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
```diff
@@ -0,0 +1,106 @@
...
+	authTokenInfo, err := agClient.GetAuthToken(ctx, dk.Name)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	coreLabels := k8slabel.NewCoreLabels(dk.Name, k8slabel.KubeMonComponentLabel)
+
+	secret, err := k8ssecret.Build(dk,
```
avorima, line 84: if this secret is recreated every time we should make it immutable. the query object API actually already supports recreating. you just have to set the MustRecreate function to always return true. could even make this change globally so that all immutable secrets get recreated

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
jdanieck: added TODO in https://github.com/Dynatrace/dynatrace-operator/pull/6977/changes/12d31e465413ba71373ed47a0f95d6aa39d75ad7; I skipped the ticket id as it feels wrong to me to add internal data to public repo

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
jdanieck: Refactored in https://github.com/Dynatrace/dynatrace-operator/pull/6977/changes/10cddf6a2c8ffd3266a02274422791442526f453.

Moved the check out of the builder into an ensureReady(dk) step at the top of Reconcile, so the builder is now a pure assembler and buildEnvs dropped its error return. Also moved the ErrImageRequired check there for the same reason — both are prerequisites without which the build fails anyway, so they short-circuit in one place before any API calls.

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
jdanieck: added validation in https://github.com/Dynatrace/dynatrace-operator/pull/6977/changes/e5ff92f55b3a4c73dd9a63063aba0b968738c192

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
jdanieck: renamed in https://github.com/Dynatrace/dynatrace-operator/pull/6977/changes/9f30abd14f64a2033c7a4f102bcad0e5ecbdd3da

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6977/changes/9f30abd14f64a2033c7a4f102bcad0e5ecbdd3da

pkg/controllers/dynakube/kubemon/authtoken/reconciler.go
```diff
@@ -0,0 +1,104 @@
...
+	SecretKey = "auth-token"
+
+	// DefaultRotationInterval mirrors the legacy unified AG: rotate at 29 days against a 60-day token
+	// validity to avoid expiry warnings in the Dynatrace UI.
+	DefaultRotationInterval = 29 * day
+)
+
+type Reconciler struct {
+	secrets          k8ssecret.QueryObject
+	rotationInterval time.Duration
```
avorima: this is actually one of the instances where a fake clock is warranted. we can't run integration tests with synctest, but the goal is also to use production values. leaving things up to sleep is not a good testing strategy

jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/6977/commits/130b14ef49f9ae5993a8390a6f9a663ebf26e2eb

avorima (approved):

andriisoldatenko (approved):

PR #6981, opened 2026-07-18
fix: prevent panic on malformed ActiveGate auth token

knQzx:
## problem

`conditionSetSecretCreated` splits the ActiveGate `auth-token` secret value on `.` and takes `tokenAllParts[:2]` with no length check

`strings.Split("", ".")` returns a one-element slice, so an empty or single-segment token value panics with `slice bounds out of range [:2]`. it runs on every reconcile (the condition message is refreshed for each present, not-outdated secret), so a malformed or tampered secret wedges that DynaKube's reconcile in a panic/backoff loop

## fix

guard the length before slicing, so a token with fewer than two segments yields an empty public part instead of panicking

## how tested

added a table test calling `conditionSetSecretCreated` with empty, single-segment and valid tokens and asserting it does not panic; it panics on the old code and passes on the new

### DIFF
--- a/pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
+++ b/pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
@@ -25,12 +25,15 @@ import (
 
 const (
 	activeGateAuthTokenSecretConditionType string = "ActiveGateAuthTokenSecret"
+	malformedAuthTokenReason               string = "MalformedAuthToken"
 
 	ActiveGateAuthTokenName = "auth-token"
 
 	// Buffer to avoid warnings in the UI
 	AuthTokenBuffer           = time.Hour * 24
 	AuthTokenRotationInterval = time.Hour*24*30 - AuthTokenBuffer
+
+	authTokenPublicParts = 2
 )
 
 type Reconciler struct {
@@ -95,9 +98,7 @@ func (r *Reconciler) reconcileAuthTokenSecret(ctx context.Context, dk *dynakube.
 		return r.ensureAuthTokenSecret(ctx, dk, agClient)
 	}
 
-	r.conditionSetSecretCreated(dk, secret) // update message once a day
-
-	return nil
+	return r.conditionSetSecretCreated(dk, secret) // update message once a day
 }
 
 func (r *Reconciler) ensureAuthTokenSecret(ctx context.Context, dk *dynakube.DynaKube, agClient agclient.Client) error {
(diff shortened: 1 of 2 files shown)


0sewa0 (dismissed):

pkg/controllers/dynakube/activegate/internal/authtoken/reconciler.go
```diff
@@ -161,7 +162,11 @@ func (r *Reconciler) conditionSetSecretCreated(dk *dynakube.DynaKube, secret *co
 	lifespan := time.Since(secret.CreationTimestamp.Time)
 	days := strconv.Itoa(int(lifespan.Hours() / 24))
 	tokenAllParts := strings.Split(string(secret.Data[ActiveGateAuthTokenName]), ".")
-	tokenPublicPart := strings.Join(tokenAllParts[:2], ".")
+
+	tokenPublicPart := ""
+	if len(tokenAllParts) >= authTokenPublicParts {
+		tokenPublicPart = strings.Join(tokenAllParts[:authTokenPublicParts], ".")
```
avorima: A malformed auth token should not be ignored

avorima (changes requested):
This should not be silently ignored. The panic is clear sign that the API is not healthy.

knQzx:
Fair point - reworked it. A token with fewer than two segments now sets the condition to False with a dedicated MalformedAuthToken reason and returns an error, so the malformed token surfaces in the DynaKube status and the reconcile backs off instead of being reported as created. I kept it as a token-specific condition rather than reusing SetDynatraceAPIError, since the check also runs when reading an already-stored secret (not only right after the API call) - happy to switch to the Dynatrace API error if you'd prefer that framing. Tests updated to assert the error and the False condition for empty and single-segment tokens.

knQzx:
@avorima rebased this on main and it needs another look when you get a chance

the package moved in the meantime - #6766 removed `conditions.go` and folded the helpers into `reconciler.go`, and the condition type const became unexported - so the change is now written against that layout: `malformedAuthTokenReason` and `setAuthSecretMalformed` sit next to `setAuthSecretCreated`

behaviour is the one you asked for, not the original silent skip. a token with fewer than two dot-separated segments sets the condition to False with a `MalformedAuthToken` reason and returns the error to the caller, so an unhealthy token surfaces in the DynaKube status instead of being swallowed

the bug is still live on main - `conditionSetSecretCreated` slices `tokenAllParts[:2]` unguarded, and the new test panics there with `slice bounds out of range [:2] with capacity 1` before the fix

PR #6983, opened 2026-07-20
Fix OA communication host handling

0sewa0:
## Description

https://dt-rnd.atlassian.net/browse/ICP-7623

The `NewCommunicationHosts` expected the endpoints to be separated via `,`, which is true for the AG endpoints, but not for the OA, which can cause issues with the networkZone check + `enableIstio`.

## How can this be tested?

The most visible way is to use `enableIstio: true` with a OneAgent, and check that the `ServiceEntries` have all endpoints properly listed.

### DIFF
--- a/pkg/controllers/dynakube/connectioninfo/communication_hosts.go
+++ b/pkg/controllers/dynakube/connectioninfo/communication_hosts.go
@@ -61,15 +61,15 @@ func NewCommunicationHost(endpoint string) (CommunicationHost, error) {
 	}, nil
 }
 
-// NewOACommunicationHosts creates CommunicationHost slice from `;` separated endpoints.
+// ParseOACommunicationHosts creates CommunicationHost slice from `;` separated endpoints.
 // It removes duplicates and sorts the result for deterministic output.
-func NewOACommunicationHosts(endpoints string) ([]CommunicationHost, error) {
+func ParseOACommunicationHosts(endpoints string) ([]CommunicationHost, error) {
 	return newCommunicationHosts(endpoints, ";")
 }
 
-// NewAGCommunicationHosts creates CommunicationHost slice from `,` separated endpoints.
+// ParseAGCommunicationHosts creates CommunicationHost slice from `,` separated endpoints.
 // It removes duplicates and sorts the result for deterministic output.
-func NewAGCommunicationHosts(endpoints string) ([]CommunicationHost, error) {
+func ParseAGCommunicationHosts(endpoints string) ([]CommunicationHost, error) {
 	return newCommunicationHosts(endpoints, ",")
 }
 
--- a/pkg/controllers/dynakube/connectioninfo/communication_hosts.go
+++ b/pkg/controllers/dynakube/connectioninfo/communication_hosts.go
@@ -61,9 +61,19 @@ func NewCommunicationHost(endpoint string) (CommunicationHost, error) {
 	}, nil
 }
 
-// NewCommunicationHosts creates CommunicationHost slice from comma separated endpoints.
+// NewOACommunicationHosts creates CommunicationHost slice from `;` separated endpoints.
 // It removes duplicates and sorts the result for deterministic output.
-func NewCommunicationHosts(endpoints string) ([]CommunicationHost, error) {
+func NewOACommunicationHosts(endpoints string) ([]CommunicationHost, error) {
+	return newCommunicationHosts(endpoints, ";")
+}
+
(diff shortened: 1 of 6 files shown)


pkg/controllers/dynakube/connectioninfo/communication_hosts.go
```diff
@@ -61,17 +61,27 @@ func NewCommunicationHost(endpoint string) (CommunicationHost, error) {
...
+// NewOACommunicationHosts creates CommunicationHost slice from `;` separated endpoints.
 // It removes duplicates and sorts the result for deterministic output.
-func NewCommunicationHosts(endpoints string) ([]CommunicationHost, error) {
+func NewOACommunicationHosts(endpoints string) ([]CommunicationHost, error) {
+	return newCommunicationHosts(endpoints, ";")
+}
+
+// NewAGCommunicationHosts creates CommunicationHost slice from `,` separated endpoints.
+// It removes duplicates and sorts the result for deterministic output.
+func NewAGCommunicationHosts(endpoints string) ([]CommunicationHost, error) {
```
andriisoldatenko: nit: why not Prefix with Split to follow go pattern, New for me is more like a constructor ?
Split is more natural -> splits strings into slice :)

andriisoldatenko:
```suggestion
func SplitAGCommunicationHosts(endpoints string) ([]CommunicationHost, error) {
```

pkg/controllers/dynakube/connectioninfo/communication_hosts.go
```diff
@@ -61,17 +61,27 @@ func NewCommunicationHost(endpoint string) (CommunicationHost, error) {
...
+	return newCommunicationHosts(endpoints, ",")
+}
+
+func newCommunicationHosts(endpoints string, sep string) ([]CommunicationHost, error) {
```
andriisoldatenko, line 76:
```suggestion
func splitCommunicationHosts(endpoints string, sep string) ([]CommunicationHost, error) {
```

pkg/controllers/dynakube/connectioninfo/communication_hosts.go
0sewa0: because it "constructs" a new struct for you

andriisoldatenko: this is not a struct `[]CommunicationHost` it;s slice.

@avorima wdyt?

avorima: i don't have a strong opinion either way. the function used to be called parse communication endpoints which is more accurate than split

andriisoldatenko: parse communication endpoints - sgtm as well.

0sewa0: [went with](https://github.com/Dynatrace/dynatrace-operator/pull/6983/commits/faf80c3cec142bf46391f23c869cc1c875d44bd6) `Parse....`

andriisoldatenko (approved):

pkg/controllers/dynakube/connectioninfo/oneagent/networkzone_test.go
```diff
@@ -74,52 +74,52 @@ func TestHasStaleNetworkZoneEndpoints(t *testing.T) {
...
 		// Cluster still advertises the previous IPv6 address.
-		endpoints := "https://10.0.0.1:443/communication,https://[2001:db8::2]:443/communication,https://" + localServiceHost + ":443/communication"
+		endpoints := "https://10.0.0.1:443/communication;https://[2001:db8::2]:443/communication;https://" + localServiceHost + ":443/communication"
 		assert.True(t, hasStaleNetworkZoneEndpoints(dk, endpoints))
 	})
 
 	t.Run("unparseable endpoint string → not stale (best effort skip)", func(t *testing.T) {
 		dk := newDynaKubeWithAG([]string{"10.0.0.1"})
-		// "garbage" makes NewCommunicationHosts return an error; the function defers
+		// "garbage" makes NewOACommunicationHosts return an error; the function defers
```
avorima: nit: search and replace NewOACommunicationHosts and NewAGCommunicationHosts

0sewa0: done

PR #6985, opened 2026-07-20
Fix update-csv-bundles: scoped git restore, missing deploy manifests, and CSV description updates

waodim:
## Description

[ICP-7626](https://dt-rnd.atlassian.net/browse/ICP-7626)

Fixes three issues with the automated OLM bundle workflow, observed in the v1.10.0 bot PR (#6965) where Chart.yaml, schema.yaml, and the kubernetes deploy manifests had to be committed manually.

**Scoped `git restore` in `bundle.mk`**
`make bundle` called `git restore config/` after each run, which reset Chart.yaml and schema.yaml to their pre-release snapshot values. The kubernetes bundle runs after the first restore, so `helm template` baked `app.kubernetes.io/version: 0.0.0-snapshot` into every kubernetes OLM manifest. The restore is now scoped to only the intermediate files that actually need cleanup (`config/crd/` and `config/helm/chart/default/templates/`), leaving Chart.yaml, schema.yaml, and the generated deploy manifests intact.

**Missing deploy manifest paths in `add-paths`**
`config/deploy/kubernetes/kubernetes.yaml` and `config/deploy/openshift/openshift.yaml` were generated by `make bundle` but never listed in the `create-pull-request` add-paths block (or the dry-run artifact upload), so they were silently excluded from the bot PR.

**CSV description updates**
- Added `cloudNativeFullStack` to the OneAgent capability list
- Added `dynatrace-api` to the ActiveGate capability list
- Added missing top-level capability sections: Log Monitoring, Extensions, and KSPM
- Removed the deprecated `paasToken` secret field from the installation instructions
- Removed the Dynatrace Managed URL reference (EOL)
- Removed stale "Currently" wording

## How can this be tested?

Trigger the `Update CSV Bundles` workflow via `workflow_dispatch` with `dry-run: true` and verify the downloaded artifact contains correct versions in `kubernetes.yaml` / `openshift.yaml` and that both files are present alongside the OLM bundle content.

[ICP-7626]: https://dt-rnd.atlassian.net/browse/ICP-7626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/update-csv-bundles/action.yml
+++ b/.github/actions/update-csv-bundles/action.yml
@@ -125,7 +125,9 @@ runs:
         add-paths: |
           config/olm/**
           config/deploy/kubernetes/kustomization.yaml
+          config/deploy/kubernetes/kubernetes.yaml
           config/deploy/openshift/kustomization.yaml
+          config/deploy/openshift/openshift.yaml
           config/manifests/bases/**
           config/helm/**
         commit-message: "[Automatic] Update OLM bundles for v${{ inputs.version }}"
@@ -141,7 +143,9 @@ runs:
         path: |
           config/olm/
           config/deploy/kubernetes/kustomization.yaml
+          config/deploy/kubernetes/kubernetes.yaml
           config/deploy/openshift/kustomization.yaml
+          config/deploy/openshift/openshift.yaml
           config/manifests/bases/
           config/helm/
         if-no-files-found: error
--- a/config/manifests/bases/dynatrace-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/dynatrace-operator.clusterserviceversion.yaml
@@ -21,10 +21,11 @@ spec:
   description: |
     The Dynatrace Operator supports rollout and lifecycle management of various Dynatrace components in Kubernetes and OpenShift.
 
-    Currently the Dynatrace Operator supports the following capabilities:
+    The Dynatrace Operator supports the following capabilities:
 
     ### OneAgent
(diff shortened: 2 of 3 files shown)


avorima:
* [x] both files appear in the archive from this branch
* [x] 0.0.0-snapshot is gone from OLM bundle

the CSV base (config/manifests/bases/dynatrace-operator.clusterserviceversion.yaml) remains modified with the new git restore. i think maybe we should start checking that in. it should change together with the generated CRD bases

waodim:
> * [x]  both files appear in the archive from this branch
> 
>     * [x]  0.0.0-snapshot is gone from OLM bundle
> 
> 
> the CSV base (config/manifests/bases/dynatrace-operator.clusterserviceversion.yaml) remains modified with the new git restore. i think maybe we should start checking that in. it should change together with the generated CRD bases

Isn't this covered already here? https://github.com/Dynatrace/dynatrace-operator/pull/6985/changes#diff-c350838c7d02df6e4909e6951ce3fe82e516b05b176764dd23e90be65c122c3aR149c

avorima:
no, previously the file was restored so there's no diff but now every time you run `make bundle` there's a diff

waodim:
> no, previously the file was restored so there's no diff but now every time you run `make bundle` there's a diff

Thanks I was confused. Added the bases back to restoring it

avorima (approved):

PR #7004, opened 2026-07-22
Remove control characters from PMC properties

Incredible-O:
## Description
Under certain circumstances, tenant settings values accept newlines, which are passed to the bootstrapper config, then written to `ruxitagentproc.conf`.
This is problematic because control characters like newlines could let property values inject additional INI sections.

Relevant ticket: ICP-6290.

## How can this be tested?
Unit tests

### DIFF
--- a/pkg/injection/namespace/bootstrapperconfig/pmc.go
+++ b/pkg/injection/namespace/bootstrapperconfig/pmc.go
@@ -14,6 +14,7 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8sconditions"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/objects/k8ssecret"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/sanitize"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


pkg/injection/namespace/bootstrapperconfig/pmc.go
```diff
@@ -84,6 +82,8 @@ func (s *SecretGenerator) preparePMC(ctx context.Context, dk *dynakube.DynaKube)
 
 	pmConfig.SortPropertiesByKey()
 
+	sanitizePMC(log, dk, pmConfig)
```
andriisoldatenko, line 85: more like design question, why not implement it inside sorting? and also a method?
```
pmConfig.sanitize()
```

pkg/injection/namespace/bootstrapperconfig/pmc.go
```diff
@@ -96,6 +96,27 @@ func (s *SecretGenerator) preparePMC(ctx context.Context, dk *dynakube.DynaKube)
 	return marshaled, nil
 }
 
+// sanitizePMC removes control characters from PMC properties to prevent
+// injection of malicious INI sections/keys in ruxitagentproc.conf.
+func sanitizePMC(log logd.Logger, dk *dynakube.DynaKube, pmConfig *oneagentclient.ProcessModuleConfig) {
```
0sewa0, line 101: Shouldn't this be fixed:
a) in the API/tenant side, so we don't get nasty values to begin with?
b) in the bootstrapper, that will [actually write the files](https://github.com/Dynatrace/dynatrace-bootstrapper/blob/main/pkg/configure/oneagent/pmc/create.go#L12)? (this would also catch things if they tamper with the secret in the user-ns, and not just if they mess on the API side)

0sewa0 (changes requested):

pkg/injection/namespace/bootstrapperconfig/pmc.go
Incredible-O: Discussed this Marcell offline; we've decided to only implement this change in the bootstrapper for now (which I was going to to anways 😄)

Incredible-O:
Will be implemented in the bootstrapper.

PR #7009, opened 2026-07-22
fix(api): don't omit explicitly set bool fields

VLazarevic:
## Description

https://dt-rnd.atlassian.net/browse/ICP-1012

The boolean fields `skipCertCheck` and `enableIstio` used `omitempty` on a
non-pointer `bool`. Because of this, an explicitly set `false` was dropped
during conversion and did not show up in the live manifest. Users who set these
fields to `false` on purpose could not see them, and in ArgoCD this caused a
permanent diff (`false` in Git, nothing in the live object).

Following a review discussion, the fix uses `*bool` + `omitempty` (tri-state):
- `nil` → field is omitted (not set by the user)
- `true` / `false` → always rendered, including an explicit `false`

Changes:
- `skipCertCheck` and `enableIstio` changed to `*bool` in `latest` and `v1beta5`.
- All readers now use `ptr.Deref(..., false)`; `v1beta4` conversion bridges
  `bool` ↔ `*bool` with `ptr.To` / `ptr.Deref`. `v1beta4` types stay `bool`
  (unserved).
- Regenerated deepcopy. No CRD change: the fields already have
  `+kubebuilder:validation:Optional`, so optionality never depended on `omitempty`.
- Added serialization tests and updated the conversion golden files.

While searching for other fields with the same issue (acceptance criteria), I
also included `ActiveGate.UseEphemeralVolume` and
`ExtensionExecutionController.UseEphemeralVolume`, since they have the exact same
bug and the fix is identical. Fields that were intentionally left as-is:
`kubemon.AppEnabled` (unused, superseded by a feature flag), all `Port uint32`
(0 is not a meaningful value), status fields (server-managed), and edgeconnect
(different CRD).

One pre-existing note: `ActiveGate.UseEphemeralVolume` has no `v1beta4`
conversion site. This is unrelated to this change and left as-is.

## How can this be tested?

The effect is only visible for a **served, non-storage version** (`v1beta5`),
because that is the version that goes through conversion. A `v1beta6` object is
the storage version and is returned as-is, so it does not show the difference.

1. Deploy the operator image from this branch (the fix is in the conversion
   webhook, not in the CRD).
2. Apply a DynaKube as `v1beta5` with the field set explicitly:
```yaml
   apiVersion: dynatrace.com/v1beta5
   kind: DynaKube
   metadata:
     name: dynakube
     namespace: dynatrace
   spec:
     apiUrl: https://ENVIRONMENTID.live.dynatrace.com/api
     skipCertCheck: false
```
3. Check that the explicit `false` survives conversion:
```bash
   kubectl -n dynatrace get dynakube dynakube -o jsonpath='{.spec.skipCertCheck}'
```
   Expected: `false` (before the fix this was empty).
4. ArgoCD: with a `v1beta5` DynaKube tracked by an app, it stays **Synced** with
   no permanent diff on `skipCertCheck`.


andriisoldatenko:
@VLazarevic is it ready for review?

avorima (approved):

PR #7017, opened 2026-07-23
Fix webhook CrashLoopBackOff on gVisor nodes

waodim:
## Description

[ICP-7883](https://dt-rnd.atlassian.net/browse/ICP-7883)

The `webhook` pod crashes immediately when scheduled on a gVisor node
(`runtimeClassName: gvisor`). gVisor interprets a trailing slash on a
volumeMount `mountPath` as an instruction to create a new directory as a mountpoint.
When the target path already exists inside the container image, gVisor returns
`file exists` and fails container startup.

Removed the trailing slash from the `certs-dir` volumeMount `mountPath`.

## How can this be tested?

1. Create a gVisor node pool on your GKE cluster:

```
   gcloud container node-pools create gvisor-pool \
     --cluster=<cluster-name> \
     --region=<region> \
     --sandbox type=gvisor \
     --num-nodes=1 \
     --machine-type=e2-standard-4
```

2. Deploy the operator:

   make deploy

3. Patch the webhook Deployment to use the gVisor runtime class:

```
   kubectl patch deployment dynatrace-webhook -n dynatrace \
     --type=merge \
     -p '{"spec":{"template":{"spec":{"runtimeClassName":"gvisor"}}}}'
```

4. Verify the rollout completes without CrashLoopBackOff:

   kubectl rollout status deployment/dynatrace-webhook -n dynatrace

   Before the fix, pods would fail immediately with:
   "failed to create directory mountpoint ... serving-certs/: file exists"

   After the fix, the rollout should complete successfully.

[ICP-7883]: https://dt-rnd.atlassian.net/browse/ICP-7883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
+++ b/config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
@@ -111,7 +111,7 @@ spec:
               {{- toYaml (.Values.webhook).limits | nindent 14 }}
           volumeMounts:
             - name: certs-dir
-              mountPath: /tmp/k8s-webhook-server/serving-certs/
+              mountPath: /tmp/k8s-webhook-server/serving-certs
           securityContext:
           {{- include "webhook.securityContext" . | nindent 12 }}
       serviceAccountName: dynatrace-webhook
--- a/config/helm/chart/default/tests/Common/webhook/deployment-webhook_test.yaml
+++ b/config/helm/chart/default/tests/Common/webhook/deployment-webhook_test.yaml
@@ -185,7 +185,7 @@ tests:
                     memory: 128Mi
                 volumeMounts:
                   - name: certs-dir
-                    mountPath: /tmp/k8s-webhook-server/serving-certs/
+                    mountPath: /tmp/k8s-webhook-server/serving-certs
                 securityContext:
                   seccompProfile:
                     type: RuntimeDefault
@@ -398,7 +398,7 @@ tests:
                     memory: 128Mi
                 volumeMounts:
                   - name: certs-dir
-                    mountPath: /tmp/k8s-webhook-server/serving-certs/
+                    mountPath: /tmp/k8s-webhook-server/serving-certs
                 securityContext:
                   seccompProfile:
                     type: RuntimeDefault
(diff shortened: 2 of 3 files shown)


config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
```diff
@@ -111,7 +111,7 @@ spec:
               {{- toYaml (.Values.webhook).limits | nindent 14 }}
           volumeMounts:
             - name: certs-dir
-              mountPath: /tmp/k8s-webhook-server/serving-certs/
+              mountPath: /tmp/k8s-webhook-server/serving-certs
```
0sewa0, line 114: wouldn't volumes-mounts done by the operator also be affected?

like https://github.com/Dynatrace/dynatrace-operator/blob/2a39d88a0ee1fbb61e2d22db02520b3dc92ffc80/pkg/controllers/dynakube/oneagent/daemonset/daemonset.go#L49
(this is the only one I found)

waodim: Good point let me test this

waodim: I verified with a minimal pod mimicking the mount:

```
volumeMounts:
  - name: test-vol
    mountPath: /mnt/dynatrace/certs/activegate/
```

With `runtimeClassName: gvisor` that pod fails with the exact same file exists error, so the trailing slash is indeed a problem in principle.

However, when I patched the actual OA DS to use `runtimeClassName: gvisor`, it never even creates the pods:


`Warning  FailedCreate      2m27s (x17 over 2m39s)  daemonset-controller  Error creating: pods "dynakube-oneagent-" is forbidden: HostPath is not allowed: "/"`

it seems gVisor blocks the pod at creation because OA mounts the host root.

I can still fix it. WDYT?

0sewa0: fix it, just for consistency

waodim: done https://github.com/Dynatrace/dynatrace-operator/pull/7017/commits/6c2f30d58d43ba00c9ed0dfc241c1fee160aac16

did not find another mount with a trailing slash

avorima (approved):

0sewa0 (approved):

PR #7024, opened 2026-07-24
ci: nightly chart and image for daily fips e2e runs

svfoxat:
## Description

https://dt-rnd.atlassian.net/browse/ICP-7887

- instead of using local chart + image, fips now also uses the `nightly` chart and the `-fips` variant of the nightly image for daily e2e testruns


avorima (approved):
we really need to fix that imageRef.tag handling. i don't like this repository string magic

PR #7025, opened 2026-07-24
Fix order of precedence for enrichment rules

avorima:
## Description

Fixes a regression in the order that enrichment rules that were applied in. The first rule that resolves to a value wins. The only exception are rules without a target field. This behavior is consistent with pre 1.10 releases and is only possible for legacy schema rules.

ICP-7916

## How can this be tested?

* Create rules for the test. The following links will take you to the environment-scoped rules. Click "Go to entity" to manage the entity-scoped rules.
  * Legacy schema: https://TENANT/ui/apps/dynatrace.settings/settings/kubernetes-telemetry-enrichment
  * New schema: https://TENANT/ui/apps/dynatrace.settings/settings/ingest-enrichment-config
* Deploy a DynaKube with metadata enrichment enabled
* Check that the rules in the status have the same order as the ones created in the UI. Environment-scoped rules should have lower precedence than entity-scoped
* Create a workload, e.g. `kubectl create deployment nginx --image nginx`
* Check the pod metadata to ensure correct values were resolved


svfoxat (approved):
- env scoped have lower precedence than entity scoped ✅ 
- resolved correctly on pod ✅

0sewa0 (approved):

PR #7026, opened 2026-07-24
Log enrichment rules that are omitted from the DynaKube status

avorima:
## Description

The backend has several more rule types available than the operator is able to process.
In some cases users might expect rules to be processed, so log the ones that are skipped to provide more context.

ICP-7797

## How can this be tested?

* Deploy the operator
* Create rules with conditions or unsupported types: https://TENANT/ui/apps/dynatrace.settings/settings/ingest-enrichment-config
* Create a DynaKube with metadata enrichment enabled
* Verify the status only contains supported rules
* Check the skipped rules are logged


pkg/clients/dynatrace/settings/enchrichment_test.go
```diff
@@ -307,3 +308,77 @@ func Test_enrichmentSchemaModel(t *testing.T) {
...
+					{Type: "K8S_NAMESPACE_LABEL", ValueSource: "d"},
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var logSink capturingLogSink
+			ctx := logr.NewContext(context.Background(), logr.New(&logSink))
```
andriisoldatenko:
```suggestion
			ctx := logr.NewContext(t.Context(), logr.New(&logSink))
```

0sewa0 (approved):

PR #7027, opened 2026-07-24
Allow connection timeout configuration for Dynatrace Client

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/ICP-7877)

## Description

The default client connection timeout is set to
-  30s -> operator
- 15m -> CSI provisioner
- 15m -> bootstrapper subcommand (hardcoded)

They can be configured via values.yaml using new 
- `operator.clientConnectionTimeout` property
- `csidriver.clientConnectionTimeout` property

DT_CLIENT_CONNECTION_TIMEOUT env var is used to pass the value to operator/provisioner containers.

## How can this be tested?

unittests

1) deploy CNFS with CSI driver. Check Operator/Provisioner logs for DT_CLIENT_CONNECTION_TIMEOUT. Change the timeout settings and repeat the test.

### DIFF
--- a/cmd/bootstrapper/download/client.go
+++ b/cmd/bootstrapper/download/client.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
 	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/oneagent"
 	"github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer/binary"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8senv"
 )
 
 type Client struct {
@@ -46,7 +47,7 @@ func New(options ...Option) *Client {
 }
 
 func (cl *Client) Do(ctx context.Context, inputDir string, targetDir string, props binary.Properties) error {
-	dtClient, err := cl.createDTClientFromFs(inputDir)
+	dtClient, err := cl.createDTClientFromFs(ctx, inputDir)
 	if err != nil {
 		return err
 	}
@@ -61,7 +62,7 @@ func (cl *Client) Do(ctx context.Context, inputDir string, targetDir string, pro
 	return err
 }
 
-func (cl *Client) createDTClientFromFs(inputDir string) (oneagent.Client, error) {
+func (cl *Client) createDTClientFromFs(ctx context.Context, inputDir string) (oneagent.Client, error) {
 	config, err := configFromFs(inputDir)
 	if err != nil {
 		return nil, err
@@ -80,7 +81,7 @@ func (cl *Client) createDTClientFromFs(inputDir string) (oneagent.Client, error)
 		options = append(options, dynatrace.WithCerts(certs))
 	}
 
-	options = append(options, dynatrace.WithBaseURL(cl.baseURL))
+	options = append(options, dynatrace.WithBaseURL(cl.baseURL), dynatrace.WithConnectionTimeout(k8senv.GetDTClientConnectionTimeout(ctx)))
 
 	dtClient, err := dynatrace.NewClient(options...)
 	if err != nil {
--- a/cmd/bootstrapper/download/client.go
+++ b/cmd/bootstrapper/download/client.go
@@ -7,12 +7,12 @@ import (
 	"context"
 	"os"
 	"path/filepath"
-	"time"
 
 	"github.com/Dynatrace/dynatrace-bootstrapper/pkg/configure/oneagent/ca"
 	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
 	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/oneagent"
 	"github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer/binary"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8senv"
 )
 
 type Client struct {
@@ -21,8 +21,6 @@ type Client struct {
 }
 
 type Option func(*Client)
-
-const InitContainerDTClientConnectionTimeout = 15 * time.Minute
 
 func WithInstaller(builder binary.NewFunc) Option {
 	return func(cl *Client) {
@@ -83,7 +81,7 @@ func (cl *Client) createDTClientFromFs(inputDir string) (oneagent.Client, error)
 		options = append(options, dynatrace.WithCerts(certs))
 	}
 
-	options = append(options, dynatrace.WithBaseURL(cl.baseURL), dynatrace.WithConnectionTimeout(InitContainerDTClientConnectionTimeout))
+	options = append(options, dynatrace.WithBaseURL(cl.baseURL), dynatrace.WithConnectionTimeout(k8senv.DefaultCSIDriverDTClientConnectionTimeout))
 
 	dtClient, err := dynatrace.NewClient(options...)
 	if err != nil {
--- a/cmd/bootstrapper/download/client.go
+++ b/cmd/bootstrapper/download/client.go
@@ -7,12 +7,12 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/Dynatrace/dynatrace-bootstrapper/pkg/configure/oneagent/ca"
 	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
 	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/oneagent"
 	"github.com/Dynatrace/dynatrace-operator/pkg/injection/codemodule/installer/binary"
-	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8senv"
 )
 
 type Client struct {
@@ -21,6 +21,8 @@ type Client struct {
 }
 
 type Option func(*Client)
+
+const InitContainerDTClientConnectionTimeout = 15 * time.Minute
 
 func WithInstaller(builder binary.NewFunc) Option {
 	return func(cl *Client) {
@@ -47,7 +49,7 @@ func New(options ...Option) *Client {
 }
 
 func (cl *Client) Do(ctx context.Context, inputDir string, targetDir string, props binary.Properties) error {
-	dtClient, err := cl.createDTClientFromFs(ctx, inputDir)
+	dtClient, err := cl.createDTClientFromFs(inputDir)
 	if err != nil {
 		return err
 	}
@@ -62,7 +64,7 @@ func (cl *Client) Do(ctx context.Context, inputDir string, targetDir string, pro
 	return err
 }
 
-func (cl *Client) createDTClientFromFs(ctx context.Context, inputDir string) (oneagent.Client, error) {
+func (cl *Client) createDTClientFromFs(inputDir string) (oneagent.Client, error) {
 	config, err := configFromFs(inputDir)
 	if err != nil {
 		return nil, err
@@ -81,7 +83,7 @@ func (cl *Client) createDTClientFromFs(ctx context.Context, inputDir string) (on
 		options = append(options, dynatrace.WithCerts(certs))
 	}
 
-	options = append(options, dynatrace.WithBaseURL(cl.baseURL), dynatrace.WithConnectionTimeout(k8senv.GetDTClientConnectionTimeout(ctx)))
+	options = append(options, dynatrace.WithBaseURL(cl.baseURL), dynatrace.WithConnectionTimeout(InitContainerDTClientConnectionTimeout))
 
 	dtClient, err := dynatrace.NewClient(options...)
 	if err != nil {
--- a/config/helm/chart/default/templates/_helpers.tpl
+++ b/config/helm/chart/default/templates/_helpers.tpl
@@ -195,9 +195,18 @@ topologySpreadConstraints:
 {{- end }}
 {{- end -}}
 
-{{- define "dynatrace-operator.connection-timeout-env" -}}
-  {{- if .Values.operator.connectionTimeout -}}
-    {{- with .Values.operator.connectionTimeout }}
+{{- define "dynatrace-operator.client-connection-timeout-env" -}}
+  {{- if .Values.operator.clientConnectionTimeout -}}
+    {{- with .Values.operator.clientConnectionTimeout }}
+- name: DT_CLIENT_CONNECTION_TIMEOUT
+  value: {{ . | quote }}
+    {{- end }}
+  {{- end -}}
+{{- end -}}
+
+{{- define "dynatrace-csidriver.client-connection-timeout-env" -}}
+  {{- if .Values.csidriver.clientConnectionTimeout -}}
+    {{- with .Values.csidriver.clientConnectionTimeout }}
 - name: DT_CLIENT_CONNECTION_TIMEOUT
   value: {{ . | quote }}
     {{- end }}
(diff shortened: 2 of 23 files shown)


avorima:
Are we going to keep the simpler fix in 1.10 like we did for #6971 and #6970 ?

config/helm/chart/default/values.yaml
```diff
@@ -58,6 +58,7 @@ operator:
   requeueAfter: "" # requeue period for the controller, defaults to 15m
   clientCacheCleanupInterval: "" # defined in the Golang time.Duration format, like "30m" == 30 minutes. Defaults to 1h
   kubemonEnableOperand: false # temporary gate; set to true to enable kubemon in dev/test. Remove when kubemon is complete.
+  connectionTimeout: "" # defined in the Golang time.Duration format, like "30m" == 30 minutes. Defaults to 15m
```
avorima: this is used used for webhook and csi as well. also name should be clientConnectionTimeout because connectionTimeout on its own is a bit ambiguous

pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -33,6 +33,12 @@ const (
 	minDTClientCacheCleanInterval     = 5 * time.Minute
 	maxDTClientCacheCleanInterval     = 100 * time.Hour
 
+	DTClientConnectionTimeout = "DT_CLIENT_CONNECTION_TIMEOUT"
+	// DefaultDTClientConnectionTimeout enough time to download an OneAgent package of about 1GB in size
+	DefaultDTClientConnectionTimeout = 15 * time.Minute
+	minDTClientConnectionTimeout     = 1 * time.Second
```
avorima: this should be a realistic timeout, e.g. 30 seconds. tests should use fake clocks or synctest

config/helm/chart/default/templates/_helpers.tpl
```diff
@@ -194,3 +194,21 @@ topologySpreadConstraints:
   value: {{ . | quote }}
 {{- end }}
 {{- end -}}
+
+{{- define "dynatrace-operator.client-connection-timeout-env" -}}
+  {{- if .Values.operator.clientConnectionTimeout -}}
```
avorima: the `with` is enough. i don't know if this warrants an extra helper it doesn't do anything special

cmd/bootstrapper/download/client.go
```diff
@@ -21,6 +22,8 @@ type Client struct {
 
 type Option func(*Client)
 
+const InitContainerDTClientConnectionTimeout = 15 * time.Minute
```
avorima: would be better to use the default CSI constant to have all values in one place

pkg/clients/dynatrace/dynakube.go
```diff
@@ -12,18 +12,29 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-type ClientFactory func(ctx context.Context, apiReader client.Reader, dk *dynakube.DynaKube, apiToken, paasToken, userAgentSuffix string) (*Client, error)
+type ClientFactory func(ctx context.Context, apiReader client.Reader, dk *dynakube.DynaKube, apiToken, paasToken, userAgentSuffix string, options ...Option) (*Client, error)
 
 // NewClientFromDynakube creates a new Dynatrace dtClient using the provided DynaKube configuration and tokens.
-func NewClientFromDynakube(ctx context.Context, apiReader client.Reader, dk *dynakube.DynaKube, apiToken, paasToken, userAgentSuffix string) (*Client, error) {
-	opts, err := optionsFromDynakube(ctx, apiReader, dk, apiToken, paasToken, userAgentSuffix)
+func NewClientFromDynakube(ctx context.Context, apiReader client.Reader, dk *dynakube.DynaKube, apiToken, paasToken, userAgentSuffix string, options ...Option) (*Client, error) {
```
avorima: this API was meant to prevent having to use options, which is why the user agent suffix is used here even though the operator doesn't use it. actually the timeout might be a required value rather than optional

pkg/controllers/csi/provisioner/controller_test.go
```diff
@@ -462,3 +466,67 @@ func createToken(t *testing.T, dk *dynakube.DynaKube) *corev1.Secret {
...
+			provisioner.dtClientFactory = testDTClientBuilder(t, testCase.timeout)
+
+			dtClient, err := buildDtc(&provisioner, context.Background(), dk)
+			require.NoError(t, err)
+			assert.NotNil(t, dtClient)
+		}
+	})
+}
+
+func testDTClientBuilder(t *testing.T, timeout time.Duration) func(ctx context.Context, apiReader client.Reader, dk *dynakube.DynaKube, apiToken, paasToken, userAgentSuffix string, options ...dynatrace.Option) (*dynatrace.Client, error) {
```
avorima: use the dynatrace.ClientFactory alias

pkg/controllers/dynakube/controller_test.go
```diff
@@ -1129,6 +1130,95 @@ func TestSetupTokensAndClientForPlatformToken(t *testing.T) {
...
+
+	dk := &dynakube.DynaKube{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testName,
+			Namespace: testNamespace,
+		},
+		Spec: dynakube.DynaKubeSpec{APIURL: server.URL},
+	}
+
+	t.Run("the default connection timeout is applied", func(t *testing.T) {
```
avorima: this should be a test case in the table

pkg/controllers/dynakube/controller_test.go
```diff
@@ -1129,6 +1130,95 @@ func TestSetupTokensAndClientForPlatformToken(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testName,
+			Namespace: testNamespace,
+		},
+		Data: map[string][]byte{
+			token.APIKey: []byte(testAPIToken),
+		},
+	})
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
avorima: let's keep the tests focused on whether the config is correct. making an actual request succeed breaks too many layers of abstraction

config/helm/chart/default/templates/_helpers.tpl
aorcholski: sorry, my bad
the condition is not needed to make sure the env var is defined only when the value is not-empty

pkg/controllers/dynakube/controller_test.go
aorcholski: `httptest` is the only option to get no error. However you are right , we know that the dtClient is configured properly before a communication error is returned. So I'm just making sure that the attempt to establish a connection cannot be made due to the lack of a scheme.

pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -143,6 +150,14 @@ func GetDTClientCacheCleanInterval(ctx context.Context) time.Duration {
 	return parseDuration(ctx, DTClientCacheCleanInterval, defaultDTClientCacheCleanInterval, minDTClientCacheCleanInterval, maxDTClientCacheCleanInterval)
 }
 
+func GetOperatorDTClientConnectionTimeout(ctx context.Context) time.Duration {
+	return parseDuration(ctx, DTClientConnectionTimeout, DefaultOperatorDTClientConnectionTimeout, minDTClientConnectionTimeout, maxDTClientConnectionTimeout)
```
andriisoldatenko: i know this function is very confusing, `parseDuration` i miss it before.

It's unclear how it works, especially when we pass 3 different values. Also i don't understand why we need 3 values? and for which purpose we need min value?

pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -33,6 +33,13 @@ const (
 	minDTClientCacheCleanInterval     = 5 * time.Minute
 	maxDTClientCacheCleanInterval     = 100 * time.Hour
 
+	DTClientConnectionTimeout = "DT_CLIENT_CONNECTION_TIMEOUT"
```
andriisoldatenko: DTClientConnectionTimeoutEnvVar may be?

it's not a timeout it's string with envVar name

pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -33,6 +33,13 @@ const (
...
+	// DefaultCSIDriverDTClientConnectionTimeout enough time to download an OneAgent package of about 1GB in size
+	DefaultCSIDriverDTClientConnectionTimeout = 15 * time.Minute
+	DefaultOperatorDTClientConnectionTimeout  = 30 * time.Second
```
waodim: For the CSI driver, a user who sets `csidriver.clientConnectionTimeout`: 45s gets exactly 45s and then sees OneAgent installations silently fail mid-download on any normal connection.

Suggestion:

- Raise `minDTClientConnectionTimeout` for the CSI context (would require separate min constants for each getter)?
- Keep the current min but add a warning to the values.yaml comment, e.g.:

```
clientConnectionTimeout: "" # ...
                            # Values below ~5m will/can cause failures when downloading large OneAgent packages (~1GB).
```

WDYT?

pkg/util/kubernetes/fields/k8senv/env.go
andriisoldatenko: @0sewa0 proposing to name as: `getSafeDurationFromEnv` for example.

pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -33,6 +33,13 @@ const (
...
+	minDTClientConnectionTimeout              = 30 * time.Second
+	maxDTClientConnectionTimeout              = 15 * time.Minute
```
andriisoldatenko, line 41: also we can say:
maxDTClientConnectionTimeout = 15 * time.Minute
minDTClientConnectionTimeout = 30 * time.Second
DefaultCSIDriverDTClientConnectionTimeout = maxDTClientConnectionTimeout
DefaultOperatorDTClientConnectionTimeout = minDTClientConnectionTimeout
just easy to read

0sewa0: > [!IMPORTANT]
>  TLDR: This `parseDuration` and related const IMO could be done better, but should be handled outside of this PR.


I did a small thing (DO NOT DO THIS IN CONTEXT OF THIS PR), just an idea how maybe we could escape this consts hell
```go
package k8senv

import (
	"context"
	"os"
	"time"

)

func GetWebhookCertsRootDurationNew(ctx context.Context) BoundedDuration {
	return BoundedDuration{
		EnvName:      "DT_WEBHOOK_CERTS_ROOT_DURATION",
		MinValue:     12 * time.Hour,
		MaxValue:     24 * time.Hour,
		DefaultValue: 12 * time.Hour,
	}.resolve(ctx)
}

type BoundedDuration struct {
	EnvName       string
	MinValue 	  time.Duration
	MaxValue      time.Duration
	DefaultValue  time.Duration
	ResolvedValue time.Duration
}

func (d BoundedDuration) resolve(ctx context.Context) BoundedDuration {
	_, log := logd.NewFromContext(ctx, "k8senv")

	rawDuration := os.Getenv(d.EnvName)
	if rawDuration == "" {
		log.Debug("no custom env set, using default", "env", d.EnvName, "default", d.DefaultValue)

		d.ResolvedValue = d.DefaultValue
		
		return d
	}

	duration, err := time.ParseDuration(rawDuration)
	if err != nil {
		log.Error(err, "invalid duration value, using default", "env", d.EnvName, "value", rawDuration, "default", d.DefaultValue)

		d.ResolvedValue = d.DefaultValue

		return d
	}

	if duration < d.MinValue || duration > d.MaxValue {
		log.Info("duration not in allowed range, using default", "env", d.EnvName, "value", duration, "min", d.MinValue, "max", d.MaxValue, "default", d.DefaultValue)

		d.ResolvedValue = d.DefaultValue

		return d
	}

	log.Info("using custom duration", "env", d.EnvName, "value", duration)
	d.ResolvedValue = duration

	return d
}
```

the details of the implementation probably would need love but the general idea is:

I want all these const values to be grouped in a struct, so I can read them easily with my 🐒  brain

So stuff like: (I even dropped the "Get" here as it is just too JAVA for me)
```go
func WebhookCertsRootDuration(ctx context.Context) BoundedDuration {
	return BoundedDuration{
		EnvName:     "DT_WEBHOOK_CERTS_ROOT_DURATION",
		MinValue:     7 * 24 * time.Hour,
		MaxValue:     10 * 365 * 24 * time.Hour,
		DefaultValue: 365 * 24 * time.Hour,
	}.resolve(ctx)
}
```

this is (IMO) far more readable than:
```go
    WebhookCertsRootDurationEnvVar  = "DT_WEBHOOK_CERTS_ROOT_DURATION"
	defaultWebhookCertsRootDuration = 365 * 24 * time.Hour
	minWebhookCertsRootDuration     = 7 * 24 * time.Hour
	maxWebhookCertsRootDuration     = 10 * 365 * 24 * time.Hour
//....

func parseDuration (ctx context.Context, envVar string, defaultValue, minValue, maxValue time.Duration) time.Duration {...}
```

I intentionally made all elements public, so for whatever testing/logging you want to do, they are available, and grouped.

andriisoldatenko: sounds like tech debt small story :)

aorcholski: nice idea, will create a ticket

pkg/util/kubernetes/fields/k8senv/env.go
aorcholski: we shouldn't allow to set it to 0s so min value is needed

pkg/util/kubernetes/fields/k8senv/env.go
aorcholski: warning added to the comment.

pkg/util/kubernetes/fields/k8senv/env.go
aorcholski: the ticket [JIRA](https://dt-rnd.atlassian.net/browse/ICP-8206)

pkg/util/kubernetes/fields/k8senv/env.go
aorcholski: right, my bad, sorry

config/helm/chart/default/templates/_helpers.tpl
```diff
@@ -194,3 +194,17 @@ topologySpreadConstraints:
...
+{{- define "dynatrace-operator.client-connection-timeout-env" -}}
+  {{- with .Values.operator.clientConnectionTimeout }}
+- name: DT_CLIENT_CONNECTION_TIMEOUT
+  value: {{ . | quote }}
+  {{- end }}
+{{- end -}}
+
+{{- define "dynatrace-csidriver.client-connection-timeout-env" -}}
+  {{- with .Values.csidriver.clientConnectionTimeout }}
+- name: DT_CLIENT_CONNECTION_TIMEOUT
```
andriisoldatenko, line 207: is it intentiolly same env var?
```
DT_CLIENT_CONNECTION_TIMEOUT
```

pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -33,6 +33,13 @@ const (
 	minDTClientCacheCleanInterval     = 5 * time.Minute
 	maxDTClientCacheCleanInterval     = 100 * time.Hour
 
+	DTClientConnectionTimeoutEnvVar = "DT_CLIENT_CONNECTION_TIMEOUT"
```
andriisoldatenko, line 36: if i read helm code correctly, we reuse same envVar name for 2 CSI/operator time out. is it intentionally?

0sewa0: yes, it influences the same thing, there is no reason to add another one (opreator/csi are separate containers)

avorima (approved):

andriisoldatenko (approved):

PR #7043, opened 2026-07-27
add support of new db extension component name for image fetching

andriisoldatenko:
## Description

Resolves [ICP-7908](https://dt-rnd.atlassian.net/browse/ICP-7908)

An inconsistency in naming was identified for `dynatrace-sql-extension-executor` component so this PR resolves this problem by introducing correct name, but keep old name for backward compatibility

> [!NOTE]
> New name takes presedense: `sql-extension-executor` and the we fallback to old name

## How can this be tested?

unit-tests or manual test if you tenant supports both:

1. Apply DK
```
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
  annotations:
    feature.dynatrace.com/use-public-registry: "true"
spec:
  customPullSecret: devregistry
  apiUrl: https://*.dev.dynatracelabs.com/api
  activeGate:
    capabilities:
      - kubernetes-monitoring
  extensions:
    databases:
      - id: testtest
```

2. Make sure db ext pods are up and running
```
k get pods
NAME                                              READY   STATUS    RESTARTS   AGE
dynakube-activegate-0                             1/1     Running   0          32m
dynakube-extension-controller-0                   1/1     Running   0          32m
dynakube-sql-ext-exec-testtest-5d57d5b6c4-lk644   1/1     Running   0          32m
dynatrace-operator-64f65dbbbf-v8htk               1/1     Running   0          60m
dynatrace-webhook-7cc5ffdccf-n6b74                1/1     Running   0          60m
dynatrace-webhook-7cc5ffdccf-x2467                1/1     Running   0          60m
```

> [!NOTE]
> My tenant supports both, so no fallabck. 
> it's not visiable anywhere only for component query, because the imageURI is the same:
```
    {
      "type": "sql-extension-executor",
      "imageUri": "****.dkr.ecr.us-east-1.amazonaws.com/dynatrace/dynatrace-sql-extension-executor:1.345.0.20260721-174746@sha256:155921e677efdc7512f9243ac7d6c0a7eeb00c3e8e6f059ba9c462b7d5008121"
    },
    {
      "type": "dynatrace-sql-extension-executor",
      "imageUri": "****.dkr.ecr.us-east-1.amazonaws.com/dynatrace/dynatrace-sql-extension-executor:1.345.0.20260721-174746@sha256:155921e677efdc7512f9243ac7d6c0a7eeb00c3e8e6f059ba9c462b7d5008121"
    },
``` 

[ICP-7908]: https://dt-rnd.atlassian.net/browse/ICP-7908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controllers/dynakube/extension/databases/reconciler.go
+++ b/pkg/controllers/dynakube/extension/databases/reconciler.go
@@ -125,7 +125,7 @@ func resolveImageURI(ctx context.Context, imageClient dtimage.Client, dk *dynaku
 
 	// fallback to old image name for backward compatibility if we can't resolve the image URI by new component name.
 	// TODO: remove this fallback in a future release
-	if apiErr := new(core.HTTPError); err != nil && !errors.As(err, &apiErr) { //nolint:revive // we want to check for a specific error type, not the whole error chain
+	if apiErr := new(core.HTTPError); err != nil && !errors.As(err, &apiErr) {
 		imageURI, err = registry.ResolveImage(ctx, imageClient, dk.PublicRegistryOverride(), dtimage.DBExecutorOldName)
 		if err != nil {
 			return "", err
--- a/pkg/controllers/dynakube/extension/databases/reconciler.go
+++ b/pkg/controllers/dynakube/extension/databases/reconciler.go
@@ -5,8 +5,10 @@ package databases
 
 import (
 	"context"
+	"errors"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/core"
 	dtimage "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/image"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/registry"
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
@@ -62,16 +64,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, imageClient dtimage.Client, 
 	} else {
 		var err error
 
-		imageURI, err = registry.ResolveImage(ctx, imageClient, dk.PublicRegistryOverride(), dtimage.DBExecutor)
+		imageURI, err = resolveImageURI(ctx, imageClient, dk)
 		if err != nil {
-			// fallback to old image name for backward compatibility
-			// TODO: remove this fallback in a future release
-			imageURI, err = registry.ResolveImage(ctx, imageClient, dk.PublicRegistryOverride(), dtimage.DBExecutorOldName)
-			if err != nil {
-				return err
-			}
-
-			log.Debug("using old image name for backward compatibility", "imageURI", imageURI)
+			return err
 		}
 	}
 
@@ -122,4 +117,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, imageClient dtimage.Client, 
 	k8sconditions.SetDeploymentsApplied(dk, conditionType, expectedDeploymentNames)
 
 	return nil
+}
+
+func resolveImageURI(ctx context.Context, imageClient dtimage.Client, dk *dynakube.DynaKube) (string, error) {
+	ctx, log := logd.NewFromContext(ctx, "resolveImageURI")
+	imageURI, err := registry.ResolveImage(ctx, imageClient, dk.PublicRegistryOverride(), dtimage.DBExecutor)
+
+	// fallback to old image name for backward compatibility if we can't resolve the image URI by new component name.
+	// TODO: remove this fallback in a future release
+	if apiErr := new(core.HTTPError); err != nil && !errors.As(err, &apiErr) { //nolint:revive // we want to check for a specific error type, not the whole error chain
+		imageURI, err = registry.ResolveImage(ctx, imageClient, dk.PublicRegistryOverride(), dtimage.DBExecutorOldName)
+		if err != nil {
+			return "", err
+		}
+
+		log.Debug("using old image name for backward compatibility", "imageURI", imageURI)
+	}
+
+	return imageURI, err
 }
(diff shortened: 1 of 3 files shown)


pkg/controllers/dynakube/extension/databases/reconciler.go
```diff
@@ -64,7 +64,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, imageClient dtimage.Client,
 
 		imageURI, err = registry.ResolveImage(ctx, imageClient, dk.PublicRegistryOverride(), dtimage.DBExecutor)
 		if err != nil {
-			return err
+			// fallback to old image name for backward compatibility
```
avorima: this path should only be taken if it's not a core.HTTPError we get the same error twice otherwise

andriisoldatenko: Agree, 👍 will fix

andriisoldatenko: 038af23fb fixed

pkg/controllers/dynakube/extension/databases/reconciler.go
```diff
@@ -116,3 +118,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, imageClient dtimage.Client,
...
 	return nil
 }
+
+func resolveImageURI(ctx context.Context, imageClient dtimage.Client, dk *dynakube.DynaKube) (string, error) {
+	ctx, log := logd.NewFromContext(ctx, "resolveImageURI")
+	imageURI, err := registry.ResolveImage(ctx, imageClient, dk.PublicRegistryOverride(), dtimage.DBExecutor)
+
+	// fallback to old image name for backward compatibility if we can't resolve the image URI by new component name.
+	// TODO: remove this fallback in a future release
+	if apiErr := new(core.HTTPError); err != nil && !errors.As(err, &apiErr) { //nolint:revive // we want to check for a specific error type, not the whole error chain
```
avorima: what linter error is this ignoring? nothing shows up when i remove this

andriisoldatenko: nothing leftover, removed in 9d262a155

pkg/controllers/dynakube/extension/databases/reconciler.go
```diff
@@ -116,3 +118,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, imageClient dtimage.Client,
 
 	return nil
 }
+
+func resolveImageURI(ctx context.Context, imageClient dtimage.Client, dk *dynakube.DynaKube) (string, error) {
+	ctx, log := logd.NewFromContext(ctx, "resolveImageURI")
+	imageURI, err := registry.ResolveImage(ctx, imageClient, dk.PublicRegistryOverride(), dtimage.DBExecutor)
+
+	// fallback to old image name for backward compatibility if we can't resolve the image URI by new component name.
```
waodim, line 126: The comment implies any failure triggers the fallback, but http errors are actually propagated immediately without retrying. The condition below only falls back on non-http errors if I understand correctly. Maybe let us go with a clearer version?

```
// only fall back on non-HTTP errors (component not in API response).
// HTTP errors (4xx, 5xx) mean the request itself failed
```

pkg/controllers/dynakube/extension/databases/reconciler_test.go
```diff
@@ -365,6 +366,46 @@ func TestPublicRegistryImage(t *testing.T) {
 		require.Equal(t, expectedImageURI, deployment.Spec.Template.Spec.Containers[0].Image)
 	})
 
+	t.Run("imageURI from public registry fallback to old name", func(t *testing.T) {
+		expectedImageURI := "my-public-registry:5000/my-test-repo:my-test-tag"
+		getReconciledDeployment := func(t *testing.T) *appsv1.Deployment {
```
waodim, line 371: This helper is defined and called exactly once in the same test. I think inlining it would keep this test structurally consistent with the "http error" test below it, which doesn't use a wrapper. IMO it would make both tests easier to read side by side

pkg/controllers/dynakube/extension/databases/reconciler_test.go
```diff
@@ -365,6 +366,46 @@ func TestPublicRegistryImage(t *testing.T) {
...
+			t.Helper()
+
+			dk := getTestDynakube()
+			dk.Annotations = map[string]string{exp.UsePublicRegistryKey: "true"}
+			dk.Spec.Templates = dynakube.TemplatesSpec{}
+			clt := fakeClient()
+
+			imageClient := imageclientmock.NewClient(t)
+			boom := errors.New("component not found")
+			imageClient.EXPECT().GetComponentLatestInfo(mock.MatchedBy(func(context.Context) bool { return true }), dtimage.DBExecutor, "").Return(nil, boom)
```
waodim, line 381: The "http error" test below uses `.Once()` on its mock. Worth being consistent and adding `.Once()` to both mocks here too, no?

andriisoldatenko: no because we do it twice, so ONce i've added to validate we don't run it twice.

pkg/controllers/dynakube/extension/databases/reconciler.go
andriisoldatenko: it's overcomplicate it, we will remove code, we agreed with @avorima that checking any error doesn't make anysense, bc we want to fallback only when component not found but http request was successful, so we exclude all httpErrors from list.
I don't think we want to have such a detailed comment here.

pkg/controllers/dynakube/extension/databases/reconciler_test.go
andriisoldatenko: you can check full test file, i'm following same pattern. removeing inline make sense only for small eeror tests, otherwise we need refactor all tests in suite.

waodim (approved):
lgtm

avorima (approved):

PR #7047, opened 2026-07-27
Delete replicated secrets on namespace deselection

Incredible-O:
## Description
Currently, replicated secrets (like `dynatrace-bootstrapper-config` or `dynatrace-otlp-exporter-config`, for example) aren't always deleted when a namespace is removed from a DynaKube's namespace selector. Fix this by reusing clean up logic from `UnmapFromDynaKube`.

Addresses ICP-7549.

## How can this be tested?
Unit tests

### DIFF
--- a/pkg/injection/namespace/mapper/dynakubes.go
+++ b/pkg/injection/namespace/mapper/dynakubes.go
@@ -31,6 +31,8 @@ type DynakubeMapper struct {
 	matchedOANamespaces   []string
 	matchedMENamespaces   []string
 	matchedOTLPNamespaces []string
+
+	deselectedNamespaces []string
 }
 
 func NewDynakubeMapper(ctx context.Context, clt client.Client, apiReader client.Reader, operatorNs string, dk *dynakube.DynaKube) DynakubeMapper {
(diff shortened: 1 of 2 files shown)


svfoxat:
while testing, when a `DynaKube` updated with a `namespaceSelector` that deselects a previously-injected namespace, the validating webhook panics with a nil pointer dereference.

my steps:
- `kubectl create namespace test-ns`
- `kubectl label namespace test-ns env=test`
- apply a `DynaKube` that selects namespaces with label `env=test`
- wait until secrets have been replicated
- apply patched `DynaKube` with `env=prod` label selector
- get ```Error from server: admission webhook "v1beta6.dynakube.webhook.dynatrace.com" denied the request:
panic: runtime error: invalid memory address or nil pointer dereference [recovered]```

pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -167,3 +154,37 @@ func (dm *DynakubeMapper) mapFromDynakube(nsList *corev1.NamespaceList, dkList *
...
+		dm.matchedOTLPNamespaces = append(dm.matchedOTLPNamespaces, namespace.Name)
+	}
+
+	updated, err := updateNamespace(dm.ctx, namespace, dkList)
+	if err != nil {
+		return false, err
+	}
+
+	if previouslyInjected && !result.IsAny() {
+		if err := dm.deleteReplicatedSecrets(namespace.Name); err != nil {
```
avorima: this must not be called during validation

Incredible-O: Addressed in a0f94eaea (cc @svfoxat)

avorima (approved):

svfoxat (approved):

PR #7048, opened 2026-07-27
Prefix kubemon operand gate with experimental

jdanieck:
## Description

Renames the temporary kubemon operand gate to carry an experimental prefix, making its maturity explicit in both the env var and the Helm value.

Summary of the change:

- Env var: `KUBEMON_ENABLE_OPERAND` → `EXPERIMENTAL_ENABLE_KUBEMON_OPERAND` 
- Helm value: flat `operator.kubemonEnableOperand` → nested `operator.experimental.enableKubemonOperand`
- Templates use nil-safe access (`| default dict`) so rendering doesn't fail when `operator.experimental` is unset
- Updated operator/webhook deployment templates, Helm unit tests (incl. null-experimental regression cases), Go tests, and the kubemon README

## How can this be tested?

Enable the operand via Helm and confirm the env var lands in the operator and webhook pods:

```sh
  --reuse-values \
  --set operator.experimental.enableKubemonOperand=true
```


config/helm/chart/default/values.yaml
```diff
@@ -57,7 +57,8 @@ operator:
   startupProbe: {} # configurable: periodSeconds, timeoutSeconds, failureThreshold
   requeueAfter: "" # requeue period for the controller, defaults to 15m
   clientCacheCleanupInterval: "" # defined in the Golang time.Duration format, like "30m" == 30 minutes. Defaults to 1h
-  kubemonEnableOperand: false # temporary gate; set to true to enable kubemon in dev/test. Remove when kubemon is complete.
+  experimental:
```
jdanieck, line 60: @StefanHauth @baichinger shall we use `.experimental` flag also to gate Scalable Prometheus changes while they are under development ? What do you think ?

config/helm/chart/default/templates/Common/operator/deployment-operator.yaml
```diff
@@ -117,8 +117,8 @@ spec:
             - name: DT_CLIENT_CACHE_CLEAN_INTERVAL
               value: "{{ .Values.operator.clientCacheCleanupInterval }}"
             {{- end }}
-            {{- if .Values.operator.kubemonEnableOperand }}
-            - name: KUBEMON_ENABLE_OPERAND
+            {{- if (.Values.operator.experimental | default dict).enableKubemonOperand }}
```
avorima: do we ever expect users to set experimental to `nil`? and if so, does that really not warrant an error?

config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
```diff
@@ -77,8 +77,8 @@ spec:
             - name: DT_EXTRACT_CODEMODULES_IMAGE_LINKS
               value: "true"
             {{- end }}
-            {{- if .Values.operator.kubemonEnableOperand }}
-            - name: KUBEMON_ENABLE_OPERAND
+            {{- if (.Values.operator.experimental | default dict).enableKubemonOperand }}
```
0sewa0: also works by only doing:
```suggestion
            {{- if (.Values.operator.experimental).enableKubemonOperand }}
```

config/helm/chart/default/templates/Common/operator/deployment-operator.yaml
jdanieck: Generally this should be rare, but technically possible, so I wanted to guard against it . I can imagine someone want to disable all experimental features by `--set operator.experimental=nil` and this shouldn't produce error.

When I do just `{{if .Values.operator.experimental.enableKubemonOperand }}`
```ssh
FAIL  test deployment of webhook       config/helm/chart/default/tests/Common/webhook/deployment-webhook_test.yaml
        - should render without error and omit EXPERIMENTAL_ENABLE_KUBEMON_OPERAND when operator.experimental is null
                Error: template: dynatrace-operator/templates/Common/webhook/deployment-webhook.yaml:80:26: executing "dynatrace-operator/templates/Common/webhook/deployment-webhook.yaml" at <.Values.operator.experimental.enableKubemonOperand>: nil pointer evaluating interface {}.enableKubemonOperand
```

config/helm/chart/default/templates/Common/webhook/deployment-webhook.yaml
jdanieck: done in https://github.com/Dynatrace/dynatrace-operator/pull/7048/changes/0479209887904b376cfca15cce5434ac8d2abc24

0sewa0 (approved):

config/helm/chart/default/templates/Common/operator/deployment-operator.yaml
avorima: the features that we toggle here should get their own sections once they are enabled by default, so this should only be used for enabling things. with this use case in mind, setting this to nil does not seem like something we should support. otherwise we should guard against disabling the csi with `--set csidriver=nil` also for example

0sewa0: AFAIK, the whole reason for checking `nil`, is due to `reuse-values` flag in helm and not that people would manually `--set` something to `nil`. Or that is the historical reason we have this in other places.

However, AFAIK we don't officially support `reuse-values` due to this "it may break things" reason.

I generally agree, that I don't want to support stuff like `--set section.subsection=nil`, if the user wants to "reset" values.

If we don't support `reuse-values` we should cleanup our `nil` checks, and this new one is not needed.
If we support `reuse-values` we should still cleanup our old `nil` checks, and this new one is also needed.

pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -162,7 +162,7 @@ func GetDTExtractCodeModulesImageLinks(ctx context.Context) bool {
 }
 
 func IsKubemonOperandEnabled() bool {
```
waodim, line 164: Food for thought: Should we add the prefix here as well?

config/helm/chart/default/templates/Common/operator/deployment-operator.yaml
jdanieck: My example of `--set experimental=null` wasn't a great one, and I agree we don't need to support that. But @0sewa0 point about `--reuse-values` is what shifts me toward keeping the guard — that's the real scenario here. Dropping it feels wrong when it's such a cheap fix; saying its not supported feels more like a workaround.

0sewa0: IMO keep the guard for now, so we can move this along

and start a separate conversation about ALL our Helm `nil` checks if they are necessary.

if it is not necessary here, it is probably not necessary anywhere

andriisoldatenko: my 5 cents i agree with @avorima that when we remove kubemon from experimental, we still can keep:
```
experimental: {}
```
but not nil. so setting `--set operator.experimental=nil` is a bit shooting our own leg.

Disable all experimental features should be default behaviour IMO (and only expicitly enabling one by one make sense to me).

> and start a separate conversation about ALL our Helm nil checks if they are necessary.

+1 for write a story or discuss independently from PR.

> --reuse-values

yes it's special case - which again we need to decided seperatly to support or not.

andriisoldatenko: @avorima are you good with fix? It seems @jdanieck already fixed as we expect for now.

0sewa0: > which again we need to decided seperatly to support or not.


FYI, in our docs we always reference to use `--reset-then-reuse-values` which is the safe version of `--reuse-values`

jdanieck: Thank you all for the context! I will keep as is in this PR and follow-up on how to unify across the codebase separately.

andriisoldatenko (approved):

pkg/util/kubernetes/fields/k8senv/env.go
jdanieck: @waodim I thought about it and chosen to do changes only to externally visible interface (helm chart, env var). This PR is open too long already so I'll keep as is :)

waodim (approved):

PR #7069, opened 2026-07-28
fix: restore OneAgent downgrade detection for public registry image source

knQzx:
## Description

`CheckForDowngrade` only fills `previousVersion` when the target source is `TenantRegistryVersionSource`. for public-registry setups (the `public-registry` feature flag) the source is `PublicRegistryVersionSource`, so `previousVersion` stays empty, `isDowngrade` always compares against an empty string and returns false - the downgrade guard is silently off

this is a regression from #6227 (commit c78fc15de), which dropped the `PublicRegistryVersionSource` branch that read the version tag from the image id. this restores it

impact: with a public-registry OneAgent image, lowering the version rolls the OLDER image across the whole DaemonSet instead of being blocked as a downgrade

## How can this be tested?

enable the public registry feature flag on a DynaKube so the OneAgent version source is `PublicRegistryVersionSource`, let it settle on an image at some version, then reconcile with a lower target version - on main the downgrade is accepted, with this change the downgrade condition is set and the rollback is blocked

the added `TestCheckForDowngrade` cases cover the public-registry downgrade and non-downgrade paths without needing a cluster

### DIFF
--- a/pkg/controllers/dynakube/version/oneagent.go
+++ b/pkg/controllers/dynakube/version/oneagent.go
@@ -130,15 +130,8 @@ func (updater *oneAgentUpdater) CheckForDowngrade(ctx context.Context, latestVer
 	var err error
 
 	switch updater.Target().Source {
-	case status.TenantRegistryVersionSource:
+	case status.TenantRegistryVersionSource, status.PublicRegistryVersionSource:
 		previousVersion = updater.Target().Version //nolint:staticcheck
-	case status.PublicRegistryVersionSource:
-		previousVersion, err = getTagFromImageID(imageID)
-		if err != nil {
-			setVerificationFailedReasonCondition(updater.dk.Conditions(), oaConditionType)
-
-			return false, err
-		}
 	}
 
 	downgrade, err := isDowngrade(ctx, updater.Name(), previousVersion, latestVersion)
--- a/pkg/controllers/dynakube/version/oneagent.go
+++ b/pkg/controllers/dynakube/version/oneagent.go
@@ -129,8 +129,16 @@ func (updater *oneAgentUpdater) CheckForDowngrade(ctx context.Context, latestVer
 
 	var err error
 
-	if updater.Target().Source == status.TenantRegistryVersionSource {
+	switch updater.Target().Source {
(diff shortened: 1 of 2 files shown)


pkg/controllers/dynakube/version/oneagent.go
```diff
@@ -129,8 +129,16 @@ func (updater *oneAgentUpdater) CheckForDowngrade(ctx context.Context, latestVer
...
+	switch updater.Target().Source {
+	case status.TenantRegistryVersionSource:
 		previousVersion = updater.Target().Version //nolint:staticcheck
+	case status.PublicRegistryVersionSource:
+		previousVersion, err = getTagFromImageID(imageID)
+		if err != nil {
+			setVerificationFailedReasonCondition(updater.dk.Conditions(), oaConditionType)
+
+			return false, err
+		}
```
0sewa0: could also just use the `updater.Target().Version`

as we set the tag as the version:
```go
func setImageFromImageInfo(ctx context.Context,
	target *status.VersionStatus,
	imageInfo *image.Info,
) {
	log := logd.FromContext(ctx)
	oldImageID := target.ImageID

	target.Version = imageInfo.Tag // HERE
	target.ImageID = imageInfo.URI

	log.Info("updated image version info",
		"image", imageInfo,
		"oldImageID", oldImageID,
		"newImageID", target.ImageID)
}
```

knQzx:
good call, that is simpler - switched to it

on the public registry path `CheckForDowngrade` runs before `setImageFromImageInfo`, so `Target().Version` still holds the tag from the previous reconcile, and since `ValidateStatus` rejects an empty version no successful public registry pass can leave it unset. so both sources share the branch now and `getTagFromImageID` is not needed here:

```go
switch updater.Target().Source {
case status.TenantRegistryVersionSource, status.PublicRegistryVersionSource:
    previousVersion = updater.Target().Version
}
```

updated the two new test cases to set `Version` next to `ImageID`, matching the tenant registry cases and how the two fields are actually written together

for context on how this got lost: the branch was dropped in #6227 when the whole public registry path was removed, and #6553 brought the feature back without it. `getTagFromImageID` has had no production caller since, which is what put me onto it

PR #7070, opened 2026-07-28
fix: prevent file descriptor leak in CSI app volume publisher

knQzx:
## Description

`preparePodInfoUpperDir` in the CSI app volume publisher opens the destination file with `os.OpenFile` and writes to it, but never closes the handle. every volume mount leaks one file descriptor on the node, so a workload that mounts and remounts (or crash-loops) slowly exhausts the CSI driver process fds until mounts start failing with "too many open files"

the fix replaces the manual open/write with `os.WriteFile`, which closes the handle on return

## How can this be tested?

hard to see directly at runtime without watching `/proc/<csi-driver-pid>/fd` count while pods mount app volumes repeatedly - it grows without bound on main and stays flat with this change

at the unit level the added `TestPreparePodInfoUpperDir` writes the pod info file and asserts the content, and the change is a straight swap to a self-closing write so no fd is left open

### DIFF
--- a/pkg/controllers/csi/server/volumes/app/publisher.go
+++ b/pkg/controllers/csi/server/volumes/app/publisher.go
@@ -37,6 +37,8 @@ type Publisher struct {
 
 const (
 	retryLimitReachedMsg = "reached max mount attempts for pod, attaching dummy volume, monitoring disabled"
+
+	podInfoFilePerm = 0666
 )
 
 func (pub *Publisher) PublishVolume(ctx context.Context, volumeCfg *csivolumes.VolumeConfig) (*csi.NodePublishVolumeResponse, error) {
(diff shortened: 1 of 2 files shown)


knQzx:
fixed the two linter findings, both were on lines this PR added:

- `gosec` G306 flagged `os.ModePerm` on the new `os.WriteFile` call. the old `os.OpenFile` used the same permission but G306 only looks at `WriteFile`, so the rule only surfaced now. switched to a named `podInfoFilePerm = 0666`, which drops the execute bits that mean nothing on a data file and matches `permAll` in `cmd/webhook/certificates/watcher.go`
- `intrange` wanted the loop as `for range 3`

while checking, the test turned out not to be testing much - it passed on the leaky code as well, since a handful of leaked descriptors are invisible in a short run. rewrote it to count open descriptors around 50 mounts, so it now fails on main with `open file descriptors leaked across mounts` and passes with the fix

also rebased on main, and `golangci-lint` (v2.12.2, as pinned in CI) is clean

PR #7072, opened 2026-07-28
Make init user/group configurable

0sewa0:
## Description
https://dt-rnd.atlassian.net/browse/ICP-7892

Adds 2 new pod level annotation, that can overrule the current "auto setting" of the init-container's `runAs*`.

```
init-container.dynatrace.com/securityContext.runAsUser: "1002"
init-container.dynatrace.com/securityContext.runAsGroup: "1002"
```

Introducing this config option for the rare usecase where the auto-setting does not comply with the user's setup.

Other option that was considered: "Never let the init-container to be root, even if the injected containers are root"
- Could be breaking change, why we decided to not do it.

## How can this be tested?

Set the mentioned annotations on a sample pod, and verify that it is respected.

### DIFF
--- a/pkg/webhook/mutation/pod/handler/injection/init.go
+++ b/pkg/webhook/mutation/pod/handler/injection/init.go
@@ -131,11 +131,11 @@ func combineSecurityContexts(ctx context.Context, baseSecurityCtx corev1.Securit
 		baseSecurityCtx.RunAsGroup = containerSecurityCtx.RunAsGroup
 	}
 
-	if user := parseAnnotationAsUint32(ctx, pod.Annotations, dtwebhook.AnnotationInitContainerRunAsUser); user != nil {
+	if user := getValidatedID(ctx, pod.Annotations, dtwebhook.AnnotationInitContainerRunAsUser); user != nil {
 		baseSecurityCtx.RunAsUser = user
 	}
 
-	if group := parseAnnotationAsUint32(ctx, pod.Annotations, dtwebhook.AnnotationInitContainerRunAsGroup); group != nil {
+	if group := getValidatedID(ctx, pod.Annotations, dtwebhook.AnnotationInitContainerRunAsGroup); group != nil {
 		baseSecurityCtx.RunAsGroup = group
 	}
 
@@ -144,16 +144,16 @@ func combineSecurityContexts(ctx context.Context, baseSecurityCtx corev1.Securit
 	return &baseSecurityCtx
 }
 
-// parseAnnotationAsUint32 will parse the provided annotation as a uint32,
-// but will return an int64 pointer for compatibility reasons with the corev1.SecurityContext type.
-// Will only log parsing errors, as we do not want to block due to incorrect user annotation.
-func parseAnnotationAsUint32(ctx context.Context, annotations map[string]string, key string) *int64 {
+// getValidatedID will parse the provided annotation as valid user/group ID between 0 and math.MaxInt32.
+// Returns *int64 for compatibility with corev1.SecurityContext
+func getValidatedID(ctx context.Context, annotations map[string]string, key string) *int64 {
 	val, ok := annotations[key]
 	if !ok {
 		return nil
 	}
 
-	parsed, err := strconv.ParseUint(val, 10, 32)
+	// uint31 == positive int32 range
+	parsed, err := strconv.ParseUint(val, 10, 31)
 	if err != nil {
 		logd.FromContext(ctx).Error(err, "failed to parse annotation value, must be a uint32", "key", key, "value", val)
 
--- a/pkg/webhook/mutation/pod/handler/injection/init.go
+++ b/pkg/webhook/mutation/pod/handler/injection/init.go
@@ -131,11 +131,11 @@ func combineSecurityContexts(ctx context.Context, baseSecurityCtx corev1.Securit
 		baseSecurityCtx.RunAsGroup = containerSecurityCtx.RunAsGroup
 	}
 
-	if user := getValidatedID(ctx, pod.Annotations, dtwebhook.AnnotationInitContainerRunAsUser); user != nil {
+	if user := getValidatedID(ctx, pod.Annotations[dtwebhook.AnnotationInitContainerRunAsUser]); user != nil {
 		baseSecurityCtx.RunAsUser = user
 	}
 
-	if group := getValidatedID(ctx, pod.Annotations, dtwebhook.AnnotationInitContainerRunAsGroup); group != nil {
+	if group := getValidatedID(ctx, pod.Annotations[dtwebhook.AnnotationInitContainerRunAsGroup]); group != nil {
 		baseSecurityCtx.RunAsGroup = group
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/webhook/mutation/pod/handler/injection/init.go
```diff
@@ -129,11 +131,35 @@ func combineSecurityContexts(baseSecurityCtx corev1.SecurityContext, pod corev1.
...
 	return &baseSecurityCtx
 }
 
+func parseAnnotationAsInt64(annotations map[string]string, key string) *int64 {
+	val, ok := annotations[key]
+	if !ok {
+		return nil
+	}
+
+	parsed, err := strconv.ParseInt(val, 10, 64)
```
andriisoldatenko: >The Linux Kernel 2.6 and above supports unsigned 32-bit integers as UIDs and GIDs. This means the maximum UID should be 4294967294

andriisoldatenko: i've quickly tried:
```
The Pod "security-context-demo" is invalid: spec.securityContext.runAsUser: Invalid value: 4294967295: must be between 0 and 2147483647, inclusi
ve
```

andriisoldatenko: so we need something like:
```
const maxUID = int64(2147483647) // math.MaxInt32

func validateRunAsUser(uid int64) error {
    if uid < 0 || uid > maxUID {
        return fmt.Errorf("runAsUser must be between 0 and %d, got %d", maxUID, uid)
    }
    return nil
}
```
for groups as well:
```
* spec.securityContext.runAsUser: Invalid value: 4294967295: must be between 0 and 2147483647, inclusive
* spec.securityContext.runAsGroup: Invalid value: 4294967295: must be between 0 and 2147483647, inclusive
```

pkg/webhook/mutation/pod/mutator/config.go
```diff
@@ -33,6 +33,12 @@ const (
 	// InstallContainerName is the name used for the install container
 	InstallContainerName = "dynatrace-operator"
 
+	AnnotationInitContainerPrefix = "init-container.dynatrace.com/"
+	// AnnotationInitContainerRunAsUser can be set on a Pod to override the RunAsUser in the init container's security context.
+	AnnotationInitContainerRunAsUser = AnnotationInitContainerPrefix + "securityContext.runAsUser"
```
waodim, line 38: Is the naming set in stone? wdyt about kebap case? `security-context.run-as-user`? I mean I would understand if you think this is overkill 😆

pkg/webhook/mutation/pod/handler/injection/init.go
andriisoldatenko:
```
"k8s.io/apimachinery/pkg/util/validation"
validation.IsValidUserID
```

validation.IsValidUserID is not a good idea, because it returns slice of strings with error messages as strings, so not fit in this purpose, also has 2 variations for same code.

pkg/webhook/mutation/pod/handler/injection/init.go
```diff
@@ -129,11 +131,35 @@ func combineSecurityContexts(baseSecurityCtx corev1.SecurityContext, pod corev1.
...
+	if err != nil {
+		logd.Get().Error(err, "failed to parse pod annotation as int64, ignoring", "annotation", key, "value", val)
```
waodim: Is this the right call here? To call `Error()` but just recovering gracefully and ignoring the bad value? But I guess calling `Info()` would be wrong as in fact there is something that has gone wrong.

pkg/webhook/mutation/pod/mutator/config.go
0sewa0: the name came from product 😅 

the idea is to "be familiar" to the actual fields

pkg/webhook/mutation/pod/handler/injection/init.go
0sewa0: the idea was that I do not want to return the err, but I want to log it, so `.Error(...)` is the way

0sewa0: but I am not using the context logger, will fix that

waodim: fair enough, thanks

pkg/webhook/mutation/pod/handler/injection/init.go
0sewa0: I will try to use it 😅

avorima: use ParseUint

andriisoldatenko:
```
if s, err := strconv.ParseUint(v, 10, 32); err == nil {
		fmt.Printf("%T, %v\n", s, s)
	}
```

0sewa0: will do, but the log message won't be as nice 😭

andriisoldatenko: you can still reconstruct it "must be between 0 and %d, inclusive", math.MaxUint32

0sewa0: just said "must be positive int32", anyone using this annotation should be knowledgeable enough to understand this

pkg/webhook/mutation/pod/handler/injection/init.go
```diff
@@ -129,11 +131,38 @@ func combineSecurityContexts(baseSecurityCtx corev1.SecurityContext, pod corev1.
...
+func getValidatedID(ctx context.Context, annotations map[string]string, key string) *int64 {
+	val, ok := annotations[key]
+	if !ok {
+		return nil
+	}
+
+	// uint31 == positive int32 range
+	parsed, err := strconv.ParseUint(val, 10, 31)
+	if err != nil {
+		logd.FromContext(ctx).Error(err, "failed to parse annotation value, must be a uint32", "key", key, "value", val)
```
waodim:
```suggestion
		logd.FromContext(ctx).Error(err, "failed to parse annotation value, must be a positive int32", "key", key, "value", val)
```

pkg/webhook/mutation/pod/handler/injection/init.go
```diff
@@ -129,11 +131,38 @@ func combineSecurityContexts(baseSecurityCtx corev1.SecurityContext, pod corev1.
...
+	}
+
 	baseSecurityCtx.RunAsNonRoot = new(isNonRoot(&baseSecurityCtx))
 
 	return &baseSecurityCtx
 }
 
+// getValidatedID will parse the provided annotation as valid user/group ID between 0 and math.MaxInt32.
+// Returns *int64 for compatibility with corev1.SecurityContext
+func getValidatedID(ctx context.Context, annotations map[string]string, key string) *int64 {
```
avorima: can we make the API a bit simpler here? the map lookup could be done by the caller so that this just takes a string. this changes the validation behavior to make the empty string ignored instead of an error, which i think this fine in this case since it would just be logged.

andriisoldatenko (dismissed):

avorima (approved):

PR #7074, opened 2026-07-28
Add image auto-update to kubemon sts

0sewa0:
## Description
https://dt-rnd.atlassian.net/browse/ICP-7295

The standalone kubemon statefulset no longer requires a custom image to be set.

Will auto resolve the image like the normal ActiveGate (without the baggage)

~Open question:~
I didn't use the `VersionStatus` that is already in the kubemon' status(but unused). ~Should I? -> discuss in https://github.com/Dynatrace/dynatrace-operator/pull/7074#discussion_r3665448051~
   - Outcome: Fully resolved `image` in the status instead

## How can this be tested?

Same as in https://github.com/Dynatrace/dynatrace-operator/pull/6977

BUT, you no longer need to set the image 😉 
- it should auto resolve the image based on the usual `dk.FF().IsPublicRegistry()` check


pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -213,8 +211,36 @@ func (r *Reconciler) delete(ctx context.Context, dk *dynakube.DynaKube) error {
...
+	case dk.KubernetesMonitoring().GetCustomImage() != "":
+		imageURI = dk.KubernetesMonitoring().GetCustomImage()
+	case dk.FF().IsPublicRegistry():
+		imageURI, err = registry.ResolveImage(ctx, imageClient, dk.PublicRegistryOverride(), image.ActiveGate)
+	default:
+		var latestVersion string
+
+		latestVersion, err = versionClient.GetLatestActiveGateVersion(ctx, installer.OSUnix)
+		imageURI = dk.KubernetesMonitoring().GetDefaultImage(latestVersion)
+	}
```
0sewa0, line 240: I intentionally didn't use the `version.Reconciler` or its output.

Both the `versionClient.GetLatestActiveGateVersion` and `imageClient.GetComponentLatestInfo` are using the dtclient's cache (as proved by the `reconciler_autoupdate_test.go`), so the `version.Reconciler` is redundant and just adds extra levels of complexity.

and in the future the `versionClient.GetLatestActiveGateVersion` scenario can be just removed.

Open question:
I didn't use the `VersionStatus` that is already in the kubemon's status(but unused). Should I?
- If yes? What parts of it?
- If no, I will remove for the status

pkg/controllers/dynakube/kubemon/statefulset/reconciler_integration_test.go
```diff
@@ -180,13 +192,106 @@ func runReEnablePhase(t *testing.T, deps *lifecycleDeps) {
...
 	assertStatefulSetShape(t, sts, deps.dk)
 	assert.Equal(t, deps.rotatedTenantTokenHash, sts.Spec.Template.Annotations[statefulset.AnnotationTenantTokenHash])
 }
 
+// TestReconcileLifecycleAutoImage walks provision → rollout-complete → switch-to-public-registry
+// with images resolved at runtime rather than hard-coded. The test confirms that:
+//   - The version client drives image resolution in the default tenant-registry path.
+//   - Adding the public-registry annotation switches resolution to the image client, and the
+//     StatefulSet is updated with the URI returned by that client.
+func TestReconcileLifecycleAutoImage(t *testing.T) {
```
0sewa0, line 207: kinda duplicates `reconciler_autoupdate_test.go`

but `reconciler_autoupdate_test.go` checks that the client's cache works as expected.

while `TestReconcileLifecycleAutoImage` is just the same as `TestReconcileLifecycle` but without a hardcoded image + I check a change between tenant to public registry

pkg/api/latest/dynakube/kubemon/props.go
```diff
@@ -66,3 +80,28 @@ func (km *Spec) GetCustomImage() string {
...
+	tag := truncatedVersion
+
+	if !strings.Contains(tag, api.RawTag) {
+		tag += "-" + api.RawTag
+	}
+
+	return apiURLHost + TenantRegistrySubPath + ":" + tag
+}
+
+func (km *KubeMon) apiURLHost() string {
```
avorima: could use APIURLHost here instead parsing it separately. we can rethink the "fail on purpose" behavior

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
avorima: we should still store the resolved image in the status, but everything else can be removed. probably don't need the existing version status

pkg/controllers/dynakube/kubemon/statefulset/reconciler_integration_test.go
avorima: i don't see anything here that isn't covered by the unit test(s)

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
0sewa0: My current plan is to do what the [KOD](https://github.com/dynatrace-apps/dynatrace-operator-internal/blob/main/docs/designs/kod-002-activegate-operand-split.md) expects, which is a single `version` status field

0sewa0: Populated the `version` field in https://github.com/Dynatrace/dynatrace-operator/pull/7074/commits/84aa758550908e4844232125327008d2582632d7 according to the KOD, so it is just the version/tag string, not the full resolved image (unless it is custom image)

However, I agree with Mario that it might be more beneficial to "store the resolved image in the status".
- It would be more consistent
- Incase of the public-registry usecase, the used registry may vary (the tenant registry was always the same)
- Simpler logic in general

But then a status field `version` is not the best name. I would just go with an `image` status field (the same as a `Deployment` would do for example)

WDYT? (EDIT: done in [version -> image](https://github.com/Dynatrace/dynatrace-operator/pull/7074/commits/49185573b9927b683cbd6ab259db916b02c37abb))

pkg/api/latest/dynakube/kubemon/props.go
```diff
@@ -66,3 +79,18 @@ func (km *Spec) GetCustomImage() string {
...
+
+func (km *KubeMon) GetDefaultImage(version string) string {
+	if km.apiURLHost == "" {
+		return ""
+	}
+
+	truncatedVersion := dtversion.ToImageTag(version)
+	tag := truncatedVersion
+
+	if !strings.Contains(tag, api.RawTag) {
```
avorima: HasSuffix would be more accurate here

pkg/api/latest/dynakube/kubemon/status.go
```diff
@@ -5,19 +5,19 @@ package kubemon
...
-	status.VersionStatus `json:",inline"`
+	// Information about the resolved image of KubernetesMonitoring that is meant to be deployed.
+	Image string `json:"image,omitempty"`
 
 	// Information about KubernetesMonitoring's connections.
 	// +kubebuilder:validation:Optional
 	ConnectionInfo communication.ConnectionInfo `json:"connectionInfo,omitzero"`
 }
 
 func (s *Status) IsZero() bool {
```
avorima: removing the version status actually obsoletes this method

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
jdanieck: +1 for fully resolved image in the status

pkg/api/latest/dynakube/kubemon/props.go
waodim: This same `strings.Contains` pattern already exists at least in the OneAgent and ActiveGate props.

IMO If we fix it here, we could fix all of them for consistency, or we'll have a mix of `Contains` and `HasSuffix` across the codebase. But just minor nit, feel free to ignore if out of scope.

0sewa0: will do a fly by

pkg/api/latest/dynakube/kubemon/status.go
0sewa0: because of:
```
// IsZero reports whether every field is zero. It is required for the `omitzero`
// JSON tag: the promoted status.VersionStatus.IsZero() only inspects the version
// fields, so without this method omitzero would drop the whole status (including
// connection info) whenever the version happens to be unset.
```

but this comment didn't exist on this `IsZero` 🥲

pkg/api/latest/dynakube/kubemon/props.go
0sewa0: [done](https://github.com/Dynatrace/dynatrace-operator/pull/7074/commits/c81ee55f82fa0e3569a61f8461e6303c5a5a5aff)

pkg/api/latest/dynakube/kubemon/status.go
0sewa0: [done](https://github.com/Dynatrace/dynatrace-operator/pull/7074/commits/4c520f1e757e58ce7d67cc7dd6617197f3dccdfc)

avorima (dismissed):
the image gets resolved correctly, but there's no image pull secret for the tenant registry. i'm fine with doing that in a follow-up

0sewa0:
> the image gets resolved correctly, but there's no image pull secret for the tenant registry. i'm fine with doing that in a follow-up

@avorima  totally missed that, will fix it in context of this, the pull-secrets were not even propagated to this statefulset

also, found some code smells related to this that I will fix in a fly by

waodim (dismissed):
Tested: 1. pub reg. 2. tenant reg. 3. custom image

Image is resolved ✅ But I can confirm, what @avorima mentioned. But we can do it in a different PR as well

jdanieck (dismissed):

aorcholski (dismissed):
I checked
- tenant registry
- use-public-registry
- custom image

Image resolved properly.

0sewa0:
Fixed the pull-secret issue.

The actual changes are trivial.
I added a bunch of additional tests. So that is most of the added changes

Actual logic:
[Propagate pull secrets to kubemon sts](https://github.com/Dynatrace/dynatrace-operator/pull/7074/commits/43a6aee8f7b29b51470914c0d4c18b5b6ea21838)
[Create dtpullsecret for standalone kubemon](https://github.com/Dynatrace/dynatrace-operator/pull/7074/commits/e63d3a146a1847311f244ec8c924928a4e8b78d4)

Fly by clean-up:
[Add imagePullSecret option to StatefulSet builder](https://github.com/Dynatrace/dynatrace-operator/pull/7074/commits/f8b4a02309ccd006aa7491784a63e48fc0c42d42)
[Remove reimplementation](https://github.com/Dynatrace/dynatrace-operator/pull/7074/commits/ee334664152c2c8fc07adca7b82774d7c1f95af9)


EDIT: Have to rebase due to https://github.com/Dynatrace/dynatrace-operator/pull/7048 🌚

pkg/controllers/dynakube/dtpullsecret/reconciler.go
```diff
@@ -36,7 +37,9 @@ func NewReconciler(clt client.Client, apiReader client.Reader) *Reconciler {
 func (r *Reconciler) Reconcile(ctx context.Context, dk *dynakube.DynaKube, tokens token.Tokens) error {
 	ctx, log := logd.NewFromContext(ctx, "pullsecret")
 
-	if dk.FF().IsPublicRegistry() || (!dk.OneAgent().IsDaemonsetRequired() && !dk.ActiveGate().IsEnabled()) {
+	isStandaloneKubemon := k8senv.IsKubemonOperandEnabled() && dk.KubernetesMonitoring().IsEnabled()
```
avorima: nit: "standalone" sounds like it's the only activegate

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -213,8 +210,45 @@ func (r *Reconciler) delete(ctx context.Context, dk *dynakube.DynaKube) error {
...
+
+		latestVersion, err = versionClient.GetLatestActiveGateVersion(ctx, installer.OSUnix)
+		if err != nil {
+			return "", err
+		}
+
+		imageURI = dk.KubernetesMonitoring().GetDefaultImage(latestVersion)
+	}
+
+	dk.KubernetesMonitoring().Image = imageURI
```
avorima: let's not do that. there's a spec field with the same name. please spend the few extra characters to be unambiguous

avorima (commented):
providing the tenant registry secret even for custom images matches the current ActiveGate reconciler, but given that we're targeting this for a version where the tenant registry will be moving towards deprecation we should may re-evaluate that

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
0sewa0: you wanna know something funny?

the linter autofixes `dk.KubernetesMonitoring().Status.Image` to `dk.KubernetesMonitoring().Image` 🌚 🤦

avorima: yeah, i think in this case we should just assign to the status without this helper. it's not worth the nolint

0sewa0: I will rename the field to `ResolvedImage` but keep the json tag as is.

I do not want to accidentally confuse these again. It is just a bug asking to happen

avorima:
The pull secret is not being created. Image: ghcr.io/dynatrace/dynatrace-operator@sha256:e3ee6470f739b10f51b1f203d42fcfacccfd917d4b988d637bc6b374f85dab3e, commit: f633dd7505be2a2d358fd1b80ef6683aa005d68b

pkg/controllers/dynakube/dtpullsecret/reconciler.go
```diff
@@ -36,7 +37,9 @@ func NewReconciler(clt client.Client, apiReader client.Reader) *Reconciler {
 func (r *Reconciler) Reconcile(ctx context.Context, dk *dynakube.DynaKube, tokens token.Tokens) error {
 	ctx, log := logd.NewFromContext(ctx, "pullsecret")
 
-	if dk.FF().IsPublicRegistry() || (!dk.OneAgent().IsDaemonsetRequired() && !dk.ActiveGate().IsEnabled()) {
+	isKubemonOperandEnabled := k8senv.IsKubemonOperandEnabled() && dk.KubernetesMonitoring().IsEnabled()
```
waodim, line 40: Would a TODO make sense here because I think the kubemonOperand field will be removed in the future, no?

pkg/controllers/dynakube/dtpullsecret/reconciler_test.go
```diff
@@ -324,6 +326,102 @@ func TestReconciler_Reconcile(t *testing.T) {
...
+		var pullSecret corev1.Secret
+		err = fakeClient.Get(t.Context(),
+			client.ObjectKey{Name: testName + "-pull-secret", Namespace: testNamespace},
+			&pullSecret)
+
+		require.NoError(t, err)
+		assert.NotEmpty(t, pullSecret.Data)
+	})
+	t.Run("Don't create with standalone kubemon when operand env is not set", func(t *testing.T) {
+		// KUBEMON_ENABLE_OPERAND is not set; kubemon spec is present but the operand is disabled
```
waodim:
```suggestion
		// EXPERIMENTAL_ENABLE_KUBEMON_OPERAND is not set; kubemon spec is present but the operand is disabled
```

pkg/controllers/dynakube/dtpullsecret/reconciler.go
0sewa0: The moment `k8senv.IsKubemonOperandEnabled()` disappears this will also, we don't need to litter the codebase with TODOs

0sewa0: (also, I already have a tiny refactor here, which I want to PR after this 👀 )

waodim (approved):
Tenant registry does now just throw error that the image does not exist (thats a me problem because of my tenant), but not that there is no auth as before, 👍

pkg/controllers/dynakube/dtpullsecret/reconciler_test.go
```diff
@@ -324,6 +326,102 @@ func TestReconciler_Reconcile(t *testing.T) {
...
+		var pullSecret corev1.Secret
+		err = fakeClient.Get(t.Context(),
+			client.ObjectKey{Name: testName + "-pull-secret", Namespace: testNamespace},
+			&pullSecret)
+
+		assert.True(t, k8serrors.IsNotFound(err))
+	})
+	t.Run("Don't create when operand env is set but kubemon is not configured", func(t *testing.T) {
+		t.Setenv(k8senv.ExperimentalEnableKubemonOperand, "true")
+		dk := addFakeTennantUUID(&dynakube.DynaKube{
```
avorima, line 370: i know you didn't write it, but you could you fix the typo in this function please?

pkg/controllers/dynakube/kubemon/reconciler_test.go
```diff
@@ -101,10 +108,10 @@ func TestReconcileConditionMapping(t *testing.T) {
 		},
 		{
 			name:           "stack-wrapped error -> error without stack trace in message",
-			statefulSetErr: pkgerrors.WithStack(kubemonstatefulset.ErrImageRequired),
+			statefulSetErr: pkgerrors.WithStack(errors.New("sentinel error")),
```
avorima, line 111: why do we need stacktraces in unit tests?

avorima (approved):
I think we can merge it. Maybe just add these changes to your follow-up PR. Close at your own discretion

PR #7078, opened 2026-07-29
Log skipped rules during injection

avorima:
## Description

We should be logging all rules that are skipped. #7026 just added the logs to the client that filters by type/condition, but I forgot to add the log in #7025.

The logs are duplicated if you use both OTLP auto config and metadata enrichment, but there's no easy way around this since they use shared code.

```
{"level":"info","ts":"2026-07-29T09:04:16.685Z","logger":"admission.pod-mutation.injection.otlp.resource-attributes","msg":"ignoring metadata enrichment rules with lower precedence","object":{"name":"","namespace":"inject"},"namespace":"inject","name":"","resource":{"group":"","version":"v1","resource":"pods"},"user":"system:serviceaccount:kube-system:replicaset-controller","requestID":"af2b1781-a66a-41c3-8755-72afdbff6017","podName":"nginx-56c45fd5ff-","namespace":"inject","rules":[{"type":"K8S_NAMESPACE_LABEL","source":"e2e-test-label","target":"dt.cost.product"}]}
{"level":"info","ts":"2026-07-29T09:04:57.186Z","logger":"admission.pod-mutation.injection","msg":"ignoring metadata enrichment rules with lower precedence","object":{"name":"","namespace":"inject"},"namespace":"inject","name":"","resource":{"group":"","version":"v1","resource":"pods"},"user":"system:serviceaccount:kube-system:replicaset-controller","requestID":"b915f9d8-d049-4c55-b0af-4d4ce4195e12","rules":[{"type":"K8S_NAMESPACE_LABEL","source":"e2e-test-label","target":"dt.cost.product"}]}
```

ICP-7797

## How can this be tested?

* Create two rules on the tenant that have the same target
* Label or annotate a namespace to ensure that one rule wins over the other
* Deploy a DynaKube with metadata enrichment or OTLP auto config (or both) enabled 
* Check that the rules appear in the correct order in the status
* Inject a workload in that namespace
* Check that the metadata annotation contains the expected  value
* Check the webhook logs to see if they mention the ignored rules

### DIFF
--- a/pkg/webhook/mutation/pod/attributes/read.go
+++ b/pkg/webhook/mutation/pod/attributes/read.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube/metadataenrichment"
+	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8senv"
 	"github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/mutator"
 	"github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/workload"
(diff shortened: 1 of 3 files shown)


pkg/webhook/mutation/pod/attributes/read.go
```diff
@@ -65,9 +68,15 @@ func (attrs *Pod) applyEnrichmentRules(namespace corev1.Namespace, dk dynakube.D
...
 				attrs.rules[rule.Target] = valueFromNamespace
+			} else {
+				skippedRules = append(skippedRules, rule)
 			}
 		}
 	}
+
+	if len(skippedRules) > 0 {
+		logd.FromContext(ctx).Info("ignoring metadata enrichment rules with lower precedence", "rules", skippedRules)
+	}
```
0sewa0, line 79: > // The last rule without a target to resolve to a value wins. 

So, technically not all "skipped" rules will show up here.

I don't know if we want to log those aswell.

avorima: Hm, I just checked and actually the API does not allow to set the target to null even with the old schema. I think we can actually remove that code, but I'd rather do that in 1.11

0sewa0: awesome, maybe an update to the comment would make sense, otherwise LGTM

0sewa0 (approved):

waodim (approved):

PR #7081, opened 2026-07-29
Add LivenessProbe and ReadinessProbe to the kubemon AG spec

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/ICP-8144)

## Description

Kubemon AG should have readinessprobe and livenessprobe configured as in case of classic AG.

## How can this be tested?

unittests

### DIFF
--- a/pkg/controllers/dynakube/activegate/statefulset/probes/probes.go
+++ b/pkg/controllers/dynakube/activegate/statefulset/probes/probes.go
@@ -1,0 +1,44 @@
+// Copyright Dynatrace LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package probes
+
+import (
+	agconsts "github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/activegate/consts"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func BuildReadinessProbe() *corev1.Probe {
+	return &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path:   "/rest/health",
+				Port:   intstr.IntOrString{IntVal: agconsts.HTTPSContainerPort},
+				Scheme: "HTTPS",
+			},
+		},
+		InitialDelaySeconds: 90,
+		PeriodSeconds:       15,
+		FailureThreshold:    3,
+		TimeoutSeconds:      2,
+		SuccessThreshold:    1,
+	}
+}
+
+func BuildLivenessProbe() *corev1.Probe {
+	return &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path:   "/rest/state",
+				Port:   intstr.IntOrString{IntVal: agconsts.HTTPSContainerPort},
+				Scheme: "HTTPS",
+			},
+		},
+		InitialDelaySeconds: 90,
+		PeriodSeconds:       30,
+		FailureThreshold:    2,
+		TimeoutSeconds:      1,
+		SuccessThreshold:    1,
+	}
+}
--- a/pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
+++ b/pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
@@ -13,6 +13,7 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/installer"
 	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/version"
 	agconsts "github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/activegate/consts"
+	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/activegate/statefulset/probes"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/connectioninfo"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/deploymentmetadata"
 	kubemonauthtoken "github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/kubemon/authtoken"
@@ -26,7 +27,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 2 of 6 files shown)


pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -32,6 +33,8 @@ const (
 	AnnotationAuthTokenHash   = api.InternalFlagPrefix + "kubemon-authtoken-hash"
 	StorageVolumeName         = "kubemon-storage"
 	AuthTokenVolumeName       = "kubemon-authtoken-secret"
+
+	HTTPSContainerPort = 9999
```
waodim: this const already exists in the consts file in the activegate package

pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go
```diff
@@ -286,6 +286,20 @@ func TestReconcileBuildsStatefulSet(t *testing.T) {
 		assert.Equal(t, statefulset.StorageVolumeName, sts.Spec.Template.Spec.Volumes[2].Name)
 		assert.NotNil(t, sts.Spec.Template.Spec.Volumes[2].EmptyDir)
 	})
+
+	t.Run("probes", func(t *testing.T) {
```
waodim: can we give a more descriptive name?

and the tests could be expanded to test 

example:

```
assert.Equal(t, int32(90), sts.Spec.Template.Spec.Containers[0].ReadinessProbe.InitialDelaySeconds)
assert.Equal(t, int32(15), sts.Spec.Template.Spec.Containers[0].ReadinessProbe.PeriodSeconds)
```

WDYT?

aorcholski: unittests cover all fields

regarding naming, I follow existing approach:

TestReconcileBuildsStatefulSet
- env vars..
- auth token..
..
- readiness probe
- liveness probe

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
aorcholski: right, I assumed everything related to kubemon has to be new. Fixed

waodim (dismissed):

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -207,6 +208,40 @@ func buildVolumeMounts(_ *dynakube.DynaKube) []corev1.VolumeMount {
 	return mounts
 }
 
+func buildReadinessProbe() *corev1.Probe {
```
andriisoldatenko: is it copy/paste from `func (statefulSetBuilder Builder) buildBaseContainer`

why not reuse it?

aorcholski: kubemon has own new directory in sources and has own method `buildDesiredStatefulSet`.
There are two small functions added because I want to make `buildDesiredStatefulSet` more readable

andriisoldatenko: nah my point is - we deploy AG and it has same livenessPRobe/rediness probe into 2 places.
So IMO we should reuse it because then it's easy to maintain if AG change it.

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -244,6 +279,8 @@ func (r *Reconciler) buildDesiredStatefulSet(ctx context.Context, dk *dynakube.D
 		Resources:       dk.KubernetesMonitoring().Resources,
 		Env:             buildEnvs(dk),
 		VolumeMounts:    buildVolumeMounts(dk),
+		ReadinessProbe:  buildReadinessProbe(),
```
andriisoldatenko: so my idea is to have something like:
```
ReadinessProbe: activegate.DefaultReadinessProbe()
```
or just
```
ReadinessProbe: activegate.ReadinessProbe()
```

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
jdanieck: if it can be re-used as is - sure make sense to do now , but if it requires AG refactoring I'd defer it and do as a separate ticket

when we initially started to work on kubemon we assumed we can re-use a lot from AG, which turned out not to be true due to some hardcoded assumptions in AG like multiple conditions or caching in .status (which we don't do in kubemon); my point is not to make kubemon work about AG refactoring - lets make kubemon feature complete with good tests, then AG refactor + unification with kubemon should be easy and safe to do

jdanieck (dismissed):

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
aorcholski: `activegate/statefulset/probes` module added because modules from the `.../internal/statefulset` can't be imported

pkg/controllers/dynakube/activegate/statefulset/probes/probes.go
```diff
@@ -0,0 +1,44 @@
...
+
+package probes
+
+import (
+	agconsts "github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/activegate/consts"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func BuildReadinessProbe() *corev1.Probe {
```
0sewa0: nitpick: nothing is being built here, there is no input even + probe is already the package name

```suggestion
func Readiness() *corev1.Probe {
```

so usage is `probes.Readiness()`

pkg/controllers/dynakube/activegate/statefulset/probes/probes.go
```diff
@@ -0,0 +1,44 @@
...
+		},
+		InitialDelaySeconds: 90,
+		PeriodSeconds:       15,
+		FailureThreshold:    3,
+		TimeoutSeconds:      2,
+		SuccessThreshold:    1,
+	}
+}
+
+func BuildLivenessProbe() *corev1.Probe {
```
0sewa0: nitpick: same as prev [comment](https://github.com/Dynatrace/dynatrace-operator/pull/7081/changes#r3682986694)
```suggestion
func Liveness() *corev1.Probe {
```

pkg/controllers/dynakube/activegate/statefulset/probes/probes.go
avorima: package name should be the singular probe -> `probe.Readiness`

andriisoldatenko: agree

andriisoldatenko: also we can just create activage/probe and probe.go

andriisoldatenko: probe is not part of STS it's just part of Container.

aorcholski: IMHO after further refactoring we will have more shared "pieces-of-STATEFULSET-spec" 
- pkg/controllers/dynakube/activegate/statefulset/probe/
- pkg/controllers/dynakube/activegate/statefulset/securitycontext
- pkg/controllers/dynakube/activegate/statefulset/< any-shared-part-of-sts-spec >
also on the statefulset level, template.spec.pod level....

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
andriisoldatenko: the way @aorcholski currently implemented to me looks good :)

andriisoldatenko (approved):

0sewa0 (approved):

PR #7100, opened 2026-07-30
feat: add omitzero to struct fields - edgeconnect - ICP-7911

VLazarevic:
## Description

https://dt-rnd.atlassian.net/browse/ICP-7911

Follow-up to ICP-973 (#6942), which introduced `omitzero` for the DynaKube CRD.
This PR does the same for the EdgeConnect CRD.

`omitempty` has no effect on non-pointer struct fields, so empty structs were
always written to the YAML (e.g. `imageRef: {}`, `resources: {}`,
`status: version: {}`). Go 1.24 added `omitzero`, which correctly drops these
fields when they are zero.

The change is applied only to the storage/hub version `v1alpha2`:
- `imageRef`, `resources`, `metadata`, `status`, `status.version` and
  `status.updatedTimestamp` now use `omitzero`
- `spec` is now marked as required (no omit)
- regenerated the EdgeConnect CRD

Since only `v1alpha2` was changed, the empty fields are dropped only for clients
using `v1alpha2`. Clients using the older `v1alpha1` version keep the previous
behavior.

I also created an integration test
(`serialization_integration_test.go`), following the DynaKube approach.

Verified that the empty {} blocks no longer cause a diff in ArgoCD for v1alpha2 resources." I left it out to keep it short, but it directly answers one of the acceptance criteria, so it may be worth including.

## How can this be tested?

The change lives in the conversion webhook, so the operator from this branch
must be deployed. `omitzero` only takes effect during conversion, so apply the
non-storage version `v1alpha1` and read the object back as the storage version
`v1alpha2`.

1. Deploy the operator from this branch.
2. Apply a minimal EdgeConnect as `v1alpha1`:

```yaml
kind: EdgeConnect
metadata:
  name: edgeconnect
  namespace: dynatrace
spec:
  apiServer: abc12345.dev.apps.dynatracelabs.com
  oauth:
    clientSecret: edgeconnect-oauth
    endpoint: https://sso.dynatrace.com/sso/oauth2/token
    resource: urn:dtaccount:abc12345
  imageRef: {}
  resources: {}
```

3. Read it back in both versions:

```bash
# v1alpha2 (storage): imageRef / resources are gone
kubectl get edgeconnects.v1alpha2.dynatrace.com edgeconnect -n dynatrace -o yaml

# v1alpha1: imageRef: {} / resources: {} are still present
kubectl get edgeconnects.v1alpha1.dynatrace.com edgeconnect -n dynatrace -o yaml
```

Expected: the empty `imageRef: {}` and `resources: {}` blocks are gone in
`v1alpha2` and still present in `v1alpha1`. (Note: the
`last-applied-configuration` annotation still shows the original manifest,
so ignore it and look at the actual `spec`.)


pkg/api/validation/edgeconnect/serialization_integration_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+			golden := filepath.Join("testdata", "serialized-"+tt.name+".yaml")
+			if *updateSerializationGolden {
+				require.NoError(t, os.WriteFile(golden, gotData, 0o600))
+			}
+
+			wantData, err := os.ReadFile(golden)
+			require.NoError(t, err)
+
+			assert.Equal(t, string(wantData), string(gotData))
```
andriisoldatenko: have tried `assert.YAMLEq(t, expected, actual, "optional message")` ?

VLazarevic: tried it out, it works fine: [b88d36f](https://github.com/Dynatrace/dynatrace-operator/pull/7100/commits/b88d36f384b3d4d2bcf51d4ad43eaef404766f93)

avorima (approved):

PR #7106, opened 2026-07-30
Introduce a limit for the value of metadata.dynatrace.com annotation

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/ICP-7924)

## Description

Adds webhook.metadatSizeLimit value.

Pod is not injected if size of combined value of the metadata annotation is greater than the limit. 
Then the Pod is annotated with the reason:
```
     dynakube.dynatrace.com/injected: "false"
     dynakube.dynatrace.com/reason: 'Value of the metadata enrichment annotation exceeds
       the maximum allowed size of 10, actual size: 87'
```

The limit has [no range](https://dt-rnd.atlassian.net/browse/ICP-7924?focusedCommentId=4762692).

## How can this be tested?

unittests

deploy operator with small limit (like < 20) , CNFS, install an app, check if the app is injected

### DIFF
--- a/pkg/util/kubernetes/fields/k8senv/env.go
+++ b/pkg/util/kubernetes/fields/k8senv/env.go
@@ -64,6 +64,9 @@ const (
 	defaultWebhookCertsRootDuration = 365 * 24 * time.Hour
 	minWebhookCertsRootDuration     = 7 * 24 * time.Hour
 	maxWebhookCertsRootDuration     = 10 * 365 * 24 * time.Hour
+
+	WebhookMetadataSizeLimitEnvVar       = "DT_METADATA_SIZE_LIMIT"
+	defaultWebhookMetadataSizeLimitValue = 24 * 1024
 )
 
 func Find(envVars []corev1.EnvVar, name string) *corev1.EnvVar {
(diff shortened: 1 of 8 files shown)


pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -237,3 +237,60 @@ func getSafeDurationFromEnv(ctx context.Context, envVar string, defaultValue, mi
...
+}
+
+type BoundedInt struct {
+	EnvName       string
+	DefaultValue  int
+	ResolvedValue int
+	ctx           context.Context
+}
+
+func (d BoundedInt) WithLog(ctx context.Context) BoundedInt {
```
avorima: it seems that the only purpose of this is to prevent logging the same thing over and over. since it's just a nil check this doesn't need to live in the struct. passing nil to `Resolve(context.Context)` would have the same result

aorcholski: Fixed. The `Resolve` function has ctx argument instead of `WithLog` method.

pkg/webhook/mutation/pod/register.go
```diff
@@ -77,6 +77,8 @@ func newWebhook( //nolint:revive
 
 	log.Info("got webhook's image from env", "image", webhookImage)
 
+	_ = k8senv.GetMetadaSizeLimit().Resolve(ctx)
```
avorima: please add a comment here that this is for logging otherwise we might accidentally remove it

aorcholski: done

pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -237,3 +237,53 @@ func getSafeDurationFromEnv(ctx context.Context, envVar string, defaultValue, mi
 
 	return duration
 }
+
+func GetMetadaSizeLimit() BoundedInt {
```
avorima: I talked with @0sewa0 and we'll do the change to use the structs later (ICP-8206). In the meantime keep using the old way with constants and normal functions

aorcholski: done

avorima (approved):

PR #7107, opened 2026-07-30
Add DtPrometheus

StefanHauth:
## Description

Add and register the new CRD DtPrometheus.

[JIRA](https://dt-rnd.atlassian.net/browse/ICP-7432)

## How can this be tested?

* unit tests
* apply CRD and CR
```
kubectl apply -f config/crd/bases/dynatrace.com_dtprometheuses.yaml
kubectl apply -f assets/samples/dtprometheus/v1alpha1/basic.yaml
```


assets/samples/dtprometheus/v1alpha1/basic.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: dynatrace.com/v1alpha1
```
waodim, line 1: DynaKube handles similar features (examples. KSPM, LogMonitoring) with a pointer: `*kspm.Spec` where nil = disabled and non-nil = enabled. Here we're using a value type + enabled bool, which is a different pattern.

The code comment on `selfMonitoring.enabled` says "Set to false to opt out" so the intent is clearly opt-out. But I tested the actual behavior on the cluster:

- Applied a CR with only `dynaKubeName` set (no `selfMonitoring` field) → `selfMonitoring` absent in the stored object
- Patched the same CR with selfMonitoring: {} → API server injected `enabled: true`

So a user who omits `selfMonitoring` entirely gets it disabled, which contradicts the documented intent. Or am I missing something?

0sewa0: I am also for `selfMonitoring: {}` == ON and `spec.selfMonitoring: nil` == OFF, with no `enabled` field.

I think it is a nicer pattern to have the user enable things and having them visible in the CR.
- We are also going that way with the `spec.kubernetesMonitoring.registration: {}`
  - this is currently (using `spec.activeGate`) is auto enabled and have to turn off with a feature-flag
  - now the user will explicitly have to opt in, or at least will be informed about it when looking at it in the sample / onboarding page.

**IMO The recommended defaults for enabled features should come from our samples and the onboarding page, and not from the CRs implementation.**

edit: defaults added via `// +kubebuilder:default=true` would make more sense for me on lower level settings such as `scrapeInterval` or `pullPolicy`

assets/samples/dtprometheus/v1alpha1/complete.yaml
```diff
@@ -0,0 +1,264 @@
...
+    # single hardware failure taking down all replicas simultaneously.
+    topologySpreadConstraints:
+      - maxSkew: 1
+        topologyKey: topology.kubernetes.io/zone
+        whenUnsatisfiable: DoNotSchedule
+
+    # Assign a PriorityClass so the TA is not evicted under node pressure.
+    # Losing the TA causes all scrapers to lose target assignments until it
+    # recovers and they re-poll - make it the highest-priority component.
+    priorityClassName: high-priority
```
gkrenn: The priority class does not exist. Either created it in the sample or remove it

pkg/api/v1alpha1/dtprometheus/common/props.go
avorima, line 1: do not use words like common, util(s), shared as an imported package name. these values are meant for the dtprometheus, so they should live there

pkg/api/v1alpha1/dtprometheus/common/spec.go
```diff
@@ -0,0 +1,63 @@
...
+	Replicas *int32 `json:"replicas,omitempty"`
+
+	// Overrides the default component image (full reference including tag or digest).
+	// When omitted the operator uses the latest available image from the public Dynatrace registry.
+	// +kubebuilder:validation:Optional
+	Image string `json:"image,omitempty"`
+
+	// Image pull policy for the component pods.
+	// +kubebuilder:validation:Optional
+	PullPolicy image.PullPolicy `json:"pullPolicy,omitempty"`
```
0sewa0: don't reinvent the wheel, in a container spec it is called `imagePullPolicy`, lets keep to that

pkg/api/v1alpha1/dtprometheus/common/props.go
```diff
@@ -0,0 +1,69 @@
...
+	return spec.PriorityClassName
+}
+
+// GetAnnotations returns the extra pod annotations for the component.
+func (spec *Spec) GetAnnotations() map[string]string {
+	return spec.Annotations
+}
+
+// GetLabels returns the extra pod labels for the component.
+func (spec *Spec) GetLabels() map[string]string {
```
avorima: let's not repeat the pattern for needless getters from the DynaKube

pkg/api/v1alpha1/dtprometheus/gateway/spec.go
```diff
@@ -0,0 +1,30 @@
...
+
+	name string
+}
+
+// +kubebuilder:object:generate=true
+
+// Spec configures the gateway pool StatefulSet (tier 2): a StatefulSet of OTel
+// Collectors that run stateful processors and export to Dynatrace via OTLP/HTTP.
+type Spec struct {
+	common.Spec `json:",inline"`
```
0sewa0: I don't know if I am a fan of this, sharing the `Spec` to safe on code duplication is not something I would do in the api, but I would need more opinions here.

don't tie the 2 things together unnecessarily.

pkg/api/v1alpha1/dtprometheus/common/spec.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// +kubebuilder:object:generate=true
+
+// Spec holds the pod-level settings shared by all DtPrometheus components.
+type Spec struct {
+	// Number of replicas for the component. At least 2 is recommended for
+	// failure tolerance.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=2
```
avorima: no defaulting here if we want to support nil as a value

pkg/api/v1alpha1/dtprometheus/scraper/props.go
```diff
@@ -0,0 +1,33 @@
...
+
+// GetName returns the base name for the scraper's Kubernetes resources.
+func (s *Scraper) GetName() string {
+	return s.name + NameSuffix
+}
+
+// GetUpdateStrategy returns the Deployment update strategy for the scraper pool.
+func (s *Scraper) GetUpdateStrategy() appsv1.DeploymentStrategy {
+	return s.UpdateStrategy
+}
```
0sewa0: is this necessary?

`UpdateStrategy` is public, unless you do something extra with it, it just noise

pkg/api/v1alpha1/dtprometheus/scraper/props.go
```diff
@@ -0,0 +1,33 @@
...
+
+// SetName sets the owning DtPrometheus name.
+func (s *Scraper) SetName(name string) {
+	s.name = name
+}
+
+// GetName returns the base name for the scraper's Kubernetes resources.
+func (s *Scraper) GetName() string {
+	return s.name + NameSuffix
+}
```
0sewa0: Confusing name
```suggestion
func (s *Scraper) GetDeploymentName() string {
	return s.name + NameSuffix
}
```

(I don't know if a scraper is a deployment or a statefulset, but I think you will get the point 😅 )

pkg/api/v1alpha1/dtprometheus/common/props.go
```diff
@@ -0,0 +1,69 @@
...
+
+package common
+
+import corev1 "k8s.io/api/core/v1"
+
+// DefaultReplicas is the replica count used when a component does not set one.
+const DefaultReplicas int32 = 2
+
+// GetReplicas returns the configured replica count, or the default when unset.
+func (spec *Spec) GetReplicas() int32 {
```
avorima: in order to handle HPA we will need to check for nilness. i don't think we actually need any getters for this PR

pkg/api/v1alpha1/dtprometheus/selfmonitoring/props.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+}
+
+// SetName sets the owning DtPrometheus name.
+func (sm *SelfMonitoring) SetName(name string) {
+	sm.name = name
+}
+
+// GetName returns the base name for the self-monitoring collector's Kubernetes resources.
+func (sm *SelfMonitoring) GetName() string {
```
0sewa0: same as https://github.com/Dynatrace/dynatrace-operator/pull/7107/changes#r3688789369

assets/samples/dtprometheus/v1alpha1/complete.yaml
```diff
@@ -0,0 +1,264 @@
...
+  # settings (apiUrl, tokens, proxy, networkZone, trustedCAs, ActiveGate, etc.).
+  dynaKubeName: dynakube
+
+  # Dynatrace preset
+  # When enabled, the operator creates a built-in ScrapeConfig that discovers
+  # pods annotated with metrics.dynatrace.com/scrape: "true" and reads the
+  # port, path, and secure annotations to construct scrape targets.
+  # This provides backwards compatibility with the ActiveGate Kubernetes
+  # module annotation-based discovery workflow.
+  # The ScrapeConfig is labelled prometheus.dynatrace.com: "true" so it is
```
gkrenn: Maybe also add a comment down at the scrape CR selector, that the dynatrace label is required for this feature to work

pkg/api/v1alpha1/dtprometheus/targetallocator/props.go
```diff
@@ -0,0 +1,72 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// NameSuffix is appended to the owning DtPrometheus name to derive the base
+	// name of the Target Allocator's Kubernetes resources.
+	NameSuffix = "-target-allocator"
+
+	// DefaultScrapeInterval is applied when no scrape interval is configured.
+	DefaultScrapeInterval = "60s"
```
0sewa0: maybe a kubebuilder default? 👀

pkg/api/v1alpha1/dtprometheus/targetallocator/props.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+// SetName sets the owning DtPrometheus name.
+func (ta *TargetAllocator) SetName(name string) {
+	ta.name = name
+}
+
+// GetName returns the base name for the Target Allocator's Kubernetes resources.
+func (ta *TargetAllocator) GetName() string {
```
0sewa0: same as https://github.com/Dynatrace/dynatrace-operator/pull/7107/changes#r3688789369

assets/samples/dtprometheus/v1alpha1/complete.yaml
```diff
@@ -0,0 +1,264 @@
...
+  # automatically picked up by the Target Allocator without any extra config.
+  dynatracePreset:
+    # Enable the built-in annotation-based ScrapeConfig.
+    # Default: false
+    enabled: true
+
+  # TLS
+  # Controls TLS for internal component communication
+  # (TA -> scraper HTTPS, scraper -> gateway OTLP/TLS).
```
gkrenn: and from all components to self monitoring if enabled

pkg/api/v1alpha1/dtprometheus/targetallocator/props.go
```diff
@@ -0,0 +1,72 @@
...
+	return ta.name + NameSuffix
+}
+
+// GetScrapeInterval returns the configured scrape interval, or the default when unset.
+func (ta *TargetAllocator) GetScrapeInterval() string {
```
0sewa0: if [prev comment is possible](https://github.com/Dynatrace/dynatrace-operator/pull/7107/changes#r3688797347) then this is not needed 👀

pkg/api/v1alpha1/dtprometheus/targetallocator/props.go
```diff
@@ -0,0 +1,72 @@
...
+// GetScrapeCRNamespaceSelector returns the configured namespace selector. A nil or
+// empty selector matches all namespaces.
+func (ta *TargetAllocator) GetScrapeCRNamespaceSelector() *metav1.LabelSelector {
+	return ta.ScrapeCRNamespaceSelector
+}
+
+// GetUpdateStrategy returns the Deployment update strategy for the Target Allocator.
+func (ta *TargetAllocator) GetUpdateStrategy() appsv1.DeploymentStrategy {
+	return ta.UpdateStrategy
+}
```
0sewa0: Getters for simply public things are IMO not needed.

(if they were nested somehow, then it can make sense)

pkg/api/v1alpha1/dtprometheus/dtprometheus_status.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/status"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// DtPrometheusStatus defines the observed state of DtPrometheus.
+type DtPrometheusStatus struct { //nolint:revive
+	// Defines the current state (Running, Deploying, Error, ...)
+	DeploymentPhase status.DeploymentPhase `json:"phase,omitempty"`
```
0sewa0: nit: 
```suggestion
	Phase status.DeploymentPhase `json:"phase,omitempty"`
```

assets/samples/dtprometheus/v1alpha1/complete.yaml
```diff
@@ -0,0 +1,264 @@
...
+  tls:
+    # Set to false to disable operator-managed TLS.
+    # Use when a service mesh (Istio, Linkerd) handles mTLS between components.
+    # Default: true
+    enabled: true
+
+  # Self-monitoring
+  # When enabled, the operator deploys a dedicated self-monitoring collector that
+  # ships the stack's own telemetry to Dynatrace. The collector's configuration
+  # (replicas, image, resources, scheduling, ...) is set on the referenced DynaKube.
```
gkrenn: Please add what happens in case nothing is set on the DynaKube

pkg/api/v1alpha1/dtprometheus/common/props.go
waodim: I'd agree. IMO a good approach would be to introduce eventual getters when the reconciler is in play and really needs them

assets/samples/dtprometheus/v1alpha1/complete.yaml
gkrenn: Same for the other sections that use it.

pkg/api/v1alpha1/dtprometheus/targetallocator/props.go
andriisoldatenko: yes please

pkg/api/v1alpha1/dtprometheus/common/props.go
andriisoldatenko: +1

pkg/api/v1alpha1/dtprometheus/common/props.go
gkrenn: Agree, let's just take the values from the struct directly

assets/samples/dtprometheus/v1alpha1/basic.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: dtprometheus
+  namespace: dynatrace
+spec:
+  # Required: Name of the DynaKube in the same namespace that provides all
+  # connection settings (apiUrl, tokens, proxy, networkZone, trustedCAs,
+  # ActiveGate, etc.).
+  #
+  # All other fields default. See complete.yaml for the full set of options.
+  dynaKubeName: dynakube
```
gkrenn, line 12: Would maybe add resource limits here

pkg/api/v1alpha1/dtprometheus/gateway/spec.go
gkrenn: I would combine them for at least gateway and scraper, as they are the same component, configured to do something else.

assets/samples/dtprometheus/v1alpha1/basic.yaml
StefanHauth: It is an requirement to have SFM as opt-out, because we want to make sure that we have observability in case something goes south. Turning SFM on for troubleshooting after it failed might be too late.

0sewa0: Then make sure that in the onboarding page / samples:
- It is always present
- The user knows the risk that comes from removing it  (comments / nice docs)

edit: (you can even add a validation warning, that explains the importance of this)

pkg/api/v1alpha1/dtprometheus/gateway/spec.go
0sewa0: if you go with what [mario suggested](https://github.com/Dynatrace/dynatrace-operator/pull/7107#discussion_r3688745195) I am ok 👍

pkg/api/v1alpha1/dtprometheus/dtprometheus_types.go
```diff
@@ -0,0 +1,112 @@
...
+	// port, path, and secure annotations to construct scrape targets.
+	// This provides backwards compatibility with the ActiveGate Kubernetes
+	// module annotation-based discovery workflow.
+	// +kubebuilder:validation:Optional
+	DynatracePreset DynatracePresetSpec `json:"dynatracePreset,omitzero"`
+
+	// Controls TLS for internal component communication
+	// (TA -> scraper HTTPS, scraper -> gateway OTLP/TLS).
+	// +kubebuilder:validation:Optional
+	TLS TLSSpec `json:"tls,omitzero"`
```
mihaitanasedt: When `TLS.Enabled` is set to false, `omitzero` treats it as a zero value and omits the entire tls block from the serialized output. The CRD defines enabled with `default: true`,making it impossible for users to persistently set `enabled: false`.

pkg/api/v1alpha1/dtprometheus/types.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha1/dtprometheus/targetallocator"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// DtPrometheusSpec defines the desired state of DtPrometheus.
+type DtPrometheusSpec struct { //nolint:revive
+	// Name of the DynaKube in the same namespace that provides all connection
+	// settings (apiUrl, tokens, proxy, networkZone, trustedCAs, ActiveGate, etc.).
+	// +kubebuilder:validation:Required
+	DynaKubeName string `json:"dynaKubeName"`
```
mihaitanasedt, line 20: lets add `// +kubebuilder:validation:MinLength=1`, to avoid dynakubeName: ""

mihaitanasedt (commented):
Should we add to `hack/helm/generate-crd.sh` a step for generating labels for DTPrometheus?

svfoxat:
edit: same as @mihaitanasedt comment above

In `hack/helm/generate-crd.sh` we inject `dynatrace-operator.commonLabels` into the `dynakube` and `edgeconnect` crds with the two `awk` blocks there, do we also need that for the new crd?

assets/samples/dtprometheus/v1alpha1/basic.yaml
StefanHauth: No, the basic sample should demonstrate the bare minimum a customer has to configure (aka. all non-optional fields).

StefanHauth:
I've tried to address all review comments:
* flatten structure
* remove over-engineered getters
* naming
* etc.

assets/samples/dtprometheus/v1alpha1/basic.yaml
StefanHauth: Flattened structure, renamed a bit [1f60b67](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/1f60b6755f12d585e46ec4a1c54c18ae7ea9a235)

pkg/api/v1alpha1/dtprometheus/common/props.go
StefanHauth: Flattened structure, renamed a bit [1f60b67](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/1f60b6755f12d585e46ec4a1c54c18ae7ea9a235)

pkg/api/v1alpha1/dtprometheus/common/props.go
StefanHauth: Removed lot's of over-engineered getters: [1f60b67](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/1f60b6755f12d585e46ec4a1c54c18ae7ea9a235)

pkg/api/v1alpha1/dtprometheus/common/spec.go
StefanHauth: [1f60b67](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/1f60b6755f12d585e46ec4a1c54c18ae7ea9a235)

pkg/api/v1alpha1/dtprometheus/common/props.go
StefanHauth: Fixed in [1f60b67](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/1f60b6755f12d585e46ec4a1c54c18ae7ea9a235)

assets/samples/dtprometheus/v1alpha1/complete.yaml
StefanHauth: Fixed in [1f60b67](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/1f60b6755f12d585e46ec4a1c54c18ae7ea9a235)

assets/samples/dtprometheus/v1alpha1/complete.yaml
StefanHauth: Let's leave out SFM for the moment until details are decided

assets/samples/dtprometheus/v1alpha1/complete.yaml
StefanHauth: Will be decided later.

StefanHauth:
> Should we add to `hack/helm/generate-crd.sh` a step for generating labels for DTPrometheus?

[1f60b67](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/1f60b6755f12d585e46ec4a1c54c18ae7ea9a235)

pkg/api/v1alpha1/dtprometheus/common.go
```diff
@@ -0,0 +1,65 @@
...
+
+package dtprometheus
+
+import (
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/shared/image"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// CommonPodSpec holds the pod-level settings shared by all DtPrometheus components.
+type CommonPodSpec struct {
```
0sewa0:
```suggestion
type PodSpec struct {
```

could also call the file `pod.go`,  because `common` is an alias for "dumpster of random stuff"

andriisoldatenko: Can we not use "common" and may be use `PodSpec`?

wdyt @avorima ? I saw you alaready asked to avoid such naming?

FYI: it's already inside package `dtprometheus.PodSpec`

pkg/api/v1alpha1/dtprometheus/common.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// Extra labels merged into the pod template metadata.
+	// +kubebuilder:validation:Optional
+	Labels map[string]string `json:"labels,omitempty"`
+}
+
+// GetPullPolicy returns the image pull policy for the component pods.
+func (spec *CommonPodSpec) GetPullPolicy() corev1.PullPolicy {
+	return corev1.PullPolicy(spec.ImagePullPolicy)
+}
```
0sewa0: why is this needed?

this turns `image.PullPolicy` to `corev1.PullPolicy`, could we just use 1 type? (also a getter should match the name of what it's getting 😅, but we don't really need a getter IMO )

edit: fyi I found this in the new kubemon spec as well, will try to fix it there

pkg/api/v1alpha1/dtprometheus/types.go
```diff
@@ -0,0 +1,105 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={}
+	// +nullable
+	TLS *TLSSpec `json:"tls"`
+
+	// Send self-monitoring for the stack's own telemetry to Dynatrace.
+	// Enabled by default; set to null to opt out.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={}
+	// +nullable
```
0sewa0: ok for now, the defaulting research will inform us if this ok for gitops tools
and in the KOD we can discuss further if we like to see `selfMonitoring: nil` in a rendered CR

from implementation perspective (for the controllers) it will be transparent

pkg/api/v1alpha1/dtprometheus/dtprometheus_props.go
```diff
@@ -0,0 +1,52 @@
...
+
+// TargetAllocator returns the Target Allocator accessor wrapping the target
+// allocator spec and the owning DtPrometheus name.
+func (dtp *DtPrometheus) TargetAllocator() *TargetAllocator {
+	return NewTargetAllocator(&dtp.Spec.TargetAllocator, dtp.Name)
+}
+
+// Scraper returns the scraper pool accessor wrapping the scraper spec and the
+// owning DtPrometheus name.
+func (dtp *DtPrometheus) Scraper() *Scraper {
```
andriisoldatenko: we have linter which we found yesterday seems don't work here:
```
        - name: var-naming
          arguments:
            - [] # AllowList
            - ["DT", "EEC", "KSPM", "OLM", "OTLP", "DK", "EC", "GRPC", "OS", "CSI"] # DenyList
```

We should use `DTPrometheus` instead

pkg/api/v1alpha1/dtprometheus/common.go
andriisoldatenko: +1

pkg/api/v1alpha1/dtprometheus/properties.go
```diff
@@ -0,0 +1,52 @@
+// Copyright Dynatrace LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package dtprometheus
```
andriisoldatenko, line 4: I know we follow this pattern in dk package, but I personally prefer to remove this duplication:
```
pkg/api/v1alpha1/dtprometheus/props.go
```
so remove prefix `dtprometheus` from all files inside package.

pkg/api/v1alpha1/edgeconnect/convert_from_test.go
```diff
@@ -230,7 +230,7 @@ func fromAreStatusesEqual(t *testing.T, src *edgeconnect.EdgeConnectStatus, dst
 
 	assert.True(t, reflect.DeepEqual(src.KubeSystemUID, dst.KubeSystemUID), "KubeSystemUID")
 
-	assert.True(t, reflect.DeepEqual(src.DeploymentPhase, dst.DeploymentPhase), "DeploymentPhase")
+	assert.True(t, reflect.DeepEqual(src.DeploymentPhase, dst.DeploymentPhase), "Phase")
```
andriisoldatenko: is this change nessesary for this PR?

pkg/api/v1alpha1/dtprometheus/common.go
0sewa0: [I found out why](https://github.com/Dynatrace/dynatrace-operator/blob/5ae3924b242dd00fa23a4132e810a1af41a1936a/pkg/api/shared/image/types.go#L68-L71)

```go
// +kubebuilder:validation:Enum=IfNotPresent;Always;Never

// PullPolicy is the image pull policy. Use a custom type to share the validation marker.
type PullPolicy string
```

this is to save on repeating `// +kubebuilder:validation:Enum=IfNotPresent;Always;Never`

IMO i would repeat the validation marker rather than redefining the pull policy with an extra getter everytime...
(cc @avorima )

pkg/api/v1alpha1/dtprometheus/common.go
StefanHauth: I agree that CommonPodSpec isn't good, but isn't PodSpec too broad? I think we all have a very specific idea of what a PodSpec is and it's not only about the few fields in this struct.

pkg/api/v1alpha1/dtprometheus/properties.go
StefanHauth: but then I'd suggest to use "properties.go" and get rid of the abbrevation.

pkg/api/v1alpha1/edgeconnect/convert_from_test.go
StefanHauth: @0sewa0 IIRC you suggested that.

pkg/api/v1alpha1/dtprometheus/common.go
0sewa0: > it's not only about the few fields in this struct

true, but calling it the `UserConfigurablePartsOfPodSpec` is a mouthful, no? 😅 

I think the name `PodSpec` is good enough, nicely scopes what can be inside it. (if we want to expand it in the future)
But adding a docu string that says `PodSpec is the parts of the corev1.PodTemplateSpec that we surface for the user to configure. We don't make the whole corev1 spec configurable to limit possible conflicts with what the Operator would need to configure for proper functionality` (or something)

pkg/api/v1alpha1/edgeconnect/convert_from_test.go
0sewa0: why was it changed for EC? 😅

pkg/api/v1alpha1/dtprometheus/common.go
StefanHauth: Yeah, I'm fine with that, will adopt.

pkg/api/v1alpha1/edgeconnect/convert_from_test.go
StefanHauth: Upsi, I read over the EC part of the file path 😅

pkg/api/v1alpha1/dtprometheus/common.go
0sewa0: I created a cleanup PR for this type https://github.com/Dynatrace/dynatrace-operator/pull/7205

I need more opinions, but I would not advise doing such "magic" to save on `// +kubebuilder: validation` declarations just to then having to compensate with extra getters and conversion

pkg/api/v1alpha1/dtprometheus/properties.go
andriisoldatenko: yes please do it.

pkg/api/v1alpha1/dtprometheus/common.go
0sewa0: FYI: because I like confusing people the smaller version https://github.com/Dynatrace/dynatrace-operator/pull/7204 got a bunch of positive feedback, but I closed it, so we can do it in a uniform way everywhere.

The big one (https://github.com/Dynatrace/dynatrace-operator/pull/7205), is also approved. Are you also ok with it?

pkg/api/v1alpha1/dtprometheus/common.go
StefanHauth: Renamed to PodSpec: [000b10b](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/000b10bf456dcf4600a63a29ce6721c11d1ed927)

pkg/api/v1alpha1/dtprometheus/dtprometheus_props.go
StefanHauth: Renamed: [000b10b](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/000b10bf456dcf4600a63a29ce6721c11d1ed927)

pkg/api/v1alpha1/dtprometheus/common.go
StefanHauth: Renamed to PodSpec and added doc: [000b10b](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/000b10bf456dcf4600a63a29ce6721c11d1ed927)

pkg/api/v1alpha1/dtprometheus/common.go
StefanHauth: Will adopt in Prometheus as well

pkg/api/v1alpha1/dtprometheus/properties.go
StefanHauth: Renamed: [000b10b](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/000b10bf456dcf4600a63a29ce6721c11d1ed927)

pkg/api/v1alpha1/edgeconnect/convert_from_test.go
StefanHauth: Corrected: [000b10b](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/000b10bf456dcf4600a63a29ce6721c11d1ed927)

pkg/api/v1alpha1/dtprometheus/common.go
StefanHauth: Fixed in [2d62922](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/2d629229a94943dd9fe206d985ca6227e0a74235)

waodim:
Codewise lgtm

Just 2 things:

- as discussed please add `publicRegistryOverride` field similar to the DK.
- and maybe keep this in mind https://github.com/Dynatrace/dynatrace-operator/pull/7107#discussion_r3689009096

Other than that - I tested, it works. Happy to approve if at least the field is added. 👍

StefanHauth:
> * as discussed please add `publicRegistryOverride` field similar to the DK.

[a4ff109](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/a4ff109f7df7440c85d151d2d8fa4c2bb6813ff9)

waodim (dismissed):

pkg/api/v1alpha1/dtprometheus/scraper.go
```diff
@@ -0,0 +1,54 @@
...
+// NewScraper wraps the given Spec together with the owning DTPrometheus name.
+func NewScraper(spec *ScraperSpec, name string) *Scraper {
+	return &Scraper{
+		ScraperSpec: spec,
+		name:        name,
+	}
+}
+
+// SetName sets the owning DTPrometheus name.
+func (s *Scraper) SetName(name string) {
```
avorima: these setter are not needed if the name is already provided in the constructor

StefanHauth: [a59b43b](https://github.com/Dynatrace/dynatrace-operator/pull/7107/commits/a59b43b4fe25dbaecfe61a0ccf74505c89e3a93d)

avorima (dismissed):

avorima (approved):

PR #7120, opened 2026-07-31
Small fixes related to dtpullsecret

0sewa0:
## Description

tiny follow up of https://github.com/Dynatrace/dynatrace-operator/pull/7074

Mainly non functional changes, fixing typos, converting "methods" into func, naming in tests.

1 functional change (kinda, but not really):
- Removed `getImageRegistryFromAPIURL`, which is basically the same as `dk.APIURLHost()` but can fail. (this fail scenario is not possible in reality as we need a tenantUUID for this whole thing to work, which means we need to be able to use the APIURL to make a call 🤷 )
- Which also caused me to stop using the `dk.APIURL()`, which is correct, because no gen3 tenants will have a tenant-registry to use this pull-secret for, so converting such URLs doesn't really help anyone.

> [!NOTE]
> I went down a rabbit hole of trying to make the tests pretty. Which I mostly discarded, as this logic is hopefully not going to be here for that much longer, so didn't want to change it for the sake of it. 

## How can this be tested?

No change in behavior

### DIFF
--- a/pkg/controllers/dynakube/dtpullsecret/generate.go
+++ b/pkg/controllers/dynakube/dtpullsecret/generate.go
@@ -6,7 +6,6 @@ package dtpullsecret
 import (
 	b64 "encoding/base64"
 	"encoding/json"
-	"fmt"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/token"
@@ -61,13 +60,13 @@ func generateData(dk *dynakube.DynaKube, tokens token.Tokens) (map[string][]byte
 	dockerCfg := newDockerConfigWithAuth(tenantUUID,
 		registryToken,
 		registry,
-		buildAuthString(tenantUUID, registryToken))
+		basicAuth(tenantUUID, registryToken))
 
 	return pullSecretDataFromDockerConfig(dockerCfg)
 }
 
-func buildAuthString(tenantUUID string, registryToken string) string {
-	auth := fmt.Sprintf("%s:%s", tenantUUID, registryToken)
+func basicAuth(username string, password string) string {
+	auth := username + ":" + password
 
 	return b64.StdEncoding.EncodeToString([]byte(auth))
 }
--- a/pkg/controllers/dynakube/dtpullsecret/generate.go
+++ b/pkg/controllers/dynakube/dtpullsecret/generate.go
@@ -7,7 +7,6 @@ import (
 	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
-	"net/url"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/token"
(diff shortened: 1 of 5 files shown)


pkg/controllers/dynakube/dtpullsecret/reconciler.go
```diff
@@ -37,9 +37,11 @@ func NewReconciler(clt client.Client, apiReader client.Reader) *Reconciler {
 func (r *Reconciler) Reconcile(ctx context.Context, dk *dynakube.DynaKube, tokens token.Tokens) error {
 	ctx, log := logd.NewFromContext(ctx, "pullsecret")
 
-	isKubemonOperandEnabled := k8senv.IsKubemonOperandEnabled() && dk.KubernetesMonitoring().IsEnabled()
+	anyRelevantOprandEnabled := dk.OneAgent().IsDaemonsetRequired() ||
+		dk.ActiveGate().IsEnabled() ||
+		(k8senv.IsKubemonOperandEnabled() && dk.KubernetesMonitoring().IsEnabled())
 
-	if dk.FF().IsPublicRegistry() || (!dk.OneAgent().IsDaemonsetRequired() && !dk.ActiveGate().IsEnabled() && !isKubemonOperandEnabled) {
+	if dk.FF().IsPublicRegistry() || !anyRelevantOprandEnabled {
```
waodim:
```suggestion
	if dk.FF().IsPublicRegistry() || !anyRelevantOperandEnabled {
```

pkg/controllers/dynakube/dtpullsecret/generate.go
```diff
@@ -65,26 +61,17 @@ func (r *Reconciler) generateData(dk *dynakube.DynaKube, tokens token.Tokens) (m
...
 		registry,
-		r.buildAuthString(tenantUUID, registryToken))
+		buildAuthString(tenantUUID, registryToken))
 
 	return pullSecretDataFromDockerConfig(dockerCfg)
 }
 
-func (r *Reconciler) buildAuthString(tenantUUID string, registryToken string) string {
+func buildAuthString(tenantUUID string, registryToken string) string {
 	auth := fmt.Sprintf("%s:%s", tenantUUID, registryToken)
```
avorima: nit: don't really need Sprintf here. here's the stdlib for reference:

```go
func basicAuth(username, password string) string {
	auth := username + ":" + password
	return base64.StdEncoding.EncodeToString([]byte(auth))
}
```

pkg/controllers/dynakube/dtpullsecret/generate_test.go
```diff
@@ -49,43 +36,42 @@ func TestReconciler_GenerateData(t *testing.T) {
...
 
 	tests := []struct {
 		name        string
 		tokens      token.Tokens
 		expectToken string
 	}{
 		{
-			"use paas token",
-			token.Tokens{
+			name: "use paas token",
```
avorima: you can leave out the field names if all fields are specified. for simple tests like these it can reduce the boilerplate. i'd actually go into the opposite direction

````go
{"use paas token", token.Tokens{token.PaaSKey: &token.Token{Value: testPaasToken}}, testPaasToken},
{"use api token", token.Tokens{token.APIKey: &token.Token{Value: testAPIToken}}, testAPIToken},

File: pkg/controllers/dynakube/kubemon/reconciler_test.go [hunk 4]
```diff
@@ -108,7 +107,7 @@ func TestReconcileConditionMapping(t *testing.T) {
 		},
 		{
 			name:           "stack-wrapped error -> error without stack trace in message",
```
waodim: big nit: but since `pkgerrors.WithStack()` was removed is this name still accurate?

0sewa0: lol no :D, will just remove the case 🙈

waodim (approved):

avorima (approved):

PR #7121, opened 2026-07-31
add rbac and crd for dtprometheus to helm chart

svfoxat:
## Description

▎ Note: Reviewable but may be subject to change while https://github.com/Dynatrace/dynatrace-operator/pull/7107 is still open.

https://dt-rnd.atlassian.net/browse/ICP-7434

- Updates helm chart for DtPrometheus with RBAC templates and unit tests for CRD deployment gate and RBAC objects.

## How can this be tested?

`make test/helm/unit`

### DIFF
--- a/config/helm/chart/default/templates/Common/prometheus/clusterrole-dtprometheus-operator.yaml
+++ b/config/helm/chart/default/templates/Common/prometheus/clusterrole-dtprometheus-operator.yaml
@@ -1,0 +1,35 @@
+# Copyright Dynatrace LLC
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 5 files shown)


0sewa0 (dismissed):
will approve as there is not much to test or verify here

if fine tuning is needed then that can be done when the controller is added

svfoxat:
> will approve as there is not much to test or verify here
> 
> if fine tuning is needed then that can be done when the controller is added

Alright, just had to rebase the branch onto main.
But I think we still have to wait for the CRD to land on main ~until we merge this, or else helm unit tests will be red on main.~

waodim (approved):

PR #7202, opened 2026-08-04
ICP-7550 Add kubemon customproperties

meibensteiner:
## Description

[ICP-7550](https://dt-rnd.atlassian.net/browse/ICP-7550)

Implementation of custom properties (inline value or referenced Secret via valueFrom) to the kubemon operand, as documented in [KOD-002](https://github.com/dynatrace-apps/dynatrace-operator-internal/blob/main/docs/designs/kod-002-activegate-operand-split.md).

## How can this be tested?

1. Enable the experimental kubemonoperand (`experimental.enableKubemonOperand=true`)
2. Apply a DynaKube with inline value: 
```yaml
kind: DynaKube
metadata:
  name: kubemon-cp-test
  namespace: dynatrace
spec:
  apiUrl: https://<tenant>.live.dynatrace.com/api
  tokens: <existing-tokens-secret>
  kubernetesMonitoring:
    customProperties:
      value: |
        [collector]
        restInterface = true
```
4. Verify secret, StatefulSet mount and the file mounted in the pod:
`kubectl -n dynatrace get secret kubemon-cp-test-kubemon-custom-properties -o jsonpath='{.data.customProperties}' | base64 -d`

[ICP-7550]: https://dt-rnd.atlassian.net/browse/ICP-7550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -346,3 +369,30 @@ func (r *Reconciler) getAuthTokenHash(ctx context.Context, dk *dynakube.DynaKube
...
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	data := secret.Data[kubemoncustomproperties.DataKey]
+	if len(data) == 0 {
+		return "", nil
+	}
+
+	hash, err := hasher.GenerateHash(string(data))
```
0sewa0: nit:
```suggestion
	hash, err := hasher.GenerateSecureHash(string(data))
```

it is more "Secure" as it uses a fancier algo, and as custom-properties is a secret, we should try to keep it secret

I see that we use `GenerateHash` always in this package, it should be a simple drop in replacement

pkg/controllers/dynakube/kubemon/README.md
```diff
@@ -34,11 +34,12 @@ workaround, and the operator can manage each capability's lifecycle separately.
 - Runtime wiring in StatefulSet (required env vars, token mount, restart trigger hash)
 - Support for core StatefulSet spec propagation (`rollingUpdate`, storage, DNS policy, priority class, termination grace period, ephemeral volume)
 - Cluster identity and deployment metadata seeded into the operand
+- Custom properties (inline `value` or referenced Secret via `valueFrom`), mounted from an optional Secret with pod-template hash annotation for rollout on change
 
 ### Missing
 
 - Self-registration in the Dynatrace UI (manual via Settings API for now)
-- KSPM-gated Service, image discovery, custom TLS, custom properties
+- KSPM-gated Service, image discovery, custom TLS
```
jdanieck: `image discovery` was done already in previous PR; I wonder if this README still makes sense, maybe we should drop it ? seems it got obsolete quickly; @0sewa0 WDYT? the original idea was to explain to external contributors what is the current state

0sewa0: feel free to remove it 👍

meibensteiner: The entire README or just the "Missing" section?

jdanieck: entire README

meibensteiner:
Forgot to run linter locally. I realize I made the tests more complex than they need to be. 👀

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	if len(data) == 0 {
+		return r.cleanup(ctx, dk)
+	}
+
+	return r.ensureSecret(ctx, dk, data)
+}
+
+func (r *Reconciler) resolveData(ctx context.Context, dk *dynakube.DynaKube) ([]byte, error) {
```
jdanieck, line 65: I wonder if we should warn when both `Value` & `ValueFrom` are set simultaneously. @0sewa0 how we handle this in other components ?

0sewa0: In the current AG custom-prop reconciler, `value` just takes precedence, and no validation is done.

Don't worry about validation for now, the [epic](https://dt-rnd.atlassian.net/browse/ICP-1814) is missing a story about validation in general (because of the tickets I moved out of it). I will collect all the validation we want to add into one story (copy paste from KOD 😅 ), and I will note this down as a potential warning. (also in the KOD)

Would make sense to document it on the docs string now, what takes precedence.

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
```diff
@@ -0,0 +1,104 @@
...
+	if src.ValueFrom == "" {
+		return nil, nil
+	}
+
+	referenced, err := r.secrets.Get(ctx, client.ObjectKey{Name: src.ValueFrom, Namespace: dk.Namespace})
+	if err != nil {
+		return nil, errors.WithMessagef(err, "failed to read custom properties secret %q", src.ValueFrom)
+	}
+
+	return referenced.Data[DataKey], nil
```
jdanieck: If the secret exists but does not contain DataKey, we return an empty slice which causes the reconciler to silently delete the managed secret and proceed as if no custom properties were configured.  The user gets no error and no log — kubemon reconciles successfully with the configuration dropped.                                                                                                    
                                                                                                                                                                                                     
Return an error instead when the key is absent, so the misconfiguration is surfaced immediately.

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
jdanieck: Is it difficult to add webhook validation in this PR ? Tracking in separate ticket is OK, but then you will need to test all those not-related functionalities together. Seems to me it would be easier to add it together with functionality PR, or am I missing smth?

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -176,6 +178,16 @@ func buildVolumes(dk *dynakube.DynaKube) []corev1.Volume {
 				},
 			},
 		},
+		{
+			Name: kubemoncustomproperties.VolumeName,
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName:  km.GetCustomPropertiesSecretName(),
+					DefaultMode: new(int32(0o640)),
+					Optional:    new(true),
```
aorcholski: if custom.properties is removed from the `spec.kubernetesMonitoring` section

```
failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/var/lib/kubelet/pods/084e7fe8-f9d7-4855-bf81-2c280a386506/volume-subpaths/kubemon-custom-properties/kubemon/2" to rootfs at "/va │
│ r/lib/dynatrace/gateway/config_template/custom.properties": mount src=/var/lib/kubelet/pods/084e7fe8-f9d7-4855-bf81-2c280a386506/volume-subpaths/kubemon-custom-properties/kubemon/2, dst=/var/lib/dynatrace/gateway/config_template/custom.properties, dstFd=/proc/thread-self/fd/11, flags=MS_RDONLY|MS_BIND|MS_REC: not a directory 
```

Anyway, assuming it works, Optional volume means we have no control over how the KUBEMON-AG is configured. The secret may be created manually and will be mounted even if there is no `kubernetesMonitoring.customProperties` section.

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
```diff
@@ -0,0 +1,104 @@
...
+
+	if src.Value != "" {
+		return []byte(src.Value), nil
+	}
+
+	if src.ValueFrom == "" {
+		return nil, nil
+	}
+
+	referenced, err := r.secrets.Get(ctx, client.ObjectKey{Name: src.ValueFrom, Namespace: dk.Namespace})
```
jdanieck, line 76: Consider mounting the referenced secret directly in the StatefulSet (with Optional: false and conditional volume inclusion) instead of copying it: avoids duplicating sensitive data, eliminates the stale second copy, and delegates missing-key validation to Kubernetes — pod fails to start if the secret or key is absent.

pkg/controllers/dynakube/kubemon/customproperties/reconciler_test.go
```diff
@@ -0,0 +1,301 @@
+// Copyright Dynatrace LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package customproperties_test
```
jdanieck, line 4: Should we add an integration test for the custom properties reconciler to match the pattern we follow in connectioninfo and authtoken? The reconciler is simple enough that the lifecycle phases won't surface much new logic, but it would give us owner reference and idempotency coverage against a real apiserver, and keeps the kubemon operand consistent. @0sewa0 @avorima @andriisoldatenko WDYT ?

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
0sewa0: it is just a [different package](https://github.com/Dynatrace/dynatrace-operator/blob/912071148a05b438db3cf19027501d2e411e3c8c/pkg/api/validation/dynakube/validation.go#L4), and can be easily done separately.

It is not needed for this PR, and I don't want to block or expand its scope further. (there are already many comments)

pkg/controllers/dynakube/kubemon/customproperties/reconciler_test.go
jdanieck: When I look at these tests now it feels we could go with just UTs for sub-reconcilers and have few more comprehensive integration test per operand - the same coverage, less code. Would be also nice to have mock Dynatrace server (there are tools that can generate one from openapi spec) alongside real API server & etcd (from envtest) - would bring us a lot closer to E2E tests without full setup needed.

This is just me thinking out loud, so not blocking/needed for this PR. I hope @andriisoldatenko will clarify things like this in our testing strategy doc.

meibensteiner: I'll do this in a followup to move this PR along 😅

pkg/controllers/dynakube/kubemon/README.md
meibensteiner: Removed!

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
meibensteiner: Returning an error now.

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
meibensteiner: Implemented and tests updated

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
meibensteiner: customProperties mount is now conditional as recommended by @jdanieck

pkg/api/latest/dynakube/kubemon/props.go
```diff
@@ -62,6 +63,18 @@ func (km *KubeMon) GetAuthTokenSecretName() string {
 	return km.name + NameSuffix + "-authtoken-secret"
 }
 
+func (km *KubeMon) GetCustomPropertiesSecretName() string {
+	return km.name + NameSuffix + "-custom-properties"
+}
+
+func (km *Spec) GetPullPolicy() corev1.PullPolicy {
```
svfoxat: Is this Getter really needed? The getter is currently never used but 
```
ImagePullPolicy: dk.KubernetesMonitoring().ImagePullPolicy
```
instead is used in `statefulset/reconciler.go:312`

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
```diff
@@ -0,0 +1,110 @@
...
+	if !dk.KubernetesMonitoring().IsEnabled() || dk.KubernetesMonitoring().CustomProperties == nil {
+		return r.cleanup(ctx, dk)
+	}
+
+	data, err := r.resolveData(ctx, dk)
+	if err != nil {
+		return err
+	}
+
+	if len(data) == 0 {
```
svfoxat, line 57: This will cleanup the Secret if `len(data) == 0`. This is also the case when CustomProperties is `&Source{}` (both Value and ValueFrom empty) -> resolveData returns a nil byte slice, so cleanup runs and no secret is created. 
The statefulset reconciler at line 188 only checks CustomProperties != nil, so it still appends the volume with Optional: false → `CreateContainerConfigError`.

Or is `customProperties: {}` validated/rejected somewhere else?

jdanieck: @0sewa0 [promised](https://github.com/Dynatrace/dynatrace-operator/pull/7202#discussion_r3711023582) to create ticket for validation so we can add yet another case to validate.

0sewa0: all the validation stuff is in https://dt-rnd.atlassian.net/browse/ICP-8385

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
jdanieck: Un-resolving as I don't see the change. 

I wanted to avoid having to `r.secrets.Get()` at all. We are reading contents of `src.ValueFrom` secret and replicating it into our managed `dk.KubernetesMonitoring().GetCustomPropertiesSecretName()` secret - this is unnecessary. We can just use `src.ValueFrom` secret directly in our managed Statefulset.

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
0sewa0: that being said, do not do a `cleanup` here, unnecessary

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
0sewa0: the nice thing about looking at the secret is you can do a verification:
```go
	data, ok := referenced.Data[DataKey]
	if !ok || len(data) == 0 {
		return nil, errors.Errorf("secret %q has no %q key", src.ValueFrom, DataKey)
	}
```
because if we do `delegate missing-key validation to Kubernetes`, we still have to handle it, but differently.

also, I think we do it in the original code is because to set the no-proxy for the AG you actually have to set a custom property, see in https://github.com/Dynatrace/dynatrace-operator/blob/20a434377aa2b34a7163b0f0db874c3b88be5bb0/pkg/controllers/dynakube/activegate/internal/customproperties/reconciler.go#L116
- we should also do this, probably logic should be extracted similarly [how the probes were](https://github.com/Dynatrace/dynatrace-operator/blob/20a434377aa2b34a7163b0f0db874c3b88be5bb0/pkg/controllers/dynakube/activegate/statefulset/probe/probe.go#L12)

sidenote: 
our validation should be a bit different IMO:
```go
	data, ok := referenced.Data[DataKey]
	if !ok  {
		return nil, errors.Errorf("secret %q has no %q key", src.ValueFrom, DataKey)
	} else if len(data) == 0 {
		return nil, errors.Errorf("secret %q has empty value under %q key", src.ValueFrom, DataKey)
	}
```

jdanieck: ok, let's keep it the way it is - if we need to modify the secret content then current implementation makes sense

that said, standard way to configure proxy is via env vars and I think we should at minimum create a ticket for AG to start supporting it; @aorcholski what do you think, possible for AG to read proxy from env vars ?

aorcholski: AG reads settings from mounted proxy-secret.
Or it could be configured via custom.properties - very low level approach. Not sure how passwords are provided (raw text, encoded....)

Internally AG (agctl) transfers proxy settings from proxy-secret VolumeMount to internal config . It should be done the same way manually using custom.properties.

jdanieck: So if we can use `proxy-secret` there is no need to modify `custom.properties` ref secret, hence my proposal is still viable :)

0sewa0: Proxy
No proxy is not part if it 😬

jdanieck: Anyways, It's a matter of judgment, @0sewa0 just make a call - I am good with both approaches; let us also keep in mind that we don't watch (or do we?) the customer secret so we propagate changes only on our 15 min (?) re-sync

0sewa0: The proxy secret we create does not contain the no proxy: https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/controllers/dynakube/proxy/reconciler.go

So we have to set it somewhere, and that place is custom properties 😥

0sewa0: > the customer secret so we propagate changes only on our 15 min (?) re-sync

Yes, I don't know a better way around it

jdanieck (dismissed):

aorcholski (dismissed):
LGTM
I added, changed, removed the secret.

pkg/api/latest/dynakube/kubemon/props.go
meibensteiner: Not needed, that was a bad merge. Fixed in https://github.com/Dynatrace/dynatrace-operator/pull/7202/commits/e3303d06739d713543afbceecd178dfe39be67ad

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
meibensteiner: Added separate logging for absent key vs present, but empty in https://github.com/Dynatrace/dynatrace-operator/pull/7202/commits/fbdbfb1d17027125a439c5f964aef18a3bd4cea5

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
meibensteiner: Fixed in https://github.com/Dynatrace/dynatrace-operator/pull/7202/commits/6ed000a0361802cae4ea318293789fd5797bd688

svfoxat (approved):

jdanieck (approved):

0sewa0 (approved):
we can implement the [noproxy handling](https://github.com/Dynatrace/dynatrace-operator/pull/7202/changes#r3729611194) separately

andriisoldatenko:
> we can implement the [noproxy handling](https://github.com/Dynatrace/dynatrace-operator/pull/7202/changes#r3729611194) separately

would be nice to trck it, do you mind to create a story for it?
cc @meibensteiner

pkg/controllers/dynakube/kubemon/customproperties/reconciler.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/objects/k8ssecret"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// Secret key holding the AG-internal ini.
+	DataKey = "customProperties"
```
andriisoldatenko, line 20: we usually call it: CustomPropertiesDataKey

andriisoldatenko: ah never mind, it's inside package name.

0sewa0:
> would be nice to trck it, do you mind to create a story for it?

@andriisoldatenko  @meibensteiner I just did https://dt-rnd.atlassian.net/browse/ICP-8504

andriisoldatenko:
found few minor issues, which will be resolved during implementing validation.

andriisoldatenko (approved):

PR #7204, opened 2026-08-04
Use `corev1.PullPolicy` + marker instead of custom type in kubemon

0sewa0:
## Description

I don't want to have
```go
func (km *Spec) GetPullPolicy() corev1.PullPolicy {
	if km == nil {
		return ""
	}

	return corev1.PullPolicy(km.ImagePullPolicy)
}
```

to save on `// +kubebuilder:validation:Enum=IfNotPresent;Always;Never`

Extended version that does this everywhere is https://github.com/Dynatrace/dynatrace-operator/pull/7205

## How can this be tested?

No change in behavior.

`imagePullPolicy: bloop` -> validation should catch it.
```
The DynaKube "marcell-dev" is invalid: spec.kubernetesMonitoring.imagePullPolicy: Unsupported value: "bloop": supported values: "IfNotPresent", "Always", "Never"
```

### DIFF
--- a/pkg/api/latest/dynakube/kubemon/props.go
+++ b/pkg/api/latest/dynakube/kubemon/props.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/dtversion"
-	corev1 "k8s.io/api/core/v1"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


aorcholski (approved):

svfoxat (approved):
i like this

jdanieck (approved):

0sewa0:
will close this for the extended change https://github.com/Dynatrace/dynatrace-operator/pull/7205 

thx for all the reviews ❤️

PR #7205, opened 2026-08-04
Remove the `image.PullPolicy` type

0sewa0:
## Description

Extended version of https://github.com/Dynatrace/dynatrace-operator/pull/7204

In order to save on a few `// +kubebuilder:validation:Enum=IfNotPresent;Always;Never` we introduced a bunch of getters and conversion. (and IMO confusion)

## How can this be tested?

No change in CRD ✅  so all enum validation are still present
`imagePullPolicy: bloop` -> validation should catch it. (any image pull policy in our CRs)

### DIFF
--- a/pkg/api/latest/dynakube/activegate/props.go
+++ b/pkg/api/latest/dynakube/activegate/props.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/dtversion"
-	corev1 "k8s.io/api/core/v1"
 )
 
 const (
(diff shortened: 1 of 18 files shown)


pkg/api/latest/dynakube/activegate/props.go
```diff
@@ -10,7 +10,6 @@ import (
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/dtversion"
-	corev1 "k8s.io/api/core/v1"
```
andriisoldatenko, line 13: please remove unit-tests and may be add/update integration test to demonstrate marker is working, should be easy to do. 
We can also create tech debt to track your effort.

andriisoldatenko:
@avorima please remind us why we created custom reusable type instead of using marker everywhere?

pkg/api/latest/dynakube/activegate/props.go
0sewa0: I can correct this to have a correct pull-policy, but this getter is one that make sense

avorima (dismissed):

avorima:
> @avorima please remind us why we created custom reusable type instead of using marker everywhere?

it's as marcell said, it's just a tradeoff between adding markers vs having to cast. <strike>i'm fine either way</strike> actually, using the shared package less is better

pkg/api/latest/dynakube/activegate/props.go
0sewa0: done in https://github.com/Dynatrace/dynatrace-operator/pull/7205/commits/621b15a5fdf620c4c6706ef4f962ac680b7dd940

avorima (approved):

andriisoldatenko (approved):

waodim (approved):
nice!

PR #6408, opened 2026-03-31
Add node taint analysis to support archive

averagenative:
## Summary
- Adds a new `nodeTaintAnalysisCollector` to the support archive that analyzes node taints against OneAgent and CSI DaemonSet tolerations
- For each DaemonSet, reports desired/ready pod counts vs total node count, lists configured tolerations, and highlights nodes with untolerated taints
- Helps diagnose scheduling issues where OneAgent or CSI pods are missing from nodes due to taint/toleration mismatches

## Test plan
- [x] Unit tests for all scenarios: all taints tolerated, untolerated taints, missing DaemonSets, no DynaKubes, wildcard tolerations, partial tolerations
- [x] Table-driven tests for toleration matching logic (exact match, Exists operator, wildcard, effect matching)
- [x] Full `go test ./cmd/supportarchive/` passes
- [x] `go build ./...` compiles cleanly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[JIRA](https://dt-rnd.atlassian.net/browse/ICP-4195)

PR #7207, opened 2026-08-04
Add node taint analysis collector to support archive

svfoxat:
## Description

https://dt-rnd.atlassian.net/browse/ICP-4195
taken over from: https://github.com/Dynatrace/dynatrace-operator/pull/6408

- Adds a new nodeTaintAnalysisCollector to the support archive that analyzes node taints against OneAgent and CSI DaemonSet tolerations
- For each DaemonSet, reports desired/ready pod counts vs total node count, lists configured tolerations, and highlights nodes with untolerated taints
- Helps diagnose scheduling issues where OneAgent or CSI pods are missing from nodes due to taint/toleration mismatches

Example output:
```
Node count: 3

--- OneAgent DaemonSet: gke-k8s-2026-07-27-agents-oneagent (DynaKube: gke-k8s-2026-07-27-agents) ---
  Desired: 1 | Ready: 1 | Total cluster nodes: 3
  Configured tolerations:
    node-role.kubernetes.io/master:NoSchedule
    node-role.kubernetes.io/control-plane:NoSchedule
  WARNING: 2 node(s) have untolerated taints:
    Node: gke-ninoe-dev-default-pool-1dba91cd-7zjg
      - dedicated=gpu:NoExecute
    Node: gke-ninoe-dev-default-pool-1dba91cd-853x
      - team=platform:NoSchedule

--- CSI Driver DaemonSet: dynatrace-oneagent-csi-driver ---
  Desired: 2 | Ready: 2 | Total cluster nodes: 3
  Configured tolerations:
    node-role.kubernetes.io/master:NoSchedule
    node-role.kubernetes.io/control-plane:NoSchedule
    kubernetes.io/arch=arm64:NoSchedule
    kubernetes.io/arch=amd64:NoSchedule
    kubernetes.io/arch=ppc64le:NoSchedule
    kubernetes.io/arch=s390x:NoSchedule
    ToBeDeletedByClusterAutoscaler:NoSchedule
  WARNING: 2 node(s) have untolerated taints:
    Node: gke-ninoe-dev-default-pool-1dba91cd-7zjg
      - dedicated=gpu:NoExecute
    Node: gke-ninoe-dev-default-pool-1dba91cd-853x
      - team=platform:NoSchedule
```

## How can this be tested?

- `make go/test`
- deploy to a cluster and execute the `troubleshoot` command [docs/troubleshooting](https://docs.dynatrace.com/docs/ingest-from/setup-on-k8s/deployment/troubleshooting)

### DIFF
--- a/cmd/supportarchive/node_taint_analysis.go
+++ b/cmd/supportarchive/node_taint_analysis.go
@@ -1,0 +1,273 @@
+package supportarchive
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	dtcsi "github.com/Dynatrace/dynatrace-operator/pkg/controllers/csi"
+	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const nodeTaintAnalysisCollectorName = "nodeTaintAnalysisCollector"
+
+type nodeTaintAnalysisCollector struct {
+	collectorCommon
+	ctx       context.Context
+	apiReader client.Reader
+	namespace string
+}
+
+func newNodeTaintAnalysisCollector(ctx context.Context, log logd.Logger, supportArchive archiver, namespace string, apiReader client.Reader) collector {
+	return nodeTaintAnalysisCollector{
+		collectorCommon: collectorCommon{
+			log:            log,
+			supportArchive: supportArchive,
+		},
+		ctx:       ctx,
+		apiReader: apiReader,
+		namespace: namespace,
+	}
+}
+
+func (c nodeTaintAnalysisCollector) Name() string {
+	return nodeTaintAnalysisCollectorName
+}
+
+func (c nodeTaintAnalysisCollector) Do() error {
+	logInfof(c.log, "Starting node taint analysis")
+
+	report, err := c.buildReport()
+	if err != nil {
+		logErrorf(c.log, err, "Failed to complete node taint analysis")
+
+		return err
+	}
+
+	if err := c.supportArchive.addFile(NodeTaintAnalysisFileName, strings.NewReader(report)); err != nil {
+		return err
+	}
+
+	logInfof(c.log, "Stored node taint analysis into %s", NodeTaintAnalysisFileName)
+
+	return nil
+}
+
+func (c nodeTaintAnalysisCollector) buildReport() (string, error) {
+	nodes, err := c.getNodes()
+	if err != nil {
+		return "", err
+	}
+
+	dynakubes, err := c.getDynaKubes()
+	if err != nil {
+		return "", err
+	}
+
+	var sb strings.Builder
+
+	sb.WriteString(fmt.Sprintf("Node count: %d\n\n", len(nodes.Items)))
+
+	c.analyzeOneAgentDaemonSets(&sb, nodes, dynakubes)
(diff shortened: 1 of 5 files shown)


cmd/supportarchive/node_taint_analysis.go
```diff
@@ -0,0 +1,268 @@
...
+	return false
+}
+
+func formatToleration(t corev1.Toleration) string {
+	key := t.Key
+	if key == "" {
+		key = "*"
+	}
+
+	if t.Operator == corev1.TolerationOpExists || t.Operator == "" && t.Value == "" {
```
waodim:
```suggestion
	if t.Operator == corev1.TolerationOpExists || (t.Operator == "" && t.Value == "") {
```

cmd/supportarchive/node_taint_analysis.go
```diff
@@ -0,0 +1,268 @@
...
+		node := &nodes.Items[i]
+		if len(node.Spec.Taints) == 0 {
+			continue
+		}
+
+		var untolerated []corev1.Taint
+
+		for _, taint := range node.Spec.Taints {
+			// PreferNoSchedule is advisory: DaemonSet controller schedules regardless of whether the taint is tolerated
+			if taint.Effect == corev1.TaintEffectPreferNoSchedule {
```
waodim, line 193: Should we add a test for exactly this case? I think no test covers it currently or am I wrong?

cmd/supportarchive/node_taint_analysis.go
```diff
@@ -0,0 +1,268 @@
...
+			})
+		}
+	}
+
+	return result
+}
+
+func isTaintTolerated(taint corev1.Taint, tolerations []corev1.Toleration) bool {
+	for i := range tolerations {
+		if tolerations[i].ToleratesTaint(klog.Background(), &taint, true) {
```
waodim, line 216: looking at `ToleratesTaint` i found this:

```
//  5. If enableComparisonOperators is false and the toleration uses 'Lt' or 'Gt'
//     operators, the toleration does not match (returns false).
```

https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#numeric-comparison-operators

So 2 things:

1. Maybe add a little comment what this param is like `true /*enableComparisonOperators*/` or so, IMO this would help understanding what is happening
2. Since the param is true `Lt` and `Gt` operators are valid, but IMO `formatToleration` doesn't deal with them. Let's say you have a toleration with `operator: Lt` and a `value: "25"` -> this would show as `cpu-score=25:NoSchedule` no? So it would interpret it as equal which could lead to confusion. Or am I wrong?

svfoxat: Good catch!

I added the `Lt` and `Gt` output handling, so it will print like this:
```
Configured tolerations:
  cpu-score<25:NoSchedule
  memory-pressure>100:NoExecute
  kubernetes.io/arch=amd64:NoSchedule
```
Also I added the comment to clarify the `true` in the function call.

waodim (approved):
Tested Lt/Gt tolerations on my gke cluster: the k8s api server rejects them at admission time (confirmed I tried `--dry-run=server` failing while `--dry-run=client` succeeds, so it's not gke only I guess). 

IMO we should keep it nevertheless, since it does not hurt and covers a case that can come into play.

svfoxat:
> Tested Lt/Gt tolerations on my gke cluster: the k8s api server rejects them at admission time (confirmed I tried `--dry-run=server` failing while `--dry-run=client` succeeds, so it's not gke only I guess).
> 
> IMO we should keep it nevertheless, since it does not hurt and covers a case that can come into play.

the `TaintTolerationComparisonOperators` feature gate is disabled by default since k8s 1.35

PR #7209, opened 2026-08-04
Validate namespace selectors

avorima:
## Description

I was making some changes in the context of fixing ICP-8342 and noticed that we don't properly validate all namespace selectors.
Do not print the exact error messages to the user, because we're mixing two concepts (field.ErrorList and our custom weird only one message stuff). We need to address that at some point.

## How can this be tested?

Deploy the operator and try to create DynaKubes with invalid namespace label selectors


0sewa0 (dismissed):

pkg/api/validation/dynakube/namespace_selector.go
```diff
@@ -42,11 +44,65 @@ func conflictingNamespaceSelector(ctx context.Context, dv *Validator, dk *dynaku
...
-	errs := validation.ValidateLabelSelector(dk.OneAgent().GetNamespaceSelector(), validation.LabelSelectorValidationOptions{AllowInvalidLabelValueInSelector: false}, field.NewPath("spec", "namespaceSelector"))
+func invalidOneAgentNamespaceSelector(ctx context.Context, _ *Validator, dk *dynakube.DynaKube) string {
+	var (
+		oneAgentSelector *metav1.LabelSelector
+		oneAgentMode     string
+	)
+
+	switch oa := dk.OneAgent(); {
+	case oa.IsCloudNativeFullstackMode():
+		oneAgentSelector = &oa.CloudNativeFullStack.NamespaceSelector
```
waodim, line 55: Why do we use a getter for the `metadataEnrichment` and for `otlp` but not for the `oneAgent` namespace selectors?

avorima: i was already doing the switch for the mode, so i added the extra lines for the selector. metadata doesn't actually need a getter

waodim (approved):

PR #7212, opened 2026-08-05
Enable client caching for OA/AG connection-info

0sewa0:
## Description

https://dt-rnd.atlassian.net/browse/ICP-7301

No longer use conditions to gate connection-info requests, but use the client's cache instead.

- (partly) Moved the networkzone IP check into the OA client, to invalidate the cache.

## How can this be tested?

No new behavior.
- The networkzone check should still work, and not cause issues. (set the `networkZone` field)

Check in the logs, that you are hitting the cache instead of making an extra call for the connection-info.


pkg/clients/dynatrace/oneagent/connection_info_test.go
```diff
@@ -73,130 +73,103 @@ func Test_GetConnectionInfo(t *testing.T) {
...
-			TenantToken: testTenantToken,
-			Endpoints:   testCommunicationEndpoint,
-		}
-
-		oaClient := setupMockedClient(t, map[string]string{}, "", dupResponse, nil)
-		connectionInfo, err := oaClient.GetConnectionInfo(ctx)
-		require.NoError(t, err)
-
-		assert.Equal(t, expected, connectionInfo)
+	t.Run("required IPsS missing from response → MissingIPs error", func(t *testing.T) {
```
svfoxat:
```suggestion
	t.Run("required IPs missing from response → MissingIPs error", func(t *testing.T) {
```

pkg/controllers/dynakube/connectioninfo/oneagent/reconciler.go
svfoxat: `timeProvider`is now only used in `NewReconciler`, I think we can remove it. (or do we use it in tests?)

pkg/controllers/dynakube/connectioninfo/oneagent/cache_test.go
```diff
@@ -0,0 +1,130 @@
...
+			connectionInfoBody(testTenantUUID, testTenantToken, testTenantEndpoints),
+			connectionInfoBody(updatedUUID, testTenantToken, testTenantEndpoints),
+		)
+		oaClient := newConnectionInfoOAClient(t, transport, testConnectionInfoCacheTTL)
+		dk := getTestDynakube()
+		fakeClient := fake.NewClient(dk)
+		r := NewReconciler(fakeClient, fakeClient)
+
+		synctest.Test(t, func(t *testing.T) {
+			require.NoError(t, r.Reconcile(context.Background(), oaClient, dk))
```
svfoxat:
```suggestion
			require.NoError(t, r.Reconcile(t.Context(), oaClient, dk))
```

pkg/controllers/dynakube/connectioninfo/oneagent/cache_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+			time.Sleep(testConnectionInfoCacheTTL + time.Second)
+
+			require.NoError(t, r.Reconcile(context.Background(), oaClient, dk))
```
svfoxat:
```suggestion
			require.NoError(t, r.Reconcile(t.Context(), oaClient, dk))
```

pkg/clients/dynatrace/oneagent/connection_info.go
```diff
@@ -16,27 +18,51 @@ const (
 	connectionInfoPath = "/v1/deployment/installer/agent/connectioninfo"
 )
 
-// connectionInfoResponse is the raw shape returned by the connectioninfo API.
-// It is unmarshalled directly from JSON and never exposed outside this package.
-type connectionInfoResponse struct {
-	TenantUUID             string   `json:"tenantUUID"`
-	TenantToken            string   `json:"tenantToken"`
-	CommunicationEndpoints []string `json:"communicationEndpoints"`
+// MissingIPs is returned by GetConnectionInfo when provided required IPs are absent
```
avorima: shouldn't this be MissingHosts?

0sewa0: `MissingIP` is my ~4. name for this. 😅 
`MissingHosts` was the ~3.

while it is true, that it checks the `host` part of the URL, but we always give IPs to this. 🤷 

I will revert it to `MissingHosts`, it is more correct in the context of the client.

pkg/controllers/dynakube/connectioninfo/oneagent/reconciler.go
```diff
@@ -82,29 +81,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, oaClient oneagent.Client, dk
...
-			return nil
-		}
-	}
-
-	k8sconditions.SetSecretOutdated(dk.Conditions(), oaConnectionInfoConditionType, secretNamespacedName.Name+" is not present or outdated, update in progress") // Necessary to update the LastTransitionTime, also it is a nice failsafe
-
-	connectionInfo, err := oaClient.GetConnectionInfo(ctx)
-	if err != nil {
+	connectionInfo, err := oaClient.GetConnectionInfo(ctx, getRequiredEndpoints(dk))
+	if err != nil && !errors.As(err, new(oneagent.MissingIPs)) {
```
waodim: IMO would make sense to store the error into a var since we use it down in the other if again

pkg/clients/dynatrace/oneagent/connection_info.go
0sewa0: I just moved/reused already existing errors from the oa connection info reconciler instead of creating another one.

pkg/controllers/dynakube/connectioninfo/activegate/cache_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+			require.NoError(t, r.Reconcile(t.Context(), agClient, dk))
+		})
+
+		assert.Equal(t, updatedUUID, dk.Status.ActiveGate.ConnectionInfo.TenantUUID)
+		transport.assertCalls(t, 2)
+	})
+}
+
+// --- Fake transport ----------------------------------------------------------
```
andriisoldatenko: why we need this line? functions already holds name: `fakeCITransport`

pkg/controllers/dynakube/connectioninfo/activegate/cache_test.go
```diff
@@ -0,0 +1,128 @@
...
+		Request:    r,
+	}, nil
+}
+
+// --- Client constructor ------------------------------------------------------
+
+// newConnectionInfoAGClient builds a real agclient.ClientImpl whose HTTP transport is
+// the given fakeCITransport wrapped in a cache round-tripper with the specified TTL.
+// PaasToken is set to t.Name() so each subtest has an isolated namespace in the global cache.
+func newConnectionInfoAGClient(t *testing.T, transport http.RoundTripper, ttl time.Duration) agclient.Client {
```
andriisoldatenko, line 93: same here, Client constructor? also i'm not sure we need this doc string ...

andriisoldatenko: it's just repeasts what's inside test helper/test factory function

pkg/controllers/dynakube/connectioninfo/oneagent/cache_test.go
```diff
@@ -0,0 +1,129 @@
+// Copyright Dynatrace LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package oaconnectioninfo
+
+// Tests for the HTTP-level response cache used by the oneagent connection-info client.
+//
+// The reconciler no longer performs its own timestamp-based caching: every Reconcile call
```
andriisoldatenko: sounds like wrong place for this doc string, better to keep it close to impl not to code which is testing it.

pkg/controllers/dynakube/connectioninfo/oneagent/cache_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+			require.NoError(t, r.Reconcile(t.Context(), oaClient, dk))
+		})
+
+		assert.Equal(t, updatedUUID, dk.Status.OneAgent.ConnectionInfo.TenantUUID)
+		transport.assertCalls(t, 2)
+	})
+}
+
+// --- Fake transport ----------------------------------------------------------
```
andriisoldatenko: same

pkg/clients/dynatrace/oneagent/connection_info.go
```diff
@@ -16,14 +18,24 @@ const (
...
+// It implements core.Cacheable so that Execute opts the request into the HTTP-layer cache.
 type connectionInfoResponse struct {
 	TenantUUID             string   `json:"tenantUUID"`
 	TenantToken            string   `json:"tenantToken"`
 	CommunicationEndpoints []string `json:"communicationEndpoints"`
 }
 
+func (r *connectionInfoResponse) IsEmpty() bool {
+	return r.TenantUUID == "" || r.TenantToken == "" || len(r.CommunicationEndpoints) == 0
+}
```
0sewa0, line 96: this is not correct, this has to handle the `requiredHosts`, because otherwise it doesn't invalidate the cache.

❤️  @aorcholski for finding it

aorcholski (commented):
Dynakube:
- AG routing
- CNFS
- network zone with fallback
- no LogMonitoring
---

OA PODs are deployed after 15min. Previously some external endpoints could be reachable immediately and deployment was successful  after 2-3 min

---

LogMonitoring and OneAgent use the same `oaconnectioninfo.NewReconciler` 
```
{"level":"info","ts":"2026-08-07T10:28:35.632Z","logger":"dynakube.logmonitoring.connectioninfo","msg":"OneAgent endpoints do not contain the local ActiveGate Service IP yet, postponing OneAgent deployment",
```
we need to remember to grep 
```
"dynakube.oneagent.connectioninfo.dtclient-oneagent\|dynakube.logmonitoring.connectioninfo.dtclien-oneagent"
```
to get full picture of the endpoint status.

"dynakube.logmonitoring.connectioninfo" <-> "postponing OneAgent deployment"

pkg/controllers/dynakube/connectioninfo/activegate/cache_test.go
0sewa0: removed all the redundant doc strings

pkg/clients/dynatrace/oneagent/connection_info.go
```diff
@@ -16,12 +19,70 @@ const (
...
 type connectionInfoResponse struct {
 	TenantUUID             string   `json:"tenantUUID"`
 	TenantToken            string   `json:"tenantToken"`
 	CommunicationEndpoints []string `json:"communicationEndpoints"`
+
+	dedupOnce sync.Once
+
+	requiredHosts []string
+	validateOnce  sync.Once
+	missing       bool
```
0sewa0, line 41: FYI:

I have implemented once with only using 1 `sync.Once` where I loop over the endpoints only once, and check/do all the things in 1 go.

I thought this was more readable, and looping over a few hundred entries twice should still be quick

pkg/controllers/dynakube/connectioninfo/oneagent/cache_test.go
```diff
@@ -0,0 +1,200 @@
...
+		// it may look suboptimal, but the likelihood of this scenario is very low, and it will fix it self as the error is returned correctly.
+		_, err = oaClient.GetConnectionInfo(t.Context(), []string{newHost})
+		require.ErrorIs(t, err, oneagentclient.StaleNetworkZoneEndpointsError)
+		transport.assertCalls(t, 1, "second call must be served from cache before invalidation")
+
+		// cache was invalidated, so a new fetch occurs, now no error
+		_, err = oaClient.GetConnectionInfo(t.Context(), []string{newHost})
+		require.NoError(t, err)
+		transport.assertCalls(t, 2, "cache must be invalidated after required host change")
+	})
```
0sewa0, line 131: important edgecase based on the current limitation of our caching logic

pkg/clients/dynatrace/oneagent/connection_info.go
```diff
@@ -16,12 +19,70 @@ const (
...
+			endpoint = strings.TrimSpace(endpoint)
+			if endpoint == "" {
+				continue
+			}
+
+			if _, ok := seen[endpoint]; ok {
+				continue
+			}
+
+			seen[endpoint] = struct{}{}
```
andriisoldatenko: i know it's well known trick, but i'm not sure you will gain a lot

what i would do (for readability purpose) : `seen[endpoint] = true

and then you can do:

```
if seen[endpoint] {
				continue
			}
```
why? because zero value for bool is false :)

pkg/clients/dynatrace/oneagent/connection_info.go
```diff
@@ -16,12 +19,70 @@ const (
...
+	missing       bool
+}
+
+func (r *connectionInfoResponse) IsEmpty() bool {
+	return r.TenantUUID == "" || r.TenantToken == "" || len(r.uniqueEndpoints()) == 0 || r.missingRequiredHosts()
+}
+
+func (r *connectionInfoResponse) uniqueEndpoints() []string {
+	r.dedupOnce.Do(func() {
+		seen := make(map[string]struct{})
```
andriisoldatenko:
```suggestion
		seen := make(map[string]bool)
```

pkg/clients/dynatrace/oneagent/connection_info.go
```diff
@@ -16,12 +19,70 @@ const (
...
+		}
+
+		r.CommunicationEndpoints = uniqueEndpoints
+	})
+
+	return r.CommunicationEndpoints
+}
+
+func (r *connectionInfoResponse) missingRequiredHosts() bool {
+	r.validateOnce.Do(func() {
```
andriisoldatenko, line 83: thats smart :)

pkg/clients/dynatrace/oneagent/connection_info.go
```diff
@@ -59,35 +122,17 @@ func (c *ClientImpl) GetConnectionInfo(ctx context.Context) (ConnectionInfo, err
...
-			continue
-		}
-
-		if _, ok := seen[endpoint]; ok {
-			continue
-		}
-
-		seen[endpoint] = struct{}{}
-		unique = append(unique, endpoint)
+		Endpoints:   strings.Join(resp.uniqueEndpoints(), ";"),
```
andriisoldatenko: can we just move ";" to constatin in top level of function to add meaning to it, i know it's OA specific.

const oneAgentDefaultSeperator = ";"

pkg/clients/dynatrace/oneagent/connection_info.go
0sewa0: that is just copied code :D 

but I also prefer `bool`, will fix

andriisoldatenko (dismissed):
tested using e2e tests

pkg/controllers/dynakube/connectioninfo/oneagent/reconciler.go
0sewa0: fixed

andriisoldatenko (approved):

PR #7215, opened 2026-08-05
Add custom annotations to the webhook service

NigelVanHattum:
The dynatrace-webhook Service had a hardcoded metadata block, so users could not attach annotations to it. This blocks common setups that are driven by Service annotations, such as internal load balancers, service mesh opt-out and Prometheus scrape configuration.

Apply the existing webhook.annotations value to the Service metadata, following the same pattern webhook.labels and webhook.annotations already use on the webhook Deployment and its pod template, so no new values key is introduced.

Add helm-unittest cases covering the annotated and the default (absent) case.

### DIFF
--- a/config/helm/chart/default/templates/Common/webhook/service.yaml
+++ b/config/helm/chart/default/templates/Common/webhook/service.yaml
@@ -5,6 +5,10 @@ kind: Service
 metadata:
   name: dynatrace-webhook
   namespace: {{ .Release.Namespace }}
+  {{- if .Values.webhook.annotations }}
+  annotations:
+    {{- toYaml .Values.webhook.annotations | nindent 4 }}
+  {{- end }}
   labels:
   {{- include "dynatrace-operator.webhookLabels" . | nindent 4 }}
 spec:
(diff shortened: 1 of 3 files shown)


avorima (approved):
👍

VLazarevic (approved):
lgtm

NigelVanHattum:
When can we expect a new release that includes these changes, @svfoxat ?

VLazarevic:
> When can we expect a new release that includes these changes, @svfoxat ?
@NigelVanHattum 
I am not svfoxat, but the next release(1.11.0) is planned to be end of september.

NigelVanHattum:
That's still a long wait. Is it possible to release a 1.10.3 version. We do require these annotations in order to deploy the operator.

VLazarevic:
> That's still a long wait. Is it possible to release a 1.10.3 version. We do require these annotations in order to deploy the operator.

I don't think this is possible, but I quoted the important people :)
@andriisoldatenko @0sewa0 ?

0sewa0:
> That's still a long wait. Is it possible to release a 1.10.3 version. We do require these annotations in order to deploy the operator.

Doing a release is not that simple, comes with a lot of fanfare and time investment.
- If there will be a `v1.10.3` due to some high-profile bugs/CVEs, then this could be included aswell.

Channels (official support and such) possibly exits to escalate such requests, but that channel is not Github 😅.

PR #7218, opened 2026-08-06
Replicate secrets according to namespace selectors

avorima:
## Description

Only replicate OTLP config secrets in namespaces were OTLP injection is enabled. This reverts a behavior introduced in #5491.

OA/ME secret is shared, so both selectors are ORed. OTLP secret can now be properly separated from OA/ME secret.

ICP-8342

## How can this be tested?

Try different combinations of oneAgent/metadataEnrichment and OTLP auto config

* No label selector -> secrets are replicated to all namespaces
* OA label selector -> secrets are replicated to all namespaces
* ME label selector -> secrets are replicated to all namespaces
* OA+ME label selector -> init container secret replicated to selected namespaces, OTLP everywhere
* OTLP label selector -> OTLP secrets are replicated to selected namespaces, OA/ME everywhere
* OA+ME+OTLP -> all secrets are only replicated in their selected namespaces
* Update label selectors -> secrets are cleaned up if their are in a namespace that is no longer selected


pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -280,3 +304,48 @@ func (r *Reconciler) cleanupOTLPSecret(ctx context.Context, namespaces []corev1.
...
+		if labels.MatchesNothing(selector) {
+			continue
+		}
+
+		selectors = append(selectors, selector)
+	}
+
+	var result []corev1.Namespace
+
+	if len(selectors) > 0 {
```
andriisoldatenko: why not re-write 
```
for _, selector := range selectors {
    for _, ns := range namespaces {
          if selector.Matches(labels.Set(ns.Labels)) {
					result = append(result, ns)

					break // continue with the next namespace
				}
```
the nyou don't need if len(selectors) > 0 {

or am i miss something?

andriisoldatenko: yeah break comment is wrong, but it more or less same code ( selectors * namespaces operations)

avorima: `selectors * namespaces` needs to handle iterating over namespaces twice without adding duplicates

pkg/api/latest/dynakube/oneagent/props.go
```diff
@@ -128,6 +128,7 @@ func (oa *OneAgent) GetInitResources() *corev1.ResourceRequirements {
 	return nil
 }
 
+// GetNamespaceSelector returns the namespace label selector when application monitoring or cloud-native fullstack is enabled, otherwise nil.
```
andriisoldatenko: what is the purpose of such comment? just replicates implementation which is already trivial.

avorima: I changed the implementation of the metadata enrichment getter, so the comment here documents the intent to keep the behavior the same

andriisoldatenko: I usually prefer comments on focus on explaining the "why" rather than the "what" code is doing.

0sewa0:
```suggestion
// GetNamespaceSelector returns the namespace label selector when application monitoring or cloud-native fullstack is enabled, otherwise `nil` as it has no meaning for classic-fullstack / host-monitoring.
```

better?

0sewa0: or 
```suggestion
// GetNamespaceSelector returns the namespace label selector only when it has relevance for the active OneAgent mode.
```

but you could put this on almost all `Get*` in the `oneagent` package 😅

andriisoldatenko: yes it's better, i'm not against comments, i just prefer to have them with some value

pkg/controllers/dynakube/injection/reconciler.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/status"
 	"github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace"
 	dtimage "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/image"
-	oaClient "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/oneagent"
+	dtoneagent "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/oneagent"
```
0sewa0, line 14: sidenote: as we don't have to fix this here (anything is better than `oaClient`), but

we should standardize this, I see a lot of `oneagentclient` (also used one myself recently)
also `tokenclient`, `agclient` 

there are a few with `dt` prefix, like `dtimage` and `dtversion` (eventhough this conflicts with an actual [util package](https://github.com/Dynatrace/dynatrace-operator/blob/9dbc769a824af261dbda7faf7bcecbd21c5fa67d/pkg/util/dtversion/dtversion.go#L11) 😅 )

**Do we have a ticket to go "all in" on the `dt` prefix for client aliases?**

Note: the `client` suffix aliases are kinda bad due to stuttering. example: `oneagentclient.Client` and `oneagentclient.NewClient`

pkg/injection/namespace/mapper/dynakubes.go
```diff
@@ -28,27 +29,39 @@ type DynakubeMapper struct {
...
+	flagOneAgent matchFlags = 1 << iota
+	flagMetadata
+	flagOTLP
+)
+
+func (f matchFlags) isAny() bool       { return f > 0 }
+func (f matchFlags) isOneAgent() bool  { return f&flagOneAgent != 0 }
+func (f matchFlags) isMetadata() bool  { return f&flagMetadata != 0 }
+func (f matchFlags) isBootstrap() bool { return f.isOneAgent() || f.isMetadata() }
+func (f matchFlags) isOTLP() bool      { return f&flagOTLP != 0 }
```
0sewa0, line 52: 🧠

0sewa0 (dismissed):
LGTM,

Image

also played around with the selector in the DK, so to check the delete logic, saw no issues.

the namespaces:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  labels:
    injected/by: hvs0985d
  name: oa-only
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    enriched/by: hvs0985d
  name: meta-only
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    configured/by: hvs0985d
  name: otlp-only
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    injected/by: hvs0985d
    enriched/by: hvs0985d
  name: oa-meta
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    enriched/by: hvs0985d
    configured/by: hvs0985d
  name: meta-otlp
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    injected/by: hvs0985d
    configured/by: hvs0985d
  name: otlp-oa
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    injected/by: hvs0985d
    configured/by: hvs0985d
    enriched/by: hvs0985d
  name: otlp-oa-meta
```

the dynakube:
```yaml
kind: DynaKube
metadata:
  name: marcell-dev
  namespace: dynatrace
spec:
  apiUrl: https://hvs0985d.dev.dynatracelabs.com/api
  metadataEnrichment:
    enabled: true
    namespaceSelector:
      matchLabels:
        enriched/by: hvs0985d
  oneAgent:
    applicationMonitoring:
      namespaceSelector:
        matchLabels:
          injected/by: hvs0985d
  otlpExporterConfiguration:
    namespaceSelector:
      matchLabels:
        configured/by: hvs0985d
    signals:
      logs: {}
      metrics: {}
      traces: {}
```

pkg/controllers/dynakube/injection/reconciler.go
avorima: way ahead of you :) ICP-8480

pkg/api/latest/dynakube/oneagent/props.go
avorima: i removed them

0sewa0 (dismissed):

0sewa0 (approved):

PR #7231, opened 2026-08-10
add rbac objects for prometheus to helm chart

svfoxat:
## Description

https://dt-rnd.atlassian.net/browse/ICP-7250

- Added RBAC objects required by the Target Allocator, OTel Collector (scraper), and OTel Collector (gateway). 
- Gated creation on `.Values.rbac.prometheus.create` 
- Added Helm unit tests.

## How can this be tested?

`make test/helm/unit`

```Bash
helm template config/helm/chart/default/ --namespace dynatrace \
  --set rbac.prometheus.create=false \
  | grep -E "^  name: dynatrace-(target-allocator|prometheus-scraper|prometheus-gateway)" | wc -l
# should be 0
```
```Bash
 helm template config/helm/chart/default/ --namespace dynatrace \
  --set rbac.prometheus.create=true \
  | grep -E "^  name: dynatrace-(target-allocator|prometheus-scraper|prometheus-gateway)" | wc -l
 # should not be 0
```

### DIFF
--- a/config/helm/chart/default/templates/Common/prometheus/clusterrole-dtprometheus-operator.yaml
+++ b/config/helm/chart/default/templates/Common/prometheus/clusterrole-dtprometheus-operator.yaml
@@ -1,6 +1,7 @@
 # Copyright Dynatrace LLC
 # SPDX-License-Identifier: Apache-2.0
 
+{{- if .Values.rbac.prometheus.create }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
@@ -33,3 +34,4 @@ roleRef:
   kind: ClusterRole
   name: dynatrace-dtprometheus-operator
   apiGroup: rbac.authorization.k8s.io
+{{- end }}
--- a/config/helm/chart/default/templates/Common/prometheus/clusterrole-dtprometheus-operator.yaml
+++ b/config/helm/chart/default/templates/Common/prometheus/clusterrole-dtprometheus-operator.yaml
@@ -1,7 +1,6 @@
+{{- if .Values.rbac.prometheus.create }}
 # Copyright Dynatrace LLC
 # SPDX-License-Identifier: Apache-2.0
-
-{{- if .Values.rbac.prometheus.create }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
--- a/config/helm/chart/default/templates/Common/prometheus/clusterrole-prometheus-gateway.yaml
+++ b/config/helm/chart/default/templates/Common/prometheus/clusterrole-prometheus-gateway.yaml
@@ -1,7 +1,6 @@
+{{- if .Values.rbac.prometheus.create }}
 # Copyright Dynatrace LLC
 # SPDX-License-Identifier: Apache-2.0
-
-{{- if .Values.rbac.prometheus.create }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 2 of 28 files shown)


andriisoldatenko (commented):
code LGTM, hard to test without reconciler, 

cc @avorima can you test it? or lets merge and you can rebase?

avorima:
Needs dtprometheuses/status patch
```
dtprometheuses.dynatrace.com \"test\" is forbidden: User \"system:serviceaccount:dynatrace:dynatrace-operator\" cannot patch resource \"dtprometheuses/status\" in API group \"dynatrace.com\" in the namespace \"dynatrace\"
```
You can actually remove any other write verbs from the RBAC. the operator should not write the spec. status can stay update/patch in case we switch

config/helm/chart/default/templates/Common/prometheus/role-dtprometheus-operator.yaml
```diff
@@ -17,13 +18,18 @@ rules:
       - get
       - list
       - watch
-      - update
```
avorima, line 20: sorry, i forgot about finalizers because we don't use them currently. please keep update here in case we decide otherwise

svfoxat: the `update` for `dtprometheuses/finalizers` is [here](https://github.com/Dynatrace/dynatrace-operator/pull/7231/changes/BASE..ec8e32555c34ddb2b3dc04ad739cd9c22c7a6ab6#diff-c75440faf42f95099929bc805dce0d08d87321087cb18775a8c8a8b0a293e027R24) 

```
kubectl auth can-i update dtprometheuses.dynatrace.com \
  --subresource=finalizers \
  --as=system:serviceaccount:dynatrace:dynatrace-operator \
  -n dynatrace

yes
```

or do you mean something else?

config/helm/chart/default/templates/Common/prometheus/clusterrole-dtprometheus-operator.yaml
```diff
@@ -1,6 +1,7 @@
 # Copyright Dynatrace LLC
 # SPDX-License-Identifier: Apache-2.0
 
+{{- if .Values.rbac.prometheus.create }}
```
andriisoldatenko: we usually do check before copyright, so helm won't render a file, otherwise it can complain about empty file

config/helm/chart/default/templates/Common/prometheus/clusterrole-prometheus-gateway.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright Dynatrace LLC
+# SPDX-License-Identifier: Apache-2.0
+
+{{- if .Values.rbac.prometheus.create }}
```
andriisoldatenko: example https://github.com/dynatrace/dynatrace-operator/blob/2a39d88a0ee1fbb61e2d22db02520b3dc92ffc80/config/helm/chart/default/templates/Common/edge-connect/clusterrole-edgeconnect.yaml#L1

config/helm/chart/default/templates/Common/prometheus/role-dtprometheus-operator.yaml
avorima: the finalizers subresource is not used when we're setting the finalizer on the dtprom. it's actually used for the resources the dtprom owns. so if we set the finalizer on the dtprom it's counted as a normal update/patch

svfoxat: [`8a1d40c` (this PR)](https://github.com/Dynatrace/dynatrace-operator/pull/7231/changes/8a1d40c8080ab1b915961558b69a5dfb1b7244ad)

avorima (approved):

PR #7235, opened 2026-08-11
Add dtprometheus reconciler

avorima:
## Description

Implement the base DTPrometheus reconciler with some changes for the follow-up TargetAllocator reconciler.
The PoC was built with both so some changes in this PR were made with this in mind, namely:

* Remove updateTimestamp. This field was copied over from the DynaKube, but the same information can be gauged from the conditions (once they are implemented). For effective use it also requires that the reconciler figures out if the status changed.
* Change duration field types. Change the field types from string to metav1.Duration. The schema remains unchanged, but in go code we can use the duration without having to parse it first.
* Add condition reasons. These are mandatory for each subreconciler to use in order for the phase detection to work.

What this actually provides:

* Base reconcile scaffolding for DTPrometheus with 404 checks and gating whether the required DynaKube is ready
* Build a (yet to be used) dynatrace API client
* Phase calculation and status updating using SSA. This uses a deprecated method from controller-runtime, because the repository needs some work to allow auto-generating the applyconfiguration. If we don't end up supporting it before this method is completely removed we can fall back to normal Update or merge patch.

ICP-7433

## How can this be tested?

* Deploy the operator and create DTPrometheus resources.

You will need this RBAC:

<details><summary>rbac.yaml</summary>

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: dynatrace-prometheus-operator-temp
rules:
  - apiGroups: ["dynatrace.com"]
    resources: ["dtprometheuses/status"]
    verbs: ["patch"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: dynatrace-prometheus-operator-temp
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: dynatrace-prometheus-operator-temp
subjects:
  - kind: ServiceAccount
    name: dynatrace-operator
    namespace: dynatrace
```

</details>

You can check the phase transitions by using `kubectl patch`:

* `kubectl patch --subresource status -n dynatrace dtp test --type merge -p '{"status":{"conditions":[{"type":"Test","status":"False","reason":"Test","message":"Test","lastTransitionTime":"2026-08-11T13:39:00Z"}]}}'`
* `kubectl patch --subresource status -n dynatrace dtp test --type merge -p '{"status":{"conditions":[{"type":"Test","status":"Unknown","reason":"Test","message":"Test","lastTransitionTime":"2026-08-11T13:39:00Z"}]}}'`
* `kubectl patch --subresource status -n dynatrace dtp test --type merge -p '{"status":{"conditions":[{"type":"Test","status":"True","reason":"Test","message":"Test","lastTransitionTime":"2026-08-11T13:39:00Z"}]}}'`


pkg/controllers/dtprometheus/reconciler.go
```diff
@@ -0,0 +1,280 @@
...
+
+type Reconciler struct {
+	client.Client
+
+	newDynatraceClient dynatrace.ClientFactory
+}
+
+// +kubebuilder:rbac:groups=dtprometheuses.dynatrace.com,resources=dtprometheuses,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=dtprometheuses.dynatrace.com,resources=dtprometheuses/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=dtprometheuses.dynatrace.com,resources=dtprometheuses/finalizers,verbs=update
```
andriisoldatenko: do we want this? or it's just for testing

pkg/controllers/dtprometheus/reconciler.go
jokuniew, line 1: rename to just `reconciler.go`?
https://github.com/Dynatrace/dynatrace-operator/blob/main/doc/coding-style-guide.md?plain=1#L43

pkg/controllers/dtprometheus/reconciler.go
avorima: we're not using it now, but i think we should down the line. it's basically the reconciler's interface definition

pkg/controllers/dtprometheus/dtprometheus_reconciler.go
```diff
@@ -0,0 +1,280 @@
...
+// +kubebuilder:rbac:groups=dtprometheuses.dynatrace.com,resources=dtprometheuses/finalizers,verbs=update
+
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the Frigate object against the actual cluster state, and then
```
jokuniew:
```suggestion
// the DTPrometheus object against the actual cluster state, and then
```

pkg/controllers/dtprometheus/reconciler.go
andriisoldatenko: Yeah I know just asking. I'm fine to use it

pkg/controllers/dtprometheus/reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+var (
+	errMissingDynaKube  = errors.New("missing dynakube")
```
jokuniew: those errors are not returned anywhere in reconciler

pkg/controllers/dtprometheus/reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+		},
+		DeleteFunc: func(event.TypedDeleteEvent[client.Object]) bool {
+			return false
+		},
+		UpdateFunc: func(e event.TypedUpdateEvent[client.Object]) bool {
+			oldDK, _ := e.ObjectOld.(*dynakube.DynaKube)
+
+			newDK, _ := e.ObjectNew.(*dynakube.DynaKube)
+			if oldDK == nil || newDK == nil {
+				panic(fmt.Sprintf("unexpected objects: %T, %T", e.ObjectOld, e.ObjectNew))
```
jokuniew: should we really panic here, or log.Error like in `newDTPrometheusFromDynaKubeMapper`?

pkg/controllers/dtprometheus/reconciler.go
```diff
@@ -0,0 +1,263 @@
...
+		dk,
+		tokens.APIToken().Value,
+		tokens.PaasToken().Value,
+		"dtprometheus",
+		k8senv.GetOperatorDTClientConnectionTimeout(ctx),
+	)
+}
+
+func setPhase(dtp *dtprometheus.DTPrometheus, err error) error {
+	if errors.Is(err, errMissingDynaKube) || errors.Is(err, errDynaKubeNotReady) {
```
jokuniew: It is a bit unclear to me
If dynakube is not ready -> it can be ready within some time -> dt prometheus is Deploying, ok

but if dynakube is missing? I assume prometheus is not doing any more progress
doesn't status.Error fit in `errMissingDynaKube` better?

pkg/controllers/dtprometheus/reconciler.go
avorima: panic is the correct approach, but we don't like that so i'll change it

pkg/controllers/dtprometheus/reconciler.go
avorima: it's not an error, because resources can be created out of order

jokuniew (approved):

svfoxat (approved):

andriisoldatenko (approved):

PR #7237, opened 2026-08-11
Implement target allocator reconciler

avorima:
## Description

Implement the target allocator reconciliation with some caveats:

* Missing TLS. This will be done later
* Missing image lookup. The image API is not ready, so an image is required for now
* The targetallocator.yaml config cannot contain user-specified values (will be done in a follow-up)

I added schema defaulting for DTPrometheus, so that it always populates the label selectors and scrape interval.
This obsoletes the getter for the label selector.

ICP-8189

## How can this be tested?

Deploy a DTPrometheus resource and check that the components are deployed correctly.
Try to modify objects and check that they get restored.


pkg/controllers/dtprometheus/targetallocator/reconciler.go
```diff
@@ -0,0 +1,425 @@
...
+	return nil
+}
+
+func (r *Reconciler) reconcileDeployment(ctx context.Context, s *reconcileScope) error {
+	log := logd.FromContext(ctx)
+	log.Debug("reconciling deployment")
+
+	if s.Spec.Image == "" {
+		// TODO: fix this once the target allocator images are available
+		// imageURI, err := registry.ResolveImage(ctx, s.Image, s.Owner.Spec.PublicRegistryOverride, image.TargetAllocator)
```
andriisoldatenko, line 216: can we link ticket, and update ticket so we do it later?

pkg/controllers/dtprometheus/targetallocator/reconciler.go
```diff
@@ -0,0 +1,425 @@
...
+		cm.Data = map[string]string{configFile: string(data)}
+
+		return controllerutil.SetControllerReference(s.Owner, cm, r.Scheme())
+	})
+	if err != nil {
+		return fmt.Errorf("reconcile configmap: %w", err)
+	}
+
+	checksum := sha256.Sum256(data)
+	s.ConfigMapHash = hex.EncodeToString(checksum[:])
```
andriisoldatenko, line 198: q why do we set Hash after crete or update?

pkg/controllers/dtprometheus/targetallocator/reconciler.go
avorima: we're a bit inconsistent with this, but in general we don't use ticket numbers for TODOs. this is an active area of development, so we'll have to change this soon anyways

pkg/controllers/dtprometheus/targetallocator/reconciler.go
```diff
@@ -0,0 +1,425 @@
...
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context, dtp *dtprometheus.DTPrometheus, dk *dynakube.DynaKube, imageClient image.Client) error {
+	ctx, _ = logd.NewFromContext(ctx, "targetallocator")
+
+	scope := &reconcileScope{
+		Owner:       dtp,
+		DynaKube:    dk,
+		Spec:        dtp.TargetAllocator(),
+		AppLabels:   k8slabel.NewAppLabels("opentelementry-target-allocator", dtp.Name, "otel-allocator", ""),
```
svfoxat: typo

```suggestion
		AppLabels:   k8slabel.NewAppLabels("opentelemetry-target-allocator", dtp.Name, "otel-allocator", ""),
```

pkg/controllers/dtprometheus/targetallocator/reconciler.go
avorima: don't need to compute the hash if we're not going to reach the deployment. i also considered making the computation only happen if we actually got a change. not too sure if it's worth it. wdyt?

pkg/controllers/dtprometheus/targetallocator/reconciler.go
```diff
@@ -0,0 +1,425 @@
...
+}
+
+func (r *Reconciler) reconcileCondition(s *reconcileScope, err error) {
+	condition := metav1.Condition{
+		Type: dtprometheus.TargetAllocatorAvailable,
+	}
+
+	switch {
+	case err != nil:
+		condition.Status = metav1.ConditionTrue
```
svfoxat: Shouldn't this be `ConditionFalse` on error?

pkg/api/v1alpha1/dtprometheus/targetallocator.go
```diff
@@ -50,11 +50,13 @@ type TargetAllocatorSpec struct {
 	// match. When omitted, the operator defaults to matching
 	// prometheus.dynatrace.com: "true". An empty selector {} matches all CRDs.
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={"matchLabels":{"prometheus.dynatrace.com":"true"}}
 	ScrapeCRSelector *metav1.LabelSelector `json:"scrapeCRSelector,omitempty"`
 
 	// Restricts which namespaces the TA watches for CRDs. An empty selector {}
 	// matches all namespaces.
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={"matchLabels":{"prometheus.dynatrace.com":"true"}}
```
gkrenn: We don't want to have a match label for the namespace per default.

Will add this to the KOD as well, as it was not defined there.

pkg/api/v1alpha1/dtprometheus/types.go
```diff
@@ -30,7 +30,8 @@ type DTPrometheusSpec struct { //nolint:revive
 	// Configures the Target Allocator, which holds all Prometheus service
 	// discovery metadata and distributes scrape targets across the scraper pool.
 	// +kubebuilder:validation:Optional
-	TargetAllocator TargetAllocatorSpec `json:"targetAllocator,omitzero"`
+	// +kubebuilder:default={}
```
gkrenn, line 33: Good catch! Please also add defaults to target allocator and scraper spec then. We shouldn't be inconsitent there.

pkg/controllers/dtprometheus/targetallocator/reconciler.go
andriisoldatenko: i prefer to merge PR without commenting code, it's fine if it fails, IMO.

pkg/controllers/dtprometheus/targetallocator/reconciler.go
andriisoldatenko: hmm.. so you are saying we don't annotate CM, we annotate deployment? True, true. we don't need to.

depends how hard it to do. Fine by me as is. Otherwise you need to return data and then compute hash?

pkg/controllers/dtprometheus/targetallocator/reconciler.go
0sewa0: why not just not comment it out?
if we know the `image.TargetAllocator` (so the component name), then it is all done, and at some point it will work

I thought that was the idea 🤷

0sewa0: (and while it doesn't work, you will also just return an error 🤷 )

pkg/controllers/dtprometheus/targetallocator/reconciler.go
```diff
@@ -0,0 +1,425 @@
...
+	TLSKeyFilePath  string `json:"tls_keyt_file_path"`
+	CAFilePath      string `json:"ca_cert_file_path"`
+}
+
+type ScrapeConfig struct {
+	Enabled                bool                  `json:"enabled"`
+	ScrapeInterval         metav1.Duration       `json:"scrape_interval"`
+	PodMonitorSelector     *metav1.LabelSelector `json:"pod_monitor_selector,omitempty"`
+	ScrapeConfigSelector   *metav1.LabelSelector `json:"scrape_config_selector,omitempty"`
+	ServiceMonitorSelector *metav1.LabelSelector `json:"service_monitor_selector,omitempty"`
```
gkrenn: ProbeSelector is missing

pkg/controllers/dtprometheus/targetallocator/reconciler.go
```diff
@@ -0,0 +1,425 @@
...
+			ServiceMonitorSelector: s.Spec.ScrapeCRSelector,
+		},
+	}
+
+	data, err := yaml.Marshal(cfg)
+	if err != nil {
+		return err
+	}
+
+	cm := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: s.Spec.GetDeploymentName(), Namespace: s.Owner.Namespace}}
```
gkrenn, line 180: One thing you didn't change, but still would to do in this PR: 

the GetDeploymentName() function uses `ta.name` => let's rename the `.name` field. It's not a target allocator name but rather the name of the dt prometheus object.

pkg/controllers/dtprometheus/targetallocator/reconciler.go
```diff
@@ -0,0 +1,425 @@
...
+	}
+
+	var err error
+
+	for _, f := range []func(context.Context, *reconcileScope) error{
+		r.reconcileConfigMap,
+		r.reconcileDeployment,
+		r.reconcileService,
+	} {
+		if err = f(ctx, scope); err != nil {
```
gkrenn, line 111: Is it wise to break and not just try everything we can? - And then log all the errors?

If not I would at least change the order to CM => Service => Deployment

pkg/controllers/dtprometheus/targetallocator/reconciler.go
```diff
@@ -0,0 +1,425 @@
...
+	default:
+		condition.Status = metav1.ConditionFalse
+		condition.Reason = status.ReasonReconciling
+		condition.Message = "target allocator is pending"
+	}
+
+	_ = meta.SetStatusCondition(&s.Owner.Status.Conditions, condition)
+}
+
+func (r *Reconciler) reconcileConfigMap(ctx context.Context, s *reconcileScope) error {
```
gkrenn, line 144: Where did you implement this AC? - "The config map can be controlled by the user if specified by an annotation on the config map and won't be updated by the operator."

pkg/api/v1alpha1/dtprometheus/types.go
avorima: it only makes sense if the object actually has a default, otherwise it will just result in an empty field like `gateway: {}`. i will add it for scraper

pkg/controllers/dtprometheus/targetallocator/reconciler.go
avorima: i don't think we gain anything by parallelizing. this models the dependency chain quite well. the service needs pods to run and the deployment needs a config

pkg/controllers/dtprometheus/targetallocator/reconciler.go
avorima: yes, the hash is only needed for the deployment annotation

pkg/controllers/dtprometheus/targetallocator/reconciler.go
avorima: the change is trivial, but we should do it once we can actually verify it

pkg/controllers/dtprometheus/targetallocator/reconciler.go
andriisoldatenko: we can do this optimization later.

pkg/controllers/dtprometheus/targetallocator/reconciler.go
avorima: I will implement it by merging the configs instead, but in a smaller follow-up PR

pkg/controllers/dtprometheus/reconciler.go
```diff
@@ -158,13 +169,19 @@ func (r *Reconciler) patchStatus(ctx context.Context, dtp *dtprometheus.DTPromet
...
 	dtp.ResourceVersion = ""
+
+	var opts []client.SubResourcePatchOption
+
 	if dtp.Kind == "" || dtp.APIVersion == "" {
 		// Set these values for unit tests
 		dtp.Kind = "DTPrometheus"
 		dtp.APIVersion = "dynatrace.com/v1alpha1"
+		// The fake client cannot set the field owner, but it's a required field.
+		opts = []client.SubResourcePatchOption{client.FieldOwner("dynatrace-operator")}
```
0sewa0: would it make sense to have a `const` for `"dynatrace-operator"`?
- could be put into the `wrapper.go` so it is `k8sobjects.FieldOwner`

(considering that it will always be the same, if you had to define it in the very beginning 😅 )

pkg/controllers/dtprometheus/targetallocator/reconciler.go
0sewa0: I don't want to block this, so lets leave as is, we will revisit this probably anyway later.

pkg/controllers/dtprometheus/targetallocator/reconciler.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context, dtp *dtprometheus.DTPrometheus, dk *dynakube.DynaKube, imageClient image.Client) error {
+	ctx, _ = logd.NewFromContext(ctx, "targetallocator")
+
+	scope := &reconcileScope{
+		Owner:       dtp,
+		DynaKube:    dk,
+		Spec:        dtp.TargetAllocator(),
+		AppLabels:   k8slabel.NewAppLabels("opentelemetry-target-allocator", dtp.Name, "otel-allocator", ""),
```
0sewa0: could use `k8slabel.New` 👀

pkg/controllers/dtprometheus/targetallocator/reconciler.go
gkrenn: Fine for me, will add it to the KOD

pkg/controllers/dtprometheus/reconciler.go
andriisoldatenko: we already have const here https://github.com/dynatrace/dynatrace-operator/blob/2a39d88a0ee1fbb61e2d22db02520b3dc92ffc80/pkg/version/version.go#L20 but not sure we want import it, but we could. or move it to global const `pkg/consts/consts.go:7`

avorima: i'll move it into wrapper.go

pkg/controllers/dtprometheus/targetallocator/testdata/deployment.yaml
```diff
@@ -0,0 +1,99 @@
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: dtp-target-allocator
```
gkrenn: Let's align the name with our KOD. You can question the name there 😄 

Same for cm and service

pkg/controllers/dtprometheus/targetallocator/reconciler.go
```diff
@@ -0,0 +1,428 @@
...
+
+func (r *Reconciler) reconcileConfigMap(ctx context.Context, s *reconcileScope) error {
+	log := logd.FromContext(ctx)
+	log.Debug("reconciling configmap")
+
+	cfg := Config{
+		ListenAddr:         ":8080",
+		AllocationStrategy: "consistent-hashing",
+		CollectorNamespace: s.Owner.Namespace,
+		CollectorSelector:  s.Spec.ScrapeCRNamespaceSelector,
```
gkrenn: The collector selector is nothing user provided and should just the label for the TA to find all our scrapers. Something like this: 

```
    collector_selector:
      matchlabels:
        app.kubernetes.io/instance: <scraper-instance-label>   # selects scraper pods
        app.kubernetes.io/name: opentelemetry-collector
```

pkg/controllers/dtprometheus/targetallocator/reconciler.go
```diff
@@ -0,0 +1,428 @@
...
+	ListenAddr      string `json:"listen_addr"`
+	TLSCertFilePath string `json:"tls_cert_file_path"`
+	TLSKeyFilePath  string `json:"tls_keyt_file_path"`
+	CAFilePath      string `json:"ca_cert_file_path"`
+}
+
+type ScrapeConfig struct {
+	Enabled                bool                  `json:"enabled"`
+	ScrapeInterval         metav1.Duration       `json:"scrape_interval"`
+	PodMonitorSelector     *metav1.LabelSelector `json:"pod_monitor_selector,omitempty"`
```
gkrenn: You are missing all the NamespaceSelector, this should be the fields that are using `spec.targetAllocator.scrapeCRNamespaceSelector`

pkg/controllers/dtprometheus/targetallocator/testdata/configmap.yaml
```diff
@@ -0,0 +1,61 @@
...
+    kind: DTPrometheus
+    name: dtp
+    uid: dtp-uid
+data:
+  targetallocator.yaml: |
+    allocation_strategy: consistent-hashing
+    collector_namespace: dynatrace
+    collector_selector:
+      matchLabels:
+        app.kubernetes.io/instance: otel-allocator
```
gkrenn: Should contain the labels of the scraper pods. Otherwise everything looks fine and I will approve then

pkg/util/kubernetes/fields/k8slabel/labels.go
```diff
@@ -36,6 +36,14 @@ const (
 	OperatorComponentLabel      = "operator"
 )
 
+func OTelTargetAllocator() *Labels {
+	return New("opentelemetry-target-allocator", "otel-allocator", "")
+}
+
+func OTelScraper() *Labels {
+	return New("opentelemetry-collector", "otel-scraper", "")
```
gkrenn, line 44: The instance name should be the name of the deployment right? - Otherwise the target allocator would use all scrapers in the cluster if there are multiple instances of dtProm, but both deployments should be separated.

avorima: These are the values from your PoC. I put them in a central place so that whoever actually does the scraper can figure it out

gkrenn: Fine for me, if you say that's the responsible of the scraper ticket 👍

gkrenn (approved):

PR #7241, opened 2026-08-12
Add istio handling for kubemon

0sewa0:
## Description

[ICP-7554](https://dt-rnd.atlassian.net/browse/ICP-7554)

The simplest way to respect the `spec.enableIstio` incase of `kubernetesMonitoring`.

Reuses `ReconcileActiveGate`, so that we do not double our ServiceEntries, when Kubemon is used alongside a Routing AG.

This will make the "standalone" Kubemon DK's status a bit ugly as it will mention ActiveGate and not Kubemon.
- The whole isito reconcile needs a rework it avoid duplicating endpoints in our ServiceEntries....

## How can this be tested?

> [!NOTE]
> - Don't forget to set the `experimental.enableKubemonOperand: true` helm value.
> - Don't forget to have isito installed on your cluster.

```yaml
kind: DynaKube
metadata:
  name: example
  namespace: dynatrace
spec:
  apiUrl: https://example.dev.dynatracelabs.com/api
 
  enableIstio: true
  kubernetesMonitoring:
    registration: {}
```

Should create "activegate" ServiceEntries.

[ICP-7554]: https://dt-rnd.atlassian.net/browse/ICP-7554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -163,7 +166,14 @@ func (r *Reconciler) ReconcileActiveGate(ctx context.Context, dk *dynakube.DynaK
 		return nil
 	}
 
-	agCommunicationHosts, err := connectioninfo.ParseAGCommunicationHosts(dk.Status.ActiveGate.ConnectionInfo.Endpoints)
+	// We shouldn't reconcile this separatly, as the endpoints of the Generic AG and the Kubemon AG will be the same.
+	// If we reconciled this separatly, then we would duplicate the endpoints in ServiceEntries, which is not something Istio recommends.
```
jokuniew:
```suggestion
	// We shouldn't reconcile this separately, as the endpoints of the Generic AG and the Kubemon AG will be the same.
	// If we reconciled this separately, then we would duplicate the endpoints in ServiceEntries, which is not something Istio recommends.
```

pkg/controllers/dynakube/kubemon/reconciler.go
```diff
@@ -110,6 +117,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, dk *dynakube.DynaKube, dtcli
 		return err
 	}
 
+	if err = r.istioReconciler.ReconcileActiveGate(ctx, dk); err != nil {
```
jokuniew: shouldn't this reconciliation be done only if activegate is not enabled to avoid double call `r.istioReconciler.ReconcileActiveGate`?

pkg/controllers/dynakube/istio/reconciler.go
```diff
@@ -163,7 +166,14 @@ func (r *Reconciler) ReconcileActiveGate(ctx context.Context, dk *dynakube.DynaK
...
+	endpoints := dk.Status.ActiveGate.ConnectionInfo.Endpoints
+	if len(endpoints) == 0 && isKubemonEnabled {
```
jokuniew:
```suggestion
	if !dk.ActiveGate().IsEnabled() && isKubemonEnabled {
```
i wonder if there will be scenario when someone disables ActiveGate middle-reconciliation so there will be leftower ConnectionInfo endpoints, so they will never be len==0?

pkg/controllers/dynakube/kubemon/reconciler.go
0sewa0: +1

pkg/controllers/dynakube/istio/reconciler.go
0sewa0: kinda unlikely but will fix to be safe

0sewa0: done in https://github.com/Dynatrace/dynatrace-operator/pull/7241/commits/b177170c8ca7db49dccd3d8610324b2751c9ab79

pkg/controllers/dynakube/kubemon/reconciler.go
0sewa0: done in https://github.com/Dynatrace/dynatrace-operator/pull/7241/commits/b177170c8ca7db49dccd3d8610324b2751c9ab79

jokuniew (approved):
With Dynakube from PR description
serviceEntry is created as well as VirtualService with respective hosts list

PR #7246, opened 2026-08-13
Add new labels structure

mihaitanasedt:
## Description
 Beginning of [ICP-8644](https://dt-rnd.atlassian.net/browse/ICP-8644), to be used with the new comming components, kubemon, prometheus stuff etc..

New approach example (in context of kubemon)

```go
"app.kubernetes.io/name": kubemon
"app.kubernetes.io/created-by": dynakube // dynakube CR name
"app.kubernetes.io/managed-by": dynatrace-operator
"app.kubernetes.io/version": 1.347.0.20260811-205912 // active gate version
"app.kubernetes.io/manager-version": 1.11.0 // operator version

```

## How can this be tested?
`make go/test`

[ICP-8644]: https://dt-rnd.atlassian.net/browse/ICP-8644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/util/kubernetes/fields/k8slabel/labels.go
+++ b/pkg/util/kubernetes/fields/k8slabel/labels.go
@@ -115,8 +115,8 @@ func NewCoreLabels(dynakubeName, component string) *CoreLabels {
 	}
 }
 
-func (labels *Labels) BuildLabels() map[string]string {
-	labelsMap := labels.BuildMatchLabels()
+func (labels *Labels) Build() map[string]string {
+	labelsMap := labels.BuildMatch()
 
 	if labels.Version != "" {
 		labelsMap[AppVersionLabel] = labels.Version
@@ -129,7 +129,7 @@ func (labels *Labels) BuildLabels() map[string]string {
 	return labelsMap
 }
 
-func (labels *Labels) BuildMatchLabels() map[string]string {
+func (labels *Labels) BuildMatch() map[string]string {
 	return map[string]string{
 		AppNameLabel:      labels.Name,
 		AppCreatedByLabel: labels.CreatedBy,
--- a/pkg/util/kubernetes/fields/k8slabel/labels.go
+++ b/pkg/util/kubernetes/fields/k8slabel/labels.go
@@ -121,7 +121,7 @@ func NewCoreLabels(dynakubeName, component string) *CoreLabels {
 }
 
 func (labels *Labels) AsMap() map[string]string {
-	labelsMap := labels.MatchAsMap()
+	labelsMap := labels.AsSelector()
 
 	if labels.Version != "" {
 		labelsMap[AppVersionLabel] = labels.Version
@@ -134,7 +134,7 @@ func (labels *Labels) AsMap() map[string]string {
 	return labelsMap
 }
 
-func (labels *Labels) MatchAsMap() map[string]string {
+func (labels *Labels) AsSelector() map[string]string {
 	return map[string]string{
 		AppNameLabel:      labels.Name,
 		AppCreatedByLabel: labels.CreatedBy,
--- a/pkg/util/kubernetes/fields/k8slabel/labels.go
+++ b/pkg/util/kubernetes/fields/k8slabel/labels.go
@@ -12,12 +12,14 @@ import (
 )
 
 const (
-	AppNameLabel           = "app.kubernetes.io/name"
-	AppCreatedByLabel      = "app.kubernetes.io/created-by"
-	AppManagedByLabel      = "app.kubernetes.io/managed-by"
-	AppComponentLabel      = "app.kubernetes.io/component"
-	AppVersionLabel        = "app.kubernetes.io/version"
-	AppManagerVersionLabel = "app.kubernetes.io/manager-version"
+	AppNameLabel     = "app.kubernetes.io/name"
+	AppInstanceLabel = "app.kubernetes.io/instance"
+	// Deprecated: Use AppInstanceLabel instead. See https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-created-by-deprecated
+	AppCreatedByLabel    = "app.kubernetes.io/created-by"
+	AppManagedByLabel    = "app.kubernetes.io/managed-by"
+	AppComponentLabel    = "app.kubernetes.io/component"
+	AppVersionLabel      = "app.kubernetes.io/version"
+	OperatorVersionLabel = "internal.dynatrace.com/operator-version"
 
 	OneAgentComponentLabel      = "oneagent"
 	CodeModuleComponentLabel    = "codemodule"
@@ -47,11 +49,11 @@ type coreMatchLabels struct {
 }
 
 type Labels struct {
-	Name           string
-	CreatedBy      string
-	ManagedBy      string
-	Version        string
-	ManagerVersion string
+	Name            string
+	Instance        string
+	ManagedBy       string
+	Version         string
+	OperatorVersion string
 }
 
 type AppLabels struct {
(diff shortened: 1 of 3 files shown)


avorima (dismissed):

avorima:
Maybe also deprecate the other constructors

pkg/util/kubernetes/fields/k8slabel/labels.go
```diff
@@ -56,6 +65,16 @@ type CoreLabels struct {
 	Version string
 }
 
+func New(appName, createdBy, appVersion string) *Labels {
```
0sewa0: we should add some docstring that gives context for this

Example: (feel free to adjust)
```suggestion
// New will return a simplified Labels struct that contains all the necessary info related to the ownership of a resource.
// It should be used instead of NewAppLabels and NewCoreLabels, as those are overcomplicating things.
// If appVersion is empty the related `version` label will be omitted. This should be done for resources that have no version in general, like a `Secret`.
func New(appName, createdBy, appVersion string) *Labels {
```

otherwise LGTM :D

pkg/util/kubernetes/fields/k8slabel/labels.go
```diff
@@ -94,6 +113,28 @@ func NewCoreLabels(dynakubeName, component string) *CoreLabels {
 	}
 }
 
+func (labels *Labels) Build() map[string]string {
```
andriisoldatenko: labels.AsMap()  -> map

pkg/util/kubernetes/fields/k8slabel/labels.go
```diff
@@ -94,6 +113,28 @@ func NewCoreLabels(dynakubeName, component string) *CoreLabels {
...
+	}
+
+	if labels.ManagerVersion != "" {
+		labelsMap[AppManagerVersionLabel] = labels.ManagerVersion
+	}
+
+	return labelsMap
+}
+
+func (labels *Labels) BuildMatch() map[string]string {
```
andriisoldatenko:
```suggestion
func (labels *Labels) MatchAsMap() map[string]string {
```

mihaitanasedt: [c45b1bb](https://github.com/Dynatrace/dynatrace-operator/pull/7246/commits/c45b1bbd83fa8cbc44129a925bdfbf704be7f9e5)

pkg/util/kubernetes/fields/k8slabel/labels.go
mihaitanasedt: [c45b1bb](https://github.com/Dynatrace/dynatrace-operator/pull/7246/commits/c45b1bbd83fa8cbc44129a925bdfbf704be7f9e5)

pkg/util/kubernetes/fields/k8slabel/labels.go
mihaitanasedt: [c45b1bb](https://github.com/Dynatrace/dynatrace-operator/pull/7246/commits/c45b1bbd83fa8cbc44129a925bdfbf704be7f9e5)

pkg/util/kubernetes/fields/k8slabel/labels.go
```diff
@@ -94,6 +120,28 @@ func NewCoreLabels(dynakubeName, component string) *CoreLabels {
...
+	}
+
+	if labels.ManagerVersion != "" {
+		labelsMap[AppManagerVersionLabel] = labels.ManagerVersion
+	}
+
+	return labelsMap
+}
+
+func (labels *Labels) MatchAsMap() map[string]string {
```
0sewa0: this reads wrong to me, like it should return a `bool` 😅 

use `AsSelector` (your idea :D), add a docs string what is the difference between `AsMap` and `AsSelector`

mihaitanasedt: [4fddfb6](https://github.com/Dynatrace/dynatrace-operator/pull/7246/commits/4fddfb696038dcd066640c871e00ee5c7f78cbc6)

pkg/util/kubernetes/fields/k8slabel/labels.go
mihaitanasedt: After additional feedback [4fddfb6](https://github.com/Dynatrace/dynatrace-operator/pull/7246/commits/4fddfb696038dcd066640c871e00ee5c7f78cbc6)

pkg/util/kubernetes/fields/k8slabel/labels.go
mihaitanasedt: After additional feedback [4fddfb6](https://github.com/Dynatrace/dynatrace-operator/pull/7246/commits/4fddfb696038dcd066640c871e00ee5c7f78cbc6)

pkg/util/kubernetes/fields/k8slabel/labels.go
```diff
@@ -12,11 +12,12 @@ import (
...
-	AppCreatedByLabel = "app.kubernetes.io/created-by"
-	AppManagedByLabel = "app.kubernetes.io/managed-by"
-	AppComponentLabel = "app.kubernetes.io/component"
-	AppVersionLabel   = "app.kubernetes.io/version"
+	AppNameLabel           = "app.kubernetes.io/name"
+	AppCreatedByLabel      = "app.kubernetes.io/created-by"
+	AppManagedByLabel      = "app.kubernetes.io/managed-by"
+	AppComponentLabel      = "app.kubernetes.io/component"
+	AppVersionLabel        = "app.kubernetes.io/version"
+	AppManagerVersionLabel = "app.kubernetes.io/manager-version"
```
svfoxat:
```suggestion
	AppManagerVersionLabel = "dynatrace.com/manager-version"
```
I wouldn't use `app.kubernetes.io/` for this custom key, as it is *technically* reserved for well-known labels listed [here](kubernetes.io/docs/reference/labels-annotations-taints)

Maybe `dynatrace.com/operator-version` would be better? wdyt?

pkg/util/kubernetes/fields/k8slabel/labels.go
```diff
@@ -12,11 +12,12 @@ import (
 )
 
 const (
-	AppNameLabel      = "app.kubernetes.io/name"
-	AppCreatedByLabel = "app.kubernetes.io/created-by"
-	AppManagedByLabel = "app.kubernetes.io/managed-by"
-	AppComponentLabel = "app.kubernetes.io/component"
-	AppVersionLabel   = "app.kubernetes.io/version"
+	AppNameLabel           = "app.kubernetes.io/name"
+	AppCreatedByLabel      = "app.kubernetes.io/created-by"
```
svfoxat:
```suggestion
	AppCreatedByLabel      = "app.kubernetes.io/instance"
```

This label has been [deprecated](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-created-by-deprecated). Maybe we can use `app.kubernetes.io/instance`?

pkg/util/kubernetes/fields/k8slabel/labels.go
0sewa0: We use `internal.dynatrace.com` in our helm labels for selectors, so that could also be considered.

`internal.dynatrace.com/operator-version` for example

0sewa0: note: in the [official docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) there is a note that this is a "recommendation and not required" so we should be probably fine anyway 😅

svfoxat: > note: in the [official docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) there is a note that this is a "recommendation and not required" so we should be probably fine anyway 😅

Yes, it would work either way and, as you said, just a recommendation but we should at least think about it 😄

mihaitanasedt: good catch [10aa004](https://github.com/Dynatrace/dynatrace-operator/pull/7246/commits/10aa004609c41cc6cc69d6237ccc3c1e36cde27d)

pkg/util/kubernetes/fields/k8slabel/labels.go
mihaitanasedt: [10aa004](https://github.com/Dynatrace/dynatrace-operator/pull/7246/commits/10aa004609c41cc6cc69d6237ccc3c1e36cde27d)

0sewa0 (approved):

andriisoldatenko (approved):

svfoxat (approved):

PR #7257, opened 2026-08-14
Validate the CM version from the API

0sewa0:
## Description

[ICP-8425](https://dt-rnd.atlassian.net/browse/ICP-8425)

Fixes an exploit the relies on the fact that the CM version we get from the API is something we trust AND use to create a filepath in the CSI driver. I did a mitigation on both the Operator and CSI level, just to be sure.

1. [Validate the CM version from the API](https://github.com/Dynatrace/dynatrace-operator/commit/cbb0e13fd7a1f2eb8b09b5a8fe8bf91f7c7b3e4f)
2. [Sanitize the targetDir in the provisioner](https://github.com/Dynatrace/dynatrace-operator/commit/cc64df49853d8b372cb34e19fe32b621368db88b)

## How can this be tested?

A detailed reproducer can be found in https://dt-rnd.atlassian.net/browse/ICP-8425?focusedCommentId=4768678.

But the tests are also pretty straight forward.

**More important to check that this doesn't break current logic.**

[ICP-8425]: https://dt-rnd.atlassian.net/browse/ICP-8425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/util/sanitize/sanitize.go
+++ b/pkg/util/sanitize/sanitize.go
@@ -5,6 +5,7 @@
 package sanitize
 
 import (
+	"os"
 	"strings"
 	"sync"
 )
@@ -42,4 +43,22 @@ func CommandLineArgs(args []string) []string {
 	}
 
 	return sanitized
+}
+
+var filePathSanitizer = strings.NewReplacer("..", "", "//", "/", ",", "", ":", "")
+
+// FilePath removes characters that are unsafe to use as a filesystem path component: "..", "//", ",", and ":".
+// Sanitization runs before the root-separator check so that inputs that disguise a bare separator
+// (e.g. "../" → "/" after stripping "..") are also stripped.
+//
+// [filepath.Clean] is not used here because it normalizes a full path rather than stripping characters
+// from an untrusted name component. A relative input such as "../secret" passes through Clean unchanged.
+func FilePath(path string) string {
+	path = filePathSanitizer.Replace(path)
+
+	if path == string(os.PathSeparator) {
+		return ""
+	}
+
+	return path
 }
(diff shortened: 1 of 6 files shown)


pkg/util/sanitize/sanitize.go
```diff
@@ -43,3 +44,21 @@ func CommandLineArgs(args []string) []string {
 
 	return sanitized
 }
+
+var filePathSanitizer = strings.NewReplacer("..", "", "//", "/", ",", "", ":", "")
+
+// FilePath removes characters that are unsafe to use as a filesystem path component: "..", "//", ",", and ":".
+// Sanitization runs before the root-separator check so that inputs that disguise a bare separator
+// (e.g. "../" → "/" after stripping "..") are also stripped.
```
andriisoldatenko, line 52: more a q: why not fail ? why do we want to sanitize something?

0sewa0:
Will close this because similar exploits have been rejected before:
```
They all share the same root: they rely on a service account that can write a DynaKube's (or EdgeConnect's) status subresource but not its spec. We reproduced the behavior and the code does trust those status fields, so the technical observation is correct.

The reason we are treating these as informative is the trust model. That status-only role is not something the operator ships or promises to honor as a security boundary. No default Helm role grants status write in isolation, our operator role bundles spec update, status, Secrets, and workloads together, so any principal that can legitimately write status in a real deployment can also write spec or more, and could reach the same outcome directly. In other words the boundary being crossed only exists in the custom role built for the PoC, and status write sits inside the same trust level as spec write by design.

We will go ahead and close these as Informative.
```

With the addition of:
AFAIK having the rights to change the `spec.apiURL` admin level privilege in our trust model.
And validating the URL is not an option, due to managed envs and environment activegate support.

PR #7266, opened 2026-08-18
Clean up license headers

avorima:
## Description

Ensure all license headers use the short form. I also removed a testdata file which was unused.

Fixed helm templating of licenses so they only get rendered if they outer if condition applies.


andriisoldatenko (dismissed):

jokuniew (dismissed):
`config/helm/*` headers are left long on purpose?

0sewa0 (dismissed):

avorima:
> `config/helm/*` headers are left long on purpose?

i just checked go files. will do helm as well

jokuniew (approved):

PR #7267, opened 2026-08-18
Set GOMEMLIMIT for targetallocator

avorima:
## Description

Implement leftover AC for targetallocator. All DTPromtheus deployed workloads should set the GOMEMLIMIT value to 90% of the memory limit (if it is configured).

ICP-8189

## How can this be tested?

Deploy targetallocator and play around with the resource values

### DIFF
--- a/pkg/util/kubernetes/fields/k8senv/env.go
+++ b/pkg/util/kubernetes/fields/k8senv/env.go
@@ -272,8 +272,8 @@ func GetMetadaSizeLimit(ctx context.Context) int {
 	return value
 }
 
-// AppendMemoryLimit appends GOMEMLIMIT env var
-func AppendMemoryLimit(envs []corev1.EnvVar, resources corev1.ResourceRequirements) []corev1.EnvVar {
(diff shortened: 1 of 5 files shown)


pkg/util/kubernetes/fields/k8senv/env.go
```diff
@@ -271,3 +271,13 @@ func GetMetadaSizeLimit(ctx context.Context) int {
 
 	return value
 }
+
+// AppendMemoryLimit appends GOMEMLIMIT env var
+func AppendMemoryLimit(envs []corev1.EnvVar, resources corev1.ResourceRequirements) []corev1.EnvVar {
```
0sewa0: nit:
```suggestion
func AppendGoMemoryLimit(envs []corev1.EnvVar, resources corev1.ResourceRequirements) []corev1.EnvVar {
```

gkrenn:
Tested and checked code, besides of what marcell mentioned, looks good and will approve afterwards.

0sewa0 (approved):

PR #7268, opened 2026-08-18
Make OTel component args configurable

avorima:
## Description

We decided to allow config overrides via command-line args.
Add sanitization to prevent AI slop vulnerability reports.

ICP-8866

## How can this be tested?

Deploy operator and set args


pkg/api/v1alpha1/dtprometheus/pod.go
```diff
@@ -61,3 +66,8 @@ type PodSpec struct {
 	// +kubebuilder:validation:Optional
 	Labels map[string]string `json:"labels,omitempty"`
 }
+
+// GetArgs returns sanitized command-line arguments.
+func (s PodSpec) GetArgs() []string {
```
0sewa0: nit
```suggestion
func (s PodSpec) SanitizedArgs() []string {
```

pkg/util/sanitize/sanitize_test.go
```diff
@@ -43,22 +44,21 @@ func TestCommandLineArg(t *testing.T) {
...
-		assert.Equal(t, []string{"foobar", "clean", "abc"}, CommandLineArgs(in))
-	})
+	tests := []struct {
+		name   string
+		in     []string
+		expect []string
+	}{
+		{"preserve nil", nil, nil},
+		{"empty slice", []string{}, []string{}},
+		{"sanitize", []string{"foo\nbar", "clean", "a\tb\x00c"}, []string{"foobar", "clean", "abc"}},
```
andriisoldatenko, line 54: can you please add one positive test case?
https://opentelemetry.io/docs/collector/configuration/
```
otelcol --config="yaml:exporters::debug::verbosity: normal"
```

andriisoldatenko: also i don't see if you can do it with TA https://github.com/Dynatrace/otel-target-allocator#configuration

andriisoldatenko: cc @gkrenn

andriisoldatenko: also sharing types (inside CRD) (agree @0sewa0 ) between components is not good, especially if you can't configure everything same way.

pkg/controllers/dtprometheus/targetallocator/testdata/deployment.yaml
```diff
@@ -37,19 +37,21 @@ spec:
...
       automountServiceAccountToken: true
       containers:
-      - env:
-        - name: OTELCOL_NAMESPACE
-          value: dynatrace
+      - name: targetallocator
         image: registry.example.com/target-allocator:1.2.3
         imagePullPolicy: Always
+        args:
+        - '--config="yaml:exporters::debug::verbosity: normal"'
```
andriisoldatenko: did you test this?

pkg/util/sanitize/sanitize_test.go
gkrenn: Oh no 😢  - Apparently there are flags for all the configuration options. Mario is currently testing that.

avorima: all components have flags

pkg/controllers/dtprometheus/targetallocator/testdata/deployment.yaml
avorima: it's not supposed to work. changed it to something that's more obviously a test value

gkrenn:
Please don't forget to update the sample in the asset folder as well

gkrenn:
Tested and checked the code, will approve after the sample is merged

assets/samples/dtprometheus/v1alpha1/complete.yaml
```diff
@@ -55,6 +55,19 @@ spec:
...
 
+    # Extra command-line flags for the Target Allocator.
+    # The operator owns the config file mounted at /conf/targetallocator.yaml, and
+    # CLI flags take precedence over it - so only logging flags are safe here.
+    # Do NOT set --config-file, --listen-addr, --enable-https-server, --listen-addr-https,
+    # --https-*, --enable-prometheus-cr-watcher or --kubeconfig-path: each one silently
+    # overrides operator-managed config, and the port flags also break the probes and Service.
+    args:
+      - --zap-log-level=debug
+      - --zap-encoder=console
```
gkrenn: Would not enable any args in this sample file. Those should be resonable defaults. 

It's fine if we just leave it as comments

avorima: The complete.yaml sets all fields. Only a few of these are reasonable IMO

gkrenn (approved):

PR #7270, opened 2026-08-18
Set DTPrometeus phase on DynaKube deletion

avorima:
## Description

The DTPrometheus reconciler only watches for DynaKube phase changes, so a deletion usually won't trigger a reconcile.
Also fixed the phase calculation to account for missing DynaKube after everything was ready.

ICP-7433

## How can this be tested?

Deploy a DK and DTP, then delete the DK -> phase error
Create the DK again -> phase running

### DIFF
--- a/pkg/controllers/dtprometheus/reconciler.go
+++ b/pkg/controllers/dtprometheus/reconciler.go
@@ -5,6 +5,7 @@ package dtprometheus
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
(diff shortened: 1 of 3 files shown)


pkg/controllers/dtprometheus/reconciler.go
```diff
@@ -237,7 +259,7 @@ func newDynaKubePhaseChangedPredicate() predicate.Funcs {
 			return false
```
gkrenn, line 259: Shouldn't we listen to creations as well? - What if the dynakube is deleted => phase of dtp goes to error => dk is created => phase phase should change back to running.

gkrenn: Ok, well, ignore me. The phase of the dynakube then anyway changes and we are listening to that event.

0sewa0 (approved):

PR #7272, opened 2026-08-18
Make AG volumes and mounts user configurable

0sewa0:
## Description

https://dt-rnd.atlassian.net/browse/ICP-8871

> [!IMPORTANT]
> **The 2000+ additions is mainly(~1600) from the CRD, so don't panic** 

[Add volume/volumeMounts to generic AG](https://github.com/Dynatrace/dynatrace-operator/pull/7272/commits/de54c741598cdd66d16153edf436f488d421a848) is trivial, and contains mainly CRD related changes.

Carefully review [Extract AG volumes related consts](https://github.com/Dynatrace/dynatrace-operator/pull/7272/commits/f369b7736278ebdef52e431fbf50f54028e81800) 
- Just extracting consts so I can use it 1 commit later for validation.
- Many small changes

[Another interesting part is the validation:](https://github.com/Dynatrace/dynatrace-operator/pull/7272/commits/5d29e488dbfcc1bf47083cbe3cd33239f4b7dfaa)
- For conflicting volumes: `activeGateHasConflictingVolumes`
   - Didn't add our injected Volumes/Mounts to the check, as that is more an internal usecase. (can do in a follow up if needed)
   - Doesn't check for "what is turned on", as that would go overboard.
- For allowed types: `activeGateHasDisallowedVolumeType`
  - I wanted first a deny list, but allow list made more sense for now. 
- General validation: `activeGateHasInvalidVolumes`
  - The general validation feels like something we should let K8s do, should I keep it?

## How can this be tested?

Minimal DK:
```yaml
kind: DynaKube
metadata:
  name: marcell-dev
  namespace: dynatrace
spec:
  apiUrl: https://hvs0985d.dev.dynatracelabs.com/api
  activeGate:
    capabilities:
    - routing
    volumeMounts:
    - mountPath: /var/run/secrets/example
      name: example
    volumes:
    - emptyDir: {}
      name: example
```
Feel free to go crazy, 90% of this change is validation.


pkg/api/validation/dynakube/activegate.go
```diff
@@ -86,3 +97,104 @@ func activeGateRollingUpdateWithOldK8sVersion(_ context.Context, _ *Validator, d
 
 	return ""
 }
+
+func activeGateHasConflictingVolumes(ctx context.Context, _ *Validator, dk *dynakube.DynaKube) string {
+	log := logd.FromContext(ctx)
+
+	for _, volume := range dk.Spec.ActiveGate.Volumes {
+		if slices.Contains(agconsts.VolumeNames, volume.Name) {
```
andriisoldatenko, line 100:
```suggestion
		if slices.Contains(agconsts.ReservedVolumeNames, volume.Name) {
```
or  `agconsts.InternalVolumeNames`

pkg/api/validation/dynakube/activegate.go
```diff
@@ -86,3 +97,104 @@ func activeGateRollingUpdateWithOldK8sVersion(_ context.Context, _ *Validator, d
...
+			log.Info("conflicting ActiveGate volume name detected", "volume", volume.Name)
+
+			return fmt.Sprintf(errorActiveGateConflictingVolumeName, volume.Name)
+		}
+	}
+
+	for _, volumeMount := range dk.Spec.ActiveGate.VolumeMounts {
+		for _, managedPath := range agconsts.VolumeMountPaths {
```
andriisoldatenko: same as above pls

pkg/api/validation/dynakube/activegate.go
0sewa0: so what would be the `agconsts.ExtrenalVolumeNames`? :)

andriisoldatenko: fine by me, just use `ReservedVolumeNames`

0sewa0: but which are the `FreeVolumeNames`? :D

andriisoldatenko: also we need somehow document in code to update the list when soneone introduce new VolumeName for AG :)

pkg/api/validation/dynakube/activegate.go
0sewa0: I am more conflicted here as the `VolumeMountPaths` should be the `MainContainerVolumeMountPaths` if we want to be super exact, but that is a mouthful

pkg/api/validation/dynakube/activegate.go
0sewa0: > also we need somehow document in code to update the list when soneone introduce new VolumeName for AG :)

now everything AG mounts related are in 1 places, so there is pattern there.
not much more I can do to enforce it. 😞  
- writing it in the style guide or something is not going to do much

0sewa0: > fine by me, just use ReservedVolumeNames

I did a local alias for the `agconsts.VolumeNames`, to match the meaning in the context of validation.

but in the global context `agconsts.VolumeNames` is more correct.

avorima:
A few things:
* agree that `activeGateHasInvalidVolumes` should not be our job. the statefulset will send a clear enough message
* `activeGateHasConflictingVolumes` should be best effort. if somebody is trying to be smart with relative paths, they're on their own.
* given that the "consts" package (which is a bad name to begin with) doesn't contain constants anymore, maybe it's about time we move these to a place where the package actually improves naming

pkg/controllers/dynakube/activegate/internal/statefulset/statefulset_test.go
```diff
@@ -682,10 +682,77 @@ func TestVolumeMounts(t *testing.T) {
...
+
+		managedVolume := corev1.Volume{
+			Name: consts.GatewayTmpVolumeName,
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		}
+		assert.Contains(t, sts.Spec.Template.Spec.Volumes, managedVolume)
+		assert.Len(t, sts.Spec.Template.Spec.Volumes, 1)
+	})
```
andriisoldatenko, line 755: can we extend integration tests https://github.com/dynatrace/dynatrace-operator/blob/3016e66a74e8599e76ba378bf58a7b6db7d0b2fe/pkg/controllers/dynakube/activegate/internal/statefulset/reconciler_integration_test.go#L65 to add some positive cases / may be negative:

- to make sure k8s api doesn't reject anything.

I know you have unit-tests - but if you check existing test, it was introduced bc unit tests doesn't cover properly volumes/volumeMounts

andriisoldatenko: Just demonstrate that feature is working all together.

0sewa0: will do

andriisoldatenko: ❤️ ❤️ ❤️

pkg/api/validation/dynakube/activegate.go
andriisoldatenko: fine by me resolved

andriisoldatenko: you added comment to slices it's cover what i want

0sewa0:
> agree that activeGateHasInvalidVolumes should not be our job. the statefulset will send a clear enough message

will remove

> activeGateHasConflictingVolumes should be best effort. if somebody is trying to be smart with relative paths, they're on their own.

will just check for exact matches

> given that the "consts" package (which is a bad name to begin with) doesn't contain constants anymore, maybe it's about time we move these to a place where the package actually improves naming

@avorima any suggestions :D 

I thought about just put it into `package activegate` so it would be `activegate.VolumeNames` and such
EDIT: this is an import cycle 😅

andriisoldatenko:
> I thought about just put it into `package activegate` so it would be `activegate.VolumeNames` and such

+1

pkg/controllers/dynakube/activegate/internal/statefulset/statefulset_test.go
0sewa0: so I did it, I don't see how it covers more than the unittests 🤷

pkg/controllers/dynakube/activegate/internal/statefulset/reconciler_integration_test.go
```diff
@@ -66,3 +66,54 @@ func TestStatefulSet(t *testing.T) {
...
+	}
+
+	userVolume := corev1.Volume{
+		Name: "my-user-volume",
+		VolumeSource: corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{},
+		},
+	}
+	userVolumeMount := corev1.VolumeMount{
+		Name:      "my-user-volume",
```
andriisoldatenko: if you use different names, unit-test will work, integration won/t

also applying not allowed vol type is better to test all together (if you run unit tests for validator - yes)

/shrug

0sewa0:
> given that the "consts" package (which is a bad name to begin with) doesn't contain constants anymore, maybe it's about time we move these to a place where the package actually improves naming

@avorima @andriisoldatenko 
I would not try to fix this now, as that is a bigger restructure
the current `package activegate` will just cause import cycles, and the whole `activegate` packages should be reorg-ed anyway, and then we can tackle it there.

pkg/controllers/dynakube/activegate/internal/statefulset/reconciler_integration_test.go
0sewa0: I just simplied our validation, and would like to rely more on k8s validation.

I added a `TestStatefulSetUserVolumesAndMountsFail`

where there is typo in the volume name.

cc @avorima this relates to
> agree that activeGateHasInvalidVolumes should not be our job. the statefulset will send a clear enough message

there won't always be a statefulset, so Operator will error, if we are fine with that then I will leave as is.

pkg/api/validation/dynakube/activegate.go
```diff
@@ -86,3 +92,54 @@ func activeGateRollingUpdateWithOldK8sVersion(_ context.Context, _ *Validator, d
...
+	for _, volume := range dk.Spec.ActiveGate.Volumes {
+		if slices.Contains(agconsts.VolumeNames, volume.Name) {
+			log.Info("conflicting ActiveGate volume name detected", "volume", volume.Name)
+
+			return fmt.Sprintf(errorActiveGateConflictingVolumeName, volume.Name)
+		}
+	}
+
+	for _, volumeMount := range dk.Spec.ActiveGate.VolumeMounts {
+		for slices.Contains(agconsts.MainVolumeMountPaths, volumeMount.MountPath) {
```
avorima: this should've caused a test timeout. is this case not covered?

pkg/api/validation/dynakube/activegate.go
```diff
@@ -86,3 +92,54 @@ func activeGateRollingUpdateWithOldK8sVersion(_ context.Context, _ *Validator, d
...
+}
+
+func activeGateHasDisallowedVolumeType(ctx context.Context, _ *Validator, dk *dynakube.DynaKube) string {
+	log := logd.FromContext(ctx)
+
+	for _, volume := range dk.Spec.ActiveGate.Volumes {
+		if !slices.ContainsFunc(allowedActiveGateVolumeSources, func(isAllowed func(corev1.VolumeSource) bool) bool {
+			return isAllowed(volume.VolumeSource)
+		}) {
+			log.Info("ActiveGate volume uses a type disallowed by the OpenShift nonroot-v2 SCC", "volume", volume.Name)
```
avorima: logging the whole volume spec might be useful

pkg/controllers/dynakube/activegate/internal/customproperties/reconciler.go
```diff
@@ -24,8 +25,8 @@ const (
 	Suffix     = "custom-properties"
 	DataKey    = "customProperties"
 	DataPath   = "custom.properties"
-	VolumeName = "custom-properties"
-	MountPath  = "/var/lib/dynatrace/gateway/config_template/custom.properties"
+	VolumeName = consts.CustomPropertiesVolumeName
+	MountPath  = consts.CustomPropertiesMountPath
```
avorima: why do we need these now that the constants are in a central place

pkg/controllers/dynakube/activegate/internal/statefulset/reconciler_integration_test.go
```diff
@@ -66,3 +66,99 @@ func TestStatefulSet(t *testing.T) {
 	err = reconciler.Reconcile(ctx, &dk, mcap)
 	require.NoError(t, err)
 }
+
+func TestStatefulSetUserVolumesAndMounts(t *testing.T) {
+	clt := integrationtests.SetupTestEnvironment(t)
```
avorima: only 1 test environment should be set up for a package, because it eats into the runtime

pkg/controllers/dynakube/activegate/internal/statefulset/statefulset_test.go
```diff
@@ -682,10 +682,77 @@ func TestVolumeMounts(t *testing.T) {
 
 		expectedVolumeMount := corev1.VolumeMount{
 			Name:      consts.GatewayTmpVolumeName,
-			MountPath: consts.GatewayTmpMountPoint,
+			MountPath: consts.GatewayTmpMountPath,
 		}
 		require.Contains(t, sts.Spec.Template.Spec.Containers[0].VolumeMounts, expectedVolumeMount)
 	})
+
+	t.Run("user-provided volume mounts are included before managed mounts", func(t *testing.T) {
```
avorima, line 690: i would've expected the mounts to be appended. where does this decision come from?

0sewa0: that is just impl detail, and claude gave it this title 😅 
because I do:
```go
	mounts := statefulSetBuilder.capability.Properties().VolumeMounts

	return append(mounts, corev1.VolumeMount{
		Name:      consts.GatewayTmpVolumeName,
		MountPath: consts.GatewayTmpMountPath,
	})
```

it shouldn't matter, as the order of volume mounts shouldn't matter in general.

pkg/api/validation/dynakube/activegate.go
0sewa0: why? Am I blind? 🙈

pkg/api/validation/dynakube/activegate.go
0sewa0: fixed

pkg/controllers/dynakube/activegate/internal/customproperties/reconciler.go
0sewa0: fixed

pkg/controllers/dynakube/activegate/internal/statefulset/reconciler_integration_test.go
0sewa0: fixed

.github/workflows/ci.yaml
```diff
@@ -124,6 +125,10 @@ jobs:
           CHANGED_HELM_CONFIG_FILES: ${{ steps.helm.outputs.all_changed_files }}
           CHANGED_MARKDOWN_FILES: ${{ steps.markdown.outputs.all_changed_files }}
           CHANGED_PREREQUISITES_MK_FILE: ${{ steps.prerequisites.outputs.all_changed_files }}
+      - name: Get Go version from Dockerfile
+        id: go-version
+        run: |
+          echo "go_version=$(grep -m1 'FROM.*golang:' Dockerfile | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"
```
andriisoldatenko, line 131: grep -m1 'FROM.*golang:' Dockerfile | sed 's/.*golang:\([0-9.]*\).*/\1/'
1.26.5

andriisoldatenko (dismissed):

pkg/api/validation/dynakube/activegate.go
avorima: oh right because you return you didn't notice
```suggestion
		if slices.Contains(agconsts.MainVolumeMountPaths, volumeMount.MountPath) {
```

0sewa0: wow :D I am blind 🙈

avorima (approved):

andriisoldatenko (approved):

PR #7282, opened 2026-08-20
ICP-8542: support multiple pull-secrets in troubleshoot command

Edivion:
## Description

Previously the troubleshoot command only respected `customPullSecret` (when configured) and didn't fall back or combine with alternative secrets in the namespace. Which caused a 401 error. 

This was previously fixed for other parts of the product. Used the then-built function for the troubleshoot command to respect 1 or many relevant secret(s). 

## How can this be tested?

1. Create a classic API token (platform will not work) for your tenant
2. Create the valid token in your cluster:
```
kubectl create secret generic classic-token-secret -n dynatrace \
  --from-literal=apiToken=<your-api-token>
```
3. Create a fake `customPullSecret`:
```
kubectl create secret docker-registry my-fake-acr -n dynatrace \
  --docker-server=some.domain.com \
  --docker-username=fake \
  --docker-password=fake
```
4. Create a minimal Dynakube that specifies my-fake-acr as the `customPullSecret`
5. After reconciliation run the troubleshoot command and verify the output:
```
kubectl -n dynatrace exec deploy/dynatrace-operator -- \
  /usr/local/bin/dynatrace-operator troubleshoot -n dynatrace \
  --dynakube minimal-dynakube
```
6. In the output you can now see that both secrets are recognized:
```
[dynakube  ]        pull secret 'dynatrace:classic-token-secret' exists
[dynakube  ]        pull secret 'dynatrace:my-fake-acr' exists
```

### DIFF
--- a/cmd/troubleshoot/cmd.go
+++ b/cmd/troubleshoot/cmd.go
@@ -121,7 +121,7 @@ func runChecksForDynakube(ctx context.Context, baseLog logd.Logger, apiReader cl
 	logNewCheckf(log, "checking if '%s:%s' Dynakube is configured correctly", dk.Namespace, dk.Name)
 	logInfof(log, "using '%s:%s' Dynakube", dk.Namespace, dk.Name)
 
-	pullSecret, err := checkDynakube(ctx, baseLog, apiReader, &dk)
+	err := checkDynakube(ctx, baseLog, apiReader, &dk)
 	if err != nil {
 		return errors.Wrapf(err, "'%s:%s' Dynakube isn't valid. %s",
 			dk.Namespace, dk.Name, dynakubeNotValidMessage())
@@ -129,7 +129,7 @@ func runChecksForDynakube(ctx context.Context, baseLog logd.Logger, apiReader cl
 
 	logOkf(log, "'%s:%s' Dynakube is valid", dk.Namespace, dk.Name)
 
-	keychain, err := dockerkeychain.NewDockerKeychain(ctx, apiReader, pullSecret)
+	keychain, err := dockerkeychain.NewDockerKeychains(ctx, apiReader, dk.Namespace, dk.PullSecretNames())
 	if err != nil {
 		return err
 	}
--- a/cmd/troubleshoot/dynakube.go
+++ b/cmd/troubleshoot/dynakube.go
@@ -29,33 +29,34 @@ const (
 
 const dynakubeCheckLoggerName = "dynakube"
 
-func checkDynakube(ctx context.Context, baseLog logd.Logger, apiReader client.Reader, dk *dynakube.DynaKube) (corev1.Secret, error) {
+func checkDynakube(ctx context.Context, baseLog logd.Logger, apiReader client.Reader, dk *dynakube.DynaKube) error {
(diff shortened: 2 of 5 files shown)


cmd/troubleshoot/cmd.go
```diff
@@ -121,15 +121,15 @@ func runChecksForDynakube(ctx context.Context, baseLog logd.Logger, apiReader cl
...
+	err := checkDynakube(ctx, baseLog, apiReader, &dk)
 	if err != nil {
 		return errors.Wrapf(err, "'%s:%s' Dynakube isn't valid. %s",
 			dk.Namespace, dk.Name, dynakubeNotValidMessage())
 	}
 
 	logOkf(log, "'%s:%s' Dynakube is valid", dk.Namespace, dk.Name)
 
-	keychain, err := dockerkeychain.NewDockerKeychain(ctx, apiReader, pullSecret)
+	keychain, err := dockerkeychain.NewDockerKeychains(ctx, apiReader, dk.Namespace, dk.PullSecretNames())
```
avorima, line 132: nit: it's still a single keychain so NewDockerKeychain still works

cmd/troubleshoot/dynakube.go
```diff
@@ -180,26 +181,32 @@ func checkAPIURLForLatestAgentVersion(ctx context.Context, baseLog logd.Logger,
 	return nil
 }
 
-func checkPullSecretExists(ctx context.Context, baseLog logd.Logger, apiReader client.Reader, dk *dynakube.DynaKube) (corev1.Secret, error) {
+func checkPullSecretExists(ctx context.Context, baseLog logd.Logger, apiReader client.Reader, dk *dynakube.DynaKube) ([]corev1.Secret, error) {
```
avorima: nit: this could be plural now

cmd/troubleshoot/dynakube.go
```diff
@@ -180,26 +181,32 @@ func checkAPIURLForLatestAgentVersion(ctx context.Context, baseLog logd.Logger,
...
 	log := baseLog.WithName(dynakubeCheckLoggerName)
 
 	query := k8ssecret.Query(nil, apiReader)
 
-	pullSecret, err := query.Get(ctx, types.NamespacedName{Namespace: dk.Namespace, Name: dk.PullSecretName()})
-	if err != nil {
-		return corev1.Secret{}, errors.Wrapf(err, "'%s:%s' pull secret is missing", dk.Namespace, dk.PullSecretName())
-	}
+	var secrets []corev1.Secret
```
avorima: use pointers when dealing with k8s objects, especially if we're building the list ourselves (ICP-1216)

cmd/troubleshoot/cmd.go
Edivion: Since this is also used in the installer, this goes beyond the scope of this ticket. Want me to create a follow-up ticket?

avorima: yes, i think we should consolidate the two functions

avorima (approved):

cmd/troubleshoot/cmd.go
Edivion: ICP-9071

PR #7238, opened 2026-08-11
Kubemon KSPM

mihaitanasedt:
## Description
The PR extends new Kubernetes Monitoring (kubemon) with KSPM

[Jira ICP-7553](https://dt-rnd.atlassian.net/browse/ICP-7553)

## How can this be tested?
New sanity e2e test  `test/e2e/kspm/kubemon`

Run unit tests `make go/test`

Manual tested :
| Scenario | Result |
|---|---|
| KSPM + ActiveGate + automation annotation + platform token | ✅ works |
| KSPM + ActiveGate + automation annotation + API token | ✅ works |
| KSPM + ActiveGate + no annotation | ✅ validation rejected |
| KSPM + ActiveGate + replicas = 1 | ✅ works |
| KSPM + ActiveGate + replicas > 1 | ✅ validation rejected |
| KSPM + KubernetesMonitoring (with registration) + platform token | ✅ works |
| KSPM + KubernetesMonitoring (with registration) + API token | ✅ works |
| KSPM + KubernetesMonitoring (no registration) | ✅ validation rejected |
| KSPM + KubernetesMonitoring + replicas = 1 | ✅ works |
| KSPM + KubernetesMonitoring + replicas > 1 | ✅ validation rejected |
| KSPM + no KubernetesMonitoring + no ActiveGate | ✅ validation rejected |
| KSPM + KubernetesMonitoring + KSPM module disabled | ✅ validation rejected |

PR #7287, opened 2026-08-20
Omit faulty version labels

mihaitanasedt:
## Description
Add validation for version labels.

Follow up from https://github.com/Dynatrace/dynatrace-operator/pull/7238 ([comment](https://github.com/Dynatrace/dynatrace-operator/pull/7238#discussion_r3811470768)). Part of https://github.com/Dynatrace/dynatrace-operator/pull/7246 and [ICP-8644](https://dt-rnd.atlassian.net/browse/ICP-8644).

## How can this be tested?
`make go/test`

[ICP-8644]: https://dt-rnd.atlassian.net/browse/ICP-8644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


avorima:
This looks ok, but why add it now. We've never had issues with the label content before

andriisoldatenko (dismissed):

mihaitanasedt:
> This looks ok, but why add it now. We've never had issues with the label content before

Accidentaly saw it, no specific reason, just to be safe.

avorima:
our testing should catch any label values that are invalid. maybe we should surface this error instead of dropping the value silently, because it's a signal that something is broken

mihaitanasedt:
> our testing should catch any label values that are invalid. maybe we should surface this error instead of dropping the value silently, because it's a signal that something is broken

A valid image version can be an invalid label. Per my understanding, version labels might be used for monitoring, so an invalid label will not break any functionality -> If image version is not a valid label just omit it.

avorima:
> A valid image version can be an invalid label. Per my understanding, version labels might be used for monitoring, so an invalid label will not break any functionality -> If image version is not a valid label just omit it.

Our versions are semver, so anything that deviates and gets pushed to a release is a bug. Dev images aren't semver, but they still are guaranteed to be a valid label. We can debate whether we want to loosen this restriction, but until then if it breaks it's a bug that would be hidden by this change.

mihaitanasedt:
> > A valid image version can be an invalid label. Per my understanding, version labels might be used for monitoring, so an invalid label will not break any functionality -> If image version is not a valid label just omit it.
> 
> Our versions are semver, so anything that deviates and gets pushed to a release is a bug. Dev images aren't semver, but they still are guaranteed to be a valid label. We can debate whether we want to loosen this restriction, but until then if it breaks it's a bug that would be hidden by this change.

Couldn't customer have a image in a private registry with a modified tag that isn't valid?

avorima:
> Couldn't customer have a image in a private registry with a modified tag that isn't valid?

That's fair, but what about the operator version

mihaitanasedt:
> > Couldn't customer have a image in a private registry with a modified tag that isn't valid?
> 
> That's fair, but what about the operator version

Operator version is always valid, I can remove it from tests, the check is for app version

pkg/util/kubernetes/fields/k8slabel/labels.go
```diff
@@ -211,7 +212,11 @@ func NotEqual(currentLabels, desiredLabels map[string]string) bool {
 
 func truncateVersion(ver string) string {
```
svfoxat: nit: the function now also validates and not only truncates it, maybe `sanitizeVersion`/`normalizeVersion` would be a more fitting name now?

pkg/util/kubernetes/fields/k8slabel/labels.go
```diff
@@ -211,7 +212,11 @@ func NotEqual(currentLabels, desiredLabels map[string]string) bool {
...
 	if len(ver) > validation.DNS1035LabelMaxLength {
-		return ver[:validation.DNS1035LabelMaxLength]
+		ver = ver[:validation.DNS1035LabelMaxLength]
+	}
+
+	if errs := content.IsLabelValue(ver); len(errs) > 0 {
```
svfoxat, line 227: nit:
```suggestion
	if errs := content.IsLabelValue(ver); len(errs) > 0 {
		log.Warn("omitting invalid version label", "version", ver, "errors", errs)
```

Would make it easier to find out what happened with the version label *IF* this would ever happen

svfoxat (commented):
just some small nits

pkg/util/kubernetes/fields/k8slabel/labels.go
avorima: this should not sanitize

mihaitanasedt: [0137494](https://github.com/Dynatrace/dynatrace-operator/pull/7287/commits/01374940ac765bf4135a8997c0625dd922ff62fc)

pkg/util/kubernetes/fields/k8slabel/labels.go
mihaitanasedt: [0137494](https://github.com/Dynatrace/dynatrace-operator/pull/7287/commits/01374940ac765bf4135a8997c0625dd922ff62fc)

pkg/util/kubernetes/fields/k8slabel/labels.go
avorima: i'm just referring to the sanitization. the truncate should still happen. that was actually the point of my initial remark

mihaitanasedt: [6b2e121](https://github.com/Dynatrace/dynatrace-operator/pull/7287/commits/6b2e121669d064f1424135bfc9c1caed823407e8) 👍

avorima (approved):

svfoxat (approved):

PR #7289, opened 2026-08-20
Implement workload metadata 

avorima:
## Description

Implement support for workload metadata from enrichment rules and direct annotations.

ICP-8266

## How can this be tested?

Create rules on your tenant and check that the precedence works.


jokuniew (approved):
workload annotations are propagated as expected to args and metadata enrichment files on pod 👍

0sewa0 (approved):

PR #7296, opened 2026-08-24
Fix kubemon KSPM e2e test

mihaitanasedt:
## Description
Fix kubemon kspm e2e test. Problem was that if you try to upgrade the operator in tests, our setup checks just for readiness (no generation) and might create a race condition between the old and new version resulting in a test can run in an outdated env. Flag ~[`--wait`](https://helm.sh/docs/intro/using_helm/#helpful-options-for-installupgraderollback)~ `--rollback-on-failure`  solves the issue globally.

[ICP-9002](https://dt-rnd.atlassian.net/browse/ICP-9002)

## How can this be tested?
`make test/e2e/kspm/kubemon`

[ICP-9002]: https://dt-rnd.atlassian.net/browse/ICP-9002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/e2e/helpers/components/operator/installation.go
+++ b/test/e2e/helpers/components/operator/installation.go
@@ -227,7 +227,7 @@ func getHelmOptions(releaseTag, platform string, withCSI bool) ([]helm.Option, e
 		helm.WithNamespace("dynatrace"),
 		helm.WithArgs("--create-namespace"),
 		helm.WithArgs("--install"),
-		helm.WithArgs("--wait"),
+		helm.WithArgs("--rollback-on-failure"),
 		helm.WithArgs("--set", fmt.Sprintf("platform=%s", platform)),
 		helm.WithArgs("--set", "installCRD=true"),
 		helm.WithArgs("--set", fmt.Sprintf("csidriver.enabled=%t", withCSI)),
--- a/test/e2e/helpers/components/operator/installation.go
+++ b/test/e2e/helpers/components/operator/installation.go
@@ -227,6 +227,7 @@ func getHelmOptions(releaseTag, platform string, withCSI bool) ([]helm.Option, e
 		helm.WithNamespace("dynatrace"),
 		helm.WithArgs("--create-namespace"),
 		helm.WithArgs("--install"),
+		helm.WithArgs("--wait"),
 		helm.WithArgs("--set", fmt.Sprintf("platform=%s", platform)),
 		helm.WithArgs("--set", "installCRD=true"),
 		helm.WithArgs("--set", fmt.Sprintf("csidriver.enabled=%t", withCSI)),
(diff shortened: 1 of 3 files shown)


test/e2e/helpers/components/operator/installation.go
```diff
@@ -227,6 +227,7 @@ func getHelmOptions(releaseTag, platform string, withCSI bool) ([]helm.Option, e
 		helm.WithNamespace("dynatrace"),
 		helm.WithArgs("--create-namespace"),
 		helm.WithArgs("--install"),
+		helm.WithArgs("--wait"),
```
andriisoldatenko: i would change it to `--rollback-on-failure` why?
- because in our release page we recommend to use `--atomic` but it has been deprecated and new `--rollback-on-failure` is recommended.
From another side:  `--rollback-on-failure`  if set, Helm will rollback the upgrade to previous success release upon failure. The `--wait` flag will be defaulted to "watcher" if --rollback-on-failure is set

So same as `"--wait"` default value is watcher.

andriisoldatenko (changes requested):

test/e2e/helpers/components/operator/installation.go
mihaitanasedt: `--rollback-on-failure` Is for helm 4, but in the helm chart we state: `This Helm Chart requires Helm 3.`, e2e pipelines use Helm `4.2.4`. Is we support helm 4 should we update the description?

mihaitanasedt: [1fa070d](https://github.com/Dynatrace/dynatrace-operator/pull/7296/commits/1fa070d390496c5674923e7a6e30b7b58a0513b5)

andriisoldatenko: we do support helm4, AFAIK. so are you saying it won't work?
but in our release notes we just saying to use `--atomic`?

andriisoldatenko (approved):

test/e2e/helpers/components/operator/installation.go
mihaitanasedt: I upgraded to latest helm (4.2.4) and it looks like it works. I say that in chart description we tell that helm 3 is required, maybe change wording so people understand that helm 4 is also supported?. 

https://github.com/Dynatrace/dynatrace-operator/blob/70d37e5ee84d0dbac2c4bd4ab7946e91e6d65e9f/config/helm/chart/default/README.md?plain=1#L5

andriisoldatenko: it's good catch, but the wording is not clear, may be we need to discuss it. And say requires atleast Helm 3?

mihaitanasedt: yes, something like this

andriisoldatenko (approved):

PR #7297, opened 2026-08-24
Add support for image volume injection

andriisoldatenko:
## Description

Resolves https://dt-rnd.atlassian.net/browse/ICP-8560

Implement image volume type injection for CloudNativeFullStack and ApplicationMonitoring modes

## How can this be tested?

> [!NOTE]
> E2e will be handle in a seperate ticket. 

Scenario 1:
0. deploy operator without CSI
1. Deploy appMon dk:
```
---
kind: DynaKube
metadata:
  name: dynakube
  namespace: dynatrace
  annotations: {}
spec:
  apiUrl: https://hbc47170.dev.dynatracelabs.com/api
  oneAgent:
    applicationMonitoring:
      codeModulesImage: public.ecr.aws/dynatrace/dynatrace-codemodules:1.343.92.20260813-103442
```

2. Add annotation to your sample `oneagent.dynatrace.com/volume-type: oci` / or use DK ff
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: go-sample-oci
  namespace: go-sample-oci
spec:
  replicas: 1
  selector:
    matchLabels:
      app: go-sample-oci
  template:
    metadata:
      annotations:
        oneagent.dynatrace.com/volume-type: image
      labels:
        app: go-sample-oci
    spec:
      containers:
        - image: asoldatenko/echo-server
          imagePullPolicy: Always
          name: app
```
3. Make sure 
```
k get pods -n go-sample-oci go-sample-oci-7fd4546bfb-nwr4j -o yaml | yq .spec.containers
```
- sample app has init container
- has LD_PRELOAD
```
root@go-sample-77659c7c89-cx9nj:/app# env | grep LD_PRELOAD
LD_PRELOAD=/opt/dynatrace/oneagent-paas/agent/lib64/liboneagentproc.so
```
- process has preloaded oneagent:
```
root@go-sample-77659c7c89-cx9nj:/# ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
root           1       0  0 05:02 ?        00:00:13 manager
root          43       0  0 09:07 pts/0    00:00:00 bash
root          63      43  0 09:08 pts/0    00:00:00 ps -ef
root@go-sample-77659c7c89-cx9nj:/# ldd /usr/local/bin/manager
        linux-vdso.so.1 (0x00007fff9bba6000)
        /opt/dynatrace/oneagent-paas/agent/lib64/liboneagentproc.so (0x00007e280cc00000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007e280ca0c000)
        /lib64/ld-linux-x86-64.so.2 (0x00007e280cfdd000)
```
- chech logs of technology specific agent
```
ls -la /var/lib/dynatrace/oneagent/log/go/ruxitagent_manager_go-sample-__1.0.log
```


hack/make/deploy/deploy.mk
```diff
@@ -32,6 +33,7 @@ deploy: manifests/crd/helm
 			--atomic \
 			--set installCRD=true \
 			--set csidriver.enabled=$(ENABLE_CSI) \
+			--set csidriver.migrationMode=$(CSI_MIGRATION_MODE) \
```
0sewa0, line 40: in honor of `deploy/no-csi` there could be a `deploy/migration-csi`

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -341,8 +341,16 @@ func findDuplicates[S ~[]E, E comparable](s S) []E {
 	return duplicates
 }
 
+func conflictingImageMode(_ context.Context, v *Validator, dk *dynakube.DynaKube) string {
+	if dk.FF().IsNodeImagePull() && dk.FF().IsImageVolume() {
+		return "node-image-pull and OCI image volume are mutually exclusive"
```
0sewa0: message should be a separate `const`

also, maybe more context `node-image-pull and OCI image volume are mutually exclusive as OCI image volume will not use the CSI driver, while node-image-pull FF only influences the CSI driver` (or something similar that also is more fitting the current validation errors 😅 )

pkg/api/exp/oneagent.go
```diff
@@ -61,6 +61,10 @@ func (ff *FeatureFlags) SkipOneAgentLivenessProbe() bool {
 	return ff.getBoolWithDefault(OASkipLivenessProbeKey, false)
 }
 
+func (ff *FeatureFlags) IsImageVolume() bool {
```
0sewa0: nitpick, now it is maybe too Java sounding 😅 
```suggestion
func (ff *FeatureFlags) IsCodeModuleImageVolume() bool {
```

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
```diff
@@ -24,6 +23,7 @@ import (
 const (
 	CSIVolumeType       = "csi"
 	EphemeralVolumeType = "ephemeral"
+	OCIVolumeImageType  = "oci"
```
0sewa0:
```suggestion
	ImageVolumeType  = "oci"
```

so either call it everywhere `OCIVolume...` or `ImageVolume...`, but don't mix it

the value can be `oci`, but I could also challenge it to call it `image` if the FF(`mount-code-modules-via-image-volume`) itself doesn't mention OCI 🤷

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
```diff
@@ -136,25 +144,27 @@ func containerIsInjected(container corev1.Container, _ *dtwebhook.BaseRequest) b
 func mutateUserContainers(request *dtwebhook.BaseRequest, installPath string, log logd.Logger) bool {
 	newContainers := request.NewContainers(containerIsInjected)
 	for _, container := range newContainers {
-		addOneAgentToContainer(request.DynaKube, container, request.Namespace, installPath, log)
+		addOneAgentToContainer(request, container, request.Namespace, installPath, log)
```
0sewa0: note: you could simplify `addOneAgentToContainer`
`mutateUserContainers` is overly simple, but `addOneAgentToContainer` is getting out of hand.

example:
- have `addOneAgentEnvsToContainer` (which would do 90% of what `addOneAgentToContainer` used to do) just for the env stuff
	- then just add the volume mounts part right in the loop

(check if it is nicer, maybe it will be just as ugly or annoying to test 😅 )

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
```diff
@@ -136,25 +144,27 @@ func containerIsInjected(container corev1.Container, _ *dtwebhook.BaseRequest) b
...
 	return len(newContainers) > 0
 }
 
-func addOneAgentToContainer(dk dynakube.DynaKube, container *corev1.Container, namespace corev1.Namespace, installPath string, log logd.Logger) {
+func addOneAgentToContainer(request *dtwebhook.BaseRequest, container *corev1.Container, namespace corev1.Namespace, installPath string, log logd.Logger) {
 	log.Info("adding OneAgent to container", "name", container.Name)
 
-	addVolumeMounts(container, installPath)
-	addDeploymentMetadataEnv(container, dk)
+	isImageVolume := hasOCIVolumeAnnotation(request) || request.DynaKube.FF().IsImageVolume()
```
0sewa0: the check `hasOCIVolumeAnnotation(request) || request.DynaKube.FF().IsImageVolume()` could be abstracted like the `isCSIVolume`

like 
```go
// feel free to format it in a more readable way :D 
func isImageVolume(mutationRequest *dtwebhook.BaseRequest) bool {
	return mutationRequest.DynaKube.OneAgent().GetCodeModulesImage() != "" &&
(maputils.GetField(mutationRequest.Pod.Annotations, AnnotationVolumeType, EphemeralVolumeType) == OCIVolumeImageType || request.DynaKube.FF().IsImageVolume())
}
```

0sewa0: sidenote, there is a `IsSelfExtractingImage` func, that is deadcode, could remove it 😉

pkg/webhook/mutation/pod/mutator/oneagent/init.go
```diff
@@ -38,7 +38,20 @@ func mutateInitContainer(mutationRequest *dtwebhook.MutationRequest, installPath
...
+	case hasOCIVolumeAnnotation(mutationRequest.BaseRequest) || mutationRequest.DynaKube.FF().IsImageVolume():
+		log.Info("configuring init-container with OCI image bin volume", "name", mutationRequest.PodName())
+		addOCIBinVolume(mutationRequest.Pod,
+			mutationRequest.DynaKube.OneAgent().GetCodeModulesImage(),
+			mutationRequest.DynaKube.OneAgent().GetCodeModulesImagePullPolicy())
+
+		customInitResources := mutationRequest.DynaKube.OneAgent().GetInitResources()
+		if customInitResources != nil {
+			mutationRequest.InstallContainer.Resources = *customInitResources
+		}
```
0sewa0: the way we handle this is dumb

1. there is a `initContainerResources` helper
2. the logic is the same for all cases

could do it once before or after the `switch` 🤷

0sewa0 (changes requested):

hack/make/deploy/deploy.mk
andriisoldatenko: i'm not sure we really need it, but it was hard to deploy it without make deploy, so decided to add, i can fix it.

andriisoldatenko: fixed in 575c8b48e

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
andriisoldatenko: this one we can ask tomorrow on operator sync? originally in docs we decided to use oci but i agree `image` is better and i'm already using it in code a lot.

pkg/api/validation/dynakube/oneagent.go
andriisoldatenko: fixed in a4cc9c982

pkg/api/exp/oneagent.go
andriisoldatenko: fixed in a4cc9c982

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
andriisoldatenko: fixed var name in a4cc9c982

~waiting for final approval to rename oci -> image~ 

done in 48808d03b

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
andriisoldatenko: >sidenote, there is a IsSelfExtractingImage func, that is deadcode, could remove it 😉

good catch removed in c0262e7ee

pkg/webhook/mutation/pod/mutator/oneagent/init.go
andriisoldatenko: yes i rememeber i also wanted to do it, but this helper:
```
func initContainerResources(dk dynakube.DynaKube) corev1.ResourceRequirements {
	customInitResources := dk.OneAgent().GetInitResources()
	if customInitResources != nil {
		return *customInitResources
	}

	return corev1.ResourceRequirements{}
}
```

reset to `orev1.ResourceRequirements{}` and i was affraid that it's not correct also in default case:

```
		// in case of no CSI, the default init resources will not work, so we must overwrite them to the custom ones from `spec.oneAgent.<mode>.initResources`, or unset them
		mutationRequest.InstallContainer.Resources = initContainerResources(mutationRequest.DynaKube)

```
based on comment we want do it, but in CSI scenario we never did it...

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
andriisoldatenko: and refactored in 5f851d027

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
andriisoldatenko: i've tried to refactor it in 90acab4df

pkg/webhook/mutation/pod/mutator/oneagent/init.go
0sewa0: should be fine for CSI aswell

0sewa0: so how it should work, because it is a bit cursed (so I wasn't fully correct)

we should only set it to `corev1.ResourceRequirements{}` (so empty) if we are in "url-download" mode and not `customInitResources` were given, because we don't want to set some super high default 400MB limit for that case.

So I would just do something like
```go
func mutateInitContainer(mutationRequest *dtwebhook.MutationRequest, installPath string) error {
	log := logd.FromContext(mutationRequest.Context)

	if !mutationRequest.DynaKube.IsCodeModulesStatusReady() {
		return dtwebhook.MutatorError{
			Err:      codeModulesStatusNotReadyError{dkName: mutationRequest.DynaKube.Name},
			Annotate: setNotInjectedAnnotationFunc(DynaKubeStatusNotReadyReason),
		}
	}

    // THIS
   	customInitResources := mutationRequest.DynaKube.OneAgent().GetInitResources()
	if customInitResources != nil {
			mutationRequest.InstallContainer.Resources = *customInitResources
	}

	switch {
		case isImageVolume(mutationRequest.BaseRequest): ...
		case isCSIVolume(mutationRequest.BaseRequest): ...
		default:
			if customInitResources == nil {
               mutationRequest.InstallContainer.Resources = corev1.ResourceRequirements{}
            }
            ...
```

even-though IMO we could use the default the limits for the `"configuring init-container with self-extracting image"` case, BUT that would be a change in behavior that I don't want to do now 😅

andriisoldatenko: okey that's how i did locally, let me push a change! thanks!

pkg/webhook/mutation/pod/mutator/oneagent/volumes.go
```diff
@@ -125,3 +141,23 @@ func addCSIBinVolume(pod *corev1.Pod, dkName string, maxTimeout string) error {
 
 	return nil
 }
+
+func addOCIBinVolume(pod *corev1.Pod, imageName string, pullPolicy corev1.PullPolicy) {
+	if k8svolume.Contains(pod.Spec.Volumes, BinVolumeName) {
```
avorima: this needs more thorough checks like `addCSIBinVolume` to prevent another security report. validate the name and source type

andriisoldatenko: good catch,

andriisoldatenko: done in 7fb954c51

pkg/webhook/mutation/pod/mutator/oneagent/init.go
andriisoldatenko: done in 7fb954c51

pkg/api/exp/oneagent_test.go
```diff
@@ -255,6 +255,44 @@ func TestSkipOneAgentLivenessProbe(t *testing.T) {
 	}
 }
 
+func TestIsImageVolume(t *testing.T) {
```
0sewa0: Mismatch with actual func name

(also kinda redundant test, but ok 😅 )

pkg/api/validation/dynakube/oneagent.go
```diff
@@ -56,6 +56,8 @@ Use a nodeSelector to avoid this conflict. Conflicting DynaKubes: %s`
 	warningDeprecatedVersionIgnored = `version field is deprecated and ignored. Please remove the version field from the DynaKube specification.`
 
 	errorImagePullRequiresCodeModulesImage = `The DynaKube specification enables node image pull, but neither a code modules image is set nor a public registry is used.`
+
+	errorConflictingImageMode = `Node image pull and OCI image volume are mutually exclusive as OCI image volume will not use the CSI driver, while node-image-pull FF only influences the CSI driver.`
```
0sewa0:
```suggestion
	errorConflictingImageMode = `Node image pull and image volume are mutually exclusive as image volume will not use the CSI driver, while node-image-pull FF only influences the CSI driver.`
```

pkg/webhook/mutation/pod/mutator/oneagent/init.go
```diff
@@ -38,7 +38,22 @@ func mutateInitContainer(mutationRequest *dtwebhook.MutationRequest, installPath
...
 
-	if isCSIVolume(mutationRequest.BaseRequest) {
+	customInitResources := mutationRequest.DynaKube.OneAgent().GetInitResources()
+	if customInitResources != nil {
+		mutationRequest.InstallContainer.Resources = *customInitResources
+	}
+
+	switch {
+	case isImageVolume(mutationRequest.BaseRequest):
+		log.Info("configuring init-container with OCI image bin volume", "name", mutationRequest.PodName())
```
0sewa0:
```suggestion
		log.Info("configuring init-container with image bin volume", "name", mutationRequest.PodName())
```

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
```diff
@@ -134,27 +140,28 @@ func containerIsInjected(container corev1.Container, _ *dtwebhook.BaseRequest) b
...
 
 	return len(newContainers) > 0
 }
 
-func addOneAgentToContainer(dk dynakube.DynaKube, container *corev1.Container, namespace corev1.Namespace, installPath string, log logd.Logger) {
-	log.Info("adding OneAgent to container", "name", container.Name)
-
-	addVolumeMounts(container, installPath)
-	addDeploymentMetadataEnv(container, dk)
+func addOneAgentEnvsToContainer(request *dtwebhook.BaseRequest, container *corev1.Container, namespace corev1.Namespace, installPath string, log logd.Logger) {
```
0sewa0:
```suggestion
func addOneAgentEnvsToContainer(dk *dynakube.DynaKube, container *corev1.Container, namespace corev1.Namespace, installPath string, log logd.Logger) {
```

you don't need the whole `request`

pkg/webhook/mutation/pod/mutator/oneagent/mutator_test.go
```diff
@@ -21,6 +20,65 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+func TestHasOCIVolumeAnnotation(t *testing.T) {
```
0sewa0: func name mismatch

pkg/webhook/mutation/pod/mutator/oneagent/volumes.go
```diff
@@ -125,3 +141,32 @@ func addCSIBinVolume(pod *corev1.Pod, dkName string, maxTimeout string) error {
 
 	return nil
 }
+
+func addOCIBinVolume(pod *corev1.Pod, imageName string, pullPolicy corev1.PullPolicy) error {
```
0sewa0:
```suggestion
func addImageBinVolume(pod *corev1.Pod, imageName string, pullPolicy corev1.PullPolicy) error {
```

0sewa0 (changes requested):
(also linting is failing)

pkg/api/exp/oneagent_test.go
andriisoldatenko: true, fixed in  28cdd4da3

pkg/api/validation/dynakube/oneagent.go
andriisoldatenko: fixed in  28cdd4da3

pkg/webhook/mutation/pod/mutator/oneagent/init.go
andriisoldatenko: fixed in  28cdd4da3

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
andriisoldatenko: fixed in  28cdd4da3

pkg/webhook/mutation/pod/mutator/oneagent/mutator_test.go
andriisoldatenko: fixed in  28cdd4da3

pkg/webhook/mutation/pod/mutator/oneagent/volumes.go
andriisoldatenko: fixed in  28cdd4da3

pkg/webhook/mutation/pod/mutator/oneagent/volumes.go
```diff
@@ -125,3 +141,32 @@ func addCSIBinVolume(pod *corev1.Pod, dkName string, maxTimeout string) error {
 
 	return nil
 }
+
+func addImageBinVolume(pod *corev1.Pod, imageName string, pullPolicy corev1.PullPolicy) error {
+	if vol := k8svolume.FindByName(pod.Spec.Volumes, BinVolumeName); vol != nil {
+		if vol.Image != nil && vol.Image.Reference != imageName {
```
avorima: it should be 
```suggestion
		if vol.Image == nil || vol.Image.Reference != imageName {
```
so that volumes that are not OCI image also get rejected

andriisoldatenko: how is it possible to set non oci image volume type?

andriisoldatenko:
```
k explain pod.spec.volumes --recursive | grep -A5 image
  image <ImageVolumeSource>
    pullPolicy  <string>
    enum: Always, IfNotPresent, Never
    reference   <string>
```
@avorima am i miss something?

avorima: a user can add a volume of any kind. we have to prevent mounting it

andriisoldatenko: or you mean any type. okey.

andriisoldatenko: fixed in 99df0ef7b

andriisoldatenko:
> (also linting is failing)

where? works fine for me locally and on CI/

0sewa0:
> where? works fine for me locally and on CI/

that was the case when I reviewed
that was before you pushed your changes that fixed it 😉 

image

andriisoldatenko:
> that was before you pushed your changes that fixed it 😉

ok ok trust you :)

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
```diff
@@ -59,6 +54,18 @@ func isCSIVolume(mutationRequest *dtwebhook.BaseRequest) bool {
...
 
+func hasImageVolumeAnnotation(mutationRequest *dtwebhook.BaseRequest) bool {
+	if mutationRequest.DynaKube.OneAgent().GetCodeModulesImage() != "" {
+		return maputils.GetField(mutationRequest.Pod.Annotations, AnnotationVolumeType, EphemeralVolumeType) == ImageVolumeType
+	}
+
+	return false
+}
+
+func isImageVolume(mutationRequest *dtwebhook.BaseRequest) bool {
```
0sewa0: tested and it works, but

there is a weird behavior that can happen if you set `oneagent.dynatrace.com/volume-type: "ephemeral"` on the Pod, but `feature.dynatrace.com/mount-code-modules-via-image-volume: "true"` on the DynaKube.

The DK FF will win, which is kinda weird, as the Pod annotations should have higher prio.

(We never really respected the `volume-type: csi`, because if you don't have CSI we can't do anything, and if you do then that is the default, so that is fine.)

you would need to do something like:
```go
func isImageVolume(mutationRequest *dtwebhook.BaseRequest) bool {
	defaultVolumeType := EphemeralVolumeType
	if mutationRequest.DynaKube.FF().IsCodeModuleImageVolume() {
		defaultVolumeType = ImageVolumeType
	}

	return hasImageVolumeAnnotation(mutationRequest) // this would also need to be changed to use ImageVolumeType as the default
}
```

andriisoldatenko: Thanks for nice catch, i fixed and added some missing tests to cover it f53764b4d

andriisoldatenko:
@avorima @0sewa0 guys PTAL!

also added more tests for default/custom init resources

pkg/controllers/csi/provisioner/controller.go
```diff
@@ -101,7 +101,7 @@ func (provisioner *OneAgentProvisioner) Reconcile(ctx context.Context, request r
 		return reconcile.Result{}, err
 	}
 
-	if !installconfig.GetModules().CSIDriver {
+	if !installconfig.GetModules().CSIDriver || dk.FF().IsCodeModuleImageVolume() {
 		log.Info("CSI driver migration mode active, running cleanup only")
```
waodim, line 105: This says "CSI driver migration mode active" even when the reason for this path is `IsCodeModuleImageVolume()` being true (and CSI is enabled)

But the image volume mode is a different operational state than the migration mode, no? Should we update the message here accordingly?

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
```diff
@@ -134,18 +143,19 @@ func containerIsInjected(container corev1.Container, _ *dtwebhook.BaseRequest) b
 }
 
 func mutateUserContainers(request *dtwebhook.BaseRequest, installPath string, log logd.Logger) bool {
+	log.Info("install path", "installPath", installPath, "isImageVolume", isImageVolume(request))
+
 	newContainers := request.NewContainers(containerIsInjected)
 	for _, container := range newContainers {
-		addOneAgentToContainer(request.DynaKube, container, request.Namespace, installPath, log)
+		addVolumeMounts(container, installPath, isImageVolume(request))
```
waodim: nit: IMO we can store the call here in a variable and use it then in the log and in the loop

pkg/controllers/csi/provisioner/controller.go
andriisoldatenko: based on KOD it's the same
>DynaKube CR feature flag annotation (feature.dynatrace.com/mount-code-modules-via-image-volume: true) for cluster-wide OCI volume injection configuration.
Note: this FF makes the CSI driver "not-enabled" in migration mode.

source https://github.com/dynatrace-apps/dynatrace-operator-internal/blob/main/docs/designs/kod-008-oci-image.md#2-scope

pkg/webhook/mutation/pod/mutator/oneagent/mutator.go
andriisoldatenko: fixed in fa3cc3e8c

pkg/controllers/csi/provisioner/controller.go
waodim: Thanks!

0sewa0 (approved):

waodim (approved):

avorima (approved):

jokuniew (approved):

PR #7298, opened 2026-08-24
Validate Kubemon section

mihaitanasedt:
## Description
Adds missing `spec.kubernetesMonitoring` related validations.

[ICP-8385](https://dt-rnd.atlassian.net/browse/ICP-8385)
- This PR only adds a subset what the ticket mentions:
   -  As the "deprecation warning for implicit AG creation when extensions are enabled" is a bigger more annoying change than anticipated.
   - The logmon related warning is not necessary, users complained about it previously, 

## How can this be tested?
Apply manually CRs 
`make go/test`

[ICP-8385]: https://dt-rnd.atlassian.net/browse/ICP-8385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/api/validation/dynakube/extensions.go
+++ b/pkg/api/validation/dynakube/extensions.go
@@ -10,11 +10,11 @@ import (
 )
 
 const (
-	warningExtensionsWithoutK8SMonitoring = "The Dynakube is configured with extensions without an ActiveGate with `kubernetes-monitoring` enabled or the `automatic-kubernetes-api-monitoring` feature flag. You need to ensure that Kubernetes monitoring is setup for this cluster."
+	warningExtensionsWithoutK8SMonitoring = "The DynaKube configures extensions, which require Kubernetes Monitoring with cluster registration. Configure `spec.kubernetesMonitoring` and add its `registration` section, or configure an ActiveGate with the `kubernetes-monitoring` capability and enable the `automatic-kubernetes-api-monitoring` feature flag."
 )
 
-func extensionsWithoutK8SMonitoring(ctx context.Context, dv *Validator, dk *dynakube.DynaKube) string {
-	if dk.Extensions().IsAnyEnabled() && (!dk.ActiveGate().IsKubernetesMonitoringEnabled() || !dk.FF().IsAutomaticK8sAPIMonitoring()) {
+func extensionsWithoutK8SMonitoring(_ context.Context, _ *Validator, dk *dynakube.DynaKube) string {
+	if dk.Extensions().IsAnyEnabled() && !dk.IsKubernetesMonitoringRegistrationEnabled() {
 		return warningExtensionsWithoutK8SMonitoring
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/api/validation/dynakube/extensions.go
```diff
@@ -10,11 +10,11 @@ import (
 )
 
 const (
-	warningExtensionsWithoutK8SMonitoring = "The Dynakube is configured with extensions without an ActiveGate with `kubernetes-monitoring` enabled or the `automatic-kubernetes-api-monitoring` feature flag. You need to ensure that Kubernetes monitoring is setup for this cluster."
+	warningExtensionsWithoutK8SMonitoring = "The DynaKube configures extensions, which require Kubernetes Monitoring with cluster registration. Configure `spec.kubernetesMonitoring` and add its `registration` section, or configure an ActiveGate with the `kubernetes-monitoring` capability and enable the `automatic-kubernetes-api-monitoring` feature flag."
```
waodim: The warning message here tells to set `spec.kubernetesMonitoring` + registration, but that path is gated by the `EXPERIMENTAL_ENABLE_KUBEMON_OPERAND` env var.

Should we mention it here for the time being?

pkg/api/validation/dynakube/kubemon_test.go
```diff
@@ -0,0 +1,91 @@
...
+						ValueFrom: "secret",
+					},
+				},
+			},
+		})
+	})
+}
+
+func TestMutualExclusiveKubernetesMonitoring(t *testing.T) {
+	t.Run("only kubernetesMonitoring", func(t *testing.T) {
```
waodim, line 69: This test never sets `EXPERIMENTAL_ENABLE_KUBEMON_OPERAND`, so `IsKubemonEnabled()` returns false and the validator always returns "" every other positive kubemon test sets the env var before `assertAllowed`. This one should too IMO.

pkg/api/validation/dynakube/extensions.go
0sewa0: I want to say "nah it will be released with 1.11", but that kinda goes against the whole "experimental" flag, so I will try to make it nicer

waodim (dismissed):
```
 1. The DynaKube configures Kubernetes Monitoring in both `spec.kubernetesMonitoring` and `spec.activeGate.capabilities`. Configure Kubernetes Monitoring in only one of these locations.
Error from server (Forbidden): error when creating "/.../../Desktop/tmp/kubemon-cr-validation-dk.yaml": admission webhook "v1beta6.dynakube.webhook.dynatrace.com" denied the request: 
 1. The DynaKube's `spec.kubernetesMonitoring.customProperties` cannot set both `value` and `valueFrom`. Set only one of these fields.
```

avorima (dismissed):

avorima (dismissed):

andriisoldatenko (dismissed):

avorima (approved):

PR #7305, opened 2026-08-25
add resourceAttributes from DynaKube to gateway config

svfoxat:
## Description

https://dt-rnd.atlassian.net/browse/ICP-8858

- adds `resourceAttributes` of the referenced `DynaKube` to the gateway `ConfigMap`
  - attributes are rendered as `set(attributes[%s], %s) where attributes[%s] == nil` to preserve precedence
  - I noticed a single `merge_maps(attributes, ParseJSON(...), "insert")` statement could replace the per-key `set()` calls and simplify the config, at the cost of readability (one packed JSON blob vs. one line per attribute), is this something we would want?
- annotations take precedence over `dk.resourceAttributes`
  - `JSON blob annotation > metadata.dynatrace.com/<key> > resourceAttributes`

## How can this be tested?

`make go lint && make go test`

### In cluster
- Apply your `DynaKube` with `spec.resourceAttributes` set, e.g.:
```yaml
spec:
  resourceAttributes:
    favorite.coffee: espresso
    deploy.mood: yolo
```

- Apply `DtPrometheus`

```yaml
kind: DTPrometheus
metadata:
  name: my-dtp
  namespace: dynatrace
spec:
  dynaKubeName: <your-dynakube-name>
  gateway:
    image: ghcr.io/dynatrace/dynatrace-otel-collector/dynatrace-otel-collector:0.51.0
```

- Check created `DtPrometheus`, `StatefulSet`, `ConfigMap`
- The `ConfigMap` should include one `set()` directive per `resourceAttribute` defined in the `DynaKube`
- Change  `spec.resourceAttributes`  on the `DynaKube`, reconcile, and confirm the `ConfigMap` content and  `ConfigMapHash`  annotation change, and the `StatefulSet `rolls out

- also verification of the generated ottl `transformer` with https://ottl.run/

### DIFF
--- a/pkg/controllers/dtprometheus/gateway/config.go
+++ b/pkg/controllers/dtprometheus/gateway/config.go
@@ -4,6 +4,11 @@
 package gateway
 
 import (
+	"fmt"
+	"maps"
+	"slices"
+	"strconv"
+
 	"github.com/Dynatrace/dynatrace-operator/pkg/otelcgen"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/pipeline"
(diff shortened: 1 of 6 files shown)


gkrenn:
@svfoxat Should we add a second gold file? - So we easily see the difference?

svfoxat:
> @svfoxat Should we add a second gold file? - So we easily see the difference?

Good Idea, will add that.

e: added in https://github.com/Dynatrace/dynatrace-operator/pull/7305/changes/95879d24213cb9cf39ece72c8c426fa6fc31eeec

gkrenn (approved):

PR #7307, opened 2026-08-25
Only replicate bootstrapper secrets if needed

avorima:
## Description

Check whether OA/metadata secrets are actually needed before replicating them in the admission webhook.

DTSEC-28206

## How can this be tested?

Unit tests

### DIFF
--- a/pkg/webhook/mutation/pod/handler/injection/handle.go
+++ b/pkg/webhook/mutation/pod/handler/injection/handle.go
@@ -15,6 +15,7 @@ import (
 	dtwebhook "github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/mutator"
 	"github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/mutator/metadata"
 	"github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/secrets"
+	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


pkg/webhook/mutation/pod/handler/injection/handle.go
```diff
@@ -61,43 +62,40 @@ func (h *Handler) Handle(mutationRequest *dtwebhook.MutationRequest) error {
 		return nil
 	}
 
-	if !h.isInputSecretPresent(mutationRequest, bootstrapperconfig.GetSourceConfigSecretName(mutationRequest.DynaKube.Name), consts.BootstrapperInitSecretName) {
+	oaEnabled := h.oaMutator.IsEnabled(ctx, mutationRequest.BaseRequest)
+
+	if installContainer := k8scontainer.FindInitInPodSpec(&mutationRequest.Pod.Spec, dtwebhook.InstallContainerName); installContainer != nil {
+		h.handlePodReinvocation(mutationRequest, installContainer, oaEnabled)
```
0sewa0: the `*dtwebhook.MutationRequest` already has an installContainer member, you could just
```suggestion
	if installContainer := k8scontainer.FindInitInPodSpec(&mutationRequest.Pod.Spec, dtwebhook.InstallContainerName); installContainer != nil {
		mutationRequest.InstallContainer = installContainer
		h.handlePodReinvocation(mutationRequest)
```

also, isn't passing around `oaEnabled` just over-optimizing? it makes everything more unreadable to me.

pkg/webhook/mutation/pod/handler/injection/handle.go
```diff
@@ -61,43 +62,40 @@ func (h *Handler) Handle(mutationRequest *dtwebhook.MutationRequest) error {
...
+
 		return nil
 	}
 
-	if mutationRequest.DynaKube.IsAGCertificateNeeded() || mutationRequest.DynaKube.Spec.TrustedCAs != "" {
+	needsConfig := oaEnabled || h.metaMutator.IsEnabled(ctx, mutationRequest.BaseRequest)
+
+	if needsConfig && !h.isInputSecretPresent(mutationRequest, bootstrapperconfig.GetSourceConfigSecretName(mutationRequest.DynaKube.Name), consts.BootstrapperInitSecretName) {
+		return nil
+	}
```
0sewa0: why not just add a check in the beginning? like
```go
if !mutationRequest.DynaKube.OneAgent().IsAppInjectionNeeded() && !mutationRequest.DynaKube.MetadataEnrichment().IsEnabled() {
		log.Debug("injection disabled", "podName", mutationRequest.PodName(), "namespace", mutationRequest.Namespace.Name)

		return nil
	}
if !h.metaMutator.IsEnabled(ctx, mutationRequest.BaseRequest) && !h.oaMutator.IsEnabled(ctx, mutationRequest.BaseRequest) {
		log.Debug("no mutator is enabled", "podName", mutationRequest.PodName(), "namespace", mutationRequest.Namespace.Name)

		return nil
	}
```

easier to read (also no funny business with adding `dynatrace-operator` as an init-container and we just bricking things)

pkg/webhook/mutation/pod/handler/injection/handle.go
```diff
@@ -111,28 +109,12 @@ func (h *Handler) Handle(mutationRequest *dtwebhook.MutationRequest) error {
...
+func (h *Handler) handlePodMutation(mutationRequest *dtwebhook.MutationRequest, needsConfig, oaEnabled bool) (bool, error) {
 	mutationRequest.InstallContainer = h.createInitContainerBase(mutationRequest.Context, mutationRequest.Pod, mutationRequest.DynaKube)
 
 	var mutated bool
 
-	oaEnabled := h.oaMutator.IsEnabled(mutationRequest.Context, mutationRequest.BaseRequest)
-
-	// metaMutator is automatically enabled if oaMutator is enabled, but it can also be enabled standalone
-	if h.metaMutator.IsEnabled(mutationRequest.Context, mutationRequest.BaseRequest) || oaEnabled {
+	if needsConfig {
```
0sewa0: technically unneeded if

similarly in `handlePodReinvocation` we do not do an if before `metadata.AddContainerAttributes`

as metadata-enrichment is always "on" if we reached this point (either because `metadataEnrichment` is on, or because `oneAgent` is on)

0sewa0 (changes requested):
TLDR:
I would keep the changes to a minimum for this
- So what I mentioned in https://github.com/Dynatrace/dynatrace-operator/pull/7307/changes#r3892836185 should be enough
- Pls don't try to optimize the `IsEnabled` calls as we even do it in the metadataMutator it self.
Most of the changes feel like putting lipstick on a pig.

There is this (super old / out of date)https://dt-rnd.atlassian.net/browse/ICP-941 ticket for fixing this whole webhook business, we should do bigger changes in context of that instead.

pkg/webhook/mutation/pod/handler/injection/handle.go
avorima: yeah, it was an optimization. will remove

avorima:
> TLDR: I would keep the changes to a minimum for this

Do you want to move the reinvocation back where it was before, so that replication gets retried every time? Removing the bool return value gets rid of the case where reinvocation failed, but we still log "no change", so I'd keep that.

> There is this (super old / out of date)https://dt-rnd.atlassian.net/browse/ICP-941 ticket for fixing this whole webhook business, we should do bigger changes in context of that instead.

Yes, I also created ICP-9093 to further reduce the work the webhook needs to do.

pkg/webhook/mutation/pod/handler/injection/handle.go
```diff
@@ -161,28 +139,25 @@ func (h *Handler) handlePodMutation(mutationRequest *dtwebhook.MutationRequest)
 	return mutated, nil
 }
 
-func (h *Handler) handlePodReinvocation(mutationRequest *dtwebhook.MutationRequest) bool {
-	log := logd.FromContext(mutationRequest.Context)
-
-	installContainer := k8scontainer.FindInitInPodSpec(&mutationRequest.Pod.Spec, dtwebhook.InstallContainerName)
-	if installContainer == nil {
-		log.Error(nil, "could not find init container during reinvoke")
+func (h *Handler) handlePodReinvocation(ctx context.Context, mutationRequest *dtwebhook.MutationRequest) {
```
0sewa0: while I agree with "don't put the context into struct", if we don't have a planned follow up to rip out the context from `dtwebhook.MutationRequest`, then doing this is not really an improvement.
- we are doing an anti pattern AND we are inconsistent about it 😬

0sewa0 (dismissed):

pkg/webhook/mutation/pod/handler/injection/handle.go
avorima: i didn't see that we reassign the context to the struct (wtf?), so that the logger name gets properly propagated. removed the context arg

this actually is a bug, because we don't reset the context and subsequent handlers inherit the name when they shouldn't

0sewa0: > struct (wtf?)

maany things could/should be fixed in this whole area 😅

0sewa0 (approved):

PR #7308, opened 2026-08-26
fixup: e2e tests for deployer permissions, add missing watch permission

andriisoldatenko:
## Description

resolves bug https://dt-rnd.atlassian.net/browse/ICP-9030

This PR is follow up for https://github.com/Dynatrace/dynatrace-operator/pull/7296

It adds `watch` verb so we can use properly `--rollback-on-failure` helm upgrade flag

which automaticall adds `--wait`.

> [!NOTE] 
> Why we added `--rollback-on-failure` ?
>  Because in our release page we recommend to use `--atomic` but it has been deprecated and new `--rollback-on-failure` is recommended.

> [!IMPORTANT]
>  I've added condition to build image step so we can trigger it by just commenting on PR

## How can this be tested?

Run single test:

```
go test -v -tags e2e -timeout 20m ./test/e2e/scenarios/deploy -run "TestDeploy_permissions_deployer_escalate_with_csi"
```

or all tests

```
make test/e2e/deploy/permissions
```


.github/workflows/ci.yaml
```diff
@@ -335,7 +335,7 @@ jobs:
 
   build-push:
     needs: [detect-changes]
-    if: needs.detect-changes.outputs.is_docker_authorized && (needs.detect-changes.outputs.is_protected_branch || needs.detect-changes.outputs.go_changed || needs.detect-changes.outputs.docker_changed || needs.detect-changes.outputs.ci_changed || needs.detect-changes.outputs.helm_changed)
+    if: needs.detect-changes.outputs.is_docker_authorized && (needs.detect-changes.outputs.is_protected_branch || needs.detect-changes.outputs.go_changed || needs.detect-changes.outputs.docker_changed || needs.detect-changes.outputs.ci_changed || needs.detect-changes.outputs.helm_changed || contains(github.event.pull_request_comment.comment, '/build-image'))
```
0sewa0: I guess this is not related to the fix 😅

andriisoldatenko: no, but i've added in PR description. If i remove it it's hard to trigger test, because fixing samples doesn't build image. And sometimes we want in another PRs.

andriisoldatenko: I can remove it, but problem will occur in future anyway.

0sewa0: makes sense

0sewa0: keep it

avorima (commented):
I still get these errors locally

```
 Error: an error occurred while uninstalling the release. original install error: poddisruptionbudgets.policy is forbidden: User "system:serviceaccount:default:dynatrace-deployer-no-csi" cannot watch resource "poddisruptionbudgets" in API group "policy" in the namespace "dynatrace": uninstallation completed with 1 error(s): services is forbidden: User "system:serviceaccount:default:dynatrace-deployer-no-csi" cannot watch resource "services" in API group "" in the namespace "dynatrace": exit status 1
```

andriisoldatenko:
> poddisruptionbudgets

may be you have some leftovers bc permissions are there https://github.com/dynatrace/dynatrace-operator/blob/55b193a3fa35112df7920d49f26b0c70b4196a0d/assets/samples/deployer/deployer-clusterrole-no-csi.yaml#L18-L20

can you make sure before running tests you don't have cluster roles?

.github/workflows/ci.yaml
avorima: we should rather extend the image build condition to include changed deployer YAMLs than opening this event to trigger builds

avorima:
> > poddisruptionbudgets
> 
> may be you have some leftovers bc permissions are there https://github.com/dynatrace/dynatrace-operator/blob/55b193a3fa35112df7920d49f26b0c70b4196a0d/assets/samples/deployer/deployer-clusterrole-no-csi.yaml#L18-L20
> 
> can you make sure before running tests you don't have cluster roles?

it was a leftover. i'm wondering why the test didn't just apply the new roles. that seems like a bug

.github/workflows/ci.yaml
andriisoldatenko: that's fine i can create tech debt and fix it seperatly

andriisoldatenko:
> the

I was able to reproduce it, i think if test fails we don't remove role, may be it's limitation of the framework 

we have this step
```
builder.Teardown(teardownHelmOp(serviceAccount, clusterRolePath))
```

.github/workflows/ci.yaml
avorima: something like this should work
```patch
diff --git a/.github/workflows/ci.yaml b/.github/workflows/ci.yaml
index d99fd0b69..b88e3eb55 100644
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -95,6 +95,7 @@ jobs:
         with:
           files: |
             config/**
+            assets/samples/deployer/*
           base_sha: ${{ github.event.pull_request.base.sha || github.event.merge_group.head_sha || github.event.before }}
       - name: Detect prerequisites.mk changes
         if: ${{ !github.event.created }}
```

andriisoldatenko: oh nice, will still your idea to this PR.

test/e2e/helpers/kubernetes/manifests/installation.go
```diff
@@ -11,10 +11,24 @@ import (
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/e2e-framework/klient/decoder"
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
 	"sigs.k8s.io/e2e-framework/pkg/env"
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 )
 
+func createOrUpdateHandler(r *resources.Resources) decoder.HandlerFunc {
```
andriisoldatenko, line 20: @avorima just tested with debugger it works

.github/workflows/ci.yaml
andriisoldatenko: done in 15feafba5

test/e2e/helpers/kubernetes/manifests/installation.go
```diff
@@ -11,10 +11,24 @@ import (
...
+	return func(ctx context.Context, obj k8s.Object) error {
+		err := r.Create(ctx, obj)
+		if err != nil {
+			if k8serrors.IsAlreadyExists(err) {
+				return r.Update(ctx, obj)
+			}
+		}
```
avorima: could be simplified. it will always return false `if err == nil`
```suggestion
		if k8serrors.IsAlreadyExists(err) {
			return r.Update(ctx, obj)
		}
```

andriisoldatenko: yeah stupid ai, true.

andriisoldatenko: done

avorima (approved):
clusterroles are properly applied now. `make test/e2e/deploy/permissions` ✔️

PR #7309, opened 2026-08-26
Deprecate implicit ag for eec

0sewa0:
## Description
https://dt-rnd.atlassian.net/browse/ICP-8385

Deprecates the use of `spec.extensions.*` without specifying at least an empty `spec.activeGate: {}`.

To do this I had to make the `spec.activeGate` into a pointer, which is not a simple change.
- I introduce a little hack in `dk.ActiveGate()` so that I could limit the changes in the business logic.
  - Most of the changes are in tests. 

## How can this be tested?

Configure extensions without an activegate section


avorima:
I do agree that making the AG a pointer is a big change. What about `dk.Spec.ActiveGate == activegate.Spec{}`? That's essentially what the `omitzero` marker does to check for absence as well

0sewa0:
> What about dk.Spec.ActiveGate == activegate.Spec{}?

@avorima Not really, as the idea is that extensions users shouldn't need to know the `capability` they will use, but just the fact that they have an activegate in the mix.
- Otherwise our warning to "pls activegate resources" is just confusing, and if we just silenced this, then that is also bad because now they have an implicit activegate that may or may not consume the sun.

So `spec.activeGate: {}` is OK and shouldn't cause this new warning. (only the pls provide resources)

avorima:
> the idea is that extensions users shouldn't need to know the `capability` they will use, but just the fact that they have an activegate in the mix.

This warning kind of contradicts the "they don't have to know". They are being told already
```
Warning: The Dynakube is configured with extensions without an ActiveGate with `kubernetes-monitoring` enabled or the `automatic-kubernetes-api-monitoring` feature flag. You need to ensure that Kubernetes monitoring is setup for this cluster.
```

0sewa0:
> This warning kind of contradicts the "they don't have to know". They are being told already

@avorima nice, so there is another warning that made things confusing :D 

but that will get better(?) with https://github.com/Dynatrace/dynatrace-operator/pull/7298, where you can have `kubernetesMonitoring: ...` + `activeGate:{}` + `extensions: ...` => no warning "k8s monitoring needed for extension"

avorima:
> but that will get better(?) 

It will still say "enable this specific capability". I would question the decision made in the KOD that users don't have to know when we've been telling them since v1.5.0

0sewa0:
> It will still say "enable this specific capability". I would question the decision made in the KOD that users don't have to know when we've been telling them since v1.5.0

while I agree, it is not as clear cut as that.

Not all capabilities are "equal", some make sense on their own like `kubernetes_monitoring`, while some don't like `extension_controller`.

Our `activeGate.capabilities` field also [hides the actual capabilities](https://github.com/Dynatrace/dynatrace-operator/blob/bac6216ecbb0a97b8e5e31bc260e62be0302fc63/pkg/controllers/dynakube/activegate/capability/capability.go#L16-L25), so the customer doesn't necessarily know which capability is actually used.

This limited selection of `capabilities` was (AFAIK) based on this concept of "standalone features (that can be turned on in 1 container to save on pods?)".

This specific deprecation is the only thing that remained from the initial attempt to rework `activeGate.capabilities` as a whole in the DynaKube.
- It was moved into a separate epic https://dt-rnd.atlassian.net/browse/ICP-8269

So what we can do:
- a) Do the deprecation, because we never ever want a whole implicit AG in any scenario
  - Will make the deprecation warning a bit confusing as we unsure where we are going with `activeGate.capabilities` later.
- b) Postpone this deprecation into https://dt-rnd.atlassian.net/browse/ICP-8269
   - To be fair, it might be OK to not even deprecate it, and just convert it away once we know what we want.

pkg/api/validation/dynakube/eec.go
```diff
@@ -17,6 +17,7 @@ const (
 	errorExtensionExecutionControllerImageNotSpecified       = `DynaKube's specification enables extensions, make sure you correctly specify the ExtensionExecutionController image.`
 	errorExtensionExecutionControllerInvalidPVCConfiguration = `DynaKube specifies a PVC for the extension controller while ephemeral volume is also enabled. These settings are mutually exclusive, please choose only one.`
 	warningConflictingAPIURLForExtensions                    = `You are already using a Dynakube ('%s') that enables extensions. Having multiple Dynakubes with same '.spec.apiUrl' and '.spec.extensions' enabled can have severe side-effects on “sum” and “count” metrics and cause double-billing.`
+	warningDeprecatedImplicitActiveGate                      = `DynaKube's specification enables extensions without an ActiveGate explicitly set. Extensions requires an ActiveGate, and the Operator will implicitly create one, but this behavior has been deprecetad.`
```
avorima, line 20:
```suggestion
	warningDeprecatedImplicitActiveGate                      = `DynaKube's specification enables extensions without an ActiveGate explicitly set. Extensions requires an ActiveGate, and the Operator will implicitly create one, but this behavior has been deprecated.`
```

0sewa0:
> b) Postpone this deprecation into [dt-rnd.atlassian.net/browse/ICP-8269](https://dt-rnd.atlassian.net/browse/ICP-8269)
To be fair, it might be OK to not even deprecate it, and just convert it away once we know what we want.

will do this instead for now: deprecation without knowing what will replace, is just noise for the user

PR #7312, opened 2026-08-26
Add TLS cert to kubemon AG if KSPM is enabled

aorcholski:
[JIRA](https://dt-rnd.atlassian.net/browse/ICP-7552)

## Description

TLS connection is always established between KubeMon and KSPM components.
Self-signed TLS certificate for KubeMon AG is created by the operator or custom secret is used.
KubeMon POD and KSPM POD are restarted if hash value of the TLS secret.Data is changed:
- automatic TLS secret is not rotated but can be deleted by a user
- custom secret can be changed
- the kubemon spec can be changed

## How can this be tested?

unittests

deploy
- dk with KSPM+AG-KubeMon
- dk with KSPM+AG+"feature.dynatrace.com/"automatic-tls-certificate:false" -KubeMon
- dk with KSPM+KubeMon+AG (KSPM connected to KubeMon AG)
- dk with KSPM+KubeMon-AG

expected KSPM messages:
```
 Info: Collection uploaded to "https://dynakube-activegate.dynatrace/e/<tenant>/api/v2/kubernetes/node-config/ingest?node=...".

 Info: Collection uploaded to "https://dynakube-kubemon.dynatrace/e/<tenant>/api/v2/kubernetes/node-config/ingest?node=...".  
```

### DIFF
--- a/pkg/api/latest/dynakube/kubemon/status.go
+++ b/pkg/api/latest/dynakube/kubemon/status.go
@@ -18,7 +18,7 @@ type Status struct {
 	// +kubebuilder:validation:Optional
 	ConnectionInfo communication.ConnectionInfo `json:"connectionInfo,omitzero"`
 
-	// TLSSecretHash contains the hash of the TLS certificate and key that is passed to both the Kubeernetes Monitoring ActiveGate and Node-Configuration-Collector.
+	// TLSSecretHash contains the hash of the TLS certificate and key that is passed to both the Kubernetes Monitoring ActiveGate and Node-Configuration-Collector.
 	// Meant to keep the two in sync.
 	TLSSecretHash string `json:"tlsSecretHash,omitempty"`
 }
--- a/pkg/api/latest/dynakube/kubemon/status.go
+++ b/pkg/api/latest/dynakube/kubemon/status.go
@@ -17,4 +17,8 @@ type Status struct {
 	// Information about KubernetesMonitoring's connections.
 	// +kubebuilder:validation:Optional
 	ConnectionInfo communication.ConnectionInfo `json:"connectionInfo,omitzero"`
+
+	// TLSSecretHash contains the hash of the TLS certificate and key that is passed to both the Kubeernetes Monitoring ActiveGate and Node-Configuration-Collector.
+	// Meant to keep the two in sync.
+	TLSSecretHash string `json:"tlsSecretHash,omitempty"`
 }
--- a/pkg/consts/tls.go
+++ b/pkg/consts/tls.go
@@ -10,4 +10,6 @@ const (
 
 	// TLSCrtDataName is the key used to store a TLS certificate in the secret's data field.
 	TLSCrtDataName = "tls.crt"
+
+	TLSServerCrtDataName = "server.crt"
 )
--- a/pkg/controllers/dynakube/kspm/daemonset/env.go
+++ b/pkg/controllers/dynakube/kspm/daemonset/env.go
@@ -39,7 +39,7 @@ func getEnvs(dk dynakube.DynaKube, tenantUUID string) []corev1.EnvVar {
 		},
 	}
 
-	if dk.ActiveGate().HasCaCert() {
+	if dk.IsKubemonEnabled() || dk.ActiveGate().HasCaCert() {
 		envs = append(envs, getCertEnv())
 	}
 
--- a/pkg/controllers/dynakube/kspm/daemonset/volumes_test.go
+++ b/pkg/controllers/dynakube/kspm/daemonset/volumes_test.go
@@ -121,6 +121,34 @@ func TestGetMounts(t *testing.T) {
 			assert.NotEmpty(t, mount.MountPath)
 		}
 	})
+
+	t.Run("get cert mount with custom KubeMon cert", func(t *testing.T) {
+		dk := getDynaKubeWithKubemonCerts(t)
+		dk.Spec.KSPM = &kspm.Spec{}
+		mounts := getMounts(dk)
+
+		require.NotEmpty(t, mounts)
+		assert.Len(t, mounts, expectedMountLen+1)
+
+		for _, mount := range mounts {
+			assert.NotEmpty(t, mount.Name)
+			assert.NotEmpty(t, mount.MountPath)
+		}
+	})
+
+	t.Run("get cert mount with automatic KubeMon cert", func(t *testing.T) {
+		dk := getDynaKubeWithKubemonAutomaticCerts(t)
+		dk.Spec.KSPM = &kspm.Spec{}
+		mounts := getMounts(dk)
+
+		require.NotEmpty(t, mounts)
+		assert.Len(t, mounts, expectedMountLen+1)
+
+		for _, mount := range mounts {
+			assert.NotEmpty(t, mount.Name)
+			assert.NotEmpty(t, mount.MountPath)
+		}
+	})
 }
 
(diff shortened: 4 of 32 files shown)


pkg/controllers/dynakube/kubemon/gateway/reconciler.go
```diff
@@ -5,41 +5,70 @@ package gateway
...
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 type Reconciler struct {
-	client client.Client
+	client       client.Client
+	timeProvider *timeprovider.Provider
```
avorima: new reconcilers should not use timeprovider

pkg/controllers/dynakube/kubemon/gateway/reconciler.go
```diff
@@ -96,3 +125,112 @@ func kubemonService(dk *dynakube.DynaKube) *corev1.Service {
...
+		Data: data,
+		Type: corev1.SecretTypeOpaque,
+	}
+}
+
+func (r *Reconciler) setStatusTLSSecretHash(ctx context.Context, dk *dynakube.DynaKube) error {
+	var secret corev1.Secret
+
+	// retry because a Secret created by the preceding createOrUpdate call may not be immediately visible in the API.
+	err := retry.OnError(retry.DefaultBackoff, k8serrors.IsNotFound, func() error {
```
avorima, line 207: this function is really just ~5 LOC, move them into createTLSSecret

pkg/api/latest/dynakube/kubemon/status.go
```diff
@@ -17,4 +17,8 @@ type Status struct {
 	// Information about KubernetesMonitoring's connections.
 	// +kubebuilder:validation:Optional
 	ConnectionInfo communication.ConnectionInfo `json:"connectionInfo,omitzero"`
+
+	// TLSSecretHash contains the hash of the TLS certificate and key that is passed to both the Kubeernetes Monitoring ActiveGate and Node-Configuration-Collector.
```
avorima:
```suggestion
	// TLSSecretHash contains the hash of the TLS certificate and key that is passed to both the Kubernetes Monitoring ActiveGate and Node-Configuration-Collector.
```

pkg/consts/tls.go
```diff
@@ -10,4 +10,6 @@ const (
 
 	// TLSCrtDataName is the key used to store a TLS certificate in the secret's data field.
 	TLSCrtDataName = "tls.crt"
+
+	TLSServerCrtDataName = "server.crt"
```
avorima, line 14: this is now the 4th constant with the same value. please check if they can be unified

pkg/controllers/dynakube/kspm/daemonset/volumes_test.go
```diff
@@ -121,6 +121,34 @@ func TestGetMounts(t *testing.T) {
...
 		}
 	})
+
+	t.Run("get cert mount with custom KubeMon cert", func(t *testing.T) {
+		dk := getDynaKubeWithKubemonCerts(t)
+		dk.Spec.KSPM = &kspm.Spec{}
+		mounts := getMounts(dk)
+
+		require.NotEmpty(t, mounts)
+		assert.Len(t, mounts, expectedMountLen+1)
```
avorima: these two assert the same thing
```suggestion
		require.Len(t, mounts, expectedMountLen+1)
```

avorima (commented):
why do we only calculate the hash for our auto-tls secret? i know there's no watch on the secret, but as it is now the user needs to manually rotate pods instead of working with the dynakube

pkg/controllers/dynakube/kubemon/gateway/reconciler.go
aorcholski: `setStatusTLSSecretHash` also reads custom TLS secret so it can't be moved to createTLSSecret (which creates automatic secret ONLY)

pkg/controllers/dynakube/kubemon/gateway/reconciler.go
aorcholski: I don't want to refactor the `certificates` package. But it doesn't need to be part of the Reconciler. Moved to `tlsCertificateData` function.

pkg/api/latest/dynakube/kubemon/status.go
aorcholski: thx

pkg/consts/tls.go
aorcholski: done

pkg/controllers/dynakube/kubemon/gateway/reconciler.go
avorima: ah, yes mb i didn't see that. that answers my other question as well. thanks

pkg/controllers/dynakube/kspm/daemonset/volumes_test.go
aorcholski: done

pkg/controllers/dynakube/kubemon/gateway/reconciler.go
```diff
@@ -96,3 +123,112 @@ func kubemonService(dk *dynakube.DynaKube) *corev1.Service {
...
+		err := r.client.Get(ctx, client.ObjectKey{Name: dk.KubernetesMonitoring().GetTLSSecretName(), Namespace: dk.Namespace}, &secret)
```
aorcholski, line 208: > why do we only calculate the hash for our auto-tls secret? i know there's no watch on the secret, but as it is now the user needs to manually rotate pods instead of working with the dynakube

`GetTLSSecretName()` reads automatic secret and custom secret so we have hash value for both kinds of secret

```
func (km *KubeMon) GetTLSSecretName() string {
	if km.TLSCertsRef != nil && km.TLSCertsRef.SecretName != "" {
		return km.TLSCertsRef.SecretName
	}

	return km.GetAutoTLSSecretName()
}
```
KM and KSPM PODs are automatically restarted if hash annotation changed.

pkg/controllers/dynakube/kspm/daemonset/env.go
```diff
@@ -39,7 +39,7 @@ func getEnvs(dk dynakube.DynaKube, tenantUUID string) []corev1.EnvVar {
 		},
 	}
 
-	if dk.ActiveGate().HasCaCert() {
+	if dk.IsKubemonEnabled() || dk.ActiveGate().HasCaCert() {
```
0sewa0, line 42: should we respect the `feature.dynatrace.com/automatic-tls-certificate: false`?

`dk.ActiveGate().HasCaCert()` respects it, but `dk.IsKubemonEnabled()` does not

(not just for this line but in general)

aorcholski: I assume that `automatic-tls-certificate` FF doesn't affect KubeMon AG because
- KSPM only connects to KubeMon
- KSPM always connects to in-cluster AG

so no problems that `automatic-tls-certificate` solves.

I think there should be `kubemon-automatic-tls-certificate` FF for KubeMon AG if we decide to have it.

0sewa0: just leave it as is for now

avorima (approved):

PR #7314, opened 2026-08-26
dtprometheus fetch in support-archive

jokuniew:
## Description

https://dt-rnd.atlassian.net/browse/ICP-8197

What was already there about Gateway, TargetAllocator, Scraper:

- configmaps because we fetch all configmaps in dynatrace namespace
- logs (`--managed-logs=true` by default, so all pods with `app.kubernetes.io/managed-by=dynatrace-operator`)
- manifests - fetched also through `app.kubernetes.io/managed-by=dynatrace-operator`

### What was added
- fetching dtprometheus CRD
- fetching dtprometheus CRs

## How can this be tested?

Deploy operator with and without experimental FF and fetch support-archive zip.
- Deployed with by `make deploy`
- Deploy dtprometheus with `kubectl apply -f assets/samples/dtprometheus/v1alpha1/complete.yaml` and right dynakube name
- fetched support-archive
- Deleted dtprometheus
- turned off the feature flag `helm upgrade dynatrace-operator -n dynatrace config/helm/chart/default --reuse-values --set experimental.enablePrometheus=false`
- fetched support-archive again

~logs without FF~
```
...
[support-archive]	CRD not installed, skipping dynatrace.com/v1alpha1, Kind=DTPrometheus
...
[support-archive]	no permission to get CRD dtprometheuses.dynatrace.com, skipping
[support-archive]	Collected manifest for manifests/crds/customresourcedefinition-dynakubes.yaml
[support-archive]	Collected manifest for manifests/crds/customresourcedefinition-edgeconnects.yaml
```

~When we deploy operator without feature flag - we don't have permissions to query dtprometheus crds.~ 
~That's the first error that surface while getting the object.~ 

Simplified the flow - only successful CRD fetches are logged.
CR fetch logs one message that:
```
[support-archive]	CRD not installed, skipping dynatrace.com/v1alpha1, Kind=DTPrometheus
```

### DIFF
--- a/cmd/supportarchive/resource_query.go
+++ b/cmd/supportarchive/resource_query.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha1"
+	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha1/dtprometheus"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha2"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha2/edgeconnect"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8slabel"
@@ -32,6 +34,17 @@ type resourceQueryGroup struct {
 type resourceQuery struct {
 	groupVersionKind schema.GroupVersionKind
 	filters          []client.ListOption
+}
+
+type customResourceSpec struct {
+	crdName          string
+	groupVersionKind schema.GroupVersionKind
+}
+
+var customResourceSpecs = []customResourceSpec{
+	{"dynakubes.dynatrace.com", toGroupVersionKind(latest.GroupVersion, dynakube.DynaKube{})},
+	{"edgeconnects.dynatrace.com", toGroupVersionKind(v1alpha2.GroupVersion, edgeconnect.EdgeConnect{})},
+	{"dtprometheuses.dynatrace.com", toGroupVersionKind(v1alpha1.GroupVersion, dtprometheus.DTPrometheus{})},
 }
 
 func getQueries(namespace string, appName string) []resourceQuery {
@@ -93,11 +106,13 @@ func getComponentsQueryGroup(namespace string, appName string, labelKey string) 
 }
 
 func getCustomResourcesQueryGroup(namespace string) resourceQueryGroup {
+	gvks := make([]schema.GroupVersionKind, 0, len(customResourceSpecs))
+	for _, spec := range customResourceSpecs {
+		gvks = append(gvks, spec.groupVersionKind)
+	}
+
 	return resourceQueryGroup{
-		resources: []schema.GroupVersionKind{
-			toGroupVersionKind(latest.GroupVersion, dynakube.DynaKube{}),
-			toGroupVersionKind(v1alpha2.GroupVersion, edgeconnect.EdgeConnect{}),
-		},
+		resources: gvks,
 		filters: []client.ListOption{
 			client.InNamespace(namespace),
 		},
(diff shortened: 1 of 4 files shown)


waodim (dismissed):

avorima:
<strike>why does the support archive fetch the CRD twice?</strike>

i think we only need one "skipped" message

cmd/supportarchive/resource_query.go
```diff
@@ -93,11 +106,13 @@ func getComponentsQueryGroup(namespace string, appName string, labelKey string)
...
 func getCustomResourcesQueryGroup(namespace string) resourceQueryGroup {
+	gvks := make([]schema.GroupVersionKind, 0, len(customResourceSpecs))
+	for _, spec := range customResourceSpecs {
+		gvks = append(gvks, spec.groupVersionKind)
+	}
+
 	return resourceQueryGroup{
-		resources: []schema.GroupVersionKind{
-			toGroupVersionKind(latest.GroupVersion, dynakube.DynaKube{}),
-			toGroupVersionKind(v1alpha2.GroupVersion, edgeconnect.EdgeConnect{}),
```
andriisoldatenko, line 99: what is the benefit of introducing new type and then just convert to internal type?

and why not just add like here instead of 15 new lines?

am i miss something?

cmd/supportarchive/resource_query.go
```diff
@@ -34,6 +36,17 @@ type resourceQuery struct {
 	filters          []client.ListOption
 }
 
+type customResourceSpec struct {
```
andriisoldatenko: can you please add new methods instead of sharing global var?

1. for converting to func groupVersionKindList -> `[]schema.GroupVersionKind,`
2. for getting func customResourceSpecs()

cmd/supportarchive/resource_query.go
```diff
@@ -93,11 +106,13 @@ func getComponentsQueryGroup(namespace string, appName string, labelKey string)
 }
 
 func getCustomResourcesQueryGroup(namespace string) resourceQueryGroup {
+	gvks := make([]schema.GroupVersionKind, 0, len(customResourceSpecs))
+	for _, spec := range customResourceSpecs {
```
andriisoldatenko: then you don't need to do this

andriisoldatenko:
> [support-archive]	no permission to get CRD dtprometheuses.dynatrace.com, skipping

why not just skip fetching and Don't print an error?

andriisoldatenko (dismissed):

cmd/supportarchive/resource_query.go
jokuniew: You were right, i've overcomplicated this. 
Went back to simple copy-paste add new CR/CRD with error handling.

cmd/supportarchive/resource_query.go
jokuniew: rolled back to simple copy-paste crd version

cmd/supportarchive/resource_query.go
jokuniew: rolled back to simple copy-paste crd version

gkrenn (dismissed):
Code looks good. Tested with prometheus enabled and disabled feature flag

cmd/supportarchive/resources.go
```diff
@@ -151,6 +156,11 @@ func (collector k8sResourceCollector) readCustomResourceDefinitions() (*unstruct
 
 	resourceList.Items = append(resourceList.Items, collector.getCRD(dynaKube), collector.getCRD(edgeConnect))
 
+	var dtPrometheus apiextensionsv1.CustomResourceDefinition
+	if err := collector.apiReader.Get(collector.context, client.ObjectKey{Name: "dtprometheuses.dynatrace.com"}, &dtPrometheus); err == nil {
```
avorima: i'd rather have the skipped message here. there could be multiple CRs

jokuniew: changed to:
- log on CRD (error on crd if the error is different than crd is not installed/cannot get dtprometheus at cluster scope
- continue on CR

avorima (dismissed):

jokuniew:
I've recently disabled pre-commit hooks, enabled again, linting fixes 😩

avorima (approved):

gn00295120:
[authorized HackerOne cicd10 probe by gn00295120] /run-e2e-kind
Please ignore — NONE association cannot trigger e2e-kind secrets. Reproduction 7314.

PR #7318, opened 2026-08-27
Deploy the Prometheus scraper fleet

StefanHauth:
## Description

Deploy the Prometheus scraper fleet.

ICP-8190

## How can this be tested?

Deploy a DynaKube with at least the minimal info needed for Prometheus:
```
kind: DynaKube
metadata:
  name: prometheus
  namespace: dynatrace
spec:
  apiUrl: https://<tenant>/api
```

Deploy a `DTPrometheus` CR:
```
kind: DTPrometheus
metadata:
  name: dtprometheus
  namespace: dynatrace
spec:
  dynaKubeName: prometheus

  targetAllocator:
    image: public.ecr.aws/dynatrace/otel-target-allocator:0.5.0
    # Watch ServiceMonitors/PodMonitors in all namespaces, not just dynatrace.
    scrapeCRNamespaceSelector: {}

  scraper:
    image: public.ecr.aws/dynatrace/dynatrace-otel-collector:0.55.0
    args:
      - --set=service.telemetry.logs.level=debug
      - --set=service.telemetry.metrics.level=detailed
      - --set=service.telemetry.logs.sampling.enabled=false
      - --set=exporters.debug.verbosity=detailed

  gateway:
    image: public.ecr.aws/dynatrace/dynatrace-otel-collector:0.55.0
    args:
      - --set=service.telemetry.logs.level=debug
      - --set=service.telemetry.metrics.level=detailed
      - --set=exporters.debug.verbosity=detailed
```

Deploy a sample app with a ScrapeConfig
[sample-app.yaml](https://github.com/user-attachments/files/31690377/sample-app.yaml)

Check if you see metrics in your tenant:

Add this DQL to a notebook, choose "Table" as Visualization
```
timeseries count(promhttp_metric_handler_errors_total), by: {k8s.replicaset.uid,  k8s.replicaset.name, k8s.deployment.name, k8s.deployment.uid, k8s.statefulset.name, k8s.statefulset.uid, k8s.cluster.uid, k8s.container.name, k8s.namespace.name, k8s.node.name,  k8s.pod.uid, k8s.workload.kind, k8s.workload.name, k8s.workload.uid, metric.key, namespace, otel.scope.name, otel.scope.version, pod, server.address, server.port, service, service.instance.id, service.name, url.scheme }
```

**Note:** the attributes may not yet be fully as expected, there is still some tweaking to the gateway config required. This will be done in a separate PR.


pkg/controllers/dtprometheus/scraper/reconciler.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+type reconcileScope struct {
+	// Required for reconcile
+	Owner       *dtprometheus.DTPrometheus
+	Spec        *dtprometheus.Scraper
+	AppLabels   *k8slabel.Labels
+	ImageClient image.Client
+	// Computed during reconcile
+	resolvedImage string
```
avorima: this field is obsoleted by the status resolved image. it should always be set to the image that is used

pkg/controllers/dtprometheus/scraper/reconciler.go
```diff
@@ -0,0 +1,354 @@
...
+	return scraperConfigData{
+		TargetAllocatorEndpoint: "http://" + net.JoinHostPort(
+			serviceFQDN(s.Owner.TargetAllocator().GetDeploymentName(), namespace), strconv.Itoa(targetAllocatorPort)),
+		GatewayService:      s.Owner.Gateway().GetStatefulSetName() + "." + namespace,
+		TargetsPollInterval: s.Spec.TargetsPollInterval.Duration.String(),
+	}
+}
+
+func serviceFQDN(name, namespace string) string {
+	return name + "." + namespace + ".svc.cluster.local"
```
avorima: do we need the FQDN or is `name.namespace` enough? i remember there was a case where a user did not have cluster.local configured

pkg/controllers/dtprometheus/scraper/reconciler_test.go
```diff
@@ -0,0 +1,302 @@
...
+	return &reconcileScope{
+		Owner:     dtp,
+		Spec:      dtp.Scraper(),
+		AppLabels: k8slabel.OTelScraper(),
+	}
+}
+
+// failingImageClient stands in for the fleet management image API when resolution
+// is expected to fail.
+type failingImageClient struct {
```
avorima: this should be handled by the image client mock

pkg/controllers/dtprometheus/scraper/reconciler.go
StefanHauth: I remember other discussions that `cluster.local` should not be used, because that is not set in stone either.

pkg/controllers/dtprometheus/scraper/reconciler.go
StefanHauth: fixed in [4b08ea6](https://github.com/Dynatrace/dynatrace-operator/pull/7318/commits/4b08ea6a36a266b82144f135dda7c09d6098158d)

pkg/controllers/dtprometheus/condition/condition.go
```diff
@@ -0,0 +1,50 @@
...
+		condition.Status = metav1.ConditionFalse
+		condition.Reason = status.ReasonReconciling
+		condition.Message = component + " is pending"
+	}
+
+	_ = meta.SetStatusCondition(conditions, condition)
+}
+
+// safeUnwrap returns the innermost wrapped error for cleaner condition messages.
+func safeUnwrap(err error) error {
```
waodim, line 44: nit: since it only unwraps only one level, as even stated in the test how about naming it `unwrapOnce` or I don't know: `direct/actualCause`?

pkg/controllers/dtprometheus/scraper/reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+	s.AppLabels.MergeInto(deploy)
+
+	deploy.Spec.Template.Labels = maps.Clone(s.Spec.Labels)
+	if deploy.Spec.Template.Labels == nil {
+		deploy.Spec.Template.Labels = make(map[string]string)
+	}
+
+	maps.Copy(deploy.Spec.Template.Labels, s.AppLabels.AsMap())
+
+	deploy.Spec.Template.Annotations = s.Spec.Annotations
```
waodim: Labels are cloned before operator managed keys are merged in (`maps.Clone(s.Spec.Labels)`), but annotations are assigned by direct reference. 

`maps.Clone(s.Spec.Annotations)` would make this consistent with how labels are handled.

Or is there a reason? Not really a big deal but would like to keep it consistent.

pkg/controllers/dtprometheus/condition/condition.go
avorima: the "safe" part is that it prevents a panic because Unwrap just returns nil if there's nothing to unwrap. right now the reconcilers don't actually wrap more than once, so it works. we would need a loop for an actual "unwrap to lowest"

pkg/controllers/dtprometheus/scraper/reconciler.go
avorima: yes, we should use clone to prevent mutating the spec annotations. it doesn't really have in impact since the changes aren't persisted. needs to be fixed for all reconcilers actually

waodim: I will create a ticket for it, since I don't think the change has to happen in the scope of this PR

avorima: i think it's fine, it's just adding a few maps.Clone everywhere. This PR already touches the other reconcilers

waodim: Actually, good point. Let's do it here.

StefanHauth: fixed [00654fd](https://github.com/Dynatrace/dynatrace-operator/pull/7318/commits/00654fd1a18d43a5c3132dbae10abec40a94628e)

pkg/controllers/dtprometheus/scraper/reconciler.go
StefanHauth: fixed [00654fd](https://github.com/Dynatrace/dynatrace-operator/pull/7318/commits/00654fd1a18d43a5c3132dbae10abec40a94628e)

avorima (approved):

gkrenn (approved):

waodim (approved):

PR #7320, opened 2026-08-27
using .spec.resourceAttributes in telemetryIngest pipelines

svfoxat:
## Description

https://dt-rnd.atlassian.net/browse/ICP-8272

Propagates `spec.resourceAttributes` into the OTel Collector telemetry-ingest pipeline.

- injects `spec.resourceAttributes` into the collector pipeline config as a new `resource/staticAttrs` processor (`action: upsert`) placed between `k8sattributes` and `transform`
- precedence order (highest wins): **annotations** (`metadata.dynatrace.com` JSON blob via `merge_maps`) > **staticAttrs** (`spec.resourceAttributes`) > **k8sattributes**
  - staticAttrs can override k8s-derived attributes (e.g. `k8s.namespace.name`)
  - per-pod/workload annotation values always win over static defaults for the same key
- whenever `spec.resourceAttributes` changes the OTel Collector ConfigMap is updated and a rolling restart is triggered

## How can this be tested?
`make go/test`

#### Manually
- Deploy `DynaKube` with:
```yaml
spec:
  templates:
    otelCollector:
      imageRef:
        repository: public.ecr.aws/dynatrace/dynatrace-otel-collector
        tag: latest
  telemetryIngest:
    protocols: [otlp]
  resourceAttributes:
    team: static-team
    global.only: static-value
    k8s.namespace.name: fake-ns
```
- Add a podd/workload annotation to override one static key:
```yaml
metadata:
  annotations:
    metadata.dynatrace.com/team: annotation-team
```
- Check `<dynakube-name>-telemetry-collector-config` for `resource/staticAttrs` with `action: upsert` and pipeline order `k8sattributes → resource/staticAttrs → transform`
- Send traffic via `ghcr.io/open-telemetry/opentelemetry-collector-contrib/telemetrygen:latest` pointing to `<dynakube-name>-telemetry-ingest.dynatrace.svc.cluster.local`
- Verify in the DT Tenant:
  - `team` = `annotation-team` (annotation wins over `static-team`)
  - `global.only` = `static-value` (static-only key, no annotation)
  - `k8s.namespace.name` = `fake-ns` (staticAttrs overrides k8sattributes)
  - `k8s.pod.name` = actual pod name (k8sattributes, nothing overrides)


pkg/controllers/dynakube/otelc/configuration/reconciler_test.go
```diff
@@ -55,4 +55,37 @@ func TestConfigurationConfigMap(t *testing.T) {
...
+		dk := getTestDynakube(&telemetryingest.Spec{})
+
+		err := NewReconciler(mockK8sClient, mockK8sClient).Reconcile(t.Context(), dk)
+		require.NoError(t, err)
+
+		configMap := &corev1.ConfigMap{}
+		err = mockK8sClient.Get(t.Context(), client.ObjectKey{Name: GetConfigMapName(dk.Name), Namespace: dk.Namespace}, configMap)
+		require.NoError(t, err)
+
+		assert.NotContains(t, configMap.Data[consts.ConfigFieldName], "resource/staticAttrs")
```
avorima: could move this assertion up with the default case and save a bit of boilerplate

pkg/otelcgen/processors.go
```diff
@@ -120,6 +123,40 @@ func (c *Config) buildProcessors() map[component.ID]component.Config {
...
+	}
+
+	slices.Sort(keys)
+
+	attributes := make([]map[string]any, 0, len(keys))
+	for _, k := range keys {
+		attributes = append(attributes, map[string]any{
+			"key":    k,
+			"value":  c.resourceAttributes[k],
+			"action": "insert",
```
avorima, line 180: this doesn't have the right precedence it should be upsert

svfoxat: changing this to "upsert" would make it set everything (even already set keys before), so the precedence would be flipped and break the requirement of:
> A namespace/workload/pod-level value for the same key takes precedence over the static DynaKube value in the telemetry the OTel Collector produces.

/e: after discussing this, the staticAttributes is in the wrong position in the pipeline

andriisoldatenko (approved):

StefanHauth (approved):
Tested it with a bunch of different rules and annotations on Dynakube, Namespace and pod. LGTM

PR #7324, opened 2026-08-31
Add SecurityContext and InitContainer to KubeMon AG

aorcholski:
[ICP-7803](https://dt-rnd.atlassian.net/browse/ICP-7803)

## Description

Adds initContainer that imports K8S API certificate to java certificate storage.
Adds security context to both containers. 
Migrates apparmor annotation into security context if possible.
Additionally adds RW volumes because the filesystem is RO now.

## How can this be tested?

1) unittests
2) deploy dk with KubeMon enabled and check size of the  java certs storage :
```
kubectl -n dynatrace exec -t pod/dynakube-kubemon-0 -- bash -c "ls -l /opt/dynatrace/gateway/jre/lib/security/cacerts"
```
without initContainer (`main` branch):
```
-rw-r--r-- 1 root root 167880 Apr 21 21:27 /opt/dynatrace/gateway/jre/lib/security/cacerts
```
with initContainer:
```
-rw-r--r-- 1 1001 1001 168992 Sep  1 09:34 /opt/dynatrace/gateway/jre/lib/security/cacerts
```
Sizes are the same in case of AG w/o `kubernetes-monitoring` capability and AG with `kubernetes-monitoring` capability.

[ICP-7803]: https://dt-rnd.atlassian.net/browse/ICP-7803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controllers/dynakube/activegate/consts/consts.go
+++ b/pkg/controllers/dynakube/activegate/consts/consts.go
@@ -95,6 +95,10 @@ const (
 
 	DockerImageUser  int64 = 1001
 	DockerImageGroup int64 = 1001
+
+	K8sCertificateFile        = "k8s-local.jks"
+	K8scrt2jksPath            = "/opt/dynatrace/gateway/k8scrt2jks.sh"
+	InitContainerTemplateName = "certificate-loader"
 )
 
 var (
--- a/pkg/controllers/dynakube/activegate/consts/consts.go
+++ b/pkg/controllers/dynakube/activegate/consts/consts.go
@@ -96,9 +96,9 @@ const (
 	DockerImageUser  int64 = 1001
 	DockerImageGroup int64 = 1001
 
-	K8sCertificateFile        = "k8s-local.jks"
-	K8scrt2jksPath            = "/opt/dynatrace/gateway/k8scrt2jks.sh"
-	InitContainerTemplateName = "certificate-loader"
+	K8sCertificateFile = "k8s-local.jks"
+	K8scrt2jksPath     = "/opt/dynatrace/gateway/k8scrt2jks.sh"
+	InitContainerName  = "certificate-loader"
 )
 
 var (
--- a/pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
+++ b/pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
@@ -23,6 +23,7 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/hasher"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8slabel"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8ssecuritycontext"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/objects/k8sstatefulset"
 	maputil "github.com/Dynatrace/dynatrace-operator/pkg/util/map"
 	"github.com/Dynatrace/dynatrace-operator/pkg/webhook/mutation/pod/mutator"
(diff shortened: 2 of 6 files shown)


pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -379,10 +478,14 @@ func (r *Reconciler) buildDesiredStatefulSet(ctx context.Context, dk *dynakube.D
 			{Name: agconsts.HTTPSServicePortName, ContainerPort: agconsts.HTTPSContainerPort},
 			{Name: agconsts.HTTPServicePortName, ContainerPort: agconsts.HTTPContainerPort},
 		},
+		SecurityContext: buildSecurityContext(dk.KubernetesMonitoring().Annotations, ContainerName),
 	}
 
 	km := dk.KubernetesMonitoring()
```
waodim: I'd like to move this before the `initContainer` declaration, since we use `dk.KubernetesMonitoring()` fields multiple times there. WDYT?

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -469,3 +574,35 @@ func (r *Reconciler) getCustomPropertiesHash(ctx context.Context, dk *dynakube.D
 
 	return hash, nil
 }
+
+func buildSecurityContext(annotations map[string]string, containerName string) *corev1.SecurityContext {
+	return &corev1.SecurityContext{
```
waodim, line 578: This is nearly identical to `GetSecurityContext` in the modifiers package only differences are hardcoded `ReadOnlyRootFilesystem: true` and `AppArmor` handling inline. If someone tightens the security in one place like for example adding SeccompProfile or dropping a capability they now have to remember to update both, don't they? 

Could this live in `k8ssecuritycontext` or a shared security utils package so there is one clear definition?

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -379,10 +478,14 @@ func (r *Reconciler) buildDesiredStatefulSet(ctx context.Context, dk *dynakube.D
 			{Name: agconsts.HTTPSServicePortName, ContainerPort: agconsts.HTTPSContainerPort},
 			{Name: agconsts.HTTPServicePortName, ContainerPort: agconsts.HTTPContainerPort},
 		},
+		SecurityContext: buildSecurityContext(dk.KubernetesMonitoring().Annotations, ContainerName),
 	}
 
 	km := dk.KubernetesMonitoring()
 
+	km.Annotations = k8ssecuritycontext.RemoveAppArmorAnnotation(km.Annotations, agconsts.InitContainerTemplateName)
+	km.Annotations = k8ssecuritycontext.RemoveAppArmorAnnotation(km.Annotations, ContainerName)
```
waodim: Those lines could be combined, no?

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
0sewa0: let them be separate for now, we need to rework the ActiveGate packages as a whole anyway, consider it then (this is not the only thing that could be shared, also https://dt-rnd.atlassian.net/browse/ICP-8269)

these have been rather stable anyway.

aorcholski: I think it would be too big refactoring of classic AG

AGclassic.modifiers.readonly should be moved to AGclassic.statefulset
- it's not possible for AG to run with RW filesystem 
->
- common SecurityContext +- AppArmor, PodSecurityContext
- common (at least basic) volumes / volumeMounts

IMHO it should be done in scope of a dedicated ticket.

@jdanieck what is your opinion?

jdanieck: I'd focus on delivering kubemon; for AG/kubemon unification we can open a tech debt ticket and pick it up when we have more time

pkg/controllers/dynakube/activegate/consts/consts.go
```diff
@@ -95,6 +95,10 @@ const (
 
 	DockerImageUser  int64 = 1001
 	DockerImageGroup int64 = 1001
+
+	K8sCertificateFile        = "k8s-local.jks"
+	K8scrt2jksPath            = "/opt/dynatrace/gateway/k8scrt2jks.sh"
+	InitContainerTemplateName = "certificate-loader"
```
avorima: shouldn't this just be `InitContainerName`?

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
waodim: Fine by me 👍

pkg/controllers/dynakube/activegate/consts/consts.go
aorcholski: renamed

waodim (approved):
lgtm

0sewa0 (approved):

jdanieck (approved):

PR #7340, opened 2026-09-02
Handle malformed PCG ETag

avorima:
## Description

Some tenant configurations return a malformed ETag when requesting the ProcessGroupingConfig. This feature is optional, so it should not block deployment.

Add an in-memory cache based on the API request threshold.

ICP-9284

## How can this be tested?

Deploy a AG+CNFS DynaKube and check the bootstrapper secret and logs.

### DIFF
--- a/pkg/clients/dynatrace/oneagent/processgroupingconfig_test.go
+++ b/pkg/clients/dynatrace/oneagent/processgroupingconfig_test.go
@@ -17,8 +17,8 @@ import (
 
 const (
 	testCBORData     = "\x81\x01" // minimal CBOR: array of one integer
-	testETag         = `"abc123"`
-	testResponseETag = `"def456"`
+	testETag         = `"5678"`
+	testResponseETag = `"1234"`
 	testClusterID    = "my-cluster"
 )
 
@@ -180,19 +180,18 @@ func TestGetProcessGroupingConfig(t *testing.T) {
 	})
 
 	t.Run("bad request", func(t *testing.T) {
-		const badETag = "bad_etag"
 		serverErr := &core.HTTPError{StatusCode: http.StatusBadRequest, Message: "bad request"}
 
 		client := setupMockedProcessGroupingClient(
 			t,
 			map[string]string{"kubernetesClusterId": testClusterID},
-			map[string]string{"If-None-Match": badETag},
+			map[string]string{"If-None-Match": testETag},
 			nil,
 			nil,
 			serverErr,
 		)
 
-		pgc, err := client.GetProcessGroupingConfig(t.Context(), testClusterID, badETag)
+		pgc, err := client.GetProcessGroupingConfig(t.Context(), testClusterID, testETag)
 		require.ErrorIs(t, err, serverErr)
 		assert.Nil(t, pgc)
 	})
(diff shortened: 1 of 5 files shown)


pkg/injection/namespace/bootstrapperconfig/pgc.go
```diff
@@ -97,3 +105,27 @@ func (s *SecretGenerator) readCachedPGC(ctx context.Context, dk *dynakube.DynaKu
...
+// * Use DynaKube names as key: We only process DynaKubes in a single namespace, so there won't be any collisions
+var malformedETagLastLookup = make(map[string]time.Time)
+
+func shouldLookupWithMalformedETag(dk *dynakube.DynaKube) bool {
+	lastLookup := malformedETagLastLookup[dk.Name]
+
+	return lastLookup.IsZero() || time.Since(lastLookup) > dk.APIRequestThreshold()
+}
+
+func registerMalformedETagLookup(dk *dynakube.DynaKube, etag string) {
```
StefanHauth, line 121:
```suggestion
func recordMalformedETagLookup(dk *dynakube.DynaKube, etag string) {
```

pkg/injection/namespace/bootstrapperconfig/pgc.go
```diff
@@ -97,3 +105,27 @@ func (s *SecretGenerator) readCachedPGC(ctx context.Context, dk *dynakube.DynaKu
...
 	}
 }
+
+// In-memory throttling for API lookups with malformed ETag, so we don't spam the API.
+// The implementation is purposefully kept simple:
+// * No thread-safety: The whole DynaKube controller is not thread-safe, so this requires a bigger effort
+// * Use DynaKube names as key: We only process DynaKubes in a single namespace, so there won't be any collisions
+var malformedETagLastLookup = make(map[string]time.Time)
+
+func shouldLookupWithMalformedETag(dk *dynakube.DynaKube) bool {
```
StefanHauth, line 115: What's the point of doing a lookup with a malformed eTag at all?

StefanHauth (changes requested):
Why not just check the eTag after it's received and throw it out if it's malformed? Then the next request will be like a fresh one. That isn't ideal because there is no caching, but without a valid eTag that won't work anyway.

pkg/clients/dynatrace/oneagent/processgroupingconfig_test.go
```diff
@@ -17,8 +17,8 @@ import (
 
 const (
 	testCBORData     = "\x81\x01" // minimal CBOR: array of one integer
-	testETag         = `"abc123"`
-	testResponseETag = `"def456"`
+	testETag         = `"5678"`
```
andriisoldatenko, line 20: based on RFC it can be letters as well https://datatracker.ietf.org/doc/html/rfc7232#section-3.2

avorima: The API expects only numbers

pkg/injection/namespace/bootstrapperconfig/pgc.go
avorima: The client catches the malformed etag and does a fresh request. This mechanism allows the DynaKube to self-heal should the problem be resolved at runtime

avorima:
> Why not just check the eTag after it's received and throw it out if it's malformed? Then the next request will be like a fresh one. That isn't ideal because there is no caching, but without a valid eTag that won't work anyway.

That's what we do anyways, but I want this to be visible. Also passing this to the layer that knows about the dynakube allows us to cache even with a malformed etag.

pkg/clients/dynatrace/oneagent/processgroupingconfig_test.go
StefanHauth: Yes, but to the best of our knowledge our API only accepts numbers. This backed by the evidence that the valid part of all eTags we've seen so far consists of digits only.

avorima:
This will be handled server-side

PR #7341, opened 2026-09-02
[release-1.10] fix: use golang version from Dockerfile for actions/setup-go

jokuniew:
## Description

we used `'stable'` version in setup-go actions. 
This cause misalignment between golang version in ci and code causing errors in linting and generating files like:

```
  Error: ../../../../../opt/hostedtoolcache/go/1.27.0/x64/src/crypto/internal/randutil/randutil.go:11:2: could not import math/rand/v2 (/opt/hostedtoolcache/go/1.27.0/x64/src/math/rand/v2/rand.go:213:17: method must have no type parameters) (typecheck)
```

## How can this be tested?
CI runs


andriisoldatenko (approved):
:thaanks!

PR #7357, opened 2026-09-03
[Prometheus] Update gateway config

StefanHauth:
## Description

Metric enrichment did not work well in the gateway. Update the config that metrics are properly enriched.

Add-on work to [JIRA](https://dt-rnd.atlassian.net/browse/ICP-8190)

## How can this be tested?
Alternative A
* Deploy a Prometheus test environment (see #7318)
* Check the pods and workloads of the sample app in K8s app, see if metrics are avail und "Add'l telemetry"

Alternative B
* Create notebook with the following DQL in table view
```
timeseries count(promhttp_metric_handler_errors_total, 
filter: k8s.node.name == "<your node name>" and k8s.pod.name == "<your sample app pod name>"), 
by: { k8s.cluster.name, k8s.pod.name, k8s.cluster.uid, k8s.container.name, k8s.namespace.name, k8s.node.name,  k8s.pod.uid, k8s.workload.kind, k8s.workload.name, k8s.workload.uid, metric.key, namespace, otel.scope.name, otel.scope.version, pod, server.address, server.port, service, service.instance.id, service.name, url.scheme }
```
* Check that all the attributes have the correct values


andriisoldatenko:
@StefanHauth please resolve conflicts?

pkg/controllers/dtprometheus/gateway/testdata/statefulset.yaml
```diff
@@ -48,6 +48,13 @@ spec:
             fieldRef:
               apiVersion: v1
               fieldPath: status.podIP
+        - name: DT_API_TOKEN
```
gkrenn: Why did you introduce the env var again? - The token is synced via the file system.

pkg/controllers/dtprometheus/gateway/testdata/configmap.yaml
```diff
@@ -88,7 +85,7 @@ data:
             check_interval: 1s
             limit_percentage: 95
             spike_limit_percentage: 5
-        metric_start_time: null
+        metric_start_time: {}
```
gkrenn, line 88: Please verify that this change does not have side effects.

pkg/controllers/dtprometheus/reconciler.go
```diff
@@ -289,8 +294,7 @@ func newDTPrometheusFromDynaKubeMapper(c client.Client) handler.MapFunc {
 	}
 }
 
-// Create [predicate.Funcs] that only return true when the DynaKube phase changed.
-func newDynaKubePhaseChangedPredicate() predicate.Funcs {
+func newDynaKubePredicate(check func(oldDK, newDK *dynakube.DynaKube) bool) predicate.Funcs {
```
gkrenn: Don't see the need to pass the function here. If we keep it I would make it a variadic function.

gkrenn (changes requested):

pkg/controllers/dtprometheus/gateway/testdata/statefulset.yaml
StefanHauth: Accidentally re-add while resolving conflicts. Will remove again

StefanHauth: Fixed [f35e222](https://github.com/Dynatrace/dynatrace-operator/pull/7357/commits/f35e222225a5a6d21f820924511f9a3a8b11fbe3)

pkg/controllers/dtprometheus/reconciler.go
StefanHauth: Refactored in [d244a0f](https://github.com/Dynatrace/dynatrace-operator/pull/7357/commits/d244a0f1144e34b70bcf3e38a1a20a3194d4178d)

gkrenn (approved):

PR #7358, opened 2026-09-03
Create deployment.properties secret for KubeMon AG

aorcholski:
[ICP-8419](https://dt-rnd.atlassian.net/browse/ICP-8419)

## Description

Create `<dynakube name>-kubemon-deployment-properties` secret that contains dynakube.Spec.ResourceAttributes.

## How can this be tested?

1) unitests
2) deploy dk with KubeMon and Spec.ResourceAttributes and check if the secret exists.

[ICP-8419]: https://dt-rnd.atlassian.net/browse/ICP-8419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -288,6 +291,25 @@ func buildVolumes(dk *dynakube.DynaKube) []corev1.Volume {
...
+					Items: []corev1.KeyToPath{
+						{
+							Key:  agconsts.DeploymentPropertiesFileName,
+							Path: agconsts.DeploymentPropertiesFileName,
+						},
+					},
+				},
+			},
+		})
+	}
```
0sewa0: what if we made this volume `Optional: true`?

so we don't have to put `if`s and just always define it, and if something was created by the sub-reconciler it will "just work"
- (these `if`s will change the moment we add proxy support due to no-proxy)

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
```diff
@@ -457,6 +488,11 @@ func (r *Reconciler) buildDesiredStatefulSet(ctx context.Context, dk *dynakube.D
 		return nil, err
 	}
 
+	deploymentPropertiesHash, err := r.getDeploymentPropertiesHash(dk)
```
0sewa0: somewhat related to previous comment https://github.com/Dynatrace/dynatrace-operator/pull/7358/changes#r3957186046

we are a bit inconsistent here:
- we save the [tlsSecretHash](https://github.com/Dynatrace/dynatrace-operator/blob/dc1845af4073ea7f04281c7d2443cf713ca2242f/pkg/api/latest/dynakube/kubemon/status.go#L23) in the status
- but nothing else is saved in the status.

IMO, we should have all of these in the status or none.

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
0sewa0: if we are don't want to go down the `Optional: true` path, we should abstract the `len(dk.GetResourceAttributes()) > 0 ` to be `dk.KubeMon().NeedsDeploymentProperties()`

note: `dk.KubeMon().NeedsDeploymentProperties()` is probably a bit ugly, as the proxy is set on the `dk` level, not inside the kubemon section, and I am not the biggest fan of [propagating fields down](https://github.com/Dynatrace/dynatrace-operator/blob/dc1845af4073ea7f04281c7d2443cf713ca2242f/pkg/api/latest/dynakube/operands_props.go#L65-L66), but we already do it so 🤷

jdanieck: Its a matter of preference, but I personally like this explicit if approach - you can easily find out when the volume is needed. Don't see any problem with extending the check later to "or proxy needed".

aorcholski: NeedsDeploymentProperties added

pkg/controllers/dynakube/kubemon/statefulset/reconciler.go
aorcholski: I will remove status.kubernetesMonitoring.tlsSecretHash in following PR.
KSPM demonset Reconciler can Get it and calculate hash the same way KubeMon statefulset Reconciler does.

pkg/api/latest/dynakube/operands_props.go
```diff
@@ -64,6 +64,7 @@ func (dk *DynaKube) KubernetesMonitoring() *kubemon.KubeMon {
 	}
 	km.SetName(dk.Name)
 	km.SetAPIURLHost(dk.APIURLHost())
+	km.SetNeedsDeploymentProperties(len(dk.GetResourceAttributes()) > 0)
```
0sewa0, line 67: we can revisit this when implementing the proxy support (https://dt-rnd.atlassian.net/browse/ICP-8504), and undo or lean into it, for now its fine as is, should be a simple change anyway.

any objections @jdanieck ?

0sewa0 (approved):

pkg/api/latest/dynakube/operands_props.go
jdanieck: no objections, go ahead

jdanieck (approved):

PR #7367, opened 2026-09-04
Reconcile DTPrometheus when relevant DynaKube fields changed

avorima:
## Description

The DynaKube resource attributes are part of the gateway config, so the reconciler needs to react when they change.
This is meant to cover the case when the DynaKube does not deploy any workload. If the ActiveGate is deployed, the change in resource attributes will cause a phase change since there will be a rolling restart.

ICP-9419

## How can this be tested?

`make go/test`

### DIFF
--- a/pkg/controllers/dtprometheus/reconciler.go
+++ b/pkg/controllers/dtprometheus/reconciler.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/status"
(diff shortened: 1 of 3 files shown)


jokuniew (dismissed):

andriisoldatenko (dismissed):

pkg/controllers/dtprometheus/reconciler.go
```diff
@@ -289,8 +290,12 @@ func newDTPrometheusFromDynaKubeMapper(c client.Client) handler.MapFunc {
 	}
 }
 
-// Create [predicate.Funcs] that only return true when the DynaKube phase changed.
-func newDynaKubePhaseChangedPredicate() predicate.Funcs {
+// Create [predicate.Funcs] that return true when DynaKube changed in a way that's relevant for a DTPrometheus.
+//   - status.phase changed
+//   - spec.token changed
+//   - spec.resourceAttributes changed
+//   - object deleted
```
0sewa0: maybe dumb question: (I don't know how much of the DK's top fields are reused)

what about:
- `trustedCAs` 
- `proxy` (/ no-proxy FF)
- `dynatraceAPIRequestThreshold`
- `skipCertCheck`
- `publicRegistryOverride`
- `networkZone`

avorima: * trustedCAs: will add
* proxy: will add
* dynatraceAPIRequestThreshold: will add
* skipCertCheck: not used
* publicRegistryOverride: not used
* networkZone: not used

StefanHauth (dismissed):

gkrenn (approved):

PR #7371, opened 2026-09-07
Use fleet management for EdgeConnect image resolution

waodim:
## Description

[ICP-8200](https://dt-rnd.atlassian.net/browse/ICP-8200)

EdgeConnect previously resolved its container image by querying the public OCI registry directly. This PR switches it to use the fleet management endpoint `/api/v2/fleetManagement/components/containerImages`

The behavior depends on configuration:

| `spec.imageRef.repository` | `publicRegistryOverride` | Result |
|---|---|---|
| not set | not set | image from default public registry via fleet management |
| not set | set | image from the specified registry via fleet management |
| set | not set | auto-update disabled, custom image used as-is |

The `fleet-management:container-images:read` OAuth scope is included when building the EdgeConnect client.

Note: The install and install_provisioner e2e tests are currently expected to fail. They go the default image resolution path via the fleet management endpoint (`/platform/classic/environment-api/v2/fleetManagement/components/containerImages`), which requires the OAuth scope `fleet-management:container-images:read`. This scope is not yet available for OAuth clients on the test tenant so it is blocked.

## How can this be tested?

1. Unit tests cover all three scenarios from the table above

2. Deploy an EdgeConnect without `spec.imageRef.repository` and verify the image in the deployment is resolved via fleet management (check `status.version.imageID` is populated with a registry digest). (Unfortunately this does not work as of now as you know)

3. Set `spec.publicRegistryOverride` to a private registry and confirm the image is pulled from that registry instead.

4. Set `spec.imageRef.repository` to a custom image and confirm it is used as-is without contacting fleet management.

[ICP-8200]: https://dt-rnd.atlassian.net/browse/ICP-8200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controllers/edgeconnect/controller.go
```diff
@@ -367,6 +356,53 @@ func (controller *Controller) updateVersionInfo(ctx context.Context, ec *edgecon
...
+	}
+}
+
+// environmentAPIURL builds the base URL under which the classic environment API (/api/v2/...) is
+// reachable on an EdgeConnect apiServer, which always points at a Dynatrace platform host.
+func environmentAPIURL(ec *edgeconnect.EdgeConnect) string {
+	return "https://" + ec.Spec.APIServer + environmentAPIPathPrefix
+}
+
+func newImageClient() imageClientBuilderType {
```
mihaitanasedt, line 430: This is called only in the `imageClientProvider`, do we need both? Also it is very similar to `newEdgeConnectClient`, how about extracting the repetitive logic ina helper  (smth like `buildOAuthClients`)

pkg/util/oci/registry/client.go
mihaitanasedt, line 1: This is not a client anymore, wdyt to rename to `transport` for example?

waodim:
Still WIP since the install and install_provisioner e2e tests are currently expected to fail. They go the default image resolution path via the fleet management endpoint (`/platform/classic/environment-api/v2/fleetManagement/components/containerImages`), which requires the OAuth scope `fleet-management:container-images:read`. This scope is not yet available for OAuth clients on the test tenant so it is blocked.

pkg/controllers/edgeconnect/controller.go
```diff
@@ -907,6 +943,7 @@ func buildOAuthScopes(k8sAutomationEnabled bool) []string {
 		"app-engine:edge-connects:write",
 		"app-engine:edge-connects:delete",
 		"oauth2:clients:manage",
+		"fleet-management:container-images:read",
```
mihaitanasedt: Any existing OAuth client will fail with the new additional scope. Customers will have to rotate the token, is this expected? Should we add the scope also when image is hardcoded or only if public registry?

pkg/controllers/edgeconnect/version/updater.go
```diff
@@ -10,51 +10,69 @@ import (
...
 	}
 
-	return isRequestOutdated && u.IsAutoUpdateEnabled()
+	return u.timeProvider.IsOutdated(version.LastProbeTimestamp, minRequestThreshold) && u.IsAutoUpdateEnabled()
+}
+
+// didPublicRegistryChange reports whether the image in the status was pulled from a registry other
+// than the one currently requested. Without an override the registry is chosen by fleet management,
+// so there is nothing to compare the status against.
+func (u updater) didPublicRegistryChange(imageID string) bool {
```
mihaitanasedt: `did` implies the method compares two historical snapshots, what about `hasRegistryOverrideChanged` or `hasPublicRegistryChanged`?

mihaitanasedt (commented):
Temporary review, lets polish so when the bug with the scopes is fixed we could just test and merge 👍

pkg/controllers/edgeconnect/controller.go
andriisoldatenko: agree with @mihaitanasedt we should not add it always, cc @avorima 

IMO we should do:
```
func buildOAuthScopes(k8sAutomationEnabled bool, isFleetManagmentNeeded bool) []string {
	oAuthScopes := []string{
		"app-engine:edge-connects:read",
		"app-engine:edge-connects:write",
		"app-engine:edge-connects:delete",
		"oauth2:clients:manage",
	}
	if k8sAutomationEnabled {
		oAuthScopes = append(oAuthScopes, "settings:objects:read", "settings:objects:write")
	}

	isFleetManagmentNeeded {
		oAuthScopes = append(oAuthScopes, "\"fleet-management:container-images:read\",")
	}

	return oAuthScopes
}


```


```
func newEdgeConnectClient() func(context.Context, *edgeconnect.EdgeConnect, oauthCredentialsType, []byte) (edgeconnectClient.Client, error) {
	return func(ctx context.Context, ec *edgeconnect.EdgeConnect, oauthCredentials oauthCredentialsType, customCA []byte) (edgeconnectClient.Client, error) {
		oAuthClients, err := dynatrace.NewOAuthClient(
			clientcredentials.Config{
				ClientID:     oauthCredentials.clientID,
				ClientSecret: oauthCredentials.clientSecret,
				TokenURL:     ec.Spec.OAuth.Endpoint,
				Scopes:       buildOAuthScopes(ec.IsK8SAutomationEnabled(), ec.IsCustomImage()),
			},
			dynatrace.WithBaseURL("https://"+ec.Spec.APIServer),
```

so if user using custom image and don't want to auto-update why he must re-generate client?

waodim: good points I agree actually. Thanks, will adapt accordingly.

andriisoldatenko:
During today sync we agreed to fallback to image registry client if fleetmanagment API return component not found.

waodim:
switching back to draft trying to debug tests

pkg/controllers/edgeconnect/version/reconciler.go
```diff
@@ -19,28 +20,39 @@ type versionStatusUpdater interface {
...
+		timeProvider:           timeProvider,
+		imageClientProvider:    imageClientProvider,
+		registryClientProvider: registryClientProvider,
 	}
 }
 
 func (reconciler *Reconciler) Reconcile(ctx context.Context) error {
 	ctx, log := logd.NewFromContext(ctx, "version")
 
 	updaters := []versionStatusUpdater{
```
avorima: this list seems like a leftover. please remove it as a fly-by

pkg/controllers/edgeconnect/controller.go
```diff
@@ -973,11 +973,20 @@ func buildOAuthScopes(k8sAutomationEnabled bool) []string {
...
 	}
 
 	return oAuthScopes
 }
+
+// TODO: the scope cannot be granted on a tenant yet, so this list has never been verified against
+// a live fleet management endpoint. Until then every EdgeConnect resolves its image through the
+// OCI registry fallback. Once the scope is grantable, check that an EdgeConnect with only this
+// scope resolves its image without the fallback, and extend the list if the endpoint asks for more.
+func buildImageOAuthScopes() []string {
```
mihaitanasedt, line 989: Lets rename the edge connect oauth scopes func to have same format: `buildEdgeConnectOAuthScopes`.

pkg/controllers/edgeconnect/version/updater.go
```diff
@@ -31,42 +36,56 @@ func newUpdater(
...
 
-// didPublicRegistryChange reports whether the image in the status was pulled from a registry other
+// hasPublicRegistryChanged reports whether the image in the status was pulled from a registry other
 // than the one currently requested. Without an override the registry is chosen by fleet management,
 // so there is nothing to compare the status against.
-func (u updater) didPublicRegistryChange(imageID string) bool {
+func (u updater) hasPublicRegistryChanged(imageID string) bool {
 	registryOverride := u.edgeConnect.Spec.PublicRegistryOverride
 	if registryOverride == "" {
 		return false
```
mihaitanasedt, line 97: If we have it set and then remove it, it will not trigger a new version update, or?

pkg/controllers/edgeconnect/version/updater.go
```diff
@@ -77,48 +96,158 @@ func (u updater) didPublicRegistryChange(imageID string) bool {
...
+		"newImageID", target.ImageID,
+		"version", target.Version)
+}
+
+func setImageIDToCustomImage(ctx context.Context, target *status.VersionStatus, imageURI string) {
+	log := logd.FromContext(ctx)
+	oldImageID := target.ImageID
+
+	target.ImageID = imageURI
+	target.Version = string(status.CustomImageVersionSource)
```
mihaitanasedt, line 243: Shoudln't it be source instead of version? For custom images we do not have versions

waodim: Source is set too in the `defer` of `Update`, I basically mirrored DKs `setImageIDToCustomImage` whcih basically does the identical thing. I'd keep as is or are you insistent on changing this?

pkg/controllers/edgeconnect/controller.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/7371/commits/5cee461e8c330d5a429f34f32fd7613159220840

pkg/controllers/edgeconnect/version/updater.go
waodim: Yes, you're right removing it doesn't trigger an update. There's nothing to compare against once the override is gone, since without it the registry is what fleet management give (it is essentially only a param for the endpoint). You want me to change it? Or what is your proposal here?

pkg/controllers/edgeconnect/version/updater.go
mihaitanasedt: I see, lets leave it as is, to be consistent

pkg/controllers/edgeconnect/controller.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/7371/changes/5cee461e8c330d5a429f34f32fd7613159220840

pkg/util/oci/registry/client.go
waodim: I think this is outdated now that we reintroduced the logic

pkg/controllers/edgeconnect/version/updater.go
waodim: https://github.com/Dynatrace/dynatrace-operator/pull/7371/commits/a961e11941b9a3ba72db945a66fd5811f78960d6

went with your proposal detecting changes/removals of `publicRegistryOverride` via a new version source

mihaitanasedt (dismissed):
lgtm 

Tested additionally with a mock tenant server ✅

mihaitanasedt (approved):

PR #7374, opened 2026-09-07
e2e tests for kubemon (split, full, +kspm)

jokuniew:
https://dt-rnd.atlassian.net/browse/ICP-7299
## Description

the ticket requires to cover the following criteria:
- **Split-AG mode (new)** - done in `test/e2e/features/kubemon/kubemon.go`
- **All-in AG mode (legacy)** - enabled kubemon experimental flag and extended checks in `test/e2e/features/activegate/activegate.go` that kubemon indeed is not deployed if not used 
- **Validation webhook** - no new code. I think we are covered in `pkg/api/validation/dynakube/kubemon_test.go`
- **Conditions** - no new code. I think we are covered in `pkg/controllers/dynakube/kubemon/reconciler_test.go`
- **Restart triggers** - done in `test/e2e/features/kubemon/kubemon.go`
- **Cleanup** -  done in `test/e2e/features/kubemon/kubemon.go` at the end
- **RBAC** - ~done in `test/e2e/features/kubemon/kubemon.go` - it's first time in e2e where we check rbac so precisely. ⚠️~ will be covered in helm unit tests
- **Integrations (with KSPM)** `test/e2e/features/kspm/kspm.go` - kspm was done already by @mihaitanasedt , i've expanded this tests with settings creation check and rbac

## How can this be tested?
``` 
make go/test (Webhook/Conditions)

make test/e2e/activegate
make test/e2e/kubemon
make test/e2e/kubemon/olm


# Test specific scenarios
make test/e2e/kubemon/split-ag
make test/e2e/kubemon/restart-triggers
```

### DIFF
--- a/hack/make/tests/e2e.mk
+++ b/hack/make/tests/e2e.mk
@@ -418,11 +418,11 @@ test/e2e/kubemon:
 
 ## Runs Kubernetes Monitoring split-AG provisioning, RBAC and cleanup e2e test
 test/e2e/kubemon/split-ag:
-	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/nocsi -run "kubemon_SplitAG" $(SKIPCLEANUP)
+	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/nocsi -run "kubemon_split_ag" $(SKIPCLEANUP)
 
 ## Runs Kubernetes Monitoring restart-triggers e2e test
 test/e2e/kubemon/restart-triggers:
-	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "kubemon_RestartTriggers" $(SKIPCLEANUP)
+	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "kubemon_kubemon_restart_triggers" $(SKIPCLEANUP)
 
 test/e2e/token/migration:
 	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "token_migration" $(SKIPCLEANUP)
--- a/hack/make/tests/e2e.mk
+++ b/hack/make/tests/e2e.mk
@@ -414,15 +414,15 @@ test/e2e/kspm/optionalscopes:
 
 ## Runs all Kubernetes Monitoring e2e tests
 test/e2e/kubemon:
-	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/nocsi -run "Kubemon" $(SKIPCLEANUP)
+	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/nocsi -run "kubemon" $(SKIPCLEANUP)
 
 ## Runs Kubernetes Monitoring split-AG provisioning, RBAC and cleanup e2e test
 test/e2e/kubemon/split-ag:
-	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/nocsi -run "Kubemon_SplitAG" $(SKIPCLEANUP)
+	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/nocsi -run "kubemon_SplitAG" $(SKIPCLEANUP)
 
 ## Runs Kubernetes Monitoring restart-triggers e2e test
 test/e2e/kubemon/restart-triggers:
-	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "Kubemon_RestartTriggers" $(SKIPCLEANUP)
+	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "kubemon_RestartTriggers" $(SKIPCLEANUP)
 
 test/e2e/token/migration:
 	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "token_migration" $(SKIPCLEANUP)
--- a/hack/make/tests/e2e.mk
+++ b/hack/make/tests/e2e.mk
@@ -412,5 +412,17 @@ test/e2e/kspm/kubemon:
 test/e2e/kspm/optionalscopes:
 	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "kspm_with_optional_scopes" $(SKIPCLEANUP)
 
+## Runs all Kubernetes Monitoring e2e tests
+test/e2e/kubemon:
+	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/nocsi -run "Kubemon" $(SKIPCLEANUP)
+
+## Runs Kubernetes Monitoring split-AG provisioning, RBAC and cleanup e2e test
+test/e2e/kubemon/split-ag:
+	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/nocsi -run "Kubemon_SplitAG" $(SKIPCLEANUP)
+
+## Runs Kubernetes Monitoring restart-triggers e2e test
+test/e2e/kubemon/restart-triggers:
+	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "Kubemon_RestartTriggers" $(SKIPCLEANUP)
+
 test/e2e/token/migration:
 	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "token_migration" $(SKIPCLEANUP)
--- a/test/e2e/features/activegate/activegate.go
+++ b/test/e2e/features/activegate/activegate.go
@@ -85,7 +85,7 @@ func Feature(t *testing.T, proxySpec *value.Source) features.Feature {
 	// make sure that if separate kubemon activegate is not used - it does not create separate statefulset and secret
 	builder.Assess("kubemon statefulset does not exist", func(ctx context.Context, t *testing.T, envConfig *envconf.Config) context.Context {
 		_, err := k8sstatefulset.Get(ctx, envConfig.Client().Resources(), testDynakube.Name+"-kubemon", testDynakube.Namespace)
-		require.True(t, k8serrors.IsNotFound(err))
+require.Truef(t, k8serrors.IsNotFound(err), "expected NotFound, got: %v", err)
 
 		return ctx
 	})
--- a/test/e2e/features/activegate/activegate.go
+++ b/test/e2e/features/activegate/activegate.go
@@ -84,7 +84,7 @@ func Feature(t *testing.T, proxySpec *value.Source) features.Feature {
 	// only activegate capabilities are used in this test
 	// make sure that if separate kubemon activegate is not used - it does not create separate statefulset and secret
 	builder.Assess("kubemon statefulset does not exist", func(ctx context.Context, t *testing.T, envConfig *envconf.Config) context.Context {
-		_, err := k8sstatefulset.Get(ctx, envConfig.Client().Resources(), testDynakube.Name+"-kubemon", testDynakube.Namespace)
+		_, err := k8sstatefulset.Get(ctx, envConfig.Client().Resources(), testDynakube.KubernetesMonitoring().GetStatefulSetName(), testDynakube.Namespace)
 		require.Truef(t, k8serrors.IsNotFound(err), "expected NotFound, got: %v", err)
 
 		return ctx
@@ -92,7 +92,7 @@ func Feature(t *testing.T, proxySpec *value.Source) features.Feature {
 
 	builder.Assess("kubemon authtoken secret does not exist", func(ctx context.Context, t *testing.T, envConfig *envconf.Config) context.Context {
 		var secret corev1.Secret
-		err := envConfig.Client().Resources().Get(ctx, testDynakube.Name+"-kubemon-authtoken-secret", testDynakube.Namespace, &secret)
+		err := envConfig.Client().Resources().Get(ctx, testDynakube.KubernetesMonitoring().GetAuthTokenSecretName(), testDynakube.Namespace, &secret)
 		require.Truef(t, k8serrors.IsNotFound(err), "expected NotFound, got: %v", err)
 
 		return ctx
--- a/test/e2e/features/activegate/activegate.go
+++ b/test/e2e/features/activegate/activegate.go
@@ -15,8 +15,10 @@ import (
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/latest/dynakube"
 	"github.com/Dynatrace/dynatrace-operator/pkg/api/shared/value"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/dynakube/activegate/consts"
+	"github.com/Dynatrace/dynatrace-operator/test/e2e/helpers"
 	"github.com/Dynatrace/dynatrace-operator/test/e2e/helpers/components/activegate"
 	dynakubeComponents "github.com/Dynatrace/dynatrace-operator/test/e2e/helpers/components/dynakube"
+	componentOperator "github.com/Dynatrace/dynatrace-operator/test/e2e/helpers/components/operator"
 	"github.com/Dynatrace/dynatrace-operator/test/e2e/helpers/kubernetes/objects/k8spod"
 	"github.com/Dynatrace/dynatrace-operator/test/e2e/helpers/kubernetes/objects/k8sstatefulset"
 	"github.com/Dynatrace/dynatrace-operator/test/e2e/helpers/proxy"
@@ -25,6 +27,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 	"sigs.k8s.io/e2e-framework/pkg/features"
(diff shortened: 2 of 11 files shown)


test/e2e/features/kubemon/kubemon.go
```diff
@@ -0,0 +1,165 @@
...
+		),
+		componentDynakube.WithKubernetesMonitoringRegistration(),
+	)
+
+	componentDynakube.Install(builder, &secretConfig, testDynakube)
+
+	builder.Assess("generic activegate statefulset is ready",
+		k8sstatefulset.IsReady(agHelper.GetActiveGateStateFulSetName(&testDynakube, "activegate"), testDynakube.Namespace))
+
+	builder.Assess("kubemon statefulset is ready",
```
avorima, line 49: do not wait. the same guarantee as the AG applies. if the DynaKube is ready, kubemon must be ready as well

andriisoldatenko (dismissed):

test/e2e/features/kubemon/kubemon.go
```diff
@@ -0,0 +1,165 @@
...
+	builder.Assess("kubemon authtoken secret exists",
+		k8ssecret.Exists(testDynakube.KubernetesMonitoring().GetAuthTokenSecretName(), testDynakube.Namespace))
+
+	builder.Assess("kubemon tenant secret exists",
+		k8ssecret.Exists(testDynakube.KubernetesMonitoring().GetTenantSecretName(), testDynakube.Namespace))
+
+	builder.Assess("KubernetesMonitoringAvailable condition is True",
+		componentDynakube.WaitForCondition(testDynakube, kubemon.KubeMonAvailableConditionType, metav1.ConditionTrue))
+
+	builder.Assess("kubemon pod uses dynatrace-activegate service account", func(ctx context.Context, t *testing.T, envConfig *envconf.Config) context.Context {
```
avorima: this should be covered by unit/integration tests

test/e2e/features/kubemon/kubemon.go
```diff
@@ -0,0 +1,165 @@
...
+
+		var cr rbacv1.ClusterRole
+		require.NoError(t, envConfig.Client().Resources().Get(ctx, "dynatrace-activegate", "", &cr))
+
+		var crb rbacv1.ClusterRoleBinding
+		require.NoError(t, envConfig.Client().Resources().Get(ctx, "dynatrace-activegate", "", &crb))
+		assert.Equal(t, "dynatrace-activegate", crb.RoleRef.Name)
+
+		return ctx
+	})
```
avorima: these RBAC existence checks should be covered by our helm unit tests. RBAC checks in e2e tests should be functional, like we do for the deployer

test/e2e/features/kubemon/kubemon.go
```diff
@@ -0,0 +1,165 @@
...
+	builder.Assess("kubemon statefulset is ready",
+		k8sstatefulset.WaitFor(testDynakube.KubernetesMonitoring().GetStatefulSetName(), testDynakube.Namespace))
+
+	builder.Assess("KubernetesMonitoringAvailable condition is True",
+		componentDynakube.WaitForCondition(testDynakube, kubemon.KubeMonAvailableConditionType, metav1.ConditionTrue))
+
+	builder.Assess("rotate authtoken secret",
+		k8ssecret.Delete(k8ssecret.New(testDynakube.KubernetesMonitoring().GetAuthTokenSecretName(), testDynakube.Namespace, nil)))
+
+	builder.Assess("kubemon statefulset re-rolls after secret rotation",
```
avorima: I don't believe the AG can do a complete restart in under a minute. Looking at the DynaKube condition may be a better signal

```
--- PASS: TestNoCSI_Kubemon_RestartTriggers/kubemon-restart-triggers/kubemon_statefulset_re-rolls_after_secret_rotation (5.74s)
```

test/e2e/features/kubemon/kubemon.go
```diff
@@ -0,0 +1,165 @@
...
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func FeatureSplitAG(t *testing.T) features.Feature {
+	builder := features.New("kubemon-split-ag-mode")
+
+	secretConfig := tenant.GetSingleTenantSecret(t)
+
+	builder.Setup(helpers.ToFeatureFunc(componentOperator.InstallLocal(false, componentOperator.EnableKubemonOperand()...), true))
```
avorima: Each test needs to clean up after itself, so this setup needs to be undone. We could think about a separate kubemon scenario

test/e2e/features/activegate/activegate.go
```diff
@@ -72,12 +75,31 @@ func Feature(t *testing.T, proxySpec *value.Source) features.Feature {
 	proxy.CutOffDynatraceNamespace(builder, proxySpec)
 	proxy.IsDynatraceNamespaceCutOff(builder, testDynakube)
 
+	builder.Setup(helpers.ToFeatureFunc(componentOperator.InstallLocal(false, componentOperator.EnableKubemonOperand()...), true))
```
avorima: in the case of KSPM this is a separate test

test/e2e/features/activegate/activegate.go
```diff
@@ -72,12 +75,31 @@ func Feature(t *testing.T, proxySpec *value.Source) features.Feature {
...
 	dynakubeComponents.Install(builder, &secretConfig, testDynakube)
 	assessActiveGate(builder, &testDynakube)
 
 	assessReadOnlyActiveGate(builder, &testDynakube)
 
+	// only activegate capabilities are used in this test
+	// make sure that if separate kubemon activegate is not used - it does not create separate statefulset and secret
+	builder.Assess("kubemon statefulset does not exist", func(ctx context.Context, t *testing.T, envConfig *envconf.Config) context.Context {
+		_, err := k8sstatefulset.Get(ctx, envConfig.Client().Resources(), testDynakube.Name+"-kubemon", testDynakube.Namespace)
+		require.True(t, k8serrors.IsNotFound(err))
```
svfoxat:
```suggestion
require.Truef(t, k8serrors.IsNotFound(err), "expected NotFound, got: %v", err)
```

This would result in a clearer assertion message rather than `True was not true` in case the `err` is not a `NotFoundError`

test/e2e/features/activegate/activegate.go
```diff
@@ -72,12 +75,31 @@ func Feature(t *testing.T, proxySpec *value.Source) features.Feature {
...
+
+		return ctx
+	})
+
+	builder.Assess("kubemon authtoken secret does not exist", func(ctx context.Context, t *testing.T, envConfig *envconf.Config) context.Context {
+		var secret corev1.Secret
+		err := envConfig.Client().Resources().Get(ctx, testDynakube.Name+"-kubemon-authtoken-secret", testDynakube.Namespace, &secret)
+		require.True(t, k8serrors.IsNotFound(err))
```
svfoxat: same as above

test/e2e/features/kubemon/kubemon.go
jokuniew: :+1: changed from `WaitFor` to `IsReady` https://github.com/Dynatrace/dynatrace-operator/pull/7374/changes/266755e246ccf17e41f396d96ae7a1f25fc5298c
one `WaitFor` is covered in different coversation here https://github.com/Dynatrace/dynatrace-operator/pull/7374/changes/f98f8fb3aacc87b3af83098e5f26b25005491806

hack/make/tests/e2e.mk
```diff
@@ -412,5 +412,17 @@ test/e2e/kspm/kubemon:
 test/e2e/kspm/optionalscopes:
 	$(GOTESTCMD) -timeout 20m ./test/e2e/scenarios/nocsi -run "kspm_with_optional_scopes" $(SKIPCLEANUP)
 
+## Runs all Kubernetes Monitoring e2e tests
+test/e2e/kubemon:
+	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/nocsi -run "Kubemon" $(SKIPCLEANUP)
```
mihaitanasedt: All other components use lowercase
```suggestion
	$(GOTESTCMD) -timeout 30m ./test/e2e/scenarios/nocsi -run "kubemon" $(SKIPCLEANUP)
```

test/e2e/features/kubemon/kubemon.go
jokuniew: it is already covered there, so this is redundant - removed https://github.com/Dynatrace/dynatrace-operator/pull/7374/changes/f5f99cfb4bf3bad42e54ef9585405b6546bd7407

Unit test — [pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go:214](https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/controllers/dynakube/kubemon/statefulset/reconciler_test.go#L214)                                                                                                                                                                                                                                                                                                                                                                                                       
Integration test — [pkg/controllers/dynakube/kubemon/statefulset/reconciler_integration_test.go:314    ](https://github.com/Dynatrace/dynatrace-operator/blob/main/pkg/controllers/dynakube/kubemon/statefulset/reconciler_integration_test.go#L314)

test/e2e/features/kubemon/kubemon.go
jokuniew: i agree that those RBAC checks are weird and I wasn't sure if should I place them here, no any other e2e test checks that (updating PR description)
removed those checks here https://github.com/Dynatrace/dynatrace-operator/pull/7374/changes/cc114f340338e018c50a9f37d883c5975497d260

in rbac tests `clusterrole-kubernetes-monitoring-default_test.yaml` (and kspm one) we test
 `dynatrace-kubernetes-monitoring-default` ClusterRole and ClusterRoleBinding
 and
`dynatrace-kubernetes-monitoring-kspm` ClusterRole
 
for `dynatrace-activegate` we check ServiceAccount only without ClusterRoles/Bindings. Will create that

test/e2e/features/kubemon/kubemon.go
jokuniew: "under a minute" - I need explanation why a minute here? For a total test time?
the statefulset's `WaitFor` has 10 minutes timeout https://github.com/Dynatrace/dynatrace-operator/blob/main/test/e2e/helpers/kubernetes/objects/k8sstatefulset/statefulset.go#L88


changed into checking condition as `False` immediately after deletion and then waiting for `True` https://github.com/Dynatrace/dynatrace-operator/pull/7374/changes/f98f8fb3aacc87b3af83098e5f26b25005491806 <- here it is 5 min timeout

test/e2e/features/kubemon/kubemon.go
avorima: the WaitForCondition after the initial DynaKube deployment should go or be replaced by a one-time check

test/e2e/features/kubemon/kubemon.go
jokuniew: by "so this setup needs to be undone" - do you mean adding Teardown like in KSPM scenario?

```
	builder.Teardown(helpers.ToFeatureFunc(componentOperator.InstallLocal(
		false,
		helm.WithArgs("--reuse-values"),
		helm.WithArgs("--set", "experimental.enableKubemonOperand=false"),
	), false))
```
regarding separate kubemon scenario - i already check in split-ag the deletion of kubemon https://github.com/Dynatrace/dynatrace-operator/pull/7374/changes#diff-564db238df861d2550e210871b6c2fbdf4b0a7c30161883ed6f84394c351dc09R73

test/e2e/features/kubemon/kubemon.go
avorima: it's just how long the activegate pod takes to get ready. in some setups it's over 3 minutes, but in our environments 1 minute is about as good as it gets. so if this test step `kubemon statefulset re-rolls after secret rotation` completes way faster than that it tells me it didn't actually wait for the right thing

test/e2e/features/activegate/activegate.go
jokuniew: applied suggestion and changed here as well from True to Truef with message
https://github.com/Dynatrace/dynatrace-operator/pull/7374/changes/7a46548add5c47f9e49b589c3ec598bfbc7c9193

test/e2e/features/kubemon/kubemon.go
avorima: yes, it should do the same as KSPM.

A "scenario" in this case is another TestMain that installs kubemon operand from the start instead of having to enable and disable it for each case. we can go with that approach later, but i think there is a point when always modifying and resetting the operator significantly eats into our runtime

hack/make/tests/e2e.mk
jokuniew: rename it to kubemon
now it catches also the `kspm_with_kubemon` but that is actually a good think imo
https://github.com/Dynatrace/dynatrace-operator/pull/7374/commits/cead6cfae35f26cbb040335b2724e8f8ad3c7f73

test/e2e/features/activegate/activegate.go
jokuniew: there is already separate test for kubemon  https://github.com/Dynatrace/dynatrace-operator/blob/main/test/e2e/features/kspm/kspm.go#L50
or do you mean to have 
```
componentDynakube.WithKSPM(),
componentDynakube.WithActiveGate(), # (all capabilites here)
```
and with experimental FF but without KubemonRegistration to see if doesn't conflict?

test/e2e/features/kubemon/kubemon.go
jokuniew: added teardown to those two test cases https://github.com/Dynatrace/dynatrace-operator/pull/7374/commits/03331652e8e31dd6cfa87cc08752a9c921735c9d

I thought about your perspective of separate scenario
imo if we are going to make this feature GA shortly - no point to do it now
if it is going to stay for a while as is - separate scenario :+1:

test/e2e/features/kubemon/kubemon.go
jokuniew: @avorima  FYI created utests here
https://github.com/Dynatrace/dynatrace-operator/pull/7374/commits/d8af35152f668afedef4b66ce310d8e042b5dc0f

test/e2e/features/kubemon/kubemon.go
jokuniew: introduced CheckCondition to do a oneCheck for them https://github.com/Dynatrace/dynatrace-operator/pull/7374/changes/50dd219a3712a3726e3cc625e44907853e3e47ef

hack/make/tests/e2e.mk
mihaitanasedt: Lets stick to sanke_casing. (e.g., `kubemon_restart_triggers`)

jokuniew: changed to snake_case and added a note in style guide
https://github.com/Dynatrace/dynatrace-operator/pull/7374/commits/e2309c7cb6cd53b1529d91afc60aa5956180a254

test/e2e/features/activegate/activegate.go
avorima: > there is already separate test for kubemon 

yes, KSPM has two cases one with and one without kubemon, but the AG test combines both in one. for grouping tests it would make sense to separate the AG test case into Feature and FeatureWithKubemon, so that `make test/e2e/kubemon` catches these two as well

test/e2e/helpers/components/operator/installation.go
```diff
@@ -320,3 +320,17 @@ func getImageRef(rootDir string, fips140 bool) (string, error) {
 
 	return imageRef, nil
 }
+
+func EnableKubemonOperand() []helm.Option {
```
avorima: `helm.WithArgs` takes varargs, so we can also just return a single option. this makes the usage simpler because you don't have to splat the list

test/e2e/features/activegate/activegate.go
```diff
@@ -70,12 +73,31 @@ func Feature(t *testing.T, proxySpec *value.Source) features.Feature {
...
 	// Register actual test
 	dynakubeComponents.Install(builder, &secretConfig, testDynakube)
 	assessActiveGate(builder, &testDynakube)
 
 	assessReadOnlyActiveGate(builder, &testDynakube)
 
+	// only activegate capabilities are used in this test
+	// make sure that if separate kubemon activegate is not used - it does not create separate statefulset and secret
+	builder.Assess("kubemon statefulset does not exist", func(ctx context.Context, t *testing.T, envConfig *envconf.Config) context.Context {
+		_, err := k8sstatefulset.Get(ctx, envConfig.Client().Resources(), testDynakube.Name+"-kubemon", testDynakube.Namespace)
```
0sewa0: ~I don't know if we want to reuse the logic or just hardcode it as is right now~ you use it [here](https://github.com/Dynatrace/dynatrace-operator/pull/7374/changes#diff-564db238df861d2550e210871b6c2fbdf4b0a7c30161883ed6f84394c351dc09R114) so 😅 

```suggestion
		_, err := k8sstatefulset.Get(ctx, envConfig.Client().Resources(), testDynakube.KubernetesMonitoring().GetStatefulSetName(), testDynakube.Namespace)
```

similar for `testDynakube.Name+"-kubemon-authtoken-secret` -> `testDynakube.KubernetesMonitoring().GetAuthTokenSecretName()`

test/e2e/helpers/components/operator/installation.go
jokuniew: simplified here https://github.com/Dynatrace/dynatrace-operator/pull/7374/commits/67cac1e1d82858fb4f282761ded17887f6a6c0a6

test/e2e/features/activegate/activegate.go
jokuniew: fetching names from methods here https://github.com/Dynatrace/dynatrace-operator/pull/7374/commits/c9e9c120d6e65c11d0b0fc4d20f942b1e65aea42

svfoxat (approved):

test/e2e/features/activegate/activegate.go
```diff
@@ -70,12 +73,31 @@ func Feature(t *testing.T, proxySpec *value.Source) features.Feature {
...
+		return ctx
+	})
+
+	builder.Assess("kubemon authtoken secret does not exist", func(ctx context.Context, t *testing.T, envConfig *envconf.Config) context.Context {
+		var secret corev1.Secret
+		err := envConfig.Client().Resources().Get(ctx, testDynakube.KubernetesMonitoring().GetAuthTokenSecretName(), testDynakube.Namespace, &secret)
+		require.Truef(t, k8serrors.IsNotFound(err), "expected NotFound, got: %v", err)
+
+		return ctx
+	})
```
0sewa0, line 99: I don't see why we need this, but they are fast requests 😅 

(following this logic we should also double check in other places)

test/e2e/features/activegate/activegate.go
avorima: am blind

0sewa0 (approved):
tested it:
```sh
> make test/e2e/kubemon/platform-token

...

ok      github.com/Dynatrace/dynatrace-operator/test/e2e/scenarios/nocsi        1325.592s
```

test/e2e/features/activegate/activegate.go
jokuniew: I've added this "make sure it doesn't exist" to fulfill kod-2 requirements, the All-in AG mode (legacy)

avorima (approved):

andriisoldatenko (approved):

PR #7376, opened 2026-09-07
Address Renovate warnings

Incredible-O:
## Description
Fixes the following warnings:
> `hashedBranchLength` must allow for at least the minimum hash length in addition to `branchPrefix`. Using the minimum hash length instead.

Fixed by changing Renovate's PR branch prefix from `renovate/` to `rnv/`

> No docker auth found - returning

Fixed by excluding `gcr.io/cloud-marketplace-tools/**` => doesn't need to be updated, and requires authentication

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -36,6 +36,7 @@
   automergeType: "pr",
   automerge: true,
   branchNameStrict: true,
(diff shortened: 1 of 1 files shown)


jokuniew (approved):

andriisoldatenko:
🚢

PR #7385, opened 2026-09-08
Remove tlsSecretHash field from status.KubernetesMonitoring section

aorcholski:
[ICP-9512](https://dt-rnd.atlassian.net/browse/ICP-9512)

## How can this be tested?

1) unittests
2) deploy dk with AG, KM and KSPM enabled
```
  activeGate:
    capabilities:
    - routing
  kubernetesMonitoring:
    registration: {}
  kspm:
    mappedHostPaths:
      - /boot
  templates:
      imageRef:
        repository: public.ecr.aws/dynatrace/dynatrace-k8s-node-config-collector
        tag: 1.5.12
```
- delete KM TLS secret - KSPM and KM PODs should be restarted
- check dynakube.status.KubernetesMonitoring section

[ICP-9512]: https://dt-rnd.atlassian.net/browse/ICP-9512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controllers/dynakube/kspm/daemonset/reconciler.go
+++ b/pkg/controllers/dynakube/kspm/daemonset/reconciler.go
@@ -11,13 +11,16 @@ import (
 	dtimage "github.com/Dynatrace/dynatrace-operator/pkg/clients/dynatrace/image"
 	"github.com/Dynatrace/dynatrace-operator/pkg/controllers/registry"
 	"github.com/Dynatrace/dynatrace-operator/pkg/logd"
+	"github.com/Dynatrace/dynatrace-operator/pkg/util/hasher"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8saffinity"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8sconditions"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8slabel"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/fields/k8ssecuritycontext"
 	"github.com/Dynatrace/dynatrace-operator/pkg/util/kubernetes/objects/k8sdaemonset"
+	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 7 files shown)


0sewa0 (approved):

pkg/controllers/dynakube/kspm/daemonset/reconciler.go
```diff
@@ -175,3 +185,43 @@ func buildPodSecurityContext() *corev1.PodSecurityContext {
...
+
+	var tlsSecretName string
+
+	switch {
+	case dk.IsKubemonEnabled():
+		tlsSecretName = dk.KubernetesMonitoring().GetTLSSecretName()
+	case dk.ActiveGate().IsEnabled():
+		tlsSecretName = dk.ActiveGate().GetTLSSecretName()
+		if tlsSecretName == "" {
+			return "", nil
```
waodim, line 200: why do we have the empty check here but not in the case above?

pkg/controllers/dynakube/kspm/daemonset/reconciler.go
```diff
@@ -175,3 +185,43 @@ func buildPodSecurityContext() *corev1.PodSecurityContext {
...
+		}
+	default:
+		return "", nil
+	}
+
+	err := r.kubeClient.Get(ctx, client.ObjectKey{Name: tlsSecretName, Namespace: dk.Namespace}, &secret)
+	if k8serrors.IsNotFound(err) {
+		return "", nil
```
waodim, line 208: Here we just return "" when the secret is not found, but previously we retried on error, is this intentional?

pkg/controllers/dynakube/kspm/daemonset/reconciler.go
aorcholski: because generic AG respects `automatic-tls-certificate: "false"` FF so it's possible that operator doesn't create TLS secret for generic AG

pkg/controllers/dynakube/kspm/daemonset/reconciler.go
aorcholski: It was a sequence of immediately consecutive calls of secret.Create -> secret.Get when gateway/reconciler.go called `setStatusTLSSecretHash` so Get repeatedly returned NotFound error.

jdanieck (approved):

waodim (approved):

PR #7404, opened 2026-09-10
fix: use golang version from Dockerfile for actions/setup-go

jokuniew:
## Description

I've introduced this on release branches, but not on main:
1.10 https://github.com/Dynatrace/dynatrace-operator/pull/7341
1.9 https://github.com/Dynatrace/dynatrace-operator/pull/7342

we used `'stable'` version in setup-go actions. 
This cause misalignment between golang version in ci and code causing errors in linting and generating files like:

```
  Error: ../../../../../opt/hostedtoolcache/go/1.27.0/x64/src/crypto/internal/randutil/randutil.go:11:2: could not import math/rand/v2 (/opt/hostedtoolcache/go/1.27.0/x64/src/math/rand/v2/rand.go:213:17: method must have no type parameters) (typecheck)
```

## How can this be tested?
CI runs

### DIFF
--- a/.github/actions/build-push-image/action.yaml
+++ b/.github/actions/build-push-image/action.yaml
@@ -42,7 +42,7 @@ runs:
       uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
     - name: Get go version
       id: go-version
-      uses: ./.github/actions/get-go-version
+      uses: $/.github/actions/get-go-version
       with:
         dockerfile: ${{ inputs.dockerfile }}
     - name: Set up Golang
--- a/.github/actions/build-push-image/action.yaml
+++ b/.github/actions/build-push-image/action.yaml
@@ -43,7 +43,9 @@ runs:
     - name: Get go version
       id: go-version
       shell: bash
-      run: echo "go_version=$(grep -m1 'FROM.*golang:' ${{ inputs.dockerfile }} | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"
+      env:
+        DOCKERFILE: ${{ inputs.dockerfile }}
+      run: echo "go_version=$(grep -m1 'FROM.*golang:' "$DOCKERFILE" | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"
     - name: Set up Golang
       uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
       with:
--- a/.github/actions/build-push-image/action.yaml
+++ b/.github/actions/build-push-image/action.yaml
@@ -40,10 +40,14 @@ runs:
       uses: docker/setup-qemu-action@1f40c72289eff860ee54a304f1438e3cff362e0a # v4.3.0
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
+    - name: Get go version
+      id: go-version
+      shell: bash
+      run: echo "go_version=$(grep -m1 'FROM.*golang:' ${{ inputs.dockerfile }} | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"
     - name: Set up Golang
       uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
       with:
-        go-version: 'stable' # Latest stable version
+        go-version: ${{ steps.go-version.outputs.go_version }}
         cache: false
     - name: Download third party licenses
       shell: bash
--- a/.github/actions/build-push-image/action.yaml
+++ b/.github/actions/build-push-image/action.yaml
@@ -42,10 +42,9 @@ runs:
       uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
     - name: Get go version
       id: go-version
-      shell: bash
-      env:
-        DOCKERFILE: ${{ inputs.dockerfile }}
-      run: echo "go_version=$(grep -m1 'FROM.*golang:' "$DOCKERFILE" | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"
+      uses: ./.github/actions/get-go-version
+      with:
+        dockerfile: ${{ inputs.dockerfile }}
     - name: Set up Golang
       uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
       with:
--- a/.github/actions/run-e2e/action.yaml
+++ b/.github/actions/run-e2e/action.yaml
@@ -128,7 +128,7 @@ runs:
       uses: azure/setup-kubectl@829323503d1be3d00ca8346e5391ca0b07a9ab0d # v5.1.0
     - name: Get go version
       id: go-version
-      uses: ./.github/actions/get-go-version
+      uses: $/.github/actions/get-go-version
       with:
         dockerfile: ./target/Dockerfile
     - name: Set up go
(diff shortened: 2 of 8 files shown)


andriisoldatenko:
wdyt about something like reusable action?
```diff
diff --git .github/actions/build-push-image/action.yaml .github/actions/build-push-image/action.yaml
index 00bc483c3..b984302ff 100644
--- .github/actions/build-push-image/action.yaml
+++ .github/actions/build-push-image/action.yaml
@@ -42,10 +42,9 @@ runs:
       uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
     - name: Get go version
       id: go-version
-      shell: bash
-      env:
-        DOCKERFILE: ${{ inputs.dockerfile }}
-      run: echo "go_version=$(grep -m1 'FROM.*golang:' "$DOCKERFILE" | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"
+      uses: ./.github/actions/get-go-version
+      with:
+        dockerfile: ${{ inputs.dockerfile }}
     - name: Set up Golang
       uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
       with:
diff --git .github/actions/run-e2e/action.yaml .github/actions/run-e2e/action.yaml
index 73aee9f4f..87f4f4e86 100644
--- .github/actions/run-e2e/action.yaml
+++ .github/actions/run-e2e/action.yaml
@@ -128,8 +128,9 @@ runs:
       uses: azure/setup-kubectl@829323503d1be3d00ca8346e5391ca0b07a9ab0d # v5.1.0
     - name: Get go version
       id: go-version
-      shell: bash
-      run: echo "go_version=$(grep -m1 'FROM.*golang:' ./target/Dockerfile | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"
+      uses: ./.github/actions/get-go-version
+      with:
+        dockerfile: ./target/Dockerfile
     - name: Set up go
       uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
       with:
diff --git .github/workflows/ci.yaml .github/workflows/ci.yaml
index 6ffed2bae..a4f0c4ee8 100644
--- .github/workflows/ci.yaml
+++ .github/workflows/ci.yaml
@@ -130,8 +130,7 @@ jobs:
           CHANGED_PREREQUISITES_MK_FILE: ${{ steps.prerequisites.outputs.all_changed_files }}
       - name: Get Go version from Dockerfile
         id: go-version
-        run: |
-          echo "go_version=$(grep -m1 'FROM.*golang:' Dockerfile | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"
+        uses: ./.github/actions/get-go-version
 
   helm-test:
     needs: [detect-changes]
diff --git .github/workflows/nightly.yaml .github/workflows/nightly.yaml
index 764756f4d..938ec2df2 100644
--- .github/workflows/nightly.yaml
+++ .github/workflows/nightly.yaml
@@ -242,7 +242,7 @@ jobs:
           persist-credentials: false
       - name: Get go version
         id: go-version
-        run: echo "go_version=$(grep -m1 'FROM.*golang:' Dockerfile | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"
+        uses: ./.github/actions/get-go-version
       - name: Setup Golang
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
diff --git .github/workflows/release.yaml .github/workflows/release.yaml
index d75d8c9b3..a45dc8894 100644
--- .github/workflows/release.yaml
+++ .github/workflows/release.yaml
@@ -386,7 +386,7 @@ jobs:
           persist-credentials: false
       - name: Get go version
         id: go-version
-        run: echo "go_version=$(grep -m1 'FROM.*golang:' Dockerfile | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"
+        uses: ./.github/actions/get-go-version
       - name: Setup Golang
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
diff --git .github/workflows/sync-openapi-schema.yaml .github/workflows/sync-openapi-schema.yaml
index 9d6714d40..de08591a9 100644
--- .github/workflows/sync-openapi-schema.yaml
+++ .github/workflows/sync-openapi-schema.yaml
@@ -62,7 +62,7 @@ jobs:
 
       - name: Get go version
         id: go-version
-        run: echo "go_version=$(grep -m1 'FROM.*golang:' Dockerfile | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"
+        uses: ./.github/actions/get-go-version
 
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:

```


and 
```
cat .github/actions/get-go-version/action.yaml

name: Get Go version
description: Extracts Go version from a Dockerfile
inputs:
  dockerfile:
    description: Path to the Dockerfile
    default: Dockerfile
    required: false
outputs:
  go_version:
    description: Go version extracted from the Dockerfile
    value: ${{ steps.go-version.outputs.go_version }}
runs:
  using: "composite"
  steps:
    - name: Get go version
      id: go-version
      shell: bash
      env:
        DOCKERFILE: ${{ inputs.dockerfile }}
      run: echo "go_version=$(grep -m1 'FROM.*golang:' "$DOCKERFILE" | sed 's/.*golang:\([0-9.]*\).*/\1/')" >> "$GITHUB_OUTPUT"

```

0sewa0 (approved):

.github/actions/build-push-image/action.yaml
```diff
@@ -40,10 +40,15 @@ runs:
       uses: docker/setup-qemu-action@1f40c72289eff860ee54a304f1438e3cff362e0a # v4.3.0
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
+    - name: Get go version
+      id: go-version
+      uses: $/.github/actions/get-go-version
```
avorima, line 45: i never saw this `$/.` notation what does the `$` do?

jokuniew: https://github.com/Dynatrace/dynatrace-operator/security/code-scanning/489
https://github.blog/changelog/2026-07-30-reference-same-repository-actions-with-self-repository-syntax/

_You can now reference an action or reusable workflow that lives in the same repository using the new self-repository syntax. A uses: value that starts with **$/** resolves to your workflow’s own repository at the exact commit that is running, with no checkout required. It works everywhere the workspace-relative ./ syntax works, including workflow steps, composite action steps, nested composition, and reusable workflow calls._

.github/actions/run-e2e/action.yaml
```diff
@@ -126,10 +126,15 @@ runs:
         persist-credentials: false
     - name: Set up kubectl
       uses: azure/setup-kubectl@829323503d1be3d00ca8346e5391ca0b07a9ab0d # v5.1.0
+    - name: Get go version
+      id: go-version
+      uses: $/.github/actions/get-go-version
+      with:
+        dockerfile: ./target/Dockerfile
```
mihaitanasedt, line 133: Should it be `./target` or `./ref`? If we will update the image and run tests it will run against the old go version.

jokuniew: > If we will update the image and run tests it will run **against the old go version**.

### target
If we upgrade Dockerfile to 1.28.0 (for example) on some branch `foo`
i can run workflow against `foo` (**new** golang version) or default `main` (old golang version)

### ref
it will always look at Dockerfile from `foo` branch (the new version).

to be honest I think both scenarios are ok, but i would stick to target. Dockerfile is part of target the same as go.mod

avorima (approved):

mihaitanasedt (approved):

PR #7405, opened 2026-09-10
Don't inject istio placeholder containers

0sewa0:
## Description

[ICP-9122](https://dt-rnd.atlassian.net/browse/ICP-9122)

> [!NOTE]
> The ticket is only about documenting it and make the user exclude it manually. Which I will do but we could check for this.

The current check only looks for `auto` as a container image, so that we don't break anyone using https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#customizing-injection.

I tried to check anyone else doing this, but this trick seems to be istio only. 

So options:
- a) Close this and just document
- b) Leave as is
- c) Also check that the container name is `istio-proxy`, to be more targeted.

> [!WARNING]
>  The main issue with b) and c) is that `istio-proxy` as a non-native sidecar container is something we can monitor, and by doing this change we would make that impossible if the user will also want to configure it through this `image: auto` method.
> - Injection into istio native-sidecar support is whole other beast, as istio will always want its native sidecar init-container to be the first, which would break our injection every time if we attempted to inject into it.

## How can this be tested?
Follow https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#customizing-injection to customize an injected samples istio-proxy

[ICP-9122]: https://dt-rnd.atlassian.net/browse/ICP-9122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/webhook/mutation/pod/mutator/container.go
+++ b/pkg/webhook/mutation/pod/mutator/container.go
@@ -20,6 +20,6 @@ func checkInjectionAnnotation(annotations map[string]string, name string) bool {
 	return false
 }
 
-func IsContainerExcludedFromInjection(dkAnnotations, podAnnotations map[string]string, name string) bool {
-	return checkInjectionAnnotation(dkAnnotations, name) || checkInjectionAnnotation(podAnnotations, name)
+func IsContainerExcludedFromInjection(dkAnnotations, podAnnotations map[string]string, containerName string) bool {
+	return checkInjectionAnnotation(dkAnnotations, containerName) || checkInjectionAnnotation(podAnnotations, containerName)
 }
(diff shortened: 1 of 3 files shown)


0sewa0:
cc @andriisoldatenko 
> The main issue with b) and c) is that istio-proxy the non-native sidecar container is something we can monitor, and by doing this change we would make that impossible if the user will also want to configure it through this image: auto method.

Because of this I would stay with only documenting now and even in the future. (I did the PR just for curiosity)

PR #7408, opened 2026-09-11
Add helm unit test for Prometheus Operator CRDs

avorima:
## Description

Add helm unit test to assert the templating works.

ICP-8620

## How can this be tested?

Run helm unit tests


mihaitanasedt (commented):
Should we add the scenario enablePrometheus=true + installCRDs=false?

mihaitanasedt (approved):
lgtm

PR #7421, opened 2026-09-15
Add E2E test for KSPM, OTel and DBExecutor via fleet management

meibensteiner:
### Description
ICP-9503

When using a Gen3 tenant, the fleetManagement endpoint is being used for E2E tests. When using a Gen2 tenant we still use the pinned imageRef for KSPM, TelemetryIngest and DBExecutor.
For KSPM and OTel we additionally assert against the image reported in the DynaKube status.

### How can this be tested?
Run E2E tests run-in-k8s-phase3.


test/e2e/features/kspm/kspm.go
```diff
@@ -27,10 +31,14 @@ func Feature(t *testing.T) features.Feature {
 	options := []componentDynakube.Option{
 		componentDynakube.WithAPIURL(secretConfig.APIURL),
 		componentDynakube.WithKSPM(),
-		componentDynakube.WithKSPMImageRef(t, componentDynakube.GetLatestKSPMImageTagURI(t)),
 		componentDynakube.WithActiveGate(),
 	}
 
+	if !tenant.UsePhase3Tenant() {
```
waodim, line 37: What is the reason for the guard being only added to `Feature()` and not to `FeatureWithKubemon()` and `OptionalScoeps()`? Am I missing something? Same goes for the other files.

meibensteiner: @andriisoldatenko mentioned adding auto-update to one of the tests per component, so I only updated the "baseline" tests. Also, if the fleetmgmt endpoint has issues we can still rely on `FeatureWithKubemon()` to provide some coverage.

meibensteiner:
@avorima notified me that Gen2 tenants also serve fleetMgmt. I'll remove the Gen3 condition.

test/e2e/features/extensions/dbexecutor/db_executor.go
mihaitanasedt, line 1: Do we need a resolved image check, similar to the kspm and telemtryingest?

aorcholski (commented):
KSPM and TI tests PASS on my gen2 tenant

test/e2e/features/extensions/dbexecutor/db_executor.go
```diff
@@ -25,9 +25,7 @@ func Feature(t *testing.T) features.Feature {
 
 	options := []componentDynakube.Option{
 		componentDynakube.WithAPIURL(secretConfig.APIURL),
-		componentDynakube.WithExtensionsEECImageRef(t, componentDynakube.GetLatestEECImageTagURI(t)),
```
avorima, line 28: if you remove the images you need to set the feature flag for phase 2 tenants

test/e2e/features/extensions/dbexecutor/db_executor.go
meibensteiner: I wanted to, but extensions has no status type at all. Followup?

mihaitanasedt: AFAIK we don't want to keep the pattern with version in status. Maybe we can check the deployment image ?

meibensteiner: I dont know what to assert against, since its fleetMgmt.
I could ask the fleetmgmt endpoint in the e2e test and assert against what the operator deployed?

andriisoldatenko (approved):

PR #7425, opened 2026-09-15
Fix schema defaulting for duration fields

avorima:
## Description

Fix schema defaulting of target allocator and scraper duration fields.
The defaulting worked when using kubectl, because it just sees that the field is not set and applies the schema default.
When using the structs with a go client, the field actually is set (zero value is a valid value) so it's not defaulted and ends up as "0s".

ICP-9794

## How can this be tested?

Install the CRDs and create a PrometheusMonitoring resource using a go client

```
make manifests/crd/install
```

```go
package main

import (
	"context"

	"github.com/Dynatrace/dynatrace-operator/pkg/api/scheme"
	"github.com/Dynatrace/dynatrace-operator/pkg/api/v1alpha1/prometheusmonitoring"
	corev1 "k8s.io/api/core/v1"
	"sigs.k8s.io/controller-runtime/pkg/client/config"
)

func main() {
	pm := &prometheusmonitoring.PrometheusMonitoring{}
	pm.Name = "test"
	pm.Namespace = corev1.NamespaceDefault
	pm.Spec.DynaKubeRef = "test"
	c, _ := client.New(config.GetConfigOrDie(), client.Options{Scheme: scheme.Scheme})
	if err := c.Create(context.Background(), pm); err != nil {
		panic(err)
	}
}
```


aorcholski (approved):
LGTM
```
spec:
    targetsPollInterval: 60s
    scrapeInterval: 60s 
```
The default values set as expected.
Intervals are set to 0 for the main branch.

PR #7433, opened 2026-09-16
bump timeout for ocp-latest job only

andriisoldatenko:
## Description

We need to bump timeout
[Run in OpenShift latest (main)](https://github.com/Dynatrace/dynatrace-operator/actions/runs/35040490034/job/104619021893#logs)
cancelled 1 minute ago in 6h 4m 52s

image


.github/workflows/e2e-tests.yaml
```diff
@@ -251,6 +251,7 @@ jobs:
   run-in-ocp:
     if: ${{ github.event.inputs.run-e2e-in-ocp != 'false' }}
     name: Run in OpenShift latest (${{ github.event.inputs.target || 'main' }})
+    timeout-minutes: 480 # default is 360 (6h), max is 21600 (6 days)
```
mihaitanasedt: Where have you found that max is 21600? For self hosted runners i see no max in docs

andriisoldatenko: it's timeout of job not tomeout for token. Github kills job after 6h, 
ah you are right :) it's AI bs.

andriisoldatenko: we need to move few tests around.

mihaitanasedt (approved):

.github/workflows/e2e-tests.yaml
```diff
@@ -251,6 +251,7 @@ jobs:
   run-in-ocp:
     if: ${{ github.event.inputs.run-e2e-in-ocp != 'false' }}
     name: Run in OpenShift latest (${{ github.event.inputs.target || 'main' }})
+    timeout-minutes: 480 # default is 360 (6h)
```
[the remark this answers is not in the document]
mihaitanasedt: missleading, 360 is max for github runners. Self hosted runners can run jobs up to 5 days

PR #7462, opened 2026-09-22
Fix order of OTTL statements for workload attributes

avorima:
## Description

Fix detection of workload name and kind by changing the order so that higher order workloads, e.g. deployments, cronjobs are set before lower order kinds, e.g. replicaset, job.

This only affects standalone telemetryingest, because the metadata JSON annotation has higher precedence and contains the correct values.

ICP-10013

## How can this be tested?

Deploy telemetryingest and check data on tenant


andriisoldatenko:
can you please fix tests?

svfoxat (approved):

andriisoldatenko (approved):
don't forget to 🍒 pick

